PR #3, opened 2020-06-17
Fix a bunch of null data, bad type resulted bugs

shifucun:


main.py
```diff
@@ -145,7 +145,12 @@ def place():
     place_dcid = request.args.get('dcid')
     if not place_dcid:
         return redirect(url_for('place', dcid='geoId/0649670'))
-    place_type = get_property_value(place_dcid, 'typeOf')[0]
+    place_types = get_property_value(place_dcid, 'typeOf')
+    place_type = None
+    for t in place_types:
+        if not place_type or place_type.startswith('AdministrativeArea'):
```
beets, line 151: do you need another not here? or add parens to scope it

main.py
```diff
@@ -145,7 +145,12 @@ def place():
...
+            place_type = t
+    logging.info(place_type)
```
beets, line 153: remove before submitting

main.py
shifucun: this is for the first iteration when place_type is not set yet.

PR #5, opened 2020-06-19
draw line chart function with d3 library

shifucun:


static/js/chart/draw.ts
```diff
@@ -404,10 +410,117 @@ function drawLineChart(
...
+  let legendElem;
+  if (hasLegend) {
+    legendElem = createLegend(legendText);
+    parentElement.appendChild(legendElem);
+    let bbox = legendElem.getBoundingClientRect();
+    legendHeight = bbox.height;
+    parentElement.removeChild(legendElem);
+  }
+  height -= legendHeight;
+  const margin = { top: 10, right: 10, bottom: 30, left: 30 };
```
beets, line 447: Could you pull this out to a const, since we will be sharing it across charts.

static/js/chart/draw.ts
```diff
@@ -404,10 +410,117 @@ function drawLineChart(
...
+  // append the svg object to the body of the page
```
beets, line 448: We can remove the comments for this function now, they were meant as a placeholder / guide.

static/js/chart/draw.ts
shifucun: Done

static/js/chart/draw.ts
shifucun: Done

PR #9, opened 2020-06-19
Remove unused functions; Use new legend creation for line chart

shifucun:
demo: https://screenshot.googleplex.com/UnqD8sJbFpm

### DIFF
--- a/static/js/chart/draw.ts
+++ b/static/js/chart/draw.ts
@@ -272,32 +272,6 @@ function drawStackBarChart(
 }
 
 /**
- * Draw group bar chart.
- *
- * @param parentId
- * @param parentWidth
- * @param parentHeight
- * @param dataGroups
- * @param unit
- */
-function drawGroupBarChartOld(
-  parentId: string,
-  parentWidth: number,
-  parentHeight: number,
-  dataGroups: DataGroup[],
-  unit?: string
-) {
-  drawComplexBarChart(
-    parentId,
-    parentWidth,
-    parentHeight,
-    dataGroups,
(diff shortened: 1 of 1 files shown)


static/js/chart/draw.ts
```diff
@@ -616,16 +509,18 @@ function drawLineChart(
...
         .append("circle")
-          .attr("class", "dot")
-          .attr("cx", (d, i) => xScale(d[0]))
-          .attr("cy", d => yScale(d[1]))
-          .attr("fill", COLORS[i])
-          .attr("r", 3);
+        .attr("class", "dot")
+        .attr("cx", (d, i) => xScale(d[0]))
+        .attr("cy", (d) => yScale(d[1]))
+        .attr("fill", COLORS[i])
```
beets, line 515: the color here should match the color in the legend too (pull up the getColorfn). i'll update that later to make it work for the line charts too

static/js/chart/draw.ts
```diff
@@ -616,16 +509,18 @@ function drawLineChart(
         .data(dataset)
         .enter()
         .append("circle")
-          .attr("class", "dot")
-          .attr("cx", (d, i) => xScale(d[0]))
-          .attr("cy", d => yScale(d[1]))
-          .attr("fill", COLORS[i])
-          .attr("r", 3);
+        .attr("class", "dot")
```
beets, line 512: could you keep the indent for the .attr please?

shifucun: i think the git UI hides some lines in there... it has extra indents

PR #11, opened 2020-06-20
For parent places, prefer non AdministrativeArea for place type

shifucun:


main.py
```diff
@@ -413,6 +413,12 @@ def get_parent_place(dcid):
     payload = dc.send_request(url, req_json=req_json)
     parents = payload[dcid].get('out', [])
     parents.sort(key=lambda x: x['dcid'], reverse=True)
+    for i in range(len(parents)):
+        place_types = parents[i]['types']
+        parents[i]['types'] = list(
+            filter(lambda x: not x.startswith('AdministrativeArea'), place_types))
+        if not parents[i]['types']:
+            parents[i]['types'] = place_types
```
beets, line 421: Hm, I'm confused again. Is there another case where we would reach this line? As I read it, parents[i][types] would be empt only if it only contained "AdministrativeArea", but if you reassign place_types, it would be the unfiltered list anyway. Is that the intent? If that's the case, it would be clearer to filter only if the length of the list is greater than 1 (and you wouldn't need the temp variable).

shifucun: updated the logic to check types length

PR #17, opened 2020-06-23
Wrap long tick labels

beets:
Also removed some unused code


static/js/chart/draw.ts
```diff
@@ -72,17 +51,60 @@ function appendLegendElem(
...
+  });
+}
+
 function addXAxis(
   svg: d3.Selection<SVGElement, any, any, any>,
   height: number,
   xScale: d3.AxisScale<any>
 ) {
-  svg
+  var axis = svg
```
shifucun, line 97: nit: let

PR #24, opened 2020-06-23
Update chart colors

beets:
Loosely based on spectral, but with some light colors removed and with a
fix for 1 or 2 categories.

### DIFF
--- a/static/js/chart/draw.ts
+++ b/static/js/chart/draw.ts
@@ -29,13 +29,28 @@ const NUM_X_TICKS = 5;
 const NUM_Y_TICKS = 5;
 const MARGIN = { top: 20, right: 10, bottom: 30, left: 35, yAxis: 3 };
 
-function getColorFn(labels: string[]) {
-  let k = labels.length;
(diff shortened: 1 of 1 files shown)


static/js/chart/draw.ts
```diff
@@ -29,13 +29,28 @@ const NUM_X_TICKS = 5;
...
 const MARGIN = { top: 20, right: 10, bottom: 30, left: 35, yAxis: 3 };
 
-function getColorFn(labels: string[]) {
-  let k = labels.length;
-  k = k < 3 || k > 11 ? 10 : k;  // Spectral colors exist for k = 3 -> 11
-  return d3
-    .scaleOrdinal<string, string>()
-    .domain(labels)
-    .range(d3.quantize(d3.interpolateSpectral, k));
+function getColorFn(labels: string[]): any {
```
shifucun, line 32: change return type to "d3.scaleOrdinal<string, string>"?

beets:
sorry could you re-review? thanks!

PR #25, opened 2020-06-24
Various clean up and test update for svg drawing modules

shifucun:


static/js/chart/draw.test.ts
```diff
@@ -14,22 +14,15 @@
...
-  computeCoordinate,
 } from "./base";
 
 import {
-  drawSingleBarChart,
+  appendLegendElem,
+  getColorFn,
 } from "./draw";
 
-/*
```
beets, line 29: thanks for updating the test!

static/package.json
```diff
@@ -18,7 +18,7 @@
     "@svgdotjs/svg.js": "^3.0.16",
```
beets, line 18: we might not need svgdot anymore?

shifucun: right, i will remove in next PR

PR #27, opened 2020-06-24
Do not scroll the left menu column

shifucun:


templates/place_overview.html
```diff
@@ -87,33 +87,30 @@ <h1><a>DataCommons.org</a></h1>
...
-      <div class="col-2 d-none d-md-flex"></div>
-      <div class="col-12 col-md-5 order-last order-md-2">
-        <h1 id="place-name">{{place_name}}</h1>
-        <h3>
-          <span id="place-type" data-pt={{place_type}}>A {{place_type}} in</span>
-          <span id="place-parents">
-          </span>
-        </h3>
+    <div id="body-row" class="row">
+      <div class="col-md-2 mt-3 mt-md-4 mt-lg-5 sidebar-outer">
```
beets, line 91: It currently doesn't preserve the behavior of collapsing the sidebar when the window is narrow (in bootstrap's -sm mode)... for that, add these classes back: d-none d-md-flex

Also, remove the mt-* classes (those add margin-top at various breakpoints)

beets: Also, sidebar-outer should be an id (instead of a class name)

shifucun: made sidebar-outer a div and added back "d-none d-md-flex".

the mt-* class is to keep it aligned with the right section. wdyt?

static/js/place_overview.js
```diff
@@ -22,13 +22,27 @@ import { MainPane, Ranking, Menu, ParentPlace } from "./place_overview.jsx";
...
   renderPage(dcid);
   initAutocomplete();
+  document.addEventListener("scroll", adjustMenuPosition);
 };
 
+function adjustMenuPosition() {
+  let topicsEl = document.getElementById('topics');
+  if (window.scrollY > Y_SCROLL_LIMIT) {
+    topicsEl.classList.add("fixed");
+    topicsEl.style.top = "20px";
```
beets, line 39: Could you move this to the css so we don't forget to update this?

It should work to use:
#sidebar-outer.fixed { position: fixed; top: 20px }

shifucun: done

PR #29, opened 2020-06-25
Organize flask server into modules; This prepares splitting routes fr…

shifucun:
…om main.py


shifucun:
> thanks! this is a great start, i think we can split them down more so it's all functionally grouped, e.g. group kg, gni, chart api related code together.

Yep, that's my plan for today. Organize the routes as blueprints. Made some update on script. Can you take a look again. Thanks

PR #38, opened 2020-06-27
Dockerize the entire app and update steps to start server locally

shifucun:


shifucun:
PTAL, merged the conflict with master

Dockerfile
```diff
@@ -17,4 +17,4 @@ RUN python -m pytest
 
 # Run the web service on container startup.
 ENV FLASK_ENV="production"
-CMD exec gunicorn --bind :$PORT --workers 1 --threads 8 --timeout 0 main:app
\ No newline at end of file
+CMD exec gunicorn --bind :8080 --workers 1 --threads 8 --timeout 0 main:app
```
beets, line 20: Was this intentional (or should it be kept as $PORT)?

shifucun: As the PORT not need to change so I made it constant in the container and removed the PORT var in https://github.com/datacommonsorg/website/pull/38/commits/fca37637a0ee322592b471a417ff948a85bfcb19#diff-04c6e90faac2675aa89e2176d2eec7d8L60.

PR #47, opened 2020-07-02
Remove the oldest appengine version when auto push a new version to s…

shifucun:
…taging

### DIFF
--- a/cloudbuild.deploy.yaml
+++ b/cloudbuild.deploy.yaml
@@ -32,7 +32,10 @@ steps:
     cd server && \
     gcloud app deploy app_staging.yaml -q \
       --project=datcom-browser-staging \
-      --version=$SHORT_SHA
+      --version=$SHORT_SHA && \
+    gcloud app versions delete --project=datcom-browser-staging \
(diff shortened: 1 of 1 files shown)


cloudbuild.deploy.yaml
```diff
@@ -32,7 +32,10 @@ steps:
     cd server && \
     gcloud app deploy app_staging.yaml -q \
       --project=datcom-browser-staging \
-      --version=$SHORT_SHA
+      --version=$SHORT_SHA && \
+    gcloud app versions delete --project=datcom-browser-staging \
+    `gcloud app versions list --project=datcom-browser-staging | \
+    sed 's/  */:/g' | cut -f 2 -d : | tail -n +2 | head -1 | tr "\n" " "`
```
beets, line 38: maybe do the following updates:
tail -n +3 (or 4) so we keep more than 1 version around if there's only 2 versions remaining
head -n 1 for clarity

should we add a --sort-by to guarantee it's listed by ascending create date?

shifucun: Added number check and if check

shifucun:
@googlebot I signed it!

PR #53, opened 2020-07-07
Add redirects for tools

beets:
browser.dc.org/gni -> dc.org/tools/timeline
browser.dc.org/scatter -> dc.org/tools/scatter

### DIFF
--- a/server/routes/redirects.py
+++ b/server/routes/redirects.py
@@ -1,0 +1,50 @@
+# Copyright 2020 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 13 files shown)


server/routes/redirects.py
```diff
@@ -0,0 +1,50 @@
...
+bp = Blueprint(
+  "redirects",
+  __name__,
+)
+
+
+from flask import request
+@bp.route('/kg')
+def kg():
+    return redirect('https://datacommons.org/browser', code=301)
```
shifucun, line 31: do you want to use relative path? "redirect('/browser, code=301)"

If this code is deployed to staging or prod, the /gni and /kg page will not route

beets: Updated to use relative paths

beets:
Updated to use relative paths, ptal

PR #54, opened 2020-07-07
Display GNI link on landing page.

shifucun:


static/js/place_overview.jsx
```diff
@@ -538,6 +554,17 @@ class Chart extends Component {
     this.fetchData();
   }
 
+  buildTimelineToolUrl() {
+    let url = "/gni#&ptpv=";
```
beets, line 558: Could you add a TODO here for me to update this once the redirects go out?

PR #57, opened 2020-07-08
/kg -> /browser

beets:
* Move kg routes into a module
* Convert dcid get param to path args
* Add tests, update links sitewide
* Add back dc.org homepage
* Remove unused template


server/templates/browser/kg_main.html
```diff
@@ -0,0 +1,123 @@
...
+    <p>
+      The Data Commons Graph (DCG) is constructed by synthesizing a single Graph from many different data sources.
+      It currently contains data from Wikipedia, the US Census, NOAA, FBI, etc.
+      The browser can be used both to explore what data is available and to understand the graph structure, to help use the API.
+    </p>
+    <p>
+      You can use the search box above or start looking around from some well known starting points:
+    </p>
+    <ul>
+      <li class="type-list">
```
shifucun, line 35: Add a todo to pass a json from the route to here and use for loop for these lists?

server/templates/homepage.html
```diff
@@ -12,112 +12,68 @@
...
+      contribute.
+    </p>
 
-{% block content %}
-  <div id="homepage-content">
     <p>
-      The Data Commons Graph (DCG) is constructed by synthesizing a single Graph from many different data sources.
-      It currently contains data from Wikipedia, the US Census, NOAA, FBI, etc.
-      The browser can be used both to explore what data is available and to understand the graph structure, to help use the API.
+      You can use the <a href="https://browser.datacommons.org">Data Commons
```
shifucun, line 62: update this?

server/templates/homepage.html
```diff
@@ -12,112 +12,68 @@
...
+      Graph Browser</a> to browse the graph. The data can be
+      programmatically accessed via <a href="http://docs.datacommons.org/api">APIs</a>.
```
shifucun, line 64: will this link change too?

shifucun:
can comment "@googlebot I signed it!" so the cla check can pass

server/templates/browser/kg_main.html
beets: Done. Will probably keep the config in python

server/templates/homepage.html
beets: Will be deleted in the next PR

server/templates/homepage.html
beets: Not for now...

beets:
Made the suggested updates. PTAL!

PR #59, opened 2020-07-09
Use typescript for place page entry and react template

shifucun:


shifucun:
> Is the goal to convert everything to ts?

Yep, that's the plan, can do it gradually. Try to setup an example for Lijuan and Hao from this.

PR #61, opened 2020-07-10
Migrate most static pages to the new base template

beets:
* Also added /about
* Moved some static templates to the main template directory
* Some content/style cleanup

Diff should be against the branch after #60 is submitted (I created a branch from that PR). Is there a way to simplify the diff?


server/main.py
```diff
@@ -565,55 +565,61 @@ def mcf_playground():
...
 
 
 @app.route('/special_announcement/faq')
 def special_announcement_faq():
   return flask.render_template(
-      'factcheck/special_announcement_faq.html')
+      'special_announcement_faq.html')
+
+
+#
```
shifucun, line 605: do you think it worth creating new module under routes for these? Eventually we can make all these routes out of main.py

server/templates/about.html
```diff
@@ -0,0 +1,61 @@
...
+{% block head %}
+  <link rel="stylesheet" href={{url_for('static', filename='css/static.min.css', t=config['GAE_VERSION'])}} />
+{% endblock %}
+
+{% block content %}
+  <div class="container">
+    <h1>About Data Commons</h1>
+    <p>
+      Publicly available data from open sources (census.gov, NOAA, data.gov,
+      etc.) are vital resources for students and researchers in a variety of
```
shifucun, line 30: should we add a larger target users here? students and researches maybe a bit narrowed?

beets: Yes, probably. The content on that page could use a lot of work.. I just pulled out the content from our current homepage there for now.

server/main.py
beets: Moved to a new route. I agree with the goal, though would like to eventually enforce url_prefix for our blueprints so that it's easy to find the handlers in the future. One could argue that these pages which are at the root of the site (according to their url), are main pages. Maybe static.py isn't the best name for the module..

beets:
Address comments and rebased past #60 -- ptal.

server/tests/static_test.py
```diff
@@ -0,0 +1,64 @@
...
+
+    @patch('routes.factcheck.list_blobs')
+    def test_special_announcement(self, mock_list_blobs):
+        mock_list_blobs.side_effect = (lambda bucket, max_blobs : [])
+        response = app.test_client().get('/special_announcement')
+        assert response.status_code == 200
+        assert b"Covid-19 Special Announcements" in response.data
+
+
+    def test_about(self):
```
shifucun, line 61: also need mock list_blobs here?

beets: Only the special announcements page uses the blob (though the test name needed to be updated)

beets:
@shifucun Sorry, could you re-approve now that I've fixed all the codacy issues? Thanks!

PR #62, opened 2020-07-11
build the menu, update url upon click

qlj-lijuan:
- build the gni menu using react js
- add the function of updating the url once the statvar is selected

### DIFF
--- a/static/css/gni.css
+++ b/static/css/gni.css
@@ -322,7 +322,7 @@ body {
   font-size: 0.95rem;
   margin: 0 -24px;
   margin-left: 6px;
-  padding: 0px 24px 0px 6px;
+  padding: 0 24px 0 6px;
   border-top-right-radius: 6px;
   border-bottom-right-radius: 6px;
   /* border-left: 4px solid #ccc;  */
--- a/static/css/gni.css
+++ b/static/css/gni.css
@@ -291,7 +291,7 @@ body {
 }
 
 #drill li {
-  font-size: 1rem;
+  font-size: 1.5rem;
   display: block;
   margin: 0;
   margin-left: 6px;
@@ -305,8 +305,8 @@ body {
 }
 
 #drill li::before {
-  height: 6px;
-  width: 6px;
+  height: 10px;
+  width: 10px;
   content: "";
   display: inline-block;
   margin-left: -7px;
@@ -322,10 +322,10 @@ body {
   font-size: 0.95rem;
   margin: 0 -24px;
   margin-left: 6px;
-  padding: 3px 24px 3px 6px;
+  padding: 0px 24px 0px 6px;
   border-top-right-radius: 6px;
   border-bottom-right-radius: 6px;
-  border-left: 4px solid #ccc;
+  /* border-left: 4px solid #ccc;  */
 }
 
 .explore-menu-header a,
@@ -340,7 +340,7 @@ body {
   cursor: pointer;
   width: 14px;
   height: 14px;
-  margin-top: 4px;
+  margin-top: -6px;
   float: right;
   border: 0;
   background: url('../images/icon-checkbox.svg');
--- a/static/css/gni.css
+++ b/static/css/gni.css
@@ -291,7 +291,7 @@ body {
 }
 
 #drill li {
-  font-size: 1.5rem;
+  font-size: 1rem;
   display: block;
   margin: 0;
   margin-left: 6px;
@@ -305,8 +305,8 @@ body {
 }
 
 #drill li::before {
-  height: 10px;
-  width: 10px;
+  height: 6px;
+  width: 6px;
   content: "";
   display: inline-block;
   margin-left: -7px;
--- a/static/css/gni.css
+++ b/static/css/gni.css
@@ -354,6 +354,7 @@ body {
 
 .right-caret {
   transition: 250ms cubic-bezier(0.4, 0, 0.2, 1);
+  width: 12px;
 }
 
 .right-caret.transform-down {
--- a/static/js/dev.ts
+++ b/static/js/dev.ts
@@ -20,8 +20,8 @@
 
 import React from "react";
 import ReactDOM from "react-dom";
-import {page} from "./pv_tree";
-import {updateUrlStatsVar} from "./util.js";
+import { page } from "./pv_tree";
+import { updateUrlStatsVar } from "./util.js";
 import { DataPoint, DataGroup } from "./chart/base";
 
 import {
@@ -51,9 +51,10 @@ window.onload = function () {
   }
   const widthMenu = 600;
   let containerId = addChartContainer(widthMenu, height);
-  ReactDOM.render(React.createElement(page, {updateUrl:updateUrlStatsVar, search:false}),
-                  document.getElementById(containerId));
-  
+  ReactDOM.render(
+    React.createElement(page, { updateUrl: updateUrlStatsVar, search: false }),
+    document.getElementById(containerId)
+  );
 
   // Draw single bar chart.
   let dataPoints = [
@@ -62,7 +63,7 @@ window.onload = function () {
     new DataPoint("California", 3002342),
     new DataPoint("United States", 9520234),
   ];
-   containerId = addChartContainer(width, height);
+  containerId = addChartContainer(width, height);
   drawSingleBarChart(containerId, width, height, dataPoints);
 
   let dataGroups = [
@@ -92,8 +93,7 @@ window.onload = function () {
       new DataPoint("2013", 2000),
     ]),
   ];
-  
- //ReactDOM.render(React.createElement(page),document.getElementById(containerId));
+
   // Draw stack bar chart.
   containerId = addChartContainer(width, height);
   drawStackBarChart(containerId, width, height, dataGroups);
@@ -167,6 +167,4 @@ window.onload = function () {
   ];
   containerId = addChartContainer(width, height);
   drawLineChart(containerId, width, height, dataGroups);
-
-
 };
--- a/static/js/dev.ts
+++ b/static/js/dev.ts
@@ -215,30 +215,29 @@ window.onload = function () {
     new DataPoint("2018", 19453769),
   ]);
 
-  let dataGroupsDict = {
+  let dataGroupsDict_1 = {
     "geoId/05": [dataGroups_geo1_1, dataGroups_geo1_2],
     "geoId/06": [dataGroups_geo2_1, dataGroups_geo2_2],
   };
-
   containerId = addChartContainer(1000, 500);
-  drawGroupLineChart(containerId, 1000, 500, dataGroupsDict);
+  drawGroupLineChart(containerId, 1000, 500, dataGroupsDict_1);
 
-  dataGroupsDict = {
+  let dataGroupsDict_2 = {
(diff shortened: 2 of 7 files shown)


static/css/gni.css
```diff
@@ -291,7 +291,7 @@ body {
 }
```
shifucun, line 291: Can you make a screenshot on how the current GNI would look like with the css change? Trying not to make too big a difference for it.

static/js/dev.ts
```diff
@@ -44,6 +49,11 @@ window.onload = function () {
     chartElem.style.width = width + "px";
     return containerId;
   }
+  let width_menu = 600
```
shifucun, line 52: const widthMenu = 600;

Try to use const as much as possible; js style guide uses camelCase for variable and functions

static/js/pv_tree.tsx
```diff
@@ -0,0 +1,163 @@
...
+
+interface NodePropType {
+  title: string; 
+  selected: string;
+  expanded: string;
+  count: number;
+  children: NodeType[];
+  type: string;
+  argString: string;
+  updateurl: (string, boolean) => void;
```
shifucun, line 29: updateUrl

static/js/pv_tree.tsx
```diff
@@ -0,0 +1,163 @@
+import React, {Component} from "react";
+import hierarchy from "../../tools/pv_tree_generator/hierarchy_golden.json";
+//import hierarchy from "../data/hierarchy.json";
+
+// Fields of each node from json file
+interface NodeType {
+  show: string; 
+  title: string; 
+  selected: string;
```
shifucun, line 9: selected, expanded, placeTypes can be removed?

static/js/pv_tree.tsx
```diff
@@ -0,0 +1,163 @@
+import React, {Component} from "react";
+import hierarchy from "../../tools/pv_tree_generator/hierarchy_golden.json";
+//import hierarchy from "../data/hierarchy.json";
+
+// Fields of each node from json file
+interface NodeType {
```
shifucun, line 6: Can call this NodeData?

static/js/pv_tree.tsx
```diff
@@ -0,0 +1,163 @@
...
+          {child}
+        </li>
+      </ul>
+    )
+  }
+}
+
+interface MenuPropType{
+  search: boolean;
+  updateurl: (string, bool) => void;
```
shifucun, line 116: updateUrl

static/js/pv_tree.tsx
```diff
@@ -0,0 +1,163 @@
...
+  type: string; // Property or Value
+  argString: string;// for statvars, argString = dcid
+}
+
+interface NodePropType {
+  title: string; 
+  selected: string;
+  expanded: string;
+  count: number;
+  children: NodeType[];
```
shifucun, line 26: Not quite sure how react handles the the memory of the Props, but the Menu object seems very heavy as each node contains the data for its subtree. Can you think if it is possible to create the tree without passing the data along the way?

Try https://reactjs.org/docs/react-api.html#createelement which takes the children elements as the third argument...

static/js/pv_tree.tsx
```diff
@@ -0,0 +1,163 @@
...
+  render() {
+    let checkboxImg;
+    let expandImg;
+    let child; 
+    if (this.props.type === "value") {
+      checkboxImg = <button className={this.state.checked? 
+                                        "checkbox checked":"checkbox"}  
+                              onClick={this._handleCheckboxClick}/>};
+
+    if (this.state.expanded) {
```
shifucun, line 67: this is quite interesting, so you only create node when it is expanded, then the actual created nodes should be much less than before where we build the entire tree?

static/js/util.js
```diff
@@ -493,6 +494,38 @@ function randDomId() {
...
 
+/**
+ * add or delete statvars from url
+ *
+ * @param {string} dcid of statvar
+ * @param {boolean} add = True, delete = False
+ * @return {string} The line chart URL.
+ */
+
+function updateUrlStatsVar(statvar, add){
```
shifucun, line 505: this would be a good case to add unit test

shifucun:
Try run the test locally and fix lint error and the "Codacy" lint error.

static/js/pv_tree.tsx
shifucun: Thinking a bit more I guess what you do is okay. Even all nodes are expanded, the total of all the props object is similar to the original object size. And if it's true that you only create the Node that is expanded, then the actual props passed down is very small. Creating node on the fly is definitely a good improvement!

static/css/gni.css
qlj-lijuan: Here's the screenshot.
https://drive.google.com/file/d/18AKyoYPRmswHM4cU3jRgu9Wyv4bPmAT6/view?usp=sharing

static/js/dev.ts
qlj-lijuan: Got it. Thanks.

static/js/pv_tree.tsx
qlj-lijuan: done.

static/js/pv_tree.tsx
qlj-lijuan: We actually don't need this interface. Removed it.

static/js/pv_tree.tsx
qlj-lijuan: removed it.

static/js/util.js
qlj-lijuan: done.

static/js/pv_tree.tsx
qlj-lijuan: right.

static/js/pv_tree.tsx
```diff
@@ -0,0 +1,169 @@
+import React, { Component } from "react";
+import hierarchy from "../../tools/pv_tree_generator/hierarchy_golden.json";
+//import hierarchy from "../data/hierarchy.json";
+
+interface NodePropType {
+  title: string;
+  selected: string;
+  expanded: string;
```
shifucun, line 8: do think selected and expanded can be removed? It's okay to do in another PR if so, and also updated the hierachy.json

static/js/pv_tree.tsx
```diff
@@ -0,0 +1,169 @@
...
+    this._handleExpandClick = this._handleExpandClick.bind(this);
+    this.state = {
+      checked: props.selected === "yes" ? true : false,
+      expanded: props.selected === "yes" ? true : false,
+      nodePath: props.nodePath + "," + props.title,
+    };
+  }
+
+  render() {
+    let checkboxImg;
```
shifucun, line 36: good to have type like: let checkboxImg: HTMLInputElement for all the let varibles

static/js/util.test.js
```diff
@@ -0,0 +1,16 @@
+import { updateUrlStatsVar } from "./util.js";
+
+test("update Url statsvar", () => {
```
shifucun, line 3: nice!

static/js/dev.ts
```diff
@@ -82,7 +92,8 @@ window.onload = function () {
       new DataPoint("2013", 2000),
     ]),
   ];
-
+  
+ //ReactDOM.render(React.createElement(page),document.getElementById(containerId));
```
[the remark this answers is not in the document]
qlj-lijuan: done.

static/js/pv_tree.tsx
qlj-lijuan: I'll do it in another PR and update the hierarchy.json.

static/js/pv_tree.tsx
qlj-lijuan: done.

static/css/gni.css
```diff
@@ -322,10 +322,10 @@ body {
   font-size: 0.95rem;
   margin: 0 -24px;
   margin-left: 6px;
-  padding: 3px 24px 3px 6px;
+  padding: 0 24px 0 6px;
   border-top-right-radius: 6px;
   border-bottom-right-radius: 6px;
-  border-left: 4px solid #ccc;
+  /* border-left: 4px solid #ccc;  */
```
beets, line 328: Was this accidentally removed?

static/js/pv_tree.tsx
```diff
@@ -0,0 +1,169 @@
+import React, { Component } from "react";
+import hierarchy from "../../tools/pv_tree_generator/hierarchy_golden.json";
+//import hierarchy from "../data/hierarchy.json";
```
beets, line 3: Please delete if not needed anymore

static/js/pv_tree.tsx
```diff
@@ -0,0 +1,169 @@
...
+  nodePath: string;
+}
+
+class Node extends Component<NodePropType, NodeStateType> {
+  constructor(props: NodePropType) {
+    super(props);
+    this._handleCheckboxClick = this._handleCheckboxClick.bind(this);
+    this._handleExpandClick = this._handleExpandClick.bind(this);
+    this.state = {
+      checked: props.selected === "yes" ? true : false,
```
beets, line 29: This is minor, but it should work to simply do:

checked: props.selected === "yes"

static/js/pv_tree.tsx
```diff
@@ -0,0 +1,169 @@
...
+        />
+      );
+    }
+
+    if (this.state.expanded) {
+      expandImg = (
+        <img
+          className="right-caret transform-down"
+          src="../images/right-caret.png"
+          width="12px"
```
beets, line 53: Could you move this parameter to the stylesheet? Looks like if we apply it to the right-caret class, it wouldn't be needed below on line 80 as well.

static/js/util.js
```diff
@@ -493,6 +494,37 @@ function randDomId() {
...
 }
 
+/**
+ * add or delete statvars from url
+ *
+ * @param {string} dcid of statvar
+ * @param {boolean} add = True, delete = False
+ * @return {string} The line chart URL.
+ */
+
```
beets, line 504: Since this comment applies to the function, the blank line isn't necessary.

static/js/util.js
```diff
@@ -493,6 +494,37 @@ function randDomId() {
...
+function updateUrlStatsVar(statvar, add) {
```
beets, line 505: Minor, but if you rename add to should_add or should_be_added or is_statvar_added, then it's very clear it's a boolean, and line 511 can be very clearly read as: if (is_statvar_added) { ... }

static/css/gni.css
qlj-lijuan: This is removed deliberately, because otherwise there're two solid lines on the left for each level. I'm no sure if there's a better way to get rid of those lines.

static/js/pv_tree.tsx
qlj-lijuan: done.

static/js/pv_tree.tsx
qlj-lijuan: done.

static/js/pv_tree.tsx
qlj-lijuan: done.

static/js/util.js
qlj-lijuan: got it.

static/js/util.js
qlj-lijuan: done.

PR #64, opened 2020-07-14
Update styling on the scatter tool

beets:
* Migrate from jquery 1.11 to the base version from base.html (3.5.1)
* Remove jquery.ui
* Cleaned up the plot selection form
* Cleaned up dialog
* Added welcome message

image

### DIFF
--- a/server/templates/tools/scatter.html
+++ b/server/templates/tools/scatter.html
@@ -120,10 +120,9 @@
         <div class="description">
           <div id="sd-y"></div>
         </div>
-      </div>
-
-      <div class="block_container">
-        <div id="dropped"></div>
+        <div class="description">
+          <div id="dropped"></div>
+        </div>
       </div>
 
     </div>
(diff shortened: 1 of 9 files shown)


server/templates/tools/scatter.html
```diff
@@ -117,8 +130,54 @@
...
-{% endblock %}
\ No newline at end of file
+
+<div id='screen'>
+  <div id="spinner"></div>
+</div>
+{% endblock %}
+
+{% block footer %}
+<script src="https://code.jquery.com/ui/1.11.1/jquery-ui.min.js"></script>
```
shifucun, line 176: do you want to remove this from the PR description?

shifucun:
Nice to fix the Codacy error as well. Wonder if there is a way to run the Codacy lint locally

server/templates/tools/scatter.html
beets: Removed from code - it's not required anymore, required elements are taken from bootstrap, which is included in base.html

beets:
Updated to also fix codacy warning, removed the Filter button, updated form validation display and some bugs as I was testing. PTAL

PR #67, opened 2020-07-14
Homepage and header updates

beets:
* Added COVID SA to Homepage's What's new
* Update searchbox
* Update header styles to match font size with tools pages

image
image


beets:
@shifucun could I get another review post merge? Thanks!

PR #76, opened 2020-07-15
Redesign place

beets:
Update place pages
* Also fixed sidebar scroll issue
* Layout updates for different window widths
* Shrink "brand" link in header
* Default to USA page instead of Mountain View
* Fixed fixed-sidebar scroll sometimes overflowing into the footer. Pin it to the main chart container.

image
image
image


beets:
This is the last change in the branch before I merge back in to master.

static/js/place.ts
```diff
@@ -30,20 +30,55 @@ import {
...
+  updatePageLayoutState();
+  maybeToggleFixedSidebar();
+  window.onresize = maybeToggleFixedSidebar;
 };
 
+/**
+ *  Make adjustments to sidebar scroll state based on the content.
+ */
+function updatePageLayoutState() {
+  Y_SCROLL_LIMIT = document.getElementById('main-pane').offsetTop;
```
shifucun, line 52: if Y_SCROLL_LIMIT not constant then should not be all caps?

PR #81, opened 2020-07-16
Add a bunch of more charts for international countries

shifucun:


server/chart_config.json
```diff
@@ -136,11 +150,25 @@
             "statsVars": [
               "Count_Person_Upto5Years",
               "Count_Person_5To17Years",
+              "Count_Person_15To19Years",
```
beets, line 153: FYI, Our charts aren't going to look good with so many labels. We'll have to aggregate some of these like we do for GNI.

beets: I know this will leave gaps, but can we omit these new ones for now? Ending at 65+ seems reasonable for the long term as well.

server/chart_config.json
```diff
@@ -153,11 +181,25 @@
...
```
beets, line 184: ditto

server/chart_config.json
shifucun: Done

server/chart_config.json
shifucun: Done

server/chart_config.json
shifucun: Note US won't use the newly added age group since census does not have these ages. 
Even it is a large group, US and EU only use a fraction of it separately.

server/chart_config.json
shifucun: this is how it looks like: https://screenshot.googleplex.com/UKW2QQGQOTj

we do need to fix too many years though...

beets: iirc, census has all of these, we just left a bunch out from these charts (you'll notice that it's not actually a complete range, 18-24 is missing.

image

beets: Let's see how it looks like across countries. I suspect we'll have to remove most of these new categories quickly..

shifucun: removed two more... we can trim afterwards..

shifucun: can you stamp again?

PR #89, opened 2020-07-16
Consolidate commands to run tests into the script

beets:
* add cmd flags to to support the various options

No rush on this, Bo. Just wanted to get this out, but can merge when we have more time to triage issues that come up.
Also, how can I test the cloudbuild yamls?

### DIFF
--- a/run_test.sh
+++ b/run_test.sh
@@ -51,7 +51,7 @@ function run_all_tests {
   run_py_test
   run_npm_build
   run_npm_lint
-  run_py_test
+  run_npm_test
 }
 
 function help {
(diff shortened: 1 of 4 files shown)


run_test.sh
```diff
@@ -13,15 +15,83 @@
...
+  pip3 install -r requirements.txt -q
+  python3 -m pytest
+  cd ..
+}
+
+function run_all_tests {
+  run_py_test
+  run_npm_build
+  run_npm_lint
+  run_py_test
```
shifucun, line 54: this should be run_npm_test

beets:
Tested locally and added some clarifications. PTAL @shifucun when you get the chance - thanks!

PR #99, opened 2020-07-16
Add api to get place names

shifucun:


shifucun:
> add a todo for a test?

Added test :)

shifucun:
> thanks for adding the test - it was incredibly useful while i made changes to all the pages. it caught quite a few things in development!

Totally agree!

PR #107, opened 2020-07-17
move timeline and scatter related ts/tsx/js files into directory of js/tools

qlj-lijuan:


static/js/tools/menu.js
```diff
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-import { dcidToPlaceType, getUrlVars, setSearchParam } from "./dc.js";
+import { dcidToPlaceType, getUrlVars, setSearchParam } from "../dc.js";
```
shifucun, line 17: how about move dc.js, util.js and scatter.js also to this folder. As we refactor, eventually dc.js can just be removed.

beets: good call, i forgot those were only used by the gni tool

qlj-lijuan: I moved dc.js, scatter.js into this folder, and add timeline related utility functions into a new file tools/timeline_util.js

PR #112, opened 2020-07-17
Add triples of statsvar as state of component Page

qlj-lijuan:


static/js/tools/timeline_page.tsx
```diff
@@ -15,28 +15,29 @@
  */
 
 import React, { Component } from "react";
-import { parseStatVarPath, parsePlace } from "./timeline_util";
+import { parseStatVarPath, parsePlace, getTriples } from "./timeline_util";
```
shifucun, line 18: how about call this getStatsVarProp and below too. As the object is not in triples format..

static/js/tools/timeline_page.tsx
```diff
@@ -47,6 +48,8 @@ class Page extends Component<PagePropType, PageStateType> {
   }
 
   handleHashChange() {
+    const svPaths = parseStatVarPath()[0];
+    const svIds = parseStatVarPath()[1];
     const placesPromise = parsePlace();
```
shifucun, line 53: before call parsePlace and getTriples, can compare if the current places, statsvars are the same as before, if the same, then no need to call them.

Also in case both of them need to be called, can use Promise.All to wait them all finished, then just setState once.

static/js/tools/timeline_page.tsx

static/js/tools/timeline_util.js
```diff
@@ -104,20 +106,42 @@ function parseStatVarPath() {
  */
 function parsePlace() {
```
shifucun, line 107: nice to split the function out

static/js/tools/timeline_page.tsx
qlj-lijuan: Added the comparison before call parsePlace and getTriples.
Any specific reason to use Promise.All ?

shifucun: The way you do here is fine. Since the place and stats var won't change in the same time.

PR #113, opened 2020-07-17
Expand child place types and re-name sitemap to place list

shifucun:


server/templates/base.html
```diff
@@ -132,7 +132,7 @@ <h6>Data Commons</h6>
               <a href="#">About Data Commons</a>
               <a href="{{ url_for('static.faq') }}">Frequently Asked Questions</a>
               <a href="http://github.com/datacommonsorg">Github Repository</a>
-              <a href="{{url_for('sitemap.index')}}">Sitemap</a>
+              <a href="{{url_for('placelist.index')}}">Place List</a>
```
beets, line 135: Could you remove this link actually? I think we wanted to both rename and remove it from this list.

server/templates/placelist.html
```diff
@@ -15,22 +15,22 @@
...
+{% set title = 'Place List' %}
 
 {% block head %}
   <link rel="stylesheet" href={{url_for('static', filename='css/static.min.css', t=config['GAE_VERSION'])}} />
 {% endblock %}
 
 {% block content %}
   <div class="container">
-    <h1>Sitemap</h1>
+    <h1>Place List</h1>
```
beets, line 27: Nit: maybe call this "Places in the Data Commons Graph"

server/templates/base.html
shifucun: Done

server/templates/placelist.html
shifucun: Done

PR #116, opened 2020-07-18
Match the style of the new timeline page to the old one.

qlj-lijuan:
Revised the menu and search part. 
image

### DIFF
--- a/static/css/dev_menu.scss
+++ b/static/css/dev_menu.scss
@@ -25,3 +25,16 @@
 #chart-region {
   margin-left: 400px;
 }
+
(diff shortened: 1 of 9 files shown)


static/css/dev_menu.scss
```diff
@@ -24,4 +24,17 @@
 
 #chart-region {
   margin-left: 400px;
+}
+
+#drill .noborder{
```
shifucun, line 29: Are these two classes used to remove the extra line?

qlj-lijuan: Yes.

PR #119, opened 2020-07-18
Complete event propagation from Page to ChartRegion

qlj-lijuan:
- pass perCapita to ChartRegion
- add the deleteStatVar function of ChartRegion
- resolve the issue that the last place chip cannot be deleted

### DIFF
--- a/static/js/tools/timeline_chart.tsx
+++ b/static/js/tools/timeline_chart.tsx
@@ -16,6 +16,7 @@
 
 import React, { Component } from "react";
 import { randDomId } from "../util";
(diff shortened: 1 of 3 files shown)


qlj-lijuan:
> very nice. Try to add/remove button in menu, search chip, chart chip (so 6 combination) and make sure the chart can update.
> 
> There is a case when a place is selected but not stats_var, we will set the stats var to be "Count_Person". Can do this in following PR

The buttons work wells. There's an issue that when all place chips are deleted, there would be an "page unresponsive" error.

PR #120, opened 2020-07-18
set default statvar = Count_Person when place is defined but statvar is empty

qlj-lijuan:
…is not defined


shifucun:
plz fix the tests

qlj-lijuan:
> plz fix the tests

fixed

PR #122, opened 2020-07-19
Fix some unpaired link tag

shifucun:


beets:
Technically, closing the tag will keep it xml compliant. Either is valid though:

PR #123, opened 2020-07-19
add static component when the page loads

qlj-lijuan:
- add the static component when the page loads, replace the old url with the new ones.
- in the url, join the node path by " ' " instead of " , " , because " , " is used in the income range such as "Less than 9,999 $"

### DIFF
--- a/static/js/tools/statsvar_menu.tsx
+++ b/static/js/tools/statsvar_menu.tsx
@@ -1,7 +1,7 @@
 import React, { Component } from "react";
 import hierarchy from "../../../tools/pv_tree_generator/hierarchy.json";
 
-export const joinPath="'";
+export const SEP="'";
 
 interface NodePropType {
   title: string;
@@ -30,7 +30,7 @@ class Node extends Component<NodePropType, NodeStateType> {
     this.state = {
       checked: false,
       expanded: false,
-      nodePath: props.nodePath + joinPath + props.title,
+      nodePath: props.nodePath + SEP + props.title,
       svPaths: [[]],
     };
   }
--- a/static/js/tools/statsvar_menu.tsx
+++ b/static/js/tools/statsvar_menu.tsx
@@ -28,7 +28,7 @@ class Node extends Component<NodePropType, NodeStateType> {
     this.state = {
       checked: false,
       expanded: false,
-      nodePath: props.nodePath + "," + props.title,
+      nodePath: props.nodePath + "'" + props.title,
       svPaths: [[]],
     };
   }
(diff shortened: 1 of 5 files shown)


static/js/tools/timeline_page.tsx
```diff
@@ -125,6 +126,7 @@ class Page extends Component<PagePropType, PageStateType> {
             <div id="search">
               <SearchBar places={this.state.places} />
             </div>
+            {(window.location.hash === "") && <Info/>}
```
shifucun, line 129: check this.state.places.length === 0? There is the case when stats var is selected but place are removed

static/js/tools/statsvar_menu.tsx
```diff
@@ -28,7 +28,7 @@ class Node extends Component<NodePropType, NodeStateType> {
     this.state = {
       checked: false,
       expanded: false,
-      nodePath: props.nodePath + "," + props.title,
+      nodePath: props.nodePath + "'" + props.title,
```
shifucun, line 31: can you make this a constant? like const SEP = "'". May need to change it in the future if we have ' in the string..

static/js/tools/timeline_page.tsx

static/js/tools/statsvar_menu.tsx

static/js/tools/timeline_util.test.ts
```diff
@@ -22,20 +23,24 @@ test("update Url statsvar", () => {
 });
 
 test("parse statvar from Url", () => {
-  window.location.hash = "#&statsvar=dc/test,Demo,prop";
+  window.location.hash = "#&statsvar=dc/test'Demo'prop";
```
shifucun, line 26: also need to change here

static/js/tools/timeline_info.tsx
```diff
@@ -0,0 +1,56 @@
...
+                    find data about places (zip codes, cities, counties, states) from a variety of
+                    sources, including the Census, FBI, Bureau of Labor Statistics, CDC and others. </p>
+                <p> Enter a place in the search box above and then pick one or more of the variables in
+                    the pane to the left. There are thousands of variables to choose from, arranged in a
+                    topical hierarchy. </p>
+                <p> Or you can start your exploration from these interesting points ...</p>
+                <ul>
+                    <li>
+                        <b>University towns</b> by
+                        <a href="#&place=geoId/0606000,geoId/2511000,geoId/2603000,geoId/1777005,geoId/1225175,geoId/4815976&statsvar=MedianAge'Demographics'Median%20age"> age</a>,
```
shifucun, line 18: please also change here. As this is used a lot, can think of a short name for it (2 or 3 letters)

static/js/tools/timeline_util.test.ts
qlj-lijuan: done

static/js/tools/timeline_info.tsx
qlj-lijuan: done

PR #133, opened 2020-07-20
add the small JSON file of top level only

qlj-lijuan:
- add the small JSON file hierarchy_top.json (5KB)
- copy the complete json file into static/data/hierarchy_statsvar.json
- load the complete json in ComponentDidMount

- correct the node path of default statsvar Count_Person


static/js/tools/statsvar_menu.tsx
```diff
@@ -182,6 +191,13 @@ class Menu extends Component<MenuPropType, {}> {
       </div>
     );
   }
+  componentDidMount(){
+    $.getJSON(jsonPath, function(data){
```
shifucun, line 195: Can also use axios for consistency. the can remove jquery from package. json

qlj-lijuan: done.

qlj-lijuan:
> Can you also try this with an url that has a few stats vars selected?

Looks good to me. Do you have any concern?

Issue #88, opened 2020-07-16
[BUG] Need to remove column in rankings for states


rvguha:
See https://staging.datacommons.org/place?dcid=geoId%2F06

PR #134, opened 2020-07-20
Update UI for ranking and continent

shifucun:
- Do not add link to continent.
- Do not show map and ranking for non-US places.
- Do not include ranking for continent.

### DIFF
--- a/static/js/place_template.tsx
+++ b/static/js/place_template.tsx
@@ -378,23 +378,26 @@ interface OverviewPropType {
 
 class Overview extends Component<OverviewPropType, {}> {
   render() {
-    return (
-      <React.Fragment>
-        <section className="factoid col-12">
-          <div className="row">
-            <div className="col-12 col-md-4">
-              <Map dcid={this.props.dcid}></Map>
+    if (!this.props.topic) {
+      return (
+        <React.Fragment>
+          <section className="factoid col-12">
+            <div className="row">
+              <div className="col-12 col-md-4">
+                <Map dcid={this.props.dcid}></Map>
+              </div>
+              <div className="col-12 col-md-8">
+                <Ranking
+                  dcid={this.props.dcid}
+                  data={{ label: [], Population: [] }}
+                ></Ranking>
+              </div>
             </div>
-            <div className="col-12 col-md-8">
-              <Ranking
-                dcid={this.props.dcid}
-                data={{ label: [], Population: [] }}
-              ></Ranking>
-            </div>
-          </div>
-        </section>
-      </React.Fragment>
-    );
+          </section>
+        </React.Fragment>
+      );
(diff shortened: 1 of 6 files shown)


static/js/place_template.tsx
```diff
@@ -82,15 +82,19 @@ interface ChartCategory {
...
 }
 
 class ParentPlace extends Component<ParentPlacePropsType, {}> {
-  constructor(props) {
+  constructor(props: ParentPlacePropsType) {
     super(props);
   }
   render() {
     const num = this.props.parentPlaces.length;
+    if (this.props.placeType === "Country" && num === 1) {
```
beets, line 95: Can we just unlink the last parent place on line 101? We need to unlink it everywhere a continent is displayed, not just on country pages.

shifucun: Now checking this by the node type

shifucun:
> To really fix #88, we'll need to drop columns in the ranking table whenever a column is empty. If all columns are empty, then drop the table.
> https://staging.datacommons.org/place?dcid=geoId/0667000

Move data fetch of ranking inside component and only render it when there are ranking data

static/js/place_template.tsx
```diff
@@ -121,53 +121,80 @@ class ParentPlace extends Component<ParentPlacePropsType, {}> {
...
+    label: string[];
+    Population: { name: {}; label: string }[];
+  };
+}
+
+class Ranking extends Component<RankingPropsType, RankingStateType> {
   constructor(props) {
     super(props);
+    this.state = {
+      data: this.props.data,
```
beets, line 142: I'm quite confused about this. Which data is this.props.data (RankingPropsType or RankingStateType)? And why is it set here when it will get filled in by the request?

shifucun: this is RankingStateType.

the sequence of action will be constructor() -> render() -> componentDidMount()

after constructor(), render() will kick in and use this.state.data in line 148, finding the length ===0 and renders empty component. Then in componentDidMount(), it fetches data and set state.data. As state is updated, render() is called again to refresh the page.

This is kinda a typical patter where data needs to be async obtained to render the content. Effectively the same as how it is before in place.ts, where we call "axios.get().then(react.Render())"

shifucun:
Updated test. Can you stamp again? thx!

static/js/place_template.tsx
beets: Got it, thanks for the explanation!

Hm, why not explicitly start it off with a blank dictionary then? Unless it's to get the right type info...

Btw, then the data property in RankingPropsType is not used anymore and should be removed.

shifucun: Good point. Remove "data" in props.

Issue #92, opened 2020-07-16
[BUG] DCID's for genomes missing from search


beets:
### Describe the bug
Missing results for ENCODE related searches.

@spiekos could you add some examples here that don't work?

spiekos:
Search on the dcid no longer works. This leads to no results being returned when you search for things like:
CHEMBL102202
TP63
ChemicalCompound

These previously would return the correct nodes under the old method in which the search operated only on the name property of entities.

beets:
@spiekos the old search functionality will be available again at datacommons.org/search_dc after our next push to prod. Sorry for the trouble!

PR #152, opened 2020-07-22
Create sitemap for places and add robots.txt

shifucun:
The setup is based on: https://www.sitemaps.org/protocol.html


server/templates/placelist.html
```diff
@@ -32,7 +32,7 @@ <h3>{{ key }}</h3> <div style="column-count: 3">
             <div>
               <a href="{{url_for('placelist.node', dcid=place['dcid'])}}">{{place['name']}}</a>
               <span>: (</span>
-              <a href="{{url_for('browser.kg_entity', dcid=dcid)}}">KG</a>
```
beets, line 35: oops - thanks for fixing that!

static/robots.txt
```diff
@@ -0,0 +1,17 @@
+Sitemap: https://datacommons.org/CongressionalDistrict.0.txt
```
beets: can we serve these from datacommons.org/sitemap? or do they need to be served from the root?

static/sitemap/Country.0.txt
```diff
@@ -0,0 +1,267 @@
...
+https://datacommons.org/place?dcid=country/GUY
+https://datacommons.org/place?dcid=country/MSR
+https://datacommons.org/place?dcid=country/BYS
+https://datacommons.org/place?dcid=country/SDN
+https://datacommons.org/place?dcid=country/DJI
+https://datacommons.org/place?dcid=country/GHA
+https://datacommons.org/place?dcid=country/GNQ
+https://datacommons.org/place?dcid=country/HVO
+https://datacommons.org/place?dcid=country/PHL
+https://datacommons.org/place?dcid=iso/NL
```
beets, line 12: oh interesting:
https://datacommons.org/browser/iso/NL works
https://datacommons.org/place?dcid=iso/NL fails

searching for netherlands though, resolves to:
https://datacommons.org/place?dcid=country%2FNLD

looks like this could be an import error @tjann ?

shifucun:
> nice! could you add a quick README as well?
> 
> btw, we might need to do it for all dcid's too: see #92

Added README. okay, i will add other nodes in follow up PR. Hopefully to check this setup can work properly first.

static/robots.txt
shifucun: Yeah, that was an error. updated to server from /sitemap

static/sitemap/Country.0.txt
shifucun: Added a fix in https://github.com/datacommonsorg/website/pull/153

PR #154, opened 2020-07-23
Wrap text for gni in-chart legend

shifucun:
image

image

image

### DIFF
--- a/static/js/chart/draw.ts
+++ b/static/js/chart/draw.ts
@@ -104,6 +104,7 @@ function wrap(
     const text = d3.select(this);
     const words = text.text().split(/\s+/).reverse();
     let line: string[] = [];
+    let lineNumber = 0;
     const lineHeight = 1.1; // ems
     const y = text.attr("y");
     const dy = parseFloat(text.attr("dy"));
@@ -126,7 +127,7 @@ function wrap(
           .append("tspan")
           .attr("x", 0)
           .attr("y", y)
-          .attr("dy", lineHeight + dy + "em")
+          .attr("dy", ++lineNumber * lineHeight + dy + "em")
           .text(word);
       }
       word = words.pop();
@@ -147,7 +148,9 @@ function addXAxis(
     .call((g) => g.select(".domain").remove());
 
   if (typeof xScale.bandwidth === "function") {
-    axis.selectAll(".tick text").call(wrap, xScale.bandwidth());
+    axis.selectAll(".tick text")
+    .attr("dy", "0em")
+    .call(wrap, xScale.bandwidth());
   }
 }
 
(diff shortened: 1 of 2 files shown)


static/js/chart/draw.ts
```diff
@@ -130,7 +126,7 @@ function wrap(
           .append("tspan")
           .attr("x", 0)
           .attr("y", y)
-          .attr("dy", ++lineNumber * lineHeight + dy + "em")
```
beets, line 133: This was added for long text x-axis ticks. Do the charts on the place pages and dev page look ok when set to narrow widths?

static/js/chart/draw.ts
```diff
@@ -556,9 +552,13 @@ function drawGroupLineChart(
   unit?: string
 ) {
   let dataGroups = Object.values(dataGroupsDict)[0];
+  const dashInLegend = Object.keys(plotParams.dashes).length !== 1;
+  const legenTextdWidth = Math.max(width * LEGEND.ratio, LEGEND.minTextWidth);
```
beets, line 556: typo: misplaced little "d" in the var name (legendTextWidth)

static/js/chart/draw.ts
```diff
@@ -629,46 +625,54 @@ function drawGroupLineChart(
...
  * @param plotParams: It contains all colors and dashes for geoIds and statVars.
  */
-function buildInChartLegend(id: string, plotParams: PlotParams) {
-  const legend = d3.select("#" + id);
-  if (Object.keys(plotParams.dashes).length === 1) {
+function buildInChartLegend(
+  legend: d3.Selection<SVGGElement, any, any, any>,
+  plotParams: PlotParams,
+  dashInLegend: boolean,
+  legenTextdWidth: number
```
beets, line 632: same typo

static/js/chart/draw.ts
```diff
@@ -629,46 +625,54 @@ function drawGroupLineChart(
...
-        .text(plotParams.title[statsVars[i]])
+        .attr("transform", `translate(0, ${yOffset})`)
+        .attr("text-anchor", "start")
+        .attr("dy", ".3em")
+        .attr("fill", `${plotParams.colors[statsVar]}`)
         .style("font-size", "14")
-        .attr("fill", `${plotParams.colors[statsVars[i]]}`);
+        .text(plotParams.title[statsVar])
+        .call(wrap, legenTextdWidth);
+      yOffset += text.node().getBBox().height + LEGEND.lineMargin;
```
beets, line 648: we should be careful with getbbox, since it wont work if the chart isn't attached to the dom. any issues on the place page?

static/js/chart/draw.ts
shifucun: revert this back. I need to add attr("y") for the legend text to work with this

static/js/chart/draw.ts
shifucun: done

static/js/chart/draw.ts
shifucun: done

static/js/chart/draw.ts
shifucun: yeah, this works fine; and this is only used in gni page. I wonder if we need to do the same thing for landing page as now there is some text cut.

PR #156, opened 2020-07-23
Place page: do not show topic on menu if there are no sub-topic charts

shifucun:
image


shifucun:
@googlebot I signed it!

PR #158, opened 2020-07-23
Remove the category if there's no child

qlj-lijuan:


static/js/tools/statsvar_menu.tsx
```diff
@@ -152,22 +158,46 @@ class Node extends Component<NodePropType, NodeStateType> {
...
-    let childCnt = 0;
-    if (this.props.cd && this.props.cd.length !== 0) {
+    if (this.props.t !== "c") {
+      return this.childCnt(this.props.cd) !== 0;
+    } else {
+      // check two levels for category nodes
+      let childCnt = 0;
       this.props.cd.map((item) => {
+        let cnt = 0;
+        if (
```
shifucun, line 168: can you add some comments on the logic here? New people  will be able to follow up on it

qlj-lijuan: Sure.

PR #162, opened 2020-07-23
add super enum for crime types

qlj-lijuan:


static/data/hierarchy_statsvar.json
```diff
@@ -6,7 +6,7 @@
   "c": 1784,
   "cd": [
    {
-    "sv": "Count_Person",
+    "sv": "TotalPopulation",
```
shifucun: Note this will change back to Count_Person very soon. I will let you know then, so need to generate the json again.

qlj-lijuan: Cool. For now, I added one line to convert it.

PR #163, opened 2020-07-23
Use all place types in contained in (child) place relation

shifucun:
Right now for containing places, only one type is selected. When all child places are grouped by the type, child place are grouped randomly. This fix the bug.

### DIFF
--- a/static/webpack.config.js
+++ b/static/webpack.config.js
@@ -21,12 +21,11 @@ const FixStyleOnlyEntriesPlugin = require("webpack-fix-style-only-entries");
 const config = {
   entry: {
     dev: [__dirname + "/js/dev.ts", __dirname + "/css/dev.scss"],
-    dev_menu: __dirname + "/js/tools/dev_menu.ts",
-    download: __dirname + "/js/tools/download.js",
     dev_menu: [
       __dirname + "/js/tools/dev_menu.ts",
       __dirname + "/css/dev_menu.scss",
     ],
+    download: __dirname + "/js/tools/download.js",
     kg: [__dirname + "/js/browser/kg.js", __dirname + "/css/kg.scss"],
     mcf_playground: __dirname + "/js/mcf_playground.js",
     place: [
--- a/server/routes/api/place.py
+++ b/server/routes/api/place.py
@@ -143,7 +143,6 @@ def child_fetch(dcid):
                     'dcid': place['dcid'],
                     'pop': pop.get(place['dcid'], 0)
                 })
-                break
     return result
 
 
--- a/static/js/browser/kg.js
+++ b/static/js/browser/kg.js
@@ -677,9 +677,6 @@ function renderKGPage(
     (t) => t["objectId"] == dcid && t["subjectId"] != dcid
   );
   inArcs = inArcs.map((t) => {
-    if (t["subjectTypes"]) {
-      t["subjectType"] = t["subjectTypes"][0];
-    }
     t["dcid"] = t["subjectId"];
     t["prov"] = t["provenanceId"];
     t["src"] = provDomain[t["provenanceId"]];
@@ -694,10 +691,20 @@ function renderKGPage(
   // Do not show observation in in-arcs as they are handled later.
   inArcs = inArcs.filter((t) => t["subjectType"] != "Observation");
 
-  const inArcsGroup = _.groupBy(inArcs, (arc) => arc["subjectType"]);
+  let inArcsGroup = {};
+  for (let inArc of inArcs) {
+    for (let sType of inArc["subjectTypes"]) {
+      if (!(sType in inArcsGroup)) {
+        inArcsGroup[sType] = [];
+      }
+      inArcsGroup[sType].push(inArc);
+    }
+  }
 
   const inArcsGroupsElem = document.getElementById("in-arcs-groups");
-  for (let parentType in inArcsGroup) {
+  let sortedTypes = Object.keys(inArcsGroup);
+  sortedTypes.sort();
+  for (const parentType of sortedTypes) {
     if (util.isPopulation(parentType)) continue;
     const nodesByPred = _.groupBy(
       inArcsGroup[parentType],
--- a/static/js/shared/util.js
+++ b/static/js/shared/util.js
@@ -20,7 +20,7 @@ const POPULATION = "StatisticalPopulation";
 const OBSERVATION = "Observation";
 const COMPARATIVE_OBSERVATION = "ComparativeObservation";
 
-const MAX_CARD_HEIGHT = 600;
+const MAX_CARD_HEIGHT = 400;
 
 const /** !Array<string> */ STATS = [
     "measuredValue",


static/webpack.config.js
```diff
@@ -21,12 +21,11 @@ const FixStyleOnlyEntriesPlugin = require("webpack-fix-style-only-entries");
 const config = {
   entry: {
     dev: [__dirname + "/js/dev.ts", __dirname + "/css/dev.scss"],
-    dev_menu: __dirname + "/js/tools/dev_menu.ts",
```
beets, line 24: was this change intended?

pradh:
@shifucun  this is better than before but, in the example I highlighted, we'll have the same places listed under AdminArea and AdminArea1. AdminArea in the presence of AdminArea1 can be ignored.

static/webpack.config.js
shifucun: it is duplicated in line 26, so removed.

shifucun:
> @shifucun this is better than before but, in the example I highlighted, we'll have the same places listed under AdminArea and AdminArea1. AdminArea in the presence of AdminArea1 can be ignored.
> 
> https://screenshot.googleplex.com/QSP1sxLv0Ao

Trying to understand is this only for "AdminstrativeArea" or for place type ranking in general? Now the state has 3 place types, shall we only pick one? Submitting this PR and we can follow up on better ranking.

pradh:
> Trying to understand is this only for "AdminstrativeArea" or for place type ranking in general? Now the state has 3 place types, shall we only pick one? Submitting this PR and we can follow up on better ranking.

There are at least two cases here:
1. A type that strictly supersedes another. This example is the one I was pointing out, where we should skip "AdminArea" completely. The other type is "Place".
2. Equivalent types. This is the case where State <=> AdminArea1, County <=> AdminArea2. This is more of product decision.
   (a) We can do your current approach of listing under State and AdminArea1.
   (b) In our imports, we use State when its equivalent to AdminArea1 or County when AdminArea2. So treat these types equivalently and show one category.

I'm okay with (2a) or (2b), but just noting that (1) is odd.

shifucun:
> > Trying to understand is this only for "AdminstrativeArea" or for place type ranking in general? Now the state has 3 place types, shall we only pick one? Submitting this PR and we can follow up on better ranking.
> 
> There are at least two cases here:
> 
> 1. A type that strictly supersedes another. This example is the one I was pointing out, where we should skip "AdminArea" completely. The other type is "Place".
> 2. Equivalent types. This is the case where State <=> AdminArea1, County <=> AdminArea2. This is more of product decision.
>    (a) We can do your current approach of listing under State and AdminArea1.
>    (b) In our imports, we use State when its equivalent to AdminArea1 or County when AdminArea2. So treat these types equivalently and show one category.
> 
> I'm okay with (2a) or (2b), but just noting that (1) is odd.

2(b) is good since this can benefit all clients. That means there is always only one equivalent place types for a place?

For (1), is there other supersedes relations other than "AdministrativeArea" and "Place"?

pradh:
> > > Trying to understand is this only for "AdminstrativeArea" or for place type ranking in general? Now the state has 3 place types, shall we only pick one? Submitting this PR and we can follow up on better ranking.
> > 
> > 
> > There are at least two cases here:
> > 
> > 1. A type that strictly supersedes another. This example is the one I was pointing out, where we should skip "AdminArea" completely. The other type is "Place".
> > 2. Equivalent types. This is the case where State <=> AdminArea1, County <=> AdminArea2. This is more of product decision.
> >    (a) We can do your current approach of listing under State and AdminArea1.
> >    (b) In our imports, we use State when its equivalent to AdminArea1 or County when AdminArea2. So treat these types equivalently and show one category.
> > 
> > I'm okay with (2a) or (2b), but just noting that (1) is odd.
> 
> 2(b) is good since this can benefit all clients. That means there is always only one equivalent place types for a place?

Maybe I didn't state (2b) correctly. What I meant was, there are a set of equivalent place types, and places may be tagged with one or more of those types (i.e., a place may be State and AdminArea1 or just State only or just AdminArea1 only).

What are the equivalent types? When I look at kPlaceTypesWithPopObs, I think these:
  - AA1: also State
  - AA2: also County
    (NOTE: We have State and County only for US geos I believe, non-US will be AA1/AA2.)
  - City: also Town, Village, Borough
    (NOTE: Census reports these as "places" with different sub-types, but we treat them all equivalently as City. Such geos will have multiple other types besides City)
  - Province is unused

> For (1), is there other supersedes relations other than "AdministrativeArea" and "Place"?

From the list noted above, I think AdminArea and Place are the two basic types that will get superseded.

Finally, there can be a (3).
(3) Places that have multiple non-equivalent types. For example it can happen that wikidata refers to a place as City and geonames refers to it as AdminArea3. These aren't equivalent types, and in this case, I think your approach of listing the place under both buckets makes sense.

So, based on that the logic could be to first trim away the superseded types, then dedup equivalent-types under a canonical place-type (AA1, AA2 or City), and then show all resulting place-types

pradh:
> > > > Trying to understand is this only for "AdminstrativeArea" or for place type ranking in general? Now the state has 3 place types, shall we only pick one? Submitting this PR and we can follow up on better ranking.
> > > 
> > > 
> > > There are at least two cases here:
> > > 
> > > 1. A type that strictly supersedes another. This example is the one I was pointing out, where we should skip "AdminArea" completely. The other type is "Place".
> > > 2. Equivalent types. This is the case where State <=> AdminArea1, County <=> AdminArea2. This is more of product decision.
> > >    (a) We can do your current approach of listing under State and AdminArea1.
> > >    (b) In our imports, we use State when its equivalent to AdminArea1 or County when AdminArea2. So treat these types equivalently and show one category.
> > > 
> > > I'm okay with (2a) or (2b), but just noting that (1) is odd.
> > 
> > 
> > 2(b) is good since this can benefit all clients. That means there is always only one equivalent place types for a place?
> 
> Maybe I didn't state (2b) correctly. What I meant was, there are a set of equivalent place types, and places may be tagged with one or more of those types (i.e., a place may be State and AdminArea1 or just State only or just AdminArea1 only).
> 
> What are the equivalent types? When I look at kPlaceTypesWithPopObs, I think these:
> 
> * AA1: also State
> * AA2: also County
>   (NOTE: We use State and County only for US geos I believe, non-US will be AA1/AA2.)
> * City: also Town, Village, Borough
>   (NOTE: Census reports these as "places" with different sub-types, but we treat them all equivalently as City. Such geos will have multiple other types besides City)
> * Province is unused
> 
> > For (1), is there other supersedes relations other than "AdministrativeArea" and "Place"?
> 
> From kPlaceTypesWithPopObs, I think AdminArea and Place are the two basic types that will get superseded.
> 
> Finally, there can be a (3).
> (3) Places that have multiple non-equivalent types. For example it can happen that wikidata refers to a place as City and geonames refers to it as AdminArea3. These aren't equivalent types, and in this case, I think your approach of listing the place under both buckets makes sense.
> 
> So, based on that the logic could be to first trim away the superseded types, then dedup equivalent-types under a canonical place-type (AA1, AA2 or City), and then show all resulting place-types

I'll have a bug to make sure the equivalent types are tagged with the canonical type, so that all clients can unambiguously use that and get all the geos.

beets:
Just to tack on that this appears in the place pages as well:
http://staging.datacommons.org/place?dcid=geoId%2F06

Is the resolution that we should do (2b) and (3)?

PR #164, opened 2020-07-23
small copy fixes

beets:


server/templates/tools/timeline.html
```diff
@@ -95,7 +95,7 @@
           <li> <b>Extremes</b> <br>
             <span>Santa Clara County vs Imperial County: <a
                 href="#&&place=geoId/06085,geoId/06025&ptpv=__Person,income,age,Years15Onwards,incomeStatus,WithIncome&pc=0">Richest
-                vs Poorest CA counties</a>, <a
+                vs. Poorest CA counties</a>, <a
```
shifucun, line 98: please also help fix here https://github.com/datacommonsorg/website/blob/master/static/js/tools/timeline_info.tsx

beets: done ptal!

shifucun: thanks!

PR #165, opened 2020-07-23
fix the PVtree node filtering logic

qlj-lijuan:
Do not render if the node is an invalid value node, even if it can be expanded


static/js/tools/statsvar_menu.tsx
```diff
@@ -75,7 +75,7 @@ class Node extends Component<NodePropType, NodeStateType> {
 
     return (
       // render the node only if it is a valid SV node or it is canExpand
-      (isValidStatsVar || canExpand) && (
+      (isValidStatsVar || (this.props.t !== 'v' && canExpand)) && (
```
shifucun, line 78: lint suggests to use double quote

PR #168, opened 2020-07-24
fix the chart when statsvar is not available

qlj-lijuan:
- fix the problem that the chart cannot be rendered when the statsvar is not available for one of the places
- fix the filtering logic of menu: need to check two levels for value nodes.

### DIFF
--- a/static/js/tools/statsvar_menu.tsx
+++ b/static/js/tools/statsvar_menu.tsx
@@ -74,7 +74,8 @@ class Node extends Component<NodePropType, NodeStateType> {
     }
 
     return (
-      // render the node only if it is a valid SV node or it is canExpand
+      // render the node only if it is a valid value node
+      // or it is a property node that can be expanded
       (isValidStatsVar || (this.props.t !== "v" && canExpand)) && (
         <ul className="noborder">
           <li className="value" id={this.props.l}>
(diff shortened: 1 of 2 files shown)


static/js/tools/statsvar_menu.tsx
```diff
@@ -158,49 +158,49 @@ class Node extends Component<NodePropType, NodeStateType> {
...
-      // if the node is the top level node i.e. this.props.t === "c"
-      // we need to check if its child node has valid nodes.
-      let childCnt = 0; // number of valid child nodes
+    if (this.props.t === "p") {
+      // a property node can be expanded if it has >= 1 children
+      return this.hasChild(this.props.cd);
+    } else if (this.props.t === "c") {
+      // the top level node is expandable if has valid value node
+      // or valid property node
+      let hasChild = false;
```
shifucun, line 167: avoid have same name for variable and function "hasChild"

static/js/tools/statsvar_menu.tsx
```diff
@@ -158,49 +158,49 @@ class Node extends Component<NodePropType, NodeStateType> {
...
+        });
+        return valid;
+      }
     }
   }
 
-  private childCnt(children) {
-    let childCnt = 0;
+  private hasChild(children) {
+    // return if a property node has valid children
```
shifucun, line 194: "return true"

static/js/tools/statsvar_menu.tsx
qlj-lijuan: Got it.

static/js/tools/statsvar_menu.tsx

qlj-lijuan:
@googlebot I fixed it.

PR #169, opened 2020-07-24
Add functionality for download chart data; Update char dash and line styles for various scenarios

shifucun:
image


static/css/timeline.scss
```diff
@@ -289,4 +289,12 @@ sup {
...
   font-size: 12px;
   color: #666;
+}
+
+#timeline-lower-pane {
+  display: flex;
+}
+
+#download-link {
+  color: #fff;
```
beets, line 299: This shouldn't be required... btn-dark already gives it white text. Can you try removing and see?

static/css/tools/statvar_menu.scss
```diff
@@ -52,6 +52,19 @@ $menu-line: #5b5e60;
   padding: 24px;
 }
 
+#drill .noborder{
```
beets, line 55: Nit: please ad a space before {
And remove "px" if the value is 0

static/css/timeline.scss
shifucun: removed and changed to use <button>

static/css/tools/statvar_menu.scss
shifucun: done

shifucun:
Now svg and chips are wrapped by a card. Also added some margin

static/js/chart/draw.ts
```diff
@@ -469,37 +469,49 @@ function drawLineChart(
...
-  dashes: { [key: string]: string };
-  title: { [key: string]: string };
+interface Style {
+  color: string;
+  dash: string;
 }
 
 /**
- * Return color and dash style given place names and stats names.
+ * Return color and dash style given place names and stats var dcids.
```
beets, line 478: could you clarify what the key is for the return value. or perhaps the key should also be part of the interface?

static/js/chart/draw.ts
```diff
@@ -469,37 +469,49 @@ function drawLineChart(
...
-    dashes[placeNames[i]] = dashFn[i];
+  statsVars: string[]
+): { [key: string]: Style } {
+  const defaultColor = getColorFn([""])("a");
+  const defaultDash = "";
+  const result = {};
+  if (statsVars.length === 1) {
+    // One stats var, colors are based on places and dash is solid.
+    const colorFn = getColorFn(placeNames);
+    result[statsVars[0]] = { color: defaultColor };
```
beets, line 490: should this get the default dash as well?

static/js/chart/draw.ts
```diff
@@ -469,37 +469,49 @@ function drawLineChart(
...
+    for (const placeName of placeNames) {
+      result[placeName] = { color: colorFn(placeName), dash: defaultDash };
+    }
+  } else if (placeNames.length === 1) {
+    // One place, colors are based on stats var and dash is solid.
+    const colorFn = getColorFn(statsVars);
+    result[placeNames[0]] = { color: "#000", dash: defaultDash };
```
beets, line 497: why is the default color black? can we make it a const?

static/js/chart/draw.ts
```diff
@@ -469,37 +469,49 @@ function drawLineChart(
...
+  } else {
+    // Multiple places, multiple stats vars, colors are based on stats var and
+    // dash are based on places.
+    const colorFn = getColorFn(statsVars);
+    const dashFn = getDashes(placeNames.length);
+    for (const statsVar of statsVars) {
+      result[statsVar] = { color: colorFn(statsVar), dash: defaultDash };
+    }
+    for (let i = 0; i < placeNames.length; i++) {
+      // All places using black color;
```
beets, line 510: The black isn't used for the graph I think?

static/js/chart/draw.ts
```diff
@@ -674,18 +679,20 @@ function buildInChartLegend(
...
+        .attr("stroke-dasharray", `${plotParams[place[1]].dash}`);
+      dashWidth = LEGEND.dashWidth;
     }
+    // Draw the text.
+    lgGroup
+      .append("text")
+      .attr("transform", `translate(${dashWidth}, 0)`)
+      .attr("y", "0.3em")
+      .attr("dy", "0em")
+      .style("font-size", "14")
```
beets, line 691: svg styles such as font-size, stroke, etc should be in the css as much as possible

static/js/chart/draw.ts
```diff
@@ -600,20 +616,24 @@ function drawGroupLineChart(
         new Date(dp.label).getTime(),
         dp.value,
       ]);
-
       const line = d3
         .line()
         .x((d) => xScale(d[0]))
         .y((d) => yScale(d[1]));
-
+      let color: string;
```
beets, line 623: The split in logic here between computePlotParams and for the callers suggests that the interface might need to change. Perhaps computePlotParams should return a closure that takes in two params (place, statvar) and returns the right dash and color.

static/js/chart/draw.ts
beets: it's also unclear from just reading this function (and comment) how to use the returned dictionary. if we keep the interface as is, please clarify.

shifucun: refactored based on the one pager

static/js/chart/draw.ts
shifucun: Fixed

static/js/chart/draw.ts
shifucun: Moved non svg style to scss

static/js/chart/draw.ts
shifucun: Made a constant

static/js/chart/draw.ts
beets: thanks! can you link that here? or add a gist of it to the comment above.

some things to point out are the keys in the returned value. it will fail if stat var display titles diverge from the actual stat var. or maybe clarify that display names are required?

shifucun: Yea, this is not ideal, added comments to clarify. The issue is due to sharing the color information with the stats var chips component. Ideally, if we can draw the chips as svg inside the drawGroupLineChart() function, this will be cleaner.

static/js/tools/timeline_chart.tsx
```diff
@@ -89,24 +96,40 @@ class ChartRegion extends Component<ChartRegionPropsType, {}> {
...
-              })}
+            <div key={domId} className="card">
+              <div id={domId} className="chart-svg"></div>
+              <div>
+                {statsVars.map(
+                  function (statsVar) {
+                    let color: string;
+                    const statsVarTitle = this.props.statsVars[statsVar].title;
+                    if (statsVars.length === 1) {
+                      color = "#999";
```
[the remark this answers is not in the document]
shifucun: Done

static/js/chart/draw.ts
```diff
@@ -606,107 +620,89 @@ function drawGroupLineChart(
...
+        .attr("stroke-dasharray", `${legendStyle.dash}`);
+      dashWidth = LEGEND.dashWidth;
     }
+    // Draw the text.
+    lgGroup
+      .append("text")
+        .attr("class", "legend-text")
+        .attr("transform", `translate(${dashWidth}, 0)`)
+      .attr("y", "0.3em")
+      .attr("dy", "0em")
```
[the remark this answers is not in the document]
shifucun: Done

static/js/chart/draw.ts
beets: Yea, gotcha.

shifucun:
merge from master, ptal

PR #174, opened 2020-07-25
add source info to the chart

qlj-lijuan:
image

### DIFF
--- a/static/js/chart/draw.ts
+++ b/static/js/chart/draw.ts
@@ -20,7 +20,7 @@ import { DataGroup, DataPoint } from "./base";
 
 const NUM_X_TICKS = 5;
 const NUM_Y_TICKS = 5;
-const MARGIN = { top: 20, right: 10, bottom: 30, left: 35, yAxis: 3, sourceText: 8 };
+const MARGIN = { top: 20, right: 10, bottom: 30, left: 35, yAxis: 3};
 const LEGEND = {
   ratio: 0.2,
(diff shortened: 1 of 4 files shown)


static/js/chart/draw.ts
```diff
@@ -20,7 +20,7 @@ import { DataGroup, DataPoint } from "./base";
 
 const NUM_X_TICKS = 5;
 const NUM_Y_TICKS = 5;
-const MARGIN = { top: 20, right: 10, bottom: 30, left: 35, yAxis: 3 };
+const MARGIN = { top: 20, right: 10, bottom: 30, left: 35, yAxis: 3, sourceText: 8 };
```
shifucun, line 23: can make this a separate variable: SOURCE_FONT = 8;

this is for the margin at page level

static/js/tools/timeline_chart.tsx
```diff
@@ -197,6 +196,7 @@ class ChartRegion extends Component<ChartRegionPropsType, {}> {
         statsVarTitle[statsVar] = this.props.statsVars[statsVar].title;
       }
       plotParams.title = statsVarTitle;
+      plotParams.source = Array.from(statsData.data.sources);
```
shifucun, line 199: In my other PR, I am changing the plotParams to be keyed by label and value with only color and style.

The source is technically not part of a "plot". so can just pass the source as argument of drawGroupLineChart(), no need to change the PlotParams struct

PR #177, opened 2020-07-27
Support stats var dcid only url parameter in timeline tool

qlj-lijuan:
- generate the mapping from statsvarId to node path
- remove node path from url


static/js/tools/timeline_util.ts
```diff
@@ -77,48 +76,55 @@ function updateUrl(param: UrlParam) {
...
-    for(const node of statvarPath){
-      encodedPath.push(encodeURI(node));
-    }
-    const statsVarUrl = encodedPath.join(SEP);
+  // update statsVar
+  if ("statsVar" in param) {
+    const statsVarUpdate = param.statsVar.statsVar;
+    const statsVarUpdateName = statsVarUpdate.split(",")[0];
+    const statsVarUpatePath = statsVarUpdate.split(",").slice(1);
     let statsVarList = [];
```
shifucun, line 87: make this a Set?

qlj-lijuan: sg. Done

PR #178, opened 2020-07-27
Use mock placeid2dcid config in TEST env

shifucun:
Should not access real GCS in the tests. This allows to add selenium test

### DIFF
--- a/server/tests/browser_test.py
+++ b/server/tests/browser_test.py
@@ -13,7 +13,6 @@
 # limitations under the License.
 
 import unittest
-import logging
 
 from main import app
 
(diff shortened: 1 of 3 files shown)


server/tests/browser_test.py
```diff
@@ -12,17 +12,16 @@
...
 from main import app
 
+
 class TestStaticPage(unittest.TestCase):
     def test_kg_static(self):
         response = app.test_client().get('/browser/')
         assert response.status_code == 200
-        assert b"The Data Commons Graph (DCG)" in response.data
-
+        assert b"The Data Commons Graph is constructed by" in response.data
```
beets, line 24: :-o

Thanks for fixing these tests! How was I able to merge these changes without hitting ci failures?

shifucun:
> Nice change, but what's the context for it?

Hao is trying to run this and found he need to access the actual gcs, so fixing that

PR #189, opened 2020-07-29
Switch current timeline tool to the new one that is based on statisti…

shifucun:
…ca variable


shifucun:
> Nice! Though should we delete the old timeline template?

Yeah, plan to let the new tool to run for sometime then delete all the old code.

shifucun:
Holding this off a bit. I will wait for this week's data release which has all the source info in GetStats() cache to get into staging and submit this PR.

shifucun:
merged with master and fixed a test. PTAL

PR #192, opened 2020-07-29
Geojson download

fpernice-google:
Script to download the desired geojson data from the DataCommons API to be simplified.


tools/geojson_simplifier/download.py
```diff
@@ -0,0 +1,59 @@
...
+
+    def __init__(self):
+        dc.set_api_key('dev')
+        self.geojsons = None
+
+    def download_data(self, place='country/USA'):
+        geolevel_below = dc.get_property_values([place],
+                                                "typeOf")
+        geolevel_below = geolevel_below[place][0]
+        geos_contained_in_place = dc.get_places_in(
```
jeffreyoldham, line 47: Does this call always succeed?  If not, what is the appropriate action?

tools/geojson_simplifier/download.py
```diff
@@ -0,0 +1,59 @@
...
+                                    [place],
+                                    self.LEVEL_MAP[geolevel_below])
+        geos_contained_in_place = geos_contained_in_place[place]
```
jeffreyoldham, line 50: It might be possible to append the "[place]" in the previous line, reducing the code length by one.

tools/geojson_simplifier/download.py
```diff
@@ -0,0 +1,59 @@
...
+        "Country": "AdministrativeArea1",
+        "AdministrativeArea1": "AdministrativeArea2",
+        "AdministrativeArea2": "City"
+    }
+
+    def __init__(self):
+        dc.set_api_key('dev')
+        self.geojsons = None
+
+    def download_data(self, place='country/USA'):
```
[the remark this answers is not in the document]
fpernice-google: Done.

tools/geojson_simplifier/download.py
```diff
@@ -0,0 +1,59 @@
...
+        geolevel_below = dc.get_property_values([place],
+                                                "typeOf")
+        geolevel_below = geolevel_below[place][0]
```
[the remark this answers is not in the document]
fpernice-google: Done.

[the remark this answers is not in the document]
fpernice-google: I think we should just let it fail.

tools/geojson_simplifier/download.py
fpernice-google: Same as above.

tools/geojson_simplifier/download.py
fpernice-google: Done.

tools/geojson_simplifier/download.py
```diff
@@ -21,8 +21,44 @@
 import geojson
 
 
-# TODO(fpernice-google): support downloading more than just US states
-class GeojsonDownloder:
+# TODO(fpernice-google): Support downloading more than just US states.
+class GeojsonDownloader:
+    """Downloads desired Geojsons files from the DataCommons Knowledge Graph.
```
jeffreyoldham, line 26:
```suggestion
    """Downloads desired GeoJSON files from the DataCommons Knowledge Graph.
```

tools/geojson_simplifier/download.py
```diff
@@ -21,8 +21,44 @@
...
+
+    Attributes:
+        geojsons: A dictionary that maps each queried area to another
+                  dictionary containing the geojson coordinate information. An
```
jeffreyoldham, line 30:
```suggestion
                  dictionary containing the GeoJSON coordinate information. An
```

tools/geojson_simplifier/download.py
```diff
@@ -21,8 +21,44 @@
...
+                  example of this is the following
+
+                {   # Multipolygon of the state of Alabama (fake).
```
jeffreyoldham, line 33: Thanks for the example.

tools/geojson_simplifier/download.py
```diff
@@ -21,8 +21,44 @@
...
+                            # Polygon 1
+                            [[ [1.5, 12.4], [5.3, 45.2], [1.1, 3.5],
+                                                            [1.5, 12.4] ]],
+                            # Polygon 2
+                            [[ [1, 2], [3, 4], [5, 6], [2, -1], [1, 2] ]],
+                            # Polygon 3
+                            [[ [53, 23], [65, 2], [31, 12], [53, 23] ]]
+                        ]
+                    }],
+                    # Polygon of the state of Illinois (fake).
```
jeffreyoldham, line 48: Wyoming is more of a rectangle.

tools/geojson_simplifier/download.py
```diff
@@ -34,17 +70,39 @@ def __init__(self):
         self.geojsons = None
 
     def download_data(self, place='country/USA'):
+        """Downloads geojson data for specified location.
```
jeffreyoldham, line 73:
```suggestion
        """Downloads GeoJSON data for a specified location.
```

tools/geojson_simplifier/download.py
```diff
@@ -34,17 +70,39 @@ def __init__(self):
...
+
+        Given the specified location, extracts the geojsons of all
```
jeffreyoldham, line 75:
```suggestion
        Given the specified location, extracts the GeoJSONs of all
```

tools/geojson_simplifier/download.py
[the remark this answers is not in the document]
jeffreyoldham: https://github.com/datacommonsorg/api-python/blob/master/datacommons/core.py indicates it will throw an exception upon failure.

tools/geojson_simplifier/download.py
```diff
@@ -34,17 +70,39 @@ def __init__(self):
...
+        administrative areas one level below it (as specified by the
+        LEVEL_MAP class constant). For example, if the input is country/USA,
+        extracts all AdministrativeArea1's within the US (US states).
+
+        Args:
+            place: A string that is a valid value for the geoId property of a
+                   DataCommons node.
+        """
```
jeffreyoldham, line 83:
```suggestion
        Raises:
          ValueError: If a Data Commons API call fails.
        """
```

tools/geojson_simplifier/download.py
```diff
@@ -34,17 +70,39 @@ def __init__(self):
...
-        geos_contained_in_place = geos_contained_in_place[place]
+                                    self.LEVEL_MAP[geolevel_below])[place]
         self.geojsons = dc.get_property_values(geos_contained_in_place,
                                                "geoJsonCoordinates")
 
     def save(self, prefix='', path='./original-data/'):
+        """Saves the downloaded geojsons to disk.
+
+        Args:
+            prefix: Prefix prepended to the geoId of a given geojson to
```
jeffreyoldham, line 100:
```suggestion
            prefix: Prefix prepended to the geoId of a given GeoJSON to
```

tools/geojson_simplifier/download.py
```diff
@@ -34,17 +70,39 @@ def __init__(self):
...
+                    determine the name of its filename. For example, if
+                    prefix='original-', a resulting filename might be
+                    'original-geoId-01.geojson'.
+            path: Path in which to save the desired files.
```
jeffreyoldham, line 104:
```suggestion
            path: Directory in which to save the desired files, as a string.
```

tools/geojson_simplifier/download.py
fpernice-google: Done.

tools/geojson_simplifier/download.py
fpernice-google: Done.

tools/geojson_simplifier/download.py
fpernice-google: No problem :)

tools/geojson_simplifier/download.py
fpernice-google: Changed name.

tools/geojson_simplifier/download.py
fpernice-google: Done.

tools/geojson_simplifier/download.py
fpernice-google: Done.

tools/geojson_simplifier/download.py
fpernice-google: Added.

tools/geojson_simplifier/download.py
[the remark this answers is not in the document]
fpernice-google: That seems reasonable to me. Are you saying we should catch the exception and report something else?

tools/geojson_simplifier/download.py
fpernice-google: Done.

tools/geojson_simplifier/download.py
fpernice-google: Done.

fpernice-google:
@beets

tools/geojson_simplifier/download.py
```diff
@@ -0,0 +1,120 @@
...
+                }
+    """
+    LEVEL_MAP = {
+        "Country": "AdministrativeArea1",
+        "AdministrativeArea1": "AdministrativeArea2",
+        "AdministrativeArea2": "City"
+    }
+
+    def __init__(self):
+        dc.set_api_key('dev')
```
beets, line 69: This isn't necessary anymore! Besides, dev isn't a correct key.

tools/geojson_simplifier/download.py
```diff
@@ -0,0 +1,120 @@
...
+        extracts all AdministrativeArea1's within the US (US states).
+
+        Args:
+            place: A string that is a valid value for the geoId property of a
+                   DataCommons node.
+
+        Raises:
+            ValueError: If a Data Commons API call fails.
+        """
+        geolevel_below = dc.get_property_values([place],
```
beets, line 87: Is there a better name you can use to describe this variable? It's really the geo level of the specified place, not below..

beets:
Also, can you respond here to the question above? re:
"Why are we storing the GeoJSON files? We may need them, but I am not sure why."

shifucun:
Storing these files will require updating them anytime new places are involved or the geo data are updated on the graph. Loading them on server start would save all these hassle.

fpernice-google:
@shifucun @beets The saving of the files is a temporary measure. Ultimately, the simplified files (as outputted by #194) would be imported into the DataCommons Graph so that they can be queried directly. These would represent a "compressed" version of the original US Census files, and would be available via the DataCommons API alongside the original files. The benefit of this is that applications like our Choropleth tool could execute queries much faster, allowing for faster response time of the tool.

As a first step though, these files will be available directly in disk. This allows us to get a fast-working Choropleth tool while we go through the process of merging the simplified files into DataCommons. Strictly speaking, the only files that need to be available locally are the simplified ones (so there shouldn't be a need for the originals). However, I decided to tentatively add them since they will be useful in certain views of the Choropleth map (like states in the US), where the rendering speed of original maps is not an issue

tools/geojson_simplifier/download.py
fpernice-google: Getting the following error when trying to run without this line:
```
Traceback (most recent call last):
  File "download.py", line 118, in <module>
    loader.download_data()
  File "download.py", line 86, in download_data
    geolevel_below = dc.get_property_values([place],
  File "/home/fpernice/.local/lib/python3.8/site-packages/datacommons/core.py", line 174, in get_property_values
    payload = utils._send_request(url, req_json=req_json)
  File "/home/fpernice/.local/lib/python3.8/site-packages/datacommons/utils.py", line 94, in _send_request
    raise ValueError(
ValueError: Request error: Must set an API key before using the API! You can call datacommons.set_api_key or assign the key to an environment variable named DC_API_KEY
```

fpernice-google: It seems like putting in any API key works though... Not sure why that is.

tools/geojson_simplifier/download.py
fpernice-google: Good catch! I had copied some of this code from Ian's code and I forgot to change the variable name :/

tools/geojson_simplifier/download.py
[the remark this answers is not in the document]
fpernice-google: Resolved offline.

tools/geojson_simplifier/download.py
fpernice-google: Resolved offline.

beets:
re: api key - looks like it's because we are using a forked library, my mistake..

beets:
@shifucun we'll address the data load before merging into master.

shifucun:
> @shifucun @beets The saving of the files is a temporary measure. Ultimately, the simplified files (as outputted by #194) would be imported into the DataCommons Graph so that they can be queried directly. These would represent a "compressed" version of the original US Census files, and would be available via the DataCommons API alongside the original files. The benefit of this is that applications like our Choropleth tool could execute queries much faster, allowing for faster response time of the tool.
> 
> As a first step though, these files will be available directly in disk. This allows us to get a fast-working Choropleth tool while we go through the process of merging the simplified files into DataCommons. Strictly speaking, the only files that need to be available locally are the simplified ones (so there shouldn't be a need for the originals). However, I decided to tentatively add them since they will be useful in certain views of the Choropleth map (like states in the US), where the rendering speed of original maps is not an issue

Got it. Thanks for clarification!

PR #193, opened 2020-07-29
Fixed test issues in Choropleth tool. 

IanCostello:


server/templates/choropleth.html
```diff
@@ -0,0 +1,33 @@
+{#
```
beets, line 1: Didn't this file move to templates/tools/choropleth.html? This looks like an old version of the file, please rm.

PR #194, opened 2020-07-29
Geojson simplify

fpernice-google:
Script to simplify the data downloaded by the script in PR #192 .

PR #195, opened 2020-07-29
Added plotting tool to visualize changes.

fpernice-google:
Script to plot a comparison between the data downloaded by script in #192  and simplified in #194 .

### DIFF
--- a/tools/geojson_simplifier/plotter.py
+++ b/tools/geojson_simplifier/plotter.py
@@ -1,0 +1,53 @@
+# Copyright 2020 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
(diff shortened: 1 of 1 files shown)


tools/geojson_simplifier/plotter.py
```diff
@@ -0,0 +1,53 @@
...
+FLAGS = flags.FLAGS
+flags.DEFINE_string('original_path', default=None,
+                    help='Path to original geojson to be compared.')
+flags.DEFINE_string('simplified_path', default=None,
+                    help='Path to simplified geojson to be compared.')
+
+
+def main(_):
+    flags.mark_flag_as_required('original_path')
+    flags.mark_flag_as_required('simplified_path')
```
jeffreyoldham, line 39: Let's move these immediately after the flags declarations.

tools/geojson_simplifier/plotter.py
```diff
@@ -0,0 +1,53 @@
...
+    python3 plotter.py --original_path original-data/geoId-01.geojson
+                       --simplified_path simplified-data/geoId-01-simple.geojson
+"""
+
+from absl import flags
+from absl import app
+import geopandas as gpd
+import matplotlib
+matplotlib.use('TkAgg')
+import matplotlib.pyplot as plt
```
jeffreyoldham, line 27: Do we need to update the requirements.txt file?

fpernice-google: Nope. These packages are already included.

tools/geojson_simplifier/plotter.py
fpernice-google: Moved.

PR #199, opened 2020-07-29
Add update condition to reduce unnecessary re-render, combine individual set states into one

qlj-lijuan:
- add update condition in Page component
- set the statsvar titles when all statsvar info are received
- remove the unused field "search" in menu
- remove place state in search bar


static/js/tools/timeline_page.tsx
```diff
@@ -54,6 +54,18 @@ class Page extends Component<PagePropType, PageStateType> {
       statsVarValid: new Set(),
     };
   }
+  shouldComponentUpdate(nextProps, nextState) {
+    if (
+      JSON.stringify(this.state.statsVarInfo) !==
+        JSON.stringify(nextState.statsVarInfo) ||
+      this.state.places !== nextState.places ||
+      this.state.perCapita !== nextState.perCapita
+    ) {
```
shifucun, line 63: just "return ()"

qlj-lijuan: done

PR #201, opened 2020-07-30
Use new GetRelatedPlace() api for similar places and ranking

shifucun:
The new api is based on statsvar dcid and support query multiple statsvars in the same time. This boost the speed of ranking data fetch.

After this is pushed to prod, we can remove the old cache @Spaceenter 

Also get rid of legacy human readable stats var handling as they are removed from cache. @qlj-lijuan


server/main.py
```diff
@@ -49,6 +48,13 @@
 GCS_BUCKET = app.config['GCS_BUCKET']
 _MAX_SEARCH_RESULTS = 1000
 
+RANKING_STATS = {
```
beets, line 51: Please add a descriptive comment for this const.

server/main.py
```diff
@@ -274,42 +252,26 @@ def api_ranking(dcid):
...
-            'data': get_related_place(
-                dcid, 'CriminalActivities', 'count', 'measuredValue',
-                pvs_string='crimeType^UCR_CombinedCrime',
-                same_place_type=True, within_place=parent, is_per_capita=True)})
-
-    result['label'] = [
-        'Population', 'Median Income', 'Median Age', 'Unemployment Rate',
-        'Crime per capita']
+        crime_statsvar = {
+            'Count_CriminalActivities_CombinedCrime': 'Crime per capita'}
```
beets, line 264: Should this be part of RANKING_STATS as well?

server/services/datacommons.py
```diff
@@ -62,7 +62,7 @@
     'get_place_obs': '/bulk/place-obs',
     'get_chart_data': '/node/chart-data',
     'get_stats': '/bulk/stats',
-    'get_related_places': '/node/related-places',
+    'get_related_places': '/node/related-locations',
```
beets, line 65: Is this also temporary, to support the rollout?

server/main.py
shifucun: This one uses perCapita option while the rest does not. The API does not support mixed options, so need to make a separate call. Made an issue to improve on mixer side: https://github.com/datacommonsorg/mixer/issues/262

server/services/datacommons.py
shifucun: Yeah, now there are both /node/related-places and /node/related-locations API and cache data. The old one will be removed after rollout.

server/main.py
shifucun: Done

server/main.py
beets: I see, thanks for the clarification. Could you add a note to that here (or in RANKING_STATS above)

server/services/datacommons.py
beets: Cool. I'd prefer it if we kept /node/related-places at the end of the switch. :)

server/main.py
shifucun: Done

server/services/datacommons.py
shifucun: okay, added a todo

PR #208, opened 2020-07-30
Use regex to filter top NAICS code

shifucun:
image


static/js/browser/kg.js
```diff
@@ -803,7 +787,10 @@ function renderKGPage(
         let popDcids = [];
         if (key.includes("naics")) {
           for (const pop of popGroup[key]) {
-            if (TOP_NAICS.includes(pop["pvs"]["naics"])) {
+            if (pop["pvs"]["naics"].match(
+                /(NAICS\/10\d*|NAICS\/\d\d|NAICS\/\d\d-\d\d)/g
```
pradh, line 791: A few comments:

- Make this a constant and add some comments what this filters?

- I'm assuming match accepts RE2 pattern. But if unit-tests are already setup, adding a case would be nice.

- While this is a big improvement, there are a list of additional whitelisted NAICS that show up in the cache. Its okay to ignore for now, but worth adding a TODO?

shifucun: moved regex as constant. no unittest fo this yet, but i verified locally and attached screenshot in the PR description.

pradh: Thanks for the screenshot.

PR #213, opened 2020-07-30
Fix empty data issue when getting ranking data

shifucun:


server/main.py
```diff
@@ -260,10 +260,9 @@ def api_ranking(dcid):
         response = get_related_place(
             dcid, '^'.join(crime_statsvar.keys()), same_place_type=True,
             within_place=parent, is_per_capita=True)
-        for stats_var, label in crime_statsvar.items():
-            result[label].append({
-                'name': parent_names[parent],
-                'data': response[stats_var]})
+        for stats_var, data in response.items():
```
beets, line 263: I'm not sure I follow this change.. are you sure you want to use the crime_statsvar for all items in response?

shifucun: This is to mimic line 251, the stats_var (Count_CriminalActivities_CombinedCrime) here is from crime_statsvar which is different than RANKING_STATS.

The issue before is: for County, there is no crime data so the result from api is empty, calling response[Count_CriminalActivities_CombinedCrime] leads to exception. This way is more general, and if we add other stats var with perCapita option, they can just be added here, um, with the need of rename "crime_statsvar"

beets: oh gotcha. thanks for the explanation, sorry i missed that.

PR #216, opened 2020-07-30
Fix 0 range in d3 y axis plot

shifucun:
Current behavior:
image

With this fix:
image


beets:
Wow, it's really all 0...

PR #217, opened 2020-07-30
[Landing page] Fix issues with place types

beets:
Fix a few issues with new place types, in both the page subtitle and the child places.

image
image

### DIFF
--- a/static/js/place/place_template.tsx
+++ b/static/js/place/place_template.tsx
@@ -414,7 +414,9 @@ class ChildPlace extends Component<ChildPlacePropType, {}> {
         {Object.keys(this.props.childPlaces).map((placeType) => (
           <div key={placeType}>
             <div className="child-place-type">{pluralize(placeType)}</div>
-            {this.props.childPlaces[placeType].map((place, i) => (
+            {this.props.childPlaces[placeType]
+            .sort((a, b) => a.name > b.name)
+            .map((place, i) => (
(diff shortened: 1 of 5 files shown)


static/js/place/place_template.tsx
```diff
@@ -411,10 +445,13 @@ class ChildPlace extends Component<ChildPlacePropType, {}> {
...
       <React.Fragment>
+        <span id="child-place-head">Places in {this.props.placeName}</span>
         {Object.keys(this.props.childPlaces).map((placeType) => (
-          <div key={placeType}>
-            <div className="child-place-type">{pluralize(placeType)}</div>
-            {this.props.childPlaces[placeType].map((place, i) => (
+          <div key={placeType} className="child-place-group">
+            <div className="child-place-type">{pluralizedDisplayNameForPlaceType(placeType)}</div>
+            {this.props.childPlaces[placeType]
+            .sort((a, b) => a.name > b.name)
```
shifucun, line 453: can remove this sort() as input as already sorted

shifucun (approved):

static/js/place/place_template.tsx
beets: It seems to be sorted by population (maybe it's done in the mixer, but it's not in the website repo), but I think alphabetical is better (easier to scan and find a child place).

PR #223, opened 2020-07-31
Support multiple statsVars in the same UI node

qlj-lijuan:
- keep all the human readable statsVars in the PV-tree
- support multiple statsVars in statsvar_menu


static/data/statsvar_path.json
```diff
@@ -8679,13 +8679,27 @@
   6,
   2
  ],
+ "Count_Person_3OrMoreYears_Female_EnrolledInPrivateSchool": [
```
shifucun, line 8682: basically these are all the duplicate nodes so far

static/js/tools/timeline_util.ts
```diff
@@ -62,38 +62,41 @@ function updateUrl(param: UrlParam) {
   }
   // update statsVar
   if ("statsVar" in param) {
```
shifucun, line 64: I feel use an object/interface to wrap url data is very needed now. Can barely follow the if and else here :)

tools/pv_tree_generator/build_tree.py
```diff
@@ -50,39 +50,53 @@ def build_tree_recursive(pos, level, pop_obs_spec, stat_vars,
...
+                        value_blob['sv_set'] |= branch['sv_set']
+                        del branch['sv_set']
+                value_blob['c'] = len(value_blob['sv_set'])
+            else:
+                # if this is a duplicated statsVar
+                # append the statsVar to the right node, increase the count by 1
+                for child in result['cd']:
+                    if child['e'] == sv.pv[property_diff]:
+                        child['sv'].append(sv.dcid)
+                        child['c'] += 1
```
shifucun, line 98: might not need to increase the count here. the count is more of the UI node rather than the stats var count.

static/js/tools/timeline_util.ts
qlj-lijuan: I moved some logics into helper functions.

static/data/statsvar_path.json
qlj-lijuan: right

static/js/tools/timeline_util.ts
qlj-lijuan: I moved some logics into helper functions, hope it's easier to follow now

tools/pv_tree_generator/build_tree.py
qlj-lijuan: done.

qlj-lijuan:
> Thanks for the quick change. Please verify locally see if the health data can show up again.

Looks good now. Count_death is available.

static/js/tools/timeline_util.ts
shifucun: looks much better!

PR #224, opened 2020-07-31
Do not fold the menu when children are unselected

qlj-lijuan:


static/js/tools/statsvar_menu.tsx
```diff
@@ -154,7 +154,7 @@ class Node extends Component<NodePropType, NodeStateType> {
   private _handleHashChange() {
     const statsVarPathNext = [];
     let check = false;
-    let expand = false;
+    let expand = this.state.expanded;
```
shifucun, line 157: To confirm, once a node is expanded, it will keep expanded

qlj-lijuan: Right. Unless the user manually close it, or the page is refreshed.

PR #226, opened 2020-07-31
Added StatVar exploration sidebar to Choropleth map

IanCostello:
Screen Shot 2020-07-31 at 9 54 08 AM


static/js/tools/statsvar_menu.tsx
```diff
@@ -104,15 +106,21 @@ class Node extends Component<NodePropType, NodeStateType> {
     );
   };
   private _handleCheckboxClick = (): void => {
```
shifucun, line 108: please sync with @qlj-lijuan as the stats var menu code is in a constant change. A few things to note:

1) There maybe some logic specific to timeline chart now. This is a good example to see how we can make it general and let each app (timeline, choropleth, scatter) to extend it.

2) For the map view, I guess the statsvar menu should use radio button instead of check box, as you only want to select one stats var each time? @beets to confirm on the UX

static/js/tools/choropleth/choropleth.ts
```diff
@@ -97,23 +113,25 @@ function generateGeoMap(geojson, plt_info) {
 function handleMapHover(geo) {
```
shifucun, line 113: I stronger suggest to make a Map React component and move these logic (together with the ones below like mouseLeave, etc) into it.

Ideally, this file should only contain a few lines as the javascript entrance file.

static/js/tools/statsvar_menu.tsx
```diff
@@ -104,15 +106,21 @@ class Node extends Component<NodePropType, NodeStateType> {
...
-    });
-    updateUrl({
-      statsVarPath: {
-        statsVar: this.props.sv + this.state.nodePath,
-        shouldAdd: !this.state.checked,
-      },
-    });
+    // Perform onClick override if provided.
+    if (this.props.onClick) {
+      this.props.onClick(this.props.sv)
```
qlj-lijuan, line 111: Can include the path info in the url too, there could be multiple nodes for the same statsvar

qlj-lijuan: Actually, we can pass updateUrl/onClick as a property function, then no need to override.

server/routes/api/place.py
```diff
@@ -15,9 +15,10 @@
 import collections
 import json
 
-from flask import Blueprint, request
-
+from flask import Blueprint, request, jsonify
```
jeffreyoldham, line 18: Per https://google.github.io/styleguide/pyguide.html#224-decision, I think we

`import flask`

and then use `flask.Blueprint`, `flask.request`, and `flask.jsonify`.  This eliminates injecting these types into the file's namespace.  It makes it easier to understand the source of these values, which is helpful when updating code.

server/routes/api/place.py
```diff
@@ -15,9 +15,10 @@
...
 from cache import cache
+import services.datacommons as dc
+# Note that significant code broke when this was changed to dc.fetch_data.
 from services.datacommons import fetch_data
 from routes.api.stats import get_stats_wrapper
```
jeffreyoldham, line 23: Likewise.

server/routes/api/place.py
```diff
@@ -15,9 +15,10 @@
 import collections
 import json
 
-from flask import Blueprint, request
-
+from flask import Blueprint, request, jsonify
 from cache import cache
+import services.datacommons as dc
+# Note that significant code broke when this was changed to dc.fetch_data.
```
jeffreyoldham, line 21: This is worrisome.  I have no mental model for why this would happen.  They should be equivalent.  May we change the note to a todo to investigate?

server/routes/api/place.py
```diff
@@ -113,6 +114,56 @@ def child(dcid):
     return json.dumps(child_places)
 
 
+# Defines the map from higher geos to their respective subgeos.
+LEVEL_MAP = {
```
jeffreyoldham, line 118: If this is the same as in another file, it would be better to define it once and import it.

server/routes/api/place.py
```diff
@@ -113,6 +114,56 @@ def child(dcid):
...
+    "Country": "AdministrativeArea1",
+    "AdministrativeArea1": "AdministrativeArea2",
+    "AdministrativeArea2": "City"
+}
+@bp.route('child/statvars')
+@cache.memoize(timeout=3600 * 24)  # Cache for one day.
+def child_statvars():
+    """
+    Gets all statistical variable avaiable for a particular sublevel of a dcid.
```
jeffreyoldham, line 127:
```suggestion
    Gets all statistical variable available for a particular sublevel of a dcid. 
```

server/routes/api/place.py
```diff
@@ -113,6 +114,56 @@ def child(dcid):
...
+    # Get required params.
+    requested_dcid = request.args.get("dcid")
+    if requested_dcid is None:
+        return jsonify({"error": "Must provide a 'dcid' field!"}, 400)
+    requested_level = request.args.get("level")
+    if requested_level is None:
+        requested_dcid_type = dc.get_property_values([requested_dcid],
+                                                "typeOf")[requested_dcid]
+        # TODO(iancostello): Handle a failed function call, e.g., returns None.
+        # TODO(iancostello): Handle the case where display_level is None.
```
jeffreyoldham, line 147:
```suggestion
        # TODO(iancostello): Handle the case where requested_level is None.
```

???  I thought this was handled above.

server/routes/api/place.py
```diff
@@ -113,6 +114,56 @@ def child(dcid):
...
+        for level in requested_dcid_type:
+            if level in LEVEL_MAP:
+                requested_level = LEVEL_MAP[level]
+                break
```
jeffreyoldham, line 151: Outdent the code?

static/js/tools/choropleth/choropleth_template.tsx
```diff
@@ -14,28 +14,122 @@
...
+    if (urlParams.has("perCapita")) {
+      isPerCapita = ["true", "t", "1"].includes(
+        urlParams.get("perCapita").toLowerCase());
     }
+    this.state = {
+      statsVarPaths: [],
+      statsVarInfo: {},
+      perCapita: isPerCapita,
+      places: [],
+      statsVarValid: new Set(),
```
jeffreyoldham, line 46: May we have comments on each of these fields?  This is the main data structure I believe.

static/js/tools/choropleth/choropleth_template.tsx
```diff
@@ -14,28 +14,122 @@
...
+              <input
+                type="checkbox"
+                id="percapita"
+                checked={this.state.perCapita}
+                name="pc"
+                onClick={this._togglePerCapita}
+              ></input>
+            </div>
+            <Menu
+              //Removed in recent PR.
```
jeffreyoldham, line 99: I am not sure how to interpret the comment.  Should it be a todo to remove the following code after updating the current branch to reflect the state of the master branch?

static/js/tools/choropleth/choropleth.ts
```diff
@@ -123,34 +141,34 @@ function mouseLeave() {
...
   }
 }
 
 /**
  * Builds a redirect link given the fields to include from search params.
- * @param fields_to_include 
+ * @param fields_to_include
  * @param from_api whether the url should be to the api or locally.
  */
 function build_choropleth_url(fields_to_include, from_api) {
```
beets, line 156: the abstraction needed is probably to just build the search string, and to either use it for the api or a page refresh.

static/js/tools/choropleth/choropleth.ts
```diff
@@ -31,20 +31,33 @@ window.onload = () => {
...
+  if (window.location.search == "") {
+    window.location.search = "statVar=Count_Person_Employed&perCapita=t&geoId=country/USA"
+  }
 
   // TODO:(iancostello) Refactor the url builder.
   // Build api call from request.
   var base_url = build_choropleth_url(
-      ['statVar', 'perCapita', 'geoId', 'level', 'mdom'], true)
-
+    ["statVar", "perCapita", "geoId", "level", "mdom"],
```
beets, line 43: this should probably just get you the maps, and the other api (/child/statvars) to get the display data. it doesn't make sense for maintainability to have both paths return stat var data.

static/js/tools/choropleth/choropleth_template.tsx
```diff
@@ -14,28 +14,122 @@
...
+      perCapita: isPerCapita,
+      places: [],
+      statsVarValid: new Set(),
+    };
+
+    // TODO(iancostello): Don't assume the property exists.
+    // Create promise to get valid stat vars for subgeos geo.
+    axios.get("/child/statvars?dcid=" + urlParams.get("dcid")).then((resp) => {
+      this.setState({
+        statsVarValid: resp.data
```
beets, line 53: do we expect invalid statsvar? is it important to clarify that it's valid?

static/js/tools/statsvar_menu.tsx
beets: +1 to radio buttons

server/routes/api/place.py
```diff
@@ -113,6 +114,56 @@ def child(dcid):
...
+        dcid -> The place dcid geoID to information for, as a string.
+        level -> The level of children to pull for, as a string.
+    Example Query:
+        api/place/child/statvars?dcid=country/USA&level=State
+        Returns all statistical variables that are value for states of the USA.
+    Returns:
+        A json list of all the available statistical variables.
+    """
+    # Get required params.
+    requested_dcid = request.args.get("dcid")
```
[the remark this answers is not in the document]
IanCostello: changed

server/routes/api/place.py
```diff
@@ -113,6 +114,56 @@ def child(dcid):
...
+    geos_contained_in_place = dc.get_places_in(
+        [requested_dcid], requested_level)
+    if requested_dcid not in geos_contained_in_place:
+        return jsonify({"error": "Internal server error."}, 500)
+    geos_contained_in_place = geos_contained_in_place[requested_dcid]
+
+    # Get all available geoIds. Assume the first 10 cover all cases.
+    geoIds = set()
+    for geoId in geos_contained_in_place[:10]:
+        geoIds = geoIds.union(statsvars(geoId))
```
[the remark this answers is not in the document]
IanCostello: Sorry this is horrible misnamed. Its returning stats vars not geoids.

static/js/tools/choropleth/choropleth_template.tsx
IanCostello: It is for only filtering statistical variables that are available for the displayed geo

static/js/tools/choropleth/choropleth.ts
IanCostello: [Changed in previous PR]

server/routes/api/place.py
```diff
@@ -113,6 +114,56 @@ def child(dcid):
...
+                break
+
+    # Get sublevels.
+    geos_contained_in_place = dc.get_places_in(
+        [requested_dcid], requested_level)
+    if requested_dcid not in geos_contained_in_place:
+        return jsonify({"error": "Internal server error."}, 500)
+    geos_contained_in_place = geos_contained_in_place[requested_dcid]
+
+    # Get all available geoIds. Assume the first 10 cover all cases.
```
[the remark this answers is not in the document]
IanCostello: See above. This should be statistical variables available. It's a decent assumption that we might need to change, but it seems to be a reasonable assumption to me.

PR #233, opened 2020-07-31
Fix some edge cases related to set statsVar title

qlj-lijuan:


qlj-lijuan:
> Can you add a TODO for test? :)

Sure

PR #238, opened 2020-08-01
Revive eslint and fix current lint errors

shifucun:
Remove tslint as it is deprecated by eslint: https://github.com/palantir/tslint/issues/4534
Add eslint config to cover recommended rules.


run_test.sh
```diff
@@ -26,7 +26,8 @@ function run_npm_test {
 # Run linter on client side code
 function run_npm_lint {
   cd static
-  npm run lint && npm run tslint
```
beets, line 29: is npm run lint now equivalent to tslint?

also, i wonder if we should just add --fix, since that will be what each developer has to do right after anyway..

btw, can we speed up npm install? it really slows things down if we have to run it every time. some options:
- add a separate option to only run npm install
- npm list eslint || npm install eslint
- https://www.npmjs.com/package/npm-cache

static/.eslintignore
```diff
@@ -0,0 +1,2 @@
+node_modules
+**/*.js
```
beets, line 2: curious what the reason is to skip js files?

static/js/tools/timeline_info.tsx
```diff
@@ -1,56 +1,183 @@
...
-                <p>Take the data and use it on your site!</p>
-                <p><a href="mailto:collaborations@datacommons.org">Send</a> us your discoveries!</p>
-            </div>
-        );
-    }
+  render(): JSX.Element {
+    return (
+      <div id="placeholder-container">
+        <p>
+          {" "}
```
beets, line 8: This space can be skipped (along with the rest of them in this file).

run_test.sh
shifucun: Yep, since this is specific to eslint. I use the option of "npm list eslint || npm install eslint"

static/js/tools/timeline_info.tsx
shifucun: Removed the spaces. Kept some of them which are used as space between <a> tags.

static/.eslintignore
shifucun: There are just too many lint to fix for the legacy js code. Add comments. Hopefully while we remove the old gni and scatter codes it will get much better.

PR #239, opened 2020-08-02
Remove code of old timeline tools. Bye, gni!

shifucun:


static/js/tools/timeline_util.ts
```diff
@@ -246,6 +245,29 @@ interface StatsVarInfo {
   title: string;
 }
 
+/*
+ * Get url params
+ */
```
beets, line 250: Could you add a clarifying comment for the return value here?

shifucun: Added comment

PR #245, opened 2020-08-04
Optimize data fetch when drawing chart

qlj-lijuan:
When drawing the chart, only fetch data of new statsVar and new places from server.


qlj-lijuan:
I'm closing this PR because the code is too out-of-date. 
We've already have the function to check if all the data is available in cache. 
The logic for comparing with previous state should also be changed. 
It would be easier to open a new PR when we want to address this issue

PR #246, opened 2020-08-04
add labels on the chart

qlj-lijuan:
- add xlabel "Date" 
- add ylabel mprop 
- add 15px space between chart and right legend
image


beets:
Thanks for adding this! There are a few alignment issues, namely the top y-label string should align with the start of the chart (the purple line), and the x-axis and horizontal grid lines, which now end at the purple line, should move right, leaving a margin between the bottom of the y-label and the start of the grid lines.

Can we also maintain the same font-size as the other strings there already (10px), or if we want slightly larger, 12px. The legend text is at 14px and should be standardized with the rest as well.

alignment

shifucun:
> thx for adding this!
> 
> A separate ask to add for this PR. Can you change the y label from "500m" to "0.5" when per capita is selected. This should be some option in the y axis option in d3

Can ignore this request. I sent a fix in https://github.com/datacommonsorg/website/pull/264

qlj-lijuan:
image

static/js/chart/draw.ts
```diff
@@ -524,6 +525,30 @@ function drawGroupLineChart(
...
+  if (unit) {
+    ylabelText = ylabelText + "(" + unit + ")";
+  }
+  svg
+    .append("text")
+    .attr("text-anchor", "middle")
+    .attr("transform", "rotate(-90)")
+    .attr("x", -height / 2)
+    .attr("y", 10)
+    .style("font-size", 13)
```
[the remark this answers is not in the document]
qlj-lijuan: got it.

static/js/chart/draw.ts
```diff
@@ -524,6 +525,30 @@ function drawGroupLineChart(
...
+  let ylabelText = mprop.charAt(0).toUpperCase() + mprop.slice(1);
+  if (unit) {
+    ylabelText = ylabelText + "(" + unit + ")";
+  }
+  svg
+    .append("text")
+    .attr("text-anchor", "middle")
+    .attr("transform", "rotate(-90)")
+    .attr("x", -height / 2)
+    .attr("y", 10)
```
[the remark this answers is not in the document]
qlj-lijuan: right, revised.

static/js/chart/draw.ts
```diff
@@ -104,7 +107,10 @@ function addXAxis(
   const axis = svg
     .append("g")
     .attr("class", "x axis")
-    .attr("transform", `translate(0, ${height - MARGIN.bottom})`)
+    .attr(
+      "transform",
+      `translate(${YLABEL.rightMargin}, ${height - MARGIN.bottom})`
```
beets, line 112: We should only apply right margin if we are about to add a label to the margin. This would currently affect the charts on the place pages, which shouldn't change.

static/js/chart/draw.ts
```diff
@@ -122,7 +128,10 @@ function addYAxis(
   svg
     .append("g")
     .attr("class", "y axis")
-    .attr("transform", `translate(${width - MARGIN.right},0)`)
+    .attr(
+      "transform",
+      `translate(${width - MARGIN.right + YLABEL.rightMargin},0)`
```
beets, line 133: Ditto

static/js/chart/draw.ts
```diff
@@ -26,15 +26,18 @@ const LEGEND = {
   minTextWidth: 100,
   dashWidth: 30,
   lineMargin: 10,
-  marginLeft: 0,
+  marginLeft: -15,
```
beets, line 29: A negative margin seems odd.. perhaps the calculation needs to be updated?

beets:
I'm not sure if this is related to this change, but I noticed an unnecessary left padding on the source, and the text (descenders, e.g. the tail of the letter y) getting cut off. We should try to maintain a strict left-alignment.

annotated

Is the x-axis label necessary? I don't believe we have any other type of data there except date...

How about just adding a y-label like this... wdyt?

annotated-more

qlj-lijuan:
image

static/js/chart/draw.ts
qlj-lijuan: changed it back

static/js/chart/draw.ts
qlj-lijuan: updated

qlj-lijuan:
> How about just adding a y-label like this... wdyt? This would maintain a stronger left alignment and should also keep the code differences with the place line charts smaller.
> 

sounds good!

beets:
very nice, thanks for making the change!

PR #250, opened 2020-08-04
Add webdriver test

wh1210:


server/webdriver_tests/place_explorer_test.py
```diff
@@ -0,0 +1,70 @@
...
+class TestPlaceExplorer(TestBase):
+
+    def test_explorer_original(self):
+        time.sleep(3)
+        self.assertEqual("United States | Place Explorer | Data Commons", self.driver.title)
+        responses = []
+        responses = self.driver.find_elements_by_class_name("chart-container")
+        self.assertEqual(len(responses), 23)
+
+    def test_more_charts_click(self):
```
shifucun, line 59: please add some comments on what is the test for and the click sequence

server/webdriver_tests/place_explorer_test.py
```diff
@@ -0,0 +1,70 @@
...
+        time.sleep(5)
+        logging.info(self.driver.current_url)
```
shifucun, line 61: remove logging?

server/webdriver_tests/timeline_test.py
```diff
@@ -0,0 +1,85 @@
...
+    def tearDown(self):
+        self.driver.quit()
+
+    def test_server_and_page(self):
+        response = urllib.request.urlopen(self.get_server_url())
+        self.assertEqual(response.getcode(), 200)
+        self.assertEqual("Timelines Explorer | Data Commons", self.driver.title)
+
+
+class TestCharts(TestBase):
```
shifucun, line 48: please add some comments on what to test for each test case

server/webdriver_tests/timeline_test.py
```diff
@@ -0,0 +1,85 @@
...
+    def test_charts_original(self):
+        response = urllib.request.urlopen(self.driver.current_url)
+        time.sleep(3)
+        self.assertEqual(response.getcode(), 200)
+        self.assertEqual("Timelines Explorer | Data Commons", self.driver.title)
+        responses = []
+        responses = self.driver.find_elements_by_class_name("card")
+        self.assertEqual(len(responses), 0)
+
+    def test_chart_numbers_1(self):
```
shifucun, line 59: can you enrich this test with click on stats var menu, then assert there is one more chart added.

server/webdriver_tests/timeline_test.py
```diff
@@ -0,0 +1,85 @@
...
+        # TODO: Leave screenshot in future PR.
+        # self.driver.save_screenshot("test_screenshots/timeline_ca_2_charts.png")
+        response = urllib.request.urlopen(self.driver.current_url)
+        self.assertEqual(response.getcode(), 200)
+        self.assertEqual("Timelines Explorer | Data Commons", self.driver.title)
+        responses = []
+        responses = self.driver.find_elements_by_class_name("card")
+        self.assertEqual(len(responses), 2)
+
+    def test_chart_numbers_2(self):
```
shifucun, line 71: for this test can add a uncheck on the stats and assert chart is removed.

Basically we want each test to test distinct functions instead of repeating similar behavior. That said, can also update the test function names to reflect the test purpose.

server/webdriver_tests/place_explorer_test.py
```diff
@@ -0,0 +1,70 @@
...
+
+    def create_app(self):
+        return app
+
+    def setUp(self):
+        """
+        Will be called before every test
+        """
+        self.driver = webdriver.Chrome()
+        self.driver.get(self.get_server_url() + '/place?dcid=country/USA')
```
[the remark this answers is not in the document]
beets: +1. we'll want to add other pages to this list

server/webdriver_tests/place_explorer_test.py
```diff
@@ -0,0 +1,70 @@
...
+
+    def tearDown(self):
+        self.driver.quit()
+
+    def test_server_and_page(self):
```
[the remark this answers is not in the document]
beets: does this test belong here? or should other inherited tests call this, to assert basic info from the page, like the page title.

server/webdriver_tests/place_explorer_test.py
```diff
@@ -0,0 +1,70 @@
...
+        response = urllib.request.urlopen(self.get_server_url())
+        self.assertEqual(response.getcode(), 200)
+        self.assertEqual("United States | Place Explorer | Data Commons", self.driver.title)
+
+
+class TestPlaceExplorer(TestBase):
+
+    def test_explorer_original(self):
+        time.sleep(3)
+        self.assertEqual("United States | Place Explorer | Data Commons", self.driver.title)
```
beets, line 54: This is duplicated from the TestBase

server/webdriver_tests/place_explorer_test.py
```diff
@@ -0,0 +1,70 @@
...
+        responses = []
+        responses = self.driver.find_elements_by_class_name("chart-container")
+        self.assertEqual(len(responses), 23)
```
beets, line 57: This will be flaky (we will be constantly adding charts to the page). Could we, instead, test that all the charts on the page are rendered, and contain a graph? We could test for the presence of the svg element at minimum.

server/webdriver_tests/place_explorer_test.py
```diff
@@ -0,0 +1,70 @@
...
+
+    def test_more_charts_click(self):
+        time.sleep(5)
+        logging.info(self.driver.current_url)
+        elem = self.driver.find_element_by_xpath("/html/body/div[1]/main/div/div/div[2]/div[3]/section[1]/h3/a")
```
beets, line 62: Can we find element by id or class name instead?

server/webdriver_tests/place_explorer_test.py
```diff
@@ -0,0 +1,76 @@
...
+import logging
+import time
+
+from main import app
+
+
+USA_URL = '/place?dcid=country/USA'
+USA_DEMOGRAPHICS_URL = '/place?dcid=country/USA&topic=Demographics'
+
+class TestBase(LiveServerTestCase):
```
shifucun, line 32: can you have a new file called base_test.py which just have this base class, and all other *_test.py will inherit it.
The TestBase does not have url specific thing but just contains create_app(), setUp() and tearDown()

server/webdriver_tests/place_explorer_test.py
```diff
@@ -0,0 +1,76 @@
...
+        """
+        Will be called before every test
+        """
+        chrome_options = webdriver.ChromeOptions()
+        chrome_options.add_argument('--headless')
+        chrome_options.add_argument('--no-sandbox')
+        chrome_options.add_argument('--disable-dev-shm-usage')
+
+        self.driver = webdriver.Chrome(chrome_options=chrome_options)
+        self.driver.get(self.get_server_url() + USA_URL)
```
shifucun, line 47: can move this and test_page_server() to the concrete test class below. The TestBase can just have env setup without any specific test.

wh1210:
Thanks for these comments! I will address these in next PR, this PR can be seen as a test PR to test the cloudbuild.

wh1210:
> Since this is going into master, please don't comment out the other tools. I would also keep set -e, so that we will know immediately if there's something in the script that's failing.

Yep! This PR is used to test, I will make an PR to commit these~~ Thanks!

PR #252, opened 2020-08-06
remove onHashChange in timeline page component

qlj-lijuan:
- remove onHashChange in page component
- remove update url related functions in statsVar menu
- pass functions for adding/removing statsVar from page to statsVar menu

I'll remove url related function in searchBar and chart in the next PR.

### DIFF
--- a/static/js/tools/statsvar_menu.test.js
+++ b/static/js/tools/statsvar_menu.test.js
@@ -40,10 +40,12 @@ it("filtering the menu", () => {
   act(() => {
     render(
       <Menu
-        selectedNodePaths={[[0, 0]]}
+        selectedNodePaths={[["0", "0"]]}
         statsVarValid={new Set(["Count_Person"])}
         filter={true}
-        setStatsVarTitle={({}) => {}}
+        setStatsVarTitle={jest.fn()}
+        addStatsVar={jest.fn()}
+        removeStatsVar={jest.fn()}
       />,
       container
     );
@@ -54,10 +56,12 @@ it("filtering the menu", () => {
   act(() => {
     render(
       <Menu
-        selectedNodePaths={[[0, 0]]}
+        selectedNodePaths={[["0", "0"]]}
         statsVarValid={new Set([])}
         filter={true}
-        setStatsVarTitle={({}) => {}}
+        setStatsVarTitle={jest.fn()}
+        addStatsVar={jest.fn()}
+        removeStatsVar={jest.fn()}
       />,
       container
     );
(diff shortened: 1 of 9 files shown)


static/js/tools/timeline_page.tsx
```diff
@@ -17,119 +17,122 @@
...
+class Page extends Component<Record<string, unknown>, PageStateType> {
+  params: TimelineParams;
+
+  constructor(props: Record<string, unknown>) {
     super(props);
-    this.handleHashChange = this.handleHashChange.bind(this);
-    this._togglePerCapita = this._togglePerCapita.bind(this);
-    this.setStatsVarTitle = this.setStatsVarTitle.bind(this);
+    this.params = getTimelineParamsFromUrl();
+    const tempStatsVarTitle = {};
```
shifucun, line 46: can just call this statsVarTitle; temp is for temporary variable while this one is actually used (and later it is override)

static/js/tools/timeline_page.tsx
```diff
@@ -17,119 +17,122 @@
...
-    }
-
+    const statsVarInfoPromise = getStatsVarInfo(this.params.getStatsVarDcids());
+    const placesPromise = getPlaceNames(this.params.placeDcids);
+    const validStatsVarPromise = getStatsVar(this.params.placeDcids);
     Promise.all([
       statsVarInfoPromise,
       placesPromise,
       validStatsVarPromise,
     ]).then((values) => {
```
shifucun, line 68: nit: here "values" can be written as [statsVarInfo, placeIdNames, statsVarValid]

static/js/tools/timeline_util.ts
```diff
@@ -302,35 +306,49 @@ class TimelineParams {
...
   }
 
-  // add one new place
-  public addPlace(placeDcid: string): void {
-    this.placeDcids.add(placeDcid);
+  // add one new place, return true if this.placeDcids changed
+  public addPlace(placeDcid: string): boolean {
+    if (this.placeDcids.includes(placeDcid)) {
+      return false;
+    } else {
```
shifucun, line 313: nit: can remove else

static/js/tools/statsvar_menu.test.js
```diff
@@ -40,10 +40,12 @@ it("filtering the menu", () => {
   act(() => {
     render(
       <Menu
-        selectedNodePaths={[[0, 0]]}
+        selectedNodePaths={[["0", "0"]]}
```
shifucun, line 43: what is the reason to change from number to string?

static/js/tools/timeline_page.tsx
qlj-lijuan: got it.

static/js/tools/timeline_page.tsx

static/js/tools/timeline_util.ts
qlj-lijuan: done

static/js/tools/statsvar_menu.test.js
qlj-lijuan: Previously we have both string and number for node paths. e.g, the node path parsed from url is string and the node path built from index is number. To be consistent, I changed everything to string, because string is more general, it would be easier to change if we want to modify the representation of node path in the future.

PR #255, opened 2020-08-06
Addressed TODO to make paths relative.

fpernice-google:
This is an experimental PR that makes a small fix by addressing a TODO. @beets

### DIFF
--- a/static/js/tools/choropleth/choropleth.ts
+++ b/static/js/tools/choropleth/choropleth.ts
@@ -33,10 +33,9 @@ window.onload = () => {
 
   // TODO:(iancostello) Refactor the url builder.
   // Build api call from request.
-  const base_url = build_choropleth_url(
-    ["statVar", "perCapita", "geoId", "level", "mdom"],
-    true
-  );
+  let base_url = document.location.pathname + '/api';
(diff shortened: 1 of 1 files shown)


static/js/tools/choropleth/choropleth.ts
```diff
@@ -135,34 +134,27 @@ function handleMapClick(geo) {
...
 /**
  * Redirects the webclient to a particular geo. Handles breadcrumbs in redirect.
  * @param {string} geoId to redirect to.
  */
 function redirectToGeo(geoId) {
-  let base_url = build_choropleth_url(
-    ["statVar", "perCapita", "level", "mdom"],
-    false
-  );
+  let base_url = document.location.pathname;
```
jeffreyoldham, line 155: Do we need to worry about the case of "geoId"'s being the empty string?

fpernice-google: This is currently handled on the server side. In `server/routes/api/choropleth.py`, we have 
```
    if requested_geoId is None:
        return jsonify({"error": "Must provide a 'geoId' field!"}, 400)
```

PR #256, opened 2020-08-06
Clean up the statsVar filtering logic in statsVar menu

qlj-lijuan:


static/js/tools/statsvar_menu.tsx
```diff
@@ -113,7 +93,26 @@ class Node extends Component<NodePropType, NodeStateType> {
                 )}
               </a>
             </span>
-            {child}
+            {this.checkExpand &&
```
shifucun, line 96: should this be this.checkExpand()?

qlj-lijuan: right. Fixed

PR #257, opened 2020-08-06
pass statsVar dcid along with node path to StatsVar menu

qlj-lijuan:


qlj-lijuan:
> I guess can update the description from stats var name to stats var dcid?

Sounds good. I'll make the change in the following PR.

PR #261, opened 2020-08-07
Update the url when statsVar, places or percapita  is changed

qlj-lijuan:


static/js/tools/timeline_util.ts
```diff
@@ -275,6 +275,10 @@ function setSearchParam(vars: VarUrl) {
   window.location.hash = newHash;
 }
 
+const placeSep = ",";
+const nodePathSep = ",";
+const statsVarSep = "__";
```
shifucun, line 280: good to use const!

PR #264, opened 2020-08-07
UI fix on d3 chart y axis format and kg chart default height

shifucun:
- Show original number when y tick is less than 1 to avoid showing 0.5 as 500m.
- Make default "show more" height larger so the "show more" box is hidden in chart view.

image

### DIFF
--- a/static/js/shared/util.js
+++ b/static/js/shared/util.js
@@ -20,7 +20,7 @@ const POPULATION = "StatisticalPopulation";
 const OBSERVATION = "Observation";
 const COMPARATIVE_OBSERVATION = "ComparativeObservation";
 
-const MAX_CARD_HEIGHT = 400;
+const MAX_CARD_HEIGHT = 450;
 
 const /** !Array<string> */ STATS = [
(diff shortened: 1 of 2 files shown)


static/js/shared/util.js
```diff
@@ -20,7 +20,7 @@ const POPULATION = "StatisticalPopulation";
 const OBSERVATION = "Observation";
 const COMPARATIVE_OBSERVATION = "ComparativeObservation";
 
-const MAX_CARD_HEIGHT = 400;
+const MAX_CARD_HEIGHT = 450;
```
beets, line 23: Are we sure we want the charts to be so tall? The pages are already very long...

shifucun: this is what I see now: 
image

Or, we can reduce the chart size?

beets: Oh got it. I thought we were making the chart itself taller. The ratio feels appropriate for the width, sounds good then.

shifucun:
thx for review!

PR #270, opened 2020-08-10
Added better USA projection and debugged references

fpernice-google:
For review @beets @IanCostello @jeffreyoldham .


fpernice-google:
Comparison of old vs. new projection:
proj1
proj2

Issue #272, opened 2020-08-11
[BUG] Chart labels too big


beets:
Please revert the chart labels font sizes (axis, etc) for place pages.

![image](https://user-images.githubusercontent.com/6052978/89842599-f85a5380-db2a-11ea-8c18-ee3df8872595.png)

Perfectly reasonable to have larger font sizes for the timelines tool, but absolutely no space for it in the place page.

This also introduces overflow of the y-axis labels on the timelines page:

![image](https://user-images.githubusercontent.com/6052978/89842621-10ca6e00-db2b-11ea-9a01-9d7d071a6c12.png)

beets:
I'm sorry I missed it in the review. For the time being, could you include a screenshot of the place page charts in PR's for chart changes as well? Thanks!

qlj-lijuan:
Sure. Will check both pages in the future!

PR #275, opened 2020-08-11
fix issue #272 Chart labels too big 

qlj-lijuan:
Place:
image
Timeline:
image


qlj-lijuan:
Kept the font size of y-axis. Use font size 12px for y-label and source text

PR #279, opened 2020-08-11
remove url related functions in Timeline tool

qlj-lijuan:
These functions are not used anymore after refactoring.


static/js/tools/timeline_util.ts
```diff
@@ -258,22 +88,6 @@ function saveToFile(filename: string, csv: string): void {
  * Parse url hash into VarUrl object that contains statsvar, place and perCapita
```
shifucun, line 88: remove the comment?

qlj-lijuan: done.

PR #282, opened 2020-08-11
Move option of per capita to chart

qlj-lijuan:
The options of each chart , i.e. per capita in this case, is kept in an object chartOptions. The key of chartOptions is mprop of the chart because each chart represents a unique mprop. The value is a dictionary of options. 

An example of chartOptions:  {count: {pc: true}}. It means:  per capita for the chart with mprop = count is true.
More options such as log scale will be added in the future. 

We encode value of chartOptions in the url. The initial value of chartOption would be decoded from the url and updated by the callback function passed to the "onClick" prop of the per Capita buttons.

image


qlj-lijuan:
> Some high level suggestions:
> 
> Binding the chart options to mprop maybe a too strong assumption. If we decided to have other ways to group the stats vars then the logic would change a lot. A more robust way is to enclose the "grouping" information in the chart options:
> 
> chartOption = [
> {
> statsVars: [sv1, sv2, sv3]
> pc: true
> log: false
> },
> {
> statsVars = [sv4]
> pc: false,
> log: true
> }
> ...
> ]
> 
> The grouping information is from groupStatsVars function and would naturally adjust when the logic in that function change.
> 
> In each Chart component, it will check if all the stats var matches (which should), unless the logic of grouping function changes and new app got old url. Need some thought on err handling here.

I think we still need a key for each chart that does not change when we add/remove statsVars. It should be the same key as grouping, which is mprop now, it can be mprop+something else in the future.

The problem of the suggested way is that, we would have to change ChartOptions when we add/remove statsVars. It would make the logic more complex. The current logic is to use the default options for each chart, and set chartOptions when the option buttons are clicked.

PR #284, opened 2020-08-12
Fixed Choropleth colorscale issue and PC toggle

IanCostello:
Fixed the following two issues:
1) On perCapita toggle, the map would re-render before the state updates resulting in an incorrectly rendered map.
2) The code to determine the domain of the color palette would have an indexing error on odd array sizes resulting in a failure of the color scheme.


static/js/tools/choropleth/choropleth.tsx
```diff
@@ -371,7 +371,7 @@ function determineColorPalette(dict, pc: boolean, popMap: []): number[] {
   });
   const len = values.length;
   const lowerValue = values[0];
-  const medianValue = values[len / 2];
+  const medianValue = values[Math.floor(len / 2)];
```
jeffreyoldham, line 374: Median is defined as the middle value if there is an odd number of values.  It is the average of the two middle values if there are an even number of values.  It is not defined if there are no values.

PR #285, opened 2020-08-12
Added human readable breadcrumbs

fpernice-google:
image


fpernice-google:
> Thanks. Good.
> 
> So the bread crumbs are placed in the URL and then parsed for display if I understand correctly.

Correct.

beets:
Also, stylistically: the breadcrumbs should go above the chart title. See our docs pages for an example

fpernice-google:
Moved breadcrumbs to top: 
image

static/js/tools/choropleth/choropleth_template.tsx
```diff
@@ -155,9 +154,9 @@ class MainPane extends Component {
         </div>
         <div id="main-content">
           <React.Fragment>
+          <div className="column" id="breadcrumbs"></div>
             <div id="heading">Loading...</div>
             <div>
-              <div className="column" id="breadcrumbs"></div>
               <div className="column" id="hover-text-display"></div>
             </div>
```
beets, line 161: I suspect it's because you're using columns, but without adding rows.
https://getbootstrap.com/docs/4.0/layout/grid/

Did you try removing column from the divs? Or using something like this:
<div classname=row><div classname=col-12 id=breadcrumbs></div></div>

static/css/tools/choropleth.scss
```diff
@@ -50,22 +50,24 @@
...
 #hover-text-display {
   float: right;
   width: 40%;
+  height: 0;
 }
 
 #heading {
   font-size: 32px;
   text-align: center;
+  width: 100%;
```
beets, line 66: div's are usually 100% width by default, so perhaps it's the column class that's doing something here.

static/js/tools/choropleth/choropleth_template.tsx
fpernice-google: Not sure what you mean here.

static/css/tools/choropleth.scss
fpernice-google: Oh you're right, this was redundant. Removed.

static/js/tools/choropleth/choropleth_template.tsx
beets: it's ok. i wasn't sure what the column classes were doing, or if it was meant to be the bootstrap grid classes.

PR #287, opened 2020-08-12
Fixes various issues with the choropleth map

IanCostello:
1. Prevents hard failures in many API calls with better handling or failure conditions.
2. Fixes the on-hover code so that subgeographies are again highlighted (feature was broken)
3. For region with subgeos, a hand pointer is displayed to indicate that you can click into the subgeo. 
4. The map no longer displays an error if no statistical variable is chosen and a message is displayed.
5. API call failures now alert the user to the failure instead of displaying loading forever


static/js/tools/choropleth/choropleth.tsx
```diff
@@ -72,7 +70,9 @@ class ChoroplethMap extends Component {
       //shouldComponentUpdate.
       this.renderGeoMap();
       this.updateGeoValues();
-    });
+    }, () => {
+      document.getElementById("heading").innerHTML = "API Request Failed!"
```
beets, line 74: whoa! that's a loud error... could you move these error messages to a different div that we can easily style later?

beets: also, error messages are generally more helpful when there's a recourse for the user. what should they do to get things working again?

beets: applies to the rest of the error messages as well

IanCostello: I added better error messages for both of the UI and moved these over to a new div.

server/routes/api/choropleth.py
```diff
@@ -59,7 +59,8 @@ def choropleth_values():
                                  flask.request.args.get("level"))
     if not display_level:
         return flask.jsonify({"error":
-           f"Failed to automatically resolve subgeo for {dcid}."}, 400) 
+           f"Failed to automatically resolve subgeo for {dcid}." +
```
jeffreyoldham, line 62: "subgeo" might not be an easily understood term.  I am not sure what a good term is.  ""geographic subdivision level" is a guess.

PR #300, opened 2020-08-13
Fixed issue with reversed() function on server.

fpernice-google:
@beets fixed according to your request.


beets:
lgtm, but curious why were things reversed initially, and how this change affects thigs?

PR #302, opened 2020-08-14
Fixed failing API calls for choropleth map. 

IanCostello:


server/services/datacommons.py
```diff
@@ -88,11 +88,12 @@ def search(query_text, max_results):
     return response.json()
 
 
-def get_stats(place_dcids, stats_var):
+def get_stats(place_dcids, stats_var, obs_dates='latest'):
```
beets, line 91: please undo this change (obs_date is not an accepted parameter for the rest api)

server/routes/api/choropleth.py
```diff
@@ -265,3 +268,17 @@ def child_statvars():
         stat_vars_for_subgeo = stat_vars_for_subgeo.union(
             place.statsvars(geoId))
     return json.dumps(list(stat_vars_for_subgeo))
+
+def get_latest_data(payload_for_geo):
```
beets, line 272: Move this helper to the top of the file

beets:
Just to add a note here that calling reverse on a dictionary is only available for python >= v3.8

PR #303, opened 2020-08-14
Fix choropleth api:

beets:
Use same methodology as py api for getting the latest data from get_stat (reversed on a dictionary is only available for python >3.8


beets:
superceeded by #302

PR #304, opened 2020-08-14
fix cloudbuild screenshot yaml error

shifucun:
Should write env variable to /workspace/*.txt instead of set it as env variable.

### DIFF
--- a/static/css/kg.scss
+++ b/static/css/kg.scss
@@ -23,7 +23,7 @@
 }
 
 hr {
-  border-width: 20px;
+  border-width: 2px;
 }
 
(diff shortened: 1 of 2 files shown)


static/css/kg.scss
```diff
@@ -23,7 +23,7 @@
 }
 
 hr {
-  border-width: 2px;
```
beets: probably an accidental update?

shifucun:
thx for review!

static/css/kg.scss
shifucun: removed

PR #305, opened 2020-08-14
Fixed minor issues in rendering choropleth at sublevels greater than 1 and adds Alpha marking

IanCostello:


IanCostello:
Screen Shot 2020-08-14 at 11 15 18 AM

PR #316, opened 2020-08-18
Add explore more url handler that redirects to place page

shifucun:


server/routes/redirects.py
```diff
@@ -52,3 +52,9 @@ def documentation():
 @bp.route('/colab')
 def colab():
     return redirect('https://docs.datacommons.org/tutorials.html', code=302)
+
+
+@bp.route('/explore/place')
```
beets, line 57: please add a comment here, a warning perhaps, to not remove this (i had removed it earlier because i found no references to it from the repo).

shifucun: Maybe add a prober test for this?

beets: yeah, that would be great to check that the redirects continue working as expected.

PR #323, opened 2020-08-20
Add scatter tool test

wh1210:
* Add scatter tool test
* Move place explorer on top of the screenshot part.

### DIFF
--- a/server/webdriver_tests/scatter_test.py
+++ b/server/webdriver_tests/scatter_test.py
@@ -33,7 +33,7 @@ class TestScatter(WebdriverBaseTest):
         self.assertEqual("Scatterplot tool | Data Commons", self.driver.title)
 
     def test_choosing_ptpvs_and_parameters_to_draw(self):
-        """Test the original timeline page. No charts in this page."""
+        """Test the scatter plot tool."""
         self.driver.get(self.url_ + SCATTER_URL)
         # Using implicit wait here to wait for loading page.
         self.driver.implicitly_wait(3)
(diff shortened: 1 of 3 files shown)


server/webdriver_tests/scatter_test.py
```diff
@@ -0,0 +1,59 @@
...
+        expanded_menu.find_element_by_id("checkbox-Person,count").click()
+        expanded_menu.find_element_by_id("checkbox-Person,age").click()
+
+        # Choose parameters.
+        selects = Select(self.driver.find_element_by_id("place-types"))
+        selects.select_by_value("CensusCoreBasedStatisticalArea")
+        time.sleep(3)
+
+        # Assert there is a chart showing up.
+        chart = self.driver.find_element_by_id("chart-div").find_elements_by_tag_name("svg")
```
shifucun, line 54: nit: can you format this by indent a new line from "chart-div"

server/webdriver_tests/scatter_test.py
```diff
@@ -0,0 +1,59 @@
...
+        self.assertEqual(len(chart), 1)
```
shifucun, line 55: can you try to assert deeper element in the svg

server/webdriver_tests/scatter_test.py
wh1210: Done

server/webdriver_tests/scatter_test.py
wh1210: The svg component is a little messy. I was trying to find some valuable things, and ending with assert there are many circles showing up in the chart.

shifucun: check circle count is good

PR #325, opened 2020-08-20
Added linear scale on dev page.

fpernice-google:
Note: this linear scale will be moved to `choropleth.tsx` eventually. This is simply a temporary place of implementation to polish details. Below is a picture of the scale:

image

### DIFF
--- a/static/js/chart/draw.ts
+++ b/static/js/chart/draw.ts
@@ -213,6 +213,11 @@ function drawSingleBarChart(
     .attr("fill", (d) => color(d.label));
 }
 
+/**
+ * Create a canvas object with a color gradient to be used as a color scale.
+ * @param color  d3.scaleLinear object that encodes the desired color gradient.
+ * @param n Number of color tones to transition between.
+ */
 function ramp(color, n = 256) {
   let canvas = document.createElement('canvas');
   canvas.width = n;
@@ -225,6 +230,13 @@ function ramp(color, n = 256) {
   return canvas;
 }
 
+/**
+ * Draw a color scale legend.
+ * @param id Id of container div for the color scale.
+ * @param width Desired scale width
+ * @param height Desired scale height. Note: about 40 pixels will be needed
+ *               at least in order to fit the title and scale marks.
+ */
 function drawColorScale(
   id: string,
   width: number,
@@ -237,7 +249,7 @@ function drawColorScale(
   const marginBottom = 16 + tickSize;
   const marginSides = 15;
 
-  // Note: This would be replaced by the color scale used in choropleth.tsx
+  // TODO(fpernice-google): Replace by the color scale used in choropleth.tsx
   const color = d3.scaleLinear()
     .domain([-0.1, 0.1])
     .range((["#deebf7", "#3182bd"] as unknown) as number[]);
--- a/static/js/chart/draw.ts
+++ b/static/js/chart/draw.ts
@@ -219,7 +219,7 @@ function drawSingleBarChart(
  * @param n Number of color tones to transition between.
  */
 function ramp(color, n = 256) {
-  let canvas = document.createElement('canvas');
+  const canvas = document.createElement("canvas");
   canvas.width = n;
   canvas.height = 1;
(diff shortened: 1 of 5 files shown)


fpernice-google:
@beets @jeffreyoldham for review.

static/js/chart/draw.ts
```diff
@@ -213,6 +213,68 @@ function drawSingleBarChart(
     .attr("fill", (d) => color(d.label));
 }
 
+function ramp(color, n = 256) {
```
jeffreyoldham, line 216: How do we add a function comment in TypeScript?  It should explain the function's purpose, the arguments, and the returned value.

static/js/chart/draw.ts
```diff
@@ -213,6 +213,68 @@ function drawSingleBarChart(
...
+  canvas.height = 1;
+  const context = canvas.getContext("2d");
+  for (let i = 0; i < n; ++i) {
+    context.fillStyle = color(i / (n - 1));
+    context.fillRect(i, 0, 1, 1);
+  }
+  return canvas;
+}
+
+function drawColorScale(
```
jeffreyoldham, line 228: Likewise.

static/js/chart/draw.ts
```diff
@@ -213,6 +213,68 @@ function drawSingleBarChart(
...
+  height: number
+): void {
+  const ticks = width / 64;
+  const tickSize = 6;
+  const title = "Color scale.";
+  const marginTop = 18;
+  const marginBottom = 16 + tickSize;
+  const marginSides = 15;
+
+  // Note: This would be replaced by the color scale used in choropleth.tsx
```
jeffreyoldham, line 240: Please change this to a todo.

jeffreyoldham (dismissed):
Thanks.

static/js/chart/draw.ts
fpernice-google: Added based on other comments in this file as well as `choropleth.tsx`.

static/js/chart/draw.ts
fpernice-google: Done.

static/js/chart/draw.ts
fpernice-google: Done.

static/js/chart/draw.ts
```diff
@@ -213,6 +213,80 @@ function drawSingleBarChart(
     .attr("fill", (d) => color(d.label));
 }
 
+/**
+ * Create a canvas object with a color gradient to be used as a color scale.
+ * @param color  d3.scaleLinear object that encodes the desired color gradient.
+ * @param n Number of color tones to transition between.
+ */
+function ramp(color, n = 256) {
```
beets, line 221: Ah, I had meant to have this in dev.ts (so leaving draw.ts untouched) so you won't have to worry about exports and such. Is the plan to move these functions to one of the choropleth modules? Please do that in this pr, or create a new file in this folder for choropleth drawing related code.

Also, please give this a more descriptive function name.

static/js/chart/draw.ts
```diff
@@ -213,6 +213,80 @@ function drawSingleBarChart(
...
+
+    svg.append("g")
+      .attr("transform", `translate(0, ${height - marginBottom})`)
+      .call(d3.axisBottom(x)
+        .tickSize(tickSize))
+      .call(g => g.selectAll(".tick line").attr("y1", marginTop + marginBottom - height))
+      .call(g => g.select(".domain").remove())
+      .call(g => g.append("text")
+        .attr("x", marginSides)
+        .attr("y", marginTop + marginBottom - height - 6)
```
beets, line 283: please try to avoid magic numbers in code -- should be a const.

static/js/chart/draw.ts
fpernice-google: My plan was to move all of this directly to `choropleth.tsx`.

beets: ok, let's do that right away then. we can close this pr (all comments are still valid, and thanks for addressing them).

PR #341, opened 2020-08-24
Added map resizing.

fpernice-google:
After investigating for a few hours, I could find no better way of resizing the map to the screen than setting it to a size that is some fraction of the screen size. I suspect there must be a way of telling a React component to "take up as much screen as it can without overstepping other components," but I have not found a way of doing this. 

Below is a picture of the finished product, on a large screen size (such that before it would have taken up a small fraction of the screen):

Screenshot 2020-08-24 at 2 37 09 PM

### DIFF
--- a/static/css/tools/choropleth.scss
+++ b/static/css/tools/choropleth.scss
@@ -19,7 +19,7 @@
 @import '../draw';
 
 #map_container {
-  margin-left: 12%;
+  margin-left: 5%;
   margin-right: 12%;
 }
 
@@ -44,6 +44,7 @@
 
 #legend {
   float: right;
+  margin-right: 20%;
 }
 
 #explore #percapita-link {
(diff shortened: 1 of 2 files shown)


beets:
It's more of a CSS thing rather than React. What you want is to get a good layout in any window size with CSS, then use React to resize/redraw the map when the window size changes. Some resources/starting points:

https://developer.mozilla.org/en-US/docs/Learn/CSS/CSS_layout/Introduction
https://getbootstrap.com/docs/4.0/layout/grid/

What you have works for an initial layout, though.

fpernice-google:
Skinny screen:
image

Wide screen:
image

I realize now that in extremely wide screens like the above, the scale overlaps with the footer. This is because the header and footer take up a large proportion of screen size in that case (unlike the sidebar, they don't get resized proportionally to the screen). I'm not sure if this should be considered an issue.

beets:
It's a bigger fix to get the layout updated to handle the footer correctly. I'd leave that out for now. Please get the tests working and add a TODO to handle window resizing

PR #342, opened 2020-08-24
Place page small fixes

beets:
* Update ranking labels so the rank direction is clearer
* use geoOverlaps for child places, to get some zip codes which might go outside of the containing place's bounds

### DIFF
--- a/server/routes/api/place.py
+++ b/server/routes/api/place.py
@@ -16,18 +16,46 @@ import collections
 import json
 import services.datacommons as dc
 
-from flask import Blueprint, request, jsonify
+from flask import Blueprint, jsonify, request, Response
 
 from cache import cache
 from services.datacommons import fetch_data
-from routes.api.stats import get_stats_wrapper
+import routes.api.stats as stats_api
 
-WANTED_PLACE_TYPES = [
-    "Country", "State", "Province", "County", "City", "Town", "Village",
-    "Borough", "CensusZipCodeTabulationArea", "EurostatNUTS1", "EurostatNUTS2",
-    "EurostatNUTS3", "AdministrativeArea1", "AdministrativeArea2",
-    "AdministrativeArea3", "AdministrativeArea4", "AdministrativeArea5"
-]
+# Place types to keep for list of child places, keyed by parent place type.
+WANTED_PLACE_TYPES = {}
+WANTED_PLACE_TYPES['Country'] = [
+                      "State",
+                      "Province",
+                      "County",
+                      "EurostatNUTS1",
+                      "EurostatNUTS2",
+                      "AdministrativeArea1"]
+WANTED_PLACE_TYPES['State'] = [
+                      "Province",
+                      "County"]
+WANTED_PLACE_TYPES['County'] = [
+                      "City",
+                      "Town",
+                      "Village",
+                      "Borough"]
+ALL_WANTED_PLACE_TYPES = ["Country",
+                      "State",
+                      "Province",
+                      "County",
+                      "City",
+                      "Town",
+                      "Village",
+                      "Borough",
+                      "CensusZipCodeTabulationArea",
+                      "EurostatNUTS1",
+                      "EurostatNUTS2",
+                      "EurostatNUTS3",
+                      "AdministrativeArea1",
+                      "AdministrativeArea2",
+                      "AdministrativeArea3",
+                      "AdministrativeArea4",
+                      "AdministrativeArea5"]
 
 # These place types are equivalent: prefer the key.
 EQUIVALENT_PLACE_TYPES = {
@@ -53,6 +81,23 @@ RANKING_STATS = {
 bp = Blueprint("api.place", __name__, url_prefix='/api/place')
 
 
+@cache.memoize(timeout=3600 * 24)  # Cache for one day.
+def get_property_value(dcid, prop, out=True):
+    return dc.get_property_values([dcid], prop, out)[dcid]
+
+
+@cache.memoize(timeout=3600 * 24)  # Cache for one day.
+def get_place_type(place_dcid):
+    place_types = get_property_value(place_dcid, 'typeOf')
+    # We prefer to use specific type like "State", "County" over
+    # "AdministrativeArea"
+    chosen_type = None
+    for place_type in place_types:
+        if not chosen_type or chosen_type.startswith('AdministrativeArea'):
+            chosen_type = place_type
+    return chosen_type
+
+
 @bp.route('/name')
 def name():
     """Get place names."""
@@ -68,7 +113,7 @@ def name():
     for dcid in dcids:
         values = response[dcid].get('out')
         result[dcid] = values[0]['value'] if values else ''
-    return json.dumps(result)
+    return Response(json.dumps(result), 200, mimetype='application/json')
 
 
 @bp.route('/statsvars/<path:dcid>')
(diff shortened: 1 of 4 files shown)


beets:
image

server/routes/api/place.py
```diff
@@ -139,7 +139,7 @@ def child_fetch(dcid):
         '/node/property-values',
         {
             'dcids': [dcid],
-            'property': 'containedInPlace',
+            'property': 'geoOverlaps',
```
shifucun, line 142: might also change the function name "child_fetch" to overlap_geo_fetch?

beets:
Sorry this morphed into a bigger fix... now also merges equivalent place types, filters geonames and places with 0 population. Also fixes bug where we always used 2018 for the population, which doesn't exist for international geos.

ptal

beets:
Some results against prod mixer:
image

image

image

I didn't create equivalent groupings for Eurostat and AA because there were some weird mappings (e.g. England is AA1 but Scotland, Wales, and Northern Ireland are AA1)

server/routes/api/place.py
```diff
@@ -16,13 +16,21 @@
...
+from flask import Blueprint, jsonify, request, Response
 
 from cache import cache
 from services.datacommons import fetch_data
-from routes.api.stats import get_stats_wrapper
+import routes.api.stats as stats_api
 
-WANTED_PLACE_TYPES = [
+# Place types to keep for list of child places, keyed by parent place type.
+WANTED_PLACE_TYPES = {}
```
shifucun, line 26: can merge this into one statement?

WANTED_PLACE_TYPES = {
  'Country': [],
  'State': [],
  'County': []
}

server/routes/api/place.py
```diff
@@ -68,7 +93,7 @@ def name():
     for dcid in dcids:
         values = response[dcid].get('out')
         result[dcid] = values[0]['value'] if values else ''
-    return json.dumps(result)
+    return Response(json.dumps(result), 200, mimetype='application/json')
```
shifucun, line 96: curious what is the difference with the bare json.dump

server/routes/api/place.py
```diff
@@ -114,36 +141,54 @@ def child(dcid):
...
+            if place_type in wanted_types and place_pop > 0:
                 result[place_type].append({
                     'name': place.get('name', place['dcid']),
                     'dcid': place['dcid'],
-                    'pop': pop.get(place['dcid'], 0)
+                    'pop': place_pop,
                 })
+
     # Filter equivalent place types
     for (preferred, equivalent) in EQUIVALENT_PLACE_TYPES.items():
```
shifucun, line 182: Please add some comments for the logic.

server/routes/api/stats.py
```diff
@@ -26,6 +26,27 @@
 
 # TODO(shifucun): add unittest for this module
 
+def get_stats_latest(dcid_str, stats_var):
```
shifucun, line 29: nice change!

server/routes/api/place.py
```diff
@@ -16,13 +16,21 @@
...
+import routes.api.stats as stats_api
 
-WANTED_PLACE_TYPES = [
+# Place types to keep for list of child places, keyed by parent place type.
+WANTED_PLACE_TYPES = {}
+WANTED_PLACE_TYPES['Country'] = [
+    "State", "Province", "County", "EurostatNUTS1", "EurostatNUTS2",
+    "AdministrativeArea1"
+]
+WANTED_PLACE_TYPES['State'] = ["Province", "County"]
```
pradh, line 31: I want to spend a few minutes to lay out the choices here.

1. A place may have equivalent place types, which one should we pick from among them?
      - AA1/State/Province: pick AA1?   (this makes it general across countries)
      - County/AA2: pick AA2?
      - City/Village/Borough/Town: pick City?
      - Neighborhood
      - Zip Codes
      Note1: EurostatNUTS classification is separate from above, but may overlap (e.g. NUTS2 might AA1, NUTS3 might be AA2 for some geos)
      Note2: There are two general types "AdministrativeArea" and "Place" (this is often a City equivalent)

2. A place may have multiple types, some equivalent and some not. Do we pick one type for a place, or multiple (one for each equivalent set)?
      - if we pick multiple then the same node may occur twice in the children listing, which is odd.
      - were we to pick one place, the preference list should be in kPlaceTypesWithPopObs.

3. For a given parent type, which child types to pick/drop?
      - Looks like this is what WANTED_PLACE_TYPES does. But I suspect the key should consider all State-equivalent types, County-equivalent types?

4. For a parent node, do we group geos by "equivalent" types?
      - If one place had a type AA2 and another had County we should probably group them.

server/routes/api/place.py
```diff
@@ -43,16 +51,33 @@
...
+
+
+@cache.memoize(timeout=3600 * 24)  # Cache for one day.
+def get_place_type(place_dcid):
+    place_types = get_property_value(place_dcid, 'typeOf')
+    # We prefer to use specific type like "State", "County" over
+    # "AdministrativeArea"
+    chosen_type = None
+    for place_type in place_types:
+        if not chosen_type or chosen_type.startswith('AdministrativeArea'):
```
pradh, line 76: We want to skip Place too/

server/routes/api/place.py
```diff
@@ -16,13 +16,21 @@
...
 from cache import cache
 from services.datacommons import fetch_data
-from routes.api.stats import get_stats_wrapper
+import routes.api.stats as stats_api
 
-WANTED_PLACE_TYPES = [
+# Place types to keep for list of child places, keyed by parent place type.
+WANTED_PLACE_TYPES = {}
+WANTED_PLACE_TYPES['Country'] = [
+    "State", "Province", "County", "EurostatNUTS1", "EurostatNUTS2",
```
pradh, line 28: Why County?

server/routes/api/place.py
beets: Chatted offline, what we'll do is:

1. ignore AAx for non US, State/County/City for US. Eurostat* will stay separate

2. we'll show the place multiple times for non-equivalent types

3. pick a preferred type from an equivalent set, before key'ing off the list. so the list should have all equivalent place types
Equivalent types for US (geoId/... or zip/...)
State == AA1
County == AA2
City == Town / Village / Burrough
Everywhere else, pick AAx

4. For non-US geos, ignore Place and AA if parent AA2 onwards - group under Place

server/routes/api/place.py
pradh, line 28: We can drop Province I think

server/routes/api/place.py
pradh: Thanks for the summary.

4. To confirm, we want to ignore child AA + Place for parent AA1 and Country, but keep for parent AA2, AA3, ...?

pradh (dismissed):
This is improvement from before. Please feel free to defer the discussed changes to future PR.

server/routes/api/place.py
beets: done

server/routes/api/place.py
beets: the browser recognizes the content as json, and displays nice formatting and folding features :) it was helpful for debugging.

server/routes/api/place.py
beets: added

server/routes/api/stats.py
beets: thanks, we were doing this too often :)

server/routes/api/place.py
beets: will do this in a follow up!

server/routes/api/place.py
beets: done

server/routes/api/place.py
beets: good catch. removed..

beets: done

server/routes/api/place.py
shifucun: nice, can gradually move the rest of the reponse

PR #343, opened 2020-08-26
Fix key error exception and update children place selection logic

shifucun:
Note there are some auto-formatting by yapf on the modified files.

webdriver tests revealed a bug in current children places selection. With new places from wikidata, there are more children place types. For example, California has a village children that has 0 population, thus making the "containing" places chart not showing.

Added a fix to pick the child place type that has the highest average population.

[1] http://staging.datacommons.org/api/place/child/geoId/06

### DIFF
--- a/static/js/place/place_template.tsx
+++ b/static/js/place/place_template.tsx
@@ -783,12 +783,24 @@ class Chart extends Component<ChartPropType, ChartStateType> {
             } else if (this.placeRelation === placeRelationEnum.CONTAINING) {
               placesPromise = this.props.childPlacesPromise.then(
                 (childPlaces) => {
-                  // TODO(boxu): figure out a better way to pick child places.
+                  // Choose the place type that has the highest average
+                  // population
+                  let avgPop = 0;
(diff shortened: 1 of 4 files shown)


static/js/place/place_template.tsx
```diff
@@ -783,12 +783,24 @@ class Chart extends Component<ChartPropType, ChartStateType> {
             } else if (this.placeRelation === placeRelationEnum.CONTAINING) {
               placesPromise = this.props.childPlacesPromise.then(
                 (childPlaces) => {
-                  // TODO(boxu): figure out a better way to pick child places.
+                  // Choose the place type that has the highest average
```
pradh, line 786: How do we handle places with multiple types?

shifucun: The children places is generated in https://github.com/datacommonsorg/website/blob/master/server/routes/api/place.py#L157-L170

- A map from place type to list of places are first generated from containedInPlace relation 
- If there are equivalent place types, then the preferred one is chosen.

PR #344, opened 2020-08-26
add imported data from the interns

qlj-lijuan:
Added new specs to include the data imported by the interns. 
The new statsVars are shown in the red boxes:
image
image
image
image


tools/pv_tree_generator/constants.py
```diff
@@ -42,6 +42,23 @@
     ('NAICS/99', 'Unclassified')
 ])
 
+NACE = collections.OrderedDict([
```
shifucun, line 45: Some of the title is too long to fit in the menu region. Where do you get them, can discuss if they have shorter names.

tools/pv_tree_generator/pop_obs_spec_common.textproto
```diff
@@ -33,7 +43,14 @@ spec: {
   obs_props: {
     mprop: "count"
     stat_type: "measuredValue"
+    name: "Count"
   }
+  #obs_props: {
+  #  mprop: "lifeExpectancy"
+  #  stat_type: "measuredValue"
+  #  name: "Life Expectancy"
+  #} # Currently not shown in the tree due to the filtered
```
shifucun, line 52: what range is filtered here? If possible, we can disable the filter for this.

tools/pv_tree_generator/pop_obs_spec_common.textproto
```diff
@@ -2024,4 +2078,83 @@ spec: {
...
+  }
+  dpv: {
+    prop: "legislationPassedBy"
+    val: "GovernmentOrganization"
+  }
+  obs_props: {
+    mprop: "policyExtent"
+  }
+  name: "COVID-19 Policy"
+  vertical: "Health"
```
shifucun, line 2098: @beets @tjann any suggestion on the vertical for this observation?

qlj-lijuan:
Revise the logic of when to add the new level, fixed issue https://github.com/datacommonsorg/website/issues/345
image

tools/pv_tree_generator/constants.py
qlj-lijuan: Got it from @eftekhari-mhs. I'll check with her to see if we can make the title shorter.

tools/pv_tree_generator/pop_obs_spec_common.textproto
qlj-lijuan: Disabled the comments. 
However, the statsVars are not showing on the tree. The reason is: we're only showing specific age ranges: eg: < 5, 5-17, 18-24. But the age ranges for lifeExpectancy are: 1, 2, 3, 4, ..; These values would be filtered when building the tree.

tools/pv_tree_generator/pop_obs_spec_common.textproto
tjann: Health seems good. One alternative is to make COVID it's own vertical.

tools/pv_tree_generator/stat_config.proto
```diff
@@ -28,6 +28,7 @@ message ObsProp {
   optional string mqual = 3;
   optional string mdenom = 4;
   optional string name = 6;
+  optional bool same_level = 7;
```
shifucun, line 31: very nice change

tools/pv_tree_generator/pop_obs_spec_common.textproto
tjann: Can we sort the values under Health?

tjann: And make plural/singular case consistent?

beets: Pulling covid out into a new category sgtm

qlj-lijuan:
Add a new category: COVID-19
image

tools/pv_tree_generator/pop_obs_spec_common.textproto
qlj-lijuan: Added a new category and always use plural cases

tools/pv_tree_generator/pop_obs_spec_common.textproto
shifucun: We have similar treatment for different ages: https://github.com/datacommonsorg/website/blob/master/tools/pv_tree_generator/constants.py#L89-L100, would this work? Can add a new age list

qlj-lijuan: It would work if we put lifeExpectancy under a different category or at least in different specs. 
We should have two "age" nodes. Each has its own age list.

shifucun: okay, might leave a todo for this.

PR #346, opened 2020-08-26
Use AppEngine flex env with 3G memory

shifucun:


cloudbuild.deploy.yaml
```diff
@@ -1,54 +1,54 @@
...
+  - id: deploy
+    name: "gcr.io/cloud-builders/gcloud"
+    entrypoint: /bin/bash
+    args:
+      - -c
+      - |
+        set -x
+        cd server
+        gcloud config set project datcom-browser-staging
+        # bdist_wheel for grpcio takes a long time, seeting timeout to be 1200s
```
beets, line 29: nit: s/seeting/setting

what happens if we hit the timeout? should we continue with the deployment?

server/app_prod.yaml
```diff
@@ -12,11 +12,19 @@
...
+entrypoint: gunicorn -b :$PORT main:app
 
-instance_class: F4_1G
-automatic_scaling:
-  min_instances: 3
+runtime_config:
+  python_version: 3
+
+manual_scaling:
+  instances: 10
```
beets, line 23: should we get more instances for prod vs staging?

cloudbuild.deploy.yaml
shifucun: Then cloudbuild would error out. Tried locally it the gcloud build takes < 10min, so this should be safe. Will need to watch for the first few times deployment

server/app_prod.yaml
shifucun: Yeah. I don't know a good number to set for now. Might let it run for a while and adjust if we saw performance issues?

beets: ok

cloudbuild.deploy.yaml
beets: ok, let's keep an eye on it

PR #352, opened 2020-08-26
Add current place to nearby place list and filter/sort nearby places

shifucun:
image

### DIFF
--- a/server/routes/api/place.py
+++ b/server/routes/api/place.py
@@ -49,6 +49,8 @@ EQUIVALENT_PLACE_TYPES = {
 
 CHILD_PLACE_LIMIT = 50
 
+MIN_POP = 5000
+
 # Contains statistical variable and the display name used for place rankings.
 RANKING_STATS = {
     'Count_Person': 'Largest Population',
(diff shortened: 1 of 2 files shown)


shifucun:
> lgtm, but could you merge in after #342? i prefer the fix there for getting population, which just uses the latest value, not necessarily 2018 (which doesn't always exist for non-US geos)

sg, I will wait for that, also looking at that now.

shifucun:
Now using the latest population

server/routes/api/place.py
```diff
@@ -137,7 +140,7 @@ def child(dcid):
     for place_type in child_places:
         child_places[place_type].sort(key=lambda x: x['pop'], reverse=True)
         child_places[place_type] = child_places[place_type][:CHILD_PLACE_LIMIT]
-    return json.dumps(child_places)
+    return Response(json.dumps(child_places), 200, mimetype='application/json')
```
beets, line 143: It doesn't work for one of these handlers which is being called by another function. Please check that all the handlers still return correctly in the chrome network inspector.

shifucun: good to know. Checked a few pages and urls, these changes are okay.

PR #353, opened 2020-08-26
fix y-axis tick label format

qlj-lijuan:
set the format the y-axis tick label when the number is negative


shifucun (dismissed):
thanks for fix!

PR #354, opened 2020-08-26
add specs for covid test

qlj-lijuan:


qlj-lijuan:
> would be nice to include a screenshot

Here's the screenshot 
image

tjann:
Tell me if this proposal is too "out-there" or not inline with the rest of the PV tree. But from a general user standpoint, I think it's ok to forgo some schema reflections here.


1. s/Medical Condition Incident/Patient Status/
2. s/Medical Test/Tests/
3. This one is tricky... see https://github.com/datacommonsorg/data/pull/243/files#diff-67900f64a2ee2fca41698113dad74ec7R33-R47 we could choose the MedicalTest and nest that under Test/ as "PCR Tests Performed" or we can choose the Person and nest under Patient Status as "PCR-Tested". Or both.

qlj-lijuan:
> Tell me if this proposal is too "out-there" or not inline with the rest of the PV tree. But from a general user standpoint, I think it's ok to forgo some schema reflections here.
> 
> 1. s/Medical Condition Incident/Patient Status/
> 2. s/Medical Test/Tests/

1 and 2 can be handled by setting the super enum 
image

> 3. This one is tricky... see https://github.com/datacommonsorg/data/pull/243/files#diff-67900f64a2ee2fca41698113dad74ec7R33-R47 we could choose the MedicalTest and nest that under Test/ as "PCR Tests Performed" or we can choose the Person and nest under Patient Status as "PCR-Tested". Or both.

We don't have a unified method to handle this type of change, but we can handle this one specifically.

tjann:
I meant substitute instead of nest under for 1-2, but this is definitely better. Can we still sort everything alphabetically? Can the sort be default ?


For 3-- If we encode both poptype schemas correctly, will that solve the issue instead of hard coding?

qlj-lijuan:
> I meant substitute instead of nest under for 1-2, but this is definitely better. Can we still sort everything alphabetically? Can the sort be default ?

Yes. Sorted them. 

image


> For 3-- If we encode both poptype schemas correctly, will that solve the issue instead of hard coding?

If it follows the definition of http://datacommons.org/browser/CumulativeCount_MedicalTest_ConditionCOVID_19_Negative, with the only difference in "testResult", we should be able to handle it without hard coding.

PR #358, opened 2020-08-27
Use Response for api handlers in place.py, and use helpers for other

beets:
callers.


shifucun (dismissed):

beets:
nice, webdriver tests caught a bug! fixed the test, ptal

shifucun:
> nice, webdriver tests caught a bug! fixed the test, ptal

nice!

PR #359, opened 2020-08-27
fix dividing by population of 0

chejennifer:
When doing per capita calculation, if population is 0, we want to set the result for that calculation to be 0, not NaN.

### DIFF
--- a/static/js/shared/data_fetcher.test.ts
+++ b/static/js/shared/data_fetcher.test.ts
@@ -191,29 +191,28 @@ test("fetch stats data with per capita with population size 0", () => {
     }
   });
 
-  return fetchStatsData(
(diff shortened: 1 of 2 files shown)


static/js/shared/data_fetcher.test.ts
```diff
@@ -160,6 +160,89 @@ test("fetch stats data", () => {
   });
 });
 
+test("fetch stats data with per capita", () => {
```
beets, line 163: please update the test name with the subject of this test (that we're testing per capita with population 0)

you can also simplify this test by doing it for only one state.

chejennifer: done

PR #362, opened 2020-08-28
[place charts] Handle missing data for some places

beets:
With some of new comparison charts, we might have data only for some geos. Skip geos so the charts will still get drawn.

@chejennifer fyi


beets:
ptal when you get a chance, thanks :)

shifucun (approved):

PR #364, opened 2020-08-28
[place charts] remove non-ratio comparison charts

beets:
Some other updates
* Load config from json on dev
* Drop unused topic from chart config 
* Move covid charts to the top of health topic
* Added a few race & disaster charts
* Some duplicate charts added from overview to topic pages -- will have a more robust solution forthcoming

Sorry for the large diff in the config json - I applied formatting to the file...

fyi @chejennifer


server/chart_config.json
```diff
@@ -398,41 +424,82 @@
...
-              "Count_Death_Neoplasms",
-              "Count_Death_DiseasesOfTheRespiratorySystem",
-              "Count_Death_ExternalCauses",
-              "Count_Death_DiseasesOfTheNervousSystem"
+              "CumulativeCount_MedicalConditionIncident_COVID_19_ConfirmedOrProbableCase"
             ],
-            "source": "CDC",
-            "url": "https://www.cdc.gov/"
+            "source": "The New York Times",
+            "url": "https://www.nytimes.com/article/coronavirus-county-data-us.html"
```
shifucun, line 440: maybe in later PR. the source domain is in the cache now, can use that instead of having them in config here.

server/routes/api/chart.py
```diff
@@ -116,12 +117,17 @@ def config(dcid):
     """
     Get chart config for a given place.
     """
+    chart_config = current_app.config['CHART_CONFIG']
+    if os.environ.get('FLASK_ENV') == 'development':
```
shifucun, line 121: to understand, in this way, when there is local config change, the server can pick it up without restart?

shifucun (dismissed):

server/routes/api/chart.py
beets: yes, that helps with config changes. though the cache is still in effect on the results, i haven't found a good way to selectively turn that off beyond commenting it out :)

server/chart_config.json
beets: yea, i wanted to start on that soon.

beets:
ptal, had to fix the webdriver tests

shifucun (approved):

PR #384, opened 2020-09-03
Use different Google Analytics account for non-prod instances

beets:
I've tried setting up filters within GA but that doesn't seem to apply, and staging data is mixed up with prod. This should keep things much cleaner.


shifucun (approved):
very nice!

PR #389, opened 2020-09-09
data for latest data with all data

chejennifer:
When getting data for a chart, get it for the latest date for which there is data for all the stat vars in the chart. If there is no such date, choose the latest date.

### DIFF
--- a/server/routes/api/chart.py
+++ b/server/routes/api/chart.py
@@ -97,6 +97,22 @@ def keep_latest_data(timeseries):
     max_date = max(timeseries)
     return { max_date: timeseries[max_date] }
 
+    
+def keep_necessary_data(timeseries, dates):
+    """Only keep the necessary timestamped data from the timeseries
+
+    Args:
+        timeseries: GetStats timeseries, dictionary of date: value
+        dates: set of dates needed for this timeseries
+
+    Returns:
+        dictionary of date: value
+    """
+    data = {}
+    for date in dates:
+        data[date] = timeseries[date]
+    return data
+
 
 def get_landing_page_data(dcid):
     response = dc_service.fetch_data(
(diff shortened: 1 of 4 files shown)


server/routes/api/chart.py
```diff
@@ -97,6 +97,22 @@ def keep_latest_data(timeseries):
     max_date = max(timeseries)
     return { max_date: timeseries[max_date] }
 
+    
+def keep_necessary_data(timeseries, dates):
```
beets, line 101: Is there a more accurate word to use? Necessary is very broad, perhaps specified?

server/routes/api/chart.py
```diff
@@ -110,6 +126,70 @@ def get_landing_page_data(dcid):
...
 
+def get_date_to_add_for_chart(chart, sv_data):
+    """Get the date for which data is needed for a specific chart
+    
+    Args:
+        chart: the chart object that we currently care about (a single chart object from chart_config.json)
+        sv_data: the object returned from get_landing_page_data
+    
+    Returns:
+        date as a string
```
beets, line 137: Note that this could potentially return an empty string (and perhaps we should return None instead).

server/routes/api/chart.py
```diff
@@ -110,6 +126,70 @@ def get_landing_page_data(dcid):
     )
     return response
 
+def get_date_to_add_for_chart(chart, sv_data):
+    """Get the date for which data is needed for a specific chart
```
beets, line 130: Could you add more description here? Which dates end up being picked?

server/routes/api/chart.py
```diff
@@ -110,6 +126,70 @@ def get_landing_page_data(dcid):
...
+        dates_initialized = True
+    sorted_dates = sorted(dates)
+    date_to_add = ''
+    if len(sorted_dates) > 0:
+        date_to_add = sorted_dates[-1]
+    return date_to_add
+    
+    
+def get_dates_for_stat_vars(chart_config, sv_data):
+    """Get the dates needed for each stat var
```
beets, line 157: could you add more here as well? mention that this checks all charts for the stat var.

server/routes/api/chart.py
beets: I'm also surprised we are not inspecting the chart at all - different chart types will have different requirements (some latest, some all dates). Perhaps we can continue ignoring given that we are going to be generating the multiple types of charts now per stat var, but we should rename this function in that case.

server/tests/chart_test.py
```diff
@@ -0,0 +1,180 @@
...
+import unittest
+import json
+from unittest.mock import patch
+
+from main import app
+
+class TestRoute(unittest.TestCase):
+    
+    @patch('routes.api.chart.dc_service.fetch_data')
+    @patch('routes.api.chart.place_api.statsvars')
```
beets, line 24: nice!

static/js/shared/data_fetcher.test.ts
```diff
@@ -160,6 +161,284 @@ test("fetch stats data", () => {
...
+        },
+      });
+    }
+  });
+
+  return fetchStatsData(["geoId/05", "geoId/06"], ["Count_Person"]).then(
+    (data) => {
+      expect(data).toEqual({
+        data: {
+          Count_Person: {
```
beets, line 194: Can we factor out this data object so that we don't have to repeat the entire chunk?

static/js/shared/data_fetcher.ts
```diff
@@ -53,18 +53,21 @@ class StatsData {
     [statsVar: string]: StatApiResponse;
   };
   sources: Set<string>;
+  dateToSelect: string;
```
beets, line 56: Maybe: latestCommonDate or something to that effect?

static/js/shared/data_fetcher.ts
```diff
@@ -420,16 +449,39 @@ function fetchStatsData(
...
+          } else {
+            dates[date] = 1;
+          }
         }
       }
     }
     result.dates = Object.keys(dates);
     result.dates.sort();
+    result.dateToSelect = result.dates[result.dates.length - 1];
+    const numPlacesWithData: number = places.filter(
```
beets, line 473: please add a comment describing this block.

static/js/shared/data_fetcher.ts
```diff
@@ -400,8 +428,9 @@ function fetchStatsData(
   }
 
   return Promise.all(apiDataPromises).then((allResp) => {
-    const result = new StatsData(places, statsVars, [], {});
-    const dates: { [key: string]: boolean } = {};
+    const result = new StatsData(places, statsVars, [], {}, "");
+    const dates: { [key: string]: number } = {};
```
beets, line 432: please rename this object - describe what it contains

static/js/shared/data_fetcher.ts
```diff
@@ -420,16 +449,39 @@ function fetchStatsData(
...
+            dates[date] = dates[date] + 1;
+          } else {
+            dates[date] = 1;
+          }
         }
       }
     }
     result.dates = Object.keys(dates);
     result.dates.sort();
+    result.dateToSelect = result.dates[result.dates.length - 1];
```
beets, line 472: Move this to line 485, so it's clear that this is the default case when we can't find a better date.

beets (changes requested):
nice, very thorough changes! some comments below.

server/routes/api/chart.py
chejennifer: You're right, the name is a little misleading. This is a function just to find the latest common date for a chart. We are not inspecting the chart because if a stat var needs all dates, there is a separate check for that elsewhere in the code.

beets (dismissed):
thanks! could you resolve the merge conflict, then we're good to go.

beets (approved):

PR #403, opened 2020-09-11
Adds ability to go back in time

edumorlom:
Changes:
1. Server now returns the entire timeseries.
2. Adds dropdown to select date with small CSS changes.
3. When new date is selected, map is re-rendered.

image


static/js/tools/choropleth/choropleth.tsx
```diff
@@ -129,6 +142,21 @@ class ChoroplethMap extends Component<PropsType, StateType> {
...
+   * @param dates: a list of dates in ISO 8601 format.
+   * Example: "2020-01-02".
+   */
+  public getMaxDate = (dates: string[]): string => {
+    if (!dates.length) {
+      return "";
+    }
+
+    return dates.reduce((a, b) => {
+      return new Date(a) > new Date(b) ? a : b;
```
jeffreyoldham, line 156: Is > or >= the proper operator?  It depends on the reduce specification.

server/routes/api/choropleth.py
```diff
@@ -174,10 +173,11 @@ def choropleth_geo():
             # Process Statistical Observation if valid.
             if ('data' in population_by_geo.get(geo_id, [])
                     and population_by_geo[geo_id]['data']):
```
jeffreyoldham, line 175: Is this check necessary given that `get_data` can handle empty input?

server/routes/api/choropleth.py
```diff
@@ -174,10 +173,11 @@ def choropleth_geo():
...
-                # Grab the latest available data.
-                latest_data = get_latest_data(population_by_geo[geo_id])
-                if latest_data:
-                    geo_feature["properties"]["pop"] = latest_data
+                # Grab the data.
+                data = get_data(population_by_geo[geo_id])
+                if data:
+                    max_date = max(data)
+                    geo_feature["properties"]["pop"] = data[max_date]
```
jeffreyoldham, line 180: I am not following the logic.  I thought the Python code would now return either the data for the desired date xor data for all dates.  This seems to return the latest date only.  What am I missing?

jeffreyoldham (commented):
Thanks.

server/routes/api/choropleth.py
edumorlom: The python code returns the entire time series except for population.
We only return the latest population, for now.

**data -> return entire time series data.
population -> only returns the latest dataset.**

Front-end expects a single integer for population, not a time series so I have to calculate the max_date before returning.

server/routes/api/choropleth.py
edumorlom: thank you. you're right, it's not necessary, if 'data' is not in payload, the value will be skipped because get_data returns None.
So check is not necessary

static/js/tools/choropleth/choropleth.tsx
edumorlom: should be >= i believe, thanks. fixed.

server/routes/api/choropleth.py
edumorlom: Added a TODO.

jeffreyoldham (dismissed):
Thanks.

server/routes/api/choropleth.py
```diff
@@ -156,48 +149,48 @@ def choropleth_geo():
...
-                    population_by_geo[geo_id]['data']):
-                # Grab the latest available data.
-                latest_data = get_latest_data(population_by_geo[geo_id])
-                if latest_data:
-                    geo_feature["properties"]["pop"] = latest_data
+                coerce_geojson_to_righthand_rule(
+                                    geojson['coordinates'],
+                                    geojson['type']))
+
+            data = get_data(population_by_geo[geo_id])
```
beets, line 175: Should previous checks be carried over (testing for geo_id in population_by_geo)

server/routes/api/choropleth.py
```diff
@@ -156,48 +149,48 @@ def choropleth_geo():
...
+            # TODO(edumorales): return all populations
+            # get the selectedDate population in front-end.
+            if data:
+                max_date = max(data)
```
beets, line 179: Please add a TODO for this - the date chosen across places should be the same (not the latest date per place).

server/routes/api/choropleth.py
```diff
@@ -156,48 +149,48 @@ def choropleth_geo():
...
-                latest_data = get_latest_data(population_by_geo[geo_id])
-                if latest_data:
-                    geo_feature["properties"]["pop"] = latest_data
+                coerce_geojson_to_righthand_rule(
+                                    geojson['coordinates'],
+                                    geojson['type']))
+
+            data = get_data(population_by_geo[geo_id])
+            # TODO(edumorales): return all populations
+            # get the selectedDate population in front-end.
```
beets, line 177: What do you consider the front-end (python code?) this comment is redundant since then (or drop the "in front-end")

static/js/tools/choropleth/choropleth.tsx
```diff
@@ -80,12 +81,17 @@ class ChoroplethMap extends Component<PropsType, StateType> {
 
     Promise.all([geoPromise, valuePromise]).then(
       (values) => {
+        // Coordinates for the map.
         const geoJson = values[0].data[0];
-        const geoIdToValues = values[1].data[0];
+        // All the timeseries data for all places.
+        const data = values[1].data[0];
+        // The latest value for all places.
+        const geoIdToValues = this.filterByDate(data, "latest");
```
beets, line 89: rather than displaying latest, let's just show the latest date in the drop down selected by default

static/js/tools/choropleth/choropleth.tsx
```diff
@@ -361,6 +389,70 @@ class ChoroplethMap extends Component<PropsType, StateType> {
...
+   *
+   * Changes the date and updates the values on the state.
+   * This causes a re-render that makes the map be updated
+   * to reflect the new data.
+   * @param event: an 'onChange' event.
+   */
+  public changeDate = (event: { target: { value: string } }): void => {
+    // Get the newly selected date as a string.
+    const newDate: string = event.target.value;
+    console.log(newDate);
```
beets, line 408: please delete debugging statements

static/js/tools/choropleth/choropleth.tsx
```diff
@@ -450,6 +542,45 @@ class ChoroplethMap extends Component<PropsType, StateType> {
...
+      return (
+        <option key={date} value={date}>
+          {date}
+        </option>
+      );
+    });
+
+    // Return the select component.
+    return (
+      <select onChange={this.changeDate} defaultValue={this.state["date"]}>
```
beets, line 573: it's good practice to keep the handler function names of the form "onDateChanged" or "onSelectedDateChanged"

static/js/tools/choropleth/choropleth.tsx
```diff
@@ -361,6 +389,70 @@ class ChoroplethMap extends Component<PropsType, StateType> {
...
           .text(title)
       );
+
+    // Place the date dropdown on the screen.
+    const datePickerComponent = this.datePicker();
+    ReactDOM.render(datePickerComponent, document.getElementById("date"));
+  };
+
+  /**
+   *
```
beets, line 399: nit: delete empty line

static/js/tools/choropleth/choropleth.tsx
```diff
@@ -361,6 +389,70 @@ class ChoroplethMap extends Component<PropsType, StateType> {
...
+    console.log(newDate);
+    const data = this.state["data"];
+    // Re-update this.values to only include data for that given date.
+    const geoIdToValue = this.filterByDate(data, newDate);
+    // Store the new date and the new values.
+    this.state.values = geoIdToValue;
+    this.setState({ date: newDate });
+
+    // Re-render component and map.
+    this.renderGeoMap();
```
beets, line 417: shouldn't updateGeoValues get called before renderGeoMap?

static/js/tools/choropleth/choropleth.tsx
```diff
@@ -361,6 +389,70 @@ class ChoroplethMap extends Component<PropsType, StateType> {
...
+    this.updateGeoValues();
+
+    // TODO(edumorales): for some reason, the setState auto-render
+    // was disabled by previous developers.
+    // Figure out why, and re-enable setState() to automatically re-render.
```
beets, line 422: Have you tried re-enabling? What happened? Ok to leave as a TODO

static/js/tools/choropleth/choropleth.tsx
```diff
@@ -361,6 +389,70 @@ class ChoroplethMap extends Component<PropsType, StateType> {
...
+    this.forceUpdate();
+  };
+
+  /**
+   *
+   * @param values: the data in the form of geoId->date->value.
+   * @param date: a date in ISO 8601 format string.
+   * Example: "2020-01-02".
+   */
+  public filterByDate = (
```
beets, line 433: this doesn't need to be public

static/js/tools/choropleth/choropleth.tsx
```diff
@@ -450,6 +542,45 @@ class ChoroplethMap extends Component<PropsType, StateType> {
     }
   };
 
+  /**
+   * Returns the complete select/dropdown component
+   * with the available dates as options.
+   */
+  public datePicker = (): JSX.Element => {
```
beets, line 549: Can this be made into a new component? Or the computation to get the set of displayed dates should be done only once - it will now happen on each render.

beets (changes requested):

server/routes/api/choropleth.py
```diff
@@ -26,11 +27,14 @@
 }
 
 # All choropleth endpoints are defined with the choropleth/ prefix.
-bp = flask.Blueprint("choropleth", __name__, url_prefix='/api/choropleth')
-
+bp = flask.Blueprint(
```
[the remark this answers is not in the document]
edumorlom: fixed. thanks

server/routes/api/choropleth.py
edumorlom: fixed.

edumorlom: fixed.

server/routes/api/choropleth.py
edumorlom: done.

server/routes/api/choropleth.py
edumorlom: fixed.

static/js/tools/choropleth/choropleth.tsx
edumorlom: "Latest" shows the latest date for each region individually.

Selecting "2020" would only show those regions with data from 2020.
Selecting "latest" would show the region's latest date, some regions would display data from 2019 others from 2020. Whatever the most up-to-date data is.

They have different functionality, would you want to get rid of "latest" functionality? Or maybe rename it?
Imagine one wants to see the map of covid that gets updated every day. This is a way of displaying all at once. Some updated yesterday, others updated today.

static/js/tools/choropleth/choropleth.tsx
edumorlom: fixed.

static/js/tools/choropleth/choropleth.tsx
edumorlom: fixed.

static/js/tools/choropleth/choropleth.tsx
edumorlom: fixed.

static/js/tools/choropleth/choropleth.tsx
edumorlom: It's now its own component that takes 3 props: dates, pickedDate and onSelectedDateChanged.

static/js/tools/choropleth/choropleth.tsx
edumorlom: I will try an open a separate PR for it if it's ok.
Don't wanna break things.

static/js/tools/choropleth/choropleth.tsx
edumorlom: Documentation for updateGeoValues says:

   ```* Requires geoJson map to be rendered and state values to be set.```

renderGeoMap() renders a blank map and then updateGeoValues draws the colors.
On all other instances, renderGeoMap() is called before updateGeoValues().

I have renamed renderGeoMap to drawBlankGeoMap.
and updateGeoValues to addColorToGeoMap.

Can change it to any other name if you'd like.

static/js/tools/choropleth/choropleth.tsx
edumorlom: fixed.

static/js/tools/choropleth/choropleth.tsx
beets: Latest date to me implies the next option (in the screenshot, 2019). We could also get widely variable ranges of dates - e.g. would it make sense to compare 2010 data with 2019? It is a correctness issue in my mind, and drawing a chart comparing data across dates is misleading.

@jeffreyoldham wdyt?

Separately, have you tested the behavior for some stat vars with very long histories? Covid, BLS are some that come to mind.

edumorlom: image
image

Long history works fine! 👍 🥇

edumorlom: @beets @jeffreyoldham

jeffreyoldham: I believe:
1) Users almost always want data from every state (or every country or every county).  They do not want to have to patch together data from different dates to see the overall picture.
2) In the real world, it is unlikely to get data from every entity at the same time.  E.g., each country conducts population censuses at different times.  Some are several years old.  We are spoiled with getting time-consistent per-US state data from US federal agencies, but we cannot expect the same for some data, e.g., counts of drivers licenses.

Thus, I think that users will be OK with having "latest" show data of different dates as long as:
1) it really is the latest data for that state and
2) the time gap for all the displayed data is not too large.  E.g., if one is looking at daily data, I think it would be hard to complain that some data is from Monday while others is from Friday.  However, if it is from a month ago, it might be too old to display.  If one is looking at monthly data, if a state's data is more than a year older than some other state's data, perhaps we should not show it.  If annual data, perhaps a decade gap may be too large.

In practice, I do not expect the gap between the newest and oldest data to be too large for US state level data, but the real world may disprove me.

Would it work to add a todo to the code?  We have identified an issue.  The feature seems like an improvement.  We expect we will produce good results almost all of the time, but there is a possibility we may not.  Should we address it in a separate pull request focussed on the time-data disparity issue?

jeffreyoldham (commented):
I added a comment.  I am not sure if I have to type this to have it get registered.

static/js/tools/choropleth/choropleth.tsx
beets: Thanks Jeffrey - that sounds good to me. I agree it is probably only a few corner cases where this might not be what users are looking for.

Sure, let's add a TODO to show the range of dates somewhere e.g. "showing data from 2010-2011"

static/js/tools/choropleth/choropleth.tsx
beets: that sounds good

beets:
please take a look at the failing tests (start with run_test.sh -f, and check that the remaining tests pass)

beets (approved):
great, thanks for making the changes!

PR #408, opened 2020-09-14
Update chart config to remove  type, axis, url and source

shifucun:


beets (approved):
much cleaner now! thanks!!

PR #409, opened 2020-09-15
Add date of observations to charts

chejennifer:
If the chart is not a line chart and does not have time as its axis, put in brackets in the title what date the data in the chart is from
Screenshot 2020-09-15 at 10 23 47 AM

### DIFF
--- a/static/css/place/place.scss
+++ b/static/css/place/place.scss
@@ -171,6 +171,7 @@ $sidebar-top-padding: 132px;  // Keep in sync with place.ts: Y_SCROLL_LIMIT
 
 .chart-container h4 .sub-title {
   color: #999;
+  padding-left: 2px;
 }
 
 .chart-container label {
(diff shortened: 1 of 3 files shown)


static/css/place/place.scss
```diff
@@ -171,6 +171,7 @@ $sidebar-top-padding: 132px;  // Keep in sync with place.ts: Y_SCROLL_LIMIT
 
 .chart-container h4 .sub-title {
   color: #999;
+  padding-left: 2px;
```
beets, line 174: Can we use .25rem instead (that way the space will scale with the base font size)

static/js/place/place_template.tsx
```diff
@@ -512,6 +512,7 @@ interface ChartStateType {
   dataPoints?: DataPoint[];
   dataGroups?: DataGroup[];
   elemWidth: number;
+  dateSelected?: string;
```
beets, line 515: since titleString doesn't seem to be used, let's remove it from the class

beets (changes requested):
thanks! just a couple of small changes

static/js/place/place_template.tsx
chejennifer: I don't see titleString, do you mean titleSuffix?

beets: oh yes, titlesuffix

beets (approved):
great, thanks!

PR #412, opened 2020-09-15
Update all python code using yapf

beets:
Also test for python lint errors, and add support for fixing that using run_test.py -f
Also reformatted yaml files using Prettier, though that is unenforced (cloudbuild.deploy.yaml already uses this style).

Might be easier to review by looking through each commit separately. Only the second commit should have logic changes, others were automatically done (first using yapf --recursive --i --style=google server/ tools/) and the latter using Prettier in vscode.


beets:
@chejennifer @edumorlom fyi - this might cause some amount of merge conflict, but should be easily fixable (run that yapf command to fix your local change before doing a merge/rebase)

server/routes/api/stats.py
```diff
@@ -20,12 +20,13 @@
 
 # Define blueprint
 bp = Blueprint(
```
shifucun, line 22: some how this one is not wrapped in one line..

run_test.sh
```diff
@@ -23,23 +23,28 @@ function run_npm_test {
...
+# Tests if lint is required
+function run_lint_test {
   cd static
   npm list eslint || npm install eslint
   if ! npm run test-lint; then
     echo "\nFix lint errors by running ./run_test.sh -f"
     exit 1
   fi
   cd ..
+  if ! yapf --recursive --diff --style=google server/ tools/; then
```
shifucun, line 35: need to install yapf first?

shifucun (dismissed):
Do we want to use 2 spaces instead of 4 space across for python? 2 spaces seems used in Google..

run_test.sh
beets: updated the script to install if not found

server/routes/api/stats.py
beets: Oh, that's odd. Manually made it one line

beets (commented):
I think we should stick with 4, so it's just default yapf, and it's also what we are using in the data repo. I also find 4 spaces easier to read with python code.

shifucun:
seeing some lint errors from the presubmit

beets:
> Hm, those are just warnings, but the error is in installing yapf through
> pip. I'll try a few other things...


The virtual environment was not created successfully because ensurepip is
not
available. On Debian/Ubuntu systems, you need to install the python3-venv
package using the following command.

apt-get install python3-venv

You may need to use sudo with that command. After installing the
python3-venv
package, recreate your virtual environment.

Failing command: ['/workspace/server/.env/bin/python3', '-Im', 'ensurepip',
'--upgrade', '--default-pip']

shifucun (approved):

beets:
yay! finally got tests passing

PR #413, opened 2020-09-16
append state code to place names when applicable

chejennifer:
For place name labels on the charts, append the state code if applicable. 

Changes:
- new endpoint for getting display names for places when given a list of dcids. For each dcid, get it's name and look for its parent places. If there is a parent place of type state, get the state code and append that to the name.
- in fetchStatsData, an additional call to get display names for each place.
- move test file for routes/api/place.py to tests so that it will be run and add test for my new endpoint

Screenshot 2020-09-16 at 9 37 34 AM


static/js/shared/data_fetcher.ts
```diff
@@ -439,11 +443,21 @@ function fetchStatsData(
...
+  const apiPromises: Promise<AxiosResponse<any>>[] = [];
+  apiPromises.push(...apiDataPromises);
+  apiPromises.push(displayNamesPromise);
+
+  return Promise.all(apiPromises).then((allResp) => {
     const result = new StatsData(places, statsVars, [], {}, "");
     const numOccurencesPerDate: { [key: string]: number } = {};
     const numStatVarsPerPlace: { [key: string]: number } = {};
+    const displayNameMapping =
+      allResp.length > 0 ? allResp[allResp.length - 1].data : {};
```
beets, line 460: Check for 'data' in resp before referencing it.

server/routes/api/place.py
```diff
@@ -459,3 +462,61 @@ def api_ranking(dcid):
...
+    """Get state code for places that have a US State as one of its parent places
+
+    Args:
+        dcid: dcid of place we are looking for the state code for
+
+    Returns:
+        state code as a string or None if there is no US state code
+    """
+    state_code = None
+    for parent_place in json.loads(parent_places(dcid)):
```
beets, line 478: we should update parent_places and get_parent_place to support a list of dcids to reduce the number of mixer calls.

server/routes/api/place.py
```diff
@@ -459,3 +462,61 @@ def api_ranking(dcid):
...
+        parent_dcid = parent_place['dcid']
+        place_types = parent_place['types']
+        for place_type in place_types:
+            if place_type in STATE_EQUIVALENTS:
+                iso_code = get_property_value(parent_dcid, 'isoCode')
```
beets, line 483: @shanth can we rely on isoCode to fetch state codes? At the moment, using this for US places only but would be great to expand internationally too.

beets (changes requested):
thanks!

server/routes/api/place.py
chejennifer: @pradh

pradh: Sorry I missed this. Using isoCode sounds good and is reasonable.

server/routes/api/place.py
```diff
@@ -464,28 +502,33 @@ def api_ranking(dcid):
     return Response(json.dumps(result), 200, mimetype='application/json')
 
 
-@cache.memoize(timeout=3600 * 24) # Cache for one day.
-def get_state_code(dcid):
-    """Get state code for places that have a US State as one of its parent places
+@cache.memoize(timeout=3600 * 24)  # Cache for one day.
+def get_state_code(dcids):
+    """Get state code for list of places
```
beets, line 507: i'd probably clarify that list of places should be state equivalents.

beets (approved):
thanks for the refactor! just one small comment, and please fix the merge conflict.

beets (approved):

beets:
thanks for the Earth fix

PR #414, opened 2020-09-16
[ranking] Add support for per-capita, migrate to get api

beets:
Also adds better error handling and cleans up client lint errors

The Mixer API was updated to get in https://github.com/datacommonsorg/mixer/commit/f6d3c6a61a9450837e00b91bf7289746a78a205a

The change doesn't affect prod since the ranking pages aren't launched yet, but it does break the screenshot tests.

### DIFF
--- a/server/routes/api/ranking.py
+++ b/server/routes/api/ranking.py
@@ -14,6 +14,7 @@
 """Place Ranking related handlers."""
 
 import flask
+import json
 import services.datacommons as dc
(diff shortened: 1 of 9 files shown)


beets:
Moved the update to requirements.txt here, so that this can be submitted first (alternatively, I can comment out the screenshot tests for ranking pages in #412 - lmk which you prefer).

shifucun (approved):

PR #415, opened 2020-09-16
[Place Page] Update chart layout to have blocks of primary + related place charts

shifucun:
- Use the new chart config that has the following update:
  - Remove chart type, axis
  - Remove provenance domain and name (they will be obtained from the data)
  - Use a flattened list with category and topic as field
  - Add isOverview field to simplify the config

- Make a ChartBlock component to hold a section of charts that displays the data for same stat vars but for different comparisons. This effectively removes the need for dropdown selection menu on each chart.

- Resolve all the related places lists before passing them into React components. The actual data is needed to determine if a chart should show or not. This also simplifies the code


server/__init__.py
```diff
@@ -58,7 +58,7 @@ def create_app():
     app.register_blueprint(stats.bp)
 
     # Load chart config
-    with open('chart_config.json') as f:
+    with open('chart_config_new.json') as f:
```
beets, line 61: should we do an in-place replacement of the config?

server/routes/api/chart.py
```diff
@@ -186,16 +185,47 @@ def get_dates_for_stat_vars(chart_config, sv_data):
...
 
 
-@bp.route('/config/<path:dcid>')
+def build_config(raw_config):
+    """Builds hierachical config based on raw config."""
+    category_map = {}
+    for config in raw_config:
+        is_overview = ('isOverview' in config and config['isOverview'])
+        if is_overview:
+            del config['isOverview']
```
beets, line 194: nitpick: if config['isOverview'] = False, then this isn't deleted.

beets: maybe it'll be cleaner to just build new config objects

static/js/place/place.ts
```diff
@@ -185,19 +185,31 @@ function renderPage(dcid: string) {
...
-          chartData: resolvedValues[0].data,
-          parentPlaces: resolvedValues[1],
-          childPlacesPromise,
-          similarPlacesPromise,
-          nearbyPlacesPromise,
+          chartConfig: chartConfigData.config,
+          chartData: chartConfigData.data,
+          parentPlaces: parentPlaces,
+          childPlaces: childPlaces,
+          similarPlaces,
```
beets, line 211: should these get explicit keys too?

static/js/place/place_template.tsx
```diff
@@ -466,6 +466,150 @@ class ChildPlace extends Component<ChildPlacePropType, unknown> {
   }
 }
 
+interface ChartBlockPropType {
```
beets, line 469: Nit: can we move the chart related components to a new file?

beets: we can also move after, in the interest of time.

static/js/place/place_template.tsx
```diff
@@ -466,6 +466,150 @@ class ChildPlace extends Component<ChildPlacePropType, unknown> {
...
+    conf.chartType = chartTypeEnum.LINE;
+    conf.title = "TREND";
+    result.push(conf);
+
+    conf = { ...config };
+    conf.chartType = chartTypeEnum.GROUP_BAR;
+    conf.axis = "PLACE";
+    if (placeType == "Country") {
+      // Containing place chart
+      conf.title = "Containing Places";
```
beets, line 558: Places within ___

static/js/place/place_template.tsx
```diff
@@ -466,6 +466,150 @@ class ChildPlace extends Component<ChildPlacePropType, unknown> {
...
+      conf.title = "Similar Places";
+      result.push(conf);
+    }
+    if (placeType != "City") {
+      // Children places
+      conf = { ...config };
+      conf.chartType = chartTypeEnum.GROUP_BAR;
+      conf.placeRelation = placeRelationEnum.CONTAINING;
+      conf.axis = "PLACE";
+      conf.title = "Children Places";
```
beets, line 597: Nit: Places within ____

static/js/place/place_template.tsx
```diff
@@ -466,6 +466,150 @@ class ChildPlace extends Component<ChildPlacePropType, unknown> {
...
+      result.push(conf);
+    } else {
+      // Parent places.
+      conf = { ...config };
+      conf.chartType = chartTypeEnum.GROUP_BAR;
+      conf.placeRelation = placeRelationEnum.CONTAINED;
+      conf.axis = "PLACE";
+      conf.title = "Parent Places";
```
beets, line 605: how about "Places that Contains ________"

beets: should also use percent

beets: also add a todo to use the right denominator for percapita

static/js/place/place_template.tsx
```diff
@@ -466,6 +466,150 @@ class ChildPlace extends Component<ChildPlacePropType, unknown> {
...
+    }
+    if (placeType != "City") {
+      // Children places
+      conf = { ...config };
+      conf.chartType = chartTypeEnum.GROUP_BAR;
+      conf.placeRelation = placeRelationEnum.CONTAINING;
+      conf.axis = "PLACE";
+      conf.title = "Children Places";
+      result.push(conf);
+    } else {
```
beets, line 599: maybe skip this for state too (there'll only be one parent place to compare)

beets (commented):
nice, i like rebuilding the config, though we should keep a small one around as a formatting reference perhaps, and have a test to check that outcome is the same. or add more places to the screenshot / webdriver tests so we can catch errors there.

server/tests/chart_test.py
```diff
@@ -90,84 +90,78 @@ def test_cache_necessary_dates(self, mock_stat_vars, mock_landing_page):
...
+                ]
             }]
-            response = app.test_client().get('api/chart/config/geoId/06')
+            response = app.test_client().get('api/chart/data/geoId/06')
             assert response.status_code == 200
             assert json.loads(response.data)['data'] == {
                 'geoId/06': {
                     'StatVar1': {
                         'data': {
+                            '2010': 200,
```
shifucun, line 113: @beets I want to confirm the original test intends to omit year 2010 as this is used for bar charts with latest date?

beets: yea, we were omitting data if the stat var was only used in bar charts. initially only keeping the latest date. then this was fixed in #389, where we are trying to get the best latest date to keep (so that the same latest date is chosen across stat vars). 

@chejennifer

server/routes/api/chart.py
shifucun: fixed to check if "isOverview" exists. Build a new object from empty would require knowing all the fields (and new fields if added). This way might be lightweight for change.

static/js/place/place_template.tsx
shifucun: sg, State is handled in the if condition

server/__init__.py
shifucun: moved chart_config_new.json to chart_config.json

static/js/place/place.ts
shifucun: done

static/js/place/place_template.tsx
shifucun: sg, added TODO

static/js/place/place_template.tsx
shifucun: Done

static/js/place/place_template.tsx
shifucun: Done

static/js/place/place_template.tsx
shifucun: Tried to use perCapita for place comparison charts and realized that need per chart config based on mprop like "age", "area per person" would not need perCapita. Added todo to update the config.

server/tests/chart_test.py
shifucun: Got it. Now primary charts are line charts with all years data. So the test update seems ok.

shifucun (commented):
Thanks for review. There are some outstanding items I have added TODO. Hopefully to get this in first and we can work on the issues.

static/js/place/place.ts
shifucun: seems the shorthand key is preferred, so change all fields to use shorthand key

server/routes/api/chart.py
```diff
@@ -74,9 +74,8 @@ def get_statsvars_need_all_dates(chart_list):
     """
```
beets, line 74: this function isn't required anymore since we need to keep all dates. could you add a note to remove this?

beets: also the get*date_for_stat_vars

beets (approved):
thanks! sorry for the late review

server/routes/api/chart.py
shifucun: Added a todo to remove them

server/tests/chart_test.py
```diff
@@ -1,13 +1,13 @@
 # Copyright 2020 Google LLC
 #
-# Licensed under the Apache License, Version 2.0 (the "License");
+# Licensed under the Apache License, Version 2.0 (the 'License');
```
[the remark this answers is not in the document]
shifucun: I did it ... when I tried to change the json " to '

shifucun:
@googlebot I signed it!

beets (approved):

PR #416, opened 2020-09-17
Remove Earth from parent places, add webdriver test

beets:
Didn't catch this before prod mixer release :(

image


beets:
The Earth place page is not too bad actually. This would still serve while we figure out how to do it nicely.
http://datacommons.org/place/?dcid=Earth

shifucun (approved):
nice test!

beets:
It's nice to be able to easily add them :)

PR #417, opened 2020-09-17
Update NPM packages

beets:
There were two outstanding security vulnerability that dependabot couldn't fix due to dependent packages, so ran an update since it's been a while. Tested our tools locally.


shifucun (approved):
nice!

PR #428, opened 2020-09-21
Use provenance domain from data in place page source footer

shifucun:
screenshot: https://staging.datacommons.org/dev/screenshot/8defc14

### DIFF
--- a/static/js/place/place_template.tsx
+++ b/static/js/place/place_template.tsx
@@ -681,6 +681,7 @@ interface ChartStateType {
   dataGroups?: DataGroup[];
   elemWidth: number;
   dateSelected?: string;
+  sources: string[];
 }
 
 class Chart extends Component<ChartPropType, ChartStateType> {
@@ -702,6 +703,7 @@ class Chart extends Component<ChartPropType, ChartStateType> {
 
     this.state = {
       elemWidth: 0,
+      sources: [],
     };
     // Consider debouncing / throttling this if it gets expensive at
     // small screen sizes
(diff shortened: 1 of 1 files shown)


static/js/place/place_template.tsx
```diff
@@ -749,7 +751,15 @@ class Chart extends Component<ChartPropType, ChartStateType> {
           <div id={this.props.id} className="svg-container"></div>
           <footer className="row explore-more-container">
             <div>
-              Data from <a href={config.url}>{config.source}</a>
+              <span>Data from </span>
+              {this.state.sources.map((source, index) => {
+                return (
+                  <span key={source}>
+                    <a href={"http://" + source}>{source}</a>
```
beets, line 758: is this safe to do for all the sources?

static/js/place/place_template.tsx
```diff
@@ -749,7 +751,15 @@ class Chart extends Component<ChartPropType, ChartStateType> {
           <div id={this.props.id} className="svg-container"></div>
           <footer className="row explore-more-container">
             <div>
-              Data from <a href={config.url}>{config.source}</a>
+              <span>Data from </span>
+              {this.state.sources.map((source, index) => {
```
beets, line 755: we might need to add a filter to ensure uniqueness

beets (commented):
This is a bit of a regression from before, particularly for EuroStat which now shows europa.eu, the home page for the EU. I know it'll be a bit of a backend change to do something better, but in the meantime, I wonder if we should add a map in python somewhere to remap some of these sources, e.g. eurostat, arcos. Can be a follow up too.

static/js/place/place_template.tsx
shifucun: checked this from bigquery Source.domain and the provenanceDomain should be domain only without http or https

static/js/place/place_template.tsx
shifucun: Checked the source in StatsData is a set so here should be good: https://github.com/datacommonsorg/website/blob/master/static/js/shared/data_fetcher.ts#L59

shifucun:
Yeah, agree we should have finer grade provenance url. Filed https://github.com/datacommonsorg/website/issues/429

static/js/place/place_template.tsx
beets: could you add a todo, that we expect to get the full url?

beets (dismissed):

shifucun:
Changed to https and added todo, PTAL

beets (approved):
thanks for investigation, and making the change

shifucun:
thx for review!

PR #436, opened 2020-09-21
Add household median income to pvtree menu

shifucun:
image


static/data/hierarchy_top.json
```diff
@@ -355,15 +355,24 @@
...
   "e": "Household",
-  "c": 680,
+  "c": 690,
   "cd": [
    {
     "sv": [
      "Count_Household"
     ],
-    "l": "Household",
+    "l": "Count",
```
beets, line 364: perhaps we can call this "population" instead?

static/data/hierarchy_top.json
```diff
@@ -355,15 +355,24 @@
...
     "t": "v",
-    "e": "Household",
+    "e": "Count",
+    "c": 1
+   },
+   {
+    "sv": [
+     "Median_Income_Household"
+    ],
+    "l": "Median Income",
```
beets, line 373: Median household income, so it won't be confused with median individual income. is this the string that will be used for the chips as well?

beets (approved):

static/data/hierarchy_top.json
shifucun: Done

static/data/hierarchy_top.json
shifucun: Done

beets (approved):

PR #437, opened 2020-09-22
speed up yapf; print list of files being checked

beets:
part of it was due to creating py3 environments in both ./ and ./server, so there were lots of python files being inspected. this updates it so it's always created in the root directory.

### DIFF
--- a/run_test.sh
+++ b/run_test.sh
@@ -40,18 +40,17 @@ function run_lint_fix {
   cd static
   npm list eslint || npm install eslint
   npm run lint
+  cd ..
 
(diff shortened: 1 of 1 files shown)


shifucun (approved):
very nice! no wonder it was slow before.

shifucun:
Also is it possible to skip some generated files like https://github.com/datacommonsorg/website/blob/master/tools/pv_tree_generator/stat_config_pb2.py ?

beets:
We might want to remove generated files from the repo (since that file is regenerated every time, we don't need to have it merged).

PR #438, opened 2020-09-22
Fix some missing data issue

shifucun:
Remove continent from parent places.
Remove null data from bar chart.
Add % as unit for unemployment rate.

### DIFF
--- a/server/chart_config.json
+++ b/server/chart_config.json
@@ -3,7 +3,8 @@
     "category": ["Economics", "Unemployment"],
     "title": "Unemployment Rate",
     "statsVars": ["UnemploymentRate_Person"],
-    "isOverview": true
+    "isOverview": true,
+    "unit": "%"
   },
   {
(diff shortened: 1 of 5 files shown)


beets (commented):
Did this trigger any d3 warnings in the console? Iirc, it was a naive fix to clear those warnings for stacked bar charts with null values.

shifucun:
> Did this trigger any d3 warnings in the console? Iirc, it was a naive fix to clear those warnings for stacked bar charts with null values.

I do not see console warnings. This is for a single bar chart case: 
image do you think the d3 warnings apply here?

beets:
The warnings were for stacked bar charts.

beets (dismissed):

beets:
There's a test that needs to be fixed tho

shifucun:
> There's a test that needs to be fixed tho

The tests are very comprehensive now.

beets (approved):

PR #439, opened 2020-09-22
Ignore null points for line charts

### LINKED ISSUE #418: [places] skip null points
E.g. health outcomes and behaviors chart in MTV (when shown as a timeline)

Don't treat null as 0

image

beets:
Also fix dots only drawn for one line chart

image

### DIFF
--- a/static/js/chart/draw.ts
+++ b/static/js/chart/draw.ts
@@ -16,7 +16,14 @@
 
 import * as d3 from "d3";
 
-import { DataGroup, DataPoint, PlotParams, Style, getColorFn } from "./base";
+import {
+  DataGroup,
+  DataPoint,
+  PlotParams,
+  Style,
+  getColorFn,
+  shouldFillInValues,
+} from "./base";
 
 const NUM_X_TICKS = 5;
 const NUM_Y_TICKS = 5;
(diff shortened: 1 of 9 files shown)


beets:
Here's another possible treatment

image

image

static/js/chart/draw.ts
```diff
@@ -485,22 +485,32 @@ function drawLineChart(
...
+
+    const line = d3
       .line()
+      .defined((d) => d[1] !== null)
       .x((d) => xScale(d[0]))
       .y((d) => yScale(d[1]));
 
-    if (shouldAddDots) {
-      line = line.curve(d3.curveMonotoneX);
+    if (hasNullValues) {
```
shifucun, line 507: This looks nice! Can you explain a bit more how this does the magic to use solid lines and dotted lines together? This is fancy d3 :)

beets:
Added warning:
image

Though I don't have a good way to know if there wasn't a fill line if the missing data is at the ends:
image

beets:
Fixed the second chart (will not show warning if there are only missing data on either end):

image

shifucun (approved):
very nice change!

PR #440, opened 2020-09-23
[Scatter2] Adding stub files

intrepiditee:
This PR
- adds a new route `tools/scatter2`
- adds stub files for the new route
- modifies `run_server.sh` and `run_npm.sh` to return to the cwd

Thanks for reviewing.


intrepiditee:
@beets Carolyn, could you review? Thanks

static/webpack.config.js
```diff
@@ -20,6 +20,10 @@ const FixStyleOnlyEntriesPlugin = require("webpack-fix-style-only-entries");
 
 const config = {
   entry: {
+    scatter2: [
```
beets, line 23: let's not forget to rename these after as well.

beets (dismissed):

static/webpack.config.js
intrepiditee: added a TODO

beets (approved):

PR #441, opened 2020-09-23
Alter chart drawing area to account for growing x-axis bounding box if required

### LINKED ISSUE #424: Chart axis line wrapping only has two lines and cut off in narrow screen size
image

In this case, the "County" is missing from "Santa Clara County".

beets:
Note: this means the x-axis isn't always aligned across charts, but only if labels are very long and it's required.

image

### DIFF
--- a/static/js/chart/draw.ts
+++ b/static/js/chart/draw.ts
@@ -296,7 +296,6 @@ function drawSingleBarChart(
     .paddingOuter(0.1);
 
   const bottomHeight = addXAxis(svg, chartHeight, x);
-  console.log(bottomHeight);
 
   const y = d3
     .scaleLinear()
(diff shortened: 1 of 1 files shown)


static/js/chart/draw.ts
```diff
@@ -247,44 +266,45 @@ function drawHistogram(
...
   const x = d3
     .scaleBand()
     .domain(textList)
-    .rangeRound([MARGIN.left, width - MARGIN.right])
+    .rangeRound([MARGIN.left, chartWidth - MARGIN.right])
     .paddingInner(0.1)
     .paddingOuter(0.1);
 
+  const bottomHeight = addXAxis(svg, chartHeight, x);
+  console.log(bottomHeight);
```
shifucun, line 299: remove logging?

shifucun (dismissed):

shifucun (approved):

PR #442, opened 2020-09-23
Add relatedChart field in chart config to indicate whether related chart can be scaled

shifucun:
relatedChart field contains the information for derived chart. It contains "scale" and "denominator" fields to indicate how scaling should be done

### DIFF
--- a/server/chart_config.json
+++ b/server/chart_config.json
@@ -577,13 +577,7 @@
     "isOverview": true,
     "relatedChart": {
       "scale": true,
-      "denominator": [
-        "Count_Person_25OrMoreYears",
-        "Count_Person_25OrMoreYears",
-        "Count_Person_25OrMoreYears",
-        "Count_Person_25OrMoreYears",
-        "Count_Person_25OrMoreYears"
-      ]
+      "denominator": "Count_Person_25OrMoreYears"
     }
   },
   {
@@ -596,7 +590,7 @@
     "isOverview": true,
     "relatedChart": {
       "scale": true,
-      "denominator": ["Count_Person_3OrMoreYears", "Count_Person_3OrMoreYears"]
+      "denominator": "Count_Person_3OrMoreYears"
     }
   },
   {
--- a/server/chart_config.json
+++ b/server/chart_config.json
@@ -3,8 +3,9 @@
     "category": ["Economics", "Unemployment"],
     "title": "Unemployment Rate",
     "statsVars": ["UnemploymentRate_Person"],
-    "isOverview": true,
-    "unit": "%"
+    "canScale": false,
+    "unit": "%",
+    "isOverview": true
   },
   {
     "category": ["Economics", "Unemployment"],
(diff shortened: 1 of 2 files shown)


beets:
i wonder if we should instead explicitly define a denominator (and if not defined, do not scale the chart). we often scale by population, when we should scale by a subpopulation.

shifucun:
Added a few denominator cases, otherwise default to total population

static/js/place/place_template.tsx
```diff
@@ -556,8 +557,18 @@ class ChartBlock extends Component<ChartBlockPropType, unknown> {
     );
   }
 
-  // TODO(shifucun): Add more config to indicate whether to use perCapita for
-  // place comparison.
+  private copyAndUpdateConfig(config: ConfigType) {
+    const conf = { ...config };
+    conf.chartType = chartTypeEnum.GROUP_BAR;
+    conf.axis = "PLACE";
+    if (conf.canScale == null || conf.canScale) {
```
beets, line 564: this should update if there conf.denominator is defined (or we can revert the extra canScale from the config - i prefer this option)

static/js/place/place_template.tsx
```diff
@@ -568,9 +579,7 @@ class ChartBlock extends Component<ChartBlockPropType, unknown> {
     conf.title = conf.title + " in " + this.props.placeName;
     result.push(conf);
 
-    conf = { ...config };
-    conf.chartType = chartTypeEnum.GROUP_BAR;
-    conf.axis = "PLACE";
+    conf = this.copyAndUpdateConfig(config);
```
beets, line 582: i missed this in my last PR - could you also use displayPlaceType for the else clause (line 588)

server/chart_config.json
```diff
@@ -35,40 +33,51 @@
...
       "UnemploymentRate_Person_Female"
-    ]
+    ],
+    "canScale": false
   },
   {
     "category": ["Economics", "Economic Activity"],
     "title": "Gross Domestic Product",
     "statsVars": ["Amount_EconomicActivity_GrossDomesticProduction_Nominal"],
+    "unit": "$",
```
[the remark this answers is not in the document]
shifucun: Use relatedChart field as discussed

static/js/place/place_template.tsx
shifucun: it should be okay there now

shifucun:
ptal with the new change

beets (commented):
i really like this new format - very easy to follow along

the data_fetcher code expects denominator, if specified, to be a string[] - will that be updated in a followup? a shortcut would be to expand that from a string to an array (of the right length) when copying the config.

shifucun:
> i really like this new format - very easy to follow along
> 
> 
> 
> the data_fetcher code expects denominator, if specified, to be a string[] - will that be updated in a followup? a shortcut would be to expand that from a string to an array (of the right length) when copying the config.

Expanding string to a list sounds good

beets (approved):

PR #443, opened 2020-09-23
Update chart and section titles

beets:
* Made each "chart" into it's own topic
* Updated styling for section headers
* Updated titles for each chart to be more explicit

Not updating the config yet in case we want to revert back to using that.

see http://staging.datacommons.org/dev/screenshot/c92a80a

### DIFF
--- a/server/chart_config.json
+++ b/server/chart_config.json
@@ -84,35 +84,10 @@
     "isOverview": true
   },
   {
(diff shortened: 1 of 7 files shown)


shifucun:
looking good

shifucun (dismissed):

shifucun (commented):
There is a test failure

shifucun (approved):

PR #447, opened 2020-09-25
[place page] a bunch of small fixes

beets:
* fixes layout for missing charts (all charts in a section should flow together)
* fix overflowing x-axis tick labels when there are long place names which are hyphenated
* drop line charts if there is only a single data point
* rename population density

### DIFF
--- a/static/js/place/place_template.tsx
+++ b/static/js/place/place_template.tsx
@@ -797,7 +797,12 @@ class Chart extends Component<ChartPropType, ChartStateType> {
     const dg = this.state.dataGroups;
     if (
       (dp && dp.length === 0) ||
-      (dg && (dg.length === 0 || (dg.length === 1 && dg[0].value.length === 0)))
+      (dg &&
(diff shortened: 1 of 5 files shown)


static/js/place/place_template.tsx
```diff
@@ -532,27 +532,25 @@ class ChartBlock extends Component<ChartBlockPropType, unknown> {
...
-                childPlaces={this.props.childPlaces}
-                similarPlaces={this.props.similarPlaces}
-                nearbyPlaces={this.props.nearbyPlaces}
-                chartData={this.props.chartData}
-              />
-            );
-          })}
-        </div>
-      </div>
+      <>
```
shifucun, line 535: good to know, so this also does not create top level parent element?

shifucun (approved):

static/js/place/place_template.tsx
beets: it's a container react fragment, but it doesn't translate to additional dom nodes. useful for returning lists :)

PR #451, opened 2020-09-25
add endpoint for getting geojson for charts

chejennifer:
added new endpoint api/chart/geojson/<path:dcid> which:
- takes the given dcid and gets the geojson for the list of places that are one level down from it

### DIFF
--- a/server/routes/api/chart.py
+++ b/server/routes/api/chart.py
@@ -24,12 +24,21 @@ import urllib
 import services.datacommons as dc_service
 import routes.api.place as place_api
 import os
+import routes.api.choropleth as choropleth_api
 
 from cache import cache
-from flask import Blueprint, current_app, Response, url_for
-
+from flask import Blueprint, current_app, Response, url_for, jsonify
+from routes.api.place import EQUIVALENT_PLACE_TYPES
 # Define blueprint
 bp = Blueprint("api_chart", __name__, url_prefix='/api/chart')
+
+# Place type to get choropleth for, keyed by place type
+# TODO: Come up with a catch all way to determine place type to get choropleth for because
+# this works for US places, but hierarchy of places in other countries may be different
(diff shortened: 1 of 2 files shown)


server/routes/api/chart.py
```diff
@@ -296,3 +305,76 @@ def config(dcid):
...
+            # Exclude geo if no or multiple renderings are present.
+            if len(json_text) != 1:
+                continue
+            geojson = json.loads(json_text[0])
+            geo_feature['geometry']['coordinates'] = (
+                choropleth_api.coerce_geojson_to_righthand_rule(
+                    geojson['coordinates'], geojson['type']))
+            features.append(geo_feature)
+    return Response(json.dumps({
+        "type": "FeatureCollection",
```
beets, line 373: curious what this is for: are we expecting other types?

server/routes/api/chart.py
```diff
@@ -296,3 +305,76 @@ def config(dcid):
...
+        if json_text and geo_id in names_by_geo:
+            geo_feature = {
+                "type": "Feature",
+                "geometry": {
+                    "type": "MultiPolygon",
+                },
+                "id": geo_id,
+                "properties": {
+                    "name": names_by_geo.get(geo_id, "Unnamed Area"),
+                    "hasSublevel": False,
```
beets, line 359: we won't need this either

beets (commented):
thanks! just a couple of questions

server/routes/api/chart.py
chejennifer: I don't think we are. I followed the format that was returned from api/choropleth/geo to keep it consistent

beets: we can remove these in a follow up (we should remove it from the other choropleth api as well since it's not used).

beets (approved):

PR #452, opened 2020-09-25
[place] chart title refinements

beets:
* also drop line charts if multiple series have only one point
* small amount of refactoring to reduce type declaration and to make it easier to manipulate parent/child places
* drop comparison chart for non-US from overview pages

probably easier to review it by looking through each commit separately.

### DIFF
--- a/static/js/place/place_template.tsx
+++ b/static/js/place/place_template.tsx
@@ -467,19 +467,18 @@ class ChildPlace extends Component<ChildPlacePropType, unknown> {
             <div className="child-place-type">
               {displayNameForPlaceType(placeType, true /* isPlural */)}
             </div>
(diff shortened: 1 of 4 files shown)


static/js/place/place_template.tsx
```diff
@@ -76,23 +76,40 @@ interface ConfigType {
   relatedChart: { scale: boolean; denominator: string };
 }
 
+type parentPlacesType = { dcid: string; name: string }[];
```
shifucun, line 79: nice type change!

shifucun (approved):

static/js/place/place_template.tsx
beets: thanks!

PR #453, opened 2020-09-27
Use cohort city and county for US city county similar places

shifucun:
image

### DIFF
--- a/server/routes/api/place.py
+++ b/server/routes/api/place.py
@@ -402,7 +402,7 @@ def api_similar_places(stats_var, dcid):
     var within the same place.
     """
     # Seed with current day of the year
-    random.seed(time.localtime().tm_yday)
+    random.seed(dcid + str(time.localtime().tm_yday))
     # Choose city from US city cohort.
     match = re.match(r'geoId/\d{7}', dcid)
     if match:
--- a/server/routes/api/place.py
+++ b/server/routes/api/place.py
@@ -16,6 +16,7 @@ import collections
 import json
 import random
 import re
+import time
 
 from flask import Blueprint, jsonify, request, Response, url_for
 
@@ -50,6 +51,7 @@ EQUIVALENT_PLACE_TYPES = {
 }
 
 CHILD_PLACE_LIMIT = 50
+SIMILAR_PLACE_LIMIT = 5
 
 # Minimal population count for a place to show up in nearby places list.
 MIN_POP = 5000
@@ -399,6 +401,8 @@ def api_similar_places(stats_var, dcid):
     selected from curated cohort. Otherwise, the similar places are by stats
     var within the same place.
     """
+    # Seed with current day of the year
+    random.seed(time.localtime().tm_yday)
     # Choose city from US city cohort.
     match = re.match(r'geoId/\d{7}', dcid)
     if match:
@@ -408,7 +412,7 @@ def api_similar_places(stats_var, dcid):
         for city in city_cohort:
             if city != dcid:
                 result.append(city)
-            if len(result) == 5:
+            if len(result) == SIMILAR_PLACE_LIMIT:
                 return Response(json.dumps(result),
                                 200,
                                 mimetype='application/json')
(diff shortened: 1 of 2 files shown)


server/routes/api/place.py
```diff
@@ -380,22 +384,60 @@ def get_related_place(dcid,
...
 
 
-@cache.memoize(timeout=3600 * 24)  # Cache for one day.
+@cache.memoize(timeout=3600 * 24)
+def get_place_cohort(cohort):
+    """Get all the place dcids for a cohort set."""
+    return dc.get_property_values([cohort], 'member')[cohort]
+
+
 @bp.route('/similar/<stats_var>/<path:dcid>')
```
beets, line 393: cache this perhaps?

server/routes/api/place.py
```diff
@@ -380,22 +384,60 @@ def get_related_place(dcid,
...
+
+    If the place is a USA city or county, the similar places are randomly
+    selected from curated cohort. Otherwise, the similar places are by stats
+    var within the same place.
     """
+    # Choose city from US city cohort.
+    match = re.match(r'geoId/\d{7}', dcid)
+    if match:
+        city_cohort = get_place_cohort(CITY_COHORT)
+        random.shuffle(city_cohort)
```
beets, line 406: let's seed this so the results are somewhat stable, maybe with the dcid?

server/routes/api/place.py
```diff
@@ -380,22 +384,60 @@ def get_related_place(dcid,
...
+        result = []
+        for city in city_cohort:
+            if city != dcid:
+                result.append(city)
+            if len(result) == 5:
```
beets, line 411: Use a const instead?

server/routes/api/place.py
```diff
@@ -380,22 +384,60 @@ def get_related_place(dcid,
...
+                return Response(json.dumps(result),
+                                200,
+                                mimetype='application/json')
+
+    match = re.match(r'geoId/\d{5}', dcid)
+    if match:
+        county_cohort = get_place_cohort(COUNTY_COHORT)
+        random.shuffle(county_cohort)
```
beets, line 419: same as above

server/routes/api/place.py
```diff
@@ -380,22 +384,60 @@ def get_related_place(dcid,
...
+        for county in county_cohort:
+            if county != dcid:
+                result.append(county)
+            if len(result) == 5:
+                return Response(json.dumps(result),
+                                200,
+                                mimetype='application/json')
+
     parents = parent_places(dcid)[dcid]
```
beets, line 428: should we return blank here, as a way to suppress comparison charts for non-US places?

server/routes/api/place.py
shifucun: caching would make the result deterministic so the result would not be random..

server/routes/api/place.py
shifucun: Added a seed with current day of the year. If with dcid then would not have the random list effect for a place..

server/routes/api/place.py

server/routes/api/place.py
shifucun: Added seed

server/routes/api/place.py
shifucun: It's a bit hard to know if a place is within United States. Shall we consolidate the disabling in client side with the function to check if a parent place is US.

server/routes/api/place.py
shifucun: Actually dcid + year of the day maybe better since each place would see different places?

server/routes/api/place.py
beets: we want it to be stable-random though, i.e. the list of similar places is varied as you look at different places. but randomness was a side-effect, rather than a feature.

server/routes/api/place.py
beets: ok that sgtm

server/routes/api/place.py
beets: ok lets try this.

beets (approved):

server/routes/api/place.py
beets: sure, sgtm

server/routes/api/place.py
shifucun: one thing forget to mention is the two server call functions used in the endpoint are cached so this endpoint won't have too much latency.

PR #454, opened 2020-09-28
Move each place component into separate files.

beets:
There should be no other diff (other than updating React import statements).
Will need to reuse some of these components for the chart embed page, plus it's getting difficult to navigate the file.


shifucun (commented):
This is nice, and we can conform to then lint error for "one class per file". Just a nit, can remove the "place_" prefix for all the files since they are all in the place folder?

beets:
Sure, I renamed the files. It might make files slightly harder to find when they have generic names (chart, ranking, map) though, but this will be consistent with the python files.

shifucun (approved):
nice!

PR #460, opened 2020-09-29
Endpoint for getting choropleth data

chejennifer:
New endpoint that will get all the data needed for all the choropleth charts for a place:
1. looks at the chart config to determine the set of stat vars we will need data for and the dictionary of stat var to set of denominators
2. gets the list of places at the display level we will be creating a choropleth for 
3. gets all the stats
4. to build the result, for each stat var to denominator combination, for each place, finds the value of dividing the value of the stat var on its latest common date by the value of the denominator on its latest common date (if there is a denominator, otherwise the value is just the value of the stat var)


server/routes/api/chart.py
```diff
@@ -378,3 +378,164 @@ def geojson(dcid):
...
+    """ Gets all the stat vars and accompanying denominators that will have a choropleth chart drawn for them
+    ie. charts with the hasChoropleth property set as true
+
+    Returns: 
+        a tuple consisting of:
+            set of all stat vars including both the stat vars of the charts and stat vars of denominators
+            dictionary of stat var: list of denominators
+    """
+    chart_config = current_app.config['CHART_CONFIG']
+    if os.environ.get('FLASK_ENV') == 'development':
```
shifucun, line 393: suggest to remove line 393-395. We have seen a bug from chart config but was not able to catch it in dev :)

Might just need to restart the server when chart config is changed, which is not that bad

server/routes/api/chart.py
```diff
@@ -378,3 +378,164 @@ def geojson(dcid):
...
+        with bp.open_resource('../../chart_config.json') as f:
+            raw_config = json.load(f)
+    all_sv = set()
+    sv_denom_mapping = {}
+    for config in chart_config:
+        if 'hasChoropleth' in config and config['hasChoropleth']:
```
shifucun, line 399: config.get('hasChoropleth', false)

server/routes/api/chart.py
```diff
@@ -378,3 +378,164 @@ def geojson(dcid):
...
+            # we should only be making choropleths for configs with a single stat var
+            sv = config['statsVars'][0]
```
shifucun, line 401: worth adding a test for chart config to ensure "hasChoropleth" is only added to chart with single statvar, can add a TODO for it.

server/routes/api/chart.py
```diff
@@ -378,3 +378,164 @@ def geojson(dcid):
...
+            all_sv.add(sv)
+            denom = ''
+            if 'denominator' in config:
+                denom = config['denominator'][0]
+                all_sv.add(denom)
+            elif 'perCapita' in config and config['perCapita']:
```
shifucun, line 407: we don't have perCapita in the config now. Also for sv like Median Age, can not be scaled. You can check config.releatedChart.scale, if it false, meaning the original chart can not be scaled.

server/routes/api/chart.py
```diff
@@ -378,3 +378,164 @@ def geojson(dcid):
...
+                # The denominator used for perCapita is total population
+                all_sv.add('Count_Person')
+                denom = 'Count_Person'
+            if sv not in sv_denom_mapping:
+                sv_denom_mapping[sv] = set()
+            sv_denom_mapping[sv].add(denom)
+    return all_sv, sv_denom_mapping
+
+
+def get_data_for_statvar(sv, geos, all_sv_data):
```
shifucun, line 417: would be good to also return the provenance url which would be used in the chart footnote. can add a TODO for that.

server/routes/api/chart.py
```diff
@@ -378,3 +378,164 @@ def geojson(dcid):
...
+        if len(sorted_dates) > 0:
+            sv_date = sorted_dates[-1]
+    if not sv_date and all_dates:
+        sorted_dates = sorted(all_dates)
+        if len(sorted_dates) > 0:
+            sv_date = sorted_dates[-1]
+    return sv_date
+
+
+@bp.route('/choroplethdata/<path:dcid>')
```
shifucun, line 475: oh, should these functions in choropleth.py and the actual path for this endpoint is /choropleth/data/<path:dcid>?

chejennifer: I put it in chart.py because the endpoint is getting choropleth data for choropleth charts (based off of chart config), and it seems like choropleth.py contains endpoints for the choropleth tool. Not sure if we want to keep those two things separated

shifucun: Ah, that's right, i was thinking the choropleth tool. Then this is right.

server/routes/api/chart.py
```diff
@@ -390,24 +390,22 @@ def get_choropleth_sv():
...
-        if 'hasChoropleth' in config and config['hasChoropleth']:
+        if config.get('hasChoropleth', False):
             # we should only be making choropleths for configs with a single stat var
+            # TODO(chejennifer) add test for chart config to ensure hasChoropleth is only added to charts with single statvar
             sv = config['statsVars'][0]
             all_sv.add(sv)
             denom = ''
-            if 'denominator' in config:
+            if config.get('relatedChart',
+                          {}) and config['relatedChart']['scale']:
```
shifucun, line 403: can also do config['relatedChart'].get('scale', False) incase the 'scale' field is not set

shifucun: and also if config.relatedChart.scale == True, but config.related.denominator is not set, it defaults to "Count_Person"

shifucun:
looking good. There are some lint errors. Please fix.

shifucun (dismissed):

shifucun (approved):

PR #461, opened 2020-09-30
Add chart config for life expectancy.

tjann:
Local screenshot: 
image

I noticed current staging/prod does not have years for median age so I did the same: 
image

Question: I had some changes to static/package.*json from just running `./run_npm.sh`. Is the correct thing to do: do not commit these changes unless you know what you are doing?

### DIFF
--- a/server/chart_config.json
+++ b/server/chart_config.json
@@ -101,7 +101,6 @@
       "denominator": "Count_Person"
     }
   },
-
   {
     "category": ["Economics", "Individual Income"],
     "title": "Median Individual Income",
@@ -159,6 +158,12 @@
       "scale": true,
       "denominator": "Count_Household"
     }
+  },
+  {
+    "category": ["Health", "Life Expectancy"],
+    "title": "Life Expectancy",
+    "statsVars": ["LifeExpectancy_Person"],
+    "isOverview": true
   },
   {
     "category": ["Health", "COVID-19"],
--- a/server/chart_config.json
+++ b/server/chart_config.json
@@ -625,15 +625,6 @@
   },
   {
     "category": ["Demographics", "Population"],
-    "title": "Life Expectancy",
-    "statsVars": ["LifeExpectancy_Person"],
-    "isOverview": true,
-    "relatedChart": {
-      "scale": false
-    }
-  },
-  {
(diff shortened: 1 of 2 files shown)


tjann:
Thanks for the idea @beets ! Here are some screenshots to show the update:

Chart gone from demo
image

Chart now in health
image

How the charts look
image


Out of scope of this PR: exact chart placements (unless someone wants to tell me where and how to update). I also found and filed #474

server/chart_config.json
```diff
@@ -160,6 +159,12 @@
       "denominator": "Count_Household"
     }
   },
+  {
+    "category": ["Health", "Life Expectancy"],
+    "title": "Life Expectancy",
+    "statsVars": ["LifeExpectancy_Person"],
+    "isOverview": true
```
beets, line 166: this is somewhat out of date now. could you please copy over relatedChart as well?

server/webdriver_tests/place_explorer_test.py
```diff
@@ -91,7 +91,7 @@ def test_demographics_link(self):
         chart_title = age_across_places_chart.find_element_by_tag_name(
             "h4").text
         self.assertEqual(
-            "Population by Gender across states near California(2018)",
+            "Median Age by Gender across states near California(2018)",
```
beets, line 94: this will have to be updated post-merge from upstream/master as well

beets (changes requested):
thanks - lg in general though there's been a ton of adjacent changes recently

server/chart_config.json
beets: please copy over the relatedChart object from the existing config. and post-merge from upstream, we no longer require the second item in the category array.

beets (commented):
don't forget to run ./run_test.py -f to fix the lint errors

server/chart_config.json
tjann: done

server/webdriver_tests/place_explorer_test.py
tjann: done

tjann (commented):
Looks ready to me

beets (approved):

PR #463, opened 2020-09-30
[PV tree] possible denominators

intrepiditee:
This PR modifes the PV tree generation script to add a "d" field to each value node that has a list of potential per capita denominators for the StatVars represented by the node.

Discussions
- How lenient do we want to be when deciding if a denominator should be included in the list. Currently a denominator is included as long as its pv set is a subset of the dpv. This means if the dpv is `age: "15YearsOnwards", incomeStatus: "WithIncome"`, the denominators included are `Count_Person`, `Count_Person_15OrMoreYears`, `Count_Person_15OrMoreYears_WithIncome`.
- For criminal activities, there isn't a StatVar for the total cases like `Count_CriminalActivities`. Should we add `Count_Person` and `Count_Household` to every value node by default?

Thanks for reviewing.


tools/pv_tree_generator/build_tree.py
```diff
@@ -104,9 +107,59 @@ def json_blob(self):
...
+                      stats_vars_all: Dict[str, List[StatsVar]],
+                      stats_vars_to_exclude: List[str]) -> List[str]:
+    """Finds possible StatVars that can be used as per capita denominators for
+    a StatVar.
+
+    The dpvs of "pop_obs" and "cpvs" are combined to form the final pv mapping.
+    A denominator is included if its pvs are a subset of the final pv mapping.
+    E.g., if dpvs is {'age': '15YearsOnwards'} and cpvs is {'gender': 'female'},
+    the denominators are 'Count_Person', 'Count_Person_Female',
+    'Count_Person_15OrMoreYears', and 'Count_Person_Female_15OrMoreYears'.
```
shifucun, line 137: Can we order the sv by the importance, ie, 
- the sv corresponding to the dpv population is at the beginning.
- the sv corresponding to 0 pv population: Count_Person, Count_Household comes next,
- then all the other population that corresponding to the cpvs.

shifucun (commented):
The PR is very nice done. As we discussed offline, for now we will use the "primary" denominator, which is dpv population or total population. But the entire change should still be kept as we will very likely to use all the other denominators later.

tools/pv_tree_generator/build_tree.py
intrepiditee: done

shifucun (approved):
thanks for the change!

intrepiditee:
@shifucun Thanks for reviewing. Could you merge? I don't think I have the permission

Issue #200, opened 2020-07-29
[BUG] Investigate single datapoint for London population


beets:
### Describe the bug
http://staging.datacommons.org/place?dcid=wikidataId%2FQ84

![image](https://user-images.githubusercontent.com/6052978/88865281-53e82100-d1bc-11ea-81ec-14f1e8f75c72.png)

pradh:
There are stats from UNData and Wikidata sources. UNData is a single 2011 data point, while Wikidata is a time-series (http://staging.datacommons.org/browser/dc/p/qlyvv2jyqszzb).

I think the cache should build in both. I know the mixer does some ranking among sources [1]. For ones that aren't listed there, is the choice arbitrary? Should it be based on some heuristics on the data (has time-series or has the latest data, etc.)


[1] - https://github.com/datacommonsorg/mixer/blob/104d4db5b44f862beda42ee268d63a0678f7caae/server/stats.go#L33-L45

beets:
@shifucun 

Should we add more to this list then? The preference for a time series is only for UI purposes, but if we weight it towards latest data, that will resolve to Wikidata which works better for both UI and analysis.

pradh:
I will note that the date on Wikidata population stats will be all over the place across places. So won't be surprised if, for another place UNData was more recent.

shifucun:
Looks like we need a bit complicated ranking on both data freshness and time span. I propose to:

1) First compare data time span. If one dataset has twice or more data than another, then pick it.
2) If two datasets have similar time span, pick the one that has latest data.

Any other ideas?

PR #464, opened 2020-10-01
Add a new api module to serve bulk place page data

shifucun:
This makes fundamental changes to the place page rendering:

- Server side prepares all the data that are ready to use for client side.
- No need to async fetch data on client side any more.
- The menu and chart rendering are based on availability of the data not on stat var in a places. This effectively makes more charts available on each page. For example, state data can now be rendered on USA page.

### DIFF
--- a/server/routes/api/landingpage.py
+++ b/server/routes/api/landingpage.py
@@ -26,11 +26,12 @@ from cache import cache
 import services.datacommons as dc_service
 from services.datacommons import fetch_data
 import routes.api.stats as stats_api
+import routes.api.place as place_api
 
 import logging
 
 # Define blueprint
-bp = Blueprint("api.place.new", __name__, url_prefix='/api/place/new')
+bp = Blueprint("api.landingpage", __name__, url_prefix='/api/landingpage')
 
 
 def get_landing_page_data(dcid):
@@ -91,6 +92,7 @@ def data(dcid):
     config_data = copy.deepcopy(config)
 
     for category in config_data:
+        filtered_charts = []
         for chart in category['charts']:
             # Populate the overview page data.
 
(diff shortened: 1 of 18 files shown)


shifucun:
PR is ready for review. Hoping to get this in first and inspect bugs on staging.

server/routes/api/landingpage.py
```diff
@@ -0,0 +1,292 @@
...
+
+def get_bar_data(cc, data, places):
+    """Get the bar data across a few places.
+
+    This will scale the value if required and pick the latest date that has the
+    most <place, stat_var> entries.
+    """
+    if not places:
+        return {}
+    date_to_data = collections.defaultdict(
```
beets, line 173: please describe the format of date_to_data

server/routes/api/landingpage.py
```diff
@@ -0,0 +1,292 @@
...
+                continue
+            if 'denominator' in cc:
+                denominator_stat_var = cc['denominator'][i]
+            if denominator_stat_var is not None:
+                denominator_raw = data[dcid].get(denominator_stat_var, {})
+                if not denominator_raw:
+                    continue
+                series = scale_series(series_raw['data'],
+                                      denominator_raw['data'])
+                sources.add(series_raw['provenanceDomain'])
```
beets, line 197: could you add a note that we should only add this to sources if we end up picking the place for the final result (this could happen with international place comparisons)

server/routes/api/landingpage.py
```diff
@@ -0,0 +1,292 @@
...
+            chosen_date = date
+    result = {'date': chosen_date, 'data': [], 'sources': list(sources)}
+    for place in places:
+        points = {}
+        for stat_var, value in date_to_data[chosen_date][place]:
+            points[stat_var] = value
+        if points:
+            result['data'].append({
+                'dcid': place,
+                # 'name': place['name'],
```
beets, line 223: delete this line

server/routes/api/landingpage.py
```diff
@@ -0,0 +1,292 @@
...
+                'data': points
+            })
+    # Should have data other than the primary place. Return empty struct to
+    # so client won't draw chart.
+    if len(result['data']) == 1:
```
beets, line 228: what if we don't have data even for the current place?

server/routes/api/landingpage.py
```diff
@@ -0,0 +1,292 @@
...
+            return {}
+
+    series = {}
+    sources = set()
+    for i, stat_var in enumerate(cc['statsVars']):
+        numerator_raw = data[place].get(stat_var, {})
+        if not numerator_raw:
+            continue
+
+        if 'denominator' in cc:
```
beets, line 251: nit: check / assert that denominator has the right length?

beets: i suppose in these cases, the array lookup failure is equivalent to an assert

static/js/place/chart.tsx
```diff
@@ -106,53 +96,31 @@ class Chart extends React.Component<ChartPropType, ChartStateType> {
...
-    if (
-      this.props.config.placeRelation === placeRelationEnum.SIMILAR &&
-      this.props.similarPlaces.length === 1
-    ) {
-      return null;
-    }
+    const sources = this.props.trend
+      ? this.props.trend.sources
+      : this.props.snapshot.sources;
+    const explorUrl = this.props.trend
```
beets, line 116: nit: exploreUrl

static/js/place/chart.tsx
```diff
@@ -194,33 +162,6 @@ class Chart extends React.Component<ChartPropType, ChartStateType> {
...
-    }
-    const dp = this.state.dataPoints;
-    const dg = this.state.dataGroups;
-    if (
-      (dp && dp.length === 0) ||
-      (dg &&
-        (dg.length === 0 || (dg.length === 1 && dg[0].value.length === 0))) ||
-      (this.props.config.chartType == "LINE" &&
-        dg &&
-        dg.reduce((accum, group) => {
```
beets, line 208: this was a check to prevent showing line charts with single points #200 (and another place that I didn't record had multiple statvars with a single point)

static/js/place/chart.tsx
```diff
@@ -308,150 +249,104 @@ class Chart extends React.Component<ChartPropType, ChartStateType> {
...
+      return a.label > b.label ? -1 : 1;
+    });
+    return result;
+  }
+
+  private processData(): void {
+    const dataGroups: DataGroup[] = [];
+    const dataPoints: DataPoint[] = [];
+    const allDates = new Set<string>();
+    const scaling = this.props.scaling ? this.props.scaling : 1;
```
beets, line 297: todo: move this to python too

static/js/place/chart_block.tsx
```diff
@@ -83,124 +58,163 @@ class ChartBlock extends React.Component<ChartBlockPropType, unknown> {
...
-      conf.placeRelation = placeRelationEnum.CONTAINED;
-      conf.title = `${conf.title} across places that contain ${this.props.placeName}`;
-      result.push(conf);
+      if (this.props.isOverview) {
+        // Show child place(state) chart for USA page, otherwise show nearby
+        // places.
+        const id = randDomId();
+        if (this.props.dcid === "country/USA") {
+          if (!_.isEmpty(this.props.data.child)) {
+            chartElements.push(
```
beets, line 106: todo: simplify this code to reduce repetition

beets (changes requested):
nice to see all the data coming in one call. mostly todos

beets (approved):

shifucun:
Thanks for the review! Resolved and added TODO for most of them.

PR #472, opened 2020-10-02
[place] Embed chart enhancements

beets:
* move svg chart styling inline from css (so that all styling information is self-contained in the svg)
* added xml namespace attributes to svg
* include title and provenance information into a hidden svg image, which is then used to get the final svg xml for download and embed
* svg chart is now embedded as an image so that it can be copied
* svg is now downloadable as a standalone image


shifucun (approved):
very nice!

PR #475, opened 2020-10-03
Some chart rendering fixes

beets:
* Fix overlapping stack bar charts when there is incomplete stat var data
* Highlight the SVG rect of the current place (it's quite subtle so we might want to tweak it)
* Make place labels clickable again (also applied to stack bar charts)
* Fix label rendering crispness and timelines tool (regression from #472)

image

### DIFF
--- a/server/templates/place.html
+++ b/server/templates/place.html
@@ -66,6 +66,20 @@
     </div>
   </div>
   <div id="modal"></div>
+  {#- Add an SVG filter to lighten bars for comparison places -#}
+  <style>
+    g rect.g-bar[data-dcid="{{ place_dcid }}"] {
+      filter: url(#lighten);
+    }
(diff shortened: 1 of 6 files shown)


server/templates/place.html
```diff
@@ -66,11 +66,25 @@ <h2 class="col-12 pt-2" id="overview">Overview</h2>
     </div>
   </div>
   <div id="modal"></div>
+  {#- Add an SVG filter to lighten bars for comparison places -#}
+  <style>
+    g rect.g-bar[data-dcid="{{ place_dcid }}"] {
+      filter: url(#lighten);
```
shifucun, line 72: good to know this feature.

shifucun (approved):

shifucun (approved):

PR #483, opened 2020-10-05
Handle sticky topic param when there is no data for that topic

shifucun:


static/js/place/subtitle.tsx
```diff
@@ -0,0 +1,49 @@
...
+ */
+
+import React from "react";
+
+interface SubtitlePropsType {
+  category: string;
+  dcid: string;
+}
+
+class Subtitle extends React.Component<SubtitlePropsType, unknown> {
```
beets, line 24: please add a component description (and can we call this PageSubtitle to be more specific?)

beets (changes requested):
nice change!

static/js/place/subtitle.tsx
shifucun: Done

beets (approved):
thanks for making the change!

PR #485, opened 2020-10-05
hook up choropleth client

chejennifer:
If chart_config.json has configs with isChoropleth set to true, choropleth charts will be shown in place explorer on the topic page
Changes:
- In place.ts, add extra promises to get geojson data and statvar/denom data
- added props in MainPane, ChartBlock, and Chart to pass the extra data
- In Chart, updated processData and drawChart with cases for choropleth chart


static/js/place/chart.tsx
```diff
@@ -70,6 +77,14 @@ interface ChartPropType {
    * Scale number
    */
   scaling?: number;
+  /**
+   * Geojson data for places one level down of current dcid.
+   */
+  geoJsonData?: unknown;
```
shifucun, line 83: would be nice to use the actual type for geoJson

static/js/place/chart.tsx
```diff
@@ -114,12 +130,26 @@ class Chart extends React.Component<ChartPropType, ChartStateType> {
...
       ? "(" + this.props.snapshot.date + ")"
       : "";
-    const sources = this.props.trend
-      ? this.props.trend.sources
-      : this.props.snapshot.sources;
-    const exploreUrl = this.props.trend
-      ? this.props.trend.exploreUrl
-      : this.props.snapshot.exploreUrl;
+    let sources = [];
+    if (this.props.chartType !== chartTypeEnum.CHOROPLETH) {
```
shifucun, line 134: does choropleth data have sources now?

shifucun (commented):
looks mostly good, just a few nits

static/js/place/chart.tsx
chejennifer: not yet, that's something I still need to add. Also need to get the exploreURL when getting choropleth data

beets (changes requested):
nice change, just add a few todos

static/js/place/chart.tsx
beets: we can also add a type for it first (which is set to unknown for now) with a todo to add the complete type later

server/routes/api/chart.py
```diff
@@ -371,6 +371,8 @@ def geojson(dcid):
                 choropleth_api.coerce_geojson_to_righthand_rule(
                     geojson['coordinates'], geojson['type']))
             features.append(geo_feature)
+    print("here\n")
```
shifucun, line 374: remove logging?

shifucun (approved):

shifucun (approved):

beets (approved):
thanks! let's get this in

PR #486, opened 2020-10-05
Actual fix for single dot line chart issue

shifucun:


tjann (dismissed):

server/routes/api/landing_page.py
```diff
@@ -194,8 +194,11 @@ def get_trend(cc, data, place):
         else:
             series[stat_var] = numerator_raw['data']
             sources.add(numerator_raw['provenanceDomain'])
-    if not series or len(series) == 1:
+    if not series:
         return {}
+    for stat_var in list(series.keys()):
+        if len(series[stat_var]) <= 1:
+            del series[stat_var]
```
beets, line 201: should we also return {} if all series are deleted?

shifucun: good catch, move it above so the return {} can come to play

tjann (approved):

PR #489, opened 2020-10-05
[Timeline] Use dpv population as per capita denominator

intrepiditee:
Features
- The first denominator in the list is passed to to the chart
- When grouped together in one chart, different StatVars have their own denominators
- The denominator is stored in the URL with the node path
- An asterisk is added next to the per capita button that redirects to the FAQ page

Screen Shot 2020-10-06 at 4 16 06 PM

Issues
- Denominators are not filtered
- We can consider also modifying `statsvar_path.json` to include the denominators with the node path so that if the denominator is missing from the URl, we can still load it


static/js/tools/timeline_page.tsx
```diff
@@ -246,6 +252,12 @@ class Page extends Component<Record<string, unknown>, PageStateType> {
                     chartOptions={this.state.chartOptions}
                     setPC={this.setChartPerCapita.bind(this)}
                     initialPC={this.params.allPerCapita}
+                    denominators={Object.entries(
+                      this.state.statsVarNodes
+                    ).reduce((denominators, [dcid, node]) => {
+                      denominators[dcid] = node.denominators;
+                      return denominators;
+                    }, {})}
```
beets, line 260: this is neat!

static/js/tools/timeline_util.ts
```diff
@@ -247,19 +266,31 @@ class TimelineParams {
...
           } else {
-            this.addStatsVar(statsVarInfo[0], []);
+            this.addStatsVar(dcid, [], []);
           }
         } else {
-          this.addStatsVar(statsVarInfo[0], statsVarInfo.splice(1));
+          const pathsAndDenominators = statsVarInfo.splice(1);
+          this.addStatsVar(
+            dcid,
+            pathsAndDenominators.filter(
```
beets, line 287: can you describe these magic numbers? in particular, why 9 (it's hard to follow the format of the json, so would be good to add some documentation here. is there a better var name for elem? and should those be numbers instead of strings?)

beets (commented):
thanks! this looks really complete, though have a small question below

server/webdriver_tests/screenshot/screenshot_test.py
```diff
@@ -115,6 +115,9 @@ def test_pages_and_sreenshot(self):
             charts = self.driver.find_elements_by_class_name(
                 test_info['test_class'])
             # Assert there are charts.
+            if not len(charts):
```
shifucun, line 118: remove log?

static/data/hierarchy_statsvar.json
```diff
@@ -127217,7 +127217,8 @@
      }
     ]
    }
-  ]
```
shifucun, line 127220: there is no pv tree code change, are these from the run of the latest pv tree tool?

static/js/tools/statsvar_menu.tsx
```diff
@@ -37,10 +38,15 @@ interface NodePropType {
...
+  d?: string[]; // possible denominators
   selectedNodes: StatsVarNode; // path to all the selected statsVars
   nodePath: string[]; // path to current node
   addStatsVarTitle: (statsVarId: string, statsVarName: string) => void; // pass the title of selected statsVar to parent
-  addStatsVar: (statsVar: string, nodePath: string[]) => void; // function for adding statsVar
+  addStatsVar: (
+    statsVar: string,
+    nodePath: string[],
+    denominators: string[]
+  ) => void; // function for adding statsVar
```
shifucun, line 49: can you check if this affects other client like choropleth tool which also uses the menu

static/js/tools/timeline_chart.tsx
```diff
@@ -98,11 +100,14 @@ class Chart extends Component<ChartPropsType, unknown> {
               this.props.setPC(this.props.groupId, !this.props.perCapita);
             }}
           ></button>
+          <a href="/faq">
+            <span>&#42;</span>
```
shifucun, line 104: can use {*} in React

shifucun (commented):
looks pretty good, some minor comments

static/js/tools/timeline_page.tsx
intrepiditee: thanks

static/data/hierarchy_statsvar.json
intrepiditee: I dropped the pv tree files from this PR

static/js/tools/timeline_chart.tsx
intrepiditee: replaced

static/js/tools/statsvar_menu.tsx
intrepiditee: You are right! It did break the choropleth tool. fixed now

server/webdriver_tests/screenshot/screenshot_test.py
intrepiditee: removed

static/js/tools/timeline_util.ts
intrepiditee: is this better?

static/js/tools/statsvar_menu.tsx
intrepiditee: Not because of the extra param though. The problem was that the values of StatsVarNodes become objects

intrepiditee:
@shifucun @beets All the comments are addressed

static/js/tools/timeline_util.ts
beets: yes, better. thanks!

beets (approved):

PR #492, opened 2020-10-05
get explore url for choropleth charts

chejennifer:
add explore url for choropleth charts which will redirect to timeline tool for the stat var for the current place

Changes:
- updated api/chart/choroplethdata/<dcid> to return the explore url for each stat var
- updated chart.tsx to get explore url for choropleth charts

### DIFF
--- a/server/routes/api/chart.py
+++ b/server/routes/api/chart.py
@@ -25,6 +25,7 @@ import services.datacommons as dc_service
 import routes.api.place as place_api
 import os
 import routes.api.choropleth as choropleth_api
+import routes.api.landing_page as landing_page_api
 
 from cache import cache
 from flask import Blueprint, current_app, Response, url_for, jsonify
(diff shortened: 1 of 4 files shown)


server/routes/api/chart.py
```diff
@@ -544,7 +544,9 @@ def choropleth_data(dcid):
             sv_denom_result = {
                 'date': sv_date,
                 'data': geo_data,
-                'numDataPoints': num_data_points
+                'numDataPoints': num_data_points,
+                # TODO (chejennifer): exploreUrl should link to choropleth tool once the tool is ready
+                'exploreUrl': build_url(dcid, [sv])
```
beets, line 549: don't forget to set is_scaled=true if there is a denominator

beets (changes requested):

server/routes/api/chart.py
chejennifer: it doesn't look like there's anywhere to set that?

beets: oh sorry! i thought i had merged #488

server/routes/api/chart.py
```diff
@@ -541,12 +542,17 @@ def choropleth_data(dcid):
                 else:
                     geo_data[geo] = sv_data[geo][sv_date]
                     num_data_points += 1
+            exploreUrl = None
```
beets, line 545: could you just call:

exploreUrl = landing_page_api.build_url([dcid], [sv], bool(denom_data))

beets (changes requested):

beets (approved):

PR #496, opened 2020-10-06
Use chosen child place type returned by mixer

### LINKED ISSUE #372: Clicking on a graph item
Consider something like the drug prescribed chart on http://datacommons.org/place?dcid=geoId/10

It would be good to click on Oxycodone and get the ranking of all places of this type (in this case, State) by that variable.

beets:
Follow up of #494


shifucun (approved):
Thx for the change!

PR #502, opened 2020-10-07
Place redirect from get to path param

beets:
Add redirects from /place?dcid=foo&topic to /place/foo?topic


server/webdriver_tests/screenshot/screenshot_test.py
```diff
@@ -28,13 +28,13 @@
         'height': 5500
     },
     {
-        'url': '/place?dcid=country/USA&topic=Demographics',
+        'url': '/place?dcid=country/USA?topic=Demographics',
```
shifucun, line 31: change here to /place/country/USA too?

shifucun (dismissed):

server/webdriver_tests/screenshot/screenshot_test.py
beets: that was done on purpose, to test that the redirect works as expected

beets:
finally fixed merge + tests, ptal :)

shifucun (approved):

PR #503, opened 2020-10-07
Use landing page method for getting data for choropleth

chejennifer:
Update the endpoint for getting choropleth data to use get_snapshot_across_places method from landing_page.py, and allow sources to be shown for choropleth charts
Changes:
- update helper functions for getting choropleth data to be able to use get_snapshot_across_places
- add sources property to ChoroplethDataGroup on client
- update geoJson endpoint to get the best resolution geojson out of the low resolution versions


beets (approved):
nice change

PR #504, opened 2020-10-07
Use predefined age aggregation patterns and scoring based on multiple places

shifucun:


server/tests/range_test.py
```diff
@@ -18,53 +18,227 @@
...
+                'Count_Person_65To74Years': ['Count_Person_65To74Years'],
+                'Count_Person_75OrMoreYears': ['Count_Person_75OrMoreYears']
+            },
+            'geoId/12345': {
+                'Count_Person_5To17Years': ['Count_Person_5To17Years'],
+                'Count_Person_18To24Years': ['Count_Person_18To24Years'],
+                'Count_Person_25To34Years': ['Count_Person_25To34Years'],
+                'Count_Person_35To44Years': ['Count_Person_35To44Years'],
+                'Count_Person_45To54Years': ['Count_Person_45To54Years'],
+                'Count_Person_65To74Years': ['Count_Person_65To74Years']
```
beets, line 130: this should also have 55-64

server/tests/range_test.py
```diff
@@ -18,53 +18,227 @@
...
+            }
+        }
+        assert lib_range.aggregate_age_stat_var(input) == expected
+
+    def test_eu_place_age(self):
+        input = {
+            "country/USA": [
```
beets, line 137: maybe update the dcid for this example so it's clearer?

server/tests/range_test.py
```diff
@@ -18,53 +18,227 @@
...
+                'Count_Person_40To49Years': [
+                    'Count_Person_45To49Years', 'Count_Person_40To44Years'
+                ],
+                'Count_Person_50To59Years': [
+                    'Count_Person_55To59Years', 'Count_Person_50To54Years'
+                ],
+                'Count_Person_60To69Years': [
+                    'Count_Person_65To69Years', 'Count_Person_60To64Years'
+                ],
+                'Count_Person_70To79Years': [
```
beets, line 215: ideally, we'd do 70+ in this case but it's ok for now.

beets (dismissed):

server/tests/range_test.py
shifucun: Updated to use 70+

server/tests/range_test.py
shifucun: done

server/tests/range_test.py
shifucun: good catch!

beets (approved):

PR #510, opened 2020-10-07
Remove duplicate chart config and add config test

shifucun:


beets (approved):
nice, thanks for adding the test!

PR #512, opened 2020-10-08
Use related place api that now only supports same place type

shifucun:
Note this has to be released together with: https://github.com/datacommonsorg/mixer/pull/328


pradh (approved):
Thanks Bo!

PR #513, opened 2020-10-08
Related chart defaults to use percentage

shifucun:


beets:
can we fix this in the config instead? we have some charts where this won't make sense, eg
government debt per capita
retail drug distribution per capita

i think there are only a handful that doesn't have this set where it applies, all in economics:
state unemployment insurance rate
percentage of people employed
labor force participation rate

shifucun:
Fixed the chart config now.

server/chart_config.json
```diff
@@ -9,15 +9,17 @@
   },
   {
     "category": "Economics",
-    "title": "State Unemployment Insurance",
+    "title": "State Unemployment Insurance Claim",
```
beets, line 12: nit (here and related title): Claims

server/chart_config.json
```diff
@@ -894,7 +907,7 @@
     ],
     "isOverview": true,
     "relatedChart": {
-      "title": "Marital Status Per Capita",
+      "title": "Marital Status Rate",
```
beets, line 910: citizenship and other pop breakdowns are currently listed as "per capita" so this was meant to be consistent with those.

beets (commented):
thanks for the udpate!

server/chart_config.json
shifucun: Find it a bit confusing if we should 20% together with "per capita", maybe choose one of the two, I feel rate is closer to what we use daily?

beets: i figure for demographics related stats, per capita is more accurate..

beets: let's keep rates:
http://www.oecd.org/social/family/SF_3_1_Marriage_and_divorce_rates.pdf

beets (approved):
could you update claim to claims? then this is good to go

server/chart_config.json
shifucun: Done. Thanks!

PR #514, opened 2020-10-08
fix tooltip positioning

chejennifer:
This change should make sure that the tooltip is always above and to the right of the mouse pointer, so it won't get in the way of hovering over the current area

### DIFF
--- a/static/js/chart/draw_choropleth.ts
+++ b/static/js/chart/draw_choropleth.ts
@@ -138,8 +138,9 @@ const onMouseMove = (
   const tooltipHeight = document
     .getElementById(containerId)
     .querySelector(`[id=${TOOLTIP_ID}]`).clientHeight;
-  const leftOffset = 5;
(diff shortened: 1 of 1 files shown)


static/js/chart/draw_choropleth.ts
```diff
@@ -135,8 +135,11 @@ const onMouseMove = (
     }
   }
   const text = placeName + ": " + value;
-  const leftOffset = 3;
-  const topOffset = -3;
+  const tooltipHeight = document
+    .getElementById(containerId)
+    .querySelector(`[id=${TOOLTIP_ID}]`).clientHeight;
+  const leftOffset = 5;
+  const topOffset = -tooltipHeight - 5;
```
beets, line 142: can we turn 5 into a const?

beets (changes requested):
thanks for the fix!

beets (approved):

PR #516, opened 2020-10-08
rearrange the charts, and removed some from overview

beets:


beets:
@tjann for places with fewer < n charts, we default to only showing an overview page. trying to thin out the USA overview page

tjann (dismissed):

shifucun (approved):

PR #532, opened 2020-10-09
add some checks

chejennifer:
- add check to only call the choropleth APIs if dcid is from USA 
- add check to only draw choropleth if there is geojson data

### DIFF
--- a/static/js/place/place.ts
+++ b/static/js/place/place.ts
@@ -91,7 +91,7 @@ async function getGeoJsonData(
   dcid: string,
   placeType: string
 ): Promise<unknown> {
-  if (placeType == "Country" || placeType == "State") {
+  if (shouldMakeChoroplethCalls(dcid, placeType)) {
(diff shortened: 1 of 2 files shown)


static/js/place/place.ts
```diff
@@ -129,6 +129,12 @@ async function getLandingPageData(dcid: string): Promise<PageData> {
   });
 }
 
+function shouldMakeChoroplethCalls(dcid: string, placeType: string): boolean {
+  const isInUSA: boolean =
+    dcid.startsWith("geoId") || dcid.startsWith("country/USA");
+  return isInUSA && (placeType == "Country" || placeType == "State");
```
beets, line 135: we should also continue adding it for country (sorry if that unclear)

beets (approved):
thanks!

static/js/place/place.ts
chejennifer: Sorry, what do you mean? This check makes sure we are calling choropleth apis for countries and states that are in the USA

PR #543, opened 2020-10-09
Add housing unit by date built chart

shifucun:
Luckily this one won't need aggregation

image


beets (approved):
we should probably aggregate these too :) can do later though

static/js/shared/stats_var_titles.ts
```diff
@@ -392,6 +392,9 @@ const STATS_VAR_TITLES: { [key: string]: string } = {
   Count_HousingUnit_HomeValue1500000To1999999USDollar: "$1.5M - $2M",
   Count_HousingUnit_HomeValue2000000OrMoreUSDollar: "Over $2M",
 
+  Count_HousingUnit_OwnerOccupied: "Owner",
```
beets, line 395: These should be complete as a standalone label (they're used as titles for ranking pages)

Count of Owner Occupied Housing Units

beets: Sorry I didn't realize this for the other housing related stat vars as well

PR #551, opened 2020-10-12
Resolves issue where setState was not updating

edumorlom:
I have done some testing, and nothing seems to have broken.

Removing **shouldComponentUpdate => false** and deleting any this.forceUpdate() after this.setState().


beets (approved):

beets:
thanks for the update!

PR #552, opened 2020-10-12
Add Crime to Arson for consistency with other crimes, and also with i…

tjann:
…t's own rate chart.

### DIFF
--- a/server/chart_config.json
+++ b/server/chart_config.json
@@ -670,7 +670,7 @@
   },
   {
     "category": "Crime",
-    "title": "Arson",
+    "title": "Arson Crime",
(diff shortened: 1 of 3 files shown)


tjann:
image

image

beets (commented):
I understand the consistency issue, but I think by definition they are considered separately:

https://ucr.fbi.gov/crime-in-the-u.s/2010/crime-in-the-u.s.-2010/property-crime
https://ucr.fbi.gov/crime-in-the-u.s/2010/crime-in-the-u.s.-2010/property-crime/arsonmain

( I've wanted to do this change in the past, but held off on it). Wdyt?

tjann:
Being addressed in larger PR #623

PR #554, opened 2020-10-12
Housing title: change from value to price

shifucun:


server/chart_config.json
```diff
@@ -941,7 +941,7 @@
   },
   {
     "category": "Housing",
-    "title": "Housing Units by Value",
+    "title": "Housing Units by Price",
```
beets, line 944: Should we call it Count of Housing Units by Price?

beets: This is ok too

beets (approved):

shifucun:
thx for review!

PR #563, opened 2020-10-13
Fix denominators.

tjann:
Replace denoms as fitting.


server/chart_config.json
```diff
@@ -28,9 +28,9 @@
     "statsVars": ["Count_Person_Employed"],
     "isOverview": true,
     "relatedChart": {
-      "title": "Percentage of People Employed",
+      "title": "Employment Rate",
```
tjann, line 31: Hi @beets , @shifucun -- I do not know if I made the right call here. Although "Employed" is a subclass of "InLaborForce", this stat is not very helpful as it is simply 1 - unemploymentRate, which we already have.

% of People Employed, on the other hand, is interesting to those who think the way "unemployment" is measured (% of labor force) is unfair.

server/chart_config.json
```diff
@@ -43,11 +43,23 @@
     "relatedChart": {
       "title": "Labor Force Participation Rate",
       "scale": true,
-      "denominator": "Count_Person",
+      "denominator": "Count_Person_16OrMoreYears",
```
tjann, line 46: @shifucun @beets 

This appears to have worked out (thanks to code elsewhere, likely by luck).

Count_Person_InLaborForce exists both from BLS and from WDI. The overlap is only on country/USA and we seem to pick WDI. Bar charts comparing USA with other countries LG since all use WDI. Bar charts for places in USA almost look good:

image


It says "places in" instead of "USA and places in" so it seems ok that USA is missing (likely bc different source?)

Furthermore, our code appears to be smart and only graphs one of the Labor Force Participation Rate charts (the relatedChart config here and the chart config below both have the same title). In this case, the one from WDI:

image

tjann:
Some sxs to help ease approval:

- Sanity check the bar chart latest dates: USA top charts [local](https://screenshot.googleplex.com/7xUqEbf9H3pebkP) [staging](https://screenshot.googleplex.com/Ah6YW7ovrewbrz8)

- Title change and denom change in employment rate; denom change for labor force participation rate (local, then staging):
image
image

- Denom change for UI Claims (local, then staging):
image
image

- Labor force participation rate splitting: [local](https://screenshot.googleplex.com/4SwqUCDZJfeWy9z) [staging](https://screenshot.googleplex.com/7m6n6vwuSikZr5E)

- Denom change for individual income (left: local, right: staging)
Screen Shot 2020-10-13 at 7 58 18 PM

tjann:
[continued]
- Labor force close up: note that the other countries chart was originally computed from the WDI Count_Person_InLaborForce with Count_Person as the denom. Now, it is using the WDI StatVar Count_Person_15To64Years_InLaborForce_AsFractionOf_Count_Person_15To64Years
Screen Shot 2020-10-13 at 8 10 42 PM

server/routes/api/landing_page.py
```diff
@@ -317,28 +321,60 @@ def scale_series(numerator, denominator):
...
     year alignment to match two date. If no denominator is found for a
-    numerator, then the data is removed.
+    numerator, then the data is removed. We allow the denominator to be
+    MAX_DENOMINATOR_BACK_YEAR stale.
     """
     data = {}
+
+    num_date_fmt = None
+    denom_date_fmt = None
+    for date in denominator.keys():
```
beets, line 331: can we do this for just one of the keys? i imagine in most cases the format will be similar across keys. this is really expensive if the format ends up being y-m-d and it's a lot of datapoints

i'll also file a bug so we always return sorted dates from mixer to avoid doing all this sorting in python

server/routes/api/landing_page.py
```diff
@@ -317,28 +321,60 @@ def scale_series(numerator, denominator):
...
+        if denom_date_fmt == '%Y':
+            for i in range(0, MAX_DENOMINATOR_BACK_YEAR + 1):
+                year = str(date_obj.year - i)
+                if year in denominator:
+                    if denominator[year] > 0:
+                        data[date] = value / denominator[year]
+                    break
+        elif denom_date_fmt == '%Y-%m':
+            for i in range(0, 12 * MAX_DENOMINATOR_BACK_YEAR + 1):
+                year_month = datetime.datetime(date_obj.year, date_obj.month,
```
beets, line 373: not sure how efficient these operations are. may be worth doing something quick and dirty with some string manipulation and abs math

tjann:
@beets  since there are many calls for me to do other things, I have reverted all date munging to Bo's original functions. I'll try your suggestions later on.

beets (approved):
ok thanks. yea let's keep an eye on the other changes that we punted

PR #564, opened 2020-10-13
Add check to prevent choropleth being drawn for puerto rico

chejennifer:
d3 doesn't support drawing choropleth maps for Puerto Rico so choropleth charts for Puerto Rico appear empty. Adding this check for now while we look into npm packages that can allow for drawing Puerto Rico choropleth.


beets (approved):
please add a comment there with an explanation or link to an issue

PR #577, opened 2020-10-14
Decouple choropleth promises from main data fetch

chejennifer:
decouple the getting geojson and getting choropleth data from the main data fetch. Will wait on the promises in Chart.tsx instead of places.tsx.


static/js/place/place.ts
```diff
@@ -147,11 +147,7 @@ function renderPage(): void {
   const chartGeoJsonPromise = getGeoJsonData(dcid, placeType);
```
beets, line 147: can we return empty promises for these if not in USA and place type != country / state / county

static/js/place/chart_block.tsx
```diff
@@ -266,39 +266,31 @@ class ChartBlock extends React.Component<ChartBlockPropType, unknown> {
       }
       if (
         !!this.props.data.isChoropleth &&
```
beets, line 268: also add the isUsaPlace check if you still see a flash of content (was removed for the other comparisons)

static/package-lock.json
```diff
@@ -10,7 +10,7 @@
       "integrity": "sha512-a9gxpmdXtZEInkCSHUJDLHZVBgb1QS0jhss4cPP93EW7s+uC5bikET2twEF3KV+7rDblJcmNvTR7VJejqd2C2g==",
```
beets, line 10: please remove this file from the PR

static/js/place/chart.tsx
```diff
@@ -161,7 +164,8 @@ class Chart extends React.Component<ChartPropType, ChartStateType> {
     }
     if (
       this.props.chartType === chartTypeEnum.CHOROPLETH &&
-      (!this.state.choroplethDataGroup ||
+      (_.isEmpty(this.state.choroplethDataGroup) ||
```
beets, line 167: move this check to 443 (i.e. don't set state if these conditions hold, and set state.display = false

beets (commented):
thanks for the change. a few optimization suggestions...

beets (approved):

PR #578, opened 2020-10-14
Add Redis service as memcache for staging. Need to verify on staging

shifucun:


server/routes/place.py
```diff
@@ -14,11 +14,8 @@
 """Place Explorer related handlers."""
 
 import flask
-import services.datacommons as dc
 import routes.api.place as place_api
 
-from cache import cache
```
beets, line 20: we should probably add more to cache yea?

beets (approved):

PR #588, opened 2020-10-14
try to fix https -> http redirect in prod

beets:


server/routes/redirects.py
```diff
@@ -30,31 +30,35 @@ def kg():
...
+                    _scheme='https')
 
 
 # This is used to handle explore more link from Google search. Do not remove.
+# arg params from search: mprop, dcid, popt
 @bp.route('/explore/place')
 def explore():
-    return redirect(url_for('place.place', dcid=request.args.get('dcid')),
+    return redirect('https://datacommons.org' +
+                    url_for('place.place', dcid=request.args.get('dcid')),
```
tjann, line 63: did you want to add  _scheme='https' here?

tjann (approved):

shifucun (approved):

server/routes/redirects.py
beets: don't need to since we are specifying it. and it turns out _scheme seems to not compile / test for us. going to try without it and see if it works. i wish there was a better way to test this :(

PR #591, opened 2020-10-14
[PV Tree Generator] restrict mprop to count for denominators

intrepiditee:
Only StatVars with an mprop of "count" can be used as denominators. (The diff for `hierarchy_statsvar.json ` can be loaded)

Thanks for reviewing.


shifucun (approved):
nice, thanks for the fix!

PR #594, opened 2020-10-14
Use first autocomplete result when pressing enter in search

### LINKED ISSUE #522: enter in search box should use first autocomplete results

chejennifer:
fixes #522 
changes:
- if autocomplete object isn't selected (aka user isn't on one of the autocomplete suggestions), get the list of autocomplete predictions for what the user has entered and redirect according to the first prediction in the list.  
searchFix

### DIFF
--- a/static/js/place/search.ts
+++ b/static/js/place/search.ts
@@ -44,6 +44,7 @@ function initSearchAutocomplete(): void {
 function placeChangedCallback(): void {
   // Get the place details from the autocomplete object.
   const place = ac.getPlace();
+  // place won't have place_id information if no autocomplete object has been selected (ie. user did not select any of the autocomplete suggestions)
   if (!place.place_id) {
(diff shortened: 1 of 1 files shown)


static/js/place/search.ts
```diff
@@ -32,30 +33,63 @@ function initSearchAutocomplete(): void {
...
 /*
- * Get place from autocomplete object and update url
+ * If the autocomplete object does not have a place_id, query the autocomplete service. Otherwise, get url for the place.
  */
-function getPlaceAndRender(): void {
+function placeChangedCallback(): void {
   // Get the place details from the autocomplete object.
   const place = ac.getPlace();
+  if (!place.place_id) {
+    queryAutocompleteService(place.name);
```
beets, line 48: when do you fall into this case? could you add a comment

beets (approved):
very cool, thanks also for the video

PR #595, opened 2020-10-14
[PV Tree Generator] allow incrementalCount and cumulativeCount

intrepiditee:
Besides "count", StatVars with mprop "incrementalCount" and "cumulativeCount" can also be used as denominators.

Thanks for reviewing.


shifucun (approved):
thanks for the update!

PR #615, opened 2020-10-15
Handle place highlight undefined object error

chejennifer:
this was the original problem: 
Screenshot 2020-10-15 at 10 48 52 AM


beets (approved):
thanks for this fix!

PR #618, opened 2020-10-15
Do not use redis cache for staging website

shifucun:
It's tedious to clear the cache for every code commit.


beets (approved):
ok, but it would be good to be able to test it some other way. maybe a different website instance?

PR #622, opened 2020-10-15
Add click target to choropleth chart

chejennifer:
- When clicking on a region in the choropleth chart, will open a new tab with the place page for the place that was clicked
clickTarget


static/js/chart/draw_choropleth.ts
```diff
@@ -152,6 +152,20 @@ const onMouseMove = (
     .style("top", d3.event.offsetY + topOffset + "px");
 };
 
+const onMapClick = (domContainerId: string) => (
+  geo: { properties: { geoDcid: string } },
+  index
+) => {
+  window.open(REDIRECT_BASE_URL + geo.properties.geoDcid, "_blank");
```
shifucun, line 159: nit: In a follow up PR, it would be even better to carry out the topic from the url, which we don't do in bar chart url either.

shifucun (approved):
nice!

static/js/chart/draw_choropleth.ts
chejennifer: should I try to make the bar chart url also carry out the topic from the url then?

shifucun: Yep!

PR #623, opened 2020-10-15
Add units and sentence casing, other English edits.

tjann:


tjann:
My first attempt resulted in kkg and kkWh (auto unit k for values over 1000). My made a change to help with that, but it's still potentially confusing:
image



Sentence casing:
image

static/js/chart/base.ts
```diff
@@ -271,9 +271,12 @@ function formatYAxisTicks(
   }
   const dollar = unit === "$" ? "$" : "";
   const percent = unit === "%" ? "%" : "";
+  const tons = unit === "t" ? "t" : "";
```
beets, line 274: should we also add these units to the chart config?

beets (approved):
just one question

beets:
also, adding a space before the unit sgtm

tjann:
I added a space before units (not $ or %, which are special). So this is what it looks like now (first screenshot showing k kg and k kWh, second screenshot showing $ and %):

image
image

server/chart_config.json
```diff
@@ -679,41 +677,38 @@
   },
   {
     "category": "Crime",
-    "title": "Total Crime",
+    "title": "Total crime",
     "statsVars": ["Count_CriminalActivities_CombinedCrime"],
     "isOverview": true,
     "relatedChart": {
-      "title": "Total Crime Rate",
+      "title": "Crimes per 100,000 people",
```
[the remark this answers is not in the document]
tjann: Since the numerator is #crimes and denominator is #people, % does not make sense. I chose 100k since it is large enough for most places and commonly used:
image

static/js/chart/base.ts
tjann: Yep, added them back! I removed them due to the kkg issue that is now addressed by adding a space.

beets (approved):

PR #626, opened 2020-10-16
Proof concept to add GO backend and co-exist with Flask

shifucun:
This is a proof of concept to use go backend together with the current Flask server. This goal is to:
- write new endpoint in golang
- gradually migrate python to golang
- finally deprecate the flask server.

I have done the gcloud app deploy locally and can see the response from go service: https://staging.datacommons.org/go/

### DIFF
--- a/dispatch.yaml
+++ b/dispatch.yaml
@@ -1,0 +1,22 @@
+# Copyright 2020 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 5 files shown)


dispatch.yaml
```diff
@@ -0,0 +1,22 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+dispatch:
+  # Send all traffic to the Flask server.
+  - url: "*/"
+    service: default
+
+  # Send specific service to go server.
+  - url: "*/go/"
```
beets, line 21: this is interesting in that ordering doesn't matter.

PR #628, opened 2020-10-16
Update type for geojson data

chejennifer:
-  for geoJsonData props, replace type of unknown with newly created type of GeoJsonData

### DIFF
--- a/static/js/place/types.ts
+++ b/static/js/place/types.ts
@@ -102,3 +102,25 @@ export interface ChoroplethDataGroup {
   exploreUrl: string;
   sources: string[];
 }
+
+export interface GeoJsonData {
+  type: string;
+  features: GeoJsonFeature[];
+  properties: {
+    current_geo: string;
+  };
+}
+
+export interface GeoJsonFeature {
+  type: string;
+  geometry: {
+    type: string;
+    coordinates: number[][][][];
+  };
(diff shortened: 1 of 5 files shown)


static/js/place/types.ts
```diff
@@ -102,3 +102,25 @@ export interface ChoroplethDataGroup {
...
+  properties: {
+    current_geo: string;
+  };
+}
+
+export interface GeoJsonFeature {
+  type: string;
+  geometry: {
+    type: string;
+    coordinates: number[][][][];
```
beets, line 118: could you add a comment that we always return MultiPolygon? So, coordinates are always an array of Polygon coordinate arrays. Not sure if we want to unpack this 4-dimension array to Polygon[] and Polygon is number[][]?

https://tools.ietf.org/html/rfc7946#section-3.1.7

beets (commented):
thanks for unpacking this!

static/js/place/types.ts
chejennifer: I can add the comment. I was also considering unpacking the 4-dimension array. I think it would make things more clear, but I wasn't sure if it was necessary since we don't use a single Polygon.

beets (approved):
this is enough granularity - thanks for the change!

PR #631, opened 2020-10-19
draw choropleth for puerto rico

chejennifer:
- use package "geo-albers-usa-territories" so puerto rico choropleth charts can be drawn
- use fitSize method from https://bl.ocks.org/HarryStevens/0e440b73fbd88df7c6538417481c9065 for fitting and centering the map
- when getting geoJson for puerto rico, use the highest resolution version of simplified geojsons because lower resolutions will look like a mosaic

image
image

### DIFF
--- a/server/routes/api/chart.py
+++ b/server/routes/api/chart.py
@@ -88,6 +88,7 @@ def get_choropleth_places(geoDcid):
                                                           display_level).get(
                                                               parent_dcid, [])
                     geo_prop = CHOROPLETH_GEOJSON_PROPERTY_MAP[display_level]
+                    # Puerto Rico (geoId/72) requires higher resolution geoJson
                     if parent_dcid == 'geoId/72':
                         geo_prop = 'geoJsonCoordinatesDP1'
                     return place_list, geo_prop
@@ -96,6 +97,7 @@ def get_choropleth_places(geoDcid):
         place_list = dc_service.get_places_in([geoDcid],
                                               display_level).get(geoDcid, [])
         geo_prop = CHOROPLETH_GEOJSON_PROPERTY_MAP[display_level]
+        # Puerto Rico (geoId/72) requires higher resolution geoJson
         if geoDcid == 'geoId/72':
             geo_prop = 'geoJsonCoordinatesDP1'
         return place_list, geo_prop
--- a/server/routes/api/chart.py
+++ b/server/routes/api/chart.py
@@ -88,12 +88,16 @@ def get_choropleth_places(geoDcid):
                                                           display_level).get(
                                                               parent_dcid, [])
                     geo_prop = CHOROPLETH_GEOJSON_PROPERTY_MAP[display_level]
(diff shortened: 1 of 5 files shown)


server/routes/api/chart.py
```diff
@@ -88,12 +88,16 @@ def get_choropleth_places(geoDcid):
                                                           display_level).get(
                                                               parent_dcid, [])
                     geo_prop = CHOROPLETH_GEOJSON_PROPERTY_MAP[display_level]
+                    if parent_dcid == 'geoId/72':
```
beets, line 91: please add a comment to explain this

static/js/chart/draw_choropleth.ts
```diff
@@ -37,6 +38,28 @@ const LEGEND_IMG_WIDTH = 10;
...
 
+/**
+ * From https://bl.ocks.org/HarryStevens/0e440b73fbd88df7c6538417481c9065
+ * scales and translates the projection to allow resizing of the choropleth map
+ */
+function fitSize(
+  width: number,
+  height: number,
+  object: any,
+  projection: any,
```
beets, line 49: could we try to get the type for this (since we are calling operations on it)
for object as well, since we have the geojson type now

static/js/chart/draw_choropleth.ts
```diff
@@ -75,12 +98,12 @@ function drawChoropleth(
...
 
-  // Scale and center the map.
-  const projection = d3
-    .geoAlbersUsa()
-    .fitSize([chartWidth - LEGEND_WIDTH, chartHeight], geoJson);
+  const projection = geo.geoAlbersUsaTerritories();
   const geomap = d3.geoPath().projection(projection);
 
+  // Scale and center the map
+  fitSize(chartWidth - LEGEND_WIDTH, chartHeight, geoJson, projection, geomap);
```
beets, line 105: is this something required for all maps, or only puerto rico? asking since it looks like an expensive operation (e.g. calling bounds on the geojson)

beets (commented):
thanks!

static/js/chart/draw_choropleth.ts
chejennifer: this is something we would need for all maps to scale and center them. We used to do this for all maps with a built in d3-geo function

beets: ah yes, sorry i missed that earlier

static/js/place/types.ts
```diff
@@ -103,27 +103,21 @@ export interface ChoroplethDataGroup {
...
-  };
-  id: string;
+export type GeoJsonFeature = GeoJSON.Feature<
+  GeoJSON.MultiPolygon,
+  GeoJsonFeatureProperties
+>;
+
+export interface GeoJsonData extends GeoJSON.FeatureCollection {
+  type: "FeatureCollection";
+  features: Array<GeoJsonFeature>;
```
beets, line 119: neat!

beets (approved):

PR #634, opened 2020-10-21
Fix deploy error: need 'cd server' so requirements-core.txt can be found

shifucun:
Also "set -e" to exit immediately

### DIFF
--- a/server/webdriver_tests/screenshot/screenshot_test.py
+++ b/server/webdriver_tests/screenshot/screenshot_test.py
@@ -91,7 +91,7 @@ TEST_URLS = [
     },
     {
         'url':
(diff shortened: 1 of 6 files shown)


beets (approved):
nice find

server/webdriver_tests/screenshot/screenshot_test.py
```diff
@@ -91,7 +91,7 @@
     },
     {
         'url':
-            '/ranking/Count_Person_BelowPovertyLevelInThePast12Months_NativeHawaiianOrOtherPacificIslanderAlone/City/geoId/06085?h=geoId%2F0649670&pc=1&scaling=100&unit=%25',
+            '/ranking/Count_Person_BelowPovertyLevelInThePast12Months_AsianAlone/City/geoId/06085?h=geoId%2F0649670&pc=1&scaling=100&unit=%25',
```
shifucun, line 94: @beets need to change this as it's not passing, seems to be a data issue?

PR #639, opened 2020-10-22
Get year in a faster way

shifucun:
get_year() function takes a substantial portion of cpu time:

image


tjann (approved):
LG for now. Still need to think about month/date later.

beets (approved):

PR #641, opened 2020-10-22
sort range aggregration list for consistent result

shifucun:
Not necessary for actual logic but ensure tests always pass


pradh (approved):
thanks

PR #649, opened 2020-10-27
Automate downloading latest version of Chrome Driver

edumorlom:
Not sure if you like this idea, but I thought I'd propose it.

Instead of manually downloading the latest Chrome Driver version.

1. Curl https://chromedriver.storage.googleapis.com/LATEST_RELEASE
2. Store the response as **CHROMEDRIVERV**.
3. Download https://chromedriver.storage.googleapis.com/${**CHROMEDRIVERV**}/chromedriver_linux64.zip


shifucun (commented):
This is good. Can you add a comment to check that the chrome driver version is compatible with local chrome version

edumorlom:
done! thanks

shifucun (approved):

shifucun:
thanks for the update!

PR #650, opened 2020-10-27
use lower res geojson for choropleth tool

chejennifer:
Use lower resolution geojson for the choropleth tool. Remove getting city level geojson for now because we don't currently have the lower resolution geojson data for that.


shifucun (approved):

shifucun:
can you check if the webdriver test is a real failure or due to short sleep(). If it's the latter case, can you increase the sleep time a bit.

PR #655, opened 2020-10-29
Add i18next to dev page

tjann:


locales/en-US/demographic_enums.json
```diff
@@ -0,0 +1,4 @@
+{
+  "Total": "Total",
+  "Male": "Masculino"
```
shifucun, line 3: this is not en-US?

static/js/chart/draw.ts
```diff
@@ -26,6 +26,60 @@ import {
   formatYAxisTicks,
 } from "./base";
 
+import i18next from "i18next";
+import Backend from "i18next-http-backend";
+import Cache from "i18next-localstorage-cache";
+import postProcessor from "i18next-sprintf-postprocessor";
+import LanguageDetector from "i18next-browser-languagedetector";
+
+i18next
```
shifucun, line 35: should this init be done in the javascript entry point? like place.ts?

static/js/chart/draw.ts
```diff
@@ -57,6 +111,15 @@ const AXIS_GRID_FILL = "#999";
 // Max Y value used for y domains for charts that have only 0 values.
 const MAX_Y_FOR_ZERO_CHARTS = 10;
 
+function translate(text: string): string {
```
shifucun, line 114: nit: name the function "translated"

static/js/chart/draw.ts
```diff
@@ -57,6 +111,15 @@ const AXIS_GRID_FILL = "#999";
...
+  if (i18next.t(text)) {
+    text = i18next.t(text);
+  }
+  return text;
+}
+function translate_key(key: { label: string; link?: string }) {
```
shifucun, line 120: nit: can make this type an interface and have a member function "translate()"

static/js/chart/draw.ts
tjann: I see, to reduce conflict with d3. Done.

static/js/chart/draw.ts
tjann: I realize this is actually not needed. Now it's just translated

shifucun (approved):
very impressive pioneer work!

tjann:
Was able to block translation for local in d648d70 by returning '' instead of the json file contents when FLASK_ENV != prod. I changed it to == prod before commiting.

PR #659, opened 2020-10-31
Adds the production flag -o to run_test and cleans package.json

edumorlom:
1. Added -o flag for production-only dependancies in run_tests.sh.
2. Cleaned package.json. All tests pass.
3. Excludes test files from npm run build in tsconfig.json.


run_test.sh
```diff
@@ -178,6 +189,7 @@ while getopts pwblcsaf OPTION; do
     esac
 done
 
-if [ $OPTIND -eq 1 ]; then
```
edumorlom, line 181: I fixed a small bug here.
According to help documentation, "no args" should run_all_tests but instead it displayed HELP.

edumorlom: Also, bash was complaining about the ; before "then". I removed it.

shifucun: I think the documentation is off.. the original behavior is better: do nothing if no option is provided. Can we just modify the documentation?

shifucun (commented):
There are some other places need to add the -o flag:

cloudbuild.deploy.yaml
cloudbuild.screenshot.yaml

edumorlom:
Thanks for the review, ready for re-review!

shifucun:
thanks for update, very good cleanup!

shifucun (approved):

PR #664, opened 2020-11-04
Choropleth Tool Improvements

chejennifer:
- Few simple changes to make title a little more readable: remove underscore, few replacements in the words in the title. More efforts will be needed to make title readable
- Show the selected stat var in the menu when loading the choropleth tool for the first time
- keep the stat var menu open and keep the current option selected when navigating to different place (ie. when clicking on a region in the choropleth or when clicking a breadcrumb)

test (14)

### DIFF
--- a/static/js/tools/choropleth/choropleth.tsx
+++ b/static/js/tools/choropleth/choropleth.tsx
@@ -28,7 +28,26 @@ const TOOLTIP_ID = "tooltip";
 const LEGEND_HEIGHT = 60;
 const LATEST_DATE = "latest";
 const MAP_PADDING = 7;
+const LEGEND_WIDTH = 300;
+const TICK_SIZE = 6;
+const NUM_TICKS = 5;
+const LEGEND_MARGIN_TOP = 18;
+const LEGEND_MARGIN_BOTTOM = 16 + TICK_SIZE;
+const LEGEND_MARGIN_SIDES = 25;
+const LEGEND_TEXT_PADDING = 6;
+const NAME_REPLACEMENT_DICT = {
+  "Count Person": "Population",
+  "Count Worker": "Workers in",
+  "Count Household": "Households with",
+  "Person": "",
+};
+
 type PropsType = unknown;
+
+type BreadCrumbType = {
+  geoId: string;
+  geoName: string;
+};
 
 // TODO(eduardo): get rid of "unknown" type.
 type StateType = {
@@ -39,6 +58,7 @@ type StateType = {
   mapContent: unknown;
   pc: boolean;
   popMap: { [geoId: string]: number };
+  breadCrumbs: BreadCrumbType[];
 };
 
 class ChoroplethMap extends Component<PropsType, StateType> {
@@ -58,6 +78,7 @@ class ChoroplethMap extends Component<PropsType, StateType> {
       pc: isPerCapita,
       popMap: {},
       values: {},
+      breadCrumbs: [],
     };
     this._handleWindowResize = this._handleWindowResize.bind(this);
   }
@@ -236,7 +257,7 @@ class ChoroplethMap extends Component<PropsType, StateType> {
     // Generate breadcrumbs.
     // TODO(fpernice-google): Derive the curGeo value from geoDcid instead
     // of embedding in url.
-    generateBreadCrumbs(this.state.geoJson.properties.current_geo);
+    this.generateBreadCrumbs(this.state.geoJson.properties.current_geo);
     addTooltip();
   };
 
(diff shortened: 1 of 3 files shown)


static/js/tools/choropleth/choropleth.tsx
```diff
@@ -576,6 +608,74 @@ class ChoroplethMap extends Component<PropsType, StateType> {
...
+  /**
+   * Redirects the webclient to a particular geo. Handles breadcrumbs in redirect.
+   * @param {string} geoDcid to redirect to.
+   * @param {string} curGeo human-readable current geo, e.g. United States when geoDcid
+   *                 is country/USA.
+   */
+  private redirectToGeo = (geoDcid: string, curGeo: string): void => {
+    const url = new URL(window.location.href);
+    const currGeoId = url.searchParams.get("geoDcid");
+    let baseUrl = "/tools/choropleth";
```
shifucun, line 620: js string template would be better:

`/tools/choropleth${buildChoroplethParams()}&geoDcid=${geoDcid}&bc=`

static/js/tools/choropleth/choropleth.tsx
```diff
@@ -576,6 +608,74 @@ class ChoroplethMap extends Component<PropsType, StateType> {
...
+            </span>
+          );
+        }
+      });
+      breadcrumbsDisplay.push(<span key={curGeo}>{curGeo}</span>);
+    }
+    // Add breadcrumbs + current geoId
+    // Example: "USA" > "FL" > "Miami-Dade"
+    // Where Miami-Dade is the current id
+    ReactDOM.render(breadcrumbsDisplay, document.getElementById("breadcrumbs"));
```
shifucun, line 676: Might not be good to call ReactDOM.render in a react Component. How about set the state of a children component for the bradcrumbs?

static/js/tools/choropleth/choropleth.tsx
beets: you could also use [URLSearchParams](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams)

static/js/tools/choropleth/choropleth.tsx
```diff
@@ -576,6 +608,74 @@ class ChoroplethMap extends Component<PropsType, StateType> {
...
+      breadCrumbsList.push({ geoId: currGeoId, geoName: curGeo });
+    } else {
+      breadCrumbsList = breadCrumbsList.slice(0, idxOfCurr);
+    }
+    this.setState({
+      breadCrumbs: breadCrumbsList,
+    });
+    for (let i = 0; i < breadCrumbsList.length; i++) {
+      baseUrl += i === 0 ? "" : ";";
+      baseUrl += breadCrumbsList[i].geoId + "~" + breadCrumbsList[i].geoName;
```
beets, line 639: would be good to use a constant for the separator (helps with readability and maintainability too)

static/js/tools/choropleth/choropleth.tsx
```diff
@@ -576,6 +608,74 @@ class ChoroplethMap extends Component<PropsType, StateType> {
...
+    history.pushState({}, null, baseUrl);
+    this.loadGeoJson();
+  };
+
+  /**
+   * Generates the breadcrumbs text from browser url.
+   * @param {string} curGeo human-readable current geo to display
+   * at end of list of hierarchy of locations.
+   */
+  private generateBreadCrumbs = (curGeo: string): void => {
```
beets, line 650: Ideally, we should turn the breadcrumbs into a component. But we can add it as a todo for now

beets (commented):
There's some logic from the timelines tool for the stat var title that we should try to bring in. Could you add a TODO for that?

static/js/tools/choropleth/choropleth.tsx
```diff
@@ -615,12 +617,14 @@ class ChoroplethMap extends Component<PropsType, StateType> {
...
-    baseUrl += "&geoDcid=" + geoDcid;
-    baseUrl += "&bc=";
+    const currGeoId = new URLSearchParams(window.location.search).get(
+      "geoDcid"
+    );
+    let baseUrl = `/tools/choropleth${buildChoroplethParams([
+      "statVar",
+      "pc",
+      "mdom",
+    ])}&&geoDcid=${geoDcid}&bc=`;
```
beets, line 627: double ampersands? also looks like you could pass in geoDcid into the list of args for buildChoroplethParams to preserve.

chejennifer: I will fix the double ampersands. buildChoroplethParams will set values to what is currently in the url, but we want the geoDcid for the next geo we're redirecting to

beets: Ah yes, thanks for catching that. Readability nit for future: the code should be localized with where it's used (i.e. get currGeoId right before using it).

I'll approve and let's submit.

beets (approved):

PR #672, opened 2020-11-12
WIP i18n string extraction

beets:
- install i18next-parser package and start extraction of place explorer
strings
- to run:
  cd static; npm install; cd ..
  static/node_modules/i18next-parser/bin/cli.js -c static/i18next-parser.config.js

### DIFF
--- a/static/i18next-parser.config.js
+++ b/static/i18next-parser.config.js
@@ -1,0 +1,26 @@
+/**
+ * Copyright 2020 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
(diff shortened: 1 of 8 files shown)


static/i18next-parser.config.js
```diff
@@ -0,0 +1,26 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * Configuration for i18next-parser. See https://github.com/i18next/i18next-parser
+ * NOTE: run i18next-parser from the root of the repo
+ */
+module.exports = {
+  locales: ["en-US"],
```
tjann, line 22: if this only exports en-US, where did the es/place.json come from?

tjann: I see, I think you handwrote the files? I ran the code and the namespace was translation.json not place.json.

static/js/place/main.tsx
```diff
@@ -96,13 +97,18 @@ class MainPane extends React.Component<MainPanePropType> {
...
                   <a href={`/place/${this.props.dcid}?topic=${topic}`}>
-                    More charts ›
+                    <Trans i18nKey="more_charts">More charts ›</Trans>
                   </a>
                 </span>
               </h3>
             );
           } else {
-            subtopicHeader = <h3 id={topic}>{topic}</h3>;
+            // XXX find a way to translate these.
```
tjann, line 106: I think maybe we don't need to have the topic be auto extracted since they are in the chart config? We can write our extractor for chart config

beets: I was testing if the topic would get translated (hence the manually crafted es.json file)

PR #674, opened 2020-11-12
Move Webdriver test from cloudbuild.py.yaml to cloudbuild.screenshot.yaml

edumorlom:
1. Per Carolyn's request, move the webdriver test step from cloudbuild.py.yaml to cloudbuild.screenshot.yaml 
2. Added appropiate waitFor attribute to cloudbuild.screenshot.yaml 
3. Removed package_js step from cloudbuild.py.yaml since it is no longer needed.
4. Improve speed of package_js step in cloudbuild.screenshot.yaml by re-using node_modules from node docker image.


shifucun (approved):
Thanks for making the change!

Any luck on parallel run the pytest for webdriver tests?

PR #677, opened 2020-11-14
Extract chart titles for translations

beets:
Added titleId and some chart descriptions to chart config
Adds a new script to extract titles from chart config
POC for using the translated string in the returned landing page response.
Lots of TODOs:
- select the right locale in the python module
- finish adding chart descriptions
- figure out how to handle the related chart titles

image


server/routes/api/landing_page.py
```diff
@@ -39,6 +40,10 @@
 MIN_CHART_TO_KEEP_TOPICS = 30
 OVERVIEW = 'Overview'
 
+l10n = gettext.translation('chart_titles', localedir='l10n', languages=['es'])
```
tjann, line 43: Could add an inline TODO here to replace es hardcoding with locale detection. Optional.

tools/i18n/chart_config_extractor.py
```diff
@@ -0,0 +1,84 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Script to generate a .pot file from the chart config, used for translations.
+
+To run: python3 tools/i18n/chart_config_extractor.py
+
+This will produce /server/l10n/chart_titles.pot
+Then, run pybabel init -l $LOCALE -i server/l10n/chart_titles.pot -d server/l10n -D chart_titles
+which will initialize a chart_titles.po file for the locale.
```
tjann, line 20: This is just for development right? We'll eventually make .po files from TC output?

tjann (approved):

server/routes/api/landing_page.py
beets: done

tools/i18n/chart_config_extractor.py
beets: yep

PR #678, opened 2020-11-17
Refactor translator into react component

shifucun:
Plan to have a more realistic playground for external users to use with their own Bigquery Dataset. This would pave the path for that.


static/js/translator/constant.ts
```diff
@@ -0,0 +1,298 @@
+/**
```
beets, line 1: nit: constants.ts

static/js/translator/page.tsx
```diff
@@ -0,0 +1,240 @@
+/**
```
beets, line 1: nit: please keep each file to one component

beets (commented):
curious about the path to that -- would we have a mixer api to any bigquery?

left small comments, mostly style nits.

shifucun (commented):
Updated

static/js/translator/page.tsx
shifucun: Done

beets (approved):
thanks for making the changes!

PR #679, opened 2020-11-17
Added descriptions for remaining charts

beets:
Also updated the description template used for the translation messages.


server/chart_config.json
```diff
@@ -419,6 +442,7 @@
     "category": "Health",
     "titleId": "CHART_TITLE-Mortality_events",
     "title": "Mortality events",
+    "description": "Count of deaths caused by various diseases",
```
tjann, line 445: Nit: I think these mortality events are not necessarily due to disease. Some are external circumstances

tjann (approved):

server/chart_config.json
beets: thanks! updated

PR #682, opened 2020-11-18
Various fix for Go service

shifucun:
- copy js bundles into /go folders as the go service need to have its own static files.
- gin-gonic uses port in the form of ":8080" instead of "8080"
- Golang template does not process {# #} comments. It can remove <-- --> in the processed html file.


beets (approved):

shifucun:
thanks!

PR #685, opened 2020-11-23
Move all the client side api call to server side

shifucun:
This is to align the behavior with place and timeline tools. It also prepares isolated mixer for website setup, so a website server uses a dedicated mixer server. 

Tested locally and do not see performance degradation. Will verify after staging push.

### DIFF
--- a/static/js/browser/kg.js
+++ b/static/js/browser/kg.js
@@ -453,6 +453,10 @@ function groupObs(obsArray) {
   return result;
 }
 
+function getTriples(dcid) {
+  return axios.get(`/api/browser/triples/${dcid}`).then((resp) => resp.data);
+}
+
 /**
  * Render KG page.
  *
@@ -464,7 +468,7 @@ function groupObs(obsArray) {
  * @param {!Iterable} outArcs The out arc triples.
  * @param {!Object} provDomain Provenance domain.
  */
-function renderKGPage(
+async function renderKGPage(
   dcid,
   type,
   name,
@@ -495,9 +499,7 @@ function renderKGPage(
     popId = dcid;
   }
   if (popId) {
-    let popTriples = util.sendRequest(`/node/triples?dcids=${popId}`, false)[
-      popId
-    ];
+    let popTriples = await getTriples(popId);
     let ts = popTriples.filter((t) =>
       ["location", "childhoodLocation"].includes(t["predicate"])
     );
@@ -513,7 +515,9 @@ function renderKGPage(
   }
 
   // Get popobs
-  const popobs = util.sendRequest(`/bulk/pop-obs?dcid=${locId}`, true);
+  const popobs = await axios
+    .get(`/api/browser/popobs/${locId}`)
+    .then((resp) => resp.data);
 
   // Get out arcs of population from popobs information.
   if (util.isPopulation(type)) {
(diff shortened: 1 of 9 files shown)


beets (commented):
I'm finding it really hard to do review this change, especially since there is also a move / change in where/how we are handling the triples request. I'm weary of adding an additional proxy for these calls in flask -- have you noticed any slowdowns?

Also, as discussed offline, this isn't blocking the container change.

Since our testing coverage for the browser is poor, could we split out the code deletion change into a separate PR (it will also help with rollbacks and reduce the diff)?

shifucun:
> I'm finding it really hard to do review this change, especially since there is also a move / change in where/how we are handling the triples request. I'm weary of adding an additional proxy for these calls in flask -- have you noticed any slowdowns?
> 
> Also, as discussed offline, this isn't blocking the container change.
> 
> Since our testing coverage for the browser is poor, could we split out the code deletion change into a separate PR (it will also help with rollbacks and reduce the diff)?

I will split the PR into small ones. Sent out https://github.com/datacommonsorg/website/pull/686/files first.

static/js/browser/kg.js
```diff
@@ -938,63 +938,60 @@ function renderKGPage(
...
-    if (prov["predicate"] == "typeOf" && !!prov["subjectName"]) {
-      provDomain[prov["subjectId"]] = new URL(prov["subjectName"]).host;
-    }
-  }
-
   const dcid = document.getElementById("bg-dcid").textContent;
+  const provPromise = getTriples("Provenance");
+  const triplesPromise = getTriples(dcid);
+
+  Promise.all([provPromise, triplesPromise]).then(([provs, triples]) => {
```
shifucun, line 945: since can not may window.onload async, here did a bit more change to use Promise.all instead of two sync block call. Performance should be improved.

static/webpack.config.js
```diff
@@ -35,7 +35,11 @@ const config = {
       __dirname + "/js/tools/timeline.ts",
       __dirname + "/css/timeline.scss",
     ],
-    kg: [__dirname + "/js/browser/kg.js", __dirname + "/css/kg.scss"],
+    kg: [
+      "babel-polyfill",
```
shifucun, line 39: babel-polyfill is needed to make async/await work with webpack

shifucun:
> > I'm finding it really hard to do review this change, especially since there is also a move / change in where/how we are handling the triples request. I'm weary of adding an additional proxy for these calls in flask -- have you noticed any slowdowns?
> > Also, as discussed offline, this isn't blocking the container change.
> > Since our testing coverage for the browser is poor, could we split out the code deletion change into a separate PR (it will also help with rollbacks and reduce the diff)?
> 
> I will split the PR into small ones. Sent out https://github.com/datacommonsorg/website/pull/686/files first.

Synced and revert the client side removal, so this PR only does the migration. Can delete all the prod staging code in client in a later PR

static/js/browser/kg.js
beets: I suppose we're trading 2 sequential calls direct to the mixer with parallel calls to flask (which is now having to decode and re-encode large json blobs which was my concern).

beets (approved):
thanks for splitting this out!

static/js/browser/kg.js
shifucun: Yes, these are raw response which are large. Ideally the data should be processed in the server as we do for place page.

PR #687, opened 2020-11-24
Add two flask endpoints to get triples and popbos cache

shifucun:
This is to prepare migrating javascript api call from client to server side.

### DIFF
--- a/server/services/datacommons.py
+++ b/server/services/datacommons.py
@@ -23,7 +23,6 @@ import urllib
 
 import requests
 from werkzeug.utils import import_string
-from google.cloud import secretmanager
 
 if os.environ.get('FLASK_ENV') == 'test':
     cfg = import_string('configmodule.TestConfig')()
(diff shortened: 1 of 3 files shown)


server/services/datacommons.py
```diff
@@ -193,8 +192,13 @@ def get_property_values(dcids,
     return results
 
 
-def get_triples(dcids, limit=_MAX_LIMIT):
-    # Generate the GetTriple query and send the request.
+def get_triples(dcids, limit=0):
```
shifucun, line 195: This API is used by pv tree tool, hence adding a new one "get_triples_raw"

beets: We should update the comment so we can tell these apart in the future. Perhaps this should be renamed instead, while the other should remain as get_triples.

beets (commented):
We can add the kg.js / util.js update here too, since it'll be good to review the api update / usage together.

shifucun:
Sg. made the change in the original PR. Closing this one.

PR #695, opened 2020-11-30
Webdriver tests in parallel

edumorlom:
1. Adds pytest-parallel to requirements.txt. It's pytest + parallel plugin.
2. Updates Dockerfile to install Selenium Grid.
3. Adds -t flag in run_tests.sh to run tests in parallel.
4. Adds check to make sure $SELENIUM_SERVER is specified.
5. Increases TIMEOUT to 60 seconds.
6. Updates all documentation and README.md.

Note, PYTEST_PARALLEL is currently an OS variable. I will research to see if we can pass it in as an argument to Pytest.
-t will start selenium grid. Otherwise, run normal test (without parallel and without selenium grid).

### DIFF
--- a/webdriver-chrome/README.md
+++ b/webdriver-chrome/README.md
@@ -2,7 +2,7 @@
 
 ## Description
 
-This is a Docker image based on python:3.7-slim, but comes with some other tools:
+This is a Docker image based on python:3.7, but comes with some other tools:
     1. Google Chrome: the browser which is used to run the tests.
     2. ChromeDriver: used to send commands to Google Chrome.
     3. Java JDK: used to run the Selenium Server, which comes as a JAR file.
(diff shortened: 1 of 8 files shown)


webdriver-chrome/README.md
```diff
@@ -2,7 +2,11 @@
 
 ## Description
 
-This is a Docker image based on python:3.7-slim, but with chrome and chromedriver preinstalled which can help using Selenium to do automation tests.
+This is a Docker image based on python:3.7-slim, but comes with some other tools:
```
shifucun, line 5: nit: based on python:3.7

shifucun (approved):
Very nice and thorough PR, well done!!

webdriver-chrome/README.md
edumorlom: fixed.

PR #697, opened 2020-12-04
Create a message json file for stat var labels

beets:
Converted stats_var_label into a json file for TC extraction. A follow on will be to use these messages instead of the static array.

### DIFF
--- a/static/js/i18n/stats_var_labels.json
+++ b/static/js/i18n/stats_var_labels.json
@@ -1,0 +1,938 @@
+{
+  "Amount_Consumption_Alcohol_15OrMoreYears_AsFractionOf_Count_Person_15OrMoreYears": {
+    "defaultMessage": "Pure Alcohol Consumption",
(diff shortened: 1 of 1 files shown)


tjann (commented):
Did we miss e.g. https://github.com/datacommonsorg/website/blob/master/static/js/shared/stats_var_labels.ts#L23 "Median_Age_Person" ?

Should we do a grep of sorts to make sure nothing got missed?

beets:
Thanks for catching that - good call. I did a count and that was the only one I missed.

tjann:
> Thanks for catching that - good call. I did a count and that was the only one I missed.

Awesome. Love it when I get lucky 😆

tjann (approved):

shifucun (approved):
Any plan to do the stat var title file?

beets:
yea, i'll do that after wrapping up place pages since it's only used in ranking pages

PR #699, opened 2020-12-05
[Scatter2] APIs, provenance, shorter URL

intrepiditee:
1. Use `/stat/collection` for getting date and `/place/stat-vars/union` for getting available statvars
2. Show provenance #702 
3. Create shorter URL hash
4. Clear hash after resetting #704


static/js/tools/scatter2/plot_options.tsx
```diff
@@ -154,6 +201,51 @@ function PlotOptions(): JSX.Element {
   );
 }
 
+function getCurrentYear(): number {
+  return new Date().getFullYear();
+}
+
+function getNToM(n: number, m: number, descending = false): Array<number> {
```
beets, line 208: please add brief descriptions for these helpers

static/js/tools/scatter2/statvar.tsx
```diff
@@ -199,40 +181,88 @@ function StatVarChooser(): JSX.Element {
...
+  x: AxisWrapper,
+  y: AxisWrapper,
+  statVars: Set<string>
+) {
+  let message = "";
+  const statVarX = _.findKey(x.value.statVar);
+  const statVarY = _.findKey(y.value.statVar);
+  if (statVarX && !statVars.has(statVarX)) {
+    x.unsetStatVar();
+    message += `Sorry, no data available for ${statVarX}`;
```
beets, line 255: Can we provide an action to the user here - try another stat var or change the constraints.

beets (commented):
Mostly looks good! Just a few comments.

static/js/tools/scatter2/statvar.tsx
intrepiditee: fixed

static/js/tools/scatter2/plot_options.tsx
intrepiditee: fixed

server/routes/api/place.py
```diff
@@ -162,6 +160,40 @@ def statsvars(dcid):
...
+    """Get all the statistical variable dcids for some places.
+
+    Args:
+        dcids: Place DCIDs separated by "^" as a single string.
+
+    Returns:
+        List of unique statistical variable dcids each as a string.
+    """
+    dcids = dcids.split("^")
+    # The two indexings are due to how protobuf fields are converted to json
```
beets, line 174: That's unfortunate... we should hide these complexities from the user. We should be able to remove that outer statVars. Please add a todo to update this.

beets (approved):
thanks for the fixes! we should hide the implementation detail to avoid the double indexing of the union api, but that's a mixer fix. i'll merge this in so we can test on staging.

PR #705, opened 2020-12-09
Extract messages from the flask base template

beets:
Also adds simple code to test using es translations (there are a few test strings in es/messages.po)


server/l10n/en/LC_MESSAGES/messages.po
```diff
@@ -0,0 +1,127 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+#. The label for a collection of exploration tools.
+#: server/templates/base.html:69 server/templates/base.html:120
+msgid "Explore"
+msgstr ""
+
```
hanlu09205, line 19: For my education, is msgstr used for translator to get a better idea about the string?

server/l10n/en/LC_MESSAGES/messages.po
```diff
@@ -0,0 +1,127 @@
...
+#. The name of a tool to browse statistics about a place.
+#: server/templates/base.html:72 server/templates/base.html:122
```
hanlu09205, line 21: If the template line number change, this would also change, right?

Seems quite flaky to me or do we care the original line number?

server/l10n/en/LC_MESSAGES/messages.po
beets: technically, the msgstr should contain the string, whereas the id is a key. the way jinja-babel is set up, there isn't a way to specify the key, and the id is used in the default language.

my understanding of the file format is more like this, with separate id and str: https://github.com/datacommonsorg/website/blob/i18n/server/l10n/chart_titles.pot

server/l10n/en/LC_MESSAGES/messages.po
beets: yea, it's flaky. there isn't an option to skip the line number though, and it seems useful to have the originating file (the only option is to drop that line altogether)

hanlu09205: Got it. SG.

hanlu09205 (approved):

hanlu09205 (approved):

beets:
Thanks for the review!

PR #706, opened 2020-12-09
api_i18n_name and corresponding functions, which we later could still…

hanlu09205:
Add api_i18n_name & get_i18n_name for getting 18n names.

---
This change is [Reviewable](https://reviewable.io/reviews/datacommonsorg/website/706)

### DIFF
--- a/server/routes/api/place.py
+++ b/server/routes/api/place.py
@@ -130,6 +130,67 @@ def api_name():
     return Response(json.dumps(result), 200, mimetype='application/json')
 
 
+@cache.memoize(timeout=3600 * 24)  # Cache for one day.
+def cached_i18n_name(dcids, locale):
+    """Returns localination names for set of dcids.
+
+    Args:
+        dcids: ^ separated string of dcids. It must be a single string for the cache.
+        locale: the desired localization language code.
+
+    Returns:
+        A dictionary of place international names, keyed by dcid.
+    """
+    dcids = dcids.split('^')
+    response = fetch_data('/node/property-values', {
+        'dcids': dcids,
+        'property': 'nameWithLanguage',
+        'direction': 'out'
+    },
+                          compress=False,
+                          post=True)
+    result = {}
+    for dcid in dcids:
+        values = response[dcid].get('out')
+        name_in_en = ''
+        result[dcid] = ''
+        for entry in values:
+            if entry['value'].endswith(locale):
+                i18nname_end_index = len(entry['value']) - len(locale) - 1
+                result[dcid] = entry['value'][:i18nname_end_index]
+            else:
+                if entry['value'].endswith('en'):
+                    name_in_en = entry['value'][:(len(entry['value']) -
+                                                  len('en') - 1)]
+        # With there isn't a name in required language code, falls back to 
+        # English name if exists.
+        if not result[dcid]:
+            result[dcid] = name_in_en
+    return result
+
+
+def get_i18n_name(dcids, locale):
+    """"Returns localization names for set of dcids.
+
+    Args:
+        dcids: A list of place dcids.
+        locale: the desired localization language code.
+
+    Returns:
+        A dictionary of  place names, keyed by dcid.
+    """
+    return cached_i18n_name('^'.join((sorted(dcids))), locale)
+
+
+@bp.route('/i18nname')
+def api_i18n_name():
+    """Get place i18n names."""
+    dcids = request.args.getlist('dcid')
+    locale = request.args.get('hl', default="en")
+    result = get_i18n_name(dcids, locale)
+    return Response(json.dumps(result), 200, mimetype='application/json')
+
+
 @bp.route('/statsvars/<path:dcid>')
 @cache.memoize(timeout=3600 * 24)  # Cache for one day.
 def statsvars_route(dcid):
--- a/server/routes/api/place.py
+++ b/server/routes/api/place.py
@@ -151,9 +151,9 @@ def cached_i18n_name(dcids, locale):
                           compress=False,
                           post=True)
     result = {}
-    # When there is no locale, fall back to default en. 
-    # When there is no exact match of locale, and locale can be broken into parts: language - region, fall back to use language part. 
-    # If there still isn't a match, fall back to use en. 
+    # When there is no locale, fall back to default en.
+    # When there is no exact match of locale, and locale can be broken into parts: language - region, fall back to use language part.
+    # If there still isn't a match, fall back to use en.
     if not locale:
         locale = ENGLISH_LANG
     fallback_lang = locale.split('-')[0] if locale.find('-') != -1 else ''
--- a/server/routes/api/place.py
+++ b/server/routes/api/place.py
@@ -150,42 +150,31 @@ def cached_i18n_name(dcids, locale):
                           compress=False,
                           post=True)
     result = {}
-    if not locale:
-        locale = 'en'
-    fallback_locale = locale.split('-')[0] if locale.find('-') != -1 else ''
+    # When there is no locale, fall back to default en. 
+    # When there is no exact match of locale, and locale can be broken into parts: language - region, fall back to use language par. 
+    # If there still isn't a match, fall back to use en. 
     english_local = 'en'
-    second_choice = ''
-    last_choice = ''
+    if not locale:
+        locale = english_local
+    fallback_locale = locale.split('-')[0] if locale.find('-') != -1 else ''
     for dcid in dcids:
         values = response[dcid].get('out')
         name_in_en = ''
         result[dcid] = ''
-        print('locale = ' + locale)
+        fallback_choice = ''
+        english_choice = ''
         for entry in values:
-            print(entry)
             if (has_locale_name(entry, locale)):
                 result[dcid] = extract_locale_name(entry, locale)
+                # Find the exact matches, no need to fall back.
                 break
-            # if entry['value'].endswith('@' + locale.lower()):
-            #     i18nname_end_index = len(entry['value']) - len(locale) - 1
-            #     result[dcid] = entry['value'][:i18nname_end_index]
-            # if result[dcid]:
-            # Already found best match, break out of the loop
-            # break
             else:
                 if (has_locale_name(entry, fallback_locale)):
-                    second_choice = extract_locale_name(entry, fallback_locale)
+                    fallback_choice = extract_locale_name(entry, fallback_locale)
                 if (has_locale_name(entry, english_local)):
-                    last_choice = extract_locale_name(entry, english_local)
-                # if entry['value'].endswith('@' + fallbacl_locale):
-                # if entry['value'].endswith('@en'):
-                #     name_in_en = entry['value'][:(len(entry['value']) -
-                #                                   len('en') - 1)]
-        # When there isn't a name in required language code, falls back to
-        # English name if exists.
+                    english_choice = extract_locale_name(entry, english_local)
         if not result[dcid]:
-            print('if not result[dcid]:')
-            result[dcid] = second_choice if second_choice else last_choice
+            result[dcid] = fallback_choice if fallback_choice else english_choice
     return result
 
 
@@ -197,9 +186,8 @@ def extract_locale_name(entry, locale):
     if entry['value'].endswith('@' + locale.lower()):
         locale_index = len(entry['value']) - len(locale) - 1
         return entry['value'][:locale_index]
-    else:
+    else :
         return ''
-
 
 def get_i18n_name(dcids, locale):
     """"Returns localization names for set of dcids.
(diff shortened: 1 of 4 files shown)


server/routes/api/place.py
```diff
@@ -130,6 +130,67 @@ def api_name():
     return Response(json.dumps(result), 200, mimetype='application/json')
 
 
+@cache.memoize(timeout=3600 * 24)  # Cache for one day.
+def cached_i18n_name(dcids, locale):
+    """Returns localination names for set of dcids.
```
beets, line 135: nit: localization

server/routes/api/place.py
```diff
@@ -130,6 +130,67 @@ def api_name():
...
+        for entry in values:
+            if entry['value'].endswith(locale):
+                i18nname_end_index = len(entry['value']) - len(locale) - 1
+                result[dcid] = entry['value'][:i18nname_end_index]
+            else:
+                if entry['value'].endswith('en'):
+                    name_in_en = entry['value'][:(len(entry['value']) -
+                                                  len('en') - 1)]
+        # With there isn't a name in required language code, falls back to 
+        # English name if exists.
```
beets, line 166: are we always guaranteed a name in english? i.e. if it's blank, it means we have no name at all for the node.

server/routes/api/place.py
```diff
@@ -130,6 +130,67 @@ def api_name():
...
+    },
+                          compress=False,
+                          post=True)
+    result = {}
+    for dcid in dcids:
+        values = response[dcid].get('out')
+        name_in_en = ''
+        result[dcid] = ''
+        for entry in values:
+            if entry['value'].endswith(locale):
```
beets, line 158: should we test for "@" as well? e.g. endswith("@en")
it would be good to also lowercase the locale

also, not sure if we have expanded locale info in the graph (e.g. pt-br). we might get those expanded codes though, so would be good to take the appropriate part of the locale.

that said, we can simplify this function and only accept simplified locale info, and do the user locale parsing elsewhere.

server/routes/api/place.py
```diff
@@ -130,6 +130,67 @@ def api_name():
...
+        dcids: A list of place dcids.
+        locale: the desired localization language code.
+
+    Returns:
+        A dictionary of  place names, keyed by dcid.
+    """
+    return cached_i18n_name('^'.join((sorted(dcids))), locale)
+
+
+@bp.route('/i18nname')
```
beets, line 185: how about "/name/i18n" (since we already have /name)

hanlu09205: Done.

server/routes/api/place.py
hanlu09205: Done.

server/routes/api/place.py
hanlu09205: There are 2408932 distinct subject_id that have i18n name, but only 1051262 have i18n name in english. 

Do you think that if there is no name in the required language, we shall fall back to english?
I think that it would be weird to show blank, if it does have english name. 
In that case, if it's blank, then there isn't a name in the required language code, neither in English.

server/routes/api/place.py
hanlu09205: Done lowercase. 

Regarding testing with endswith("@en"), shall the url be like hl=en instead of hl=@en? 
If yes, don't think endswith("@en") is needed then. 

We have pt-br as well as pt, and we also have both zh-cn & zh. 
So don't think we need to tweak the parameter here except lower case?

beets: my thought with adding the "@" sign in the tests are to capture cases such as:
hl=ru
i18nname=foobar@ru | foobar@ru-RU | foobar@tt-ru

(we don't have those languages, but there might be a collision between locale and language?)

btw, are we storing locale as uppercase or lowercase? we might need to lowercase the returned value too.

(yes, it should be hl=en not hl=@en)

beets (commented):
thanks for the updates!

server/routes/api/place.py
hanlu09205: As far as I could tell, all the name is extracted from @en name. So if a place has a name, then it will also have an @en name. 
In that case, I don't think that we need to append the english name to the list of i18n names, since there will be an i18n english name. 
Just to confirm, there isn't another api call here. If we didn't find the required i18n name, we fall back to english name, which is already fetched in the same api call.

server/routes/api/place.py
hanlu09205: Ic. Tests added. 
We store all the locale as lower case. So convert the parameter to lower case shall be good.

hanlu09205: I added both fallback case, from it-eu fall back to it then fall back to en. 
Also updated the test to capture @en-ru when we want @ru, nice catch!
(I have to use ends with but not finds otherwise @en-ru would be return when we want @en.

beets:
___
*[server/routes/api/place.py, line 154 at r4](https://reviewable.io/reviews/datacommonsorg/website/706#-MOHz2km-Tw7u_h6uvPz:-MOHz2km-Tw7u_h6uvQ-:bgp5gpt) ([raw file](https://github.com/datacommonsorg/website/blob/ceb57b4dfcc9a48d138902097c57d8b05b0288b6/server/routes/api/place.py#L154)):*
> ```Python
>     result = {}
>     # When there is no locale, fall back to default en. 
>     # When there is no exact match of locale, and locale can be broken into parts: language - region, fall back to use language par. 
> ```

nit: language part

server/routes/api/place.py
```diff
@@ -130,6 +130,86 @@ def api_name():
...
+    response = fetch_data('/node/property-values', {
+        'dcids': dcids,
+        'property': 'nameWithLanguage',
+        'direction': 'out'
+    },
+                          compress=False,
+                          post=True)
+    result = {}
+    # When there is no locale, fall back to default en. 
+    # When there is no exact match of locale, and locale can be broken into parts: language - region, fall back to use language par.
```
beets, line 154:
```suggestion
    # When there is no exact match of locale, and locale can be broken into parts: language - region, fall back to use language part. 
```

server/routes/api/place.py
```diff
@@ -130,6 +130,86 @@ def api_name():
...
+    fallback_locale = locale.split('-')[0] if locale.find('-') != -1 else ''
+    for dcid in dcids:
+        values = response[dcid].get('out')
+        result[dcid] = ''
+        fallback_choice = ''
+        english_choice = ''
+        for entry in values:
+            if (has_locale_name(entry, locale)):
+                result[dcid] = extract_locale_name(entry, locale)
+                # Find the exact matches, no need to fall back.
```
beets, line 168:
```suggestion
                # Found the exact match, no need to fall back.
```

beets (changes requested):
Thanks for the updates! Just a few small nits...

> *[Reviewable](https://reviewable.io/reviews/datacommonsorg/website/706#-:-MOI16DSEg8P3hZcC5p9:b1diwoh)* status: 0 of 3 files reviewed, 5 unresolved discussions (waiting on @beets, @hanlu09205, and @shifucun)

___
*[README.md, line 146 at r4](https://reviewable.io/reviews/datacommonsorg/website/706#-MOHyYY81AXG70VnEAJ2:-MOHyYY81AXG70VnEAJ3:b-wswy5v) ([raw file](https://github.com/datacommonsorg/website/blob/ceb57b4dfcc9a48d138902097c57d8b05b0288b6/README.md#L146)):*
> ```Markdown
> 
> ```bash
> ./run_server.sh
> ```

oh whoops! thanks for the fix
___
*[server/routes/api/place.py, line 158 at r1](https://reviewable.io/reviews/datacommonsorg/website/706#-MOE3cwEXd3MBdt9-pGC-r1-158:-MOI0WmyBU3CycLB2PFC:b9y27z0) ([raw file](https://github.com/datacommonsorg/website/blob/6cc64b1ac9f8473260ed21248fa13f67f97ae50b/server/routes/api/place.py#L158)):*
<details><summary><i>Previously, hanlu09205 wrote…</i></summary><blockquote>

I added both fallback case, from it-eu fall back to it then fall back to en. 
Also updated the test to capture @en-ru when we want @ru, nice catch!
(I have to use ends with but not finds otherwise @en-ru would be return when we want @en. 
</blockquote></details>

great, thanks!
___
*[server/routes/api/place.py, line 166 at r1](https://reviewable.io/reviews/datacommonsorg/website/706#-MOE3cwEXd3MBdt9-pGC-r1-166:-MOI0c0K1pMU_Pc91t9r:bkrfy0s) ([raw file](https://github.com/datacommonsorg/website/blob/6cc64b1ac9f8473260ed21248fa13f67f97ae50b/server/routes/api/place.py#L166)):*
<details><summary><i>Previously, hanlu09205 wrote…</i></summary><blockquote>

As far as I could tell, all the name is extracted from @en name. So if a place has a name, then it will also have an @en name. 
In that case, I don't think that we need to append the english name to the list of i18n names, since there will be an i18n english name. 
Just to confirm, there isn't another api call here. If we didn't find the required i18n name, we fall back to english name, which is already fetched in the same api call. 
</blockquote></details>

Ok, gotcha!
___
*[server/routes/api/place.py, line 156 at r4](https://reviewable.io/reviews/datacommonsorg/website/706#-MOHzsDs7ff-k8wAvCSZ:-MOHzsDs7ff-k8wAvCS_:b-19ehr8) ([raw file](https://github.com/datacommonsorg/website/blob/ceb57b4dfcc9a48d138902097c57d8b05b0288b6/server/routes/api/place.py#L156)):*
> ```Python
>     # When there is no exact match of locale, and locale can be broken into parts: language - region, fall back to use language par. 
>     # If there still isn't a match, fall back to use en. 
>     english_local = 'en'
> ```

Should we make 'en' a const?
___
*[server/routes/api/place.py, line 159 at r4](https://reviewable.io/reviews/datacommonsorg/website/706#-MOI-7ZZAx7dM11DQTYL:-MOI-7ZZAx7dM11DQTYM:b-5cs1t9) ([raw file](https://github.com/datacommonsorg/website/blob/ceb57b4dfcc9a48d138902097c57d8b05b0288b6/server/routes/api/place.py#L159)):*
> ```Python
>     if not locale:
>         locale = english_local
>     fallback_locale = locale.split('-')[0] if locale.find('-') != -1 else ''
> ```

This is technically the language (and after the "-" is the region)

hanlu09205 (commented):
> Reviewed 1 of 3 files at r1, 2 of 2 files at r3.
> *[Reviewable](https://reviewable.io/reviews/datacommonsorg/website/706)* status: 1 of 3 files reviewed, 5 unresolved discussions (waiting on @beets, @hanlu09205, and @shifucun)

___
*[server/routes/api/place.py, line 156 at r4](https://reviewable.io/reviews/datacommonsorg/website/706#-MOHzsDs7ff-k8wAvCSZ:-MOIEqHAAWuYO0pWlAcN:b-896fix) ([raw file](https://github.com/datacommonsorg/website/blob/ceb57b4dfcc9a48d138902097c57d8b05b0288b6/server/routes/api/place.py#L156)):*
<details><summary><i>Previously, beets wrote…</i></summary><blockquote>

Should we make 'en' a const?
</blockquote></details>

Done.
___
*[server/routes/api/place.py, line 159 at r4](https://reviewable.io/reviews/datacommonsorg/website/706#-MOI-7ZZAx7dM11DQTYL:-MOIIHgTDBzS7UFpNqLv:btv3g0w) ([raw file](https://github.com/datacommonsorg/website/blob/ceb57b4dfcc9a48d138902097c57d8b05b0288b6/server/routes/api/place.py#L159)):*
<details><summary><i>Previously, beets wrote…</i></summary><blockquote>

This is technically the language (and after the "-" is the region)
</blockquote></details>

Renamed fallback_locale to fallback_lang.
___
*[server/routes/api/place.py, line 168 at r4](https://reviewable.io/reviews/datacommonsorg/website/706#-MOE3cwEXd3MBdt9-pGC-r4-168:-MOIIR056hRpw_8Lglzm:b-896fix) ([raw file](https://github.com/datacommonsorg/website/blob/ceb57b4dfcc9a48d138902097c57d8b05b0288b6/server/routes/api/place.py#L168)):*
<details><summary><i>Previously, beets wrote…</i></summary><blockquote>

```suggestion
                # Found the exact match, no need to fall back.
```
</blockquote></details>

Done.

beets (approved):
Thanks! Looks like the linter check is failing... you can run ./run_test.py -f to fix the errors

hanlu09205:
> Thanks! Looks like the linter check is failing... you can run ./run_test.py -f to fix the errors

Done. I am still seeing codacy static code analysis not getting a check, and this message makes me wonder whether it's turned on for this branch. https://screenshot.googleplex.com/AKFW5hhUAx3wmkg
Will merge for now, and fix this later if needed. 

Thanks for the review and suggesting reviewable, much better. :)

PR #707, opened 2020-12-10
[Scatter2] Fix formatDate and isDateChosen

intrepiditee:
1. Fix `formatDate` not adding extra zeros before single digit months and days
2. Fix `isDateChosen` not checking if month or day is not set


intrepiditee:
@beets @shifucun Could you review? Thanks

beets (approved):

PR #712, opened 2020-12-10
Extract strings from place_landing.html

beets:
- Updated extract/update commands to use longer line widths
- Sorted the strings (and caught a few errors that way)
- Added placeholder es strings for testing

Note that it was tricky to add the intro paragraph (PLACE_EXPLORER_INTRO). I'm testing this out by using a key instead, and adding the message in the en bundle. It makes string updates more cumbersome, but it doesn't make sense to use that entire paragraph as a key. This helps get the strings out for translation, at least.


beets:
Note, it might be easier to track the diffs by looking at each commit. The first only updates the sort order, the next commit adds the landing page strings.

tjann (approved):

PR #716, opened 2020-12-11
Update GKE deployment steps: this is a working version on a brand new GCP project

shifucun:
- Set the necessary and minimal IAM roles for service account.
- Add resource project to config.yaml, so it will be easy to migrate later.
- Separate into multiple small script for incremental update.


deployment/config.yaml
```diff
@@ -14,15 +14,15 @@
 
 project:
   dev: datcom-mixer-staging # datacommons-website-dev
-  staging: datacommons-website-staging
+  staging: dc-test-1-298300
```
beets, line 17: Should we re-make a new project with the name we'd like? Or add a comment to update it.

deployment/config.yaml
```diff
@@ -109,3 +114,12 @@ cpu:
       dev: 100m
       staging: 500m
       prod: 500m
+
+bind_project:
+  bigquery: google.com:datcom-store-dev
+  bigtable: google.com:datcom-store-dev
+  branch_cache: google.com:datcom-store-dev
+  website_gcs: datcom-browser-staging # TODO(shifucun): use `gsutil acl ...`
```
beets, line 122: Should we also add a TODO to migrate that content to the new project?

beets (approved):

deployment/gke/first_time_setup.sh
```diff
@@ -0,0 +1,66 @@
+#!/bin/bash
```
pradh, line 1: Is this same as one_time_setup.sh?

deployment/gke/config_project.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
```
pradh, line 1: If you're not around, I cannot imagine navigating this to run the right thing :)

Should we have a deployment/gke/README.md to give an overview of what all these scripts are and the flow in general?  (follow-up PR after you figure all this out is fine)

pradh (approved):
I don't understand most of it.  But I know one of us do :)

deployment/config.yaml
shifucun: that's the plan, added TODO

deployment/config.yaml
shifucun: done

deployment/gke/first_time_setup.sh
shifucun: Yes, I put each sub-tasks in a separate script since need to re-run some of them independently (like creating new cluster)

deployment/gke/config_project.sh
shifucun: Yeah, sure. Also I can host a tech-talk for this, and do a live deployment?

shifucun:
thanks for review!

PR #717, opened 2020-12-11
website app /healthz endpoints used for kubernetes health check

shifucun:
GKE use readiness check to to pod management and load balancing:

https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/

### DIFF
--- a/server/main.py
+++ b/server/main.py
@@ -74,6 +74,11 @@ def search():
     return flask.render_template('search.html')
 
 
+@app.route('/healthz')
+def healthz():
+    return "very healthy"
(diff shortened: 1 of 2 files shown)


server/main.py
```diff
@@ -74,6 +74,11 @@ def search():
     return flask.render_template('search.html')
 
 
+@app.route('/healthz')
+def healthz():
+    return "very healthy"
```
beets, line 79: :D

beets (approved):

PR #719, opened 2020-12-12
GKE does not support startupProbe yet, so only keep readiness probe

shifucun:
See discussion: https://stackoverflow.com/a/61936155.

Basically without startupProbe, the liveness probe can timeout if server starts slow and restart it, making it an infinite loop.


pradh (approved):

shifucun:
thanks!

PR #722, opened 2020-12-15
Trim values for "nameWithLanguage" predicate on browser

pradh:
I did not find an existing unit-test setup, so tested manually.

Since "nameWithLanguage" isn't on prod yet, tested this change with a different predicate:

```outArcs = trimArcsForPredicate(dcid, "geoOverlaps", 2, outArcs);```

With that change, the page https://datacommons.org/browser/zip/95008 shows up as:

Screen Shot 2020-12-14 at 6 33 47 PM

### CODE REVIEW

**Review by shifucun — APPROVED** [review]
thanks for change!

File: static/js/browser/kg.js [hunk 1]
```diff
@@ -457,6 +457,40 @@ function getTriples(dcid) {
...
+  outArcs = outArcs.filter(
+      (p) => !(p["predicate"] == predicate && ++numSeen > maxValues));
+
+  // If trimmed, add an indicator.
+  if (nArcs > maxValues) {
+    const extra = (nArcs - maxValues).toString();
+    outArcs.push({
+      "subjectId": dcid,
+      "predicate": predicate,
+      "objectValue": "(... " + extra + " more ...)",
```
hanlu09205, line 488: Thanks for making this change!

Just wonder whether we need to worry about translating "extra more" one day?

beets: We would, and there are a few other similar strings embedded in JS throughout.

beets (approved):
Would be nice to have an option to expand and see the rest of the values. Maybe add a TODO to add that option?

Thanks for the fix!

pradh:
> Would be nice to have an option to expand and see the rest of the values. Maybe add a TODO to add that option?

Added TODO.

pradh:
Thanks for the review. Merging...

PR #723, opened 2020-12-16
Fix the steps for create standalone cluster

shifucun:
Need to make sure project and cluster are correct.
Need to make sure the local robot keys are correct.
Need to update the yaml files before a new deployment.


beets (approved):

shifucun:
thx for review!

PR #727, opened 2020-12-16
[MAPS_API_KEY] Use distinct keys from GCP projects corresponding to the  environment

shifucun:
Set the maps api key as environment variable instead of hard coded.

### DIFF
--- a/deployment/website.yaml.tmpl
+++ b/deployment/website.yaml.tmpl
@@ -71,5 +71,5 @@ spec:
               value: <FLASK_ENV>
             - name: GCS_BUCKET
               value: <GCS_BUCKET>
-            - name: MAPS_API_KEY
-              value: <MAPS_API_KEY>
+            - name: PROJECT
+              value: <PROJECT>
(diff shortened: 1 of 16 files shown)


beets (commented):
Should we also use secret manager for staging and prod? I don't think we should put keys in github (I've received warnings about other keys I've owned in the past).

shifucun:
> Should we also use secret manager for staging and prod? I don't think we should put keys in github (I've received warnings about other keys I've owned in the past).

Put all api key in secret manager now

deployment/website.yaml.tmpl
```diff
@@ -70,4 +70,6 @@ spec:
             - name: FLASK_ENV
               value: <FLASK_ENV>
             - name: GCS_BUCKET
-              value: <GCS_BUCKET>
\ No newline at end of file
+              value: <GCS_BUCKET>
+            - name: PROJECT
```
beets, line 74: project is very generic -- do you mean GCP project here?
Maybe we can add comments to the yaml.tmpl files?

beets (commented):
thanks for moving the keys!

deployment/website.yaml.tmpl
shifucun: renamed to SECRET_PROJECT. I like this suggestion as I once have to remember what project this is.

beets: 👍

beets (approved):

PR #731, opened 2020-12-17
Fetchchild

hanlu09205:
api place ranking
 http://127.0.0.1:8080/api/place/ranking/geoId/06?hl=zh 
https://screenshot.googleplex.com/6i8e7NUYfzNxVnb

fetch_child
http://127.0.0.1:8080/api/place/child/geoId/06
This would take a long long time to load fetch children name are all fetched before filtering. 
As far as I could tell, only placelist uses this. If we don't care about placelist i18n, I am thinking revert this change. 
Or filter children first, before getting their i18n name.

---
This change is [Reviewable](https://reviewable.io/reviews/datacommonsorg/website/731)

### DIFF
--- a/server/routes/api/place.py
+++ b/server/routes/api/place.py
@@ -17,6 +17,7 @@ import json
 import random
 import re
 import time
+import logging
 
 from flask import Blueprint, request, Response, url_for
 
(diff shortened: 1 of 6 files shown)


beets:
placelist doesn't have to be i18n - it's essentially a sitemap for the search index.
if you'd like to keep the code, filtering first sgtm

beets (approved):
same comments apply from the previous PR regarding ensuring the cache includes the locale in the key

hanlu09205:
Updated per comments and verified http://127.0.0.1:8080/api/place/ranking/geoId/06?hl=zh
submitting...

PR #732, opened 2020-12-18
Upgrades to React 17 and Webpack 5

edumorlom:
**NOTE: This PR doesn't introduce any optimizations. First, we need to make sure we can get bare Webpack 5 + React 17 working without issues.**

**NOTE 2: There are several optimizations that can be done to improve speed. Webpack 5 introduces Persistent Cache, which dramatically improves building time but _I have not enabled it._**

I have tested this myself and didn't find any issues, but please do some more testing! You guys know more than I do.

1. Upgrades to React 17.
2. Upgrades to Webpack 5.
3. Upgrades Webpack loaders and plugins.
3. Switches from `enzyme-adapter-react-16 `to `@wojtekmaj/enzyme-adapter-react-17`.
4. Switches from `webpack-shell-plugin `to `webpack-shell-plugin-next`.
     - `webpack-shell-plugin` was last updated 4 YEARS ago and doesn't support Webpack 5.
5. Switches from `webpack-fix-style-only-entries` to `webpack-remove-empty-scripts`.
     - `webpack-fix-style-only-entries` doesn't support Webpack 5.
6. `-p` flag is no longer supported, switch to `--mode=production` instead. [See Here](https://webpack.js.org/guides/production/#cli-alternatives)
     - As per documentation, `-p` also enabled `optimization.minimize` ONLY when in production.
7. Switches from `__dirname + path` to `path.resolve(__dirname, path)`.
8. Minor dependency version bumps.
9. Fix linting errors to make sure tests pass.


shifucun:
@edumorlom I will close this PR as there have been substantial changes in the code base. We can come back to the upgrade later. Thanks for sending this change!

PR #744, opened 2021-01-06
Move build notifier to tools folder

shifucun:


beets (approved):
thanks! could you add a quick readme as well while you're there?

shifucun:
> thanks! could you add a quick readme as well while you're there?

Done

PR #755, opened 2021-01-06
Have config file for each gke environment and update cluster setup process

shifucun:
This is to prepare integrating with "Kustomize" which is also environment separated. This also makes it easy to maintain these configs.

Also delete the steps to download service account key json as the website would use Workload Identity to authenticate.

Note the folder and file layout is aligned with mixer repo now: https://github.com/datacommonsorg/mixer/tree/master/gke


gke/README.md
```diff
@@ -32,19 +50,13 @@ Each node is a configured to e2-highmem-4 machine type that has 4 vCPU and 32G m
 
 ## DNS setup
 
-- Make sure the managed SSL certificate is "ACTIVE" by checking ["Load balancing" in GCP](https://pantheon.corp.google.com/net-services/loadbalancing/advanced/sslCertificates/list?project=PROJECT_ID&sslCertificateTablesize=50). This can take minutes up to hours.
+- Make sure the managed SSL certificate is "ACTIVE" by checking ["Load balancing" in GCP](https://pantheon.corp.google.com/net-services/loadbalancing/advanced/sslCertificates/list?project=PROJECT_ID&sslCertificateTablesize=50). This can take minutes up to hours. Note the SSL need to be used in an Ingress to be "ACTIVE".
```
beets, line 53: Could you describe more what this means here? I'm not sure what using the SSL in an Ingress means...

beets (approved):
nice refactor! just one question for the readme..

gke/README.md
shifucun: Add some more text and a screenshot for this.

shifucun:
Thanks for review!

PR #762, opened 2021-01-09
Add Redis to k8s production

shifucun:


beets (approved):
Should we reconsider using redis in staging as well? Would be nice if staging closely mirrors the prod setup.

shifucun:
> Should we reconsider using redis in staging as well? Would be nice if staging closely mirrors the prod setup.

The redis cache persist after new deployment and needs to be invalidated. Might just use in-memory cache for staging?

PR #763, opened 2021-01-11
Skaffold file sync and local dev improvement

shifucun:
Explanation of the changes:

- Skaffold "File Sync" can avoid re-build/deploy docker containers (which can take 10s of seconds) so local "hot reload" can happen instantly. This feature does not support multi-staging docker file yet. So change the docker base image to "python-nodejs".
- "File Sync" can only work with debug Flask server, but not gunicorn, hence using a "if else" in the "CMD" command.
- "File Sync" for webpack build files relies on the existence of "server/dist" folder. Track it in Git but only with a dummy file (and excluding all built files)
- readiness probe does not work with "python main.py" command somehow, disabling it for local development.
- Disable mixer liveness probe as it will cause repeated restart during mixer branch cache reload


deploy/base/deployment.yaml
```diff
@@ -114,26 +109,21 @@ spec:
             exec:
               command: ["/bin/grpc_health_probe", "-addr=:12345"]
             periodSeconds: 5
-          livenessProbe:
-            exec:
-              command: ["/bin/grpc_health_probe", "-addr=:12345"]
-            periodSeconds: 5
-            initialDelaySeconds: 10
+            failureThreshold: 60
+          # livenessProbe:
```
beets, line 113: Add a note here on why it's commented out?

developer_guide.md
```diff
@@ -0,0 +1,155 @@
...
+[Mixer](https://github.com/datacommonsorg/mixer) is a submodule of this Git repo.
+The exact commit of the submodule is deployed together with the website so it may
+not be the same API version as `api.datacommons.org`. Make sure to update and
+track the mixer change for a new deployment:
+
+```bash
+git submodule foreach git pull origin master
+```
+
+## Prerequisit
```
beets, line 19:
```suggestion
## Prerequisites
```

developer_guide.md
```diff
@@ -0,0 +1,155 @@
...
+```
+
+### Update React test snapshots
+
+```bash
+cd static
+npm test testfilename -- -u
+```
+
+## Develop with local Kubernets (Recommended)
```
beets, line 85:
```suggestion
## Develop with local Kubernetes (Recommended)
```

beets (approved):
Thanks Bo.

Btw, should we sync the mixer submodule only in separate PR's (unless required for a feature being developed)? wdyt?

deploy/base/deployment.yaml
shifucun: Done

developer_guide.md
shifucun: Done

developer_guide.md
shifucun: Done

shifucun (commented):
thanks for review!

shifucun:
> Thanks Bo.
> 
> Btw, should we sync the mixer submodule only in separate PR's (unless required for a feature being developed)? wdyt?

I think that makes sense. We can follow that for our future release

PR #766, opened 2021-01-12
Ranking i18n

hanlu09205:
Verified with screenshot https://screenshot.googleplex.com/8oP5wkMEyCMWKVf.


static/js/chart/types.ts
```diff
@@ -89,6 +89,7 @@ export interface PageData {
   nearbyPlaces: string[];
   categories: { string: string };
   names: { string: string };
+  locale: string[];
```
beets, line 92: It doesn't look like this is used (and it shouldn't be necessary). Please revert this change.

beets (approved):
thanks! lgtm except for one extra locale prop.

static/js/chart/types.ts
hanlu09205: Thanks for the review. Done.

PR #778, opened 2021-01-21
[i18n] Bundle in remaining strings

beets:
Sorry this got quite big! Quite a few changes bundled here:

- Bundle in messages for stats_var_labels and stats_var_titles (for en only, while translations are underway). This is quite manual at the moment - each module needs to specify up front which message bundle is required for that app so we only add required translations to each package.
- Use server-specified locale on the client
- Update bundle for new server-side strings
- Submit compiled messages, since they add time to builds and don't change often
- Update tests / cleanup code.


beets:
Sample output:

image

hanlu09205 (approved):

PR #789, opened 2021-01-26
Create a cloudbuild for website autopush

shifucun:
This will be used by the pubsub trigger


script/autopush.sh
```diff
@@ -0,0 +1,38 @@
+#!/bin/bash
```
beets, line 1: Nit: can we rename this folder to 'scripts'? It will probably grow to hold more scripts in the future

shifucun: Done

beets (approved):
thx!

PR #781, opened 2021-01-22
[i18n] Propagate hl parameters

beets:
- append hl parameter to all flask links that use url_for 
- append hl parameter to relevant js links (added a new react functional component for this)

in both cases, only propagates hl parameter if it's not 'en'.

- also disables 'explore more' link for non-en languages, since it won't be translated

---
This change is [Reviewable](https://reviewable.io/reviews/datacommonsorg/website/781)

PR #790, opened 2021-01-26
[i18n] Propagate hl params for related places

beets:
Left out a set of links in #781


beets:
Also delete straggler package-lock.json at the root dir (should only be in /static)

PR #791, opened 2021-01-26
Move popobs cache unzip from server to client

shifucun:
The raw popobs cache for geoId/06 is 9.7Mb, which is not bad. Unzip it on flask server is very slow and leads to timeout. 

This way reduces the memcache object size, also user can leverage the browser cache which also use a smaller object.

Tested this on local instance, California page loads in about 15s, compared with timeout after 40s.


beets (approved):

beets:
Do we know of any other users of that API?

shifucun:
> Do we know of any other users of that API?

nope, this is the only one

pradh (approved):
Nice! Thanks for the fix!

PR #793, opened 2021-01-27
Add browser page loading and error hint

shifucun:


static/js/browser/kg.js
```diff
@@ -554,6 +554,10 @@ async function renderKGPage(
     .get(`/api/browser/popobs/${locId}`)
     .then((resp) => {
       return JSON.parse(util.unzip(resp.data));
+    })
+    .catch(() => {
+      const loadingElem = document.getElementById("page-loading");
+      loadingElem.innerText = "Error Fetch PopObs data";
```
beets, line 560:
```suggestion
      loadingElem.innerText = "Error fetching data about this node";
```

beets: or "Error fetching observations about this node"

beets: Add a comment here to also update the check on line 739, or use a variable instead of testing for the message (in case we change the message and forget to update the check later)

shifucun: Updated text and use an error variable

beets (approved):
thx!

shifucun:
thanks for review!

PR #794, opened 2021-01-27
[i18n] A few more hl parameters to propagate

beets:
* Propagate hl parameters to related places and choropleth click throughs
* Use hl parameter for geojson to get translated place names in choropleth
* Fix parsing of admin area

image


beets:
Also fix a bug in chart titles (used the wrong variable for place name):
image

beets:
friendly bump

shifucun (approved):

beets:
thx!

PR #800, opened 2021-01-28
Update minikube instructions

beets:


beets:
Thanks for your help today! It was working after the docker update was installed

developer_guide.md
```diff
@@ -111,10 +117,15 @@ Run the following command to get the javascript code recompiled when changed:
 
 Python code change will trigger a restart of the Flask server automatically.
 
-### Disable Mixer branch cache
+### Disabled Mixer branch cache
```
shifucun, line 120: thanks for the update! this section can be removed now. The branch cache is read from BT and does not impact startup time or mem usage

shifucun (approved):

developer_guide.md
beets: ok! removed

PR #805, opened 2021-02-02
Add lite dev mode so GCP auth is not required

shifucun:


developer_guide.md
```diff
@@ -143,6 +144,14 @@ Start the flask webserver locally at localhost:8080
 ./run_server.sh
 ```
 
+If you don't have DataCommons GCP permissions, can run
```
beets, line 147:
```suggestion
If you don't have DataCommons GCP permissions, run
```

beets, line 147: Should we move the flask section up (before prerequisites) and change it to "simple / lite" instead of not recommended?

beets (approved):
thanks for this bo!

should we also avoid updating the mixer submodule unless it's release related or required dependency change?

developer_guide.md
shifucun: Done

shifucun: Done

shifucun:
> thanks for this bo!
> 
> should we also avoid updating the mixer submodule unless it's release related or required dependency change?

oops, that's not intended. Reverted. Thanks for the catch!

PR #808, opened 2021-02-03
[i18n] use state names instead of state codes for i18n display names

beets:
This helps with most American places, and is in line with how Maps treats these place names.

image
image
image

Unchanged in en:
image


beets:
friendly bump

hanlu09205 (approved):

beets:
thanks for the review!

PR #809, opened 2021-02-04
[i18n] Remove extra caret from translations

beets:
The caret is now added in the template, so this caused double carets in some languages.


beets:
On staging:
image
image

On localhost:
image
image

shifucun (approved):

PR #815, opened 2021-02-11
Use provenance url instead domain for sources

shifucun:
Still shows domain in both place page and timeline page.
This bundles the mixer change that removes provenance domain as well. So client and api changes are within one commit.

image
image

### DIFF
--- a/server/routes/api/chart.py
+++ b/server/routes/api/chart.py
@@ -199,7 +199,7 @@ def process_choropleth_data(all_sv_data):
                     most_recent_date = sorted_dates[-1]
                     result[geo][sv] = {
                         'data': source_values,
-                        'provenanceDomain': source.get('provenanceDomain', '')
+                        'provenanceUrl': source.get('provenanceUrl', '')
                     }
     return result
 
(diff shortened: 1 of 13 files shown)


beets:
there could be multiple sources per chart in the timeline tool, so would be good to display only the domain there as well.
this pr would be a case where the mixer + website change should be in the same pr, so that they can be synced? but i'll approve the mixer pr, so that the mixer submodule in this repo can be brought up to date with the non-backwards compatible change.

pradh:
> there could be multiple sources per chart in the timeline tool, so would be good to display only the domain there as well.
> this pr would be a case where the mixer + website change should be in the same pr, so that they can be synced? but i'll approve the mixer pr, so that the mixer submodule in this repo can be brought up to date with the non-backwards compatible change.

nice catch!

shifucun:
> there could be multiple sources per chart in the timeline tool, so would be good to display only the domain there as well.
> this pr would be a case where the mixer + website change should be in the same pr, so that they can be synced? but i'll approve the mixer pr, so that the mixer submodule in this repo can be brought up to date with the non-backwards compatible change.

changed to use domain and updated mixer submodule in this PR too.

beets (approved):
thx!

PR #816, opened 2021-02-11
[i18n] localize displayed numbers

beets:
Includes numbers shown in chart y-axis labels, choropleth tooltips, ranking tables. Localization is achieved with a combination of the built-in [Intl.NumberFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat) and putting together translations for units (though MDN says otherwise, unit support is not available across browsers, and doesn't have all the units we use).

The translations for units came from [CLDR](http://cldr.unicode.org/) - [summary table](https://unicode-org.github.io/cldr-staging/charts/38/summary/ru.html) and [source](https://github.com/unicode-org/cldr/blob/release-38-1/common/main/de.xml).

Some other improvements:
- use "USD" for currency, instead of the ambiguous $
- y-axis units are now tight around the labels (no extra spaces, and no overflows)
- ranking tables now display the full unit name
- Fixed up some margin inconsistencies

Some regressions:
- we can't use d3's precision prefix anymore, which standardizes the number of digits shown in the labels, so it might go from 1M to 800k on the same legend (as opposed to 0.8M before)

Note: We could consider switching from FormatJS to [Globalize](https://github.com/globalizejs/globalize) which uses CLDR baked in. Our usage of FormatJS is mostly imperative anyway.

This should wrap up the last major feature for i18n! Woohoo!


beets:
image

image

image

image

image

regressions:
image

image

static/js/i18n/compiled-lang/de/units.json
```diff
@@ -0,0 +1,82 @@
+{
```
shifucun, line 1: Maybe add instructions on how to update these units in developer guide?

shifucun (approved):
This looks very cool!

Might worth to add some webdriver tests for various i18n pages, especially on these units.

beets:
thanks! yes, lots of webdriver tests required for units and other strings. will add them separately.

PR #818, opened 2021-02-11
Fix loading hint invalidate error

shifucun:
The render process terminates for Class node, hence the loading hint erasing should happen before that.

image

### DIFF
--- a/static/js/browser/kg.js
+++ b/static/js/browser/kg.js
@@ -705,8 +705,8 @@ async function renderKGPage(
   // Add "More" to existing cards.
   util.appendMoreToAll();
 
-  const loadingElem = document.getElementById("page-loading");
   if (!hasPopObsError) {
+    const loadingElem = document.getElementById("page-loading");
     loadingElem.style.display = "none";
   }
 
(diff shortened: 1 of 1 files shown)


pradh:
A screenshot validation link might be nice, given there are no tests...

static/js/browser/kg.js
```diff
@@ -705,7 +705,14 @@ async function renderKGPage(
   // Add "More" to existing cards.
   util.appendMoreToAll();
 
-  if (NO_POP_TYPES.includes(type)) return;
+  const loadingElem = document.getElementById("page-loading");
```
beets, line 708: nit - move this into the if block since it's not used otherwise

pradh:
LG, and thanks for screenshot and the PR. Defer approval to cyin...

static/js/browser/kg.js
shifucun: Done

shifucun (commented):
> A screenshot validation link might be nice, given there are no tests...

added screenshot

beets (approved):
thx!

PR #819, opened 2021-02-12
Fix broken links to the tutorials page

pradh:
Testing: brought up a local server and navigated to "Tutorials" under "Documentation". Also, checked http://localhost:8080/colab.


beets (approved):
thanks for the fix! this was new after yesterday's merge of the refactor

pradh:
> thanks for the fix! this was new after yesterday's merge of the refactor

makes me wonder if we should have some general prober that scrapes our top-level pages, navigates to links and checks for 404 or something. I'll add to agenda for prod mtg next week.

PR #822, opened 2021-02-17
fix highlight facts

beets:
- Linkify provenance links
- Display only the domain

was:
image

now:
image


shifucun (approved):
thanks for the fix!

PR #823, opened 2021-02-17
[i18n] update some translations

beets:
A few updated strings from TC


shifucun (approved):
need to learn some foreign language now :)

beets:
my usage of google translated has skyrocketed this month :)

PR #826, opened 2021-02-18
[i18n] Update chart coloring functions to handle i18n strings

beets:
The former method leaves many charts in a deep purple, since most of the other languages have characters that are above the Latin-1 range. This brings the label into stable position between a-z so pages should look more colorful again (esp in ru, hi, ja, ko).

Note that this will change the colors for the en charts as well.

before:
image

new:
image


shifucun:
Does the colors of group bar charts change?

shifucun (approved):

beets:
Bar charts are unchanged - since there are domain arrays of various lengths, the colors continue to be evenly spread throughout the predefined color range.

image

PR #827, opened 2021-02-24
[i18n] set and use locale globally in templates

beets:
this allows us to set the language variable in the html tag of all pages


shifucun (approved):
nice!

PR #829, opened 2021-02-24
Fix axis label line wrapping

beets:
There were a few issues with line wrapping before, the biggest of which was sometimes empty lines were added.

This should fix most of the issues, and there are now tests to verify the behavior.

### DIFF
--- a/static/js/chart/base.test.ts
+++ b/static/js/chart/base.test.ts
@@ -14,7 +14,8 @@
  * limitations under the License.
  */
 
-import { shouldFillInValues } from "./base";
+import * as d3 from "d3";
+import { shouldFillInValues, wrap } from "./base";
 
 test("shouldFillInValues", () => {
   let series = [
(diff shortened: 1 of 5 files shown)


static/js/chart/base.test.ts
```diff
@@ -65,3 +66,43 @@ test("shouldFillInValues", () => {
   ];
   expect(shouldFillInValues(series)).toBe(true);
 });
+
+beforeEach(() => {
+  // JSDom does not define SVGTSpanElements, and use SVGElement instead. Defines
+  // a shim for getComputedTextLength where each character is 1 px wide.
+  (window.SVGElement as any).prototype.getComputedTextLength = function () {
```
shifucun, line 73: this is nice! I remember trying this last year but got no luck...

shifucun (approved):

static/js/chart/base.test.ts
beets: yea, it was quite a project to get the tests going...

PR #833, opened 2021-03-02
Forward other parameters from search explore more

beets:
The URL's from explore more use the dcid parameter, rather than path, e.g. https://datacommons.org/place?utm_medium=explore&dcid=geoId/1304000&mprop=count&popt=Person&hl=en

We were correcting that to /place/geoId/1304000 and only forwarding along the topic parameter. That's why the utm_medium wasn't picked up by GA.

This fixes things so all parameters are forwarded. If this works well from an analytics standpoint, I'll fix the URL's on search to avoid the redirect.

### DIFF
--- a/server/configmodule.py
+++ b/server/configmodule.py
@@ -19,6 +19,7 @@ class Config:
     API_PROJECT = ''
     GA_ACCOUNT = ''
     MAPS_API_KEY = ''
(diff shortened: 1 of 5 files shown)


shifucun (approved):
thanks for the fix!

PR #835, opened 2021-03-03
[bulk download] add place names, and handle missing data

beets:


static/js/tools/timeline_bulk_download.ts
```diff
@@ -75,13 +76,18 @@ function downloadBulkData(
...
+): void {
   const results = [];
-  results.push("placeDcid," + statVars.join(","));
+  results.push("placeDcid,placeName," + statVars.join(","));
   for (const place of placeDcids.sort()) {
-    const row = [place];
+    const row = [`"${place}"`, `"${placeNames[place]}"`];
     for (const sv of statVars) {
-      row.push(data[sv]["stat"][place]["value"]);
+      row.push(data[sv]["stat"][place]["value"] || "-");
```
pradh, line 90: Is an empty CSV cell value better than "-" I wonder? (e.g, utils like pandas have built-ins to skip empty values)

beets: good point, i can leave it blank.

pradh (approved):
Thanks for the name fix!

shifucun (approved):

PR #837, opened 2021-03-04
Extract cached_name to a shared library

chejennifer:
- Create a new file for shared flask APIs and functions
- Extract the cached_name function from place api into the new shared file


shifucun (approved):
thanks for the update!

PR #839, opened 2021-03-05
Override covid test denominator to be the total population

shifucun:
Currently the denominator is the total test case which is not accurate when specify "per capita". Also we don't have the data for total test case data.


static/data/hierarchy_statsvar.json
```diff
@@ -238403,11 +238403,11 @@
      },
      {
       "sv": [
-       "Count_HousingUnit_HomeValue1000000To1499999USDollar"
+       "Count_HousingUnit_HomeValue1000000OrMoreUSDollar"
```
shifucun, line 238406: this is a swap of two stat vars, should be harmless

beets (commented):
+1 to the housing unit swap

for covid though, we were considering this before, and both are meaningful in different ways. we landed on # total tests because people take the test multiple times, and this is a measure of disease prevalence

shifucun:
> +1 to the housing unit swap
> 
> for covid though, we were considering this before, and both are meaningful in different ways. we landed on # total tests because people take the test multiple times, and this is a measure of disease prevalence

Oh I see. So the issue with using the total test is that data is not present for many places, which makes the perCapita not working: https://datacommons.org/tools/timeline#place=geoId%2F06085&statsVar=CumulativeCount_MedicalTest_ConditionCOVID_19_Positive%2C5%2C2%2C0%2C1%2CCumulativeCount_MedicalTest_ConditionCOVID_19

beets (approved):
gotcha. i suppose it's also "per capita" and not "positive rate", so dividing by population is expected for per capita. thanks fro the explanation!

shifucun:
thanks for review and explanation :)

PR #844, opened 2021-03-10
Improve flask logging message

shifucun:
* Remove DEBUG message!
* Let mixer error message show up..

### DIFF
--- a/server/services/datacommons.py
+++ b/server/services/datacommons.py
@@ -433,8 +433,9 @@ def send_request(req_url,
 
     if response.status_code != 200:
         raise ValueError(
-            'Response error: An HTTP {} code was returned by the mixer. '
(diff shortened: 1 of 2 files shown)


server/services/datacommons.py
```diff
@@ -435,8 +434,8 @@ def send_request(req_url,
     if response.status_code != 200:
         raise ValueError(
             'Response error: An HTTP {} code was returned by the mixer. '
-            'Printing response\n{}'.format(response.status_code,
-                                           response.reason))
+            'Printing response:\n{}'.format(response.status_code,
```
beets, line 437: should we keep reason as well?

beets (approved):
nice update

server/services/datacommons.py
shifucun: Added back

shifucun:
Thanks!

PR #846, opened 2021-03-10
[Graph Browser 2] use drawLineChart function for observationChart

chejennifer:
Changes:
- update the function drawLineChart to take in handler function for dots being clicked and change the data format of the dots
- update the observationChart component to use drawLineChart function.

screen-recording (2)

### DIFF
--- a/static/js/browser2/observation_chart.tsx
+++ b/static/js/browser2/observation_chart.tsx
@@ -19,21 +19,16 @@
  */
 
 import React from "react";
-import _ from "lodash";
 import * as d3 from "d3";
 import axios from "axios";
+import { DataGroup } from "../chart/base";
+import { drawLineChart } from "../chart/draw";
+import { DotDataPoint } from "../chart/types";
 
 // Chart size
 const WIDTH = 500;
 const HEIGHT = 250;
 
-// Chart margin.
-const TOP = 50;
-const RIGHT = 230;
-const BOTTOM = 50;
-const LEFT = 100;
-
-const LINE_STROKE_COLOR = "rgb(31, 119, 180)";
 const URI_PREFIX = "/browser2/";
 const TOOLTIP_ID = "tooltip";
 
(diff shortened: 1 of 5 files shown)


static/js/browser2/observation_chart.tsx
```diff
@@ -81,159 +76,61 @@ export class ObservationChart extends React.Component<
...
-    x.domain(d3.extent(plotData, (d) => d.time));
-
-    const yMax = d3.max(plotData, (d) => d.value);
-    const yMin = d3.min(plotData, (d) => d.value);
-
-    // When yMax and yMin are almost equal, need have a non-zero range for d3 to
-    // plot the y axis.
-    // If the min value is 0, then always start from it. Otherwise choose a tick
-    // range that covers yMin and yMax.
-    if (yMax - yMin < 0.0001) {
```
beets, line 117: I'm not sure of exact nodes where these problems manifested, but it would be good to see if these are fixed by the shared charting code.

static/js/browser2/observation_chart.tsx
```diff
@@ -81,159 +76,61 @@ export class ObservationChart extends React.Component<
...
+      });
+    });
+    const dataGroups = [new DataGroup("", data)];
+    const svgContainerId: string = "svg-container" + this.props.idx;
+    drawLineChart(
+      svgContainerId,
+      WIDTH,
+      HEIGHT,
+      dataGroups,
+      true,
```
beets, line 93: please add a comment that even though all dots are drawn, they are not displayed except on hover (and that is controlled via css).

beets (approved):
thanks for the update!

static/js/chart/draw.ts
```diff
@@ -660,8 +665,7 @@ function drawLineChart(
     });
     const hasGap = shouldFillInValues(dataset);
     hasFilledInValues = hasFilledInValues || hasGap;
-    const shouldAddDots = dataset.length < 12;
-
+    const shouldAddDots = dataset.length < 12 || showAllDots;
```
shifucun, line 668: nit: make 12 a constant

shifucun (approved):

static/js/browser2/observation_chart.tsx
chejennifer: Actually, all the dots are displayed even when not hovering https://screenshot.googleplex.com/B8sxhGugVGBqPG2

static/js/browser2/observation_chart.tsx
chejennifer: Done

static/js/chart/draw.ts
chejennifer: Done

chejennifer:
thanks for the reviews!

static/js/browser2/observation_chart.tsx
beets: Ah.. we might want to have a limit for that, since it won't work out so well for dense timelines (like unemployment)

chejennifer: oh, I see what you mean. I can look into only showing dots on hover when there is a dense timeline

PR #848, opened 2021-03-10
Add a config flag to indicate svobs mode (for sv migration)

shifucun:


server/configmodule.py
```diff
@@ -10,6 +10,7 @@ class Config:
     WEBDRIVER = False
     DEVELOPMENT = False
     LITE = False
+    SVOBS = False
```
beets, line 13: Add a note here indicating what it's used for?

beets (approved):

PR #852, opened 2021-03-11
Add a local development mode for statvarobs migration

shifucun:


server/services/datacommons.py
```diff
@@ -23,23 +23,9 @@
...
-elif os.environ.get('FLASK_ENV') == 'staging':
-    cfg = import_string('configmodule.StagingConfig')()
-elif os.environ.get('FLASK_ENV') == 'autopush':
-    cfg = import_string('configmodule.AutopushConfig')()
-elif os.environ.get('FLASK_ENV') == 'minikube':
-    cfg = import_string('configmodule.MinikubeConfig')()
-elif os.environ.get('FLASK_ENV') == 'gke':
-    cfg = import_string('configmodule.GKEConfig')()
-else:
-    cfg = import_string('configmodule.DevelopmentConfig')()
```
chejennifer, line 42: I'm not sure if this could be a problem, but it looks like libconfig.get_config() will raise a ValueError if no FLASK_ENV specified,  but here, the developmentConfig is used if no FLASK_ENV specified

chejennifer (approved):

server/services/datacommons.py
shifucun: Fallback to developmentConfig was okay at the beginning but as we add more config it is better to make it explicit. This consolidation also helps on it as the server should always have a valid FLASK_ENV.

chejennifer: Makes sense, thanks for the explanation!

server/configmodule.py
```diff
@@ -65,6 +65,16 @@ class DevelopmentLiteConfig(Config):
     SCHEME = 'http'
 
 
+class DevelopmentSvObsConfig(Config):
```
beets, line 68: could you add a note at the top of this file that config.py should be updated if new configs are added?

beets (approved):
I wonder if there's a way for us to write a test that all configs have the appropriate flags set. Maybe a test for config.py which uses reflection to load all config classes, and test that no error is thrown?

server/configmodule.py
shifucun: Done

shifucun:
> I wonder if there's a way for us to write a test that all configs have the appropriate flags set. Maybe a test for config.py which uses reflection to load all config classes, and test that no error is thrown?

Added test for development configs. The gke configs used environment variables from yaml files.  Those could be covered in a test for those yaml files. I will add a test in later PR.

PR #853, opened 2021-03-11
[Graph Browser 2] Add weather charts

chejennifer:
Changes:
- update browser.ts to get the node type before rendering a page
- add a new weatherChartSection component that gets weather data and renders charts for that data

screen-recording (3)


server/templates/browser/node.html
```diff
@@ -22,7 +22,6 @@
    {% set subpage_url = url_for('browser2.browser2') %}
 
    {% block head %}
-     <link rel="stylesheet" href={{url_for('static', filename='css/kg.min.css', t=config['VERSION'])}} >
```
beets, line 25: thx!

static/js/browser2/browser.ts
```diff
@@ -19,22 +19,52 @@ import ReactDOM from "react-dom";
...
+
+function getNodeType(listOfTypes, statVarId): string {
+  if (!_.isEmpty(statVarId)) {
+    return nodeTypeEnum.PLACE_STAT_VAR;
+  }
+  let type = "";
+  if (listOfTypes.length > 0) {
+    type = listOfTypes[0];
+  }
+  for (const targetType of ["State", "County", "City"]) {
```
beets, line 57: could we pull out the list of types to a file constant? that also helps to make it self documenting

static/js/browser2/browser.ts
```diff
@@ -19,22 +19,52 @@ import ReactDOM from "react-dom";
...
+          nodeName,
+          statVarId,
+        }),
+        document.getElementById("node")
+      );
+    })
+    .catch(() => removeLoadingMessage());
 };
+
+function getNodeType(listOfTypes, statVarId): string {
```
beets, line 49: this util belongs together with the definition of nodeTypeEnum. and perhaps both belong to shared.ts (or a new types.ts)

i also think it would be much clearer if this strictly returns nodetypeEnum (and that should not map to string and perhaps be a strict numeric enum). this now introduces a few new node types, and it's not really clear what the universe of return values could be. 

additionally, i think this function would be much simpler if there was a prioritized mapping for nodeType (from kg) to pageDisplayType (for what type of browser page to display). then this is a simple matter of walking down the map and returning the first match without any special cases.

please also add a one-liner comment describing the function

static/js/browser2/browser.ts
beets: it's also not immediately clear why state and county are part of "place_stat_var", please add a comment

static/js/browser2/browser.ts
```diff
@@ -19,22 +19,52 @@ import ReactDOM from "react-dom";
...
+  if (!_.isEmpty(statVarId)) {
+    return nodeTypeEnum.PLACE_STAT_VAR;
+  }
+  let type = "";
+  if (listOfTypes.length > 0) {
+    type = listOfTypes[0];
```
beets, line 55: this initialization doesn't seem necessary if line 57 includes zip codes

static/js/browser2/weather_chart_section.tsx
```diff
@@ -0,0 +1,110 @@
...
+/**
+ * Component for rendering charts for various weather related properties.
+ */
+
+import React from "react";
+import axios from "axios";
+import _ from "lodash";
+import { ObservationChart } from "./observation_chart";
+
+const WEATHER = [
```
beets, line 26: please be more specific with the variable names. these are weather_property_names?

static/js/browser2/weather_chart_section.tsx
```diff
@@ -0,0 +1,110 @@
...
+  "snowfall",
+  "barometricPressure",
+];
+
+interface WeatherChartSectionPropType {
+  dcid: string;
+}
+
+interface WeatherChartSectionStateType {
+  // TODO (chejennifer): get rid of the any type
```
beets, line 39: it's useful to do this early enough, to catch errors in programming

beets (commented):
it's really nice to start seeing this as components. very much simpler to understand the logic and the many pieces being pulled together.

static/js/browser2/browser.ts
chejennifer: what do you mean state and county are part of place_stat_var? Right now, the nodeType is considered place_stat_var if there is a statvar defined. But I guess place_stat_var is more of a pageType than a nodeType and I should do a better job of distinguishing the two

chejennifer: I'm not sure pulling out the list of types to a file constant is necessary. This line was copied from the old kg. I believe this line is there because a node can have multiple typeOf values, but if the list contains one of those 3 types, then we wanted to use that to decide what is rendered for the node.

shifucun: I agree here is more about the "page type"

static/js/browser2/browser_page.tsx
```diff
@@ -53,6 +56,10 @@ export class BrowserPage extends React.Component<BrowserPagePropType> {
           ) : (
             <ArcSection dcid={this.props.dcid} nodeName={this.props.nodeName} />
           )}
+          {this.props.nodeType === nodeTypeEnum.CITY ||
+          this.props.nodeType == nodeTypeEnum.CENSUS_ZIPCODE_TABULATION_AREA ? (
```
shifucun, line 60: Would be nice to have a util function that centralize the logic of the rendering "components" given the node/page type.

static/js/browser2/weather_chart_section.tsx
chejennifer: done

static/js/browser2/weather_chart_section.tsx
chejennifer: done

static/js/browser2/browser.ts
chejennifer: done. Changed how to figure out the type of page to render for a nodeType

static/js/browser2/util.ts
```diff
@@ -0,0 +1,76 @@
...
+  PLACE_STAT_VAR,
+  PLACE_WITH_WEATHER_INFO,
+  GENERAL,
+}
+
+/**
+ * Mapping for nodeTypes that need to render a special page type
+ */
+export const nodeTypeToPageDisplayTypeMapping = {
+  CensusZipCodeTabulationArea: PageDisplayType.PLACE_WITH_WEATHER_INFO,
```
beets, line 43: do we need a mapping  of other place types to PLACE_STAT_VAR? or add a note that it might be required as more display types are added.

beets (approved):
thanks for making the changes!

static/js/browser2/util.ts
chejennifer: we don't need a mapping of other place types to PLACE_STAT_VAR because PLACE_STAT_VAR page type is dependent on a statVarId being provided. But yes, there may be more mappings as more display types are added

static/js/browser2/observation_chart.tsx
```diff
@@ -33,11 +34,11 @@ const URI_PREFIX = "/browser2/";
 const TOOLTIP_ID = "tooltip";
 
 interface ObservationChartPropType {
-  // TODO (chejennifer): get rid of the any type
-  sourceSeries: any;
+  sourceSeries: SourceSeries;
```
shifucun, line 37: nice!

shifucun (approved):
looking good!

static/js/browser2/util.ts
```diff
@@ -0,0 +1,76 @@
...
+  City: PageDisplayType.PLACE_WITH_WEATHER_INFO,
+};
+
+/**
+ * Returns the type of page to display
+ * @param listOfTypes
```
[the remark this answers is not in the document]
chejennifer: done

PR #854, opened 2021-03-12
[Graph Browser 2] Open browser2 when in svobs mode

chejennifer:
Update the browser route so that it opens browser2 when in svobs mode and remove the browser2 route.


server/routes/browser.py
```diff
@@ -26,4 +28,12 @@ def kg_main():
 
 @bp.route('/<path:dcid>')
 def kg_entity(dcid):
```
shifucun, line 30: can update the function names "kg_main" and "kg_entity" as well?

shifucun (approved):

server/routes/browser.py

PR #856, opened 2021-03-15
[Graph Browser 2] Small fixes/changes

chejennifer:
Some small fixes/changes:
- limit height of in-arcs sections and allow scrolling
- only show dots on hover on the observation charts when there are too many dots
- update prefix of links to "/browser/" instead of "/browser2/" 

screen-recording (4)
screen-recording (5)

### DIFF
--- a/static/css/browser.scss
+++ b/static/css/browser.scss
@@ -60,7 +60,7 @@ td a {
 .card {
   box-shadow: 0 2px 4px 0 rgba(0, 0, 0, 0.2);
   padding: 7px 15px;
-  padding-right: 0px;
+  padding-right: 0;
   margin: 12px 0;
   position: relative;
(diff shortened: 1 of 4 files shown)


static/css/browser.scss
```diff
@@ -87,4 +93,9 @@ td a {
 
 .no-click circle {
   cursor: auto;
-}
\ No newline at end of file
+}
+
+.in-arc-table {
+  max-height: 450px;
```
beets, line 99: i'd suggest a multiple of the lineheight / table row (also specified in the same unit as lineheight - usually em)

beets (approved):

static/css/browser.scss
chejennifer: done

PR #857, opened 2021-03-16
[Graph Browser 2] Show arcs for place stat var pages

chejennifer:
For place stat var pages, show the arcs for the stat var by updating BrowserPage to also render the ArcSection for the PLACE_STAT_VAR pageDisplayType. 
Screen Shot 2021-03-15 at 8 14 33 PM

### DIFF
--- a/static/js/browser2/arc_section.tsx
+++ b/static/js/browser2/arc_section.tsx
@@ -24,10 +24,12 @@ import _ from "lodash";
 
 import { InArcSection } from "./in_arc_section";
 import { OutArcSection } from "./out_arc_section";
+import { PageDisplayType } from "./util";
 
 interface ArcSectionPropType {
   dcid: string;
   nodeName: string;
+  pageDisplayType: PageDisplayType;
 }
 
 interface ArcSectionStateType {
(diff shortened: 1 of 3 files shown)


static/js/browser2/arc_section.tsx
```diff
@@ -64,12 +66,14 @@ export class ArcSection extends React.Component<
           labels={this.state.outLabels}
           provDomain={this.state.provDomain}
         />
-        <InArcSection
-          nodeName={this.props.nodeName}
-          dcid={this.props.dcid}
-          labels={this.state.inLabels}
-          provDomain={this.state.provDomain}
-        />
+        {this.props.pageDisplayType !== PageDisplayType.PLACE_STAT_VAR ? (
```
shifucun, line 69: It might be better to pass this boolean from the parent. This way this component is standalone (does not depend on parent component logic)

shifucun (approved):

static/js/browser2/arc_section.tsx
chejennifer: done

PR #861, opened 2021-03-18
[Graph Browser 2] Show image for biological specimen pages

chejennifer:
- Add new ImageSection component that displays images from the imageUrl property of a node
- Render ImageSection component when the PageDisplayType is BiologicalSpecimen
screen-recording (7)


shifucun (approved):
thanks for adding it back!

beets (approved):
v nice!

PR #862, opened 2021-03-19
remove populationType filter

chejennifer:


static/js/browser2/out_arc_section.tsx
```diff
@@ -40,7 +40,6 @@ const IGNORED_OUT_ARC_PROPERTIES = new Set([
   "geoJsonCoordinatesDP2",
   "geoJsonCoordinatesDP3",
   "censusACSTableId",
```
pradh, line 42: why are we filtering censusACSTableId?

chejennifer: I think that was something I had added by accident before, I will remove that too.

pradh (approved):
Thanks!

PR #863, opened 2021-03-19
[cleanup] delete geojson_simplifier which was moved to the data repo

beets:
The code is now at:
https://github.com/datacommonsorg/data/tree/master/scripts/us_census/geojsons_low_res


chejennifer (approved):
Thanks for the cleanup!

PR #864, opened 2021-03-22
[Graph Browser 2] updated observation chart tooltip and hover 

chejennifer:
- Increased the hover area to the whole chart
- added a line to highlight which point is currently being looked at
- updated the tooltip to be fixed either at the top of the page or at the bottom of the page

Note: I feel like the structure of how I added these changes might not be the most clean/reusable, let me know if you have any suggestions to improve this

screen-recording (8)


static/js/browser2/observation_chart.tsx
```diff
@@ -157,29 +155,34 @@ export class ObservationChart extends React.Component<
...
-  private handleDotLeave = (svgContainerId: string) => (): void => {
-    d3.select("#" + svgContainerId)
       .select(`#${TOOLTIP_ID}`)
-      .style("display", "none");
+      .text(text);
+    const rect = (tooltipSelect.node() as HTMLDivElement).getBoundingClientRect();
+    const width = rect.width;
+    const height = rect.height;
+    const topOffset = 5;
+    let left = datapointX - width / 2;
```
beets, line 174: some comments here would be helpful to explain the expected outcome.

static/js/chart/draw.ts
```diff
@@ -605,8 +605,11 @@ function drawGroupBarChart(
  * @param height
  * @param dataGroups
  * @param showAllDots
+ * @param highlightOnHover
+ * @param tooltipId
  * @param unit
  * @param handleDotClick
+ * @param handleDotHighlight
```
beets, line 612: please document expected parameters

beets (commented):
thanks so much for adding this!

i wonder if more can be moved into the draw module, since creating a tooltip and placing it within the chart should be handled in a similar way across our charts. the calling code should only need to handle data-related code, while all visual / layout code belongs in the draw module.

static/js/browser2/observation_chart.tsx
```diff
@@ -94,28 +99,21 @@ export class ObservationChart extends React.Component<
         value: Number(values[key]),
       });
     });
-    if (data.length > MAX_DOTS) {
-      document
-        .getElementById("svg-container" + this.props.idx)
-        .classList.add("hide-dots");
-    }
-    const dataGroups = [new DataGroup("", data)];
     const svgContainerId: string = "svg-container" + this.props.idx;
```
shifucun, line 102: can make "svgContainerId" a private variable, then can remove it from the argument of addTooltip() and handleDotHighlight()

static/js/chart/draw.ts
```diff
@@ -724,6 +737,76 @@ function drawLineChart(
...
+              "transform",
+              "translate(" + datapointX + "," + datapointY + ")"
+            );
+            highlightedLine.attr(
+              "transform",
+              "translate(" + datapointX + "," + "0)"
+            );
+            if (handleDotHighlight) {
+              handleDotHighlight(
+                dataGroup.value[idx],
```
shifucun, line 800: Do you think it possible to let handleDotHighlight() compute some of these inputs instead of the caller?

Ideally, we can use this function in different chart (e.g. scatter plot) and the call function only need minimal computation.

chejennifer: currently, handleDotHighlight is a function that gets passed in from the observationChart component and it's the things that we want to do once a dot has been highlighted. I'm not sure it makes the most sense to compute some of these inputs there because the computed inputs are also needed for the positioning of the line and dot that highlights the point of interest. Maybe I could extract the computation into its own function.

chejennifer: This current implementation is specific to line chart and a lot of the code for highlighting is inside the drawLineChart function. Maybe I can see if I can extract things out so that it can be adapted in the future for other charts.

shifucun: I see. It would be nice if this function is used as a shared library and try to decouple it from the drawLineChart. 

But if you feel they are very tightly coupled, it's fine to absorb the function just inside drawLineChart() too.

chejennifer: ok, I think I will keep it absorbed inside drawLineChart() for now, but in the future extract it out to be used by multiple-line charts as well

static/js/browser2/observation_chart.tsx
chejennifer: done

static/js/browser2/observation_chart.tsx
chejennifer: done

static/css/browser.scss
```diff
@@ -117,4 +117,13 @@ td a {
 .error-message {
   font-size: 14px;
   color: var(--dc-red-lite);
-}
\ No newline at end of file
+}
+
+#tooltip {
+  background-color: #fff;
```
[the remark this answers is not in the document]
chejennifer: done

shifucun (commented):
Thanks for the change! How does the place page line chart looks with it?

chejennifer:
> Thanks for the change! How does the place page line chart looks with it?

Just tried it out and the tooltip works well when there is only a single datagroup, but doesn't work when there are multiple. I'll try to figure out how to make it work for multiple datagroups.

static/js/chart/draw.ts
```diff
@@ -737,6 +822,60 @@ function drawLineChart(
...
+  relativeBoundary: Boundary
+): void {
+  const tooltipSelect = d3
+    .select("#" + containerId)
+    .select(`#${TOOLTIP_ID}`)
+    .text(text);
+  const rect = (tooltipSelect.node() as HTMLDivElement).getBoundingClientRect();
+  const width = rect.width;
+  const height = rect.height;
+  const topOffset = 5;
```
beets, line 861: please use a constant for this

static/js/chart/draw.ts
```diff
@@ -737,6 +822,60 @@ function drawLineChart(
...
+ */
+function showTooltip(
+  text: string,
+  containerId: string,
+  datapointX: number,
+  datapointY: number,
+  relativeBoundary: Boundary
+): void {
+  const tooltipSelect = d3
+    .select("#" + containerId)
```
beets, line 855: nit: consider standardizing to format strings like the next line

static/js/chart/draw.ts
```diff
@@ -737,6 +822,60 @@ function drawLineChart(
   return !hasFilledInValues;
 }
 
+/**
+ * Adds tooltip element within a given container.
+ *
+ * @param containerId
+ */
+function addTooltip(containerId: string): void {
```
beets, line 830: nit: these two utility functions should go to the top of the file (defined before they are referenced)

static/js/chart/draw.ts
```diff
@@ -724,6 +735,80 @@ function drawLineChart(
         dots.on("click", handleDotClick);
       }
     }
+
+    if (highlightOnHover) {
+      addTooltip(id);
+      highlight.style("opacity", "0");
+      const highlightedDot = highlight
+        .append("circle")
+        .attr("r", 5)
```
beets, line 744: nit: consider making 5 a constant.

static/js/chart/draw.ts
```diff
@@ -737,6 +822,60 @@ function drawLineChart(
...
+function showTooltip(
+  text: string,
+  containerId: string,
+  datapointX: number,
+  datapointY: number,
+  relativeBoundary: Boundary
+): void {
+  const tooltipSelect = d3
+    .select("#" + containerId)
+    .select(`#${TOOLTIP_ID}`)
```
beets, line 856: should the css for #tooltip move to a common scss file as well?
also consider making it a more specific ID (e.g. draw-tooltip or chart-tooltip since we have a few other similarly id'ed elements in our codebase)

static/js/chart/draw.ts
```diff
@@ -724,6 +735,80 @@ function drawLineChart(
...
+              "transform",
+              "translate(" + datapointX + "," + datapointY + ")"
+            );
+            highlightedLine.attr(
+              "transform",
+              "translate(" + datapointX + "," + "0)"
+            );
+            const dataPoint: DataPoint = dataGroup.value[idx];
+            const tooltipText = `${dataPoint.label}: ${dataPoint.value} ${
+              unit ? unit : ""
```
beets, line 793: for place + timeline explorer, this would need to use the i18n functions. could you please add a todo for that?

static/js/chart/draw.ts
```diff
@@ -724,6 +735,80 @@ function drawLineChart(
         dots.on("click", handleDotClick);
       }
     }
+
+    if (highlightOnHover) {
```
beets, line 739: i see bo's point - this is pretty long and it affects the readability of the overall function.

it would be nice to move it into a helper, even if it does take in a long list of params.  i think passing in both bounding areas for the chart (one for svg transforms, and another for the tooltip / div) could help. and it's fine if it only works for this line chart use case for now.

static/js/chart/draw.ts
```diff
@@ -724,6 +735,80 @@ function drawLineChart(
...
+        .on("mouseover", () => {
+          highlight.style("opacity", "1");
+          tooltip.style("display", "block");
+        })
+        .on("mouseout", () => {
+          highlight.style("opacity", "0");
+          tooltip.style("display", "none");
+        })
+        .on("mousemove", () => {
+          // calculation of data point corresponding to mouse position from https://bl.ocks.org/Qizly/8f6ba236b79d9bb03a80.
```
beets, line 766: this could also be lifted up to a helper and is maybe a first candidate for the small refactor.

beets (commented):
added some suggestions on how to make this more readable.

static/js/chart/draw.ts
chejennifer: done

static/js/chart/draw.ts
chejennifer: done

static/js/chart/draw.ts
chejennifer: done

static/js/chart/draw.ts
chejennifer: done

static/js/chart/draw.ts
chejennifer: done

static/js/chart/draw.ts
chejennifer: done

static/js/chart/draw.ts
chejennifer: done

static/js/chart/draw.ts
chejennifer: done

static/js/chart/draw.ts
```diff
@@ -88,6 +99,200 @@ function appendLegendElem(
...
+    .attr("id", TOOLTIP_ID)
+    .attr("style", "position: absolute; display: none; z-index: 1");
+}
+
+/**
+ * Position and show the tooltip.
+ *
+ * @param contentHTML innerHTML of the tooltip as a string
+ * @param containerId
+ * @param datapointX
```
beets, line 120: please describe the other fields as well

static/js/chart/draw.ts
```diff
@@ -88,6 +99,200 @@ function appendLegendElem(
...
+        ? idx - 1
+        : idx;
+  } else if (idx === listOfTimePoints.length) {
+    idx = idx - 1;
+  }
+  return listOfTimePoints[idx];
+}
+
+/**
+ * Adds highlighting and showing a tooltip on hover for a chart.
```
beets, line 185: specify line chart here for now, esp. given the usage of time points.

static/js/chart/draw.ts
```diff
@@ -88,6 +99,200 @@ function appendLegendElem(
...
+ *
+ * @param svg
+ * @param xScale
+ * @param yScale
+ * @param containerId
+ * @param dataGroups
+ * @param colorFn
+ * @param listOfTimePoints
+ * @param highlight
```
beets, line 194: could you describe these quickly? especially the highlight param. (i realize we haven't been doing this so far...)

also consider making the param name more specific - highlightArea? highlightGroup? highlightPlane?

beets (approved):
thanks for the refactor! it's easier to parse now :)

static/js/chart/draw.ts
chejennifer: done

static/js/chart/draw.ts
chejennifer: done

static/js/chart/draw.ts
chejennifer: done

static/js/chart/draw.ts
```diff
@@ -661,9 +868,12 @@ function drawLineChart(
   const colorFn = getColorFn(legendText);
 
   let hasFilledInValues = false;
+  const timePoints = new Set<number>();
   for (const dataGroup of dataGroups) {
     const dataset = dataGroup.value.map((dp) => {
-      return [new Date(dp.label).getTime(), dp.value];
+      const time = new Date(dp.label).getTime();
```
shifucun, line 874: There are few places the label is converted to time. Would it help to add an optional "time" field in DataPoint struct and set it during initialization (and only line chart will set it)?

chejennifer: I think the only problem is that since this field would be optional, when trying to access the time of a DataPoint, I would have to check that this field is set and if not, I would still want to try converting the label to time. Or can I just assume that if the time field is not set, then this datapoint is not for a time.

shifucun: Yeah, when building the DataPoint, should ensure "time" is set if the label is about time. DrawLineChart could assume upstream has set it, since we only draw time series in line chart.

chejennifer: Then if we're assuming "time" is set if the label is about time, I should also update DrawGroupLineChart to use this time field and set it during initialization of the DataPoints for a group line chart? Maybe I should do this in a separate PR.

shifucun: separate PR sounds good

shifucun (approved):

PR #865, opened 2021-03-22
[Graph Browser 2] Show units in chart titles

chejennifer:
- update weather chart titles and observation chart titles to include the unit 

Screen Shot 2021-03-22 at 1 45 35 PM


static/js/browser2/util.ts
```diff
@@ -88,3 +88,21 @@ export function removeLoadingMessage(): void {
     loadingElem.style.display = "none";
   }
 }
+
+/**
+ * Returns the unit for a sourceSeries if there is a unit.
+ * @param sourceSeries
+ */
+export function getUnit(sourceSeries: SourceSeries): string {
```
shifucun, line 96: would be nice to start adding unittest for util functions

chejennifer: done

shifucun: thanks!

shifucun (approved):

PR #866, opened 2021-03-25
[Graph Browser 2] Use triples to get out arcs for observation nodes

chejennifer:
Observation nodes have no property labels or property values, so in order to get out arcs, fetch data from api/browser/triples

### DIFF
--- a/static/js/browser2/arc_section.tsx
+++ b/static/js/browser2/arc_section.tsx
@@ -20,7 +20,6 @@
 
 import React from "react";
 import axios from "axios";
-import _ from "lodash";
 
 import { InArcSection } from "./in_arc_section";
 import { OutArcSection } from "./out_arc_section";
--- a/static/js/browser2/arc_section.tsx
+++ b/static/js/browser2/arc_section.tsx
@@ -35,7 +35,8 @@ interface ArcSectionStateType {
   inLabels: string[];
   outLabels: string[];
   provDomain: { [key: string]: URL };
-  dataFetched: boolean;
+  isDataFetched: boolean;
+  errorMessage: string;
 }
 
 export class ArcSection extends React.Component<
@@ -45,10 +46,11 @@ export class ArcSection extends React.Component<
   constructor(props: ArcSectionPropType) {
     super(props);
     this.state = {
+      errorMessage: "",
       inLabels: [],
+      isDataFetched: false,
       outLabels: [],
       provDomain: {},
-      dataFetched: false,
     };
   }
 
@@ -57,7 +59,7 @@ export class ArcSection extends React.Component<
   }
 
   render(): JSX.Element {
-    if (!this.state.dataFetched) {
+    if (!this.state.isDataFetched) {
       return null;
     }
     return (
@@ -74,6 +76,9 @@ export class ArcSection extends React.Component<
             labels={this.state.inLabels}
             provDomain={this.state.provDomain}
           />
+        ) : null}
+        {this.state.errorMessage ? (
+          <div className="error-message">{this.state.errorMessage}</div>
         ) : null}
       </>
     );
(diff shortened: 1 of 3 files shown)


static/js/browser2/arc_section.tsx
```diff
@@ -35,6 +35,7 @@ interface ArcSectionStateType {
   inLabels: string[];
   outLabels: string[];
   provDomain: { [key: string]: URL };
+  dataFetched: boolean;
```
beets, line 38: i find prefixing boolean parameters with "is" or "has" helpful for readability, e.g. isDataFetched.

static/js/browser2/arc_section.tsx
```diff
@@ -97,8 +98,13 @@ export class ArcSection extends React.Component<
...
           outLabels: labelsData["outLabels"],
           provDomain,
+          dataFetched: true,
         });
-      }
-    );
+      })
+      .catch(() => {
+        this.setState({
+          dataFetched: true,
```
beets, line 106: i suppose it's whether the fetch attempt has completed, not that the data has been fetched. should an error message be displayed?

static/js/browser2/out_arc_section.tsx
```diff
@@ -174,4 +179,43 @@ export class OutArcSection extends React.Component<
...
+      outArcs.forEach((outArc) => {
+        const predicate = outArc.predicate;
+        if (IGNORED_OUT_ARC_PROPERTIES.has(predicate)) {
+          return;
+        }
+        if (!outArcsByPredicateAndProvenance[predicate]) {
+          outArcsByPredicateAndProvenance[predicate] = {};
+        }
+        const provId = outArc.provenanceId;
+        if (!(provId in outArcsByPredicateAndProvenance[predicate])) {
```
beets, line 199: before this line, consider pulling out "outArcsByPredicateAndProvenance[predicate]" to a local var, which you can reference more easily here and below

static/js/browser2/arc_section.tsx
chejennifer: done

static/js/browser2/arc_section.tsx
chejennifer: done

static/js/browser2/out_arc_section.tsx
chejennifer: done

static/js/browser2/arc_section.tsx
chejennifer: done

static/js/browser2/arc_section.tsx
chejennifer: done

static/js/browser2/out_arc_section.tsx
chejennifer: done

static/js/browser2/out_arc_section.tsx
```diff
@@ -141,17 +147,19 @@ export class OutArcSection extends React.Component<
             if (!(predicate in outArcsByPredicateAndProvenance)) {
               outArcsByPredicateAndProvenance[predicate] = {};
             }
+            const outArcsOfPredicate =
```
shifucun, line 150: Both "Predicate" and "Property" are used which is a bit in-consistent, might be using "property" throughout? And I feel the names here a bit too long, something like "outArcsByPropProv" might work as well.

shifucun (commented):
Overall looks good! Just a minor comment.

static/js/browser2/out_arc_section.tsx
chejennifer: I think using either "predicate" or "property" will present the same problem. The data we get back from the triples api uses "predicate", but the data we get back from the propVals api uses "property". I believe this component mainly uses "predicate" right now, do you want me to change it to "property"?

shifucun: I see. That's a good point. We should use "property" for our next version of API so the naming are aligned. Then no need to do the conversion here.

chejennifer: ok, sounds good! But, I'll shorten the names

shifucun (approved):
Thanks for the update!

beets (approved):

beets:
thanks for the updates!

PR #871, opened 2021-03-29
Add optional time field for DataPoint

chejennifer:
Adding an optional time field for DataPoint that should be set if a DataPoint has a label that is a data string. This was mentioned in: https://github.com/datacommonsorg/website/pull/864


shifucun (approved):
Thanks for the change, looks much simpler!

PR #872, opened 2021-03-29
Add tooltip and highlight for all line charts and group line charts

chejennifer:
- update drawGroupLineChart, drawLineChart, and addHighlightOnHover to show highlight and tooltip for both group line charts and line charts
- update place page line charts to show the highlighting/tooltips

timeline tool:
screen-recording (12)

place page charts:
screen-recording (10)

international place page charts:
screen-recording (11)

dev page with all the special charts:
screen-recording (9)

### DIFF
--- a/static/js/chart/draw.ts
+++ b/static/js/chart/draw.ts
@@ -253,10 +253,7 @@ function addHighlightOnHover(
       let minDataPointY = chartAreaBoundary.bottom;
       highlightDots
         .attr("transform", (d: DataGroup) => {
-          const dataPoint = d.value.find(
-            (dataPoint) =>
-              new Date(dataPoint.label).getTime() === highlightedTime
-          );
+          const dataPoint = d.value.find((val) => val.time === highlightedTime);
           if (dataPoint) {
             const dataPointY = yScale(dataPoint.value);
             minDataPointY = Math.min(minDataPointY, dataPointY);
@@ -264,9 +261,7 @@ function addHighlightOnHover(
           }
         })
         .style("opacity", (d: DataGroup) => {
-          const dataPoint = d.value.find(
-            (data) => new Date(data.label).getTime() === highlightedTime
-          );
+          const dataPoint = d.value.find((val) => val.time === highlightedTime);
           if (dataPoint && dataPoint.value) {
             return "1";
           } else {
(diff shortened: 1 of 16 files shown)


static/js/chart/draw.ts
```diff
@@ -924,25 +916,52 @@ function drawLineChart(
...
+      for (const dataGroup of dataGroups) {
+        const dataPoint = dataGroup.value.find(
+          (val) => val.time === highlightedTime
+        );
+        if (dataPoint && dataPoint.value) {
+          const label =
+            dataGroups.length === 1 ? dataPoint.label : dataGroup.label;
+          tooltipDate = dataPoint.label;
+          tooltipContent =
+            tooltipContent +
```
shifucun, line 946: can use "+="

static/js/chart/draw.ts
```diff
@@ -1135,6 +1159,57 @@ function drawGroupLineChart(
...
+  const chartAreaBoundary = {
+    bottom: height - bottomHeight,
+    left: leftWidth,
+    right: width - MARGIN.right,
+    top: 0,
+  };
+  const highlightColorFn = (place: string, dataGroup: DataGroup) => {
+    return plotParams.lines[place + statsVarsTitle[dataGroup.label]].color;
+  };
+  const getTooltipContent = (highlightedTime: number): string => {
```
shifucun, line 1175: maybe define this as "function" to be consistent with elsewhere

static/js/chart/draw.ts
```diff
@@ -1135,6 +1159,57 @@ function drawGroupLineChart(
...
+        if (dataPoint && dataPoint.value) {
+          tooltipDate = `${dataPoint.label}<br/>`;
+          let rowLabel = "";
+          if (dataGroups.length > 1) {
+            rowLabel += dataGroup.label;
+          }
+          if (Object.keys(dataGroupsDict).length > 1) {
+            rowLabel += ` ${place}`;
+          }
+          tooltipContent =
```
shifucun, line 1193: can use "+="

static/js/chart/draw.ts
shifucun: also would be nice to add function comment about what content it generates

shifucun: oh I see, there are two "getTooltipContent". maybe better two make two functions like getTooltipContentforLine() and getTooltipContentforGroupLine() as standalone functions? And would be better if can consolidate them

PR #875, opened 2021-04-01
Create a dev instance and rename previous 'dev' env to 'local'

shifucun:


chejennifer (approved):
Thanks for adding this!

PR #883, opened 2021-04-15
[Graph Browser 2] Observation chart updates

chejennifer:
Various Updates related to Observation charts:
- ability to toggle between chart and table view for observation charts
- allow observations (dots in chart view and rows in table view) to be clickable when rendered
    - if mapping of observationDate to observation dcid hasn't been retrieved yet, send query to get observation id for clicked observation. While waiting for the request to complete, show spinner
- move provenance to underneath the chart/table
- in the expanded stat var node in the stat var hierarchy, show links to the stat var browser page and the place stat var browser page

the dev instance has been updated with these changes (eg. [https://dev.datacommons.org/browser/geoId/12086](https://dev.datacommons.org/browser/geoId/12086))


pradh:
This looks so cool Jennifer!

server/routes/api/browser.py
```diff
@@ -64,42 +64,90 @@ def get_property_labels(dcid):
...
-        return Response(json.dumps("error: must provide a statVar field"),
-                        400,
-                        mimetype='application/json')
-    measurement_method = request.args.get("measurementMethod", "")
-    observation_period = request.args.get("obsPeriod", "")
+def get_sparql_query(place_id, stat_var_id, date, measurement_method,
+                     observation_period):
     measurement_method_triple = ""
     if measurement_method:
         measurement_method_triple = f"""?svObservation measurementMethod {measurement_method} ."""
```
beets, line 71: nit: you don't need a multiline string for this (here and below), f"..." should be sufficient

server/routes/api/browser.py
```diff
@@ -64,42 +64,90 @@ def get_property_labels(dcid):
...
     observation_period_triple = ""
     if observation_period:
         observation_period_triple = f"""?svObservation observationPeriod {observation_period} ."""
+    date_triple = f"""?svObservation observationDate ?obsDate ."""
+    date_selector = "?obsDate"
+    if date:
+        date_triple = f"""?svObservation observationDate "{date}" ."""
+        date_selector = ""
     sparql_query = '''
```
beets, line 80: this could be a f-string tho:

f"""
SELECT ?dcid {date_selector}
WHERE ...
"""

static/js/browser2/observation_chart.tsx
```diff
@@ -55,44 +55,111 @@ export class ObservationChart extends React.Component<
...
+                  </tr>
+                  {this.sortedDates.map((date) => {
+                    if (date in this.props.sourceSeries.val) {
+                      return (
+                        <tr
+                          className={obsTableRowClass}
+                          key={date}
+                          onClick={() => this.redirectToObsPage(date)}
+                        >
+                          <td width="50%">{date}</td>
```
beets, line 122: it could be 40/60 (or another ratio) since the dates will be fixed length, but stat var names pretty long..

static/js/browser2/statvar_node.tsx
```diff
@@ -71,4 +93,14 @@ export class StatVarNode extends React.Component<
       </Collapsible>
     );
   }
+
+  private onClickStatVarLink = () => {
+    const uri = URI_PREFIX + this.props.statVar.id;
+    window.open(uri);
```
beets, line 99: this triggers a pop-up warning because it's no longer the result of a user action. could you add a todo to improve this? (on suggestion is to save the context of the window, and update the URL when the result comes back).

beets (changes requested):
this looks nice, and the scroll on the observation table is great for our longer time series

mostly small comments

shifucun (approved):
new demo looks very good, the code is also very clean.

server/routes/api/browser.py
chejennifer: done

server/routes/api/browser.py
chejennifer: done

static/js/browser2/observation_chart.tsx
chejennifer: done

static/js/browser2/statvar_node.tsx
chejennifer: I think the pop-up warning was in the redirectToObsPage in observationChart component, so I added a todo there

server/routes/api/browser.py
```diff
@@ -64,42 +64,90 @@ def get_property_labels(dcid):
...
+        return Response(json.dumps("error: must provide a place field"),
+                        400,
+                        mimetype='application/json')
+    stat_var_id = request.args.get("statVar")
+    if not stat_var_id:
+        return Response(json.dumps("error: must provide a statVar field"),
+                        400,
+                        mimetype='application/json')
+    date = request.args.get("date", "")
+    print(date)
```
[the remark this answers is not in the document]
chejennifer: done

beets (approved):
please try to avoid force-pushes once a pr has started so we're able to see old commits and diffs between commits.

PR #884, opened 2021-04-16
[Graph Browser 2] Only Generate Stat Var Hierarchy Search Index in Svobs mode

chejennifer:


shifucun (approved):
thanks for quick fix!

PR #886, opened 2021-04-16
Fix i18n webdriver test

beets:
Update test to  look at *most* of the string, ignoring date which could change.


shifucun:
I wonder if this could be helpful (while it's also annoying..) and worth keeping?

For this case, we have new India data imported which should be ranked lower. And I sent a fix for the ranking: https://github.com/datacommonsorg/mixer/pull/498.

chejennifer (approved):

beets:
ok, i'll close this PR then, and the mixer fix should unblock the tests

PR #887, opened 2021-04-19
[Graph Browser 2] More observation chart updates/fixes

chejennifer:
- Fix bug where if an observation chart has no measurement method or observation period, clicking on a point on the chart may redirect to the wrong observation node. 
    - Fixed by updating the sparql query to return more fields and then filtering on those fields
- Update the table view column name to be the stat var display name instead of the stat var id because the stat var id may be a non human curated string
- Update tooltip to show "N/A" for null values (used to just not show the value)


beets (approved):
nice! this would be a good thing to have a test for in the webdriver, or at least a test exercising one of the sparql related flows.

shifucun (approved):
thanks for the fix!

PR #889, opened 2021-04-20
Clean up ssl setup steps to be consistent with current deployment

shifucun:
This is update on one-time setup and does not have any effect on future deployment.

In current setup, "website-certificate" is Google managed cert and "multi-domain" is the self managed cert. This change is to align the script with actual deployment.


beets (approved):
nit: maybe managed-certificate?
ok to leave as is too, since it's already deployed..

shifucun:
> nit: maybe managed-certificate?
> ok to leave as is too, since it's already deployed..

Ok, we should update when we touch them next time.

PR #890, opened 2021-04-20
[Graph Browser 2] UI updates

chejennifer:
- update page headers
- update each row to show 5 items (instead of 10) when collapsed
- fix weather observations table showing error when clicking on a row (rows for weather observations tables should not be clickable)
- make observation table values look clickable (not just on hover)
- update stat var hierarchy search input box placeholder to be "filter" instead of "search" & add ability to clear input
- filter out observations with measurement method of GoogleKGHumanCurated or HumanCuratedStats
- fix observation charts spinner problem where if we have 2 of the same stat vars expanded in the stat var hierarchy, when expanding the second one, the spinner will appear on the chart that is already expanded 
    - fixed this by giving each observation chart section a randomId

the dev instance has been updated with these changes (eg. https://dev.datacommons.org/browser/geoId/12086)

### DIFF
--- a/static/js/browser2/browser_page.tsx
+++ b/static/js/browser2/browser_page.tsx
@@ -44,18 +44,20 @@ export class BrowserPage extends React.Component<BrowserPagePropType> {
         <div className="browser-page-header">
           {this.props.pageDisplayType === PageDisplayType.PLACE_STAT_VAR && (
             <div className="browser-header-title">
-              <span>Statistical Variable: </span>
+              Statistical Variable:
               <a href={URL_PREFIX + this.props.statVarId}>
-                {this.props.statVarId}
+                {" " + this.props.statVarId}
               </a>
             </div>
           )}
           <div className="browser-header-title">
-            <span>About: </span>
+            About:
             {this.props.pageDisplayType === PageDisplayType.PLACE_STAT_VAR ? (
-              <a href={URL_PREFIX + this.props.dcid}>{this.props.nodeName}</a>
+              <a href={URL_PREFIX + this.props.dcid}>
+                {" " + this.props.nodeName}
+              </a>
             ) : (
-              <span>{this.props.nodeName}</span>
+              <span>{" " + this.props.nodeName}</span>
             )}
           </div>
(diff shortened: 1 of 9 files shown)


static/js/browser2/browser_page.tsx
```diff
@@ -36,12 +38,37 @@ interface BrowserPagePropType {
...
-    const pageTitle = this.getPageTitle();
     const arcDcid = this.getArcDcid();
     return (
       <>
-        <div className="node-about">{"About: " + pageTitle}</div>
-        <div className="node-type">{"type: " + this.props.nodeType}</div>
+        <div className="browser-page-header">
+          {this.props.pageDisplayType === PageDisplayType.PLACE_STAT_VAR && (
+            <div className="browser-header-title">
+              <span>Statistical Variable: </span>
```
beets, line 47: Not all text needs to be wrapped in a span -- e.g. you could leave out the span tags, keep the text and it should render the same.

static/js/browser2/observation_chart.tsx
```diff
@@ -229,14 +230,14 @@ export class ObservationChart extends React.Component<
 
   private loadSpinner(): void {
     document
-      .getElementById(this.svgContainerId() + "-content-area")
+      .getElementById(this.containerId + "-content-area")
```
beets, line 233: maybe use a helper or constant for building that id to reduce future errors?

beets (approved):
thanks for these fixes!

static/js/browser2/browser_page.tsx
chejennifer: done

static/js/browser2/observation_chart.tsx
chejennifer: I use the shared util function randDomId() to create the id

shifucun (approved):

static/js/browser2/observation_chart.tsx
beets: yes, but you're also doing the same string concat multiple times. a shared util that will do that string concat, or saving it in a class var, will help.

beets (approved):

PR #891, opened 2021-04-21
[Graph Browser 2] Fix bug with the stat var hierarchy search

chejennifer:
There was a bug with the stat var hierarchy search where the results returned was an intersection between the results of each of the tokens except for the first token. This happened because the variable used to save all the results for the tokens was set to be the object in the search index for the first token instead of a new object that read in the object in the search index for the first token. Fixed and added a test.


beets (approved):
nice catch!

shifucun (approved):

PR #893, opened 2021-04-22
[Graph Browser 2] weather chart updates

chejennifer:
- add provenance to weather chart
- make weather chart table values not look clickable

Screen Shot 2021-04-21 at 5 13 42 PM

### DIFF
--- a/server/main.py
+++ b/server/main.py
@@ -125,14 +125,15 @@ def get_weather():
     if not prop:
         flask.abort(400, 'Missing url parameter "prop"')
 
-    query_string = ('SELECT ?date ?mean ?unit '
+    query_string = ('SELECT ?date ?mean ?unit ?provId '
                     'WHERE {{'
                     ' ?o typeOf WeatherObservation .'
                     ' ?o observedNode {dcid} .'
                     ' ?o measuredProperty {prop} .'
                     ' ?o observationDate ?date .'
                     ' ?o unit ?unit .'
-                    ' ?o meanValue ?mean .}}').format(dcid=dcid, prop=prop)
+                    ' ?o meanValue ?mean .'
+                    ' ?o provenance ?provId .}}').format(dcid=dcid, prop=prop)
 
     _, rows = dc.query(query_string)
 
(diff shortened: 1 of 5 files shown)


server/main.py
```diff
@@ -125,14 +125,15 @@ def get_weather():
     if not prop:
         flask.abort(400, 'Missing url parameter "prop"')
 
-    query_string = ('SELECT ?date ?mean ?unit '
+    query_string = ('SELECT ?date ?mean ?unit ?provId '
```
beets, line 128: nice find

beets (approved):

static/js/browser2/weather_chart_section.tsx
```diff
@@ -105,11 +114,15 @@ export class WeatherChartSection extends React.Component<
...
+          let provId = "";
           weatherData.forEach((data) => {
             values[data.observationDate] = data.meanValue;
+            if (_.isEmpty(provId)) {
+              provId = data.provId;
+            }
           });
           const sourceSeries = {
-            provenanceDomain: "",
+            provenanceDomain: provId,
```
shifucun, line 125: why is domain assigned with id here? Need some name tweek?

static/js/browser2/arc_section.tsx
```diff
@@ -96,24 +95,13 @@ export class ArcSection extends React.Component<
...
-    Promise.all([labelsPromise, provenancePromise])
-      .then(([labelsData, provenanceData]) => {
-        const provDomain = {};
-        for (const prov of provenanceData) {
-          if (prov["predicate"] === "typeOf" && !!prov["subjectName"]) {
-            provDomain[prov["subjectId"]] = new URL(prov["subjectName"]).host;
-          }
-        }
+      .then((resp) => {
+        const labelsData = resp.data;
```
shifucun, line 101: So now each page only needs to do one API call to fetch the provenance domain?

shifucun (approved):

static/js/browser2/weather_chart_section.tsx
chejennifer: I updated it to be assigned with actual domain

static/js/browser2/arc_section.tsx
chejennifer: We originally only needed to do one API call to fetch the provenance domain for the arcs. But now that we also need the provenance domain for weather charts, I moved fetching the provenance domain to a parent component so that we still only fetch the provenance domain once.

shifucun (approved):

PR #896, opened 2021-04-22
[Graph Browser 2] Place stat var ui updates

chejennifer:
- Update title to STAT_VAR - PLACE_NAME - Graph Browser - Data Commons
- Update properties header to "Statistical Variable Properties"
- Update observations header to "Observations for PLACE_NAME"

Screen Shot 2021-04-22 at 9 18 10 AM

### DIFF
--- a/static/js/browser2/arc_section.tsx
+++ b/static/js/browser2/arc_section.tsx
@@ -24,11 +24,13 @@ import _ from "lodash";
 
 import { InArcSection } from "./in_arc_section";
 import { OutArcSection } from "./out_arc_section";
+import { PageDisplayType } from "./util";
 
 interface ArcSectionPropType {
   dcid: string;
   nodeName: string;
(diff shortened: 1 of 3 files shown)


static/js/browser2/arc_section.tsx
```diff
@@ -24,11 +24,13 @@ import _ from "lodash";
 
 import { InArcSection } from "./in_arc_section";
 import { OutArcSection } from "./out_arc_section";
+import { PageDisplayType } from "./util";
```
beets, line 27: fyi, these should be in a separate module, types.ts? they're not really utility... i have a pr coming up to do that.

beets (approved):

static/js/browser2/arc_section.tsx
chejennifer: Ok, sounds good!

PR #899, opened 2021-04-22
[Graph Browser 2] extract types and related functions to a separate module

beets:
I didn't update a few naming inconsistencies (e.g. Type suffix)


chejennifer (approved):
Thanks for extracting! But looks like you need to update the util tests

beets:
np! updated tests

PR #902, opened 2021-04-23
Use svobs mode in all website deployment config

shifucun:


server/configmodule.py
```diff
@@ -11,10 +11,6 @@ class Config:
     WEBDRIVER = False
     LOCAL = False
     LITE = False
-    # Use StatVarObs instead of PopObs data model.
-    # TODO(shifucun): change this to True after setting svobs_mode=true in mixer
-    # by default.
-    SVOBS = False
```
beets, line 17: Should it be set to True in the default config? There are still a few more references to this var around (e.g. kg_main.html). 

Let's flip it to True by default first, and clean up by deleting it in a follow up (similar to the protocol for an experiment launch).

server/configmodule.py
```diff
@@ -40,11 +36,11 @@ class AutopushConfig(Config):
 
 
 class SvObsConfig(Config):
```
beets, line 38: Is this config class still required?

shifucun: This is for svobs.datacommons.org. Can remove this with the flag after migration?

server/configmodule.py
shifucun: Good catch. Set it to True now

server/configmodule.py
beets: yup, sg

beets (approved):

PR #903, opened 2021-04-23
[Graph Browser 2] add webdriver tests

chejennifer:


server/routes/browser.py
```diff
@@ -29,7 +29,8 @@ def browser_main():
 @bp.route('/<path:dcid>')
 def browser_node(dcid):
     if os.environ.get('FLASK_ENV') == 'svobs' or os.environ.get(
```
beets, line 31: you could also do
```
if os.environ.get('FLAS_ENV') in ['svobs', 'local-svobs', 'webdriver']
```

beets (approved):
Thanks for adding these tests!

We should also add tests for each PageDisplayType. For nybg, we can test that images are able to load.
There should be a basic test for the stat var hierarchy too (maybe exercise a simple search path)
For the charts, maybe a test for clicking on the point? It's possible to search for svg nodes via xpath.

server/routes/browser.py

server/webdriver_tests/browser_test.py
```diff
@@ -186,8 +188,107 @@ def test_page_serve_ca_population(self):
...
+        WebDriverWait(self.driver, self.TIMEOUT_SEC).until(element_present)
+        first_result = self.driver.find_element_by_xpath(
+            '//*[@id="stat-var-hierarchy-section"]/div/div[1]/div[2]/div/div[1]'
+        )
+        first_result.click()
+
+        # Assert that the section Count_Person_Male_AsianAlone opened and shows at least one chart
+        element_present = EC.presence_of_element_located((
+            By.XPATH,
+            '//*[@id="stat-var-hierarchy-section"]/div/div[2]/div/div/div/div/div/div/div/div/div/div/div/div/div/div/div/div/div/div/div/div[11]/div/div/div/div/div/div[1]/div[2]'
```
beets, line 275: this is incredibly brittle... are there classes you could add to make this easier to update in the future (e.g. is-open)?

beets (commented):
thanks for adding the additional tests!

server/webdriver_tests/browser_test.py
chejennifer: done

beets (approved):
thx!

server/webdriver_tests/browser_test.py
beets: i wonder if it could be simplified more, but this is much better

PR #904, opened 2021-04-23
[Graph Browser 2] Stat Var Hierarchy Updates

chejennifer:
- Before rendering browser page, find out if the page has stat var hierarchy by making call to get all stat vars for current dcid
- Move stat var hierarchy section after out arcs section & add loading spinner
- If a node has a "Variables" child stat var group, move that group to the top of the section

dev instance has been updated with these changes (eg. https://dev.datacommons.org/browser/geoId/12086)

### DIFF
--- a/static/js/browser2/browser.ts
+++ b/static/js/browser2/browser.ts
@@ -34,11 +34,15 @@ window.onload = () => {
   if (!_.isEmpty(statVarId)) {
     document.title = `${statVarId} - ${document.title}`;
   }
-  axios
+  const typesPromise = axios
     .get(`/api/browser/propvals/typeOf/${dcid}`)
-    .then((resp) => {
-      const values = resp.data.values;
-      const types = values.out ? values.out : [];
+    .then((resp) => resp.data);
+  const statVarsPromise = axios
+    .get(`/api/place/statsvars/${dcid}`)
+    .then((resp) => resp.data);
+  Promise.all([typesPromise, statVarsPromise])
+    .then(([typesData, statVarsData]) => {
+      const types = typesData.values.out ? typesData.values.out : [];
       const listOfTypes = types.map((type) => type.dcid);
       const nodeType = getNodeType(dcid, listOfTypes, statVarId);
       const pageDisplayType = getPageDisplayType(listOfTypes, statVarId);
@@ -49,6 +53,8 @@ window.onload = () => {
           nodeType,
           pageDisplayType,
           statVarId,
+          shouldShowStatVarHierarchy:
+            !_.isEmpty(statVarsData) && _.isEmpty(statVarId),
         }),
         document.getElementById("node")
       );
@@ -61,6 +67,7 @@ window.onload = () => {
           nodeType: getNodeType(dcid, [], statVarId),
           pageDisplayType: PageDisplayType.GENERAL,
           statVarId,
+          shouldShowStatVarHierarchy: false,
         }),
         document.getElementById("node")
       );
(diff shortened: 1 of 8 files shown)


static/js/browser2/browser.ts
```diff
@@ -34,11 +34,15 @@ window.onload = () => {
...
   }
-  axios
+  const typesPromise = axios
     .get(`/api/browser/propvals/typeOf/${dcid}`)
-    .then((resp) => {
-      const values = resp.data.values;
-      const types = values.out ? values.out : [];
+    .then((resp) => resp.data);
+  const statVarsPromise = axios
+    .get(`/api/place/statsvars/${dcid}`)
```
beets, line 41: although we should not make a new endpoint in the mixer, we should make a new endpoint in flask for this (don't make the client load all that data just for a boolean -- do this on the server)

static/js/browser2/statvar_group_node.tsx
```diff
@@ -226,6 +226,17 @@ export class ChildStatVarGroupSection extends React.Component<
   ChildStatVarGroupSectionPropType
 > {
   render(): JSX.Element {
+    const childStatVarGroups = this.props.data[this.props.statVarGroupId]
```
beets, line 229: since there is only one such node, it may be more efficient to do a linear scan to find the node, and move it up top.

beets (commented):
seems to work well in practice! some suggestions for making it more performant

pradh:
Thanks for making the change!

static/js/browser2/browser.ts
chejennifer: done

static/js/browser2/statvar_group_node.tsx
chejennifer: done

static/js/browser2/browser.ts
```diff
@@ -34,11 +34,15 @@ window.onload = () => {
...
-    .then((resp) => {
-      const values = resp.data.values;
-      const types = values.out ? values.out : [];
+    .then((resp) => resp.data);
+  const numStatVarsPromise = axios
+    .get(`/api/browser/num_stat_vars/${dcid}`)
+    .then((resp) => resp.data);
+  Promise.all([typesPromise, numStatVarsPromise])
+    .then(([typesData, numStatVars]) => {
+      const types = typesData.values.out ? typesData.values.out : [];
```
shifucun, line 45: const types = typesData.values.out || []

static/js/browser2/browser_page.tsx
```diff
@@ -62,7 +69,13 @@ export class BrowserPage extends React.Component<
     if (!this.state.dataFetched) {
       return null;
     }
-
+    const showInArcSection =
+      this.props.pageDisplayType !== PageDisplayType.PLACE_STAT_VAR &&
+      !_.isEmpty(this.state.inLabels);
+    const outArcHeader =
+      this.props.pageDisplayType === PageDisplayType.PLACE_STAT_VAR
+        ? "Statistical Variable Properties"
```
shifucun, line 77: nit: define const string

shifucun (approved):

beets (approved):
thx for the update

static/js/browser2/browser.ts
chejennifer: done

static/js/browser2/browser_page.tsx
chejennifer: done

PR #909, opened 2021-04-27
Update browser to look at svobs config

chejennifer:


server/routes/browser.py
```diff
@@ -15,7 +15,7 @@
 """
 
 import os
```
shifucun, line 17: "import os" can be removed

shifucun (approved):
thanks for making the update!

server/routes/browser.py

PR #910, opened 2021-04-29
[Graph Browser 2] Update in arc header and sparql queries

chejennifer:
- Update in arc header to be "Subject Type: <TYPE>"
- For observation charts, don't fetch observation dcids on render, but instead only retrieve observation dcid when user clicks on an observation
- When loading a page with weather charts, start the weather chart section off collapsed and don't fetch the data unless user opens the section

dev instance updated with these changes (eg. https://dev.datacommons.org/browser/geoId/0649670)[eg. https://dev.datacommons.org/browser/geoId/0649670]


beets (approved):
thanks for the fixes

shifucun (approved):
thanks for the fix!

PR #912, opened 2021-04-30
Website clean up for Stat Var migration

shifucun:


chejennifer (approved):
Thanks for the clean up! 

deploy/git/mixer_hash.txt and deploy/git/website_hash.txt seem to be empty files, so maybe they can be deleted?

shifucun:
> Thanks for the clean up!
> 
> deploy/git/mixer_hash.txt and deploy/git/website_hash.txt seem to be empty files, so maybe they can be deleted?

They are used in the deployment script. In GKE deployment, they are copied over by script but when running minikube locally, the files need to exist, hence adding them here.

shifucun:
Thanks!

PR #916, opened 2021-05-05
API update for scatter tool

chejennifer:
- mixer /stat/collection was updated to stat/set/within-place so make the corresponding changes on flask and client


shifucun (approved):
thanks for updating!

PR #918, opened 2021-05-06
Website server start should check mixer readiness

shifucun:


beets (approved):
thanks for this!

beets:
the two codeql errors look like good candidates for a fix, but they are not related to the PR at all.

pradh (approved):

shifucun:
> the two codeql errors look like good candidates for a fix, but they are not related to the PR at all.

Okay, looks like they are not trivial to fix. Going to look at the codeql errors in batch later.

shifucun:
Thanks for review!

PR #921, opened 2021-05-06
scatter2 updates to api calls

chejennifer:
- use post instead of get to get population data 
- use post instead of get to get stat-set-within-place

### DIFF
--- a/server/routes/api/stats.py
+++ b/server/routes/api/stats.py
@@ -87,7 +87,6 @@ def stats(stats_var):
     if request.method == 'GET':
         place_dcids = request.args.getlist('dcid')
     result = get_stats_wrapper('^'.join(place_dcids), stats_var)
-    print(result)
     return Response(result, 200, mimetype='application/json')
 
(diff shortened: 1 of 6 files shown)


server/routes/api/stats.py
```diff
@@ -81,8 +81,13 @@ def stats(stats_var):
...
         with value to be the observation time series.
     """
-    place_dcids = request.args.getlist('dcid')
+    place_dcids = []
+    if request.method == 'POST':
+        place_dcids = request.json.get('dcid', [])
+    if request.method == 'GET':
+        place_dcids = request.args.getlist('dcid')
     result = get_stats_wrapper('^'.join(place_dcids), stats_var)
+    print(result)
```
shifucun, line 90: nit: remove print

shifucun (approved):

server/routes/api/stats.py
chejennifer: done

PR #927, opened 2021-05-07
Various updates for scatter2

chejennifer:
- add examples to info page
- show directions to pick stat vars after user has picked place and place type
- fix the error message that flashes right before chart loads
- fix a bug where if per capita is chosen but the data is unavailable, options section and chart section both disappears so user can't change their selected options
- fix the filter by population input boxes where it wouldn't let user clear the input
- update place search input so that it doesn't mislead users into thinking they can search for additional places once they've chosen a place

updated dev instance

### DIFF
--- a/static/js/tools/scatter2/info.tsx
+++ b/static/js/tools/scatter2/info.tsx
@@ -37,9 +37,96 @@ function Info(): JSX.Element {
           to choose from, arranged in a topical hierarchy.
         </li>
       </ol>
-
-      {/* TODO(intrepiditee): Add descriptions examples */}
-
+      <p>Or you can start your exploration from these interesting points ...</p>
+      <ul>
+        <li>
+          <b>Asians Per Capita vs Median Income</b> for counties in
+          <a
+            href={
+              "#%26svx%3DMedian_Income_Person%26svpx%3D0-3%26svnx%3DMedian%20income%26svy%3DCount_Person_AsianAlone%26svpy%3D0-14-1%26svdy%3DCount_Person%26svny%3DAsian%20Alone%26pcy%3D1%26epd%3DgeoId%2F06%26epn%3DCalifornia%26ept%3DCounty"
+            }
+          >
+            {" "}
+            California
+          </a>
+          ,
+          <a
+            href={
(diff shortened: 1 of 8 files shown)


static/js/tools/scatter2/info.tsx
```diff
@@ -37,9 +37,96 @@ function Info(): JSX.Element {
...
-
-      {/* TODO(intrepiditee): Add descriptions examples */}
-
+      <p>Or you can start your exploration from these interesting points ...</p>
+      <ul>
+        <li>
+          <b>Asians Per Capita vs Median Income</b> for counties in
+          <a
+            href={
+              "#%26svx%3DMedian_Income_Person%26svpx%3D0-3%26svnx%3DMedian%20income%26svy%3DCount_Person_AsianAlone%26svpy%3D0-14-1%26svdy%3DCount_Person%26svny%3DAsian%20Alone%26pcy%3D1%26epd%3DgeoId%2F06%26epn%3DCalifornia%26ept%3DCounty"
```
shifucun, line 46: I think you can use the regular (no-encoded) chars in the url here. That would be easy for maintain.

static/js/tools/scatter2/info.tsx
```diff
@@ -37,9 +37,96 @@ function Info(): JSX.Element {
           to choose from, arranged in a topical hierarchy.
         </li>
       </ol>
-
-      {/* TODO(intrepiditee): Add descriptions examples */}
-
+      <p>Or you can start your exploration from these interesting points ...</p>
```
shifucun, line 40: We should run these examples with the teams too after it's pushed to auotpush.

shifucun (approved):
Thanks for the update!

static/js/tools/scatter2/info.tsx
chejennifer: done

static/js/tools/scatter2/info.tsx
chejennifer: yeup, sounds good. It's also already on dev

static/js/tools/scatter2/info.tsx
```diff
@@ -37,9 +37,96 @@ function Info(): JSX.Element {
...
+      <ul>
+        <li>
+          <b>Asians Per Capita vs Median Income</b> for counties in
+          <a
+            href={
+              "#&svx=Median_Income_Person&svpx=0-3&svnx=Median_income&svy=Count_Person_AsianAlone&svpy=0-14-1&svdy=Count_Person&svny=Asian_Alone&pcy=1&epd=geoId%2F06&epn=California&ept=County"
+            }
+          >
+            {" "}
```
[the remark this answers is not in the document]
chejennifer: done

PR #930, opened 2021-05-10
Update get_statvar_groups() to take multiple places

shifucun:
This is a follow up of https://github.com/datacommonsorg/mixer/pull/519.

Also did some format update.

### DIFF
--- a/server/lib/statvar_hierarchy_search.py
+++ b/server/lib/statvar_hierarchy_search.py
@@ -107,7 +107,7 @@ def get_statvar_search_index() -> Dict[str, Dict[str, RankingInformation]]:
     },
     ...
     """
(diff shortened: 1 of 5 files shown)


chejennifer (approved):

beets (approved):
thanks!

beets:
webdriver is picking up a new, related, error - it's a browser test for stat var hierarchy search.

shifucun:
> webdriver is picking up a new, related, error - it's a browser test for stat var hierarchy search. so this is a temporary disablement until we push the update to staging.
> 
> we should update webdriver to test against autopush.api in the meantime?

I agree. Use autopush for webdriver test makes sense.

PR #933, opened 2021-05-11
Add back cpu/mem limit to avoid pod eviction

shifucun:
More context in https://github.com/datacommonsorg/mixer/pull/520

### DIFF
--- a/deploy/base/deployment.yaml
+++ b/deploy/base/deployment.yaml
@@ -57,6 +57,13 @@ spec:
               port: 8080
             failureThreshold: 1
             periodSeconds: 10
+          resources:
+            limits:
+              memory: "2G"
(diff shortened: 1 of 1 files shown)


deploy/base/deployment.yaml
```diff
@@ -97,6 +104,13 @@ spec:
         - name: mixer
           image: gcr.io/datcom-ci/datacommons-mixer:latest
           imagePullPolicy: Always
+          resources:
```
pradh, line 107: What fraction of the total node mem-size does this represent?

pradh (approved):

deploy/base/deployment.yaml
shifucun: Ok. Sg

PR #934, opened 2021-05-11
Clean up unused json data files and util exports

shifucun:
hierarchy.json and placeObject.json was used in scatter tool.

most of the util.js functions are not used anymore, removing them and convert the file as .ts


chejennifer (approved):
thanks for the cleanup!

shifucun:
Thanks!

PR #937, opened 2021-05-11
Bump memory limit and reduce cpu limit

shifucun:
CPU limit is over what can be scheduled. Apparently need to reserve cpu for non application resources


shifucun:
thanks for review!

PR #948, opened 2021-05-21
stat var hierarchy bugs

chejennifer:
fixed bugs:
- when section is expanded and user selects a search result within that section, result won't show up.
- when user clears search result, the expanded section (from the selected search result) is rendered and then closed

screen-recording (16)

### DIFF
--- a/static/js/browser2/statvar_group_node.tsx
+++ b/static/js/browser2/statvar_group_node.tsx
@@ -57,7 +57,7 @@ interface StatVarGroupNodePropType {
 }
 
 interface StatVarGroupNodeStateType {
-  isRendered: boolean;
+  toggledOpen: boolean;
 }
 
 export class StatVarGroupNode extends React.Component<
@@ -69,7 +69,7 @@ export class StatVarGroupNode extends React.Component<
   constructor(props: StatVarGroupNodePropType) {
     super(props);
     this.state = {
-      isRendered: this.props.open,
+      toggledOpen: false,
     };
     this.highlightedStatVar = React.createRef();
     this.scrollToHighlighted = this.scrollToHighlighted.bind(this);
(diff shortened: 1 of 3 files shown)


static/js/browser2/statvar_group_node.tsx
```diff
@@ -53,7 +53,7 @@ interface StatVarGroupNodePropType {
 }
 
 interface StatVarGroupNodeStateType {
-  isRendered: boolean;
+  toggledOpen: boolean;
```
beets, line 56: add a comment here? specifically how this relates to the prop `open`. maybe the prop should be renamed to something like `startsOpened`

beets (approved):
thanks for the bug fix

static/js/browser2/statvar_group_node.tsx
chejennifer: done

static/js/browser2/statvar_hierarchy.tsx
```diff
@@ -86,7 +89,16 @@ export class StatVarHierarchy extends React.Component<
...
 
+  componentDidUpdate(): void {
+    if (this.state.searchSelectionCleared) {
+      this.setState({ searchSelectionCleared: false });
+    }
+  }
+
   render(): JSX.Element {
+    if (this.state.searchSelectionCleared) {
+      return null;
```
shifucun, line 100: why return null here? add some comments?

shifucun (commented):
Tricky states management! Can you also help check the hierarchy in /dev works as intended.

static/js/browser2/statvar_hierarchy.tsx
chejennifer: done.

chejennifer:
> Tricky states management! Can you also help check the hierarchy in /dev works as intended.

Yeup, I have checked /dev

shifucun (approved):

shifucun:
> > Tricky states management! Can you also help check the hierarchy in /dev works as intended.
> 
> Yeup, I have checked /dev

Thanks!

PR #954, opened 2021-06-02
rename scatter and browser folders

chejennifer:
scatter2 -> scatter 
browser2 -> browser


beets (approved):
thanks for the update!

PR #955, opened 2021-06-07
Update landing page to use new API format

shifucun:


beets (approved):
thanks for the update!

shifucun:
Thanks! Also updated how additional stat vars are fetched. Made it part of the config, which defaults to empty

PR #957, opened 2021-06-09
Update for place explorer choropleth endpoints

chejennifer:


server/routes/api/chart.py
```diff
@@ -213,48 +210,74 @@ def choropleth_data(dcid):
...
     result = {}
-    for cc in choropleth_configs:
+    for cc in configs:
         # we should only be making choropleths for configs with a single stat var
         sv = cc['statsVars'][0]
-        cc_data, statvar_denom = landing_page_api.get_snapshot_across_places(
-            cc, processed_data, geos)
-        data_values = cc_data.get('data', [])
-        data_dict = dict()
+        cc_sv_data_values = sv_data.get(sv, {}).get('stat', {})
```
shifucun, line 234: Would be helpful to abstract a few functions out of this for loop, so it's easy to follow. 

Also could add comments before each sections to explain what it does. It's a bit overwhelming for someone new to the code now :)

shifucun (commented):
Thanks for the quick update!

server/routes/api/chart.py
```diff
@@ -213,48 +210,74 @@ def choropleth_data(dcid):
...
+            value as object with date,dictionary of place: value, number of
+            data points, exploreUrl, list of sources
     """
-    all_stat_vars, choropleth_configs = get_choropleth_sv()
-    geos, _ = get_choropleth_places(dcid)
-    if not all_stat_vars or not geos:
+    configs = get_choropleth_configs()
+    stat_vars, denoms = get_stat_vars(configs)
+    display_dcid, display_level = get_choropleth_display_level(dcid)
+    geos = dc_service.get_places_in([display_dcid],
```
shifucun, line 219: Do you see this approach faster than before?

server/routes/api/chart.py
```diff
@@ -147,61 +138,67 @@ def geojson(dcid):
...
-                    most_recent_date = sorted_dates[-1]
-                    result[geo][sv] = {
-                        'data': source_values,
-                        'provenanceUrl': source.get('provenanceUrl', '')
-                    }
-    return result
+    stat_vars = set()
+    denoms = set()
+    for config in configs:
+        stat_vars.add(config['statsVars'][0])
```
beets, line 170: check for existence of, and length of 'statsVars' first? or use .get so there is a default value to fall back on

server/routes/api/chart.py
```diff
@@ -147,61 +138,67 @@ def geojson(dcid):
...
+        if 'relatedChart' in config and config['relatedChart'].get(
+                'scale', False):
+            denoms.add(config['relatedChart'].get('denominator',
+                                                  'Count_Person'))
+        if 'denominator' in config:
+            denoms.add(config['denominator'][0])
```
beets, line 176: ditto

server/routes/api/chart.py
```diff
@@ -213,48 +210,74 @@ def choropleth_data(dcid):
...
+            import_name = dcid_sv_data.get("metadata", {}).get("importName", "")
+            source = cc_sv_data_metadata.get(import_name,
+                                             {}).get("provenanceUrl", "")
+            sources.add(source)
+            if denom:
+                if not cc_denom_data.get(place_dcid, {}):
+                    continue
+                denom_val = get_denom_val(
+                    date, cc_denom_data[place_dcid].get('data', {}))
+                val = val / denom_val
```
beets, line 260: check for value of denom_val before dividing?

server/routes/api/chart.py
```diff
@@ -213,48 +210,74 @@ def choropleth_data(dcid):
...
+                sources.add(cc_denom_data[place_dcid].get('provenanceUrl', ""))
+            val = val * scaling
+            data_dict[place_dcid] = val
         is_scaled = (('relatedChart' in cc and
                       cc['relatedChart'].get('scale', False)) or
                      ('denominator' in cc))
-        exploreUrl = landing_page_api.build_url([dcid], statvar_denom,
-                                                is_scaled)
+        exploreUrl = landing_page_api.build_url([dcid], {sv: denom}, is_scaled)
+        dates = filter(lambda x: x != "", dates)
```
beets, line 268: interesting -- we shouldn't be returning blank dates or sources...

beets (commented):
mostly found some error handling we should do

server/routes/api/chart.py
```diff
@@ -46,16 +46,16 @@
 
 
 @cache.memoize(timeout=3600 * 24)  # Cache for one day.
-def get_choropleth_places(geoDcid):
-    """ Get the list of places to show on a choropleth chart for a given place
+def get_choropleth_display_level(geoDcid):
```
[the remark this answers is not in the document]
chejennifer: done. I think in this case, we want something general because the geoDcid isn't necessarily a parent or child, it's just the dcid of the place we're trying to get this information for.

server/routes/api/chart.py
```diff
@@ -106,10 +93,14 @@ def geojson(dcid):
     """
     Get geoJson data for a given place
```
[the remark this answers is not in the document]
chejennifer: done

server/routes/api/chart.py
```diff
@@ -213,48 +210,74 @@ def choropleth_data(dcid):
     API Returns:
         Dictionary with
             key as stat var
-            value as object with date,dictionary of place: value, number of data points, exploreUrl, list of sources
+            value as object with date,dictionary of place: value, number of
```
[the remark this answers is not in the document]
chejennifer: done

server/routes/api/chart.py
```diff
@@ -213,48 +210,74 @@ def choropleth_data(dcid):
...
-    all_sv_data = dc_service.get_stats_all(geos, list(all_stat_vars))
-    if not 'placeData' in all_sv_data:
-        return Response(json.dumps({}), 200, mimetype='application/json')
-    processed_data = process_choropleth_data(all_sv_data['placeData'])
-
+    sv_data = dc_service.get_stat_set_within_place(display_dcid, display_level,
+                                                   list(stat_vars),
+                                                   "").get('data', {})
+    denoms_data = {}
+    for denom in denoms:
```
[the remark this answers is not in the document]
chejennifer: For denom data, we want data from multiple dates so that we can find the date that best matches the stat var. Therefore, I don't think we should use get_stat_set_within_place

server/routes/api/chart.py
chejennifer: done

server/routes/api/chart.py
chejennifer: I actually haven't noticed it being faster

server/routes/api/chart.py
```diff
@@ -106,10 +93,14 @@ def geojson(dcid):
...
     """
-    geos, geojson_prop = get_choropleth_places(dcid)
+    display_dcid, display_level = get_choropleth_display_level(dcid)
+    geos = dc_service.get_places_in([display_dcid],
+                                    display_level).get(display_dcid, [])
     if not geos:
         return Response(json.dumps({}), 200, mimetype='application/json')
-
+    geojson_prop = CHOROPLETH_GEOJSON_PROPERTY_MAP.get(display_level, "")
+    if dcid == 'geoId/72':
```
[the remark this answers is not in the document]
chejennifer: done

server/routes/api/chart.py
chejennifer: done

server/routes/api/chart.py
chejennifer: done

server/routes/api/chart.py
chejennifer: done

server/routes/api/chart.py
chejennifer: you're right, I don't think it should, but I put that just in case there was something weird in the data .. should I remove it?

server/routes/api/chart.py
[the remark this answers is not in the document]
shifucun: I see, and this is where the complexity comes from

server/routes/api/chart.py
shifucun: Thanks for the update!

shifucun (approved):

beets (approved):
thanks for the updates!

PR #958, opened 2021-06-11
[map] add per capita and breadcrumbs

chejennifer:
added new ChartOptions component that contains the per capita toggle and breadcrumbs

dev is updated with these changes (eg. http://dev.datacommons.org/tools/map#%26sv%3DCount_Person_EducationalAttainmentBachelorsDegree%26pc%3D1)

### DIFF
--- a/static/js/tools/map/chart_loader.tsx
+++ b/static/js/tools/map/chart_loader.tsx
@@ -24,6 +24,7 @@ import _ from "lodash";
 import { GeoJsonData } from "../../chart/types";
 import {
   getPopulationDate,
+  getUnit,
   PlacePointStat,
   SourceSeries,
 } from "../shared_util";
@@ -31,27 +32,30 @@ import { Context, IsLoadingWrapper, PlaceInfo, StatVarInfo } from "./context";
 import { Chart } from "./chart";
 import axios from "axios";
 
-interface MapRawData {
+interface ChartRawData {
   geoJsonData: GeoJsonData;
   statVarData: PlacePointStat;
   populationData: { [dcid: string]: SourceSeries };
 }
 
 interface ChartData {
-  dataValues: { [dcid: string]: number };
+  mapDataValues: { [dcid: string]: number };
+  breadcrumbDataValues: { [dcid: string]: number };
   sources: Set<string>;
   statVarDates: { [dcid: string]: string };
   geoJsonData: GeoJsonData;
+  unit: string;
 }
 
 export function ChartLoader(): JSX.Element {
   const { placeInfo, statVarInfo, isLoading } = useContext(Context);
-  const [rawData, setRawData] = useState<MapRawData | undefined>(undefined);
+  const [rawData, setRawData] = useState<ChartRawData | undefined>(undefined);
   const [chartData, setChartData] = useState<ChartData | undefined>(undefined);
   useEffect(() => {
     const placesLoaded =
       !_.isEmpty(placeInfo.value.enclosingPlace.dcid) &&
-      !_.isEmpty(placeInfo.value.enclosedPlaces);
+      !_.isEmpty(placeInfo.value.enclosedPlaces) &&
+      !_.isNull(placeInfo.value.parentPlaces);
     if (placesLoaded && !_.isEmpty(statVarInfo.value.statVar)) {
       fetchData(placeInfo.value, statVarInfo.value, isLoading, setRawData);
     } else {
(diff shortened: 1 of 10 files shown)


static/js/tools/map/chart_loader.tsx
```diff
@@ -145,29 +157,47 @@ function loadChartData(
...
   for (const dcid in statVarData.stat) {
     if (_.isEmpty(statVarData.stat[dcid])) {
       continue;
     }
-    const statVarValue = statVarData.stat[dcid].value;
-    const importName = statVarData.stat[dcid].metadata.importName;
-    sources.add(statVarData.metadata[importName].provenanceUrl);
-    statVarDates[dcid] = statVarData.stat[dcid].date;
+    let statVarValue = statVarData.stat[dcid].value;
     if (isPerCapita && dcid in populationData) {
```
beets, line 172: this could be more straightforward to read if it was structured:

```
if (isPerCapita) {
  if (dcid in population) {
  ...
  } else {
  ...
  }
}
```

static/js/tools/map/chart_loader.tsx
```diff
@@ -145,29 +157,47 @@ function loadChartData(
...
       const popDate = getPopulationDate(
         populationData[dcid],
         statVarData.stat[dcid]
       );
-      dataValues[dcid] = statVarValue / populationData[dcid].data[popDate];
+      statVarValue = statVarValue / populationData[dcid].data[popDate];
```
beets, line 177: pls add check for div by zero

static/js/tools/map/chart_loader.tsx
```diff
@@ -145,29 +157,47 @@ function loadChartData(
...
       sources.add(populationData[dcid].provenanceUrl);
-    } else if (!isPerCapita) {
-      dataValues[dcid] = statVarValue;
+    } else if (isPerCapita && !(dcid in populationData)) {
+      continue;
```
beets, line 180: this case should be handled and shown as a data unavailable (though the hover can still show the unadjusted values)

static/js/tools/map/chart_options.tsx
```diff
@@ -0,0 +1,113 @@
...
+  MAP_REDIRECT_PREFIX,
+} from "./util";
+
+interface ChartOptionsPropType {
+  dataValues: { [dcid: string]: number };
+  placeInfo: PlaceInfo;
+  statVarDates: { [dcid: string]: string };
+  unit: string;
+}
+export function ChartOptions(props: ChartOptionsPropType): JSX.Element {
```
beets, line 35: please add a comment for each component

static/js/tools/map/context.ts
```diff
@@ -28,6 +28,12 @@ interface Setter<T> {
   (value: T): void;
 }
 
+export interface ParentPlace {
```
beets, line 31: please add comments for shared interfaces

beets (approved):
please keep the major components / interfaces documented as you go along. otherwise lg

static/js/tools/map/chart_loader.tsx
chejennifer: done

static/js/tools/map/chart_loader.tsx
chejennifer: when drawing the map, if there are no values for a certain place dcid, it will show as "data missing". I think it would be confusing for the user if the tooltip showed the unadjusted values

static/js/tools/map/chart_options.tsx
chejennifer: done

static/js/tools/map/context.ts
chejennifer: done

static/js/tools/map/chart_loader.tsx
```diff
@@ -169,15 +169,21 @@ function loadChartData(
...
+    if (isPerCapita) {
+      if (dcid in populationData) {
+        const popDate = getPopulationDate(
+          populationData[dcid],
+          statVarData.stat[dcid]
+        );
+        const popValue = populationData[dcid].data[popDate];
+        statVarValue =
+          popValue === 0
+            ? statVarValue
```
beets, line 181: sorry, on second thought - population = 0 should be treated the same way. they're both invalid values.

so we could display it the same as "data missing", but perhaps in the tooltip, we should also say 'missing population data'.

we might want to differentiate between these "data missing" and "data invalid" states as well - ok to mark this as a todo for now.

beets (approved):

static/js/tools/map/chart_loader.tsx
chejennifer: done. For now, the tooltip content can't be customized to say "missing population data", but that is the next thing I will be working on (tooltip content to be customizable). And added a todo for differentiating between "data missing" and "data invalid".

PR #962, opened 2021-06-15
Increase Flask worker and GKE load balancer timeout to 60s

shifucun:
The default timeout is 30s, which prevents a large stat-var-hierarchy cache to serve.

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -36,4 +36,4 @@ COPY server/. /workspace/server/
 
 # Run server
 WORKDIR /workspace/server
-CMD if [ "$ENV" = "local" ] ; then python3.7 main.py ; else exec gunicorn --bind :8080 main:app ; fi
+CMD if [ "$ENV" = "local" ] ; then python3.7 main.py ; else exec gunicorn --timeout 60 --bind :8080 main:app ; fi
--- a/gke/mcs.yaml
+++ b/gke/mcs.yaml
@@ -18,6 +18,8 @@ metadata:
   name: website-mcs
   namespace: website
   annotations:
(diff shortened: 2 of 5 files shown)


gke/mcs.yaml
```diff
@@ -17,6 +17,8 @@ kind: MultiClusterService
 metadata:
   name: website-mcs
   namespace: website
+  annotations:
+    cloud.google.com/backend-config: '{"ports": {"8080":"backendconfig"}}'
```
pradh, line 21: This is supposed to refer to the yaml? worth commenting so?

build/Dockerfile
```diff
@@ -36,4 +36,4 @@ COPY server/. /workspace/server/
 
 # Run server
 WORKDIR /workspace/server
-CMD if [ "$ENV" = "local" ] ; then python3.7 main.py ; else exec gunicorn --bind :8080 main:app ; fi
\ No newline at end of file
+CMD if [ "$ENV" = "local" ] ; then python3.7 main.py ; else exec gunicorn --timeout 60 --bind :8080 main:app ; fi
```
pradh, line 39: for my understanding, is the 60s in the yaml for prod and this one for local/dev?

pradh (approved):

gke/mcs.yaml
shifucun: Done

build/Dockerfile
shifucun: This is for Kubernetes deployment, ie when "ENV" != "local". So this applies to prod/dev

Local instance will use "python3.7 main.py" which does not have a timeout.

shifucun (commented):
Thanks for review!

PR #964, opened 2021-06-16
Place Page Choropleth Fixes

chejennifer:
- for date range in title, use "[start date] to [end date]" instead of "[start date] - [end date]" (eg. "2020-03 to 2020-04" vs "2020-03-2020-04")
- the function for getting tooltip html for place pages had a bug causing it to always show "data missing"

### DIFF
--- a/server/routes/api/chart.py
+++ b/server/routes/api/chart.py
@@ -262,7 +262,7 @@ def get_date_range(dates):
     sorted_dates_list = sorted(list(dates))
     date_range = sorted_dates_list[0]
     if len(sorted_dates_list) > 1:
-        date_range = f'{sorted_dates_list[0]} - {sorted_dates_list[-1]}'
+        date_range = f'{sorted_dates_list[0]} to {sorted_dates_list[-1]}'
     return date_range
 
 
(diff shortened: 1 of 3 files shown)


shifucun (approved):

server/routes/api/chart.py
```diff
@@ -262,7 +262,7 @@ def get_date_range(dates):
     sorted_dates_list = sorted(list(dates))
     date_range = sorted_dates_list[0]
     if len(sorted_dates_list) > 1:
-        date_range = f'{sorted_dates_list[0]} - {sorted_dates_list[-1]}'
+        date_range = f'{sorted_dates_list[0]} to {sorted_dates_list[-1]}'
```
beets, line 265: is this change strictly necessary? this would require translations for place pages.

how about the [endash](https://www.thepunctuationguide.com/en-dash.html) instead?

beets (approved):

server/routes/api/chart.py
chejennifer: This isn't necessary but it just might look a little confusing if there is a date range like "2020-03-2020-04". However, there usually aren't date ranges on the place pages so I think I'll just use the endash instead.

PR #966, opened 2021-06-16
Add check when getting date range for choropleth charts

chejennifer:
There was a case where the list of dates was empty which caused a list index out of range error. This check will prevent future errors like this when getting date range.


shifucun (approved):
Thanks for the fix!

PR #968, opened 2021-06-17
[map] leaf pages

chejennifer:
- For leaf pages, zoom in on the place of interest 

dev has been updated (http://dev.datacommons.org/tools/map#%26sv%3DCumulativeCount_MedicalConditionIncident_COVID_19_PatientDeceased%26pc%3D1)
Screen Shot 2021-06-16 at 10 42 04 PM
Screen Shot 2021-06-16 at 10 42 21 PM


static/js/chart/draw_choropleth.ts
```diff
@@ -98,23 +104,46 @@ function drawChoropleth(
...
+    if (geoJsonFeature) {
+      fitSize(
+        chartWidth - legendWidth,
+        chartHeight,
+        geoJsonFeature,
+        projection,
+        geomap,
+        ZOOMED_SCALE_AMOUNT
+      );
+    }
```
beets, line 136: how should we handle the case where a feature wasn't found?

beets (commented):
lg! just one error handling case to handle..

static/js/chart/draw_choropleth.ts
chejennifer: I guess when a feature isn't found, we can just fit it like we were drawing the map without zooming in?

beets:
yeah, that sounds good.

static/js/chart/draw_choropleth.ts
chejennifer: Or maybe we could have an error message that says "Sorry, unable to draw the map for this place"

beets:
either way works. we can start with an error message - that would be less
confusing when it does happen.

static/js/chart/draw_choropleth.ts
chejennifer: or we can have an error message that says something like "Sorry, unable to draw map for this place"

chejennifer: done

beets (approved):
lg - thanks for the update

PR #969, opened 2021-06-17
Add some checks

chejennifer:
Add some checks to prevent the following errors that have been caught in the last few days: 
- KeyError 'name' in api_ranking
- KeyError 'statVars' in get_stat_vars_union
- bad mixer request in geojson


beets (approved):
this is great - thanks for these fixes!

shifucun (approved):
thanks for the fix!

PR #972, opened 2021-06-18
Increase wait time for mixer start up

shifucun:
Mixer now takes longer time to start due to the increased size of stat var group cache

### DIFF
--- a/server/__init__.py
+++ b/server/__init__.py
@@ -121,7 +121,7 @@ def create_app():
     app.config['BABEL_TRANSLATION_DIRECTORIES'] = 'i18n'
 
     if not cfg.TEST:
-        timeout = 30  # seconds
+        timeout = 120  # seconds
         counter = 0
         isOpen = False
         while not isOpen:


beets (approved):
wow, 4x

pradh (approved):

shifucun:
> wow, 4x

Give some buffer...

pradh:
To confirm, does this affect the clients in any way? For e.g., GKE thinks it has rolled out everything and stops previous versions, but Mixer is still not ready... or will GKE think it has rolled out *after* mixer loads up the full cache?

shifucun:
> To confirm, does this affect the clients in any way? For e.g., GKE thinks it has rolled out everything and stops previous versions, but Mixer is still not ready... or will GKE think it has rolled out _after_ mixer loads up the full cache?

Only when all the containers in a pod started, the pod is in ready state and can serve. So only when mixer is ready, and then website is ready, the pod is in ready state.

PR #973, opened 2021-06-18
[map] zoom in on map

chejennifer:
-added ability to zoom and pan. To zoom in and out, can use buttons or mouse wheel

dev updated (http://dev.datacommons.org/tools/map#%26sv%3DCumulativeCount_MedicalConditionIncident_COVID_19_PatientDeceased%26pc%3D1)

screen-recording (17)

### DIFF
--- a/static/js/chart/draw_choropleth.ts
+++ b/static/js/chart/draw_choropleth.ts
@@ -68,6 +68,24 @@ function fitSize(
   projection.scale(s).translate([translateX, translateY]);
 }
 
+/** Draws a choropleth chart
+ *
+ * @param containerId id of the div to draw the choropleth in
+ * @param geoJson the geojson data for drawing choropleth
+ * @param chartHeight height for the chart
+ * @param chartWidth width for the chart
+ * @param dataValues data values for plotting
+ * @param unit the unit of measurement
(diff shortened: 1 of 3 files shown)


static/js/chart/draw_choropleth.ts
```diff
@@ -81,7 +81,9 @@ function drawChoropleth(
   canClick: boolean,
   getRedirectLink: (geoDcid: GeoJsonFeatureProperties) => string,
   getTooltipHtml: (place: NamedPlace) => string,
-  zoomDcid?: string
+  zoomDcid?: string,
```
beets, line 84: please document this growing list of parameters :)

beets (approved):
very very cool!

i think the changes aren't on dev anymore, but looking at the video - would be nice if hover is disabled during the zoom transitions.

static/js/chart/draw_choropleth.ts
chejennifer: done

chejennifer:
> very very cool!
> 
> i think the changes aren't on dev anymore, but looking at the video - would be nice if hover is disabled during the zoom transitions.

Thanks! I will try to disable hover during zoom transitions in a separate pr

PR #981, opened 2021-06-23
[map] resize legend separately from the map

chejennifer:
Separate how window resize is handled for the legend from how it's handled for the map. Otherwise, going from small screen to large screen will cause the legend to become too large. 
screen-recording (18)


beets (approved):
nice refactor!

PR #984, opened 2021-06-24
Move stat var hierarchy componenets into a separate folder

shifucun:
As this will be a shared UI widget across tools.

Change file name from statvar_* to stat_var_* to be consistent with our code naming


beets (approved):
nice!

PR #986, opened 2021-06-24
[Naming Clean Up] Change statsVar to statVar in timeline tool

shifucun:


beets (approved):
nice - long awaited change!

we have lots of places to update (including bookmarks, python), so we should skip the URL updates since that seems unnecessarily risky at this point

shifucun:
> nice - long awaited change!
> 
> we have lots of places to update (including bookmarks, python), so we should skip the URL updates since that seems unnecessarily risky at this point

Good point, reverted url changes

PR #990, opened 2021-06-25
Update website GKE deployment config

shifucun:
- Relax liveness/readiness prober time
- Add bounce config


deploy/base/deployment.yaml
```diff
@@ -101,6 +101,11 @@ spec:
                 configMapKeyRef:
                   name: store-configmap
                   key: bigtable.version
+            - name: BOUNCE
```
beets, line 104: what does this do? is it possible to add a comment here?

shifucun: Done

beets (approved):

PR #995, opened 2021-06-29
[map] Few updates

chejennifer:
- add examples to the landing page
- add a tooltip beside the title explaining the date range
- disable per capita checkbox for medianValue stat vars

dev instance updated (https://dev.datacommons.org/tools/map)

Screen Shot 2021-06-29 at 9 14 10 AM

### DIFF
--- a/static/js/tools/map/chart_options.tsx
+++ b/static/js/tools/map/chart_options.tsx
@@ -30,6 +30,8 @@ import {
 } from "./util";
 import { DataPointMetadata } from "./chart_loader";
 
+const NO_PER_CAPITA_TYPES = ["medianValue"];
+
 interface ChartOptionsPropType {
   dataValues: { [dcid: string]: number };
   placeInfo: PlaceInfo;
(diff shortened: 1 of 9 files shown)


static/js/tools/map/info.tsx
```diff
@@ -39,6 +39,170 @@ export function Info(): JSX.Element {
...
+          <a
+            href={
+              "#&sv=Count_Person_EducationalAttainmentBachelorsDegreeOrHigher&pc=1&pd=geoId/08&pn=Colorado&pt=State&ept=County"
+            }
+          >
+            Colorado
+          </a>
+        </li>
+        <li>
+          <b>Foreign Born</b> across counties in{" "}
```
beets, line 146: maybe we should skip this since it doesn't have a nicely formatted title.

i noticed these examples are mostly from demographics. how about UnemploymentRate?

static/js/tools/map/chart_options.tsx
```diff
@@ -30,6 +30,8 @@ import {
 } from "./util";
 import { DataPointMetadata } from "./chart_loader";
 
+const NO_PER_CAPITA_TYPES = ["medianValue"];
```
beets, line 33: most things with "FractionOf" or "Rate" are good candidates to add here too.

beets (approved):
looks great!

static/js/tools/map/chart_options.tsx
chejennifer: I don't think FractionOf or Rate are values for the stat type predicate though .. I checked a few statistical variables that are rates or fractionOf, but their statType is just "measuredValue"

static/js/tools/map/info.tsx
chejennifer: done

static/js/tools/map/chart_options.tsx
beets: gotcha - right, those are baked into the sv dcids.

PR #1000, opened 2021-06-29
[place] fix chart geojson call for place pages

beets:
the [import of country boundaries](https://github.com/datacommonsorg/data/pull/448) includes AA1 territories of the US, which is now returned with geojsons. unfortunately, the import included the full geojson, not just the coordinates which results in key errors in the API call.

temporarily fixing so we specifically look for State children for US


chejennifer (approved):
Thanks for the fix!

shifucun:
Congrats for PR 1000 :)

PR #1001, opened 2021-06-29
Add stat var count in hierarchy UI

shifucun:
* Show descendent stat var count as super script 
* Gray out nodes that have no descendent stat vars
* Disable input when stat var has no data.

image

image


chejennifer (approved):
nice!

static/css/tools/stat_var_widget.scss
```diff
@@ -97,3 +97,7 @@ $transparent-red: rgba(240, 230, 231, 0.2);
   cursor: pointer;
   background-color: $transparent-gray;
 }
+
+#hierarchy-section .node-no-data {
```
chejennifer, line 101: looks like .node-no-data with opacity 30% is also set in stat_var_hierarchy.tsx. Does it need to be set in both places?

chejennifer (approved):

static/css/tools/stat_var_widget.scss
shifucun: Removed it here.

beets (commented):
i think it's confusing having both numbers there. some options below. 

image
image
image

it's unfortunate that we use red so much, since it could be construed as an error .. here is a try with green.
image

static/js/stat_var_hierarchy/node_header.tsx
```diff
@@ -29,7 +29,8 @@ const DOWN_ARROW_HTML = <i className="material-icons">remove</i>;
 const RIGHT_ARROW_HTML = <i className="material-icons">add</i>;
 
 interface StatVarHierarchyNodeHeaderPropType {
-  count: number;
+  statVarCount: number;
```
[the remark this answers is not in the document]
shifucun: Done

shifucun (commented):
> i think it's confusing having both numbers there. some options below.
> 
> image
> image
> image
> 
> it's unfortunate that we use red so much, since it could be construed as an error .. here is a try with green.
> image

Use highlighted blue now.

static/css/stat_var_hierarchy.scss
```diff
@@ -106,6 +106,10 @@ $horizontal-spacing: 15px;
   background-color: var(--dc-red-fade);
 }
 
+.selected-node-title {
+  color: #67adff;
```
beets, line 110: make the color a var?

beets (approved):

static/css/stat_var_hierarchy.scss
shifucun: Done

PR #1002, opened 2021-06-29
Choropleth files cleanup

chejennifer:
- delete the old choropleth tool files
- remove unused choropleth endpoints & consolidate the choropleth endpoints for map and place page into one


server/__init__.py
```diff
@@ -76,11 +76,10 @@ def create_app():
     app.register_blueprint(ranking.bp)
     app.register_blueprint(redirects.bp)
     app.register_blueprint(tools.bp)
-    from routes.api import (browser as browser_api, chart, choropleth, place as
+    from routes.api import (browser as browser_api, choropleth, place as
```
shifucun, line 79: shall we rename all the "choropleth" as "map"?

shifucun (approved):
Thanks for the clean up!

server/__init__.py
chejennifer: I can, but would I want to rename choropleth stuff in the place page to map as well? Or we could keep this api as choropleth and the name "map" would just be specific to the map explorer tool which plots a choropleth map

beets: +1 to keeping map as just the product name

beets (approved):
nice, thanks for the cleanup

server/__init__.py
shifucun: Change all the place page would be too many.. Yea, non-place change sounds good. This is not a must have, can do after all the feature is done.

PR #1003, opened 2021-06-29
[tree widget] small font size updates

beets:
made font sizes smaller, alignment fixes for inputs
made the widget ever-so-slightly wider

image


chejennifer (approved):

shifucun (approved):
wonder do we want less line spacing?

beets:
gave that a shot. also restored font sizes in browser pages

image

image

Issue #1006, opened 2021-06-30
[tree widget] browser pages should only show available nodes


beets:
we should exclude browser pages from showing *all* stat vars.

![image](https://user-images.githubusercontent.com/6052978/124010101-4078a380-d993-11eb-9cea-f1dca9d28b3e.png)

Issue #1005, opened 2021-06-30
[tree widget] entire path is not highlighted


beets:
in the timeline on autopush: https://autopush.datacommons.org/tools/timeline#place=geoId%2F06&statsVar=Count_HousingUnit__dc%2Fbqbxc8gc6tge4
- select a sv that is nested multiple levels down in the hierarchy.

observed:
- only the top two nodes are highlighted (the remaining nodes flash blue, then white again) -- possibly a state update after a rest call / re-render that doesn't retain the highlight

expected:
- entire path is highlighted

![image](https://user-images.githubusercontent.com/6052978/124004066-80885800-d98c-11eb-9e1a-d14be7811592.png)

cc @chejennifer

beets:
side note: the label of the leaf node (in this case, "16 years or more ...") should also be highlighted to help the user learn of the meaning of that highlight.

PR #1010, opened 2021-07-01
[tree widget] Various fixes

chejennifer:
- fixed [issue 1005](https://github.com/datacommonsorg/website/issues/1005):  entire path not highlighted. This was actually because a different path that goes to the same variable was being highlighted. Fixed this by making sure the path that user took to get to a stat var is the path that gets highlighted
- fixed [issue 1006](https://github.com/datacommonsorg/website/issues/1006): browser should only show available nodes
- fixed problem where when a tree widget section had been opened and then there was a places change, the contents of the opened section didn't update. 
- fixed problem where when the input in the search box was longer than the width of search box, it would overlay over the clear search button
- removed auto selecting the option user selects from the list of search results when user uses the search

dev updated with these changes

### DIFF
--- a/static/js/stat_var_hierarchy/stat_var_group_node.tsx
+++ b/static/js/stat_var_hierarchy/stat_var_group_node.tsx
@@ -27,6 +27,7 @@ import {
   StatVarInfo,
   StatVarGroupInfo,
   StatVarHierarchyNodeType,
+  StatVarHierarchyType,
 } from "../shared/types";
 import { StatVarHierarchyNodeHeader } from "./node_header";
 import { StatVarSection } from "./stat_var_section";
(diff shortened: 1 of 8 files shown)


static/js/stat_var_hierarchy/stat_var_group_node.tsx
```diff
@@ -191,31 +184,37 @@ export class StatVarGroupNode extends React.Component<
   }
 
   private fetchData(): void {
-    if (this.state.dataFetched) {
-      return;
-    }
     // stat var (group) dcid can contain [/_-.&], need to encode here.
```
shifucun, line 187: I remember there could be multiple fetchData() for a render, hence using this "if" statement to guard. Can you check if that's still valid.

shifucun (commented):
Thanks for making all these changes!

static/js/stat_var_hierarchy/stat_var_group_node.tsx
chejennifer: I have checked and there is only one fetchData()

shifucun (approved):

Issue #1008, opened 2021-06-30
[tree widget] add tooltips


beets:
We should try to describe some of the new features:
- counts: "Number of available statistical variables"
- grayed out: "There are no statistical variables for the chosen places. Please choose a different type of place" -- would be super cool to be able to suggest a place here.

I was actually confused about the counts -- since we are showing all sv's, it seemed as if that was the count for ALL sv's, not just the available ones.

PR #1013, opened 2021-07-07
[tree widget] Add tooltips

chejennifer:
- add an info tooltip to explain the count and the greyed out stat var groups ([issue 1008](https://github.com/datacommonsorg/website/issues/1008))
- add tooltip for greyed out stat vars that gives list of placetypes and/or place names that have data for this stat var

dev updated with these changes

Screen Shot 2021-07-07 at 2 19 51 PM
Screen Shot 2021-07-07 at 2 19 58 PM

### DIFF
--- a/server/routes/api/stats.py
+++ b/server/routes/api/stats.py
@@ -224,3 +224,15 @@ def get_stats_all():
     return Response(json.dumps(dc.get_stats_all(dcids, stat_vars)),
                     200,
                     mimetype="application/json")
+
+
+@bp.route('/api/stats/stat-var-summary', methods=["POST"])
+@cache.cached(timeout=3600 * 24, query_string=True)
+def get_statvar_summary():
+    """Gets the summaries for a list of stat vars.
+    """
+    stat_vars = request.json.get("statVars")
+    result = dc.get_statvar_summary(stat_vars)
+    return Response(json.dumps(result.get("statVarSummary", {})),
+                    200,
+                    mimetype='application/json')
--- a/server/routes/api/stats.py
+++ b/server/routes/api/stats.py
@@ -227,11 +227,11 @@ def get_stats_all():
 
 
 @bp.route('/api/stats/stat-var-summary', methods=["POST"])
-@cache.cached(timeout=3600 * 24, query_string=True)
 def get_statvar_summary():
     """Gets the summaries for a list of stat vars.
     """
     stat_vars = request.json.get("statVars")
+    print(stat_vars)
     result = dc.get_statvar_summary(stat_vars)
     return Response(json.dumps(result.get("statVarSummary", {})),
                     200,
--- a/static/js/stat_var_hierarchy/stat_var_hierarchy.tsx
+++ b/static/js/stat_var_hierarchy/stat_var_hierarchy.tsx
@@ -302,7 +302,7 @@ export class StatVarHierarchy extends React.Component<
     const html =
       "<ul><li>The number in parentheses represents the number of available" +
       "stat vars within the group that we have for the chosen place(s).</li>" +
-      "<li>Greyed out stat var groups have no available stat vars for the" +
+      "<li>Greyed out stat var groups have no available stat vars for the " +
       "chosen place(s), but can still be expanded for you to explore.</li></ul>";
     const topOffset = 30;
     const margin = 5;
--- a/static/js/stat_var_hierarchy/stat_var_hierarchy.tsx
+++ b/static/js/stat_var_hierarchy/stat_var_hierarchy.tsx
@@ -44,6 +44,8 @@ import {
 const SORTED_FIRST_SVG_ID = "dc/g/Demographics";
 const SORTED_LAST_SVG_ID = "dc/g/Miscellaneous";
 const ROOT_SVG = "dc/g/Root";
+const TOOLTIP_TOP_OFFSET = 30;
+const TOOLTIP_MARGIN = 5;
 
 interface StatVarHierarchyPropType {
   type: string;
@@ -300,12 +302,10 @@ export class StatVarHierarchy extends React.Component<
 
   private onMouseOverInfoIcon = () => {
     const html =
-      "<ul><li>The number in parentheses represents the number of available" +
-      "stat vars within the group that we have for the chosen place(s).</li>" +
+      "<ul><li>The number in parentheses represents the number of stat vars " +
+      "within the group where we have data for the chosen place(s).</li>" +
       "<li>Greyed out stat var groups have no available stat vars for the " +
       "chosen place(s), but can still be expanded for you to explore.</li></ul>";
-    const topOffset = 30;
-    const margin = 5;
     const containerY = (d3
       .select("#explore")
       .node() as HTMLElement).getBoundingClientRect().y;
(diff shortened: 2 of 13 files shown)


server/routes/api/stats.py
```diff
@@ -224,3 +224,15 @@ def get_stats_all():
     return Response(json.dumps(dc.get_stats_all(dcids, stat_vars)),
                     200,
                     mimetype="application/json")
+
+
+@bp.route('/api/stats/stat-var-summary', methods=["POST"])
+@cache.cached(timeout=3600 * 24, query_string=True)
```
shifucun, line 230: can you check if the cache works for POST? I remember it has issue with memoize, but not sure for "query_string=True"

static/js/stat_var_hierarchy/stat_var_hierarchy.tsx
```diff
@@ -283,4 +297,25 @@ export class StatVarHierarchy extends React.Component<
         return _.cloneDeep(resp.data["path"]).reverse();
       });
   }
+
+  private onMouseOverInfoIcon = () => {
+    const html =
+      "<ul><li>The number in parentheses represents the number of available" +
```
shifucun, line 303: need a space after "available"

static/js/stat_var_hierarchy/stat_var_hierarchy.tsx
```diff
@@ -283,4 +297,25 @@ export class StatVarHierarchy extends React.Component<
...
+      "stat vars within the group that we have for the chosen place(s).</li>" +
```
shifucun, line 304: "have data for"?

static/js/stat_var_hierarchy/stat_var_hierarchy.tsx
```diff
@@ -283,4 +297,25 @@ export class StatVarHierarchy extends React.Component<
...
+      "<li>Greyed out stat var groups have no available stat vars for the " +
+      "chosen place(s), but can still be expanded for you to explore.</li></ul>";
+    const topOffset = 30;
```
shifucun, line 307: moving the const to the top of the file?

static/js/stat_var_hierarchy/stat_var_section_input.tsx
```diff
@@ -97,12 +106,53 @@ export class StatVarSectionInput extends React.Component<
...
+          onMouseOut={() => hideTooltip()}
         >
           {this.props.statVar.displayName}
         </label>
       </form>
     );
   }
+
+  private mouseMoveAction = (e) => {
+    let html = "loading stat var summary...";
```
shifucun, line 121: how long does it load?

static/js/stat_var_hierarchy/stat_var_section_input.tsx
```diff
@@ -97,12 +106,53 @@ export class StatVarSectionInput extends React.Component<
...
+      }
+      html =
+        availablePlaceTypes.length > 0
+          ? "This stat var has no data for any of the chosen places." +
+            "You can try these types of places instead. <ul>"
+          : "Sorry, this stat var is not supported by this tool.";
+      for (const placeType of availablePlaceTypes) {
+        const placeSummary = this.props.summary.placeTypeSummary[placeType];
+        const placeList = placeSummary.topPlaces.map((place) => place.name);
+        html +=
```
shifucun, line 140: do you think it works to make the place names hyper link? Then user can directly go to the timeline page of that place.

static/js/stat_var_hierarchy/stat_var_section_input.tsx
```diff
@@ -97,12 +106,53 @@ export class StatVarSectionInput extends React.Component<
...
+          this.context.statVarHierarchyType === StatVarHierarchyType.TIMELINE
+            ? `<li>${placeType} (eg. ${placeList.join(", ")})</li>`
+            : `<li>${placeType}</li>`;
+      }
+      html += "</ul>";
+    }
+    const left = e.pageX;
+    const topOffset = 10;
```
shifucun, line 148: nit: put constant at top of the file

server/routes/api/stats.py
chejennifer: yes, you're right, the cache has problems. I removed the caching.

chejennifer: you're right, cache has issues, so I removed the caching.

static/js/stat_var_hierarchy/stat_var_hierarchy.tsx
chejennifer: done

static/js/stat_var_hierarchy/stat_var_hierarchy.tsx
chejennifer: done

static/js/stat_var_hierarchy/stat_var_hierarchy.tsx
chejennifer: done

static/js/stat_var_hierarchy/stat_var_section_input.tsx
chejennifer: The longest I've seen is 300ms

static/js/stat_var_hierarchy/stat_var_section_input.tsx
chejennifer: I think we've discussed how links don't really belong in tooltips..

static/js/stat_var_hierarchy/stat_var_section_input.tsx
chejennifer: done

static/js/stat_var_hierarchy/stat_var_section_input.tsx
```diff
@@ -97,12 +109,51 @@ export class StatVarSectionInput extends React.Component<
...
+          (placeType) =>
+            placeType in USA_CHILD_PLACE_TYPES && placeType !== "Country"
+        );
+      }
+      html =
+        availablePlaceTypes.length > 0
+          ? "This stat var has no data for any of the chosen places." +
+            "You can try these types of places instead. <ul>"
+          : "Sorry, this stat var is not supported by this tool.";
+      for (const placeType of availablePlaceTypes) {
```
shifucun, line 140: another thing is to the order of place types, can defer to later PR

shifucun (approved):

static/js/stat_var_hierarchy/stat_var_section_input.tsx
chejennifer: I think we've discussed that tooltips aren't really supposed to be used for the user to take actions? It should be more of a simple informational kind of thing ..

static/js/stat_var_hierarchy/stat_var_section_input.tsx
chejennifer: Ok, what would be the order of place types?

static/js/stat_var_hierarchy/stat_var_section_input.tsx
shifucun: I can see that, but from usability point of view, navigating to the hint places take some more effort (search place, then find the stat vars). We can gather thoughts from the team too, can update in later PR if necessary.

chejennifer: Yeah, I see what you mean, but in that case, maybe tooltip isn't the best way to show this information. Also trying to give users examples to navigate to will be a little more complicated for scatter and map tools where we need to know the parent place.

But yes, let's gather thoughts from the team!

PR #1015, opened 2021-07-09
Put Uncategorized at the bottom of stat var hierarchy

shifucun:


pradh (approved):
I'm also making a backend change to order them all. At which point we might remove this?

shifucun:
> I'm also making a backend change to order them all. At which point we might remove this?

Okay, just let autopush looking better now. Once cache is changed, can remove all the ordering logic in client

PR #1021, opened 2021-07-13
Only set per capita for measured property related to count

shifucun:


static/js/tools/timeline/chart.tsx
```diff
@@ -23,6 +23,8 @@ import { setChartPerCapita } from "./util";
 
 const CHART_HEIGHT = 300;
 
+const PER_CAPITA_MPROP = ["cumulativeCount", "incrementalCount", "count"];
```
beets, line 26: please add documentation for this

beets (approved):
thx for the fix!

PR #1027, opened 2021-07-13
[tree widget] handle selected stat vars being unavailable

chejennifer:
- update stat var union flask endpoint
- add handling for each of the tools timeline, map, and scatter

dev updated


static/js/tools/timeline/util.ts
```diff
@@ -42,13 +47,14 @@ export function getTokensFromUrl(name: string, sep: string): Set<string> {
   return tokens;
 }
 
-export function setTokensToUrl(
-  name: string,
-  sep: string,
-  tokens: Set<string>
-): void {
+export function setTokensToUrl(tokensMap: Record<string, TokenSet>): void {
```
shifucun, line 50: maybe leave the function as it was, so all the functions in this file have consistent signatures. I think the client can just do a for loop to accomplish the change here?

static/js/tools/timeline/page.tsx
```diff
@@ -125,6 +129,40 @@ class Page extends Component<unknown, PageStateType> {
       </>
     );
   }
+
+  private addPlaceAction(place: string): void {
+    // We only need to check the availability of selected stat vars when adding
```
shifucun, line 134: nit: more explicit "as we take the union of the eligible stat vars for all places"

static/js/tools/timeline/util.ts
chejennifer: I made this change so that the window.location.hash is only set once even if different params need to be updated. If the client does this in a for loop, then the window.location.hash will be set each time, which will cause an intermediate state in the window history..

server/routes/api/place.py
```diff
@@ -212,7 +212,7 @@ def statsvars(dcid):
 
 
 @cache.memoize(timeout=3600 * 24)  # Cache for one day.
-def get_stat_vars_union(dcids):
+def get_stat_vars_union(dcids, stat_vars):
```
shifucun, line 215: unfortunately we use "dcids" in the REST api; Here for clarity, should change "dcids" to "places"

static/js/tools/timeline/util.ts
shifucun: I see you want to set the url once. In that case can make the input a list of this struct:

export interface TokenInfo {
  name: string;
  sep: string;
  tokens: Set<string>;
}

static/js/tools/timeline/page.tsx
chejennifer: done

server/routes/api/place.py
chejennifer: done

static/js/tools/timeline/util.ts
chejennifer: done

shifucun (approved):

PR #1044, opened 2021-07-14
Use /v1/stat/set/series instead of old /bulk/stat API to fetch time series data

shifucun:
The old /bulk/stat API only returns provenance but not other metadata fields for StatVarObservation. For timeline tool, we want to show the "unit" in chart which requires the new /v1/stat/set/series API.

This is also a good step to deprecate (long pending task) /bulk/stat as the response is in "payload" format, with the response fields not consistent with other new APIs.

This PR also removes some unused StatData manipulation functions which were used for stack bar chart and group bar chart. They are now fetched from the landing page cache, and computed in the web server.

Also did code clean up:

* Change "Stats" to "Stat" whenever possible.
* Only use "Count_Person" as denominator as we don't have pre-defined denominators now.

### DIFF
--- a/static/js/shared/data_fetcher.ts
+++ b/static/js/shared/data_fetcher.ts
@@ -169,12 +169,6 @@ export function fetchStatData(
   }
   denomStatVars = Array.from(new Set(denomStatVars));
 
-  console.log("stat data");
-  console.log({
-    places: places,
-    statVars: statVars,
-  });
-
   const statDataPromise: Promise<StatApiResponse> = axios
     .post(`/api/stats`, {
       places: places,
@@ -185,12 +179,6 @@ export function fetchStatData(
     });
   let denomDataPromise: Promise<StatApiResponse>;
   if (denomStatVars.length > 0) {
-    console.log("stat data");
-    console.log({
-      places: places,
-      statVars: denomStatVars,
-    });
-
     denomDataPromise = axios
       .post(`/api/stats`, {
         places: places,
--- a/static/js/shared/data_fetcher.ts
+++ b/static/js/shared/data_fetcher.ts
@@ -86,7 +86,7 @@ export class StatData {
     const result: DataGroup[] = [];
     for (const statVar of this.statVars) {
       const dataPoints: DataPoint[] = [];
-      if (!this.data[place].data[statVar]) {
+      if (!this.data[place] || !this.data[place].data[statVar]) {
         continue;
       }
       const timeSeries = this.data[place].data[statVar];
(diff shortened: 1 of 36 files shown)


static/js/shared/data_fetcher.ts
```diff
@@ -387,108 +146,112 @@ function computePerCapita(
...
+    })
+    .then((resp) => {
+      return resp.data;
     });
+  let denomDataPromise: Promise<StatApiResponse>;
+  if (denomStatVars.length > 0) {
+    denomDataPromise = axios
+      .post(`/api/stats`, {
+        places: places,
+        statVars: denomStatVars,
```
beets, line 185: should we combine this in the previous call to /api/stats to reduce the number of requests? or a todo, for a followup if there are performance issues

static/js/shared/data_fetcher.ts
```diff
@@ -387,108 +146,112 @@ function computePerCapita(
...
-        result.data[sv] = computePerCapita(
-          allResp[i].data,
-          allResp[i + numStatsVars].data,
-          scaling
-        );
+    const statResp = allResp[0] as StatApiResponse;
+    const denomResp = allResp[1] as StatApiResponse;
+    const displayNameMapping = (allResp[2]
+      ? allResp[2]
+      : {}) as DisplayNameApiResponse;
```
beets, line 219: why is this special handling required only for display names? could there be empty response for the stat api calls too?

beets (approved):
whoa, big change! thanks for the cleanup as well.

static/js/shared/data_fetcher.ts
```diff
@@ -126,257 +84,58 @@ class StatsData {
       place = this.places[0];
     }
     const result: DataGroup[] = [];
-    for (const statsVar of this.statsVars) {
+    for (const statVar of this.statVars) {
       const dataPoints: DataPoint[] = [];
-      if (!this.data[statsVar][place]) continue;
-      const timeSeries = this.data[statsVar][place];
-      if (Object.keys(timeSeries.data).length !== 0) {
+      if (!this.data[place].data[statVar]) {
```
chejennifer, line 89: is it guaranteed that this.data[place] won't be null or undefined? I think this will cause an error if this.data[place] is null or undefined.

static/js/shared/data_fetcher.ts
```diff
@@ -126,257 +84,58 @@ class StatsData {
...
+  const yearMin = years[0];
+  const yearMax = years[years.length - 1];
+  for (const date in statSeries.val) {
+    const year = date.split("-")[0];
+    let pop: number;
+    if (year in popSeries.val) {
+      pop = popSeries.val[year];
+    } else if (year < yearMin) {
+      pop = popSeries.val[yearMin];
+    } else {
```
chejennifer, line 132: Could `year` be between yearMin and yearMax, but not in popSeries.val? In that case, maybe pop should be based off of the latest year in popSeries.val before `year`

static/js/shared/data_fetcher.ts
```diff
@@ -387,108 +146,112 @@ function computePerCapita(
...
+        if (perCapita) {
+          placeData[statVar] = computePerCapita(
+            placeData[statVar],
+            denomResp[place].data[TOTAL_POPULATION_SV],
+            scaling
+          );
+        } else if (statVar in denominators) {
+          placeData[statVar] = computePerCapita(
+            placeData[statVar],
+            denomResp[place].data[denominators[statVar]],
```
chejennifer, line 233: is it guaranteed that place will be in denomResp? If not, probably want to add a check

chejennifer (approved):
nice!

static/js/shared/data_fetcher.ts
shifucun: I tried to do that but that makes the code a bit hard to understand. Need to split the response based into denominator and numerator groups and they could have overlap. This is not too bad since this is adding at most one API call (only happens for per capita) and executed in parallel.

static/js/shared/data_fetcher.ts
shifucun: Double checked this could not be empty, so removing the check. Stat resp will always return non-null value too.

static/js/shared/data_fetcher.ts
shifucun: Yeah, the API will always return entries for given places, ex: https://api.datacommons.org/v1/stat/set/series?places=dummy&stat_vars=Count_Person

static/js/shared/data_fetcher.ts
shifucun: Done

static/js/shared/data_fetcher.ts
shifucun: very good catch! Picking the closest year and added a test.

shifucun (commented):
Thanks for the review!

PR #1045, opened 2021-07-15
Add unit to timeline tool Y label; Also fix data format in Download.

shifucun:
image


static/js/tools/timeline/chart.tsx
```diff
@@ -200,7 +200,20 @@ class Chart extends Component<ChartPropsType> {
...
     }
     // use mprop as the ylabel
-    const ylabelText = mprop.charAt(0).toUpperCase() + mprop.slice(1);
+    let ylabelText = mprop.charAt(0).toUpperCase() + mprop.slice(1);
+
+    // TODO(shifucun):
+    // Pick unit from one place and one stat var. This assumes all the stat var
+    // observation have same unit, which is mostly wrong but not strictly true.
+    // To do this accurately need a bigger change that group the stat vars at
+    // an early stage.
```
beets, line 209: in the meantime, could we list all the available units? it'll make the issue obvious (or not, depending on the severity)

beets (approved):
it'll be good to also use the units in the tooltip, which can be our source of truth for each observation

static/js/tools/timeline/chart.tsx
shifucun: Done

shifucun: Also added unit in tooltip

chejennifer (approved):

PR #1046, opened 2021-07-15
[timeline] fix a few deref errors for null series

beets:
follow up from #1044 - there are cases where a place will not have data for one of the stat vars, e.g.
https://autopush.datacommons.org/tools/timeline#statsVar=Count_Person_Employed__dc%2Fnm9hcklgg5zb3&place=nuts%2FES61%2CgeoId%2F56&chart=%7B%22consumption%22%3Afalse%7D


shifucun (approved):
Thanks for the fix!

PR #1051, opened 2021-07-15
[timeline] fix a few issues with missing plots

### LINKED ISSUE #1031: [timeline] Line not showing up for some charts
https://autopush.datacommons.org/tools/timeline#statsVar=dc%2Fnm9hcklgg5zb3__Count_Person_Employed&place=nuts%2FES61%2CgeoId%2F56
- the line for purple employed population doesn't show up although when hovering, we can see that there are data points for the purple employed population.

beets:
Fixes #1031. The [line.defined](https://github.com/d3/d3-shape/blob/main/README.md#line_defined) call is meant for filtering out undefined points, but since we fill in null values for all timestamps across datasets, this causes issues with cases like this:
timeseries a
- 2000: 1.0, 2001: 1.0, 2002: 1.0
timeseries b
- 2000-01: 2.0, .... 2000-12: 2.0, ...

in the example above, we filled in timeseries a with null for all the dates in timeseries b. then, the call to "line.defined" will find the lines incomplete and plot disjoint dots that are not visible.

for another [example](http://localhost:8080/tools/timeline#statsVar=Count_Person_Employed__dc%2Fnm9hcklgg5zb3&place=nuts%2FES61%2CgeoId%2F56&chart=%7B%22consumption%22%3Afalse%7D), here it is without the call to line.defined:
image
with the fix:
image

this is a bit of a compromise solution, where we manually filter out all the null values. this solves issues such as above, but also means we will interpolate missing points (such as in [this chart](https://autopush.datacommons.org/tools/timeline#&place=geoId/0606000,geoId/2511000,geoId/2603000,geoId/1777005,geoId/1225175,geoId/4815976&statsVar=Count_CriminalActivities_ViolentCrime)). i think that's ok, especially now that we have tooltips.

also included a fix for single points in datasets by drawing a circle, instead of showing a blank chart.

http://localhost:8080/tools/timeline#place=geoId%2F06&statsVar=LifeExpectancy_Person&chart=%7B%22genderIncomeInequality%22%3Afalse%7D
image

### DIFF
--- a/static/js/chart/draw.ts
+++ b/static/js/chart/draw.ts
@@ -1079,25 +1079,43 @@ function drawGroupLineChart(
   for (const place in dataGroupsDict) {
     dataGroups = dataGroupsDict[place];
     for (const dataGroup of dataGroups) {
-      const dataset = dataGroup.value.map((dp) => {
-        timePoints.add(dp.time);
-        return [dp.time, dp.value];
-      });
+      const dataset = dataGroup.value
+        .map((dp) => {
+          timePoints.add(dp.time);
+          return [dp.time, dp.value];
+        })
+        .filter((dp) => {
+          return dp[1] !== null;
+        });
       const line = d3
         .line()
-        .defined((d) => d[1] != null)
         .x((d) => xScale(d[0]))
         .y((d) => yScale(d[1]));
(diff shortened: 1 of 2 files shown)


chejennifer (approved):
thanks for the fix!

PR #1052, opened 2021-07-15
[tree widget] rename euronuts3

chejennifer:


beets (approved):
do we need to update the examples too? #1053

chejennifer:
> do we need to update the examples too? #1053

The examples come from the statVarSummary cache so if we need to update those, we would need to do it in the cache

PR #1055, opened 2021-07-15
[tree widget] update blocklist

chejennifer:
- update the groups being blocklisted
- enable blocklisting for staging


beets (approved):

pradh (approved):

server/routes/api/browser.py
```diff
@@ -30,9 +30,7 @@
 NO_OBSPERIOD_KEY = 'no_obsPeriod'
 
 # Temporary fix for messy svgs. Remove once svgs have been fixed.
-BLOCKLISTED_STAT_VAR_GROUPS = {
-    "dc/g/Person_EmploymentStatus", "dc/g/Establishment", "dc/g/Person_Industry"
-}
+BLOCKLISTED_STAT_VAR_GROUPS = {"dc/g/Establishment_Industry"}
 EMPLOYMENT_NUM_DESCENDENTS = 1759
```
pradh, line 34: This needs to be updated?

chejennifer: updated

pradh: Thanks!

Might be worth a comment explaining the magic number for the future "us" to know how it came about.

It came from this query....

```
# Number of Establishment by naics StatVars, with an additional payrollStatus constraint.
# To avoid double-counting curated stat-vars, count only generated ones.
SELECT COUNT(*) FROM `datcom-store.dc_kg_latest.StatisticalVariable`
WHERE population_type IN ('USCEstablishment', 'BLSEstablishment') AND
num_constraints <= 2 AND p1 = 'naics' AND id LIKE 'dc/%';
```

PR #1063, opened 2021-07-19
Update chart sorting in place page

shifucun:
* "Overview" tab contains sections of "categories", the order of the section should in the original order.
* "non Overview" tab contains sections of "chart groups", the order of the chart groups should be based on the title.


pradh (approved):

shifucun:
Thanks!

PR #1069, opened 2021-07-20
Add denominator to timeline page url

shifucun:
This was removed in recent stat var hierarchy change, where stat var denominator was removed and only Count_Person is used.

However, place page uses specific denominator for stat vars and that carries over in the explore more link. This fixes the broken link. Also fixes the "per capita" checkbox bug

### DIFF
--- a/static/js/tools/timeline/util.ts
+++ b/static/js/tools/timeline/util.ts
@@ -56,7 +56,8 @@ export function setTokensToUrl(tokens: TokenInfo[]): void {
   window.location.hash = urlParams.toString();
 }
 
-// add one statVar
+// Add a token to the url. A token could either be a place dcid, a stat var or
+// a [stat var,denominator] pair separated by comma.
 export function addToken(name: string, sep: string, token: string): void {
   const tokens = getTokensFromUrl(name, sep);
   if (tokens.has(token)) {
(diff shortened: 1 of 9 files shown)


static/js/tools/timeline/util.ts
```diff
@@ -56,7 +56,8 @@ export function setTokensToUrl(tokens: TokenInfo[]): void {
   window.location.hash = urlParams.toString();
 }
 
-// add one statVar
+// Add a token to the url. A token could either be a place dcid, a stat var or
+// a [stat var,denominator] pair separated by comma.
```
chejennifer, line 60: from the landing page api, it looks like [stat var, denominator] pair is now separated by "|" instead of comma?

chejennifer (approved):

static/js/tools/timeline/util.ts
shifucun: Yes, "," is used as the separator for places, want to have a different symbol to differentiate here. @pradh can you confirm is there "|" used in stat var dcid?

pradh: It turns out bio DCIDs do contain '|', but I do want to get rid of it... this may be fine.

Issue #1026, opened 2021-07-13
[timeline] ordering of graphs wrt tree widget


beets:
From Ajai:

![image](https://user-images.githubusercontent.com/6052978/125523563-d924a00c-d57b-4b0d-ab6e-b558acfa8a6e.png)

The ordering of the charts seems arbitrary - if they are sorted, it should be on something visible on the screen. Could we add each new chart to the top? That would also solve the confusion of charts "not appearing" on the screen, but are in the scroll area. Thoughts?

chejennifer:
Adding each new chart to the top makes sense to me. I guess when user navigates to the page via URL, we would show the charts in the order of the stat vars in the URL?

PR #1070, opened 2021-07-20
[timeline] Order the charts

chejennifer:
- Order timeline charts so that the newly selected chart gets added to the top of the page [issue 1026](https://github.com/datacommonsorg/website/issues/1026)

updated on dev


beets (approved):
thanks for the fix!

shifucun (approved):
Nit: (for future PRs) can start the change in a clean branch so the commit history is cleared

PR #1076, opened 2021-07-30
Remove old hierarchy json files as the new Stat Var Menu is launched

shifucun:
Now npm does not complain on large compiled files :)


beets (approved):
nice, thanks for the cleanup!

PR #1077, opened 2021-07-30
Add barebones for bio data commons page

shifucun:
This is to setup two pages of protein pages:
- landing: /bio/protein
- node page: /bio/protein/<dcid>

This is to unblock future work to add biology related graph.


server/routes/bio.py
```diff
@@ -0,0 +1,34 @@
...
+"""
+
+from flask import Blueprint, render_template
+import routes.api.shared as shared_api
+
+bp = Blueprint('bio', __name__, url_prefix='/bio')
+
+
+@bp.route('/')
+def main():
```
beets, line 24: nit: restrict these to non-staging/prod

beets (approved):
also, should we leave room for more bio-related pages? i know there's a protein and disease browser being developed -- are those both different views based on type? or should be structure it as two separate browsers. if so, we might want to add a suffix / subdir to these so that it will be easy to add the others in the future.

suhana13 (approved):
Looks good to me ! To answer Carolyn's question (@beets ), we're planning to have two different browsers: protein and disease. So, once we have a template ready for proteins, we can easily reuse it for diseases. Thank you so much Bo @shifucun !

server/routes/bio.py
shifucun: Done

shifucun (commented):
> also, should we leave room for more bio-related pages? i know there's a protein and disease browser being developed -- are those both different views based on type? or should be structure it as two separate browsers. if so, we might want to add a suffix / subdir to these so that it will be easy to add the others in the future.

I think we can show different views based on the type? I don't have an idea of how much this would scale. If it turns out to be too much, we can split in the future? It's kind of easy to manage for now.

shifucun:
> Looks good to me ! To answer Carolyn's question (@beets ), we're planning to have two different browsers: protein and disease. So, once we have a template ready for proteins, we can easily reuse it for diseases. Thank you so much Bo @shifucun !

I see. I think the question here is if we need two url like `datacommons.org/bio/protein/<dcid>` and `datacommons.org/bio/gene/<dcid>`, or just `datacommons.org/bio/<dcid>`. I will not reserve the sub-space for different types for now. If we need, we can easily expand that in the future.

beets:
If this is meant to be protein specific, let's split it up for now? Since we're planning both protein and disease already.

khoahoang1891999 (approved):

shifucun:
Okay, update the folder and naming so this is only for protein.

server/routes/bio.py
spiekos: There will be more bio-related pages including a slightly modified version of the place browser. The Disease and Protein browser will contain summary graphs representing the data related to each of these entitites.

shifucun:
> If this is meant to be protein specific, let's split it up for now? Since we're planning both protein and disease already. If we are to expand on it, we should expand it soon, before it gets too complicated...
> 
> Code wise as well, it would be good to name files bio/gene_browser so we don't have a big renaming to do when we get to the next one..

Moved protein into its own folders. The access url is still under bio/protein.

spiekos:
There will be more bio-related pages including a slightly modified version of the place browser. The Disease and Protein browser will contain summary graphs representing the data related to each of these entities. I think it would make sense to anticipate additional browser / viewers to be developed in the future.

spiekos (approved):

beets (approved):
thanks for the clarification, @spiekos and for the updates @shifucun !

PR #1078, opened 2021-08-02
[i18n] add fallback messages

beets:
Added a script to add fallback messages for untranslated stat var labels and titles. Additionally,

- extracted and compiled newly added strings
- added descriptions for new stat var labels and titles (please see static/js/i18n/strings/en/stats_var_*.json)
- sort keys in json files. this made the diffs rather large :(
- added test to check that the right client files are updated with new chart config changes

Before:
image

After:
image


shifucun (approved):
Thanks for adding this! Sorry I forgot to follow up with the new chart i18n work :(

Would be nice to add some tests later to prevent incomplete change.

server/chart_config.json
```diff
@@ -59,7 +59,7 @@
   },
   {
     "category": "Economics",
-    "titleId": "Unemployment rate: urban vs rural",
+    "titleId": "CHART_TITLE-Unemployment_rate-urban-v-rural",
```
shifucun, line 62: wonder if we could have some automatic schemes to update the IDs (and also enforce the ID format from test?)

shifucun (approved):

server/chart_config.json
beets: yea, tests would be good. not sure about automatic scheme - really the prefix is all that's required, and a test could catch that (and easy to add a prefix).

i added a basic test to check that the right files were updated in https://github.com/datacommonsorg/website/pull/1078/commits/2801550f6450fb5792f8117375de6b6b8ff74831

PR #1079, opened 2021-08-04
Pipe in real data for protein page

shifucun:
This gets the page working e2e. Next step is to use d3 to draw charts instead of showing plain text

### DIFF
--- a/static/js/protein/page.tsx
+++ b/static/js/protein/page.tsx
@@ -19,22 +19,76 @@
  */
 
 import React from "react";
+import axios from "axios";
+
+import { GraphNode } from "../shared/types";
 
 interface PagePropType {
   dcid: string;
   nodeName: string;
 }
 
-export class Page extends React.Component<PagePropType> {
+interface PageStateType {
+  data: GraphNode;
+}
+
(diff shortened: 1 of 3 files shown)


static/js/protein/page.tsx
```diff
@@ -19,22 +19,76 @@
...
 
   componentDidMount(): void {
-    // TOOD: add data fetching here.
+    this.fetchData();
   }
 
   render(): JSX.Element {
-    return <div> Protein Data</div>;
+    const tissueScore = this.getTissueScore();
+    // TODO: use d3 to draw bar chart here.
```
shifucun, line 47: @suhana13 @khoahoang1891999 This will be where you call the d3 library. We have some examples like https://github.com/datacommonsorg/website/blob/master/static/js/chart/draw_choropleth.ts#L113.

You can create a similar function to draw the new chart and call it in this file.

suhana13 (approved):
Looks good to me! I'll start off on the D3 plots with dummy data and then we can substitute it with the real data queried from data commons!

beets (approved):

PR #1086, opened 2021-08-05
[i18n] sync changes with tc to avoid re-translating due to updated descriptions

beets:
some changes from #1078 were incompatible with how we pull in strings for translation (the descriptions needed to be kept in the en *.po file due to the silly hack i added for the long homepage strings.

there were also some string / description updates made directly to the translation bundle that are out of sync with the repo -- updating them now to avoid retranslations.


shifucun (approved):
Thanks for the fix!

PR #1087, opened 2021-08-06
Use a dummy chart for bio protein page

shifucun:
Next step is to use the d3 code in observable and draw the real chart

### DIFF
--- a/static/js/protein/chart.ts
+++ b/static/js/protein/chart.ts
@@ -1,0 +1,53 @@
+/**
+ * Copyright 2021 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


suhana13 (approved):

static/js/protein/chart.ts
```diff
@@ -0,0 +1,53 @@
...
+const WIDTH = 500;
+
+/**
+ * Draw bar chart for tissue score.
+ */
+export function drawTissueScoreChart(
+  id: string,
+  data: { name: string; value: string }[]
+): void {
+  // TODO: convert data to tmpData.
```
beets, line 32:
```suggestion
  // TODO: convert data to tmpData format.
```

beets (approved):

shifucun:
Thanks!

PR #1089, opened 2021-08-09
[map] remove boundary lines for granularities 2 levels down from the selected place

chejennifer:
eg. counties in USA (country) will have its boundary lines removed

Screen Shot 2021-08-09 at 1 45 44 PM

but counties in California (State) will keep its boundary lines
Screen Shot 2021-08-09 at 1 46 07 PM

### DIFF
--- a/static/js/tools/map/util.ts
+++ b/static/js/tools/map/util.ts
@@ -30,6 +30,8 @@ export const USA_CHILD_PLACE_TYPES = {
   County: ["County"],
 };
 
+export const USA_PLACE_HIERARCHY = ["Country", "State", "County"];
+
 const URL_PARAM_VALUE_SEPARATOR = "-";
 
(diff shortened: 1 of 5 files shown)


static/js/tools/map/util.ts
```diff
@@ -30,6 +30,8 @@ export const USA_CHILD_PLACE_TYPES = {
   County: ["County"],
 };
 
+export const USA_PLACE_HIERARCHY = ["Country", "State", "County"];
```
beets, line 33: mind adding a comment here?

beets (approved):
nice change :) 

might be worth a simple test so we don't mess the logic in the future.

thanks!

static/js/tools/map/util.ts
chejennifer: done

PR #1090, opened 2021-08-09
[scatter] add option for quadrant lines for scatterplots

beets:
quadrant lines are plotted based off the mean of both x and y values

image

adds a new "qd" query parameter, with display option:

image


chejennifer (approved):
nice!

PR #1091, opened 2021-08-09
[scatter] add option for text labels

beets:
v0 - no logic to control for text overlap
also added all the display options to a single row

image

### DIFF
--- a/static/js/tools/scatter/context.ts
+++ b/static/js/tools/scatter/context.ts
@@ -91,11 +91,13 @@ const EmptyPlace: PlaceInfo = Object.freeze({
   upperBound: 1e10,
 });
 
-interface IsQuadrantWrapper {
-  value: boolean;
+interface DisplayOptionsWrapper {
+  showQuadrants: boolean;
+  showLabels: boolean;
 
   // Setters
-  set: Setter<boolean>;
+  setQuadrants: Setter<boolean>;
+  setLabels: Setter<boolean>;
 }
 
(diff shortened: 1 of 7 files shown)


static/js/tools/scatter/context.ts
```diff
@@ -164,8 +166,9 @@ const FieldToAbbreviation = {
   lowerBound: "lb",
   upperBound: "ub",
 
-  // Quadrant fields
+  // DisplayOptions fields
   isQuadrant: "qd",
```
chejennifer, line 170: nit: maybe these can also be showQuandrant and showLabels

static/js/tools/scatter/context.ts
```diff
@@ -176,6 +179,7 @@ function useContextStore(): ContextType {
   const [y, setY] = useState(EmptyAxis);
   const [place, setPlace] = useState(EmptyPlace);
   const [isQuadrants, setQuadrants] = useState(false);
```
chejennifer, line 181: nit: maybe change to showQuadrants and showLabels to be consistent?

chejennifer (approved):
nice!

static/js/tools/scatter/context.ts
beets: done!

static/js/tools/scatter/context.ts
beets: done!

PR #1093, opened 2021-08-10
[energy] fix api calls for new emissions stat vars

beets:
the frontend was getting key errors from looking for the value of the 'baseDate' in 'objectId' instead of 'objectValue' in the triples call. lmk if this is where it's expected generally, and i can make it more general (instead of the special case there for baseDate)


server/routes/api/stats.py
```diff
@@ -112,26 +112,41 @@ def stats_var_property_wrapper(dcids):
...
+                mprop = objId
+            if predicate == 'populationType':
+                pt = objId
+            if predicate == 'measurementDenominator':
+                md = objId
+            if predicate == 'statType':
+                st = objId
+            if predicate == 'name':
+                name = objVal
+            if predicate == 'baseDate':
```
pradh, line 132: baseDate, if it appears, should be part of constraintProperties. so do we need this given 137?

server/routes/api/stats.py
```diff
@@ -112,26 +112,41 @@ def stats_var_property_wrapper(dcids):
...
+                base = objVal
+            if predicate == 'measurementQualifier':
+                mq = objId
+            if predicate in pvs:
+                if predicate == 'baseDate':
+                    v = objVal
+                else:
```
pradh, line 139: Here too, should we generalize by looking for objId first, and if not falling back to objVal?

Typically constraint-prop values are IDs, but there is no hard rule that it shouldn't be Text/Date/Number, and this would allow for it.

pradh (approved):
Thanks for the fix!

static/js/shared/stat_var.ts
```diff
@@ -17,12 +17,22 @@ import axios from "axios";
...
+  // populationType
   pt?: string;
+  // constraint properties { property: value }
   pvs?: { [key: string]: string };
+  // name
   title?: string;
+  // statType
   st?: string;
+  // baseDate
+  bd?: string;
```
pradh, line 33: nit: baseDate can be removed from here?

pradh (approved):

static/js/shared/stat_var.ts
beets: ah yes, thanks for catching that. done.

PR #1094, opened 2021-08-10
[timeline] change chart background for forecasted data

beets:
this pr builds off of #1093 
adds a faint, labeled data for projected data (has a baseDate). 

image

did not use emissionsScenario as a test since [1990 data does not have it set](https://autopush.datacommons.org/browser/geoId/06031?statVar=DifferenceRelativeToBaseDate1990_Daily_PrecipitationRate), though let me know if that's a bug and should be considered. here's both 2006 and 1990 base data together, with the area from 1990 onwards marked as projection.

image


pradh:
> this pr builds off of #1093
> adds a faint, labeled data for projected data (has a baseDate).
> 
> image
> 
> did not use emissionsScenario as a test since [1990 data does not have it set](https://autopush.datacommons.org/browser/geoId/06031?statVar=DifferenceRelativeToBaseDate1990_Daily_PrecipitationRate), though let me know if that's a bug and should be considered. here's both 2006 and 1990 base data together, with the area from 1990 onwards marked as projection.
> 
> image

pradh:
Sorry, I do not fully understand why stuff to the RHS of 1990 is considered projected, but on LHS?

My vague understanding is that everything here is projected.  And we sort of picked a "baseDate" to be 1990 and 2006 to compare rest of all the data points in the time-series against.

chejennifer (approved):
nice!

beets:
ok, updated to:

image

pradh (approved):
Thanks for the update!

PR #1099, opened 2021-08-11
Remove API key requirement

shifucun:


server/templates/static/about.html
```diff
@@ -57,12 +57,7 @@ <h2>Who can use it?</h2>
       Data Commons can be accessed by anyone via the tools available on
       datacommons.org. Students, researchers and developers can use the
       REST and Python APIs, both of which are free for educational,
```
beets, line 59:
```suggestion
      <a href="https://docs.datacommons.org/api/" target="_blank">REST, Python and Google Sheets APIs</a>, both of which are free for educational,
```

beets (approved):
thanks! i also had a fix but forgot to make a pr.

shifucun:
Thanks!

PR #1109, opened 2021-08-21
Display unit in map tool legend

shifucun:
image


chejennifer (approved):
thanks for the change!

beets (approved):
thanks for the change! maybe add 3-5 more px between the unit and the legend color block?

shifucun:
> thanks for the change! maybe add 3-5 more px between the unit and the legend color block?

Done

PR #1111, opened 2021-08-23
[map] add units to breadcrumbs

chejennifer:
Screen Shot 2021-08-23 at 10 22 52 AM
Screen Shot 2021-08-23 at 10 23 21 AM

### DIFF
--- a/static/js/tools/map/chart_options.tsx
+++ b/static/js/tools/map/chart_options.tsx
@@ -29,6 +29,7 @@ import {
   MAP_REDIRECT_PREFIX,
 } from "./util";
 import { DataPointMetadata } from "./chart_loader";
+import _ from "lodash";
 
 const NO_PER_CAPITA_TYPES = ["medianValue"];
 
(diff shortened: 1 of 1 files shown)


static/js/tools/map/chart_options.tsx
```diff
@@ -41,9 +42,10 @@ interface ChartOptionsPropType {
 export function ChartOptions(props: ChartOptionsPropType): JSX.Element {
   const { statVar } = useContext(Context);
   const selectedPlace = props.placeInfo.selectedPlace;
+  const unitString = _.isEmpty(props.unit) ? "" : ` ${props.unit}`;
   const selectedPlaceValue =
     selectedPlace.dcid in props.dataValues
-      ? formatNumber(props.dataValues[selectedPlace.dcid], props.unit)
+      ? formatNumber(props.dataValues[selectedPlace.dcid], "") + unitString
```
shifucun, line 48: @beets now we have unit like "MicrogramsPerCubicMeter" and "μg/m3", I wonder shall we have symbol property for all the unit in the format of "ug/m3" or is this something that i18n should handle?

beets: I like the idea of the latter (adding SI units to schema). We can add [translations](https://github.com/datacommonsorg/website/blob/master/static/js/i18n/strings/hi/units.json) as they get added to place explorer, since that will be for a smaller number of units. wdyt?

shifucun: That sounds good. Thanks!

shifucun (approved):

PR #1117, opened 2021-09-09
[Private DC] GKE deployment config and util functions update to make it run

shifucun:
- There is no provenance from TMCF+CSV, may need to enforce a manifest for each import.
- Color change is not reflected in this PR. Will do a separate one for color and homepage change.

### DIFF
--- a/server/__init__.py
+++ b/server/__init__.py
@@ -122,7 +122,7 @@ def create_app():
     app.config['BABEL_TRANSLATION_DIRECTORIES'] = 'i18n'
 
     if not cfg.TEST:
-        timeout = 60  # seconds
+        timeout = 120  # seconds
         counter = 0
         isOpen = False
         while not isOpen:
--- a/server/__init__.py
+++ b/server/__init__.py
@@ -122,7 +122,7 @@ def create_app():
     app.config['BABEL_TRANSLATION_DIRECTORIES'] = 'i18n'
 
     if not cfg.TEST:
-        timeout = 120  # seconds
+        timeout = 60  # seconds
         counter = 0
         isOpen = False
         while not isOpen:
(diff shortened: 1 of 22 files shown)


server/__init__.py
```diff
@@ -122,7 +122,7 @@ def create_app():
     app.config['BABEL_TRANSLATION_DIRECTORIES'] = 'i18n'
 
     if not cfg.TEST:
-        timeout = 120  # seconds
+        timeout = 60  # seconds
```
beets, line 125: i don't remember why we bumped up the timeout - were there long API calls that we rely on? even so, best to do this in a separate PR

server/webdriver_tests/base_test.py
```diff
@@ -49,7 +49,7 @@ def setUp(self):
         chrome_options.add_argument('--disable-dev-shm-usage')
 
         # Maximum time, in seconds, before throwing a TimeoutException.
-        self.TIMEOUT_SEC = 60
+        self.TIMEOUT_SEC = 120
```
beets, line 52: ditto, please move this to a new PR

server/templates/base.html
```diff
@@ -29,7 +30,7 @@
 -#}
 <!DOCTYPE html>
 
-<html lang={{ locale }}>
+<html lang={{ locale }} class={{ DEPLOYMENT }}>
```
beets, line 33: i'd prefer it if we only set an overriding class in the private case, so public remains the default (besides some performance concerns, also to avoid us relying on the public class)

beets (approved):
very cool bo!

server/webdriver_tests/base_test.py
shifucun: Done

server/templates/base.html
shifucun: Done

server/__init__.py
shifucun: Done

PR #1118, opened 2021-09-10
Homepage update for private DC

shifucun:
image

### DIFF
--- a/server/routes/static.py
+++ b/server/routes/static.py
@@ -31,17 +31,16 @@ _HOMEPAGE_PLACE_DCIDS = [
 
 @bp.route('/')
 def homepage():
-    if current_app.jinja_env.globals['DEPLOYMENT'] == "private":
+    if current_app.config['PRIVATE']:
         return render_template('static/private.html')
-    else:
-        place_names = place_api.get_display_name(
-            '^'.join(_HOMEPAGE_PLACE_DCIDS), g.locale)
-        blog_date = babel_dates.format_date(date(2021, 7, 26),
-                                            format='long',
-                                            locale=g.locale)
-        return render_template('static/homepage.html',
-                               place_names=place_names,
-                               blog_date=blog_date)
+    place_names = place_api.get_display_name('^'.join(_HOMEPAGE_PLACE_DCIDS),
+                                             g.locale)
+    blog_date = babel_dates.format_date(date(2021, 7, 26),
+                                        format='long',
+                                        locale=g.locale)
+    return render_template('static/homepage.html',
+                           place_names=place_names,
+                           blog_date=blog_date)
 
 
 @bp.route('/about')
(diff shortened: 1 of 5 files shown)


server/routes/static.py
```diff
@@ -31,14 +31,17 @@
...
-                                             g.locale)
-    blog_date = babel_dates.format_date(date(2021, 7, 26),
-                                        format='long',
-                                        locale=g.locale)
-    return render_template('static/homepage.html',
-                           place_names=place_names,
-                           blog_date=blog_date)
+    if current_app.jinja_env.globals['DEPLOYMENT'] == "private":
+        return render_template('static/private.html')
+    else:
```
beets, line 36: else not needed here, since there's a return statement in the if clause above.

server/templates/base.html
```diff
@@ -49,7 +49,7 @@
       <nav class="navbar navbar-dark navbar-expand-lg col" id="main-nav">
         <div class="container-fluid">
           <div class="navbar-brand">
-            <a href="{{ url_for('static.homepage') }}">Data Commons</a>
+            <a href="{{ url_for('static.homepage') }}">{% if DEPLOYMENT == "private" %} International Energy Agency {% else %} Data Commons {% endif %}</a>
```
beets, line 52: ok for now, but we should pass the name in as a variable, easiest through a global (we could set the name Data Commons or the private instance name through a single global, so there doesn't need to be a check for DEPLOYMENT)

server/templates/static/private.html
```diff
@@ -0,0 +1,57 @@
...
+<link rel="stylesheet" href={{url_for('static', filename='css/static.min.css', t=config['VERSION'])}}>
+{% endblock %}
+
+{% block content %}
+<section id="homepage-top">
+  <div class="container">
+    <div class="row">
+      <div class="lead-col col-12 col-lg-6">
+        <p class="lead mr-lg-3">
+          Private Data Commons for International Energy Agency. The source data is in TMCF and CSV format stored in <a href="https://pantheon.corp.google.com/storage/browser/datcom-public/iea">Google Cloud Storage</a>.
```
beets, line 31: maybe instead of "stored in", say "loaded from"?

server/templates/static/private.html
```diff
@@ -0,0 +1,57 @@
+{#
+ Copyright 2020 Google LLC
```
beets, line 2: nit: 2021

beets (approved):

server/routes/static.py
shifucun: Done

server/templates/base.html
shifucun: Good point! Moved this and "private" to configmodule.py as variables.

server/templates/static/private.html
shifucun: Done

server/templates/static/private.html
shifucun: Done

shifucun:
Thanks for review!

PR #1119, opened 2021-09-10
[map] add points on top of map

chejennifer:
- add functionality where if the url contains &ppt=<placeType>, we will plot points for places of that type that are contained within the current map

dev updated (eg. https://dev.datacommons.org/tools/map#%26sv%3DMedian_Age_Person%26pc%3D0%26pd%3DgeoId%2F06027%26pn%3DInyo%20County%2C%20CA%26pt%3DCounty%26ept%3DCounty%26ppt%3DCity)

Screen Shot 2021-09-09 at 11 48 51 PM

### DIFF
--- a/server/routes/api/choropleth.py
+++ b/server/routes/api/choropleth.py
@@ -445,8 +445,8 @@ def get_map_points():
         if len(latitude_by_geo) == 0 or len(longitude) == 0:
             continue
         map_point = {
-            "dcid": geo_id,
-            "name": names_by_geo.get(geo_id, "Unnamed Place"),
+            "placeDcid": geo_id,
+            "placeName": names_by_geo.get(geo_id, "Unnamed Place"),
             "latitude": float(latitude[0]),
             "longitude": float(longitude[0])
         }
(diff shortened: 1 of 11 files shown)


server/routes/api/choropleth.py
```diff
@@ -412,3 +412,45 @@ def choropleth_data(dcid):
...
+    longitude_by_geo = dc_service.get_property_values(geos, "longitude")
+
+    map_points_list = []
+    for geo_id, latitude in latitude_by_geo.items():
+        longitude = longitude_by_geo.get(geo_id, [])
+        if len(latitude_by_geo) == 0 or len(longitude) == 0:
+            continue
+        map_point = {
+            "dcid": geo_id,
+            "name": names_by_geo.get(geo_id, "Unnamed Area"),
```
beets, line 449: nit: perhaps "Unnamed place" since these are points

static/css/tools/map.scss
```diff
@@ -235,3 +235,12 @@
...
   padding: 0.3rem;
 }
+
+.dot {
+  r: 4;
+  stroke: var(--dc-gray-lite);
+}
+
+.dot:hover {
+  r: 6;
```
beets, line 245: a slight transition would be nice.

beets (approved):
nice refactor!

server/routes/api/choropleth.py
chejennifer: done

static/css/tools/map.scss
chejennifer: done

Issue #1103, opened 2021-08-17
[scatter] keep axis for existing var when a third is added


beets:
in scatter plot explorer, add 2 variables and place, see the chart shown, then add a third var (dropping the existing x-axis).

observed: the existing y-axis var is now on the x-axis

expected: the axis of the "kept" var stays the same

PR #1120, opened 2021-09-10
[scatter] keep axis for existing var when a third is added

chejennifer:
- fix for [issue #1103](https://github.com/datacommonsorg/website/issues/1103)

screen-recording (20)


beets (approved):
thx for the fix!

PR #1125, opened 2021-09-13
[map] keep dots constant size even when hovering

chejennifer:


shifucun (approved):

beets:
Did we explore a changed outline color at least, to indicate which point the tooltip refers to?

chejennifer:
> Did we explore a changed outline color at least, to indicate which point the tooltip refers to?

I have not explored that, but I can try some stuff out

PR #1127, opened 2021-09-13
Remove parallel run for webdriver test; Cleanup screenshot tests

shifucun:
- There are consistent timeout error when running webdriver tests under "parallel" mode. It runs well (and not slow) in normal mode, hence removing the the Selenium Grid setup.
- Screenshot tests have not been used for a long time and does not seem to have the value. Remove related code


chejennifer (approved):
Thanks for the fix and the cleanup!

PR #1128, opened 2021-09-14
Check inconsistent units for data from different sources

shifucun:
When there are inconsistent units, UI would alert but still show the charts without the unit in label


static/js/tools/shared_util.ts
```diff
@@ -87,12 +87,18 @@ export function getUnit(placePointStat: PlacePointStat): string {
...
-  } else {
-    return "";
+  let unit = null;
+  for (const key in metadata) {
+    const u = metadata[key].unit;
+    if (unit === null) {
+      unit = u;
+    } else if (u !== unit) {
+      // TODO: throw exception in this function and better handle it in caller.
+      alert(`Found inconsistent unit: ${unit}, ${u}`);
```
beets, line 97: can we do this only in dev / autopush? we should not alert in prod, and there'll be large imports where we won't be able to check all charts

shifucun: Some background of this change. One example of inconsistent unit is from the electricity generation data, which has both Megawatts and Gigawatts from two imports. When plotted in scatter plot, it gives obvious wrong information. In this case, we could 1) filter and leave only one unit data, 2) alert that the data are not comparable, 3) do unit conversion.

3) seems to be a better solution, but may require some more efforts (and should considering scaling factor as well). In the mean time, I feel it's useful to let users know the data are not comparable?

chejennifer: Maybe short term solution of filtering and leaving only one unit data might be better. Although there will be less data points, at least the plot will still be valid and users can still use it to make a comparison. Compared to giving users a plot that they can't really use.

beets: i think giving a usable chart would be preferable, since there isn't much the user can do to fix the issue. ideally, we'd add a note that some points were not shown because of unit mismatch. 

we should think about how to do (3) long term..

PR #1129, opened 2021-09-14
[map] update hover behaviour for dots

chejennifer:
- on hover, make the dots outline and fill colors opposite that of the regular state

screen-recording (21) (1)


beets (approved):
thx!

PR #1130, opened 2021-09-14
[map] Grey out dots for places that have no data

chejennifer:
- if a place has no data, make the dot grey. I'm not sure how I feel about the fact that these grey dots look very similar to the hover state, what are your thoughts @beets @shifucun 

screen-recording-_22_

### DIFF
--- a/static/css/tools/map.scss
+++ b/static/css/tools/map.scss
@@ -239,10 +239,9 @@
 .dot {
   r: 4;
   stroke: var(--dc-gray-lite);
-  fill: black;
 }
 
-.dot:hover {
-  stroke: black;
+.dot:hover, .dot-missing-values:hover{
   fill: var(--dc-gray-lite);
+  stroke: black;
 }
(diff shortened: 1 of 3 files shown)


beets (approved):
i'd been thinking that the hover state over a map feels very "heavy", and close to the "no data" color.

something we did with the place explorer was to [apply a filter to the svg shape](https://github.com/datacommonsorg/website/blob/175e9c62f64e62b89f8a1bcc2ed4bfb599c7c6a0/server/templates/place.html#L70) (we do that to highlight the current place in comparison charts). i wonder if something similar would work for the base area hover. but this can be a follow on, not blocking.

chejennifer:
> i'd been thinking that the hover state over a map feels very "heavy", and close to the "no data" color.
> 
> something we did with the place explorer was to [apply a filter to the svg shape](https://github.com/datacommonsorg/website/blob/175e9c62f64e62b89f8a1bcc2ed4bfb599c7c6a0/server/templates/place.html#L70) (we do that to highlight the current place in comparison charts). i wonder if something similar would work for the base area hover. but this can be a follow on, not blocking.

Thanks for the review! I created an issue to look into updating the hover behavior: https://github.com/datacommonsorg/website/issues/1131

PR #1133, opened 2021-09-15
Update on IEA DC

shifucun:
- remove footer
- do not use "private"

### DIFF
--- a/server/templates/static/private.html
+++ b/server/templates/static/private.html
@@ -17,6 +17,8 @@
 
 {% set main_id = 'homepage' %}
 {% set title = 'Home' %}
+{% set is_hide_full_footer = true %}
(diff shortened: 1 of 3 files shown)


server/templates/static/private.html
```diff
@@ -38,10 +41,9 @@
 <section class="container">
   <h2>{% trans %}Explore the data{% endtrans %}</h2>
   <p>
-    Your private data is merged with public data from Data Commons and served in a unified format.
+    The IEA data is merged with public data from Data Commons and served in a unified format.
```
beets, line 44:
```suggestion
    IEA data is merged with public data from Data Commons and served in a unified format.
```

beets (approved):

shifucun:
Thanks!

PR #1140, opened 2021-09-22
Update translator schema mapping to use stat var observation

shifucun:


pradh (approved):
Thanks!

pradh:
Looks like the default https://datacommons.org/translator page includes the "Place Observation" scenario, and it produces a query that refers to old Observation table.

@beets

PR #1141, opened 2021-09-27
Split place page chart config json file by category

shifucun:
This is to make it easy to 
- maintain the config 
- pull configs that are only relevant to a deployment
- sort the categories

After this is checked in. Need to have a script to sync the concatenated json in g3


server/__init__.py
```diff
@@ -32,6 +32,17 @@
 
 propagator = google_cloud_format.GoogleCloudFormatPropagator()
 
+RANKED_CATEGORIES = [
```
beets, line 35: nit: maybe ORDERED_PLACE_EXPLORER_CATEGORIES or just PLACE_EXPLORER_CATEGORIES

static/js/shared/util.ts
```diff
@@ -14,6 +14,17 @@
  * limitations under the License.
  */
 
+export const topics = [
```
beets, line 17: add a note that this needs to stay in sync with server/__init__.py?

server/__init__.py
beets: also add a note that this needs to stay in sync with util.ts?

beets (approved):
thx for splitting it up!

static/js/shared/util.ts
shifucun: Done

shifucun:
Thanks for review!

PR #1143, opened 2021-09-28
Add air quality index chart to place page

shifucun:
Note the map chart time range is too wide. Time trend is also too dense to view

image


beets (commented):
this change looks good - but i wonder if we should start a new branch for this (if you'd like to work incrementally) while we fix the data density issue, as well as add other climate charts? that way we won't block pushes to prod if those are necessary and we could collaborate on adding charts.

shifucun:
> this change looks good - but i wonder if we should start a new branch for this (if you'd like to work incrementally) while we fix the data density issue, as well as add other climate charts? that way we won't block pushes to prod if those are necessary and we could collaborate on adding charts.

Changed the base to "place" branch

beets (approved):

server/__init__.py
```diff
@@ -41,6 +41,7 @@
     "education",
     "demographics",
     "housing",
+    "climate",
     "environment",
```
pradh, line 45: climate and environment seem very similar sounding.

seems like environment only contains disasters now?

pradh (approved):

server/__init__.py
shifucun: environment has energy and co2 emission now. I am thinking split it into "energy" and "environment" afterwords and adding more energy charts?

pradh: That sounds better.  Also, it might be nicer to line up these sections with verticals on SV hierarchy: https://screenshot.googleplex.com/9ijDDViorHF7AZJ

PR #1144, opened 2021-09-28
[place] fix layout for charts without comparisons

beets:
We were adding div's around each chart block, which creates an entire row (even if there could be opportunity for a chart on the next row to float up).

Before:
image

After: 
image

### DIFF
--- a/docs/developer_guide.md
+++ b/docs/developer_guide.md
@@ -258,7 +258,7 @@ the same region.
 1. Run these commands:
    ```bash
    ./scripts/extract_messages.sh
-   ./scripts/compiled_messages.sh
+   ./scripts/compile_messages.sh
    ```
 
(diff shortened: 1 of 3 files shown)


shifucun:
The original purpose of this is to have related charts in one line. Some chart does not have related chart, so it will just have itself in the block. Here is an example below. I wonder what would be a good way to let users see the "block" of (related) charts? If the div is removed, then all charts will be together

image

beets:
Agreed, it was because of cases like that that makes it look awkward, because it's not immediately clear that the intent is to have a comparison chart on the right of the line chart. I think each chart stands on it's own, so it's ok to have them float in.

There's also the three column view, and it works out here that the display is more dense. It also adds a bit more variety to the page with the different chart types..
 
image

shifucun (approved):
3 columns looks good!

PR #1146, opened 2021-09-29
[Place Page Chart] Daily max temperature change from year 2006

shifucun:
image


scripts/compile_messages.sh
```diff
@@ -20,7 +20,7 @@ set -e
 cd static
 npm list @formatjs/cli || npm install formatjs
 
-# for LOCALE in de en es fr hi it ja ko pt-BR ru zh-CN;
+# for LOCALE in de en es fr hi it ja ko ru;
```
beets, line 23: you can delete the commented out line -- it should have been a TODO to add support for those languages (pybabel relies on local Locale files with underscore separators, and I couldn't figure out how to get around that).

server/chart_config/climate.json
```diff
@@ -10,5 +10,19 @@
       "scale": false
     },
     "isChoropleth": true
+  },
+  {
+    "category": "Climate",
+    "titleId": "CHART_TITLE-Daily_max_temperature_rcp85_diff_to_2006",
+    "title": "Daily max temperature change (from year 2006)",
```
beets, line 17: we should add that it's under RCP 8.5 scenario. maybe:

Projected max daily temperature change under RCP 8.5 (based on year 2006)

static/js/i18n/strings/en/stats_var_labels.json
```diff
@@ -1062,5 +1062,9 @@
   "AirQualityIndex_AirPollutant": {
     "defaultMessage": "AQI",
     "description": "Label used to describe statistical variables in chart legends, string should be short. Describes the air quality index."
+  },
+  "DifferenceRelativeToBaseDate2006_Daily_Max_Temperature_RCP85": {
+    "defaultMessage": "Max Temperature Change",
```
beets, line 1067: please keep "RCP 8.5"  and base year 2006 in the string

static/js/i18n/strings/en/stats_var_titles.json
```diff
@@ -1062,5 +1062,9 @@
   "AirQualityIndex_AirPollutant": {
     "defaultMessage": "AQI",
     "description": "Title for a page about a statistical variable. Variable describes the air quality index."
+  },
+  "DifferenceRelativeToBaseDate2006_Daily_Max_Temperature_RCP85": {
+    "defaultMessage": "Projected Max Temperature Change from Year 2006",
```
beets, line 1067: please keep RCP 8.5 in the string

server/chart_config/climate.json
```diff
@@ -10,5 +10,19 @@
...
+    "description": "Daily max temperature change from year 2006 based on RCP 8.5",
+    "statsVars": [
+      "DifferenceRelativeToBaseDate2006_Daily_Max_Temperature_RCP85"
+    ],
+    "unit": "K",
```
beets, line 22: this usually needs additions to units.json (which was manually curated :-/), but looks like we get lucky here :)

beets (commented):
should we skip adding the localization files until we get all the charts in? 

i also wonder if it's worth adding the other emissions scenario for comparison..

scripts/compile_messages.sh
shifucun: Done

server/chart_config/climate.json
shifucun: Done

server/chart_config/climate.json
shifucun: Got. So only default symbol is used, we don't need to add it to units.json?

static/js/i18n/strings/en/stats_var_labels.json
shifucun: Done

static/js/i18n/strings/en/stats_var_titles.json
shifucun: Done

shifucun:
> should we skip adding the localization files until we get all the charts in?

If I don't run `./scripts/extract_messages.sh && ./scripts/compile_messages.sh`, then the title will be showing ad titleId with "CHART_TITLE-...". Is there a way to avoid that?
> 
> i also wonder if it's worth adding the other emissions scenario for comparison..

Sure, what other scenarios should we add?

shifucun:
> > should we skip adding the localization files until we get all the charts in?
> 
> If I don't run `./scripts/extract_messages.sh && ./scripts/compile_messages.sh`, then the title will be showing ad titleId with "CHART_TITLE-...". Is there a way to avoid that?
> 
> > i also wonder if it's worth adding the other emissions scenario for comparison..
> 
> Sure, what other scenarios should we add?

I tried 4 different RCP scenario, wdyt?

image

beets:
> > should we skip adding the localization files until we get all the charts in?
> 
> If I don't run `./scripts/extract_messages.sh && ./scripts/compile_messages.sh`, then the title will be showing ad titleId with "CHART_TITLE-...". Is there a way to avoid that?

There isn't an easy way to avoid that -- you could update the list of locales in both scripts to only include en for now? and we can do a sweep of the other languages at one go later

beets:
> > > should we skip adding the localization files until we get all the charts in?
> > 
> > 
> > If I don't run `./scripts/extract_messages.sh && ./scripts/compile_messages.sh`, then the title will be showing ad titleId with "CHART_TITLE-...". Is there a way to avoid that?
> > > i also wonder if it's worth adding the other emissions scenario for comparison..
> > 
> > 
> > Sure, what other scenarios should we add?
> 
> I tried 4 different RCP scenario, wdyt?
> 
> image

this looks great - but for now, can we split out the 4 line charts? we might also leave out comparison charts for now, since it's a little odd to only compare the last date, but rather we'd want a range, or the mean for the year.

long term, we'll really want to show the mean, and bands rather than actual temperature projections:
image

shifucun:
> > > should we skip adding the localization files until we get all the charts in?
> > 
> > 
> > If I don't run `./scripts/extract_messages.sh && ./scripts/compile_messages.sh`, then the title will be showing ad titleId with "CHART_TITLE-...". Is there a way to avoid that?
> 
> There isn't an easy way to avoid that -- you could update the list of locales in both scripts to only include en for now? and we can do a sweep of the other languages at one go later

Okay, now only do i18n for en

shifucun:
Make only one chart for each scenario now.

shifucun:
> Make only one chart for each scenario now.

image

server/chart_config/climate.json
beets: Yea, if we want to keep it as "K"

Maybe we might want to do that as a follow up anyway, since K isn't a commonly used, and it could be confused for "thousands" the way it's displayed now. wdyt? maybe we should spell it out

beets (approved):
thanks for all the updates!

server/chart_config/climate.json
shifucun: Changed to "celsius" and use "°c" as symbol

shifucun:
> thanks for all the updates!

Now I know i18n much better :)

shifucun:
Thanks for all the suggestions!

PR #1148, opened 2021-09-30
[Place Page Chart] PM 2.5

shifucun:
image


server/chart_config/climate.json
```diff
@@ -54,5 +54,18 @@
     ],
     "unit": "celsius",
     "isOverview": false
+  },
+  {
+    "category": "Climate",
+    "titleId": "CHART_TITLE-Mean_concentration_PM2.5",
+    "title": "Mean concentration of PM 2.5",
```
beets, line 61: " Mean PM2.5 concentration" seems to be the more common way to describe this.

server/chart_config/climate.json
```diff
@@ -54,5 +54,18 @@
...
+    "description": "Mean concentration of PM 2.5",
```
beets, line 62: please be more descriptive for this.. maybe add that it's for air quality, and measures concentrations of fine inhalable particles, with diameters that are generally 2.5 micrometers and smaller in the air. you could add this link too: https://www.epa.gov/air-trends/particulate-matter-pm25-trends 

this doesn't get displayed -- and goes straight to the translators

static/js/i18n/strings/en/units.json
```diff
@@ -34,5 +34,11 @@
   },
   "celsius-display": {
     "defaultMessage": "°c"
+  },
+  "micro-gram-per-cubic-meter": {
+    "defaultMessage": "{0} μg/m3"
```
beets, line 39: should we use the superscript character for this? µg/m³

static/js/i18n/strings/en/units.json
```diff
@@ -34,5 +34,11 @@
   },
   "celsius-display": {
     "defaultMessage": "°c"
```
beets, line 36: sorry just noticed this. should be °C (capital C)

beets (approved):
thanks - this is a good chart to add

server/chart_config/climate.json
shifucun: Done

server/chart_config/climate.json
shifucun: Done

static/js/i18n/strings/en/units.json
shifucun: Done

static/js/i18n/strings/en/units.json
shifucun: Good catch!

shifucun:
Thanks for review!

PR #1152, opened 2021-10-01
stat var explorer tweaks

n-h-diaz:
includes 

* changes timeline charts to redirect to stat var explorer
* removes param from hash
* changes url to "statvar" and adds redirect from "stat-var"
* makes places column smaller in provenance tables
* changes example column to "Example Places"


server/routes/redirects.py
```diff
@@ -66,6 +66,11 @@ def get_involved():
     return redirect('https://docs.datacommons.org/contributing/', code=302)
 
 
+@bp.route('/tools/stat-var')
```
shifucun, line 69: If this tool has not been exposed yet, this is not needed? Also this only works for prod but not other deploy.

shifucun (approved):
Thanks for the update!

server/routes/redirects.py
n-h-diaz: I'll keep for now since /tools/stat-var currently works on prod
note: this will temporarily break autopush.../tools/stat-var until this change is pushed to prod

n-h-diaz:
thanks for the review!

PR #1157, opened 2021-10-05
[Place Page Chart] Change Environment to Energy and move charts accordingly

shifucun:
This is to align with the categories in stat var widget as well


pradh (approved):
Thank you!

beets (approved):
thanks!

PR #1160, opened 2021-10-06
Merge "place" branch with all the place page charts update

shifucun:
A follow up would be to enable i18n for all the languages and update the translation files.


shifucun:
Thanks!

PR #1163, opened 2021-10-06
Remove the disaster by type chart as the scale of each type are far off

shifucun:
Also this chart appears together with Earthquake and Cyclone charts, which is odd. Guha suggested to take this down.


shifucun:
thanks!

PR #1164, opened 2021-10-07
[map] support countries

chejennifer:
allow placetype of country to be selected and when country is selected, show place as Earth

screen-recording (24)

### DIFF
--- a/static/package-lock.json
+++ b/static/package-lock.json
@@ -6327,23 +6327,6 @@
         }
       }
     },
-    "d3-geo-projection": {
-      "version": "4.0.0",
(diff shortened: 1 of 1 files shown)


shifucun:
> > Looking at the screencast, I feel the map is elongated vertically?
> 
> This is because I used an equal area projection which according to this website is good for choropleth maps (https://medium.com/multiple-views-visualization-research-explained/designing-choropleth-maps-what-projection-to-choose-e3ceb1314b69). But I can also use other projections that might look better?

I personally feel this scaling is not familiar, good to run with more people :)

chejennifer:
Thanks for the reviews!

PR #1172, opened 2021-10-09
[sustainability] add ability to run_server to start multiple instances

beets:
Added the following flags to run_server.sh:
-e for the environment
-p for the localhost port


shifucun (approved):
Thanks for adding this!

PR #1176, opened 2021-10-12
[scatter] Add density coloring

chejennifer:
screen-recording-_26_


beets:
Let's go ahead and merge this, and we can play with it on autopush :)

PR #1184, opened 2021-10-15
[cleanup] move types out of data_fetcher, and remove some dead code

beets:
data fetcher is included in most tools, so changes there are slow to compile, even though all imports except for timeline are only for types

also downgraded sort-import to a warning


shifucun (approved):
thanks for the clean up!

PR #1186, opened 2021-10-15
Temporarily use private dc config for sustainability.dc.org

beets:
So it can be auto-deployed to iea.datacommons.org


shifucun (approved):
Nice. Can monitor after this PR is submitted. Should auto pushed to iea.datacommons.org

PR #1188, opened 2021-10-15
[timeline] update to timeline chart controls

beets:
the current chart controls are perhaps too prominent, in that they are first on the top-left, which should be for "most important" things. moving it to below the chart, in line with our other tools, and under the legend where it gets a lot more "breathing room". all are totally technical terms :)

also renamed "delta" to "consecutive differences", and added labels so you can click on the text as well, expanding the click area of the checkbox.

image

### DIFF
--- a/static/js/chart/base.ts
+++ b/static/js/chart/base.ts
@@ -402,13 +402,6 @@ function dataGroupsToCsv(dataGroups: DataGroup[]): string {
   return result;
 }
 
-interface Range {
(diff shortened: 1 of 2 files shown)


shifucun:
In the screenshot, the band color is out of the chart box?

beets:
in the interest of time, i also fixed the overflowing model lines. sorry for skipping review on the commit..

image

PR #1190, opened 2021-10-18
[scatter] Add bivariate choropleth view

chejennifer:
- enable scatter plot tool to view the data as a bivariate choropleth
    - add toggle for user to choose between scatter and bivariate choropleth
    - add code to generate the bivariate choropleth legend
    - data processing to use drawChoropleth method
- move the code for drawing scatter plot into its own file so that chart.tsx calls either drawScatter or drawChoropleth depending on the chart type

dev updated (eg. https://dev.datacommons.org/tools/scatter#%26svx%3DMedian_Income_Person%26svy%3DCount_Person_AsianAlone%26pcy%3D1%26epd%3Dcountry%2FUSA%26epn%3DUnited%20States%20of%20America%26epts%3DCountry%26ept%3DCounty%26qd%3D0%26ld%3D0%26dd%3D0%26ct%3D1)

### DIFF
--- a/static/css/tools/scatter.scss
+++ b/static/css/tools/scatter.scss
@@ -297,8 +297,15 @@ circle {
 .chart-title {
   display: flex;
   justify-content: center;
+  flex-wrap: wrap;
 }
 
 .chart-title h3 {
   text-align: center;
+  padding: 0px 5px;
+}
+
+.chart-title span {
+  font-size: 1rem;
+  margin-bottom: 0.5rem;
 }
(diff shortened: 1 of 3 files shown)


shifucun:
> > > > Thoughts from the demo page:
> > > > 
> > > > * title: we may want to make A vs B clearer by emphasizing A and B, since they all could be long strings. Can we try make it 3 lines with "A", "vs", "B" in separate lines
> > > > * Round the float value in bi-variate legend
> > > 
> > > 
> > > If we make the title into 3 separate lines, the title will end up taking up more of the screen space and I think it might be better for the chart to have more space
> > 
> > 
> > Or, any other visual hint to make A and B more prominent? I found it hard to tell apart especially for two long names
> 
> How about this with the "vs" as a smaller and non bolded text? Screen Shot 2021-10-18 at 11 58 15 AM

this is better!

PR #1192, opened 2021-10-19
Remove "daily" from IPCC StatVars in chart config

pradh:
This PR requires a corresponding change in the Cache.

Could you confirm if I should wait until the corresponding StatVar changes have reached autopush before merging this PR?


shifucun:
Thanks! Can you also update the stat vars here https://github.com/datacommonsorg/website/blob/master/server/configmodule.py#L28-L31.

Yes, this PR needs to wait until the cache is build.

pradh:
@shifucun could you help me understand why the selenium test fails? 

There was a binary `server/i18n/fr/LC_MESSAGES/*.mo` that had a match for daily but I didn't know how it got updated.

PR #1200, opened 2021-10-20
Update website for IPCC changes

shifucun:
- Pull in mixer data and code change so this snapshot is complete
- Remove additional stat vars for place page. New data already has those.
- Fix timeline too ipcc data emptiness check.


shifucun:
Thanks!

PR #1205, opened 2021-10-22
Allow drawing india maps for map and scatter

chejennifer:
Screen Shot 2021-10-21 at 5 07 32 PM
Screen Shot 2021-10-21 at 5 10 31 PM

Note: it looks like we may be missing geojsons for some of the AdministrativeArea2 places, I will look into that
Screen Shot 2021-10-21 at 5 07 47 PM
)


pradh:
This is very cool!  Looks like Telengana has very less data!

PR #1210, opened 2021-10-25
[map] add support for europe

beets:
Add basic support for displaying sub-european maps.

I added continents back to the parent places list so that it will be possible to test for european countries on the client. So they'll start showing up in the "contained places" breadcrumbs on the bottom.

We'll need a better way to add support for other regions as they come, I tried to make the changes minimal.

Known issues:
- European countries aren't clickable from the Earth view
- Only France, Germany and UK are searchable (there is a Maps API limit of 5 country restrictions)
- Continent views, accessible via breadcrumbs from countries above + India + US, include territories that are far away from the physical continent (and Russia crosses the date line), so the views are rather skewed.

image


chejennifer (approved):

chejennifer (approved):
Thanks for adding this!

PR #1228, opened 2021-10-26
[stat var] apply drawer toggle to stat var menus in map and timeline tools

beets:
also fixes layout issue in the other tools (needed to apply flex-grow to plot-container)


chejennifer (approved):
nice!

shifucun (approved):

PR #1239, opened 2021-10-28
[map] fix legend labeling issue

chejennifer:
- map legend labels in both the map explorer and place page were sometimes quite off
    - eg. in autopush: https://screenshot.googleplex.com/BBkeRzg5ENaxEdz, with this change: https://screenshot.googleplex.com/CA5UpYuytkUcTCk
- However, now that the legend labels are correct, we can see that it's unintuitive in terms of the spacing of the colors. In the example above, the distance between 0 and 0.02 is much greater than the distance between any of the other labels


beets:
i modified our color scale function to produce linear scales (only using the extents) except in the case of manually updating the domain via url param

the results look reasonable.. we lose some color saturation but on the whole looks correct.

also flipped the scale so highest is at the top

earlier example:
image

chejennifer:
agreed that it looks correct now! However, there will be maps that will look like this (which still look accurate):
Screen Shot 2021-11-02 at 10 37 50 AM

shifucun:
> agreed that it looks correct now! However, there will be maps that will look like this (which still look accurate): Screen Shot 2021-11-02 at 10 37 50 AM

I wonder how the domain is computed? Should it be based on some quantile of the values instead of just min and max?, In this case, should the max be the 90% quantile and let the highest value saturate?

shifucun:
> > agreed that it looks correct now! However, there will be maps that will look like this (which still look accurate): Screen Shot 2021-11-02 at 10 37 50 AM
> 
> I wonder how the domain is computed? Should it be based on some quantile of the values instead of just min and max?, In this case, should the max be the 90% quantile and let the highest value saturate?

In extreme case of [0.1, 0.2, 0.5, 0.3, 0.6, 10], use 10 as the max value does not seem make sense, it will be the darkest value anyway, no matter you pick 0.7 or 10 as domain_max

shifucun:
I also feel the actual value and color of the legend does not help much for a user. It's almost not possible to really map a number and a color from the legend and then correlate that in the actual map.. So we should aim for best contrast of the image first, and maybe just show the min and max value for the legend, just to let users know the range of the data.

beets:
image
image

image
image

static/js/chart/draw_choropleth.ts
```diff
@@ -674,14 +673,19 @@ function generateLegendSvg(
...
   canvas.width = 1;
-  canvas.height = n;
+  canvas.height = height;
   const context = canvas.getContext("2d");
-  for (let i = 0; i < n; ++i) {
-    context.fillStyle = (color(i / (n - 1)) as unknown) as string;
+  for (let i = 0; i < height; ++i) {
+    // yScale maps from color domain values to height values. Therefore, to get
+    // the color at a certain height, we want to first get the color domain
+    // value for that height and then get hte color for that value.
```
shifucun, line 691: nit: s/hte/the

shifucun (approved):

static/js/chart/draw_choropleth.ts
beets: done

beets:
thx for the reviews and suggestions!

PR #1243, opened 2021-10-28
[Browser] Improve usability by showing in/out arc dcid

shifucun:
When I look up stat var property in browser, turns out the display text is different from enum, which is not obvious. Even if user knows it, has to click in to copy the dcid. This is try to improve that by showing the dcid as well. It makes a row a bit busier but makes it more clear and easy to use.

image


shifucun:
Using smaller font weight now

shifucun:
@beets PTAL

beets (approved):
sorry missed this. thx for the update!

shifucun:
thanks!

PR #1248, opened 2021-10-29
[scatter] fix jumpiness option changes

beets:
use css to set size of the chart drawing area, to reduce jumpiness on replotting.
also updated container to fluid for some options that were left out with earlier layout fixes.

will work on debouncing next..

### DIFF
--- a/static/js/tools/scatter/chart.tsx
+++ b/static/js/tools/scatter/chart.tsx
@@ -257,18 +257,12 @@ function plot(
   geoJsonData: GeoJsonData
 ): void {
   const svgContainerRealWidth = svgContainerRef.current.offsetWidth;
-  // TODO: Use CSS to set the height of the chart so it's visible (< 1 vh).
-  const scatterHeight = Math.min(
-    window.innerHeight * 0.6,
-    svgContainerRealWidth
-  );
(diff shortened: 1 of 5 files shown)


static/js/tools/scatter/chart.tsx
```diff
@@ -257,18 +257,12 @@ function plot(
...
 ): void {
   const svgContainerRealWidth = svgContainerRef.current.offsetWidth;
-  // TODO: Use CSS to set the height of the chart so it's visible (< 1 vh).
-  const scatterHeight = Math.min(
-    window.innerHeight * 0.6,
-    svgContainerRealWidth
-  );
-  const scatterWidth = svgContainerRealWidth;
-  const chartHeight = scatterHeight;
+  const chartHeight = svgContainerRef.current.offsetHeight;
```
chejennifer, line 260: on initial load, will the svgContainerRef have an offsetHeight?

chejennifer (approved):

static/js/tools/scatter/chart.tsx
beets: yep, it's set from css, so it should always be 60vh.

beets:
thanksf or teh review!

PR #1250, opened 2021-10-29
[scatter] update links for covid examples

chejennifer:


beets (approved):
thanks!!

while you're here, could you fix the reported typo in our about page? thanks!

PR #1256, opened 2021-10-30
[tools] small layout updates

beets:
- Add debounce to scatter resize updates (and remove window resize handler now that height is set via css)
- Add a few more "container-fluid" to plot options
- Remove resize handler from map which doesn't need it anymore (viewbox takes care of resizing too!)

https://user-images.githubusercontent.com/6052978/139513671-bb2ab953-8ab1-4cd7-8280-51a1998e42fa.mp4


chejennifer:
I just realized, but is it possible to not use viewbox for map? I remember switching from viewbox to resize handler for map because with viewbox, when we load the page on a small screen and then enlarge it, the lines become much thicker:
- loaded on small screen and then enlarged the screen:
Screen Shot 2021-11-01 at 7 36 47 AM

 - loaded on the full size screen:
Screen Shot 2021-11-01 at 7 37 53 AM

chejennifer (approved):

beets:
done - updated map to use resize observer
i also split the useEffect block for both scatter and map, as they are listening to different changes. ptal

chejennifer:
looks good, thanks for the changes!

PR #1260, opened 2021-10-31
Revert "Add website login and require login for non-(prod/local) instances"

shifucun:
Reverts datacommonsorg/website#1258

After deploying to GKE I realized auth need more complicated handling than this. There are a few issues:

1) The user info/session is only stored at per instance, so the login information is lost for consecutive requests, hence requires a lot of re-login. Then a distributed key-value store is needed to store the user data. I tried Redis which works.
2) https is handled by GCP load balancer as a reverse proxy, but. the Oauth callback url has to be an https. This can be solved by "app.wsgi_app = ProxyFix(app.wsgi_app, x_for=1, x_host=1)" though.
3) Need to store the "allowed users" somewhere
4) Need to make  /healthz and other ports open to GCP probers, which adding another complexity.

After trying all of these, there are still weird behavior from the chained re-direct from web pages and log in pages.

I realized all we want can be easily achieved from https://cloud.google.com/iap.. and this is verified at dev.datacommons.org


pradh (approved):

shifucun:
thanks!

PR #1263, opened 2021-11-01
[scatter] various updates/fixes

chejennifer:
- add text wrap for tooltip
Screen Shot 2021-11-01 at 7 27 53 AM

- on dot click, go to place page
- when switching to new stat var, reset per capita and scale log for that axis


beets (approved):
pls let guha know about the update in the scatter sv change behavior

PR #1266, opened 2021-11-01
[scatter] only allow choosing log scale for axis with all positive or all negative numbers

chejennifer:
Screen Shot 2021-11-01 at 1 02 58 PM

- also remove source information from tooltip


static/js/tools/scatter/plot_options.tsx
```diff
@@ -84,36 +92,40 @@ function PlotOptions(): JSX.Element {
...
-                  checked={x.value.log}
-                  onChange={(e) => checkLog(x, e)}
-                />
-                {display.chartType === ScatterChartType.SCATTER
-                  ? "X-axis"
-                  : x.value.statVarInfo.title || x.value.statVarDcid}
-              </Label>
-            </FormGroup>
-          </Col>
+          {yMinMax[0] * yMinMax[1] > 0 && (
```
beets, line 95: nice! we should set a disabled state to the input instead - but this is good for now.

add a short comment though, about why this is disabled, for a future reader.

beets (approved):

static/js/tools/scatter/plot_options.tsx
chejennifer: done

PR #1274, opened 2021-11-02
[tools] various fixes

chejennifer:
- update bivariate map legend so that the min and max doesn't change when using log scale
- fix flashing of points when switching stat vars in scatter explorer
    - original: screen-recording-_34_
    - with fix: screen-recording-_35_
- fix tooltip in the tree widget
    - original:  
Screen Shot 2021-11-02 at 12 38 01 PM
    - with fix: 
Screen Shot 2021-11-02 at 12 38 42 PM


beets (approved):
thx for these fixes!

PR #1276, opened 2021-11-02
[map] fix webdriver tests

chejennifer:


server/webdriver_tests/map_test.py
```diff
@@ -146,6 +138,8 @@ def test_manually_enter_options(self):
         element_present = EC.presence_of_element_located(
             (By.ID, 'Median_Age_Persondc/g/Demographics-Median_Age_Person'))
         WebDriverWait(self.driver, self.TIMEOUT_SEC).until(element_present)
+        screen_hidden = EC.invisibility_of_element_located((By.ID, 'screen'))
```
shifucun, line 141: oh, this is nice. Later can make this as a function in base? I found other tools were also flakey due to flash or hidden screen when updating stat var menu

shifucun (approved):

server/webdriver_tests/map_test.py
chejennifer: done, added a todo to make this a function in base

PR #1278, opened 2021-11-03
[Map] few fixes

chejennifer:
- update naming of IPCCPlace_50 to "0.5 Arc Degree" 
- remove boundaries for places of type "0.5 Arc Degree"
- highlight on hover for places that can't be clicked but just no pointer for those places
- update tooltip to not show stat var dcid

eg.
Screen Shot 2021-11-02 at 5 01 01 PM

- fixed weird things happening with installations and per capita
    - original: screen-recording-_36_
    - with these changes:  
screen-recording-_37_


static/js/tools/map/chart.tsx
```diff
@@ -433,23 +435,19 @@ const getTooltipHtml = (
...
+  const footer = showPopDateMessage
+    ? `<footer><sup>1</sup> Uses population data from: <wbr>${metadata.popDate}</footer>`
     : "";
   const html =
-    titleHtml +
-    `${statVarTitle} (${metadata.statVarDate}): <wbr><b>${value}</b><br />` +
-    `<footer>${popDateHtml}</footer>`;
+    `${titleHtml} (${metadata.statVarDate}): <wbr><b>${value}</b>${
+      showPopDateMessage ? "<sup>1</sup>" : ""
+    }<br />` + footer;
```
beets, line 450: just pointing it out (no need to update your code), but in this case - you could also just always add ${footer} to the format string.

beets (approved):
thx for these fixes!

PR #1293, opened 2021-11-03
Homepage tweaks

pradh:
* Drop wet-bulb number-of-months, pending average computation
* Clarify that Max Temperature charts are actually change in max temperature
* Use projection statvar for difference across models

Screen Shot 2021-11-03 at 4 16 57 PM


beets (approved):
thx for these fixes!

PR #1294, opened 2021-11-03
[map] add special projections for europe & asia

beets:
Only applied at the continent level for now.

image

image

### DIFF
--- a/static/js/tools/map/chart.tsx
+++ b/static/js/tools/map/chart.tsx
@@ -346,7 +346,7 @@ function draw(
         USA_PLACE_DCID,
         props.placeInfo.parentPlaces
       ),
-      props.placeInfo.selectedPlace.dcid == EUROPE_NAMED_TYPED_PLACE.dcid,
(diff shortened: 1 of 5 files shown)


static/js/tools/map/chart.tsx
```diff
@@ -346,6 +346,7 @@ function draw(
         USA_PLACE_DCID,
         props.placeInfo.parentPlaces
       ),
+      props.placeInfo.selectedPlace.dcid,
```
chejennifer, line 349: nit: this probably won't cause any problems, but want to use props.placeInfo.enclosingPlace.dcid because that's the actual enclosing place of the map (eg. for leaf pages, the enclosing place will be different from the selected place)

chejennifer (approved):

pradh:
Very cool!

static/js/tools/map/chart.tsx
beets: thanks for the catch!

beets:
thx for the reviews!

PR #1296, opened 2021-11-04
[scatter] allow select country for Europe

chejennifer:
- also populate place type options after place has been selected

For now, the place type options are general to the selected place type which means some of these options may not be real options for the selected place. 

eg.
Screen Shot 2021-11-03 at 6 18 25 PM

### DIFF
--- a/static/css/tools/scatter.scss
+++ b/static/css/tools/scatter.scss
@@ -140,6 +140,7 @@
 #enclosed-place-type {
   height: 100%;
   font-size: 1rem;
+  width: 225px;
 }
 
 .plot-options-row {
--- a/static/js/tools/scatter/place_and_type_options.tsx
+++ b/static/js/tools/scatter/place_and_type_options.tsx
@@ -43,38 +43,43 @@ const USA_CHILD_PLACE_TYPES = {
   City: USA_CITY_CHILD_TYPES,
 };
 
-const NON_USA_PLACE_TYPES = [
-  "EurostatNUTS1",
-  "EurostatNUTS2",
-  "EurostatNUTS3",
+const AA4_CHILD_PLACE_TYPES = ["AdministrativeArea5"];
+const AA3_CHILD_PLACE_TYPES = ["AdministrativeArea4", ...AA4_CHILD_PLACE_TYPES];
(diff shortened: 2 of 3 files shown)


pradh:
Lay audience question:  will we at some point make it like map tool, which seems to know better:

Screen Shot 2021-11-03 at 6 21 56 PM

chejennifer:
> Lay audience question: will we at some point make it like map tool, which seems to know better:
> 
> Screen Shot 2021-11-03 at 6 21 56 PM

I think we definitely should and we could, but scatter is a little harder to tell what child places are possible for a selected place. For example, with Europe, it looks like there actually are stat vars for AdministrativeArea1/2 in Europe
Screen Shot 2021-11-03 at 6 28 42 PM

static/css/tools/scatter.scss
```diff
@@ -140,6 +140,7 @@
 #enclosed-place-type {
   height: 100%;
   font-size: 1rem;
+  width: 225px;
```
beets, line 143: nit: for a followup - we could prefer to use a more scalable unit, like rem for this (to accommodate large fonts, etc).

static/js/tools/scatter/place_and_type_options.tsx
```diff
@@ -150,11 +155,12 @@ function PlaceAndTypeOptions(): JSX.Element {
               className="pac-target-input"
             >
               <option value="">Select a place type</option>
-              {childPlaceTypes.map((type) => (
-                <option value={type} key={type}>
-                  {type}
-                </option>
-              ))}
+              {childPlaceTypes &&
```
beets, line 158: was there a case you encountered where this list was empty? it seems like it should at least be set to [] by default.

beets (approved):
thanks for the fix!

static/css/tools/scatter.scss
chejennifer: Ok, will do this in a followup

static/js/tools/scatter/place_and_type_options.tsx
chejennifer: scatter app.test.tsx kept failing complaining about it being undefined .. even though I don't think I've actually ever seen it undefined

PR #1316, opened 2021-11-04
[place] update place page climate chart titles

beets:
this diff is bigger than i would like because we hadn't updated the string bundles with the homepage changes.


beets:
localhost_8080_place_geoId_48201_topic=Climate

shifucun (approved):

pradh (approved):
Thanks for the quick change!

PR #1317, opened 2021-11-04
Refactor and update place page "category" and "topic" concepts; Have better ordering based on the grouping.

shifucun:
This PR involves some refactor and clean up of the concept in the menu:

- "category" refers to the top level grouping, which also corresponds to the menu category like "climate", "economics"
- "topic" refers to a second level of grouping. It does not have a strict correspondence to UI representation though. Topics in "Overview" page refers to the categories. Topics in non-overview page is a virtual grouping of similar charts. Examples are like "Air", "Water" in the "Climate" category. As of now, these topics are meant to aid for grouping and ordering, but they could show up as a second level menu later on.

Also updated variables and names to make the naming consistent everywhere. It was quite confusing before..


server/tests/test_data/golden_menu_text.json
```diff
@@ -0,0 +1,122 @@
...
+  ],
+  [
+    "Housing units",
+    "Housing units by householder race",
+    "Housing units by price",
+    "Housing units by date built",
+    "Housing units by rental status"
+  ],
+  [
+    "[Air] Air quality index",
```
beets, line 97: These shouldn't have the group in the titles after the string change?

static/js/place/util.ts
```diff
@@ -247,3 +248,47 @@ export function displayNameForPlaceType(
   }
   return intl.formatMessage(retMessage);
 }
+
+// Get ordered display topics give a set of chart configs.
+export function getDisplayTopics(
```
beets, line 253: wdyt of moving up this up to the server? since we are in control of the endpoints, we should be able to return the data in display order. can we add a todo? we'd have to do that along with splitting data for each page/topic.

beets (approved):
thanks - this is nice! sorry for the late review

server/tests/test_data/golden_menu_text.json
shifucun: These are the actual title displayed. It's still nice to show them?

beets: I thought it'd be just a temporary thing to work around what we had. We can work on adding the group to the left-menu which might help.

I also missed that it's now being added in code below. I only noticed the change in the strings files.

beets (approved):

static/js/place/util.ts
shifucun: I made the change and ended up with a larger refactor and clean up. They are noted in the PR description

server/tests/test_data/golden_menu_text.json
shifucun: I reverted the change to make "topic" only an internal representation but don't show it in the UI. Later on we can add another level of menu from it if needed.

server/routes/api/landing_page.py
```diff
@@ -75,23 +75,34 @@ def fill_translation(chart):
 
 
 # TODO: add test for chart_config for assumption that each combination of stat vars will only have one config in chart_config.
-def build_spec(chart_config):
+def build_spec(chart_config, i18n=True):
```
beets, line 78: could you add a comment about the param (and that it should only be used for testing?)

server/tests/landing_page_test.py
```diff
@@ -69,6 +73,20 @@ def test_chart_config_transform(self):
                 expected = json.load(f)
                 assert expected == result
 
+    def test_menu_text(self):
+        chart_config = libutil.get_chart_config()
+        spec = landing_page.build_spec(chart_config, False)
```
beets, line 78: i worry that this test will pass, even though the page, in english, will have different strings (e.g. if we are in the middle of updating titles). i was able to test the pybabel flow e2e in static_test.py, but that required loading a full request through the app. there is a ton of mocking to do though, in that case.

maybe if the test was renamed to test_menu_hierarchy?

beets (approved):
thanks for the refactor!!!

server/routes/api/landing_page.py
shifucun: Done

server/tests/landing_page_test.py
shifucun: Updated the file. I don't fully understand the context here, but it would be nice to test this in english and another language setting.

beets: since this bypases fill_translation in landing_page.py, it skips pybabel altogether and will read the title directly from the config, not the string files (that's what gettext does). in prod, strings are read from the .mo file even for english.

+1 to testing this for a couple of languages

PR #1320, opened 2021-11-05
[map] fix webdriver tests

chejennifer:
- fix broken map webdriver test
- add shared function that should be able to help with flaky tests

### DIFF
--- a/server/webdriver_tests/shared.py
+++ b/server/webdriver_tests/shared.py
@@ -17,15 +17,15 @@ from selenium.webdriver.common.by import By
 from selenium.webdriver.support import expected_conditions as EC
 
 LOADING_WAIT_TIME = 3
-
 """Common library for functions used by multiple webdriver tests"""
 
-def wait_for_loading(driver): 
-  screen_present = EC.visibility_of_element_located((By.ID, 'screen'))
-  screen_hidden = EC.invisibility_of_element_located((By.ID, 'screen'))
-  while (True):
-    try:
-      WebDriverWait(driver, LOADING_WAIT_TIME).until(screen_present)
-      WebDriverWait(driver, LOADING_WAIT_TIME).until(screen_hidden)
-    except:
-      break
+
+def wait_for_loading(driver):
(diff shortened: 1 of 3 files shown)


server/webdriver_tests/shared.py
```diff
@@ -0,0 +1,31 @@
...
+
+LOADING_WAIT_TIME = 3
+"""Common library for functions used by multiple webdriver tests"""
+
+
+def wait_for_loading(driver):
+    screen_present = EC.visibility_of_element_located((By.ID, 'screen'))
+    screen_hidden = EC.invisibility_of_element_located((By.ID, 'screen'))
+    while (True):
+        try:
```
shifucun, line 27: do not quite understand the logic here? Can you add some comment?

shifucun (approved):
Thanks for the fix, super helpful!

server/webdriver_tests/shared.py
```diff
@@ -0,0 +1,31 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from selenium.webdriver.support.ui import WebDriverWait
+from selenium.webdriver.common.by import By
+from selenium.webdriver.support import expected_conditions as EC
+
+LOADING_WAIT_TIME = 3
```
beets, line 19: it's usually good practice to add the unit to these time based constants (e.g. loading_wait_time_ms)

server/webdriver_tests/shared.py
beets: should we also add some max retries?

beets (approved):
thanks for the fix!

server/webdriver_tests/shared.py
chejennifer: added comments and max retries

server/webdriver_tests/shared.py
chejennifer: done

PR #1329, opened 2021-11-11
Allow both &topic and &category as place page url param

shifucun:


server/routes/place.py
```diff
@@ -47,14 +47,17 @@
 @bp.route('/<path:place_dcid>', strict_slashes=False)
 def place(place_dcid=None):
     dcid = flask.request.args.get('dcid', None)
-    topic = flask.request.args.get('topic', None)
+    category = flask.request.args.get('category', None)
+    if not category:
+        # Used to use topic as the url param.
+        category = flask.request.args.get('topic', None)
```
beets, line 53: I think we should also issue a redirect, if not from search, from &topic to &category so we don't increase the number of url's in the search index (and for consistency)

beets (commented):
can we also update the homepage url's? thx!

server/routes/place.py
shifucun: Always redirect to ?category now.

server/tests/place_test.py
```diff
@@ -96,6 +101,11 @@ def test_place(self, mock_get_place_type, mock_get_i18n_name):
         assert response.status_code == 200
```
beets, line 101: does this need follow_redirects=True? (likewise for the other test that still uses the topic param)

beets (approved):
thanks for adding the webdriver!

server/tests/place_test.py
shifucun: Done

PR #1331, opened 2021-11-12
[browser] small change in dcid refs

beets:
made it even lighter, and added dcid: prefix

image


shifucun (approved):
dcid prefix is good!

Issue #1323, opened 2021-11-08
[tools] Improve URL params


chejennifer:
- right now, we're passing too much information in the URL params, look into which params are actually needed

Issue #704, opened 2020-12-07
[Scatter2] clicking on "scatter2" in the header bar should clear options


beets:
### Feedback
Clicking on "Scatter2" in the header bar should reset all options and clear the URL

To reproduce:
1. Click https://staging.datacommons.org/tools/scatter2#x%3D%7B%22statVar%22%3A%7B%7D%2C%22name%22%3A%22%22%2C%22data%22%3A%5B%5D%2C%22populations%22%3A%5B%5D%2C%22log%22%3Afalse%2C%22perCapita%22%3Afalse%7D%26y%3D%7B%22statVar%22%3A%7B%7D%2C%22name%22%3A%22%22%2C%22data%22%3A%5B%5D%2C%22populations%22%3A%5B%5D%2C%22log%22%3Afalse%2C%22perCapita%22%3Afalse%7D%26place%3D%7B%22enclosingPlace%22%3A%7B%22name%22%3A%22%22%2C%22dcid%22%3A%22%22%7D%2C%22enclosedPlaceType%22%3A%22%22%2C%22enclosedPlaces%22%3A%5B%5D%2C%22lowerBound%22%3A0%2C%22upperBound%22%3A10000000000%7D
1. Click Scatter2
--> expect the same state as navigating directly to staging.datacommons.org/tools/scatter2

(in fact, manually clearing the URL field doesn't reset either, and the options are added back to the URL)

shifucun:
Now most params are cleared except for "&qd=0&ld=0"

PR #1332, opened 2021-11-16
[scatter/map] URL Params fixes

chejennifer:
- clear URL when clicking "Scatter" in the header bar [(issue #704)](https://github.com/datacommonsorg/website/issues/704)
- remove unnecessary URL params (place name and place types) for Scatter and Map tools [(issue #1323)](https://github.com/datacommonsorg/website/issues/1323)
- clean up redundant API calls having to do with place types and parent places for Scatter and Map tools [(issue #1324)](https://github.com/datacommonsorg/website/issues/1324)


static/js/tools/map/info.tsx
```diff
@@ -44,15 +44,15 @@ export function Info(): JSX.Element {
           <b>Max Temperature Relative To 2006, RCP 4.5</b> across counties in{" "}
           <a
             href={
```
shifucun, line 46: Maybe add a webdriver test for the map landing page link, can just click one (first) link and confirm chart shows up. Maybe similar for scatter plot.

static/js/tools/scatter/place_and_type_options.tsx
```diff
@@ -71,25 +77,31 @@ const CHILD_PLACE_TYPES = {
...
-  }, [place.value.enclosingPlace.dcid]);
 
   /**
-   * Reloads child places if the enclosing place or child place type changes.
+   * Watch and update place info
    */
   useEffect(() => {
-    if (place.value.enclosingPlace.dcid && _.isNull(place.value.parentPlaces)) {
+    if (!place.value.enclosingPlace.dcid) {
+      return;
```
beets, line 86: does the context need to be updated in this case?

static/js/tools/shared_util.ts
```diff
@@ -192,3 +196,30 @@ export function isChildPlaceOf(
...
+ * @param placeDcid
+ */
+export function getNamedTypedPlace(
+  placeDcid: string
+): Promise<NamedTypedPlace> {
+  if (placeDcid === EARTH_NAMED_TYPED_PLACE.dcid) {
+    return Promise.resolve(EARTH_NAMED_TYPED_PLACE);
+  }
+  const placeTypePromise = axios
+    .get(`/api/place/type/${placeDcid}`)
```
beets, line 212: should we add a new call to flask that does both of these together? possibly with the results of parent places as well

beets (approved):
thanks for the cleanup!

static/js/tools/shared_util.ts
shifucun: The new cache with all the parents, name, type information could be used here. Maybe add a TODO to use that

static/js/tools/map/info.tsx
chejennifer: done

static/js/tools/scatter/place_and_type_options.tsx
chejennifer: nothing needs to happen in this case because it means no place has been chosen

chejennifer: Nothing needs to happen in this case because it just means no place has been chosen yet

static/js/tools/shared_util.ts
chejennifer: We could, that might be more clean. But that would also mean we have to wait for one call to finish before we can call the other. Whereas by doing it this way, we get to call both asynchronously

chejennifer: added TODO

beets: re: sequential calls: these calls between flask + mixer should be much faster than the async calls from client to flask, especially since flask and mixer are in the same container. and once it's computed once, it'll be cached for future calls.

but new cache will be much faster than that :)

static/js/tools/scatter/place_and_type_options.tsx
beets: could you add a comment to that effect?

static/js/tools/shared_util.ts
chejennifer: Ah, ok, that makes sense, I'll add the new flask endpoint then and we can use later use the new cache, but I'll do this in a separate PR.

static/js/tools/scatter/place_and_type_options.tsx
chejennifer: done

PR #1334, opened 2021-11-19
[scatter] allow log scales for all domain values

chejennifer:
- before, we only allowed log scale option if the values for that axis are all positive or all negative (and no 0)
- Now:
    - if the axis has 0 as the min or max value, use log scale but map 0 to a very small non zero number
    - if the axis has both positive and negative values, use a [symlog](https://www.researchgate.net/publication/233967063_A_bi-symmetric_log_transformation_for_wide-range_data) scale 

eg. the x axis has 0 as the min value & the y axis has both positive and negative values Screen Shot 2021-11-18 at 5 06 47 PM


static/js/tools/scatter/draw_scatter.ts
```diff
@@ -110,6 +119,57 @@ function addXLabel(
...
+ * @param scaleType type of scale to return
+ * @param domainExtent the min and max of the domain
+ * @param rangeExtent the min and max of the range
+ */
+function getScale(
+  scaleType: ScaleType,
+  domainExtent: [number, number],
+  rangeExtent: [number, number]
+):
+  | d3.ScaleSymLog<number, number>
```
beets, line 133: could we create a new type that encompasses this? especially since it's repeated throughout this file

beets (approved):
sorry for missing this earlier

static/js/tools/scatter/draw_scatter.ts
chejennifer: done

PR #1347, opened 2021-12-01
[browser] fix webdriver test

chejennifer:
- webdriver tests failed because with the series ranking update for `stat/set/series/all` API endpoint ([from this pr](https://github.com/datacommonsorg/mixer/pull/665)), the order of series shown on the browser page changed & so the redirected page also changed


beets (approved):
thanks for the fix!

PR #1349, opened 2021-12-09
Use recon API to get dcid from place id

shifucun:
This completely removes the need of using placeid2dcid.json in website! A very big simplification for maintenance


server/configmodule.py
```diff
@@ -14,7 +14,8 @@ class Config:
     VERSION = '{}-{}'.format(os.environ.get('WEBSITE_HASH'),
                              os.environ.get('MIXER_HASH'))
     API_ROOT = 'http://127.0.0.1:8081'  # Port for Kubernetes ESP.
-    GCS_BUCKET = os.environ.get('GCS_BUCKET') or ''
+    # TODO: Use prod recon api in prod.
+    RECON_API_URL = 'https://autopush.recon.datacommons.org/id/resolve'
```
shifucun, line 18: @Spaceenter can we use prod recon service for prod website now?

shifucun:
@Spaceenter @pradh can you take a look at this change too?

server/configmodule.py
pradh: +1 to using prod here.

The recon service also uses the BT cache, and stores a version.

Suppose we make a DCID map change, what's the rollout process?  For example, during go/dcrelease, should we do something like this:

1. Send PR to update the cache version in recon service
2. Push to staging recon service
3. Validate staging website (which uses staging.recon)
4. Push to prod recon

pradh (commented):
Nice change!

server/configmodule.py
shifucun: Updated mixer to serve recon APIs here as well. So there is no special handling for the recon APIs.

shifucun:
Finally updated this PR now as recon service is merged into mixer.

beets (approved):
thanks bo!

server/configmodule.py
```diff
@@ -14,7 +14,7 @@ class Config:
     VERSION = '{}-{}'.format(os.environ.get('WEBSITE_HASH'),
                              os.environ.get('MIXER_HASH'))
     API_ROOT = 'http://127.0.0.1:8081'  # Port for Kubernetes ESP.
-    GCS_BUCKET = os.environ.get('GCS_BUCKET') or ''
+    RECON_API_ROOT = 'http://127.0.0.1:8081'
```
pradh, line 17: I see, so this is the instance of Recon+Mixer API service running co-located with Website?

pradh (approved):
Very nice!

(We should also update release doc)

server/configmodule.py
shifucun: Yes, that's right.

Issue #1357, opened 2021-12-14
[map] title does not change when per capita option is selected


beets:
In scatter, we update the title of the chart to reflect that it's now showing per capita data. We don't do this for map.

https://datacommons.org/tools/map#%26sv%3DCount_Person_Female_BelowPovertyLevelInThePast12Months%26pc%3D1%26denom%3DCount_Person%26pd%3Dcountry%2FUSA%26ept%3DState

On that page, the expected title is: Female, Below Poverty Level In the Past 12 Months Population Per Capita (2019)

PR #1358, opened 2021-12-15
[fix: map] title does not change when per capita option is selected

sharadshriram:
This PR attempts to resolve issue #1357 

In my local tests, I am able to see the desired chart title captured in the following screenshots

## Before the proposed change
image

### After review comments incorporation
image

### Before review
image


static/js/tools/map/chart.tsx
```diff
@@ -89,10 +89,11 @@ export function Chart(props: ChartProps): JSX.Element {
   const statVarInfo = props.statVar.value;
   const [errorMessage, setErrorMessage] = useState("");
   const [denomInput, setDenomInput] = useState(props.statVar.value.denom);
-  const title = getTitle(
+  const chart_title=   getTitle(
     Array.from(props.dates),
     statVarInfo.info.title ? statVarInfo.info.title : statVarInfo.dcid
   );
+  const title = props.statVar.value.perCapita? 'Per Capita ' + chart_title : chart_title;
```
chejennifer, line 96: In Scatter Tool, we append "Per Capita" to the end of the stat var name. To keep things consistent, let's move "Per Capita" to the end of the chart title?

chejennifer (commented):
Thanks for the fix!

static/js/tools/map/chart.tsx
beets, line 96: please follow the convention from scatter (per capita comes after the sv name)

this edit also belongs to the getTitle function, passing in props.statVar.value.perCapita

also, as an fyi - we use camelcase in typescript

beets (commented):
thanks for the bug fix!

static/js/tools/map/chart.tsx
sharadshriram: Thank you for the review. I have moved 'Per Capita' to the end of the StatVar's name.

sharadshriram: Thank you for the review. I have moved 'Per Capita' to the end of the StatVar's name with the edit being made in the `getTitle` function.

chejennifer: With the edit in the `getTitle` function, line 97 should be removed. And maybe change `chart_title` back to `title`?

sharadshriram: Thanks for the review, Jennifer. I have cleaned up the code. This particular issue when I tried resolving a git divergent branches error. :)

chejennifer:
Thanks for the changes! There are some linting issues causing tests to fail. Running `./run_test -f` should fix those issues!

chejennifer (approved):

PR #1362, opened 2021-12-21
[Place Page] Only process data for the current topic

shifucun:
The entire cache data is still load once in the web server, however only the data for the current category is processed.

This also removes the previous behavior that merges all the categories into overview when total number of charts is too small.

Some other updates:

- Removes the new stat vars for place page as they are already in cache.

### DIFF
--- a/server/routes/api/landing_page.py
+++ b/server/routes/api/landing_page.py
@@ -41,11 +41,11 @@ OVERVIEW = 'Overview'
 
 
 def get_landing_page_data(dcid, new_stat_vars):
-    return get_landing_page_data_helpfer(dcid, '^'.join(new_stat_vars))
+    return get_landing_page_data_helper(dcid, '^'.join(new_stat_vars))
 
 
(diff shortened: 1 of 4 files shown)


server/routes/api/landing_page.py
```diff
@@ -42,10 +41,16 @@
 
 
 def get_landing_page_data(dcid, new_stat_vars):
-    response = dc_service.fetch_data('/landing-page', {
-        'place': dcid,
-        'newStatVars': new_stat_vars,
-    },
+    return get_landing_page_data_helpfer(dcid, '^'.join(new_stat_vars))
```
chejennifer, line 44: helpfer -> helper

chejennifer (approved):
Nice changes!

server/routes/api/landing_page.py
shifucun: Fixed

beets (commented):
nice!

PR #1363, opened 2021-12-21
Add topic page proto and config; Read the config on server start

shifucun:


server/topic_page.proto
```diff
@@ -0,0 +1,83 @@
...
+  bool show_decrease = 4;
+  string diff_base_date = 5; // For comparisons
+}
+
+message HighlightMetadata {
+  string description = 1;
+}
+
+message Tile {
+  enum TitleType {
```
chejennifer, line 53: TitleType -> TileType?

chejennifer (commented):
Nice!

chejennifer (approved):

server/topic_page.proto
shifucun: Fixed

shifucun:
Thanks for review!

server/topic_page_config.textproto
```diff
@@ -0,0 +1,98 @@
+# Copyright 2021 Google LLC
```
beets, line 1: we'll have a set of these, one per topic x country, so would be nice to move them into a separate folder as we add more config files.

beets (commented):
yay, protos :)

PR #1364, opened 2021-12-21
Add skeleton of topic page

shifucun:
This PR adds html template, react entrance, http handlers for topic page.


chejennifer (commented):
Thanks for creating the skeleton!

server/routes/topic_page.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+@bp.route('', strict_slashes=False)
+@bp.route('/<path:place_dcid>', strict_slashes=False)
+def topic_page(place_dcid=None):
+    if os.environ.get('FLASK_ENV') == 'production':
+        flask.abort(404)
+    topic = flask.request.args.get('topic', '')
+    # TODO: should use place metadata API to fetch these data in one call.
+    place_type = place_api.get_place_type(place_dcid)
```
chejennifer, line 31: should first check that place_dcid is not None. I think both `place_api.get_place_type` and `place_api.get_i18n_name` will throw an error if place_dcid is None.

shifucun: Added a check.

chejennifer (approved):

server/routes/topic_page.py
```diff
@@ -0,0 +1,43 @@
...
+
+import flask
+import os
+import routes.api.place as place_api
+
+bp = flask.Blueprint('topic_page', __name__, url_prefix='/topic')
+
+
+@bp.route('', strict_slashes=False)
+@bp.route('/<path:place_dcid>', strict_slashes=False)
```
beets, line 24: we should add topic to the path -- since each page will be about a specific topic:

/<topic:string>/<path:place_dcid>

static/js/topic_page/topic_page.ts
```diff
@@ -0,0 +1,43 @@
...
+import { MainPane } from "./main_pane";
+
+window.onload = () => {
+  renderPage();
+};
+
+function renderPage(): void {
+  const urlParams = new URLSearchParams(window.location.search);
+  // Get topic and render menu.
+  const topic = urlParams.get("topic");
```
beets, line 29: as we pass the config proto to the client, we can read the topic from there too.

beets (commented):
thanks for adding the framework!

Issue #1303, opened 2021-11-04
[BUG] In map view, if the enclosed place type is not available for a selected place, should clear the place type


chejennifer:
https://staging.datacommons.org/tools/map#%26sv%3DCount_Person%26pc%3D0%26denom%3DCount_Person%26pd%3Dcountry%2FUKR%26pn%3DUkraine%26pt%3DCountry%26ept%3DEurostatNUTS1

chejennifer:
new vis tool has a new flow

PR #1367, opened 2022-01-01
[Fix] In map view, if the enclosed place type is not available for a selected place, should clear the place type

sharadshriram:
This PR resolves the bug posted in issue #1303.

After the code changes are merged, the following are the expected results on the map explorer webpage:

https://user-images.githubusercontent.com/5391555/148154621-b0b969d5-1e95-474d-9178-dedd71d7aee9.mov


static/js/tools/map/place_options.tsx
```diff
@@ -242,6 +242,11 @@ function loadEnclosedPlaces(
           `Sorry, ${place.value.enclosingPlace.name} does not contain places of type ` +
             `${enclosedPlaceType}. Try picking another type or place.`
         );
+        document.getElementById("place-list").innerHTML = "";
```
chejennifer, line 245: I don't think we want to clear the selected place because the alert came after user selected a place type so it should only clear the place type.

static/js/tools/map/place_options.tsx
```diff
@@ -242,6 +242,11 @@ function loadEnclosedPlaces(
...
+        const select = document.getElementById(
+          "enclosed-place-type"
+        ) as HTMLInputElement;
+        select.value = "";
```
chejennifer, line 249: instead of directly manipulating the DOM, we should be updating the context which stores information about the options that have been chosen for this tool because we want the context to be in sync with what is displayed. PlaceInfoWrapper has a function called "setEnclosedPlaceType" which allows the enclosedPlaceType to be updated so you should be able to just do `place.setEnclosedPlaceType("")`

static/js/tools/scatter/place_and_type_options.tsx
```diff
@@ -256,6 +256,11 @@ async function loadPlaces(
             `Sorry, ${place.value.enclosingPlace.name} does not contain places of type ` +
               `${childPlaceType}. Try picking another type or place.`
           );
+          document.getElementById("place-list").innerHTML = "";
```
chejennifer, line 259: same comments as those for map

chejennifer (commented):
thanks for the changes!

static/js/tools/map/place_options.tsx
sharadshriram: Thanks for the comment, I have modified the code such that the place type is only cleared and the `place` is not modified.

static/js/tools/map/place_options.tsx
sharadshriram: Got it! I have removed the direct DOM manipulation with `place.setEnclosedPlaceType("")`

static/js/tools/scatter/place_and_type_options.tsx
sharadshriram: In the scatter, the clearing of the place-type is already done -- hence, reverting this file in this PR.

chejennifer (approved):
Thanks for updating!

PR #1369, opened 2022-01-04
Add deployment config for tidal

shifucun:
Also generalized the private deployment to make it able to extend (still a bit lengthy and worth some refactor later on).


beets (approved):
We might want to bring back the "blue" private dc mode for Tidal and others, or add it as part of the config :) Not blocking though.

shifucun:
> We might want to bring back the "blue" private dc mode for Tidal and others, or add it as part of the config :) Not blocking though.

Also as we have more use case, can generalize the way to customize page style and text too.

PR #1370, opened 2022-01-04
[topic page] add map tile component

chejennifer:
component that fetches and processes data, and renders the map tile

caveats 
- was not sure if we wanted to make the regions on the map clickable/where they should redirect to, so left it as unclickable for now
- tooltip content is the same as place explorer maps for now. Not sure if we wanted this or similar to the map explorer tool (map explorer tool will have additional information on the date of the population data if it differs from the date of the stat value for per capita)

example (from dummy prop values and the width/layout of the tiles will later be done in the parent Block component):
Screen Shot 2022-01-04 at 2 49 51 PM


static/js/tools/map/util.ts
```diff
@@ -122,6 +128,13 @@ export const ENCLOSED_PLACE_TYPE_NAMES = {
 export const USA_PLACE_HIERARCHY = ["Country", "State", "County"];
 export const MAP_REDIRECT_PREFIX = "/tools/map";
 
+export interface DataPointMetadata {
```
beets, line 131: please add brief comments for these, especially for shared interfaces

static/js/topic_page/map_tile.tsx
```diff
@@ -0,0 +1,260 @@
...
+    .then(([geoJsonData, placeStatData, populationData]) => {
+      setRawData({
+        geoJsonData,
+        placeStatData: placeStatData.data[statVarDcid],
+        metadataMap: placeStatData.metadata,
+        populationData,
+      });
+    })
+    .catch(() => {
+      setRawData(null);
```
beets, line 162: Add a TODO for an error message in this case?

beets (approved):
I had imagined that the data will be fetched per block (e.g. if we combine a map tile with ranking, they would use the same data). It's hard to glean that from a block, though, so we could also update the definition of a block to include time series data or not.

static/js/topic_page/map_tile.tsx
```diff
@@ -0,0 +1,260 @@
...
+  geoJsonData: GeoJsonData;
+  chartTitle: string;
+}
+
+export function MapTile(props: MapTilePropType): JSX.Element {
+  const [rawData, setRawData] = useState<RawData | undefined>(null);
+  const [chartData, setChartData] = useState<ChartData | undefined>(null);
+  const sourcesJsx = chartData ? getSourcesJsx(chartData.sources) : [];
+
+  useEffect(() => {
```
shifucun, line 77: I really like the structure here to use "fetchData", "processData" and "drawMap". All other tiles can follow same pattern, which makes the maintenance easier.

shifucun (approved):

shifucun:
> I had imagined that the data will be fetched per block (e.g. if we combine a map tile with ranking, they would use the same data). It's hard to glean that from a block, though, so we could also update the definition of a block to include time series data or not.

I also discussed (debated) this with Jennifer on should "block" or "tile" fetch data. Then I was convinced that letting tile fetch data might be better at this stage.

- Biggest advantage is the flexibility to add more tiles in different blocks, as they are all independent.
- There will be redundant API calls within the same block. But they are all raw mixer API calls that are cached with small data size.
- We can optimize the performance later if that becomes an issue.

static/js/tools/map/util.ts
chejennifer: done

static/js/topic_page/map_tile.tsx
chejennifer: done

static/js/tools/map/util.ts
```diff
@@ -29,8 +29,14 @@ import {
   PAKISTAN_PLACE_DCID,
   USA_PLACE_DCID,
 } from "../../shared/constants";
+import { StatApiResponse } from "../../shared/stat_types";
 import { NamedPlace } from "../../shared/types";
-import { isChildPlaceOf } from "../shared_util";
+import {
```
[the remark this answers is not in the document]
chejennifer: Agreed, but this can be in a separate PR/for future PRs

static/js/topic_page/map_tile.tsx
```diff
@@ -0,0 +1,260 @@
...
+  statVarDcid: string;
+  chartTitle: string;
+  isUsaPlace: boolean;
+  denomDcid?: string;
+  unit?: string;
+  scaling?: number;
+}
+
+interface RawData {
+  geoJsonData: GeoJsonData;
```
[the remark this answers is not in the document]
chejennifer: done

PR #1372, opened 2022-01-06
[scatter] Add date param

chejennifer:
- Add date url param to control the date for the data for each axis in the plot. (param keys are "datex" and "datey")
- If date is specified, include it in the title

Notes
- not sure if including the dates in the title is useful or confusing (would love some thoughts)
    - Could be useful to quickly see that all the data will be from this specific date. Otherwise, will need to hover over the points on the graph to see the date of the data in the tooltip.
    - Could be confusing if sometimes there is a date in the title and sometimes there isn't

Screen Shot 2022-01-06 at 2 05 28 PM

### DIFF
--- a/static/js/tools/scatter/chart.tsx
+++ b/static/js/tools/scatter/chart.tsx
@@ -51,7 +51,8 @@ interface ChartPropsType {
   yUnits?: string;
   placeInfo: PlaceInfo;
   display: DisplayOptionsWrapper;
-  date: string;
+  xDate: string;
+  yDate: string;
 }
 
 const DOT_REDIRECT_PREFIX = "/place/";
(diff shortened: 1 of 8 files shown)


static/js/tools/scatter/chart.tsx
```diff
@@ -51,6 +51,8 @@ interface ChartPropsType {
   yUnits?: string;
   placeInfo: PlaceInfo;
   display: DisplayOptionsWrapper;
+  xDate: string;
```
beets, line 54: could you clarify here that this is the requested date (and not the date from the data). or a comment for the interface that these are all requested chart params.

static/js/tools/scatter/chart.tsx
```diff
@@ -200,9 +202,15 @@ function Chart(props: ChartPropsType): JSX.Element {
       <Row>
         <Card id="no-padding">
           <div className="chart-title">
-            <h3>{props.yLabel}</h3>
+            <h3>
+              {props.yLabel}
+              {props.yDate && ` (${props.yDate})`}
```
beets, line 207: if it's the requested date, we might not be able to use it directly in the title, and it might be better to use the dates from data like we do elsewhere.

beets (approved):
thanks for the cleanup too.

qq regarding what we should put in the title: i think it's better to use the date from data -- thoughts?

chejennifer:
> thanks for the cleanup too.
> 
> qq regarding what we should put in the title: i think it's better to use the date from data -- thoughts?

So when there is no date specified and we're just getting the latest data, do we want to include the date range for each stat var in the title?

beets:
Consistency would be good so we should use dates from data -- we could think of other indicators (e.g. we don't specify in the chart that we are filtering places by population, though that would be helpful).

chejennifer:
> Consistency would be good so we should use dates from data -- we could think of other indicators (e.g. we don't specify in the chart that we are filtering places by population, though that would be helpful).

Sounds good, updated to always show dates in the title.
eg. when using latest: 
Screen Shot 2022-01-06 at 4 59 32 PM

beets (approved):
thanks for the update!

PR #1374, opened 2022-01-10
Update cloudbuild node version to 16.13.1

shifucun:


beets (approved):

shifucun:
This was the cause of the i18n test failure.. node 12 was apparently too old..

shifucun:
Thanks for review!

PR #1375, opened 2022-01-10
Add the bare bone for ranking tile

shifucun:
This allows building the Block component for completeness.

Also fixes a state issue in line tile, where "chartData" updates causes an extra rendering cycle.

### DIFF
--- a/static/js/topic_page/line_tile.tsx
+++ b/static/js/topic_page/line_tile.tsx
@@ -44,30 +44,34 @@ export function LineTile(props: LineTilePropType): JSX.Element {
 
   useEffect(() => {
     fetchData(props, setRawData);
-  }, []);
+  }, [props]);
 
   useEffect(() => {
     if (rawData) {
       processData(props, rawData, setChartData);
     }
-  }, [rawData]);
+  }, [props, rawData]);
 
   useEffect(() => {
     if (chartData) {
       draw(props, chartData);
     }
-  });
+  }, [props, chartData]);
 
   return (
     <div className="chart-container">
-      <div className="line-title">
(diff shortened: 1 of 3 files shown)


static/js/topic_page/ranking_tile.tsx
```diff
@@ -0,0 +1,71 @@
...
+    }
+  }, [rawData]);
+
+  return (
+    <div className="chart-container">
+      {rankingData && <>{Object.keys(rankingData).join(",")}</>}
+    </div>
+  );
+
+  function fetchData(statVarMetaData: StatVarMetadata): void {
```
chejennifer, line 56: It's better to put helper functions outside of the react component when possible so that they're not re-created on every re-render. We might want to keep these outside of the component.

chejennifer (approved):

static/js/topic_page/line_tile.tsx
```diff
@@ -56,22 +56,18 @@ export function LineTile(props: LineTilePropType): JSX.Element {
     if (chartData) {
       drawChart(chartData);
     }
-  }, [chartData]);
```
beets, line 59: merge error? should keep the reference to chartData

beets (approved):

static/js/topic_page/ranking_tile.tsx
shifucun: Done

static/js/topic_page/line_tile.tsx
shifucun: I intended to fix some deps lint warnings. Completed them now.

PR #1376, opened 2022-01-10
Align the Component style for line and map tile

shifucun:
- Use statVarMetadata as Component prop field.
- Move helper functions into the Component scope.
- Consolidate prop naming
- Set "isUSAPlace" to true always

### DIFF
--- a/static/js/topic_page/map_tile.tsx
+++ b/static/js/topic_page/map_tile.tsx
@@ -48,6 +48,7 @@ interface MapTilePropType {
   title: string;
   placeDcid: string;
   enclosedPlaceType: string;
+  isUsaPlace: boolean;
   statVarMetadata: StatVarMetadata;
 }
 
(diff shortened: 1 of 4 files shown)


static/js/topic_page/map_tile.tsx
```diff
@@ -121,141 +112,143 @@ export function MapTile(props: MapTilePropType): JSX.Element {
       )}
     </div>
   );
-}
```
chejennifer, line 124: It's better to put helper functions outside of the react component when possible so that they're not re-created on every re-render. Since these helper functions can exist outside of the component, I think we should keep them outside of the component.

chejennifer (approved):

static/js/topic_page/map_tile.tsx
```diff
@@ -121,141 +112,143 @@ export function MapTile(props: MapTilePropType): JSX.Element {
...
+      width,
+      chartData.dataValues,
+      props.statVarMetadata.unit,
+      colorScale,
+      _.noop,
+      getTooltipHtml,
+      () => false,
+      true,
+      true,
+      true,
```
beets, line 231: we will need to pass in the bool isUsaPlace, since we'll want these pages for non-us places (we will have for india as a start)

beets (approved):

static/js/topic_page/map_tile.tsx
shifucun: That's a good point!

static/js/topic_page/map_tile.tsx
shifucun: Reverted

PR #1379, opened 2022-01-11
Add Block componenet and get topic page working end to end

shifucun:
Page config is stored in initial page load as an html data field. Then each block (tile) renders async with their own data fetching.

Things to add later:
- real css
- ranking tile
- bar tile
- better config for place type, child place type and similar place type; place based stat vars..

### DIFF
--- a/static/js/topic_page/block.tsx
+++ b/static/js/topic_page/block.tsx
@@ -32,6 +32,7 @@ export interface BlockPropType {
   placeDcid: string;
   enclosedPlaceType: string;
   title: string;
+  description: string;
   leftTiles: Tile[];
   rightTiles: Tile[];
   statVarMetadata: StatVarMetadata;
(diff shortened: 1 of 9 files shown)


beets (approved):
nice!

static/js/topic_page/block.tsx
```diff
@@ -0,0 +1,94 @@
...
+import React from "react";
+
+import { randDomId } from "../shared/util";
+import { StatVarMetadata } from "../types/stat_var";
+import { LineTile } from "./line_tile";
+import { MapTile } from "./map_tile";
+import { RankingTile } from "./ranking_tile";
+import { Tile } from "./tile";
+
+export interface BlockPropType {
```
chejennifer, line 30: from the proto, it looks like a block can also contain a description, probably want to include a field for that here in the BlockPropType

chejennifer (approved):
nice!

static/js/topic_page/block.tsx
shifucun: Done

shifucun:
Thanks!

PR #1380, opened 2022-01-12
Implement ranking tile with highest and lowest comparison

shifucun:
I fetch the raw stat and do the ranking instead of calling the ranking API as this involves computation between stat vars. Speed wise fairly good. This PR has minimal css layout just to get the data in.


beets (approved):
thanks!!

chejennifer (approved):
nice!

PR #1381, opened 2022-01-12
[Topic Page] Add bar tile component

shifucun:


static/js/topic_page/bar_tile.tsx
```diff
@@ -0,0 +1,160 @@
+/**
+ * Copyright 2021 Google LLC
```
beets, line 2:
```suggestion
 * Copyright 2022 Google LLC
```
:)

beets (approved):
thanks!

static/js/topic_page/bar_tile.tsx
shifucun: Oh!

static/js/topic_page/bar_tile.tsx
```diff
@@ -0,0 +1,160 @@
...
+
+  useEffect(() => {
+    fetchData(props, setRawData);
+  }, [props]);
+
+  useEffect(() => {
+    if (rawData) {
+      processData(props, rawData, setChartData);
+    }
+  }, [props, rawData]);
```
chejennifer, line 57: I think we would only need to re-run processData when rawData changes so I don't think this useEffect needs to also listen for props changes. Same with the following useEffect that does the drawing, I think we only need to re-run draw when the chartData changes

chejennifer (approved):
nice!

static/js/topic_page/bar_tile.tsx
shifucun: This is from the eslint-plugin-react-hooks plugin. There are some discussions about in this thread https://github.com/facebook/create-react-app/issues/6880#issuecomment-485912528.

I yet to find sometime to fully understand it, but from initial impression, it's better to include any props, state that this hook refers.

chejennifer: oh, interesting, I didn't know that!

shifucun: If a prop change, it should be a re-render from parent component, then this hook should be triggered too.

PR #1382, opened 2022-01-12
[Topic Page] Add highlight tile

shifucun:
Note, since overview block has no difference with the other block, merge it into the general blocks.


chejennifer (approved):
looks good!

beets:
thanks!

the thought with the overview block was to call it out (and not assume the first block is the overview) since it's treated in the UI. but in all likelihood the first will always be the overview, so we can keep this.

shifucun:
> 

Thanks for the clarification!

PR #1383, opened 2022-01-12
Topic page styling v1

beets:
A starting point for styling changes. Sending this before this PR gets even bigger. Most large layout changes are in, though there are some still missing, namely the header, and placeholder titles, etc.

image


static/js/topic_page/bar_tile.tsx
```diff
@@ -42,37 +43,41 @@ interface BarTilePropType {
...
+    return null;
+  }
   return (
-    <div className="chart-container">
-      {chartData && (
-        <>
-          <div className="line-title">
-            <h4>{props.title}</h4>
-          </div>
+    <ChartTileContainer
```
shifucun: Here has no indent, whether should run the lint fix.

shifucun (approved):
This looks super cool!! Thanks for bearing with my incoming PR while you doing this :)

static/js/topic_page/bar_tile.tsx
beets: thanks, just ran lint

PR #1385, opened 2022-01-13
[topic page] add scatter tile

chejennifer:
- added a scatter tile
- tried to move a few files/functions around and added a utils folder

Questions:
- in our config, should scaling and units be instead attached to each statvar? Because in scatter, I think it would make sense for us to want different scaling and units for each axis.
- Also in our config, do we want to allow for log as well? Because I know sometimes log makes the charts look nicer.

Example of scatter tile (just using same config as another tile):
Screen Shot 2022-01-13 at 12 25 17 PM

### DIFF
--- a/static/css/topic_page.scss
+++ b/static/css/topic_page.scss
@@ -16,6 +16,7 @@
 @import "./base";
 @import "./draw";
 @import "./draw_choropleth";
+@import "./draw_scatter";
 @import "./shared/story_block";
 @import "./shared/story_chart";
 
@@ -92,29 +93,6 @@ $chart-container-top-margin: 20px;
 .highlight-tile .desc {
   padding-left: .4em;
   font-size: 1.2rem;
-}
-
-.scatter-tooltip {
-  position: absolute;
-  background-color: var(--dc-gray-lite);
-  box-shadow: 0 3px 6px 0 rgba(0, 0, 0, 0.2);
-  font-size: 0.75rem;
-  padding: 0.4rem;
-  border-radius: 8px;
-  opacity: 0.9;
-  z-index: 2;
-  border: 0.5px solid var(--dc-gray-lite);
-  word-break: break-word;  
-  white-space: normal;
-}
-
-.scatter-tooltip header {
-  font-weight: bold;
-  margin-bottom: 0.4rem;
-}
-
-.scatter-tooltip footer {
-  margin-top: 0.4rem;
 }
 
 .scatter-svg-container {
--- a/static/css/topic_page.scss
+++ b/static/css/topic_page.scss
@@ -93,3 +93,39 @@ $chart-container-top-margin: 20px;
   padding-left: .4em;
   font-size: 1.2rem;
 }
+
+.scatter-tooltip {
+  position: absolute;
+  background-color: var(--dc-gray-lite);
+  box-shadow: 0 3px 6px 0 rgba(0, 0, 0, 0.2);
+  font-size: 0.75rem;
+  padding: 0.4rem;
+  border-radius: 8px;
+  opacity: 0.9;
+  z-index: 2;
+  border: 0.5px solid var(--dc-gray-lite);
+  word-break: break-word;  
+  white-space: normal;
+}
+
+.scatter-tooltip header {
+  font-weight: bold;
+  margin-bottom: 0.4rem;
+}
+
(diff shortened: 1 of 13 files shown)


static/css/topic_page.scss
```diff
@@ -92,4 +92,40 @@ $chart-container-top-margin: 20px;
 .highlight-tile .desc {
   padding-left: .4em;
   font-size: 1.2rem;
+}
+
+.scatter-tooltip {
```
beets, line 97: we should try to align all our tooltips site wide :) are these truly specific to scatter? if so, we can move it to a scatter specific scss file.

static/js/topic_page/scatter_tile.tsx
```diff
@@ -0,0 +1,258 @@
...
+}
+
+function fetchData(
+  placeDcid: string,
+  enclosedPlaceType: string,
+  statVarMetadata: StatVarMetadata,
+  setRawData: (data: RawData) => void
+): void {
+  if (statVarMetadata.statVars.length < 2) {
+    return;
```
beets, line 131: todo: add a warning / error.

static/js/utils/number_utils.ts
```diff
@@ -0,0 +1,41 @@
+/**
```
beets, line 1: can we keep these in js/shared/ instead of adding a new utils folder?

beets (approved):
nice refactors!

re questions: +1 to both. we can add log to the scatter options, and update units to be per sv. we can do both as followups.

static/js/utils/number_utils.ts
chejennifer: right now, the shared folder has constants, utils, types, and various different kinds of files. I think it might make more sense to create this utils folder at this level and start moving files out of the shared folder (seeing how we also have a types folder at this level). I think the name shared doesn't tell us much about what's in the folder and if we were to clean up the contents of that folder into subfolders, that will create unnecessary additional nesting. What do you think?

static/css/topic_page.scss
chejennifer: I agree with aligning all our tooltips site wide, but that will be for a future pr (right now we have different tooltips for map, scatter, line chart, and stat var hierarchy)

For now, I've moved to a scatter specific scss file.

static/js/topic_page/scatter_tile.tsx
chejennifer: done

static/js/chart/draw_scatter.ts
```diff
@@ -536,18 +550,37 @@ function addRegressionLine(
...
 export function drawScatter(
-  svgContainerRef: React.MutableRefObject<HTMLDivElement>,
-  tooltipRef: React.MutableRefObject<HTMLDivElement>,
+  svgContainerRef: React.RefObject<HTMLDivElement>,
+  tooltipRef: React.RefObject<HTMLDivElement>,
   chartWidth: number,
   chartHeight: number,
-  props: ChartPropsType,
+  points: { [placeDcid: string]: Point },
+  yLabel: string,
```
shifucun, line 579: Maybe have a struct to hold "chartWidth", "chartHeight", "x/y Label", "x/y Unit"? In general, we prob should reduce the number of argument of functions in general..

shifucun (approved):

static/css/topic_page.scss
beets: yes, definitely in a future pr

static/js/utils/number_utils.ts
beets: ok, though we might want a more specific name than utils (maybe data_utils? or chart_utils)? let's keep in mind that want to avoid getting it to be too much of a grab bag of functionality.

beets (approved):

static/js/utils/number_utils.ts
chejennifer: maybe we can keep it like this for now and as more files get added to the folder, we'll have a better idea of how to split it up?

static/js/chart/draw_scatter.ts
chejennifer: done

static/js/utils/number_utils.ts
beets: sure

PR #1386, opened 2022-01-14
[tree widget] fix redundant api calls bug

chejennifer:
- fix bug where large number of redundant `api/stats/stat-var-summary` and `/api/browser/statvar/group` were being made


shifucun (approved):
Thanks for the fix!!

PR #1387, opened 2022-01-18
[topic page] add bivariate tile

chejennifer:
- added a bivariate tile
- extracted bivariate chart drawing into its own file

Example of bivariate tile (just using same config as another tile):
Screen Shot 2022-01-18 at 12 29 51 PM

### DIFF
--- a/static/js/topic_page/bivariate_tile.tsx
+++ b/static/js/topic_page/bivariate_tile.tsx
@@ -1,0 +1,332 @@
+/**
+ * Copyright 2022 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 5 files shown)


static/js/topic_page/bivariate_tile.tsx
```diff
@@ -0,0 +1,332 @@
...
+  geoJson: GeoJsonData;
+  placeStats: GetStatSetResponse;
+  population: StatApiResponse;
+  placeNames: { [placeDcid: string]: string };
+  place: NamedTypedPlace;
+  parentPlaces: NamedTypedPlace[];
+}
+
+interface ChartData {
+  xStatVar: {
```
beets, line 65: can we use StatVarMetadata here (or extract the needed part and reuse that). add a todo for that, so it can go w/ the proto change we talked about

beets (approved):
nice, thanks for the refactor!

static/js/topic_page/bivariate_tile.tsx
chejennifer: yes, I was planning to update that to StatVarMetadata when making the proto change. But added a todo for now

PR #1389, opened 2022-01-19
[topic page] update config proto

chejennifer:
- attach unit, scaling, and log to each stat var instead of to a list of stat vars

### DIFF
--- a/static/js/topic_page/ranking_tile.tsx
+++ b/static/js/topic_page/ranking_tile.tsx
@@ -30,7 +30,9 @@ import { RankingMetadata } from "./tile";
 const RANKING_COUNT = 5;
 
 interface RankingData {
-  [key: string]: Point[]; // Key is main statVarDcid
+  points: Point[];
+  unit: string;
+  scaling: number;
 }
 
 interface RankingTilePropType {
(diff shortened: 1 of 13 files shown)


static/js/topic_page/ranking_tile.tsx
```diff
@@ -30,20 +30,24 @@ import { RankingMetadata } from "./tile";
...
   title: string;
-  statVarMetadata: StatVarMetadata;
+  statVarMetadata: StatVarMetadata[];
   rankingMetadata: RankingMetadata;
 }
 
 export function RankingTile(props: RankingTilePropType): JSX.Element {
-  const [rankingData, setRankingData] = useState<RankingData | undefined>(null);
+  const [rankingData, setRankingData] = useState<
+    Record<string, RankingData> | undefined
```
beets, line 49: nit: i might rename the current RankingData to RankingPoint, and wrap the Record<string, RankingPoint> to RankingData. it'll be easier to document the keys of RankingData in that way, update types, etc. we could also pass RankingPoint to RankingUnit component directly.

beets (approved):
thanks for the refactor!

static/js/topic_page/ranking_tile.tsx
chejennifer: I agree with this approach but I'm not sure RankingPoint is the best name for it because it holds the values for multiple points in the ranking. Maybe RankingGroup?

chejennifer: done

PR #1390, opened 2022-01-19
Fix local runs of webdriver tests with Python 3.8+

beets:
When running webdriver tests on MacOS using Python 3.8+, tests fail with:
```
AttributeError: Can't pickle local object 'LiveServerTestCase._spawn_live_server.<locals>.worker'
```

There was an [update in Python 3.8](https://docs.python.org/3/library/multiprocessing.html#contexts-and-start-methods) where the multiprocessing start method default was changed from 'fork' to 'spawn'. I came across that hint on a [pytest issue](https://github.com/pytest-dev/pytest-flask/issues/104)


shifucun (approved):
nice find!

PR #1392, opened 2022-01-20
Support multiple topic configs

beets:
- created a top-level config folder for both chart config and topic pages (we should rename the former to place explorer..)
- add support for loading configs as protos, and updating URL paths to /topic/<topic_id>/<place_dcid>
- add placeholder topic landing page
- add placeholder climate config (buggy)


server/config/topic_page/climate/USA.textproto
```diff
@@ -0,0 +1,36 @@
+# Copyright 2021 Google LLC
```
shifucun, line 1: 2022

server/config/topic_page/poverty/USA.textproto
```diff
@@ -12,8 +12,14 @@
...
 # limitations under the License.
 
-## This file specifies the config for topic page
+## This file specifies the topic page config for poverty in USA
 
+metadata {
+  topic_id: 'poverty'
+  topic_name: 'Poverty'
+  place_dcid: 'country/USA'
+  place_dcid: 'geoId/06'
```
shifucun, line 21: what is geoId/06 used for here?

shifucun (approved):

server/config/topic_page/climate/USA.textproto
```diff
@@ -0,0 +1,36 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+## This file specifies the topic page config for climate in USA
+
+metadata {
+  topic_id: 'climate'
+  topic_name: 'Climate'
+  place_dcid: 'country/USA'
+  place_dcid: 'geoId/06'
```
chejennifer, line 21: I noticed in the design doc that there is an additional field in metadata for a map of contained_place_types. Is that something we're still going to include? I think it's something we'll need for a few of the tiles like map and scatter

chejennifer (approved):
thanks!

server/config/topic_page/climate/USA.textproto
beets: :) done

server/config/topic_page/poverty/USA.textproto
beets: we should specify all pages that the topic page works for. i added CA for testing, and also hoping that the config works for both country and state..

server/config/topic_page/climate/USA.textproto
beets: yes! that's what it was meant for and we should absolutely include it. there are a few other issues to debug with the configs, and i was hoping to keep that separate from this.

beets:
thanks for the reviews!

PR #1398, opened 2022-01-22
[topic page] topic page selector

chejennifer:
- add selector for place and topic
screen-recording

### DIFF
--- a/server/routes/topic_page.py
+++ b/server/routes/topic_page.py
@@ -30,7 +30,7 @@ def topic_page(topic_id=None, place_dcid=None):
     if os.environ.get('FLASK_ENV') == 'production':
         flask.abort(404)
 
-    topics_summary = json.dumps(get_topics_summary())
+    topics_summary = json.dumps(current_app.config['TOPIC_PAGE_SUMMARY'])
     # Redirect to the landing page.
     if not place_dcid and not topic_id:
         return flask.render_template('topic_page_landing.html')
@@ -76,18 +76,3 @@ def topic_page(topic_id=None, place_dcid=None):
         topic_name=topic_place_config.metadata.topic_name or "",
         config=MessageToJson(topic_place_config),
         topics_summary=topics_summary)
-
-
-def get_topics_summary():
-    topic_place_map = {}
-    topic_name_map = {}
-    all_configs = current_app.config['TOPIC_PAGE_CONFIG']
-    for topic, config_list in all_configs.items():
-        if len(config_list) < 1:
-            continue
-        topic_name_map[topic] = config_list[0].metadata.topic_name
-        if topic not in topic_place_map:
-            topic_place_map[topic] = []
-        for config in config_list:
-            topic_place_map[topic].extend(config.metadata.place_dcid)
-    return {"topicPlaceMap": topic_place_map, "topicNameMap": topic_name_map}
--- a/server/routes/topic_page.py
+++ b/server/routes/topic_page.py
@@ -16,6 +16,7 @@
 from flask import current_app
 import flask
 import os
+import json
 import routes.api.place as place_api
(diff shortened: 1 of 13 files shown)


server/routes/topic_page.py
```diff
@@ -16,25 +16,42 @@
...
         flask.abort(404)
 
+    topics_summary = json.dumps(get_topics_summary())
     # Redirect to the landing page.
-    if not place_dcid or not topic_id:
+    if not place_dcid and not topic_id:
         return flask.render_template('topic_page_landing.html')
 
-    # Find the config for the topic & place.
     all_configs = current_app.config['TOPIC_PAGE_CONFIG']
```
beets, line 38: we could use topics_summary for this too

server/routes/topic_page.py
```diff
@@ -55,5 +72,22 @@ def topic_page(topic_id=None, place_dcid=None):
...
+        topic_name=topic_place_config.metadata.topic_name or "",
+        config=MessageToJson(topic_place_config),
+        topics_summary=topics_summary)
+
+
+def get_topics_summary():
+    topic_place_map = {}
+    topic_name_map = {}
+    all_configs = current_app.config['TOPIC_PAGE_CONFIG']
+    for topic, config_list in all_configs.items():
```
beets, line 85: nit: topic_id for clarity

```suggestion
    for topic_id, config_list in all_configs.items():
```

server/routes/topic_page.py
```diff
@@ -55,5 +72,22 @@ def topic_page(topic_id=None, place_dcid=None):
...
         place_dcid=place_dcid,
-        topic_name=topic_place_config.metadata.topic_name,
-        config=MessageToJson(topic_place_config))
+        topic_id=topic_id,
+        topic_name=topic_place_config.metadata.topic_name or "",
+        config=MessageToJson(topic_place_config),
+        topics_summary=topics_summary)
+
+
+def get_topics_summary():
```
beets, line 81: we could add this to the flask cache since we'll need it for each topic page render

static/js/topic_page/page_selector.tsx
```diff
@@ -0,0 +1,153 @@
...
+  selectedTopic: string,
+  topicsSummary: TopicsSummary,
+  setPlaceOptions: (placeOptions: Record<string, string>) => void
+): void {
+  const placeOptionDcids = topicsSummary.topicPlaceMap[selectedTopic] || [];
+  const placeParams = placeOptionDcids
+    .map((place) => "dcid=" + place)
+    .join("&");
+  axios
+    .get(`/api/place/name?${placeParams}`)
```
beets, line 119: we can call this server-side in flask and pass it down (since it's a fixed map) instead of calling it from the client. add a TODO for now?

beets (approved):
thanks for the empty data bugfixes across tiles too!

server/routes/topic_page.py
chejennifer: I don't think we can use the summary here because the summary doesn't have the actual configs

server/routes/topic_page.py
chejennifer: done

server/routes/topic_page.py
chejennifer: Did you mean add it to the app config or adding the `@cache.memoize`. I added it to the app config because the topics summary will always be the same

static/js/topic_page/page_selector.tsx
chejennifer: done

server/routes/topic_page.py
beets: app config sgtm

server/routes/topic_page.py
beets: ah yes, good point.

PR #1399, opened 2022-01-24
Topic page updates for poverty-USA

beets:
Config updates
- Add support for custom stat var names
- Add support for custom ranking unit titles (uses statVar placeholder var)

Page updates
- Add support for undefined scaling
- Add border around graphical charts
- Add support for block descriptions
- Add topic name to page title
- Update poverty-USA config

Dev updates
- Hot reload compiled config protos in local env

localhost_8081_topic_poverty_country_USA


server/routes/topic_page.py
```diff
@@ -36,6 +37,8 @@ def topic_page(topic_id=None, place_dcid=None):
         return flask.render_template('topic_page_landing.html')
 
     all_configs = current_app.config['TOPIC_PAGE_CONFIG']
+    if os.environ.get('FLASK_ENV') == 'local':
+        all_configs = libutil.get_topic_page_config()
```
chejennifer, line 41: curious why you're getting all_configs here. Does local flask environment not have the configs saved in the app config?

server/config/topic_page/poverty/USA.textproto
```diff
@@ -105,15 +150,165 @@ blocks {
     type: LINE
     title: "Poverty rate by race in ${place}"
   }
+  right_tiles {
+    type: RANKING
+    description: "Some note about how US Census defines race…"
```
chejennifer, line 155: Looks like the ranking tile doesn't show the description and not sure if you were going to actually update this description

chejennifer (approved):

server/routes/topic_page.py
beets: i was making many updates to the configs and wanted a hot reload (instead of fully restarting the server). so locally, you can separately compile the protos and the latest config will be used

server/config/topic_page/poverty/USA.textproto
beets: ah yes, that isn't done yet. i'll remove and add a note to implement

PR #1400, opened 2022-01-24
[tree widget] Hide uncategorized section

chejennifer:
- hide the uncategorized section of the tree widget on staging and prod. Autopush will still be able to see the uncategorized section.
- move the stat var hierarchy related flask endpoints from browser.py to stats.py because they're not specific to the graph browser


beets (approved):
thanks for the refactor!

PR #1402, opened 2022-01-24
[map] fix bug for some new RCP stat vars

chejennifer:
- some of the new RCP stat vars only have data for 2100 so we don't want to cap the date for those stat vars
- add some checks for empty data so that the tool doesn't error out


beets (approved):
thanks for the fix!

PR #1404, opened 2022-01-24
[map] url param to set the stat var for installations on a map

chejennifer:
- add a url param to set the stat var used for the installations on a map. If no stat var set, will default to using the same stat var as being shown for the map regions.
- URL params to show glaciers: &mapsv=SurfaceArea&ppt=Glacier
- Now that installations can show a different stat var than the rest of the map, remove the dateParam when retrieving installations data because it can be hard to make the date work for both the map stat var and the installations stat var. For example, RCP stat vars use 2050 as the date, but not many other stat vars have data for 2050.
Screen Shot 2022-01-24 at 1 02 31 PM

### DIFF
--- a/static/js/tools/map/chart_loader.tsx
+++ b/static/js/tools/map/chart_loader.tsx
@@ -101,6 +101,7 @@ export function ChartLoader(): JSX.Element {
     statVar.value.dcid,
     statVar.value.info,
     statVar.value.denom,
+    statVar.value.mapPointSv,
   ]);
 
   useEffect(() => {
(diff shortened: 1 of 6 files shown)


static/js/tools/map/chart_loader.tsx
```diff
@@ -339,10 +340,11 @@ function fetchData(
       return resp.data;
     });
 
+  const mapPointSv = statVar.mapPointSv || statVar.dcid;
   const mapPointDataPromise: Promise<GetStatSetResponse> = placeInfo.mapPointPlaceType
     ? axios
         .get(
-          `/api/stats/within-place?parent_place=${placeInfo.enclosingPlace.dcid}&child_type=${placeInfo.mapPointPlaceType}&stat_vars=${statVar.dcid}${dateParam}`
+          `/api/stats/within-place?parent_place=${placeInfo.enclosingPlace.dcid}&child_type=${placeInfo.mapPointPlaceType}&stat_vars=${mapPointSv}`
```
beets, line 347: should we keep the dateParam?

beets (approved):
nice!

static/js/tools/map/chart_loader.tsx
chejennifer: keeping dateParam doesn't work well when the stat var for the map is different for the installations .. for example, RCP statvars will use 2050, but not many other stat vars will have data for 2050

beets: that makes sense. could you update the pr description with the update? it's good to list all these changes out, which helps with future debugging, and/or a comment in code.

chejennifer: done, updated pr description

PR #1407, opened 2022-01-24
[map] add a min size for map installations

chejennifer:
- the min size of the installations on the map is calculated depending on the sizes of the regions on the map. For IPCCPlace50 placetype, the size of each region causes the installations to become too small to see. Therefore, we want to set a minimum size for map installation points
Screen Shot 2022-01-24 at 3 56 56 PM


pradh (approved):
Thanks for the fix!

beets (approved):
nice catch!

PR #1411, opened 2022-01-25
Do not cache for private instance and update local private instance API_ROOT

shifucun:
As private instance will have frequent data refresh, disable caching so it won't have stale data in the frontend.

To bring up a local private instance, run

`run_server.sh -e private`


chejennifer (approved):
thanks!

PR #1412, opened 2022-01-25
[scatter] Fix bug when there is no StatVarInfo available

chejennifer:
- fix bug causing scatter chart to not show up when one of the chosen stat vars doesn't have StatVarInfo


beets (commented):
what is the app behavior in this case? without setting statvarinfo, are we still able to render a chart?

chejennifer:
> what is the app behavior in this case? without setting statvarinfo, are we still able to render a chart?

intended behaviour/behaviour with this fix: chart will render after statvarinfo is fetched for both stat vars regardless if there is statvarinfo available for the stat vars or not
before this fix: chart not rendering if either of the stat vars don't have any statvarinfo available because we're not updating that field after fetching if there is no statvarinfo

beets (approved):
gotcha, lgtm!

PR #1413, opened 2022-01-25
Topic page scatter updates

beets:
Config updates
- Flip ordering of scatter stat vars (y first, then x)
- Add support for contained_place_types (and added a tile-level override)

Scatter updates
- Removed extra top-margin (and it looks fine on scatter tool)
- Add support for configured stat var names

Equity updates
- Added health scatters

### DIFF
--- a/static/js/topic_page/block.tsx
+++ b/static/js/topic_page/block.tsx
@@ -148,6 +148,7 @@ function renderTiles(tiles: Tile[], props: BlockPropType): JSX.Element {
             id={id}
             title={tile.title}
             place={props.place}
+            // enclosedPlaceType="County"
             enclosedPlaceType={props.enclosedPlaceType}
             statVarMetadata={
               tile.statVarOverride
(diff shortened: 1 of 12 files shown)


static/js/topic_page/block.tsx
```diff
@@ -67,6 +67,10 @@ export function Block(props: BlockPropType): JSX.Element {
 function renderTiles(tiles: Tile[], props: BlockPropType): JSX.Element {
   const tilesJsx = tiles.map((tile) => {
     const id = randDomId();
+    const enclosedPlaceType =
+      tile.containedPlaceTypes && tile.containedPlaceTypes[props.place.types[0]]
```
chejennifer, line 71: might want to check that props.place.types isn't an empty array before trying to access it. And same in the MainPane component

chejennifer (approved):
nice!

static/js/topic_page/block.tsx
beets: done - added fallback code in both places, and especially in topic_page.ts so we have a consistent fallback

beets (commented):
thanks for the review!

PR #1415, opened 2022-01-26
DevOps update for private instance

shifucun:
Also simplified the deploy script file generation, so it does not mess up with local git versioning.


chejennifer (approved):

shifucun:
thanks!

PR #1417, opened 2022-01-26
[homepage] add a "by the numbers" section

beets:
Add a section to the homepage highlighting the size of the graph

localhost_8080_ (4)

### DIFF
--- a/server/templates/static/homepage.html
+++ b/server/templates/static/homepage.html
@@ -90,6 +90,23 @@
   </div>
 </section>
 
+<section id="by-the-numbers">
+  <div class="container">
+    <div class="highlight">
+      <h3>By the numbers</h3>
+      <p>We have continuously added data to the Data Commons graph over the last 5 years. As of Jan 2022, the graph contains:</p>
+      <ul>
+        <li>515 B triples</li>
+        <li>2.8 B time series</li>
+        <li>2.9 M places</li>
+        <li>90 K variables</li>
+        <li>250 datasets</li>
+      </ul>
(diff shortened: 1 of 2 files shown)


server/templates/static/homepage.html
```diff
@@ -90,6 +90,23 @@ <h2>Explore the data</h2>
   </div>
 </section>
 
+<section id="by-the-numbers">
+  <div class="container">
+    <div class="highlight">
+      <h3>By the numbers</h3>
+      <p>We have continuously added data to the Data Commons graph over the last 5 years. As of Jan 2022, the graph contains:</p>
```
pradh, line 97: Wonder if its 4 years?  Roughly from Aug 2018...

server/templates/static/homepage.html
```diff
@@ -90,6 +90,23 @@ <h2>Explore the data</h2>
...
+      <ul>
+        <li>515 B triples</li>
```
pradh, line 99: should we expand - B, M, K - to make it clear?

pradh: Perhaps `2.8 billion`, `2.8 million`, `90,000`, `200,000` ...

pradh (approved):
Looks great!

server/templates/static/homepage.html
beets: thanks for the catch!

server/templates/static/homepage.html
beets: yea that's a good suggestion!

beets:
thanks for the review -- i'll merge so we can see it on autopush

beets:
good catch! let me send a separate pr for that.

PR #1418, opened 2022-01-26
[topic] add more maps to equity page

beets:
also fix the stat on the homepage

image

### DIFF
--- a/server/config/topic_page/equity/USA.textproto
+++ b/server/config/topic_page/equity/USA.textproto
@@ -121,6 +121,26 @@ blocks {
 }
 
 blocks {
+  title: "Income inequality between sexes"
+  description: "Income inequality is computed as the ratio of the difference between median income for each sex, with the combined median incomes. A value of 0 means complete equality of median income."
+  stat_var_metadata {
+    stat_var: "GenderIncomeInequality_Person_15OrMoreYears_WithIncome"
+    name: "Income inequality between sexes"
+  }
+  left_tiles {
+    type: MAP
+    title: "Income inequality between sexes (${date})"
+  }
+  right_tiles {
+    type: RANKING
+    ranking_metadata {
+      show_highest: true
+      show_lowest: true
+    }
+  }
+}
+
+blocks {
   title: "Income and employment, by sex"
   description: "Median incomes of individuals, broken down by sex"
   stat_var_metadata {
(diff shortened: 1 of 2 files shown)


server/config/topic_page/equity/USA.textproto
```diff
@@ -120,6 +120,26 @@ blocks {
   }
 }
 
+blocks {
+  title: "Income inequality between sexes"
+  description: "Income inequality is computed as the ratio of the difference between median income for each sex, with the combined median incomes. A value of 0 means complete equality of median income."
+  stat_var_metadata {
+    stat_var: "GenderIncomeInequality_Person_15OrMoreYears_WithIncome"
```
pradh, line 127: I think this needs refresh... data is from 2018 :(

server/config/topic_page/equity/USA.textproto
```diff
@@ -120,6 +120,26 @@ blocks {
   }
 }
 
+blocks {
+  title: "Income inequality between sexes"
```
pradh, line 124: maybe fine, but I do wonder if we less controversial to just say between "men and women" instead of "sexes"? (given gender identity questions)

pradh (approved):
Nice!

server/config/topic_page/equity/USA.textproto
beets: yea, we do have it both ways. but the census uses "sex" as biological sex where as gender is a social construct (and they are [updating to ask both](https://www.census.gov/library/stories/2021/08/household-pulse-survey-updates-sex-question-now-asks-sexual-orientation-and-gender-identity.html))

how about we add a link to this in the description: https://www.census.gov/topics/population/age-and-sex/about.html

server/config/topic_page/equity/USA.textproto
beets: oh, let me try to run that pipeline...

server/config/topic_page/equity/USA.textproto
pradh: adding that reference sounds great!  we sort of have a modeling bug then, in that we map it to "gender" property.

server/config/topic_page/equity/USA.textproto
pradh: sorry I didn't mean it as blocking this change, but something to refresh

server/config/topic_page/equity/USA.textproto
beets: ah yes, we should clarify it there (and in this case, update the name of the stat var?)

beets:
thanks for the review! i'll merge this and work on followups

PR #1419, opened 2022-01-26
/version endpoints: display base tables and branch table

shifucun:


server/templates/version.html
```diff
@@ -58,5 +70,4 @@
       </tbody>
     </table>
   </div>
-  <script src={{url_for('static', filename='translator.js', t=config['VERSION'])}}></script>
```
pradh, line 61: is this relevant to this change?

pradh (approved):
Does this change require that mixer + website are updated together?

server/templates/version.html
shifucun: This was an old bug. The version page should not include translator js

shifucun:
> Does this change require that mixer + website are updated together?

Yep, updated mixer submodule

PR #1420, opened 2022-01-26
[place] add link to graph browser

beets:
added a link to the graph browser for a place from the overview block. i also considered putting it under the place searchbox (so it will be visible on every page, and also differentiated from the block links) but it felt messy / overwhelming

option 1:
image
 
option 2 (this pr):
image

also fixed a few recent css bugs

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,3 +33,4 @@ deploy/overlays/**/kustomize-build.yaml
 deploy/overlays/**/kustomize-deployed.yaml
 
 deploy/git/mixer_hash.txt
+deploy/git/website_hash.txt
--- a/server/config/topic_page.proto
+++ b/server/config/topic_page.proto
@@ -26,7 +26,7 @@ message PageMetadata {
   // List of places supported by this config.
   repeated string place_dcid = 3;
   // Contained places to load for each parent place type.
-  // This might need to be overwritten by blocks.
+  // This can be overwritten by tiles.
   map<string, string> contained_place_types = 4;
 }
 
@@ -82,6 +82,8 @@ message Tile {
   repeated StatVarMetadata stat_var_override = 4;
   RankingMetadata ranking_metadata = 5;
   HighlightMetadata highlight_metadata = 6;
+  // Overrides the contained_place_types for the page.
+  map<string, string> contained_place_types = 7;
 }
 
 message Block {
--- a/server/config/topic_page/equity/USA.textproto
+++ b/server/config/topic_page/equity/USA.textproto
@@ -121,6 +121,26 @@ blocks {
 }
 
 blocks {
+  title: "Income inequality between sexes"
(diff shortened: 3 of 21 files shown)


beets:
Maybe these screenshots are more illustrative:

option 1:
image


option 2 (this pr):
image

pradh:
Thanks for the quick change!

Should we make the link text "dcid: <..>" sort of like https://datacommons.org/tools/statvar#Count_Household ?

pradh:
To me, Option 2 looks better visually, but not sure if it gives the sense that the graph browser is a continuation of the info in overview (graph, ranking), which doesn't exist.

beets:
> Should we make the link text "dcid: <..>" sort of like https://datacommons.org/tools/statvar#Count_Household ?

I didn't explicitly put the DCID since it's very "technical" and this page is supposed to be more user friendly. I figured the link will bring users to a page that includes both the dcid and the stat vars available?

I had also initially tried putting the link under "Population: 1234" but that felt too much (and important, hierarchically, but not relevant to most users of the page).

beets:
> To me, Option 2 looks better visually, but not sure if it gives the sense that the graph browser is a continuation of the info in overview (graph, ranking), which doesn't exist there?

Yeah, I agree. I've been staring at it more and option 2 looks too "hidden" and lost in the flow. One might expect more "overview" data.

pradh:
> > Should we make the link text "dcid: <..>" sort of like https://datacommons.org/tools/statvar#Count_Household ?
> 
> I didn't explicitly put the DCID since it's very "technical" and this page is supposed to be more user friendly. I figured the link will bring users to a page that includes both the dcid and the stat vars available?
> 
> I had also initially tried putting the link under "Population: 1234" but that felt too much (and important, hierarchically, but not relevant to most users of the page).

I see, ok.  We still probably want something more convenient for our API users (perhaps improved overall search), but this is definitely an improvement from before :)

pradh (approved):

pradh:
Actually if there is some way to get the DCID from the timeline tool (without parsing the URL), it will be so convenient for API users.

They can keep searching for multiple places, navigate LHS variables and know the place DCIDs... all in one place.

Right now they need to search in place page and click through to graph browser for every place they care about.

beets:
> Actually if there is some way to get the DCID from the timeline tool (without parsing the URL), it will be so convenient for API users.

From both places (timeline and place) we can add a modal for "how to query this data via api" because you'd need both stat var and place dcid's.. we can even pre-fill with some sample calls.

pradh:
> > Actually if there is some way to get the DCID from the timeline tool (without parsing the URL), it will be so convenient for API users.
> 
> From both places (timeline and place) we can add a modal for "how to query this data via api" because you'd need both stat var and place dcid's.. we can even pre-fill with some sample calls.

That would be cool!  Though we might need to consider how to handle the different APIs/flavors (point/series in basic py, pandas, rest, etc.)

PR #1421, opened 2022-01-26
[timeline] fix bulk download formatting

beets:
there is a display-flex on #main-pane which caused the weird column layout


chejennifer (approved):
thanks for the fix!

beets:
sure, np!

PR #1424, opened 2022-01-28
feat: add protein chart1

suhana13:
Added .html and .js files for chart 1 of protein browser. 
@shifucun @beets @spiekos


beets:
can we close this pr, since this was a proof of concept to get going?

PR #1425, opened 2022-01-28
Use newer version of place stat vars api in website

shifucun:
Did some clean up for naming.

### DIFF
--- a/server/routes/api/place.py
+++ b/server/routes/api/place.py
@@ -192,23 +192,23 @@ def statsvars_route(dcid):
     Returns:
       A list of statistical variable dcids.
     """
-    return Response(json.dumps(statsvars(dcid)),
+    return Response(json.dumps(statVars(dcid)),
                     200,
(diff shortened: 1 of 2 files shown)


server/routes/api/place.py
```diff
@@ -192,23 +192,23 @@ def statsvars_route(dcid):
...
     """
-    return Response(json.dumps(statsvars(dcid)),
+    return Response(json.dumps(statVars(dcid)),
                     200,
                     mimetype='application/json')
 
 
 @cache.memoize(timeout=3600 * 24)  # Cache for one day.
-def statsvars(dcid):
+def statVars(dcid):
```
chejennifer, line 201: looking at the other file you changed, maybe for consistency we should do stat_vars?

chejennifer (approved):
thanks for the update!

server/routes/api/place.py
shifucun: Updated

PR #1426, opened 2022-01-28
[map/scatter] reduce randomness of available statvars in the stat var hierarchy

chejennifer:
- add curated list of sample places to use for countries in the world
- In map tool, increase the sample size for list of places to pass to the stat var hierarchy to 50


beets (approved):
thanks for the quick fix!

PR #1431, opened 2022-02-01
[map] fix bug caused by empty metahash in data

chejennifer:


chejennifer:
> For which query request, there is empty metadata?

It's not empty metadata, but the stat doesn't have a metahash attached to it.

the request: api/stats/within-place/all?parent_place=country/USA&child_type=County&stat_vars=IncrementalCount_MedicalConditionIncident_COVID_19_ConfirmedOrProbableCase&date=2021-10

the response:
Screen Shot 2022-02-01 at 1 38 41 PM

shifucun:
I sent a fix in mixer https://github.com/datacommonsorg/mixer/pull/778.

PR #1435, opened 2022-02-03
Update website to use the new ranking API response

shifucun:
The ranking API has been updated to not using payload but a protobuf (json) in mixer: https://github.com/datacommonsorg/mixer/pull/723


chejennifer (approved):
thanks for the fix!

PR #1439, opened 2022-02-04
fix webdriver test

chejennifer:
- the update to the stat var search mixer result caused the stat vars to be nested under stat var groups, so selecting the first result was not giving a chart


beets (approved):
thanks for the fix!

PR #1440, opened 2022-02-04
feat: add chart 1

suhana13:
@chejennifer , @beets , @shifucun @spiekos 
I added the code for the first chart. However, there is some issue on the formatting end which I'm unable to fix. Any help debugging it, would be greatly appreciated. 
Here's the link to the observable hq version of the notebook on how it should look. Also the chart is found on 'bio/protein/bio/P53_HUMAN' on the local version of the website. 
https://observablehq.com/d/a51755b9159a6baf

### DIFF
--- a/static/js/protein/chart.ts
+++ b/static/js/protein/chart.ts
@@ -57,12 +57,12 @@ export function drawTissueScoreChart(
     }
   }
 
-  const reformattedData = data.map((item) => {
-    const newObj = {} as ProteinPropDataNumType;
-    //var newObj: any = {};
-    newObj["name"] = item.name;
-    newObj["value"] = convertData(item.value);
-    return newObj;
+  var reformattedData = {} as ProteinPropDataNumType[];
+  reformattedData = data.map((item) => {
+    return {
+    name: item.name,
+    value: convertData(item.value),
+    }
   });
 
   const arr_name = reformattedData.map((x) => {
--- a/static/js/protein/chart.ts
+++ b/static/js/protein/chart.ts
@@ -14,40 +14,151 @@
  * limitations under the License.
  */
 
-import * as d3 from "d3";
+ import * as d3 from "d3";
 
-const SVGNS = "http://www.w3.org/2000/svg";
-const XLINKNS = "http://www.w3.org/1999/xlink";
-
-const HEIGHT = 224;
-const WIDTH = 500;
-
-/**
- * Draw bar chart for tissue score.
- */
-export function drawTissueScoreChart(
-  id: string,
-  data: { name: string; value: string }[]
-): void {
-  // TODO: convert data to tmpData format.
-  const tmpData = [{ x1: 20, x2: 60, y1: 30, y2: 50 }];
-  const svg = d3
-    .select("#" + id)
-    .append("svg")
-    .attr("xmlns", SVGNS)
-    .attr("xmlns:xlink", XLINKNS)
-    .attr("width", WIDTH)
-    .attr("height", HEIGHT);
-
-  svg
-    .append("g")
-    .selectAll("rect")
-    .data(tmpData)
-    .enter()
-    .append("rect")
-    .attr("x", (d) => d.x1)
-    .attr("y", (d) => d.y1)
-    .attr("width", (d) => d.x2 - d.x1)
-    .attr("height", (d) => d.y2 - d.y1)
-    .attr("fill", "teal");
-}
+ const SVGNS = "http://www.w3.org/2000/svg";
+ const XLINKNS = "http://www.w3.org/1999/xlink";
+ 
+ const HEIGHT = 224;
+ const WIDTH = 500;
+ 
+ /**
+  * Draw bar chart for tissue score.
+  */
+ export function drawTissueScoreChart(
+   id: string,
+   data: { name: string; value: string }[]
+ ): void {
+     const tmpData = [
+       { name: "Cartilage", value: 3 },
+       { name: "Eye", value: 2 },
+       { name: "Liver", value: 1 },
+       { name: "Skin", value: 0 },
+     ];
+     const size = Object.keys(tmpData);
+     const barHeight = 35;
+     const margin = { top: 70, right: 50, bottom: 10, left: 50 };
+     const height = 400 - margin.top - margin.bottom;
+     const width = 460 - margin.left - margin.right;
+     const x = d3.scaleLinear()
+         .domain([0, d3.max(tmpData, (d) => d.value)])
+         .range([margin.left, width - margin.right]);
+     const y = d3.scaleBand()
+         .domain(size)
+         .rangeRound([margin.top, height - margin.bottom])
+         .padding(0.1);
+     function formatTick(d) {
+         if (d == 0) {
+             return "NotDetected";
+         } else if (d == 1) {
+             return "Low";
+         } else if (d == 2) {
+             return "Medium";
+         } else {
+             return "High";
+         }
+         };
+       const svg = d3.selectAll("#" + id).append('svg')
+         .attr("xmlns", SVGNS)
+         .attr("xmlns:xlink", XLINKNS)
+         .attr("width", WIDTH)
+         .attr("height", HEIGHT);
+         //.attr("viewBox", [0, 0, width, height]);
+     
+     svg.append("g")
+         .selectAll("rect")
+         .data(tmpData.sort((a, b) => d3.descending(a.value, b.value)))
+         .enter()
+         .append("rect")
+         .attr("x", x(0))
+         .attr("y", (d) => y(d.name)) //y(i)
+         .attr("width", (d) => x(d.value) - x(0))
+         .attr("height", y.bandwidth())
+         .style("fill", function (d) {
+             if (d.name == "Eye") {
+             return "green";
+             } else if (d.name == "Cartilage") {
+             return "yellow";
+             } else if (d.name == "Liver") {
+             return "steelblue";
+             } else {
+             return "red";
+             }
+         })
+         .on("mouseover", function (d, i) {
+             d3.select(this).transition().duration(50).style("opacity", 0.5);
+         })
+         .on("mouseout", function (d, i) {
+             d3.select(this).transition().duration(50).style("opacity", 1);
+         });
+     
+     svg.append("g")
+         .attr("fill", "white")
+         .attr("text-anchor", "end")
+         .attr("font-family", "sans-serif")
+         .attr("font-size", 3)
+         .selectAll("text")
+         .data(tmpData.sort((a, b) => d3.descending(a.value, b.value)))
+         .join("text")
+         .attr("x", (d) => x(d.value))
+         .attr("y", (d, i) => i + y.bandwidth() / 2)
+         .attr("dy", "0.35em")
+         .attr("dx", -4)
+         .call((text) =>
+             text
+             .filter((d) => x(d.value) - x(0) < 20) // short bars
+             .attr("dx", +2)
+             .attr("fill", "black")
+             .attr("text-anchor", "start")
+         );
+ 
+     svg.append("g")
+         .attr("transform", `translate(0,${margin.top})`)
+         .call(d3.axisTop(x).ticks(3).tickFormat(formatTick))
+         .call((g) => g.select(".domain").remove())
+         .on("mouseover", function () {
+             d3.select(this).select("rect").style("fill", "red");
+         })
+         .on("mouseout", function (d, i) {
+             d3.select(this).select("rect").style("fill", "blue");
+         });
+ 
+         svg.append("g")
+         .attr("transform", `translate(${margin.left},0)`)
+         .call(
+             d3
+             .axisLeft(y)
+             .tickFormat((i) => tmpData[i].name)
+             .tickSizeOuter(0)
+         )
+         .on("mouseover", function () {
+             d3.select(this).select("rect").style("fill", "red");
+         })
+         .on("mouseout", function (d, i) {
+             d3.select(this).select("rect").style("fill", "blue");
+         });
+   // TODO: convert data to tmpData format.
+   /*
+   const tmpData = [{ x1: 20, x2: 60, y1: 30, y2: 50 }];
+   const svg = d3
+     .select("#" + id)
+     .append("svg")
+     .attr("xmlns", SVGNS)
+     .attr("xmlns:xlink", XLINKNS)
+     .attr("width", WIDTH)
+     .attr("height", HEIGHT);
+ 
+   svg
+     .append("g")
+     .selectAll("rect")
+     .data(tmpData)
+     .enter()
+     .append("rect")
+     .attr("x", (d) => d.x1)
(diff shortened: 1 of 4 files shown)


static/js/protein/chart.ts
```diff
@@ -14,40 +14,151 @@
...
+         .attr("height", HEIGHT);
+         //.attr("viewBox", [0, 0, width, height]);
+     
+     svg.append("g")
+         .selectAll("rect")
+         .data(tmpData.sort((a, b) => d3.descending(a.value, b.value)))
+         .enter()
+         .append("rect")
+         .attr("x", x(0))
+         .attr("y", (d) => y(d.name)) //y(i)
```
beets, line 74: this is different from the observable code:

```
      .attr("y", (d, i) => y(i) + y.bandwidth() / 2)
```

this line sets the y position for the rectangle, so i'd check here. you could:
- manually return different values
- add print statements: `(d) => { console.log(y(d.name)), return y(d.name); }`

static/js/protein/chart.ts
```diff
@@ -14,40 +14,151 @@
...
+     const size = Object.keys(tmpData);
+     const barHeight = 35;
+     const margin = { top: 70, right: 50, bottom: 10, left: 50 };
+     const height = 400 - margin.top - margin.bottom;
+     const width = 460 - margin.left - margin.right;
+     const x = d3.scaleLinear()
+         .domain([0, d3.max(tmpData, (d) => d.value)])
+         .range([margin.left, width - margin.right]);
+     const y = d3.scaleBand()
+         .domain(size)
```
chejennifer, line 47: I'm guessing the domain you want here should be the list of names in tmpData? But the object `size` that you're passing in is actually a list of 0, 1, 2, 3. This is because tmpData is an array and when you call Object.keys on an array, the keys are actually the indices of the array.

beets: nice find!

suhana13: I added the changes of chart 1 to this PR. I will add the rest of the charts in separate PR for ease!

suhana13:
@beets @chejennifer , I ran the listing and cl checks and removed console statements from the first graph. It should be good for a review !

static/js/protein/chart.ts
```diff
@@ -29,10 +29,61 @@ export function drawTissueScoreChart(
...
 ): void {
-  // TODO: convert data to tmpData format.
-  const tmpData = [{ x1: 20, x2: 60, y1: 30, y2: 50 }];
+  const tmpData = [
+    { name: "Cartilage", value: 3 },
+    { name: "Eye", value: 2 },
+    { name: "Liver", value: 1 },
+    { name: "Skin", value: 0 },
+  ];
+  const testData = data;
```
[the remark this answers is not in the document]
beets: also, since this is a rather long function, could you please add some comments within for the various "chunks" of code, e.g. when drawing axis, bars, etc

also helpful to comment on what the arguments are that are passed in to the function.

beets (commented):
looks nice suhana! thanks for updating the pr for this chart

static/js/protein/chart.ts
```diff
@@ -29,10 +29,61 @@ export function drawTissueScoreChart(
...
+    } else if (val == "ProteinExpressionLow") {
+      return 1;
+    } else if (val == "ProteinExpressionMedium") {
+      return 2;
+    } else if (val == "ProteinExpressionHigh") {
+      return 3;
+    }
+  }
+  const reformat = data.map((item) => {
+    const newObj: any = {};
```
chejennifer, line 51: usually we don't want to use `any` as the data type. Maybe create a new interface for {name: string, value: number}

static/js/protein/chart.ts
```diff
@@ -29,10 +29,61 @@ export function drawTissueScoreChart(
...
+      return 0;
+    } else if (val == "ProteinExpressionLow") {
+      return 1;
+    } else if (val == "ProteinExpressionMedium") {
+      return 2;
+    } else if (val == "ProteinExpressionHigh") {
+      return 3;
+    }
+  }
+  const reformat = data.map((item) => {
```
chejennifer, line 50: typically, we want to use nouns for variable names and verbs for function names.

so here, maybe processedData or something?

static/js/protein/chart.ts
```diff
@@ -29,10 +29,61 @@ export function drawTissueScoreChart(
...
+    const newObj: any = {};
+    newObj["name"] = item.name;
+    newObj["value"] = convertData(item.value);
+    return newObj;
+  });
+
+  const size = reformat.map((x) => {
```
chejennifer, line 57: nit: this looks like an array of names so maybe `names` instead of `size` for the variable name

static/js/protein/chart.ts
```diff
@@ -42,12 +93,72 @@ export function drawTissueScoreChart(
...
+    .on("mouseover", function (d, i) {
+      d3.select(this).transition().duration(50).style("opacity", 0.5);
+    })
+    .on("mouseout", function (d, i) {
+      d3.select(this).transition().duration(50).style("opacity", 1);
+    });
+
+  svg
+    .append("g")
+    .attr("fill", "white")
```
chejennifer, line 123: would be nice to put these style attributes in a css file instead

chejennifer (commented):
Nice job Suhana!

static/js/protein/chart.ts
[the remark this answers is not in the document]
suhana13: Added comments and removed the unused code.

static/js/protein/chart.ts
suhana13: Added interface !

static/js/protein/chart.ts
suhana13: fixed the names.

static/js/protein/chart.ts
suhana13: done !

static/js/protein/chart.ts
```diff
@@ -29,10 +29,61 @@ export function drawTissueScoreChart(
...
+    newObj["value"] = convertData(item.value);
+    return newObj;
+  });
+
+  const size = reformat.map((x) => {
+    return x.name;
+  });
+
+  const barHeight = 35;
+  const margin = { top: 70, right: 50, bottom: 10, left: 50 };
```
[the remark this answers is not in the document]
suhana13: Done !

static/js/protein/chart.ts
```diff
@@ -29,10 +31,63 @@ export function drawTissueScoreChart(
   id: string,
   data: { name: string; value: string }[]
 ): void {
-  // TODO: convert data to tmpData format.
-  const tmpData = [{ x1: 20, x2: 60, y1: 30, y2: 50 }];
+  const testData = data;
```
chejennifer, line 34: I believe this is an unused line so it can be deleted

static/js/protein/chart.ts
```diff
@@ -29,10 +31,63 @@ export function drawTissueScoreChart(
...
+    } else if (val == "ProteinExpressionLow") {
+      return 1;
+    } else if (val == "ProteinExpressionMedium") {
+      return 2;
+    } else if (val == "ProteinExpressionHigh") {
+      return 3;
+    }
+  }
+
+  interface newData{
```
chejennifer, line 51: - this interface should be defined outside of this function and exported because it can be used in mulitple places across multiple files. eg. line 32 in this file,  line 73 and 91 in page.tsx

- this name is very ambiguous. When defining interfaces, we want to use a name that gives other develops an idea of what the kind of object could be. Here maybe something along the lines of ProteinChartDataPoint
    - also when naming interfaces, we want to use pascal case

static/js/protein/chart.ts
```diff
@@ -29,10 +31,63 @@ export function drawTissueScoreChart(
...
+    newObj["name"] = item.name;
+    newObj["value"] = convertData(item.value);
+    return newObj;
+  });
+
+  const arr_name = reformattedData.map((x) => {
+    return x.name;
+  });
+
+  const height = 400 - MARGIN.top - MARGIN.bottom;
```
chejennifer, line 67: 400 and 460 should be defined constants

chejennifer (commented):
Thanks for the updates! A few more things

static/js/protein/chart.ts
suhana13: Replaced the number values with constants.

static/js/protein/chart.ts
suhana13: Added two interfaces because the one used in the chart.ts returns a number value while the one used in page.tsx returns string values.

static/js/protein/chart.ts
suhana13: Done !

static/js/protein/chart.ts
```diff
@@ -21,18 +21,80 @@ const XLINKNS = "http://www.w3.org/1999/xlink";
...
+      return 1;
+    } else if (val == "ProteinExpressionMedium") {
+      return 2;
+    } else if (val == "ProteinExpressionHigh") {
+      return 3;
+    }
+  }
+
+  const reformattedData = data.map((item) => {
+    const newObj = {} as ProteinPropDataNumType;
```
chejennifer, line 61: nit: lines 61-65 can be replaced with:

```
return {
    name: item.name,
    value: convertData(item.value),
}
```

and to enforce type, you can specify that the type of `reformattedData` is `ProteinPropDataNumType[]`

static/js/protein/chart.ts
```diff
@@ -21,18 +21,80 @@ const XLINKNS = "http://www.w3.org/1999/xlink";
...
+    return newObj;
+  });
+
+  const arr_name = reformattedData.map((x) => {
+    return x.name;
+  });
+
+  const max_height = 400;
+  const max_width = 460;
+  const height = max_height - MARGIN.top - MARGIN.bottom;
```
chejennifer, line 74: it looks like you have constants defined at the top of the page for HEIGHT and WIDTH, but you're not using it here (or anywhere in this file)?

maybe update those constants with this height and width & use those variables here in this function.

chejennifer (approved):
thanks for the updates! Few additional small things

static/js/protein/chart.ts
suhana13: Done !

static/js/protein/chart.ts
suhana13: Thank you for pointing that out. Yes, I have made a bunch of changes to the way constants are declared and used, but that also changes a bunch of other things too. I was thinking of addressing that in the next PR where we have the final chart ? If not, please let me know and I'll add it here!

static/js/protein/chart.ts
```diff
@@ -21,18 +21,80 @@ const XLINKNS = "http://www.w3.org/1999/xlink";
...
+    .scaleLinear()
+    .domain([0, d3.max(reformattedData, (d) => d.value)])
+    .range([MARGIN.left, width - MARGIN.right]);
+  const y = d3
+    .scaleBand()
+    .domain(arr_name)
+    .rangeRound([MARGIN.top, height - MARGIN.bottom])
+    .padding(0.1);
+  function formatTick(d) {
+    if (d == 0) {
```
beets, line 86: would be nice if some of these were defined in constant dictionaries at the top of the file, so updates in the future to colors, etc, would be easier to make.
```js
const TISSUE_SCORE_TO_LABEL = {
   0: "NotDetected",
   1: "Low",
   ...
}
...
.tickFormat((d) => TISSUE_SCORE_TO_LABEL[d])
```

static/js/protein/page.tsx
```diff
@@ -92,8 +89,15 @@ export class Page extends React.Component<PagePropType, PageStateType> {
         }
         result[tissue] = score;
       }
-      return result;
+      //const data: { name: string; value: string }[] = [];
```
beets, line 92: delete these comments?

beets (approved):
thanks for the updates suhana. approving to unblock this pr, but please do address the comments in followups as it makes sense.

static/js/protein/page.tsx
```diff
@@ -92,8 +89,14 @@ export class Page extends React.Component<PagePropType, PageStateType> {
...
-      return result;
+      const data: ProteinPropDataStrType[] = [];
+
+      for (const tissue in result) {
+        data.push({ name: tissue, value: result[tissue] });
+      }
+      return data;
     }
-    return {};
+    console.log(data);
```
chejennifer, line 99: nit: remove this console.log statement

static/js/protein/chart.ts
```diff
@@ -19,20 +19,69 @@ import * as d3 from "d3";
...
-  const tmpData = [{ x1: 20, x2: 60, y1: 30, y2: 50 }];
+  let reformattedData = {} as ProteinPropDataNumType[];
+  reformattedData = data.map((item) => {
+    return {
+      name: item.name,
+      value: TISSUE_VAL_TO_SCORE[item.value],
+    };
+  });
+
+  const arr_name = reformattedData.map((x) => {
```
chejennifer, line 65: nit: variables should be camel case so arr_name -> arrName

static/js/protein/chart.ts
```diff
@@ -42,12 +91,72 @@ export function drawTissueScoreChart(
...
-    .attr("y", (d) => d.y1)
-    .attr("width", (d) => d.x2 - d.x1)
-    .attr("height", (d) => d.y2 - d.y1)
-    .attr("fill", "teal");
+    .attr("x", x(0))
+    .attr("y", (d) => y(d.name))
+    .attr("width", (d) => x(d.value) - x(0))
+    .attr("height", y.bandwidth())
+    .style("fill", function (d) {
+      if (d.name == "Eye") {
```
chejennifer, line 102: in typescript/javascript, we want to use "===" instead of "=="

static/js/protein/chart.ts
```diff
@@ -42,12 +91,72 @@ export function drawTissueScoreChart(
...
+        return "green";
+      } else if (d.name == "Cartilage") {
+        return "yellow";
+      } else if (d.name == "Liver") {
+        return "steelblue";
+      } else {
+        return "red";
+      }
+    })
+    .on("mouseover", function (d, i) {
```
chejennifer, line 112: nit: neither d or i are used so replace with just ()
nit: function should have a type definition, so here it would be

so this should look like & same for all the other mouseover and mouseout functions (line 115, 145, 148, 156, 159):
`function (): void {...`

static/js/protein/protein_chart.test.tsx
```diff
@@ -0,0 +1,712 @@
+import Enzyme, { shallow } from "enzyme";
+import Adapter from "enzyme-adapter-react-16";
+import _ from "lodash";
+import React from "react";
+
+Enzyme.configure({ adapter: new Adapter() });
+
+import { GraphNodes } from "../shared/types";
+import { GraphNode } from "../shared/types";
```
chejennifer, line 9: remove unused import statements

chejennifer (commented):
Thanks for adding the tests! A few more nits that were caught by Codacy & please run `./run_test -f ` to fix some of the styling

PR #1441, opened 2022-02-05
[topic] Proto update - move stat_var_metadata to the category layer

beets:
I found that svm definitions were duplicated multiple times across blocks, and specifying overrides for some tiles were cumbersome. This also removes some ambiguity as to which svm is used for tiles like highlight and map which only use the first svm. Eventually, we can centralize all data fetching for a category.

Proof of concept -- let me know what you think! I've only updated one textproto. Will update tests / other configs.

Also moved all typed interfaces for the config proto to a single file that mirrors the proto.


chejennifer (approved):
nice change, this feels cleaner!

beets:
thanks for the review!

PR #1444, opened 2022-02-10
[i18n] update strings for translation, update instructions for adding place charts

beets:
it's been a while since we added strings for translations:
- updated path in chart_config_extractor.py (moved in #1392)
- removed some strings that were marked for translation (but not required)
- update instructions for adding new place charts

/cc @rbhoot


chejennifer (approved):
thanks for the update!

PR #1448, opened 2022-02-16
Update website to use new field from stat var group response

shifucun:
The proto change is introduced in https://github.com/datacommonsorg/mixer/pull/741


chejennifer (approved):

pradh (approved):

beets (approved):

shifucun:
There is an issue with https://autopush.datacommons.org/api/stats/stats-var-property?dcid=Count_Person, that the title is not set correctly. A few webdriver tests are failing because of this. I will submit this and continue to debug in mixer.

PR #1449, opened 2022-02-16
Update deployment and version to match recent mixer import group change

shifucun:
- Remove cpu limit in GKE
- And import groups args for mixer container
- Update /version to reflect mixer response change

### DIFF
--- a/deploy/base/deployment.yaml
+++ b/deploy/base/deployment.yaml
@@ -60,10 +60,8 @@ spec:
           resources:
             limits:
               memory: "3G"
-              cpu: "500m"
             requests:
               memory: "3G"
-              cpu: "500m"
           env:
             - name: FLASK_ENV
               valueFrom:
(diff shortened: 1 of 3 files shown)


beets (approved):
this is a good opportunity to test dropping cpu limits. is there a way for us to test this change? should we write a quick 1-pager to describe how we will roll this out (maybe a pod/cluster to test/stage out this new setting)

also submitting this with the base deployment would make the next push risky. can we do this for autopush first?

i know you've closed the [bug](https://github.com/datacommonsorg/mixer/issues/661), but would be good to refer to it here since it lists the articles with explanation about the cpu limit

shifucun:
> this is a good opportunity to test dropping cpu limits. is there a way for us to test this change? should we write a quick 1-pager to describe how we will roll this out (maybe a pod/cluster to test/stage out this new setting)
> 
> also submitting this with the base deployment would make the next push risky. can we do this for autopush first?
> 
> i know you've closed the [bug](https://github.com/datacommonsorg/mixer/issues/661), but would be good to refer to it here since it lists the articles with explanation about the cpu limit

I think it's a good to have a benchmark testing framework for things like this. I reverted the cpu change in this PR so other things are not blocked. We can test this out when we have the benchmark setup ready.

beets:
sounds like a plan! thanks

PR #1452, opened 2022-02-17
Fix temperature projection timelines

beets:
- Use model data only for the same obs period as primary stat data
- Fix transformed data structures returned
- Factor out to common util (with other StatData processing logic)
- Added tests

### DIFF
--- a/static/js/tools/timeline/data_fetcher.ts
+++ b/static/js/tools/timeline/data_fetcher.ts
@@ -22,6 +22,7 @@ import {
   DisplayNameApiResponse,
   StatApiResponse,
   TimeSeries,
+  StatAllApiResponse,
 } from "../../shared/stat_types";
 
 const ZERO_POPULATION = 0;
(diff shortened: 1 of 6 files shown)


static/js/tools/timeline/data_fetcher.ts
```diff
@@ -292,3 +293,74 @@ export function fetchStatData(
...
+  if (!mainStatData.dates.length) {
+    return [mainStatData, modelData];
+  }
+  for (const place in modelStatAllResponse.placeData) {
+    const placeData = modelStatAllResponse.placeData[place];
+    modelData.places.push(place);
+    modelData.data[place] = { data: {} };
+    for (const sv in modelStatAllResponse.placeData[place].statVarData) {
+      const mainObsPeriod =
+        mainStatData.data[place].data[sv].metadata.observationPeriod;
```
chejennifer, line 329: might want to check that place is in mainStatData.data first

static/js/tools/timeline/data_fetcher.ts
```diff
@@ -292,3 +293,74 @@ export function fetchStatData(
...
+          for (const date of Object.keys(series.val).sort()) {
+            means[date] = date in means ? means[date] : [];
+            means[date].push(series.val[date]);
+          }
+          const newSv = `${sv}-${series.measurementMethod}`;
+          modelData.data[place].data[newSv] = { val: series.val };
+          modelData.statVars.push(newSv);
+        }
+        for (const date in means) {
+          means[date] = _.mean(means[date]);
```
chejennifer, line 354: seems like this is changing the type of the means object from Record<string, number[]> to Record<string, number>. Might want to create a new object or add a comment noting the change in type.

static/js/tools/timeline/data_fetcher.ts
```diff
@@ -292,3 +293,74 @@ export function fetchStatData(
...
+        }
+        svData.sourceSeries = keepSeries;
```
chejennifer, line 356: looks like this would be updating modelStatAllResponse, is that what you wanted? If so, maybe add a comment to explain this part

chejennifer (approved):
Nice!

static/js/tools/timeline/data_fetcher.ts
beets: done

static/js/tools/timeline/data_fetcher.ts
beets: good catch - done

static/js/tools/timeline/data_fetcher.ts
beets: added a comment

beets (commented):
thanks for the review!

PR #1454, opened 2022-02-18
[tree widget] fix result count string

chejennifer:
- handle cases where there is only one sv/svg or only one of sv/svg


beets (approved):
thanks for adding a test too!

PR #1455, opened 2022-02-22
cap the date for SSP stat vars

chejennifer:


beets (approved):
thanks - could you add a few more tests cases for this as well? there is already a test set up, just needs more vars to add.

@pradh are there other sv's which we wouldn't want to cap the date, along the lines of NumberOfMonths_5CelsiusOrMore_Percentile10AcrossModels_

static/js/shared/util.ts
```diff
@@ -94,8 +94,8 @@ export function isDateTooFar(date: string): boolean {
 }
 
 export function getCappedStatVarDate(statVar: string): string {
-  // Only want to cap stat var date for stat vars with RCP.
-  if (!statVar.includes("_RCP")) {
+  // Only want to cap stat var date for stat vars with RCP or SSP.
+  if (!statVar.includes("_RCP") && !statVar.includes("_SSP")) {
     return "";
   }
   for (const svSubstring of NO_DATE_CAP_RCP_STATVARS) {
```
pradh, line 101: Could you remind me why this is not "NumberOfMonths_5CelsiusOrMore_", but "NumberOfMonths_5CelsiusOrMore_Percentile10AcrossModels_" ?

We have Percentile10, Percentile90 and Median...

pradh (approved):
Thanks for the fix!

pradh:
> thanks - could you add a few more tests cases for this as well? there is already a test set up, just needs more vars to add.
> 
> @pradh are there other sv's which we wouldn't want to cap the date, along the lines of NumberOfMonths_5CelsiusOrMore_Percentile10AcrossModels_

Yup, just left a comment... I was wondering if it should be `NumberOfMonths_5CelsiusOrMore_`...

static/js/shared/util.ts
chejennifer: Looking through our chat history, it looks like it was because "That class of stat-vars includes decadal aggregates and are available for dates 2030, 2050 and 2100.  Now the "percentile90" and "median" have data for 2030 and 2050.  But "percentile10" only has data for 2100."

pradh: I see, so for percentile90 and median we want to show 2050. So maybe we can add that comment to the list?

chejennifer: done

chejennifer:
> thanks - could you add a few more tests cases for this as well? there is already a test set up, just needs more vars to add.
> 
> @pradh are there other sv's which we wouldn't want to cap the date, along the lines of NumberOfMonths_5CelsiusOrMore_Percentile10AcrossModels_

done, added more test cases

beets (approved):

PR #1456, opened 2022-02-22
Pin markupsafe version

chejennifer:
- flask depends on jinja2 which depends on markupsafe and the latest markupsafe version has a [breaking change](https://github.com/pallets/markupsafe/issues/286) causing an error to be thrown when trying to run flask locally because [Jinja2 depends on the latest version of markupsafe](https://github.com/pallets/jinja/issues/1585)


beets (approved):
thanks - separately from this, we should look into a bigger version update... our packages are at least one major version behind

PR #1458, opened 2022-02-23
Remove GKE cpu limit

shifucun:
Need to remove the cpu limit as the mixer would restart a few times currently when processing the stat var hierarchy cache.

### DIFF
--- a/deploy/base/deployment.yaml
+++ b/deploy/base/deployment.yaml
@@ -60,10 +60,8 @@ spec:
           resources:
             limits:
               memory: "3G"
-              cpu: "500m"
             requests:
               memory: "3G"
-              cpu: "500m"
           env:
             - name: FLASK_ENV
               valueFrom:
@@ -101,10 +99,8 @@ spec:
           resources:
             limits:
               memory: "6G"
-              cpu: "800m"
             requests:
               memory: "6G"
-              cpu: "800m"
           args:
             - --mixer_project=$(MIXER_PROJECT)
             - --store_project=$(STORE_PROJECT)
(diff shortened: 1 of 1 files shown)


beets (approved):
exciting!

shifucun:
thanks!

pradh:
> Need to remove the cpu limit as the mixer would restart a few times currently when processing the stat var hierarchy cache.

In your research, if there are any downsides of removing this, that might be worth noting.  (One thing that comes to mind is effect of 1 task hogging CPU on others on the same host)

shifucun:
> > Need to remove the cpu limit as the mixer would restart a few times currently when processing the stat var hierarchy cache.
> 
> In your research, if there are any downsides of removing this, that might be worth noting. (One thing that comes to mind is effect of 1 task hogging CPU on others on the same host)

I have an open issue to later benchmark this change with A/B testing. Hopefully to come back to this when new features settle.

pradh:
> > > Need to remove the cpu limit as the mixer would restart a few times currently when processing the stat var hierarchy cache.
> > 
> > 
> > In your research, if there are any downsides of removing this, that might be worth noting. (One thing that comes to mind is effect of 1 task hogging CPU on others on the same host)
> 
> I have an open issue to later benchmark this change with A/B testing. Hopefully to come back to this when new features settle.

Benchmark SG... but given we're making this change in prod, just wanted to check if we anticipate any issues that we should keep an eye on (e.g., based on experience on the Internets).

shifucun:
> > > > Need to remove the cpu limit as the mixer would restart a few times currently when processing the stat var hierarchy cache.
> > > 
> > > 
> > > In your research, if there are any downsides of removing this, that might be worth noting. (One thing that comes to mind is effect of 1 task hogging CPU on others on the same host)
> > 
> > 
> > I have an open issue to later benchmark this change with A/B testing. Hopefully to come back to this when new features settle.
> 
> Benchmark SG... but given we're making this change in prod, just wanted to check if we anticipate any issues that we should keep an eye on (e.g., based on experience on the Internets).

From the current usage. The cpu usage is far below the limit. I will look into if there is alert to signal cpu issues.

image

PR #1461, opened 2022-02-25
Better check of browser page provenance triples

shifucun:
In import group mode, subject name is added to all triples. In the the "Provenance" triples data, when "Provenance" is subject, its name "Provenance" is used as URL and lead to errors. This makes the check more robust.


beets (approved):
this is a good candidate for a helper function (that can help with self-documenting / testing)..

shifucun:
> this is a good candidate for a helper function (that can help with self-documenting / testing)..

Updated

chejennifer (approved):

PR #1466, opened 2022-03-01
[tree widget] fix invalid regex problem

chejennifer:
- there was a bug where searching for mortality in the tree widget threw an error. This was because one of the matching tokens returned for this query was "mortality)" and the ")" at the end was causing an invalid regex error.


beets (commented):
it would be worth adding a test for this, since this ended up causing an irrecoverable bug.

something else worth adding is an error handler around the regexp code, which would be a good failsafe given regex's are tricky and we do a lot of string munging here now

chejennifer:
> it would be worth adding a test for this, since this ended up causing an irrecoverable bug.
> 
> something else worth adding is an error handler around the regexp code, which would be a good failsafe given regex's are tricky and we do a lot of string munging here now

done

beets (approved):
thanks for the updates!

PR #1470, opened 2022-03-03
[place page] add allowlist for rankings on overview page

chejennifer:
with allowlist: 
Screen Shot 2022-03-03 at 11 15 16 AM

without allowlist: 
Screen Shot 2022-03-03 at 11 15 53 AM


shifucun (approved):
Thanks for the quick fix!

PR #1472, opened 2022-03-07
Website autopush release with import group

shifucun:


pradh (approved):

beets (approved):
Could you expand access to datcom-control?

shifucun:
> Could you expand access to datcom-control?

It should already has access as the current version files are read from there.

beets:
> > Could you expand access to datcom-control?
> 
> It should already has access as the current version files are read from there.

Ah sorry for the confusion! I thought it was a pantheon project (not just a gcs bucket)

PR #1484, opened 2022-03-15
UI for rendering search results

eisenjulian:
This PR adds a new search interface that allows for selection of a place and search query jointly. State is kept in sync with the hash url for easy sharing of links. It talks to an existing API and currently ignores the search part of the query, but in future versions we can integrate with a functional backend.

Screenshot 2022-03-15 at 19 38 39


shifucun:
A high level question is what to show after the search. It's nice to use the place page charts here, but that may restrict the UI to it in an early stage before we come up with a final solution. In current stage, maybe we want to show a list of ranking stat vars (could just be plain text in an expanded form of stat var definitions). Then once we figure out a UI solution, we can add the charting part?

eisenjulian:
I have fixed the tests and uploaded a new version with only a stat var list. PTAL

Screenshot 2022-03-17 at 22 52 43

static/js/tools/rich_search/charts.tsx
```diff
@@ -0,0 +1,63 @@
...
+  const category = "Health";
+  const categoryData = pageData.pageChart[category];
+  const locale = document.getElementById("locale").dataset.lc;
+  const dcid = places[0];
+  const blocks = _.flatten(Object.values(categoryData));
+  return (
+    <section className="block col-12">
+      <div className="row row-cols-xl-3 row-cols-md-2 row-cols-1">
+        {blocks.map((data) => (
+          <ChartBlock
```
shifucun, line 41: why is there still ChartBlock now?

static/js/tools/rich_search/app.tsx
```diff
@@ -0,0 +1,231 @@
...
+  places: string[];
+  addPlace: (string) => void;
+  removePlace: (string) => void;
+  loading: boolean;
+  onClickSearch: () => void;
+  inputInvalid: boolean;
+  chartsData: StatVarsPropType;
+}
+
+// function getPlaceTypes(places: string[]): Promise<{ [key: string]: string }> {
```
shifucun, line 53: Remove the commented code?

run_npm.sh
```diff
@@ -14,7 +14,7 @@
 # limitations under the License.
 
 cd static
-export NODE_OPTIONS=--openssl-legacy-provider
+# export NODE_OPTIONS=--openssl-legacy-provider
```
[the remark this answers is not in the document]
eisenjulian: Done. I am running v17.6.0, which I think it's the latest version. Which version should be used?

server/routes/tools.py
```diff
@@ -20,6 +20,15 @@
 bp = flask.Blueprint("tools", __name__, url_prefix='/tools')
 
 
+@bp.route('/rich_search')
```
beets, line 23: Nit: this should go in the search module?

https://github.com/datacommonsorg/website/blob/master/server/routes/search.py

static/css/tools/rich_search.scss
```diff
@@ -0,0 +1,591 @@
+/**
```
beets, line 1: is this a copy of the place css? please only pull in the css you need, and add more as you go along.. it's harder to extract after the fact and its ok to have a UI that's not-so-great while we iterate..

static/js/tools/rich_search/rich_search.ts
```diff
@@ -0,0 +1,36 @@
...
+import { AppWithContext } from "./app";
+
+window.onload = () => {
+  const locale = document.getElementById("locale").dataset.lc;
+  loadLocaleData(locale, [
+    import(`../../i18n/compiled-lang/${locale}/place.json`),
+    // TODO(beets): Figure out how to place this where it's used so dependencies can be automatically resolved.
+    import(`../../i18n/compiled-lang/${locale}/stats_var_labels.json`),
+    import(`../../i18n/compiled-lang/${locale}/units.json`),
+  ]);
```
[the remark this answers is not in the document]
beets: You can also remove the RawIntlProvider components from app.tsx for that reason..

static/js/place/chart.tsx
```diff
@@ -116,6 +116,10 @@ interface ChartPropType {
    * If the primary place is in USA.
    */
   isUsaPlace: boolean;
+  /**
```
beets, line 119: Please revert the changes to /place/... I believe it's not required anymore?

beets (commented):
Hi Julian - sorry for the late review. Just a few high-level comments..

server/routes/tools.py
eisenjulian: I moved the route to search.py but kept `tools` as part of the path. Should I remove it? Should I also move all other files out of the tools directories?

beets:
Yes, please remove /tools from the path (should be /search or /rich_search
for now), and all the corresponding files should move out of /tools. We use
/tools for our standalone exploration interfaces, while this is
complementary to our entire site (I hope!). Thanks!

eisenjulian:
Thanks, done!

shifucun (approved):
thanks for the update!

server/templates/rich_search.html
```diff
@@ -0,0 +1,38 @@
...
+{% extends 'base.html' %}
+
+{% set is_hide_full_footer = true %}
+{% set subpage_title = "Rich Search" %}
+{% set subpage_url = url_for('search.rich_search') %}
+{% set title = "Rich Search" %}
+
+{% block head %}
+  <link rel="stylesheet" href={{url_for('static', filename='css/rich_search.min.css')}} >
+  <link rel="stylesheet" href="https://fonts.googleapis.com/icon?family=Material+Icons">
```
beets, line 26: Note that you might not need these, and if so, only one of them? Depending on the UI of the page.

beets (approved):
thanks for the updates!

beets:
there are a few errors with building the js - they are related to the files moving out of /tools. you should be able to see this locally too - ./run_test.sh -lb

beets:
Ah, sorry - looks like you'll also need to run `./run_test.sh -f` (that fixes our lint errors)

beets:
Thanks for persisting through these changes!

PR #1488, opened 2022-03-23
Translator page clean up

shifucun:
- Remove old Observation table example
- Clean up unused files
- Remove stale BQ link


static/js/translator/page.tsx
```diff
@@ -140,10 +132,6 @@ export class Page extends React.Component<PagePropType, PageStateType> {
           bindings={this.state.bindings}
           constraints={this.state.constraints}
         ></Translation>
-        <a href="https://console.cloud.google.com/bigquery?sq=443333369001:c5d2972c96704d65a3091ee15d6f0c87">
```
pradh, line 143: Are we removing BQ link in general?  That seems useful?

static/js/translator/page.tsx
```diff
@@ -91,7 +91,7 @@ export class Page extends React.Component<PagePropType, PageStateType> {
           <label id="select-input">
             <input
               type="radio"
-              value="place"
+              value="observation"
```
pradh, line 94: nit: s/Place Observation/Observation/ (below)?

pradh (approved):
Thanks for the quick fix!

static/js/translator/page.tsx
shifucun: This link can not update with the actual query which is misleading here. I want to update the translator here to use a real table name so we can copy over the SQL into GCP console.

static/js/translator/page.tsx
shifucun: Done

shifucun:
Thanks for review!

PR #1489, opened 2022-03-23
Adds support for ML models

nopper:
Hi there,

this PR adds support for deploying and querying custom trained ML models using the [Vertex AI platform](https://cloud.google.com/vertex-ai).

The main contributions are two bash scripts and a python library:
 - `setup_vpc.sh` which sets up the VPC peering allowing to deploy private endpoints in the current infrastructure.
 - `deploy_ai.sh` which deploys a model reading its name and bucket from a yaml file.
 - `ai.py` python library that offers an `InferenceClient` object for interacting with a given model, using different protocols (rest or grpc supported).
 
The model is configured using the `ai.yaml` file. In there you need specify the `endpoint_id` and `deployed_model_id`. This last attribute is optional, as public endpoints over rest api don't need this parameter.

Couple of notes:
1. This can be generalized to multi-model deployment pretty easily.
2. I modified the search interface just to demo that this is working fine. Do you have any suggestions regarding a more thorough testing strategy?


deploy/gke/dev.yaml
```diff
@@ -21,3 +21,7 @@ region:
     - europe-west2
 nodes: 1
 storage_project: datcom-store
+# Network name used for VPC peerings.
+network_name: "default"
+# Assumes a directory with this name exists in gs://datcom-website-dev-modelstore/
```
shifucun, line 26: I guess you also want a field for the model bucket?

deploy/overlays/prod/kustomization.yaml
```diff
@@ -40,6 +40,9 @@ configMapGenerator:
   - name: redis-config
     files:
       - redis.json
+  - name: ai-config
```
shifucun, line 43: Instead of "prod" maybe can first put these in "autopush" folder, which is an instance used for testing.

scripts/setup_nlp.sh
```diff
@@ -0,0 +1,83 @@
...
+
+if [[ $REGION == "" ]]; then
+  REGION=$(yq eval '.region.primary' $ROOT/deploy/gke/$ENV.yaml)
+fi
+
+PROJECT_ID=$(yq eval '.project' $ROOT/deploy/gke/$ENV.yaml)
+PROJECT_NUMBER=$(gcloud projects list --filter="${PROJECT_ID}" --format="value(PROJECT_NUMBER)")
+
+MODEL_NAME=$(yq eval '.model_name' $ROOT/deploy/gke/$ENV.yaml)
+NETWORK_NAME=$(yq eval '.network_name' $ROOT/deploy/gke/$ENV.yaml)
```
shifucun, line 37: maybe we should just use "default" for NETWORK_NAME? Also can remove it from yaml file until we need this to be a variable.

scripts/setup_nlp.sh
```diff
@@ -0,0 +1,83 @@
...
+
+MODEL_BUCKET="gs://${PROJECT_ID}-modelstore"
```
shifucun, line 39: MODEL_BUCKET can be moved to yaml

scripts/setup_nlp.sh
```diff
@@ -0,0 +1,83 @@
...
+MODEL_PATH="${MODEL_BUCKET}/${MODEL_NAME}"
+
+if ! gsutil -q stat "${MODEL_PATH}/*" ; then
+  echo "Model path does not exists ${MODEL_PATH}"
+  
+  # This might fail, meaning that the bucket already exists.
```
shifucun, line 45: If bucket exists, it can not get through the "if" check?

scripts/setup_nlp.sh
```diff
@@ -0,0 +1,83 @@
...
+gcloud beta ai endpoints create \
+    --project="${PROJECT_ID}" \
+    --region="${REGION}" \
+    --network="projects/${PROJECT_NUMBER}/global/networks/${NETWORK_NAME}" \
+    --display-name="${VERTEX_ENDPOINT_NAME}"
+
+ENDPOINT_ID=$(gcloud ai endpoints list --project="${PROJECT_ID}" --region="${REGION}" --filter "display_name=${VERTEX_ENDPOINT_NAME}" --format="value(ENDPOINT_ID.scope())")
+MODEL_ID=$(gcloud ai models list --project="${PROJECT_ID}" --region="${REGION}" --filter "display_name=${VERTEX_MODEL_NAME}" --format="value(MODEL_ID.scope())")
+
+gcloud beta ai endpoints deploy-model "${ENDPOINT_ID}" \
```
shifucun, line 77: Does this script contains one time setup or it is used for each new deployment. If it has one time setup, should move it to $ROOT/gke as "setup_nlp.sh". Then the remaining part can be renamed as "deploy_nlp.sh" since it's not really a setup?

scripts/setup_vpc.sh
```diff
@@ -0,0 +1,62 @@
...
+# limitations under the License.
+
+# This scripts sets up a VPC peering connection to servicenetworking.
+
+set -e
+
+ENV=$1
+REGION=$2
+
+if [[ $ENV != "staging" && $ENV != "prod" && $ENV != "autopush" && $ENV != "dev" && $ENV != "feeding-america" && $ENV != "tidal" && $ENV != "karnataka" ]]; then
```
shifucun, line 23: I think this is one time setup. Can move to $ROOT/gke/setup_vpc.sh

server/configmodule.py
```diff
@@ -102,6 +103,8 @@ class LocalConfig(Config):
     SECRET_PROJECT = 'datcom-website-dev'
     API_ROOT = 'https://autopush.api.datacommons.org'
     RECON_API_ROOT = 'https://autopush.recon.datacommons.org'
+    AI_CONFIG_PATH = os.path.abspath(
```
shifucun, line 106: You can mount it in the GKE. An example is redis.json https://github.com/datacommonsorg/website/blob/master/deploy/overlays/prod/patch_deployment.yaml#L38. Then you don't have do this for each environment as below

server/routes/search.py
```diff
@@ -45,8 +46,10 @@ def search_dc():
     else:
         search_response = {}
 
-    # Convert from search results to template dictionary.
     results = []
+    results.append(ai.search(query_text))
```
shifucun, line 50: Can create a new route like '/search_nlp' which only has result from this?

server/services/ai.py
```diff
@@ -0,0 +1,181 @@
...
+if not cfg.TEST and cfg.AI_CONFIG_PATH:
+    _INFERENCE_CLIENT = create_inference_client(cfg.AI_CONFIG_PATH)
+
+
+def search(query: str) -> Optional[Any]:
+    global _INFERENCE_CLIENT
+
+    if _INFERENCE_CLIENT:
+        response = _INFERENCE_CLIENT.request(query)
+        # This is a fake response for the moment.
```
shifucun, line 179: To confirm, does this return real inference for dev instance?

shifucun:
Thanks for sending this. Overall looks very good and structured! Just some minor comments

scripts/setup_nlp.sh
nopper: Done

nopper: Done.

scripts/setup_nlp.sh
nopper: Done.

scripts/setup_nlp.sh
nopper: This checks if a model path exists in the specified bucket. If the bucket exists but a model is not there this will cause the script to exit.

I removed the bucket creation in case this fails as it might not be useful. It should be the user responsibility I guess to make sure everything is in check.

scripts/setup_nlp.sh
nopper: Moved it to `deploy_ai.sh` which is a more generic name.

scripts/setup_vpc.sh
nopper: Done.

deploy/gke/dev.yaml
nopper: Done.

server/services/ai.py
nopper: Yes. I tested it using the public endpoint. But I also like to test it on a dev environment to make sure the grpc client works as expected.

Do you have any instructions on how to do that?

server/routes/search.py
nopper: Done.

server/services/ai.py
nopper: Tested on dev. Seems to work fine.

nopper:
PTAL. I've addressed all your comments and tested it on dev.

deploy/overlays/autopush/ai.yaml
```diff
@@ -0,0 +1,4 @@
+us-central1:
```
shifucun, line 1: I don't see model deployed in datcom-website-autopush?

deploy/overlays/dev/kustomization.yaml
```diff
@@ -50,3 +53,6 @@ patchesStrategicMerge:
           maxSurge: 2
           # maximum number of Pods that can be unavailable during the update process
           maxUnavailable: 75%
+
+patchesStrategicMerge:
```
shifucun, line 57: Can you move line 41-54 to patch_deployment.yaml too?

deploy/overlays/dev/patch_deployment.yaml
```diff
@@ -0,0 +1,36 @@
+# Copyright 2019 Google LLC
```
shifucun, line 1: 2022

deploy/overlays/prod/patch_deployment.yaml
```diff
@@ -31,8 +31,13 @@ spec:
         - name: redis-config
           configMap:
             name: redis-config
+        - name: ai-config
```
shifucun, line 34: Maybe should move this to $ROOT/deploy/base/deployment.yaml? So this will be a shared pattern. Then a lot of patch above can be saved.

gke/setup_vpc.sh
```diff
@@ -0,0 +1,54 @@
...
+
+set -e
+
+ENV=$1
+REGION=$2
+
+if [[ $ENV != "staging" && $ENV != "prod" && $ENV != "autopush" && $ENV != "dev" && $ENV != "feeding-america" && $ENV != "tidal" && $ENV != "karnataka" ]]; then
+  echo "First argument should be 'staging' or 'prod' or 'autopush' or 'dev' or 'feeding-america' or 'tidal' or 'karnataka'"
+  exit
+fi
```
shifucun, line 26: I have removed this check in other scripts too, since this is not scalable.. Can remove there.

shifucun (commented):
Can you revert the change for mixer submodule update?

scripts/deploy_ai.sh
```diff
@@ -0,0 +1,79 @@
...
+set -e
+
+ENV=$1
+REGION=$2
+
+if [[ $ENV != "staging" && $ENV != "prod" && $ENV != "autopush" && $ENV != "dev" && $ENV != "feeding-america" && $ENV != "tidal" && $ENV != "karnataka" ]]; then
+  echo "First argument should be 'staging' or 'prod' or 'autopush' or 'dev' or 'feeding-america' or 'tidal' or 'karnataka'"
+  exit
+fi
+
```
shifucun, line 25: Can remove the check here...

server/configmodule.py
shifucun: Got it, that makes sense.

shifucun (commented):
Overall looks good! Just some nit about the ai.yaml config change. Really appreciate sending the PR and testing it throughout!

gke/setup_vpc.sh
nopper: Done.

deploy/overlays/autopush/ai.yaml
nopper: Changed with an empty file for now.

nopper:
Hi @shifucun . Addressed your comments. PTAL.

deploy/git/mixer_hash.txt
```diff
@@ -0,0 +1 @@
+be02c54
```
shifucun, line 1: please revert this change

deploy/git/website_hash.txt
```diff
@@ -0,0 +1 @@
+5d47748
```
shifucun, line 1: please revert this change

deploy/overlays/dev/kustomization.yaml
```diff
@@ -34,6 +34,9 @@ configMapGenerator:
     literals:
       - mixerProject=datcom-website-dev
       - serviceName=website-esp.endpoints.datcom-website-dev.cloud.goog
+  - name: ai-config
+    files:
+      - ai.yaml
```
shifucun, line 39: can move this to base/kustomization.yaml as well and define empty ai.yaml for all env that has no model deployed

deploy/overlays/prod/patch_deployment.yaml
shifucun, line 34: This can be removed since it's added to base

deploy/overlays/prod/patch_deployment.yaml
```diff
@@ -31,8 +31,13 @@ spec:
...
+          configMap:
+            name: ai-config
       containers:
         - name: website
           volumeMounts:
             - name: redis-config
               mountPath: /datacommons/redis
+            - name: ai-config
```
shifucun, line 42: can be removed as well

shifucun (commented):
The mixer submodule is still not reverted. Please revert.

deploy/overlays/prod/patch_deployment.yaml
nopper: Done.

deploy/overlays/dev/kustomization.yaml
nopper: Resolved offline.

shifucun (approved):
Thanks for making all the update!

PR #1490, opened 2022-03-24
[Search] new search page 

chejennifer:
- create the new search page (you can view this on [dev](https://dev.datacommons.org/search2))

screen-recording-_5_

Some refactoring along the way:
- file rename of our various search components
- extract out some of the axios calls to functions that return a promise
- update placeIdToDcid flask endpoint to take a list of placeId instead of just single placeId


shifucun:
Instead of "Search Data Commons", should we have other example search prompt?

chejennifer:
> 

1. Do you mean for the places section? or the stat var section as well? I think 2 options for adding link to browser would be:
    a. another line that says "Open in Graph Browser" 
    b. make the dcid part clickable and make clicking that take the user to the Graph Browser
2. I think that makes sense but does that mean user has to search something in that search bar in order to get to the search page? 
3. Ok, that can be easily changed!

chejennifer:
> This looks is really cool!
> 
> peanut gallery:
> 
> * should we include link to browser page?
> * now that we have a single search bar, should we allow people to search on datacommons.org itself (more convenient/natural) than click to the search page?
> * I think for many places maps/scatter won't be helpful, but timeline always will... we can improve later, but for now, should we put timeline before the other two?

1. Do you mean for the places section? or the stat var section as well? I think 2 options for adding link to browser would be:
    a. another line that says "Open in Graph Browser"
    b. make the dcid part clickable and make clicking that take the user to the Graph Browser
2. I think that makes sense but does that mean user has to search something in that search bar in order to get to the search page?
3. Ok, that can be easily changed!



> Instead of "Search Data Commons", should we have other example search prompt?

I'm not sure we need an example search prompt because there is no structure that user has to follow. User can essentially type anything they want.

pradh:
> 
> 1. Do you mean for the places section? or the stat var section as well? I think 2 options for adding link to browser would be:
>    a. another line that says "Open in Graph Browser"
>    b. make the dcid part clickable and make clicking that take the user to the Graph Browser

(b) sounds better to me.

> 2. I think that makes sense but does that mean user has to search something in that search bar in order to get to the search page?

Oh I just mean however it was before :).  User types in something in the search bar then the search page shows up with results...

I really like this unified search btw!

static/js/search/place_results.tsx
```diff
@@ -0,0 +1,121 @@
...
+            <div className="search-results-list">
+              {props.places.map((place) => {
+                return getResultItemContentJsx(place, placeNames[place]);
+              })}
+            </div>
+            <a
+              className="search-results-place-explore-all"
+              href={PLACE_REDIRECT_PREFIX}
+            >
+              Explore All Places
```
beets, line 74: We should work on preloading the search box there so it's not too abrupt of a change (since the other actions on this page leads to more results). Could you add a todo?

static/js/search/place_results.tsx
```diff
@@ -0,0 +1,121 @@
...
+        <div className="search-results-item-byline">dcid: {placeDcid}</div>
+        <a href={PLACE_REDIRECT_PREFIX + placeDcid}>Open in place explorer</a>
+      </div>
+      <div className="search-results-place-links">
+        <a
+          href={MAP_REDIRECT_PREFIX + placeDcid}
+          className="search-results-link-icon"
+          id={`map-${placeDcid}`}
+        >
+          <i className="material-icons-outlined">public</i>
```
beets, line 102: nit: public --> Map  Explorer (these strings show up before the icons are loaded so we should put something meaningful).

static/js/search/results.tsx
```diff
@@ -0,0 +1,102 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * Results section of the search page
```
beets, line 18: Nit: maybe clarify that this is for place and stat var results. Should we rename this file to all_results.tsx?

static/js/search/search_input.tsx
```diff
@@ -0,0 +1,285 @@
...
+              redirectAction(inputText, "");
+            }}
+            tabIndex={-1}
+          >
+            <input
+              className="search-input-box"
+              type="text"
+              value={inputText}
+              onChange={(event) => onInputChanged(event.target.value)}
+              placeholder="Search Data Commons"
```
beets, line 78: Maybe the placeholder can be long, eg "Search for places, variables and more on Data Commons"

static/js/search/place_results.tsx
```diff
@@ -0,0 +1,121 @@
...
+  }, [props]);
+
+  if (_.isEmpty(placeNames)) {
+    return <></>;
+  }
+
+  return (
+    <div className="search-results-place search-results-section">
+      {props.selectedPlace ? (
+        <div className="search-results-place-selected">
```
beets, line 55: Nit: we should match this behavior for stat var results too (when a user selects a particular stat var, we can display it highlighted on enter). You could add a TODO for now.

static/js/search/search_input.tsx
```diff
@@ -0,0 +1,285 @@
...
+          if (query !== latestQuery.current) {
+            return;
+          }
+          setShowResults(false);
+          setResults({ placeResults: [], svResults: [] });
+        });
+    }
+  }
+
+  function onAutocompleteCompleted(
```
beets, line 184: nit: maybe clarify in the handler name that this is for PlaceAutocomplete

static/js/search/search_input.tsx
```diff
@@ -0,0 +1,285 @@
...
+const REDIRECT_PREFIX = "/search2?";
+
+export function SearchInput(props: SearchInputPropType): JSX.Element {
+  const delayTimer = useRef(null);
+  const placeAutocompleteService = useRef(null);
+  const latestQuery = useRef(null);
+  const [hoveredIdx, setHoveredIdx] = useState(0);
+  const [results, setResults] = useState({ placeResults: [], svResults: [] });
+  const [inputText, setInputText] = useState(props.query);
+  const [showResults, setShowResults] = useState(false);
```
beets, line 45: nit: please add some comments for these properties?

static/js/search/search_input.tsx
```diff
@@ -0,0 +1,285 @@
...
+        }
+        setShowResults(false);
+        setResults({ placeResults: [], svResults: [] });
+      });
+  }
+
+  function handleKeydownEvent(
+    event: React.KeyboardEvent<HTMLDivElement>
+  ): void {
+    if (showResults) {
```
beets, line 230: maybe do an early return here if !showResults to simplify this handler

static/js/search/statvar_results.tsx
```diff
@@ -0,0 +1,83 @@
...
+        {canExpand && (
+          <div
+            onClick={() => setShowAll(true)}
+            className="search-results-sv-show-more"
+          >
+            show more
+          </div>
+        )}
+        {showAll && (
+          <div
```
beets, line 73: In this mode, should we go full width of the results container?

static/js/shared/types.ts
```diff
@@ -145,3 +145,15 @@ export interface LinkedNodes {
 export interface GraphNodes {
   nodes: GraphNode[];
 }
+
+export interface SvgSearchResult {
```
beets, line 149: Nit: can we document which API call these are from?

static/js/utils/search_utils.test.tsx
```diff
@@ -101,63 +89,3 @@ test("getHighlightedJSX", () => {
     }
   }
 });
-
-test("getResultCountString", () => {
```
beets, line 105: can we keep this test?

beets (approved):
thanks for the refactor too!

beets:
> Oh I just mean however it was before :). User types in something in the search bar then the search page shows up with results...

+1: we can do that when we are ready to "launch" this, since that will be in our headers. while we're developing, this page is only in non-prod

server/templates/search2.html
```diff
@@ -0,0 +1,42 @@
...
+  See the License for the specific language governing permissions and
+  limitations under the License.
+ #}
+
+{% extends 'base.html' %}
+
+{% set main_id = 'search' %}
+{% set title = 'Search' %}
+{% set subpage_title = 'Search' %}
+{# TRANSLATORS: Title of a page for searching the site, or search results. #}
```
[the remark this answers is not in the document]
chejennifer: done

server/templates/search2.html
```diff
@@ -0,0 +1,42 @@
...
+
+{% block head %}
+<link rel="stylesheet" href={{url_for('static', filename='css/search2.min.css', t=config['VERSION'])}} >
+<link rel="stylesheet" href="https://fonts.googleapis.com/icon?family=Material+Icons">
+<link rel="stylesheet" href="https://fonts.googleapis.com/icon?family=Material+Icons+Outlined">
```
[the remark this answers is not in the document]
chejennifer: done

static/js/search/place_results.tsx
chejennifer: done

static/js/search/place_results.tsx
chejennifer: I don't think that's something that can be changed .. public is the name of the icon, it's defined by material icons

static/js/search/results.tsx
chejennifer: done. Since this could have other results in the future, clarified that this is all results besides the custom search engine results

static/js/search/search_input.tsx
chejennifer: done

static/js/search/place_results.tsx
chejennifer: done

static/js/search/search_input.tsx
chejennifer: done

static/js/search/search_input.tsx
chejennifer: done

static/js/search/search_input.tsx
chejennifer: done

static/js/search/statvar_results.tsx
chejennifer: I'm curious what the advantages of going the full width of the results container would be? currently, users don't need to scroll horizontally & when clicking show more on a list, I don't think expanding horizontally makes as much sense as expanding vertically

static/js/shared/types.ts
chejennifer: done

static/js/utils/search_utils.test.tsx
chejennifer: ah yes, that was an accidental deletion

static/js/search/place_results.tsx
beets: oh yes, i'd forgotten that.

static/js/search/statvar_results.tsx
beets: ah, i meant to expand both vertically and horizontally (or rather, we can keep it full width always). it seems awkward that it's half-width. we could update this in another iteration too

beets (approved):
thanks for the updates!

static/js/search/statvar_results.tsx
chejennifer: ok, will do this in next iteration

PR #1491, opened 2022-03-24
[webdriver] fix tests

chejennifer:
- fix test break caused by new "Agriculture" vertical added to stat var widget


shifucun (approved):
this is nice!

PR #1493, opened 2022-03-25
[topic] generate list of child places per config; update landing page

beets:
- Add support for contained places (instead of listing all place id's per config)
- Added better landing page
- Brought back page selector (with a minor edit of all places for a topic listed on the right, and no support for topics without a place specified)
- Lazy load topic configs
- Add a config: equity for all us states (mostly similar as USA equity, but slight differences due to missing sv's)

image

image


server/config/topic_page.proto
```diff
@@ -20,14 +20,23 @@ package datacommons;
...
   string topic_name = 2;
   // List of places supported by this config.
   repeated string place_dcid = 3;
-  // Contained places to load for each parent place type.
+  // List of places within a containing place to generate topics for.
+  repeated PlacesWithin places_within = 4;
+  // Contained places to load for each parent place type, for use in charts.
   // This can be overwritten by tiles.
-  map<string, string> contained_place_types = 4;
+  map<string, string> contained_place_types = 5;
```
chejennifer, line 39: is this still needed now that there is the new PlacesWithin with a parent place and child type?

server/config/topic_page/equity/USA_states.textproto
```diff
@@ -0,0 +1,1215 @@
...
+    # right_tiles {
+    #   title: "Unemployment rate by sex"
+    #   type: LINE
+    #   stat_var_key: "unemployment_female"
+    #   stat_var_key: "unemployment_male"
+    # }
+  }
+  blocks {
+    right_tiles {
+      title: "Median earnings of females, down by establishment"
```
chejennifer, line 84: "down by establishment" -> "by establishment"

server/config/topic_page/equity/USA_states.textproto
```diff
@@ -0,0 +1,1215 @@
...
+        show_highest: true
+      }
+      stat_var_key: "disability_rate_female"
+      stat_var_key: "disability_rate_male"
+    }
+  }
+
+  # Note: This data is not available by race.
+  blocks {
+    title: "Education attainment by sex"
```
chejennifer, line 207: nit: for consistency with all other titles, maybe "Education attainment, by sex"

server/config/topic_page/equity/USA_states.textproto
```diff
@@ -0,0 +1,1215 @@
...
+        show_lowest: true
+      }
+      stat_var_key: "associates_female"
+      stat_var_key: "associates_male"
+    }
+  }
+
+  blocks {
+    # Note: title intentionally left out
+    description: "Percentage of the population of each sex with an bachelors's degree"
```
chejennifer, line 228: nit: "with a bachelor's degree"

chejennifer (approved):
Nice changes!

server/config/topic_page.proto
beets: yes, this specifies what type of child places to include in the charts, whereas PlacesWithin specifies which places to generate for. the intent was to support the use case where they are different sometimes (or we might not always use PlacesWithin to generate a list of places for the config).

server/config/topic_page/equity/USA_states.textproto
beets: done

server/config/topic_page/equity/USA_states.textproto
beets: thanks for the catch, fixed this elsewhere too

server/config/topic_page/equity/USA_states.textproto
beets: done

beets (commented):
thanks for the review!

server/config/topic_page.proto
chejennifer: oh, I see, ok, makes sense!

PR #1496, opened 2022-03-30
[timeline] add source selector

chejennifer:
- add source selector for timeline tool
- separate data fetching from data processing so that data only needs to be fetched if place, stat vars, or denom changes
- move stat api response types from shared_utils.ts to stat_type.tsx
screen-recording-_11_


pradh:
This is very cool!

Question as a lay audience (not having read the code), does the source choice get encoded in the URL?  Should we encode the default choice too?  (I just asked a related question on datcom-core, so this is very timely)

static/js/shared/stat_types.ts
```diff
@@ -14,6 +14,8 @@
  * limitations under the License.
  */
 
+import { StatMetadata } from "../tools/shared_util";
```
beets, line 17: nit: since it's a type from our API response, should we keep that here (in stat_types), and import into shared_util?

static/js/tools/timeline/data_fetcher.ts
```diff
@@ -221,77 +249,170 @@ export function fetchStatData(
...
+  });
+}
+
+/**
+ * Given raw data and timeline chart options, gets the StatData for the chart.
+ * @param rawData raw timeline data
+ * @param places places to get data for
+ * @param statVars stat vars to get data for
+ * @param metahashMap map of stat var dcid to metahash
+ * @param isRatio whhether or not to calculate ratio
```
beets, line 326:
```suggestion
 * @param isRatio whether or not to calculate ratio
```

static/js/tools/timeline/data_fetcher.ts
```diff
@@ -221,77 +249,170 @@ export function fetchStatData(
...
+        for (const series of sourceSeries) {
+          const metadata = {
+            ...series,
+            provenanceUrl: series.provenanceDomain,
+          };
+          // a SourceSeries object holds the data values in the val field and
+          // the rest of the object contains the metadata of that source
+          // series.
+          delete metadata.val;
+          // StatMetdata uses provenanceUrl for what SourceSeries stores in
```
beets, line 296:
```suggestion
          // StatMetadata uses provenanceUrl for what SourceSeries stores in
```

static/js/tools/timeline/data_fetcher.ts
```diff
@@ -221,77 +249,170 @@ export function fetchStatData(
...
-          const mmethod = timeSeries.metadata.measurementMethod;
-          if (mmethod) {
-            result.measurementMethods.add(mmethod);
-          }
+
+      if (isRatio) {
+        const hasDenomSeries =
+          place in rawData.denomData && denom in rawData.denomData[place].data;
+        if (!hasDenomSeries) {
+          continue;
```
beets, line 382: we should show an error message in this case. add a todo, or a simple message in below the chart?

static/js/tools/timeline/util.ts
```diff
@@ -152,3 +153,35 @@ export function getDenom(mprop: string): string {
   }
   return "";
 }
+
+// Set metahash map (a map of stat var dcid to hash representing the source to
```
beets, line 157:
```suggestion
// Set metahash map (a map of stat var dcid to hash representing the source
```

static/js/tools/timeline/util.ts
```diff
@@ -152,3 +153,35 @@ export function getDenom(mprop: string): string {
...
+// of data to use) in the URL
+export function setMetahash(metahash: Record<string, string>): void {
+  const urlParams = new URLSearchParams(window.location.hash.split("#")[1]);
+  let metahashMap = JSON.parse(urlParams.get("metahash"));
```
beets, line 161: can we start adding constants for the URL params? and maybe a shorter string here, `mh` or `meta`?

beets (approved):
thanks for the refactors!

static/js/tools/timeline/data_fetcher.ts
chejennifer: added a TODO. We should think more about what we want in the error message and when to show the error message. For example, if there are multiple places chosen and only one of the places is missing a denomSeries, I don't think we necessarily want to show an error message

static/js/tools/timeline/util.ts
chejennifer: done

static/js/tools/timeline/util.ts
chejennifer: done

static/js/tools/timeline/data_fetcher.ts
chejennifer: added a TODO. We should think more about what we want in the error message and when to show the error message. For example, if there are multiple places chosen and only one of the places is missing a denomSeries, I don't think we necessarily want to show an error message

static/js/tools/timeline/data_fetcher.ts
chejennifer: done

static/js/tools/timeline/data_fetcher.ts
chejennifer: done

static/js/shared/stat_types.ts
chejennifer: done

PR #1497, opened 2022-03-31
Add IITM specific page setup in website

shifucun:
Also updated the instruction on how to setup a private instance.

We should consider simplify the process as it really involves a lot of config change now.


beets (approved):
oh wow, yes - thanks for setting this up

PR #1500, opened 2022-03-31
Remove weather observation section in browser page

shifucun:
Weather data has been added as a regular stat var observations and can be used from the variable hierarchy.


beets (approved):
thanks!

pradh:
Should we wait till daily data is in?

PR #1503, opened 2022-04-01
[place] handle empty dcids when getting i18n name

chejennifer:
- some non english place pages were erroring out (caught by webdriver) because of trying to get cached_i18n_name for an empty list of dcids

### DIFF
--- a/server/routes/api/place.py
+++ b/server/routes/api/place.py
@@ -113,7 +113,7 @@ def cached_i18n_name(dcids, locale, should_resolve_all):
     Returns:
         A dictionary of place names, keyed by dcid (potentially sparse if should_resolve_all=False)
     """
-    if not dcids: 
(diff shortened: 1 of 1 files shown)


beets (approved):
Was it a place explorer test that was having this issue? It sounds like there might be a setup issue if an empty dcid was in the request. But this is a fair change to make.

chejennifer:
> Was it a place explorer test that was having this issue? It sounds like there might be a setup issue if an empty dcid was in the request. But this is a fair change to make.

The problem occurred when trying to get display names. This is because when getting display name, we try to get the state codes for each place and to get state codes for each place when locale is not "en", we use cached_i18n_names to get the name for each state. However, there were no states to get names for causing an error to be thrown. I'm not sure why this error wasn't being thrown before though ..

PR #1504, opened 2022-04-02
[analytics] disable consent manager (and just disble client storage)

beets:
Also made some corresponding changes to tag manager. I'm not sure if that's why the stats dropped, since the diff from the release seems benign except for the new settings.

https://github.com/datacommonsorg/website/compare/v1.13.1...v1.14.0

Disabling client_storage should turn off cookies too. Will test this post release (also added back our existing GA to staging for testing).

### DIFF
--- a/server/templates/base.html
+++ b/server/templates/base.html
@@ -194,6 +194,7 @@
   <script src="https://maxcdn.bootstrapcdn.com/bootstrap/4.0.0/js/bootstrap.min.js"
     integrity="sha384-JZR6Spejh4U02d8jOt6vLEHfe/JQGiRRSQQxSfFWpi1MquVdAyjUar5+76PVCmYl"
     crossorigin="anonymous"></script>
+  {# Enable Google Analytics with cookieless tracking. #}
   {% if GA_ACCOUNT %}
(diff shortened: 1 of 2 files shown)


server/templates/base.html
```diff
@@ -194,18 +194,15 @@ <h6>{% trans %}Data Commons{% endtrans %}</h6>
...
   <script>
     window.dataLayer = window.dataLayer || [];
-    function gtag() { dataLayer.push(arguments); }
+    function gtag(){dataLayer.push(arguments);}
     gtag('js', new Date());
-    gtag('config', '{{ GA_ACCOUNT }}');
-    gtag('consent', 'default', {'ad_storage': 'denied'});
-    gtag('consent', 'default', {'analytics_storage': 'denied'});
-    gtag('set', {'allow_google_signals': false});
+    gtag('config', '{{ GA_ACCOUNT }}', { client_storage: 'none', anonymize_ip: true });
```
pradh, line 203: Maybe add some comments on rationale for setting these params?

pradh (approved):
Thanks!

Just to understand, is the summary that we will find out if this change restores original behavior when we release it?

beets:
yup, we'll have to deploy to test. i had tested our existing settings with tag manager and nothing was flagged :(

PR #1506, opened 2022-04-04
[Search] search page tweaks

chejennifer:
- add spinner when place and stat var results are loading
- add result count for stat vars
- make stat var results section full width
- highlight stat var result if it is chosen from autocomplete dropdown (same behavior as place being chosen)
- preload the search bar

dev updated: [dev.datacommons.org/search2](https://dev.datacommons.org/search2)

screen-recording-_12_


static/js/search/all_results.tsx
```diff
@@ -82,32 +83,57 @@ export function AllResults(props: ResultsProps): JSX.Element {
...
-      })
-      .catch(() => {
-        setStatVars([]);
-      });
+    } else {
+      setStatVars([]);
+    }
   }, [props]);
 
   if (_.isNull(places) || _.isNull(statVars)) {
```
beets, line 108: could you add some comments as to the required states for these? null is pre-request and [] if there are no results. this would also be easier to read if it was "placeResults" and "statVarResults"

beets (approved):
nice updates!

another small request is if we can left align the text in the search box with the results (you can start the search box 5px to the right of where it is.

static/js/search/all_results.tsx
chejennifer: done

PR #1508, opened 2022-04-04
Add missing library for AI model service

shifucun:
I will discuss later with @nopper and see if can remove "numpy" deps as it is only used for one function.

This will fix the deployment error for now.


beets:
ok, please add a comment to the file and let's have numpy removed soon..

shifucun:
> ok, please add a comment to the file and let's have numpy removed soon..

Yea, I added note to replace them with raw operations.

chejennifer (approved):

PR #1511, opened 2022-04-05
[homepage] Highlight sustainability datasets

beets:
Highlight new sustainability datasets
Update DC numbers

image

### DIFF
--- a/server/templates/static/homepage.html
+++ b/server/templates/static/homepage.html
@@ -93,17 +93,23 @@
 <section id="by-the-numbers">
   <div class="container">
     <div class="highlight">
-      <h3>By the numbers</h3>
-      <p>We have continuously added data to the Data Commons graph over the last 4 years. As of April 2022, the graph contains:</p>
-      <ul>
-        <li>570 billion triples</li>
-        <li>2.8 billion time series</li>
-        <li>2.9 million places</li>
-        <li>90,000 variables</li>
-      </ul>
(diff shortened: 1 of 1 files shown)


server/templates/static/homepage.html
```diff
@@ -93,20 +93,26 @@ <h2>Explore the data</h2>
...
-      <ul>
-        <li>515 billion triples</li>
-        <li>2.8 billion time series</li>
-        <li>2.9 million places</li>
-        <li>90,000 variables</li>
-        <li>250 datasets</li>
-      </ul>
-      <p>Separately, the Biomedical Data Commons includes 200,000 variables and 850 billion triples.</p>
+      <h3>New update</h3>
+      <p>Data Commons now includes <a href="https://docs.datacommons.org/datasets/sustainability.html">100+ sources of Sustainability data</a>, covering topics from climate predictions (CMIP 5 and CMIP 6) from NASA, emissions from EPA, energy from EIA, NREL and UN, disasters from USGS and USFS, health from CDC and <a href="https://docs.datacommons.org/datasets/sustainability.html">more</a>.</p>
```
shifucun, line 97: covering topics of? (instead of from)

shifucun (approved):

server/templates/static/homepage.html
beets: sorry i had automerge turned on. i believe "from" is better here (though i can't point you to the exact grammar rule for this). i've also run this change by comms.

shifucun: Got it. Thanks!

PR #1515, opened 2022-04-05
Fix webdriver tests

shifucun:
And I plan to make webdriver tests enforced again.

### DIFF
--- a/server/routes/api/place.py
+++ b/server/routes/api/place.py
@@ -710,4 +710,4 @@ def placeid2dcid():
             if outIds and inId:
                 result[inId] = outIds[0]
         return Response(json.dumps(result), 200, mimetype='application/json')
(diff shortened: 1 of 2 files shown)


server/routes/api/place.py
```diff
@@ -709,4 +710,4 @@ def placeid2dcid():
             if outIds and inId:
                 result[inId] = outIds[0]
         return Response(json.dumps(result), 200, mimetype='application/json')
-    abort(404, 'no valid dcid not found for %s' % place_id)
\ No newline at end of file
+    abort(404, 'no valid dcid not found for %s' % place_ids)
```
chejennifer, line 713: nit: no valid dcid found for

chejennifer (approved):
thanks for the fix!

server/routes/api/place.py

PR #1516, opened 2022-04-05
[Search] Update to use new field in search stat var mixer endpoint

chejennifer:
- new field sv_only was added to the stat-var/search mixer endpoint. These are the website changes to use that new field


shifucun (approved):
This looks much simpler

PR #1521, opened 2022-04-07
Removes GRPC inference client

nopper:
* Removes GRPC inference client and all dependencies.
* Pros: reduced docker size as lots of heavy dependencies are not required anymore (tensorflow, numpy).
* Cons: increased latency.


shifucun (approved):
Thanks for making the change! I wonder what is the latency comparison for grpc and rest. Do you have some number from local testing?

Issue #1525, opened 2022-04-08
[BUG] SV Hierarchy name prefix shouldn't split words


pradh:
![image](https://user-images.githubusercontent.com/4375037/162455655-615d3b26-3717-40b3-9b8e-00b445972972.png)

PR #1527, opened 2022-04-08
[tree widget] fix prefix bug

chejennifer:
fixes [issue #1525](https://github.com/datacommonsorg/website/issues/1525)


beets (approved):
thanks for the fix!

can we add a test for this function too?

chejennifer:
> thanks for the fix!
> 
> can we add a test for this function too?

done

PR #1529, opened 2022-04-11
IITM DC: First release

SenthamizhanV:
- Added homepage and about page for IITM DC
- Edited base.html to generalize navbar and footer for IITM DC
- Add CSS for IITM DC


SenthamizhanV:
Adding screenshots for reference

iitm-dc
iitm-dc1
iitm-dc2
iitm-dc4

PR #1530, opened 2022-04-11
add tick marks to time slider

n-h-diaz:
looks like
Screen Shot 2022-04-11 at 2 46 36 PM


static/js/tools/map/time_slider.tsx
```diff
@@ -161,6 +164,27 @@ export function TimeSlider(props: TimeSliderProps): JSX.Element {
                 x1={SLIDER_MARGIN}
                 x2="100%"
               ></line>
+              {document.getElementById("time-slider-slide") &&
+                ticks.map((element) => {
```
chejennifer, line 168: nit: "element" is a bit of a confusing name for the variable because each tick isn't yet an element .. maybe just "tick" or "date"

chejennifer (approved):

PR #1531, opened 2022-04-12
Borrow charts from category page when overview page category has no data

shifucun:


beets (approved):
nit: can we add a todo for a test for this? perhaps webdriver is the best.. at the very least, we should add some of these pages to our release test set.

thanks for the quick fix!

shifucun:
> nit: can we add a todo for a test for this? perhaps webdriver is the best.. at the very least, we should add some of these pages to our release test set.
> 
> thanks for the quick fix!

Sure. I will make a follow up PR for test and update the release doc

PR #1534, opened 2022-04-12
Fix data sparsity issue in place page

shifucun:
- Borrow category data from all categories into Overview page.
- For a set of India cities, use the District place as search result.
- [Bonus]: remove additional stat vars for place page data fetch.

### DIFF
--- a/static/js/place/place_autocomplete.ts
+++ b/static/js/place/place_autocomplete.ts
@@ -19,6 +19,53 @@ import { getPlaceDcids } from "../utils/place_utils";
 let ac: google.maps.places.Autocomplete;
 let acs: google.maps.places.AutocompleteService;
 
+// Some India cities have limited data, override to show the corresponding district page.
+const PLACE_OVERRIDE = {
+  "wikidataId/Q1156": "wikidataId/Q2341660",
+  "wikidataId/Q987": "wikidataId/Q1353",
+  "wikidataId/Q1355": "wikidataId/Q806463",
+  "wikidataId/Q1361": "wikidataId/Q15340",
+  "wikidataId/Q1070": "wikidataId/Q401686",
+  "wikidataId/Q1352": "wikidataId/Q15116",
+  "wikidataId/Q1348": "wikidataId/Q2088496",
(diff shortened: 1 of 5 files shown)


static/js/place/place_autocomplete.ts
```diff
@@ -19,6 +19,53 @@ import { getPlaceDcids } from "../utils/place_utils";
 let ac: google.maps.places.Autocomplete;
 let acs: google.maps.places.AutocompleteService;
 
+// Some India cities have limited data, override to show the corresponding district page.
```
beets, line 22: can we put a todo to move this to flask? (so we do the override in the place2dcid call - to catch all our place searches, and to reduce the size of our js).

beets (approved):
thanks for the fix!

is there also a case where we might have linked to category pages (which are now blank or have been merged into overview) and we would need to redirect? for straggler links to places with categories selected. a lower priority fix, in the grand scheme of things..

static/js/place/place_autocomplete.ts
shifucun: Moved in this PR

shifucun:
> thanks for the fix!
> 
> is there also a case where we might have linked to category pages (which are now blank or have been merged into overview) and we would need to redirect? for straggler links to places with categories selected. a lower priority fix, in the grand scheme of things..

Right now we don't show empty category pages. If there is no data, then the category page will not show up. Now only category with data is merged into the Overview page.

shifucun:
> > > thanks for the fix!
> > > is there also a case where we might have linked to category pages (which are now blank or have been merged into overview) and we would need to redirect? for straggler links to places with categories selected. a lower priority fix, in the grand scheme of things..
> > 
> > 
> > Right now we don't show empty category pages. If there is no data, then the category page will not show up. Now only category with data is merged into the Overview page.
> 
> Ah yes, but for these pages that will be "losing" their category pages, we should redirect back to the overview if there are old links online / bookmarked.. e.g. https://autopush.datacommons.org/place/wikidataId/Q15116?category=Health

Now I see what you mean.. I will bundle this with webdriver tests in a PR

beets:
thank you!!

PR #1537, opened 2022-04-12
update homepage links

### LINKED ISSUE #1523: [BUG] Scatter picture doesn't match what's on the front page
The image on the front page:

image

Is a bit different from what you see when you click on the page:

image

The couple of outlier counties do show up, but there is one more... perhaps we could update the screenshot?

beets:
removed fixed denom for map charts
updated filter for some links to match images (fixes #1523)


chejennifer (approved):
thanks for the fix!

PR #1538, opened 2022-04-12
[tree widget] update flask endpoint to match mixer stat-var/search endpoint change

chejennifer:


shifucun (commented):
There is also "ENABLE_BLOCKLIST" in configmodule.py, should remove that? Need to sync mixer submodule so this commit is self-complete

chejennifer:
> There is also "ENABLE_BLOCKLIST" in configmodule.py, should remove that? Need to sync mixer submodule so this commit is self-complete

No, that is used for getting the groups and stat vars for the hierarchy part. & ok will sync mixer submodule

beets (approved):

shifucun:
> > There is also "ENABLE_BLOCKLIST" in configmodule.py, should remove that? Need to sync mixer submodule so this commit is self-complete
> 
> No, that is used for getting the groups and stat vars for the hierarchy part. & ok will sync mixer submodule

Got it. LG then. Please sync mixer module before merge.

shifucun (approved):

PR #1541, opened 2022-04-13
[homepage] link and caption updates

beets:
- fix scatter images that have gotten out of sync (fixed the date in the link)
- update some wordings (we use percentage, where it's really a fraction)
- `wh` is not a unit (i'd meant for it to be `vh`)


chejennifer (approved):
thanks for the fixes!

PR #1546, opened 2022-04-13
[tools] chart options footer update

chejennifer:
- update "ratio of" option to "per capita" checkbox
- start with chart options expanded
screen-recording-_20_
screen-recording-_19__1
screen-recording-_18_


beets (approved):
thanks for the quick fix

PR #1550, opened 2022-04-13
[map/scatter] add per capita in title

chejennifer:
- when per capita is chosen, add per capita into chart title
Screen Shot 2022-04-13 at 2 11 01 PM
Screen Shot 2022-04-13 at 2 11 28 PM


beets (approved):
thanks!

PR #1554, opened 2022-04-14
Some UI updates

chejennifer:
- add feedback link at the bottom of charts 
Screen Shot 2022-04-14 at 10 27 55 AM

- timeline bug where not all IPCC models have projection lines [http://shortn/_WIqRo2A2n7](http://shortn/_WIqRo2A2n7) vs:
Screen Shot 2022-04-14 at 10 28 43 AM

- Change climate category to environment on place pages
- Add link to map explorer from place page map charts
screen-recording-_21_


static/js/place/place.ts
```diff
@@ -40,6 +40,9 @@ const Y_SCROLL_WINDOW_BREAKPOINT = 992;
 // Margin to apply to the fixed sidebar top.
 const Y_SCROLL_MARGIN = 100;
 const placeTypesWithChoropleth = new Set(["Country", "State", "County"]);
+const CATEGORY_REDIRECT = {
+  Climate: "Environment",
```
beets, line 44: nit: can we do the redirect in flask? it'll be much faster (no need to download and execute js).

beets (approved):
thanks for these fixes!

can we also capture a todo for a webdriver test that tests that the explore more links for maps loads the right choropleth?

static/js/place/place.ts
chejennifer: done

beets: perfect - lgtm

PR #1558, opened 2022-04-15
Reconfigure private instances into separate modules

beets:
- Private instances can specify separate base.html, which can pull in separate css files
- Move all templates into a subfolder
- Remove some handlers that only belong in main dc - protein browser, special announcement.
- Fix a stat var widget width / height overflow issue in tools

Removing the FA specific css - will rewrite it / pull in only what we need

/cc @SenthamizhanV - I'll merge and send a PR to pull in your changes, but you might want to skip merging this PR for now..


server/__init__.py
```diff
@@ -44,30 +44,38 @@ def createMiddleWare(app, exporter):
...
-def register_routes_main_app(app):
-    # apply the blueprints for main and private app
-    from routes import (protein, browser, dev, factcheck, place, placelist,
-                        ranking, redirects, search, static, tools, topic_page)
+def register_routes_common(app):
+    # apply the blueprints for main app
+    from routes import (browser, dev, factcheck, place, placelist, ranking,
+                        search, static, tools, topic_page)
     app.register_blueprint(browser.bp)
     app.register_blueprint(dev.bp)
```
shifucun, line 67: should dev, placelist and topic_page go to base_dc?

server/configmodule.py
```diff
@@ -57,7 +58,8 @@ class PrivateConfig(Config):
 
 class FeedingAmericaConfig(PrivateConfig):
     NAME = "Feeding America"
-    FEEDING_AMERICA = True
+    ENV_NAME = 'FEEDING_AMERICA'
+    BASE_HTML_PATH = 'private_dc/feeding_america/base.html'
```
shifucun, line 62: This is nice! Note after merge, should change to "feedingamerica" throughout

shifucun (commented):
This is very nice! Glad to see this work in action.

Maybe use "feedingamerica" for all folder names, now has "feeding_america" and "fa"

server/configmodule.py
shifucun: With this change, I think we can remove "PrivateConfig" class above which sets "PRIVATE=True". There PRIVATE is really an environment name. Note here FeedingamericaConfig inherits that. Then this is used in static.py when checking the env

shifucun (commented):
I guess a good TODO is adding webdriver tests to make sure the home page loads correctly for each ENV.

server/configmodule.py
beets: +1 - updated

server/routes/static.py
```diff
@@ -14,21 +14,22 @@
...
 
 
 @bp.route('/')
 def homepage():
+    # TODO: Update this to use ENV_NAME
     if current_app.config.get('PRIVATE', None):
-        return render_template('static/private.html')
+        return render_template('private_dc/default/homepage.html')
+    if current_app.config.get('FEEDING_AMERICA', None):
+        return render_template('private_dc/feeding_america/homepage.html')
```
jehangiramjad: Same comment as an earlier one: we should use "feedingamerica" as one word now :)

server/__init__.py
beets: i wasn't sure.. we can move it.

beets:
> I guess a good TODO is adding webdriver tests to make sure the home page loads correctly for each ENV.

yes, +1. static tests too, which we had for sustainability. catches it faster.

shifucun (approved):

beets:
@SenthamizhanV you might want to separately pull in the changes in base_tools.scss and stat_var_widget.scss. it will fix some overflow bugs..

SenthamizhanV:
Sure, @beets. I will pull those changes. Thanks a lot!

PR #1561, opened 2022-04-15
[BQ] add bq button for map and timeline

chejennifer:
screen-recording-_23_
screen-recording-_24_


server/templates/tools/map.html
```diff
@@ -34,4 +34,8 @@
 {% block footer %}
   <script src={{url_for('static', filename='map.js', t=config['VERSION'])}}></script>
   <script src="https://maps.googleapis.com/maps/api/js?key={{maps_api_key}}&libraries=places" async defer></script>
+{% endblock %}
+
+{% block subfooter_extra %}
+<button id="bq-link" class="btn-subfooter btn btn-primary" style="display: none">Big Query</button>
```
beets, line 40: Nit: Should be BigQuery without a space.

Should we add an action word -- Open in BigQuery?

```suggestion
<button id="bq-link" class="btn-subfooter btn btn-primary" style="display: none">BigQuery</button>
```

server/templates/tools/timeline.html
```diff
@@ -37,6 +37,7 @@
 {% endblock %}
 
 {% block subfooter_extra %}
+<button id="bq-link" class="btn-subfooter btn btn-primary" style="display: none">Big Query</button>
```
beets, line 40: ditto

beets (approved):
this is nice! we should also add a todo to add a webdriver test for this. something to check that the queries will continue to work.

server/templates/tools/map.html
chejennifer: done

server/templates/tools/timeline.html
chejennifer: done

PR #1565, opened 2022-04-19
Integrate match

nopper:
Hi @shifucun, 

PTAL. I am still having some troubles with the API in the mixer. I got different results when testing the statvar match api using the mixer project and when I am using it here. Is this expected?


server/routes/api/stats.py
```diff
@@ -324,7 +325,8 @@ def search_statvar():
     query = request.args.get("query")
     places = request.args.getlist("places")
     sv_only = request.args.get("svOnly", False)
-    result = dc.search_statvar(query, places, sv_only)
+    # result = dc.search_statvar(query, places, sv_only)
```
shifucun, line 328: remove line 328?

shifucun (approved):

nopper:
Sorry I forgot to run the tests. I had to create another api to avoid breaking the tests.

PR #1581, opened 2022-04-21
Restore comment block in BQ query

pradh:
Guha suggests adding back the comment in the query block.


pradh:
Holding off pending more clarification...

PR #1583, opened 2022-04-21
[bigquery] small fixes to the bq modal style

beets:
I noticed an overflow at narrow sizes. Factored out some style from the chart export modal to share a consistent look.

Since textareas can't expand to fit content without js, I thought it would be simplest to make the entire modal super tall, since our queries do get really long.

localhost_8080_tools_timeline (4)
localhost_8080_tools_timeline (5)

### DIFF
--- a/static/css/place/place.scss
+++ b/static/css/place/place.scss
@@ -20,6 +20,7 @@
 @import "place_search";
 @import "../shared/story_block.scss";
 @import "../shared/story_chart.scss";
+@import "../shared/modal.scss";
 
 $vertical-section-border: 1px solid var(--dc-gray-lite);
 $vertical-section-margin: 2rem;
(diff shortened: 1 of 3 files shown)


beets:
i also have a pending change to make the text easier to read, wdyt (maybe this can wait till next week)

localhost_8080_tools_timeline (6)

chejennifer (approved):
these are nice changes, thanks!

chejennifer:
> i also have a pending change to make the text easier to read, wdyt (maybe this can wait till next week)
> 
> localhost_8080_tools_timeline (6)

I think making it easier to read is not a risky change and it looks cleaner, so it should be fine to make that change?

In that change, since you're making the instructions in a list, maybe you can remove the ", then" in step 1

beets:
thanks!

beets:
lmk if we should update the header text to be a list

pradh:
For future, I slightly wonder if we should reduce font size of the query?

That would reduce the wrapping of the query and more of then than not people won't carefully read the query in our window, rather copy/paste to BQ.

image

Perhaps the middle size?

beets:
how about .8rem, which works out to 12px for most ppl:

image

pradh:
LGTM!

PR #1584, opened 2022-04-21
Adds plots for rich search interface

eisenjulian:
This PR leverages the timeline tool charts components to group and render a generic list of StatVars as returned by a search API.

8mD29uCa6kJZryw

In the process I found and fixed two issues with the existing timeline components:
1. They broke if some of the data download buttons were missing despite some existing logic trying to account for it
2. For StatVars like incremental_count they rendered empty plots.


static/js/rich_search/app.tsx
```diff
@@ -33,35 +33,69 @@ import {
...
   addPlace: (string) => void;
   removePlace: (string) => void;
   loading: boolean;
-  onClickSearch: () => void;
-  inputInvalid: boolean;
+  onSearch: (string) => void;
   chartsData: StatVarsPropType;
 }
 
+interface TextSearchBarPropType {
```
beets, line 48: Please add each new component into a separate file for easier management, and some documentation.

static/js/rich_search/app.tsx
```diff
@@ -33,35 +33,69 @@ import {
   setTokensToUrl,
 } from "../tools/timeline/util";
 import { getPlaceNames } from "../utils/place_utils";
-import { MemoStatVars, StatVarsPropType } from "./stat_vars";
+import { StatVars, StatVarsPropType } from "./stat_vars";
```
beets, line 36: please add a top-level comment for the file / module.

static/js/rich_search/app.tsx
```diff
@@ -131,14 +151,15 @@ function App({
 async function getStatVars(
```
beets, line 151: please add documentation for new functions, especially those which are exported.

static/js/rich_search/app.tsx
```diff
@@ -131,14 +151,15 @@ function App({
...
-): Promise<NamedNode[]> {
+): Promise<[string[], { [key: string]: StatVarInfo }]> {
   const params = new URLSearchParams({ query });
   for (const place of dcids) {
     params.append("place", place);
   }
   return axios
     .get(`/api/stats/stat-var-search-ai?${params.toString()}`)
-    .then((resp) => resp.data.statVars || []);
+    .then((resp) => (resp.data.statVars || []).map((v) => v.dcid).slice(0, 6))
```
beets, line 161: are dcid's guaranteed in the response?

static/js/rich_search/stat_vars.tsx
```diff
@@ -31,21 +36,31 @@ function getURL(places: string[], statsVar: string): string {
...
+          <li key={sv}>
+            <a href={getURL(places, sv)}>{statVarInfo[sv].title}</a>
           </li>
         ))}
       </ul>
     </section>
   );
 }
-
-export const MemoStatVars = memo(StatVars);
```
beets, line 51: i'm curious why you've chosen lodash's memo over react's? were there performance differences observed?

i also see that there is now no memoization for stat vars, but only for place names. could you comment here, or in the pr description, on some of these internal changes?

static/js/rich_search/stat_vars.tsx
```diff
@@ -31,21 +36,31 @@ function getURL(places: string[], statsVar: string): string {
   return `/tools/timeline#${params}`;
 }
 
-function StatVars({ places, statVars }: StatVarsPropType): JSX.Element {
-  if (!statVars.length) {
+export function StatVars({
```
beets, line 39: please add documentation

beets, line 39: is this more appropriately named "StatVarResults"?

static/js/tools/timeline/chart.tsx
```diff
@@ -135,8 +135,15 @@ class Chart extends Component<ChartPropsType, ChartStateType> {
     const placeName = Object.values(this.props.placeNames)[0];
     const deltaCheckboxId = `delta-cb-${this.props.mprop}`;
     const sourceSelectorSvInfoList = this.getSourceSelectorSvInfo(statVars);
+    // Hides empty plots
```
beets, line 138: we'd prefer to display an error message in this case, rather than hiding the entire chart.

static/js/tools/timeline/chart_region.tsx
```diff
@@ -155,12 +155,13 @@ class ChartRegion extends Component<ChartRegionPropsType> {
 
   private onDataUpdate(groupId: string, data: StatData) {
     this.allStatData[groupId] = data;
-    if (this.downloadLink && Object.keys(this.allStatData).length > 0) {
-      this.downloadLink.style.display = "inline-block";
-      this.bulkDownloadLink.style.display = "inline-block";
-    } else {
-      this.downloadLink.style.display = "none";
-      this.bulkDownloadLink.style.display = "none";
+    const displayStyle =
```
beets, line 158: thanks for this fix!

static/js/tools/timeline/chart.tsx
eisenjulian: How should the error be rendered, do you have pointers? Is it actually an error, or is it expected that some StatVars do not have data? An example of this is selecting Increment Population in the timeline tool. Perhaps we can move to separate PR

static/js/rich_search/app.tsx
eisenjulian: They are, but I added a filter just in case

static/js/rich_search/stat_vars.tsx
eisenjulian: Changed.

static/js/rich_search/stat_vars.tsx
eisenjulian: My understanding is that React's memo is for components and controls when the rendering will happen. I added it originally since the state of the search box was always in sync with the hash url, which re-forced renders of this component unnecessarily and produced a non responsive experience. After rethinking the design, I changed the logic so that the state of the search box was only updated when the search button or the enter key were pressed, which removed the need for the memo call.

Independently, I added _.memoize (which is similar to React.useMemo but React independent) as a way to cache the network results for a backed query, since both components (StatVars and PlaceSearch) needed the output of this promise. Refactoring the logic would have made it much more complicated, and PlaceSearch will probably be removed in a future version due to the two search boxes being unified.

Let me know if that makes sense.

eisenjulian: Actually since the search API was updated to extract locations in a previous PR I went ahead and removed the place search bar which simplified the logic overall

beets: thanks for the explanation - that makes sense

static/js/tools/timeline/chart.tsx
beets: it should be an error, at least in the timeline tool, as users should not be able to select variables which we don't have data for.

a separate PR would be great. i was thinking of doing something similar in this case too, when it came up in another context.

eisenjulian: I undid the changes in this file.

beets (approved):
thanks for the updates

PR #1587, opened 2022-04-22
Google Cloud NLP integration

nopper:
This PR Integrates Google Cloud NLP service for identifying place mentions in the search query.

This is the first step towards having a single search bar instead of two one for the question and one for the place selection.

A big limitation is that currently the id associated to the place is a Freebase MID. We will therefore need a mechanism to map back to dcids at some points. Do you have any suggestions @shifucun ?

### DIFF
--- a/server/services/ai.py
+++ b/server/services/ai.py
@@ -173,7 +173,8 @@ def _iterate_property_value(
             yield (key, value)
 
 
-def get_places(query: str) -> Sequence[language_v1.Entity]:
+def _get_places(query: str) -> Sequence[language_v1.Entity]:
+    """Returns a list of entities that are of type LOCATION."""
     global _LANGUAGE_CLIENT
     if not _LANGUAGE_CLIENT:
         return []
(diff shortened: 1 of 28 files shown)


shifucun (approved):

shifucun:
Please fix the lint error with `./run_test.sh -f`

server/services/ai.py
```diff
@@ -171,20 +173,60 @@ def _iterate_property_value(
             yield (key, value)
 
 
+def get_places(query: str) -> Sequence[language_v1.Entity]:
```
beets, line 176: can we start adding documentation for newly added functions?

server/services/ai.py
```diff
@@ -31,6 +32,7 @@
 # ----------------------------- INTERNAL FUNCTIONS -----------------------------
 
 _INFERENCE_CLIENT = None
+_LANGUAGE_CLIENT = language_v1.LanguageServiceClient()
```
beets, line 35: i believe this isn't thread safe, are you seeing a lot of overhead to this call? this should go in the constructor and stay within the class.

nopper: I think imports are thread safe so variable module initialization should be thread safe. Which constructor are you suggesting?

server/services/ai.py
nopper: Done.

server/services/ai.py
nopper: Let me know if this is a thing we should fix. In that case I'll do it in a later PR.

beets: yes please, could you move it into the InferenceClient constructor?  since there is a way to avoid the global construction

https://stackoverflow.com/questions/35309042/python-how-to-set-global-variables-in-flask
https://stackoverflow.com/questions/32815451/are-global-variables-thread-safe-in-flask-how-do-i-share-data-between-requests

PR #1590, opened 2022-04-27
Use ai.Context for clients instead of global vars.

nopper:
Hi @beets!

Following up with PR that should address your concerns around thread safety. Let me know if this is the right approach.


beets (commented):
thanks - yes this is a better approach. could you add documentation to the Context class? Thanks!

nopper:
Gentle ping.

Issue #1582, opened 2022-04-21
[BUG] Some timeline BQ queries can be expensive


pradh:
This processes 2.25TB.
```
SELECT
  O.observation_about AS PlaceId,
  O.variable_measured AS VariableId,
  O.observation_date AS Date,
  CAST(O.value AS FLOAT64) AS Value,
  O.measurement_method AS MeasurementMethod,
  O.unit AS Unit
FROM `data_commons.Observation` AS O
WHERE
  ((O.variable_measured = 'WithdrawalRate_Water_Thermoelectric'
      AND O.measurement_method IS NULL
      AND O.unit = 'MillionGallonsPerDay'
      AND O.observation_period = 'P1Y'
      AND O.scaling_factor IS NULL)
    OR (O.variable_measured = 'WithdrawalRate_Water_PublicSupply'
      AND O.facet_rank = 1))
  AND O.observation_about = 'geoId/06'
ORDER BY
  PlaceId,
  VariableId,
  Date
```

But we separate out the `OR` condition into a UNION ALL of two queries, this only processes 500MB.

```
SELECT
  O.observation_about AS PlaceId,
  O.variable_measured AS VariableId,
  O.observation_date AS Date,
  CAST(O.value AS FLOAT64) AS Value,
  O.measurement_method AS MeasurementMethod,
  O.unit AS Unit
FROM `data_commons.Observation` AS O
WHERE
  (O.variable_measured = 'WithdrawalRate_Water_Thermoelectric'
   AND O.measurement_method IS NULL
   AND O.unit = 'MillionGallonsPerDay'
   AND O.observation_period = 'P1Y'
   AND O.scaling_factor IS NULL)
  AND O.observation_about = 'geoId/06'

UNION ALL

SELECT
  O.observation_about AS PlaceId,
  O.variable_measured AS VariableId,
  O.observation_date AS Date,
  CAST(O.value AS FLOAT64) AS Value,
  O.measurement_method AS MeasurementMethod,
  O.unit AS Unit
FROM `data_commons.Observation` AS O
WHERE
  (O.variable_measured = 'WithdrawalRate_Water_PublicSupply'
   AND O.facet_rank = 1)
  AND O.observation_about = 'geoId/06'

ORDER BY
  PlaceId,
  VariableId,
  Date
```

PR #1591, opened 2022-04-27
[bq] various updates

chejennifer:
- updated the query for timeline to be less expensive (fixes [issue #1582](https://github.com/datacommonsorg/website/issues/1582))
- updated the queries for timeline and map to check import name when source is selected
- fixed query alignment/formatting
- extracted some redundant query code
- added unit tests


static/js/tools/map/bq_query_utils.test.ts
```diff
@@ -0,0 +1,485 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { StatMetadata } from "../../shared/stat_types";
+import { getNonPcQuery, getPcQuery } from "./bq_query_utils";
+import { PlaceInfo, StatVar } from "./context";
+
+const sv: StatVar = {
```
beets, line 21: since these are constants, you should uppercase the var names, e.g. TEST_SV

static/js/tools/map/bq_query_utils.ts
```diff
@@ -70,17 +71,21 @@ function getLatestObsDateView(
...
+           O.observation_about as PlaceId,
+           MAX(O.Observation_date) as LatestDate
+    FROM \`data_commons.Observation\` as O
+    JOIN ${CHILD_PLACE_VIEW_NAME} ON TRUE
+    JOIN \`data_commons.Provenance\` as I ON TRUE
+    WHERE O.prov_id = I.prov_id AND
+          O.variable_measured = '${sv.dcid}' AND
+          O.observation_about = ${CHILD_PLACE_VIEW_NAME}.PlaceId AND
+          ${getSvMetadataPredicates("O", "I", metadata).join(
+            " AND\n          "
```
beets, line 86: i'm curious why there's so many trailing spaces after the newline. if it's to maintain the indentation, could you leave a comment?

static/js/tools/timeline/bq_query_utils.test.ts
```diff
@@ -0,0 +1,415 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { DEFAULT_POPULATION_DCID } from "../../shared/constants";
+import { getTimelineSqlQuery } from "./bq_query_utils";
+import { ChartGroupInfo } from "./chart_region";
+
+const metahashMap = {
```
beets, line 21: same comment as above, re: constant var naming.

beets (approved):
these test are nicely laid out, very much like the mixer tests

static/js/tools/map/bq_query_utils.test.ts
chejennifer: done

static/js/tools/map/bq_query_utils.ts
chejennifer: done

static/js/tools/map/bq_query_utils.test.ts
```diff
@@ -0,0 +1,485 @@
...
+JOIN PlaceObsDatesAndDenomRank AS PODDR ON TRUE
+WHERE ONum.variable_measured = 'Count_Person_Female' AND
+      ONum.observation_about = PODDR.PlaceId AND
+      ONum.observation_about = P.id AND
+      ONum.prov_id = I.id AND
+      ONum.measurement_method = 'testMMethod' AND
+      ONum.unit IS NULL AND
+      ONum.observation_period = 'P1Y' AND
+      ONum.scaling_factor IS NULL AND
+      I.name IS NULL AND
```
pradh, line 238: I don't think we should do this.  If importNAme is not provided, we should not include it in the join.

pradh (commented):
Sorry for the late review.  Flushing out first set of comments I had.

PR #1593, opened 2022-04-28
Fix some key errors

shifucun:
These are found from the GCP error reporting page


chejennifer (approved):
thanks for the fixes!

beets (approved):
thanks for the fixes!

PR #1596, opened 2022-04-29
India DC fixes

SenthamizhanV:
- Adopted the new file structure for private DC templates
- Removed individual CSS files and used only what's already in the bundle
- Added integrity hashes for one newly added library
- Added FAQ and Feedback pages for IITM env
- Fixed minor bugs after CSS sync


server/templates/private_dc/iitm/base.html
```diff
@@ -58,11 +56,12 @@
 
 <body id="iitm-india">
   <div id="main">
+    
+    {% if is_homepage %}
+    <nav class="navbar navbar-light navbar-expand-lg fixed-top " id="mainNav" style="padding: 0px 0px;">
+      <div class="container-fluid mr-5" style="padding: 0px;">
+    {% else %}
     <header id="main-header">
```
beets, line 64: I couldn't figure this out  -- why is the header element dropped from the html for the homepage?

beets (approved):
thanks for the updates!!

server/templates/private_dc/iitm/base.html
SenthamizhanV: Homepage uses a different navbar design than other pages. We don't use the solid color bar for navbar in homepage. It's just the gradient and logos. So, we added the header element directly in homepage.html

Normal navbar with solid purple bar
image

Homepage navbar with gradient and logos
image

beets: ok got it. i mistook it for the header element which is now used for the hero image. thanks!

PR #1601, opened 2022-05-02
[map] fix non float string error in getting map points data

chejennifer:
- when getting map points, sometimes error thrown because longitude/latitude string was not a float. Should skip those map points


server/tests/api_shared_test.py
```diff
@@ -47,3 +47,22 @@ def test_cached_name(self, mock_data_fetcher):
...
+        cases = [{
+            'query': 'abc',
+            'expected': False
+        }, {
+            'query': '1.26',
+            'expected': True
+        }, {
+            'query': '-',
+            'expected': False
+        }]
```
beets, line 64: Also add a test for 0.0 and a non-float, but numeric value?

beets (approved):
nice

server/tests/api_shared_test.py
chejennifer: done

PR #1605, opened 2022-05-04
Handle the case no places returned by the API

eisenjulian:
This change properly handles and shows errors when no places were specified in the query. We also add support for displaying generic debug messages.


beets (approved):
this is nice, thanks

PR #1607, opened 2022-05-05
Dates in chart titles made more readable

SenthamizhanV:
The YYYY-MM format in chart titles can be potentially misinterpreted in some cases.

For example, the below screenshot shows the [Max. Temp.data for December 2012 in India](https://datacommons.org/tools/map#%26sv%3DDifferenceRelativeToBaseDate2006_Max_Temperature_RCP45%26pc%3D0%26denom%3DCount_Person%26dt%3D2011-12%26pd%3Dcountry%2FIND%26ept%3DAdministrativeArea2) but it is easy to misconstrue the date as 2011-2012:
image

Hence, the month format is changed from MM to MMM format. 

Example screenshot of the title with new format:
image (1)

Also, this change is only implemented when the returned date is in 'YYYY-MM'. It will not work if the mixer returns 'YYYY-MM-DD'.


static/js/utils/string_utils.ts
```diff
@@ -16,12 +16,30 @@
 
 import _ from "lodash";
 
+/**
+ * If the date is in YYYY-MM format, returns YYYY-MMM format.
+ * This is to improve readability of dates in chart titles.
+ */
+export function formatDate(strDate: string): string {
+  const regexDate = /^\d{4}-\d{1,2}$/;
```
beets, line 24: since our dates are always ISO 8601, you could do a simpler test for length.

static/js/utils/string_utils.ts
```diff
@@ -16,12 +16,30 @@
...
+  if (!regexDate.test(strDate)) {
+    return strDate;
+  } else {
+    const dt = new Date(strDate);
+    return (
+      new Intl.DateTimeFormat("en-US", { year: "numeric" }).format(dt) +
+      "-" +
+      new Intl.DateTimeFormat("en-US", { month: "short" }).format(dt)
```
beets, line 32: I think this format also looks unnatural. How about using the formatter to generate the entire final string:

```
new Intl.DateTimeFormat("en-US", { year: "numeric", month: "short"}).format(dt)
```

beets (commented):
sorry for the late review. please also add a test for the utils, thanks!

static/js/utils/string_utils.ts
```diff
@@ -16,12 +16,28 @@
 
 import _ from "lodash";
 
+
+/**
+ * If the date is in YYYY-MM format, returns YYYY-MMM format.
+ * This is to improve readability of dates in chart titles.
+ */
+export function formatDate(strDate: string): string {
+  var regex_date = /^\d{4}\-\d{1,2}$/;
```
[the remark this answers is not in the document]
SenthamizhanV: Replaced regex with length test.

static/js/utils/string_utils.ts
SenthamizhanV: We tried your suggestion of using the formatter once. But the resulting date appears in the format "MMM YYYY" (with a space in between). We couldn't find any options to force our format of YYYY-MMM.

Also, added a test for this function in utils.

beets: Yes, the dash there looks unusual as well, when paired with the spelled out date. We can try this for now and adjust as we go.

beets (approved):
thanks for adding the test!

PR #1609, opened 2022-05-05
Updates to new statvar-match API.

nopper:
Updates the new statvar-match API introduced in:
https://github.com/datacommonsorg/mixer/pull/842

### DIFF
--- a/server/services/ai.py
+++ b/server/services/ai.py
@@ -213,23 +213,40 @@ def _delexicalize_query(query: str,
     output.append(query_bytes[prev:].decode('utf8'))
     return ''.join(output)
 
+_KEY_NORMALIZATION = {
+    "measuredProperty": "mp",
+    "populationType": "pt",
+    "statType": "st",
+    # Is this all?
+}
+
+def _normalize_keys(key_values: Iterator[Tuple[str, str]]) -> Iterator[Tuple[str, str]]:
+    for key, value in key_values:
+        yield (_KEY_NORMALIZATION.get(key, key), value)
+
 
 def search(context: Context, query: str) -> Sequence[Mapping[str, str]]:
     if not context.inference_client:
         return {"statVars": [], "places": []}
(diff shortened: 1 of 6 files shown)


server/services/ai.py
```diff
@@ -213,23 +213,40 @@ def _delexicalize_query(query: str,
     output.append(query_bytes[prev:].decode('utf8'))
     return ''.join(output)
 
+_KEY_NORMALIZATION = {
+    "measuredProperty": "mp",
+    "populationType": "pt",
+    "statType": "st",
+    # Is this all?
```
nopper, line 220: @shifucun is this the full list of transformation that you apply in the mixer?

nopper:
PTAL.

server/services/ai.py
```diff
@@ -214,22 +214,66 @@ def _delexicalize_query(query: str,
     return ''.join(output)
 
 
+_KEY_NORMALIZATION = {
```
shifucun, line 217: Can add a comment that this matches the flume cache builder

server/services/ai.py
```diff
@@ -214,22 +214,66 @@ def _delexicalize_query(query: str,
...
+
+
+def _build_query(
+        query: str,
+        key_values: Iterator[Tuple[str, str]]) -> Iterator[Tuple[str, str]]:
+    terms = []
+    for token in query.split():
+        terms.append(f'sn:"{token}"')
+    for key, value in key_values:
+        key = _KEY_NORMALIZATION.get(key, key)
```
shifucun, line 233: should the default "key" be ""? rather than the key?

shifucun, line 233: better have a new variable name here?

nopper: No this is correct. We normalize only if we have a normalization entry, otherwise the key is considered valid, as it can be a constrained property key.

shifucun (approved):

PR #1616, opened 2022-05-11
Updates model.

nopper:
- Fixes request deadlock.
- Increase wait time for mixer to 5 minutes.
- Set startupProbe for mixer.

### DIFF
--- a/deploy/base/deployment.yaml
+++ b/deploy/base/deployment.yaml
@@ -57,6 +57,12 @@ spec:
           args: []
           ports:
             - containerPort: 8080
+          startupProbe:
+            httpGet:
+              path: /healthz
+              port: 8080
+            failureThreshold: 30
+            periodSeconds: 10
           readinessProbe:
             httpGet:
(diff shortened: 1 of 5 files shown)


deploy/base/deployment.yaml
```diff
@@ -146,6 +146,11 @@ spec:
                   key: mixer_hash.txt
           ports:
             - containerPort: 12345
+          startupProbe:
+            exec:
+              command: ["/bin/grpc_health_probe", "-addr=:12345"]
+            failureThreshold: 30
```
shifucun, line 152: Can you try to deploy this to dev. I tried this 2 years back but was not supported in old GKE cluster. I assume it should work now (cluster in 1.21) but should confirm.

shifucun (approved):

deploy/base/deployment.yaml
shifucun: Also can you add startupProbe to the website container below as well?

PR #1617, opened 2022-05-13
Only allow places with MIDs from NLP API

eisenjulian:
The NLP API that we use in the search UI can return non-places like the word houses or cities, we can filter those examples by using the MID attribute.


shifucun (approved):

shifucun:
Thanks!

PR #1620, opened 2022-05-17
Add choropleth map to the earth page

raebo1017:
For the visualization part, it turns out that only the chart fig that has complete data can be rendered. So I was only able to render the choropleth for the population under demographics. And there are many sources listed under the chart as well.

image

### DIFF
--- a/server/routes/api/choropleth.py
+++ b/server/routes/api/choropleth.py
@@ -72,6 +72,8 @@ def get_choropleth_display_level(geoDcid):
     # Territories of the US, e.g. country/USA are also AA1. Restrict this view to only States.
     if geoDcid == 'country/USA':
         return geoDcid, 'State'
+    if geoDcid == 'Earth':
+        return geoDcid, 'Country'
     place_type = place_api.get_place_type(geoDcid)
     display_level = None
     if place_type in CHOROPLETH_DISPLAY_LEVEL_MAP:
--- a/server/routes/api/choropleth.py
+++ b/server/routes/api/choropleth.py
@@ -41,9 +41,9 @@ CHOROPLETH_DISPLAY_LEVEL_MAP = {
 }
 
 # Place type to get choropleth for, keyed by geoDcid
-# This works for special geoDcid.
-# 1. country/USA are also AA1. Restrict this view to only States.
-# 2. display the chorpleth for earth in the country level.
+# These are the special cases.
+# 1. Territories of the US, e.g. country/USA, are also AA1. Restrict this view to only States.
+# 2. For Earth, display the countries.
 SPECIAL_CHOROPLETH_DISPLAY_LEVEL_MAP = {
     "country/USA": "State",
     "Earth": "Country"
--- a/server/routes/api/choropleth.py
+++ b/server/routes/api/choropleth.py
@@ -39,6 +39,16 @@ CHOROPLETH_DISPLAY_LEVEL_MAP = {
     "AdministrativeArea2": "AdministrativeArea2",
     "AdministrativeArea3": "AdministrativeArea3"
 }
+
+# Place type to get choropleth for, keyed by geoDcid
+# This works for special geoDcid.
+# 1. country/USA are also AA1. Restrict this view to only States.
+# 2. display the chorpleth for earth in the country level.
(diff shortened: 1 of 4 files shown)


raebo1017:
@chejennifer  @beets

server/routes/api/choropleth.py
```diff
@@ -72,6 +72,8 @@ def get_choropleth_display_level(geoDcid):
     # Territories of the US, e.g. country/USA are also AA1. Restrict this view to only States.
     if geoDcid == 'country/USA':
         return geoDcid, 'State'
+    if geoDcid == 'Earth':
+        return geoDcid, 'Country'
```
chejennifer, line 76: as more special handling of geoDcids gets added, I think it makes sense to create a map for these special geoDcids. The map would be similar to CHOROPLETH_DISPLAY_LEVEL_MAP but have key be geoDcid and value be display_level

static/js/place/chart_block.tsx
```diff
@@ -195,6 +195,7 @@ class ChartBlock extends React.Component<ChartBlockPropType> {
     // Prepare parameters for related charts.
     let unit = this.props.data.unit;
     let scaling = this.props.data.scaling;
+    const isEarth = this.props.dcid === "Earth";
```
chejennifer, line 198: we actually have a constant for Earth here: https://github.com/datacommonsorg/website/blob/5e84f21c7378880b02d3a1e1e75b869ca263957f/static/js/shared/constants.ts#L35 

Instead of using the string "Earth", it will be better to use the dcid field of this constant. This applies to all other instances in this file where "Earth" is used.

static/js/place/chart_block.tsx
```diff
@@ -254,10 +255,11 @@ class ChartBlock extends React.Component<ChartBlockPropType> {
       let gotChart = false;
       if (
         !!this.props.data.isChoropleth &&
-        this.props.isUsaPlace &&
-        (this.props.placeType === "Country" ||
-          this.props.placeType === "State" ||
-          this.props.placeType === "County")
+        (isEarth ||
+          (this.props.isUsaPlace &&
+            (this.props.placeType === "Country" ||
```
chejennifer, line 260: minor cleanup: I see that place.ts has a constant `placeTypesWithChoropleth` which is a set of all the possible place types to render a choropleth for.  Let's move that constant to the util.ts file in this same folder, rename it to USA_PLACE_TYPES_WITH_CHOROPLETH, export it and also use it here

chejennifer (commented):
Great job! 

Regarding only chart configs with complete data being rendered, in the chart.tsx file, there is a constant (https://github.com/datacommonsorg/website/blob/d698c38c9cec6fa98cf74531928f4cb36cf49fe1/static/js/place/chart.tsx#L55) that controls how many data points needs to be returned in order for a chart to be rendered. Can you double check that the charts not rendered are because they have less datapoints than that?

Regarding many sources showing up, that makes sense because the source used for the different countries can be different. However, from looking at the screenshot, we should fix the placement of the export and explore more buttons when the source list is long like that. Can you create a github issue?

server/routes/api/choropleth.py
raebo1017: Done

static/js/place/chart_block.tsx
raebo1017: Done

static/js/place/chart_block.tsx
raebo1017: Done

chejennifer: I see that you created the `USA_PLACE_TYPES_WITH_CHOROPLETH`, but didn't use it here?

raebo1017:
> 
1. Checked the number of datapoints returned for each chart and the logic for rendering the choropleth. It is expected that only the population chart should show up.
2. Created an issue for the placement of explore and explore more buttons.

static/js/place/chart_block.tsx
raebo1017: Pushed a new commit and modified the code

chejennifer (approved):
Thanks for the updates!

Issue #1522, opened 2022-04-07
[BUG] Map tool should either not list non-existent type, or allow deselecting non-existent SV


pradh:
1. Pick a StatVar that exists in county-level

![image](https://user-images.githubusercontent.com/4375037/162261091-a492d839-47e6-4d12-826a-b9205ddb900c.png)

2. The SV doesn't exist at IPCCPlace level.  If I select IPCCPlace, then it gets into a state where there is no way to deselect the SV (without perhaps using browser back button)

![image](https://user-images.githubusercontent.com/4375037/162261278-fc8cb0d2-5d18-4364-b5d5-0ef49155dfda.png)

If possible/simple, we should not list a type with no data for the chosen SV.  Or, we should offer the user a way to deselect the SV...

PR #1625, opened 2022-05-24
[download] Add stat var selection alerts

chejennifer:
- Add alert when place changes and chosen stat vars are unavailable
    - Refactored stat var widget to do this so that all tools are consistent in behavior
    - This also fixes: [issue #1522 ](https://github.com/datacommonsorg/website/issues/1522)
    
screen-recording-_42_

- Add modal that pops up when more than 5 stat vars are chosen for download tool

screen-recording-_41_


beets (approved):
this is great - thanks for the cleanup!

PR #1627, opened 2022-05-25
Update npm test cloudbuild yaml

chejennifer:
- update the yaml file for npm tests on cloudbuild to copy over node_modules from the node docker image before building and running the tests. This should decrease the time needed for building the static files.
- also update docker image to 2022-04-14 version


shifucun:
Please verify the effect after this is checked in.

beets (approved):

PR #1630, opened 2022-06-01
Increase mixer memory to accomadate for more data

shifucun:
Seeing OOM during the deployment. This is likely due to the increased data from stat var (group). Before a scalable fix, increase the memory to let deployment work smoothly.


n-h-diaz (approved):
thank you!

PR #1631, opened 2022-06-01
Implement api and write test for the ranking chart

raebo1017:
Create a new flask endpoint for the ranking chart. By adding a flag called isRankingChart in chart configs (will do in later PRs), the endpoint will be able to take a place dcid and read chart configs to figure out what stat vars to get data for. The new endpoint calls [/api/stats/within-place](https://github.com/datacommonsorg/website/blob/master/server/routes/api/stats.py) to get data and calculate the rankings.

For now, we only support the simpler single SV, no denominator case first.

Prototype design doc:
https://docs.google.com/document/d/1FSojZpkbOgtm9sraHrhKevsfRMDTgMvyU3dm04kYeeM/edit?usp=sharing&resourcekey=0-rJedtMgcx0RlD2GuGaYXZA


server/routes/api/place.py
```diff
@@ -763,3 +766,124 @@ def placeid2dcid():
...
+    for config in configs:
+        sv = config["statsVars"][0]
+        sv_data_stat = sv_data_values.get(sv, {}).get("stat", {})
+        if not sv_data_stat:
+            continue
+        sources = set()
+        dates = set()
+        data_points = []
+        for place_dcid in sv_data_stat:
+            # dcid_data is a dictionary including date, value and metahash.
```
chejennifer, line 830: would be nice to add an example to be able to see the format of dcid_data

server/routes/api/place.py
```diff
@@ -763,3 +766,124 @@ def placeid2dcid():
...
+            dcid_data = sv_data_stat[place_dcid]
+            dcid_value = dcid_data.get("value", None)
+            dcid_place_name = get_name([place_dcid]).get(place_dcid, "")
```
chejennifer, line 833: instead of making the get_name call for every place for every config, it'll be better to make the call once for all the places you will need names for, save the result in a variable, and just read that variable. 

This can be done before looping through the configs and the list of places that need names can be found by looking through sv_data_values

server/routes/api/place.py
```diff
@@ -763,3 +766,124 @@ def placeid2dcid():
...
+            # Value and place name are required for the calculation of ranking and the display.
+            if dcid_value is None or not dcid_place_name:
```
chejennifer, line 835: I don't think dcid_place_name is necessary for showing the ranking, if there's no name for a place, could just show the dcid

server/routes/api/place.py
```diff
@@ -763,3 +766,124 @@ def placeid2dcid():
...
+        sv = config["statsVars"][0]
+        sv_data_stat = sv_data_values.get(sv, {}).get("stat", {})
+        if not sv_data_stat:
+            continue
+        sources = set()
+        dates = set()
+        data_points = []
+        for place_dcid in sv_data_stat:
+            # dcid_data is a dictionary including date, value and metahash.
+            dcid_data = sv_data_stat[place_dcid]
```
chejennifer, line 831: nit: dcid_data -> data

& same for the next two variables, can remove the dcid prefix

server/routes/api/place.py
```diff
@@ -763,3 +766,124 @@ def placeid2dcid():
...
+            dcid_value = dcid_data.get("value", None)
+            dcid_place_name = get_name([place_dcid]).get(place_dcid, "")
+            # Value and place name are required for the calculation of ranking and the display.
+            if dcid_value is None or not dcid_place_name:
+                continue
+            data_point = dict()
```
chejennifer, line 837: could just do 
```
data_point = {
"placeDcid": place_dcid,
"value": dcid_value,
"placeName": dcid_place_name
}
```

chejennifer (commented):
Nice job adding an endpoint! 

& Awesome that you found functions that can be re-used and moved those to the shared file!

server/routes/api/place.py
```diff
@@ -763,3 +766,124 @@ def placeid2dcid():
...
+            "numDataPoints":1,
+            "exploreUrl":"/ranking/UnemploymentRate_Person_Rural/State/country/USA?h=geoId/06&unit=%",
+            "sources":[
+                "https://www.bls.gov/lau/"
+            ]
+        }
+    }
+    """
+    result = {}
+    isEarth = dcid == "Earth"
```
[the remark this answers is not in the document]

server/routes/api/place.py
```diff
@@ -763,3 +766,124 @@ def placeid2dcid():
...
+        parent_place_dcid = "Earth"
+        place_type = "Country"
+    configs = get_ranking_chart_configs()
+    # Get the first stat var of each config.
+    stat_vars, _ = shared_api.get_stat_vars(configs)
+    sv_data = dc.get_stat_set_within_place(parent_place_dcid, place_type,
+                                           list(stat_vars), "")
+    sv_data_values = sv_data.get("data", {})
+    if not sv_data or not sv_data_values:
+        flask.abort(500)
```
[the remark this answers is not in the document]

server/routes/api/place.py

server/routes/api/place.py
raebo1017: Good point! Done

server/routes/api/place.py

server/routes/api/place.py

server/routes/api/place.py

server/routes/api/place.py
```diff
@@ -763,3 +766,124 @@ def placeid2dcid():
...
+            data_points.append(data_point)
+            dates.add(dcid_data.get("date", ""))
+            metadata_hash = dcid_data.get("metaHash", "")
+            sources.add(
+                sv_metadata.get(str(metadata_hash),
+                                {}).get("provenanceUrl", ""))
+        # Build URL for "explore more".
+        scaling = config.get("scaling", None)
+        unit = config.get("unit", None)
+        parent_place_dcid = None if isEarth else parent_place_dcid
```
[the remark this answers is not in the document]

server/routes/api/place.py
```diff
@@ -763,3 +766,134 @@ def placeid2dcid():
...
+            "exploreUrl":"/ranking/UnemploymentRate_Person_Rural/State/country/USA?h=geoId/06&unit=%",
+            "sources":[
+                "https://www.bls.gov/lau/"
+            ]
+        }
+    }
+    """
+    result = {}
+    # Get the parent place.
+    if not dcid == EARTH_DCID:
```
chejennifer, line 799: nit: can we switch the order of the if statement and just do `if dcid == EARTH_DCID: ... else: ...`

server/routes/api/place.py
```diff
@@ -763,3 +766,134 @@ def placeid2dcid():
...
+    stat_vars, _ = shared_api.get_stat_vars(configs)
+    sv_data = dc.get_stat_set_within_place(parent_place_dcid, place_type,
+                                           list(stat_vars), "")
+    sv_data_values = sv_data.get("data", {})
+    if not sv_data or not sv_data_values:
+        return Response(json.dumps(result), 200, mimetype='application/json')
+    ## Get all the place names of dcids in the sv_data
+    place_dcids = set()
+    for value in sv_data_values.values():
+        for place_dcid in value["stat"]:
```
chejennifer, line 823: nit: instead of iterating through all the place_dcids, we can do a union on the keys like: https://github.com/datacommonsorg/website/blob/master/server/routes/api/stats.py#L456-L461

chejennifer (commented):
Thanks for updating! just some small nits

chejennifer (approved):
thanks for the changes!

PR #1634, opened 2022-06-06
feat: add blurbs and fix error chart data retrieval

suhana13:
Fixed the data retrieval for error bar chart and added blurbs for each graph 
Please review @chejennifer !

### DIFF
--- a/static/js/protein/data_processing_utils.ts
+++ b/static/js/protein/data_processing_utils.ts
@@ -263,23 +263,27 @@ export function getVarGeneAssoc(data: GraphNodes): ProteinVarType[] {
               continue;
             }
             if (n2.property === "log2AllelicFoldChange") {
-              if (n2.nodes[0].value !== null) {
-                score = n2.nodes[0].value;
+              if (_.isEmpty(n2.nodes[0].value)) {
+                continue;
               }
+              score = n2.nodes[0].value;
             } else if (n2.property === "referenceSNPClusterID") {
-              if (n2.nodes[0].value !== null) {
-                variant = n2.nodes[0].value;
+              if (_.isEmpty(n2.nodes[0].value)) {
+                continue;
               }
+              variant = n2.nodes[0].value;
             } else if (
               n2.property === "log2AllelicFoldChangeConfidenceInterval"
             ) {
-              if (n2.nodes[0].value !== null) {
-                interval = n2.nodes[0].value;
+              if (_.isEmpty(n2.nodes[0].value)) {
+                continue;
               }
+              interval = n2.nodes[0].value;
             } else if (n2.property === "associatedTissue") {
-              if (n2.nodes[0].value !== null) {
-                tissue = n2.nodes[0].value;
+              if (_.isEmpty(n2.nodes[0].value)) {
+                continue;
               }
+              tissue = n2.nodes[0].value;
             }
           }
           if (!seen.has(variant) && !!score) {
--- a/static/js/protein/data_processing_utils.ts
+++ b/static/js/protein/data_processing_utils.ts
@@ -135,7 +135,7 @@ export function getProteinInteraction(
           for (const n1 of n.nodes) {
             for (const n2 of n1.neighbors) {
               if (n2.property === "value") {
-                if (_.isEmpty(n.nodes)) {
+                if (_.isEmpty(n2.nodes)) {
                   continue;
                 }
                 const num = Number(n2.nodes[0].value);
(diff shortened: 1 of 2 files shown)


static/js/protein/data_processing_utils.ts
```diff
@@ -287,6 +287,34 @@ export function getVarGeneAssoc(data: GraphNodes): ProteinVarType[] {
...
+          continue;
+        }
+        for (const n1 of n.nodes) {
+          for (const n2 of n1.neighbors) {
+            if (n1.neighbors.length !== 4) {
+              continue;
+            }
+            console.log(n1.neighbors);
+            if (n2.property === "log2AllelicFoldChange") {
+              if (_.isEmpty(n2.nodes[0].value)) {
```
chejennifer, line 301: curious why here you're checking that the value is not empty & in the rest of the code you're checking that the value is not null? is there a reason for the difference in pattern?

chejennifer, line 301: should check that `n2.nodes` is not an empty list before accessing `n2.nodes[0].value` & same for the rest of the code. You can actually do the check before line 300 because it looks like the code following line 300 requires `n2.nodes` to not be an empty list.

static/js/protein/page.tsx
```diff
@@ -100,22 +100,60 @@ export class Page extends React.Component<PagePropType, PageStateType> {
   }
 
   render(): JSX.Element {
+    const name = this.props.nodeName;
+    const result = name.split("_");
```
chejennifer, line 104: since `.split()` doesn't modify the original string, you can just call the `.split()` method on `this.props.nodeName` in line 103 instead of creating 2 new variables. 

also add a comment about the format of this.props.nodeName and what will happen when the split happens

chejennifer (commented):
nice!

static/js/protein/data_processing_utils.ts
```diff
@@ -287,6 +287,34 @@ export function getVarGeneAssoc(data: GraphNodes): ProteinVarType[] {
...
+        */
+        if (n.property !== "geneSymbol") {
+          continue;
+        }
+        for (const n1 of n.nodes) {
+          for (const n2 of n1.neighbors) {
+            if (n1.neighbors.length !== 4) {
+              continue;
+            }
+            console.log(n1.neighbors);
```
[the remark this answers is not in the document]
suhana13: Done !

static/js/protein/data_processing_utils.ts
suhana13: Added the check !

chejennifer: actually null and empty are not the same, isEmpty catches more things as documented here: https://lodash.com/docs/4.17.15#isEmpty

static/js/protein/page.tsx
```diff
@@ -100,22 +100,62 @@ export class Page extends React.Component<PagePropType, PageStateType> {
   }
 
   render(): JSX.Element {
+    /* this.props.nodeName is formatted as ProteinName_SpeciesName
+    Using the split we get the ProteinName and SpeciesName separately
+    */
+    const result = this.props.nodeName.split("_");
```
chejennifer, line 106: nit: the variable name `result` isn't great here because there's no context for what result it's talking about. Maybe splitNodeName or processedNodeName or something about the name

static/js/protein/data_processing_utils.ts
```diff
@@ -192,6 +192,9 @@ export function getDiseaseGeneAssoc(data: GraphNodes): ProteinNumData[] {
         for (const n1 of n.nodes) {
           for (const n2 of n1.neighbors) {
             if (n2.property === "diseaseOntologyID") {
+              if (_.isEmpty(n2.nodes)) {
```
chejennifer, line 195: I don't think this is necessary because if n2.nodes is empty, the code won't go through the next for loop anyways

chejennifer (commented):
thanks for updating! Few more small things

static/js/protein/data_processing_utils.ts
suhana13: But wouldn't we want to check that n2.nodes is non-empty before we run the next for loop with n3 ?

chejennifer: the next for loop looks at each node in n2.nodes, but if there's no nodes, it just won't do anything because there's no nodes to look at?

suhana13: I see ! You're right, removed the unnecessary check !

chejennifer (approved):
Thanks for the changes!

PR #1636, opened 2022-06-09
Ranking chart frontend part 

raebo1017:
The frontend part of the ranking chart is implemented according to the previous design doc and the feedback received.  Except for the requirements listed in the design doc, one more thing is newly added: the image of the ranking chart won't be displayed in the chart embed.

To-do in the following PR:
1. update configs
2. web driver test

Screenshot of the chart:
https://screenshot.googleplex.com/egjufX6U986ZKw5

Screenshot of the chart embed:
https://screenshot.googleplex.com/AQ28oh9fEpAyXCM

Design and feedback Doc:
https://docs.google.com/document/d/1L5ighRK7mT0Z96nxii-_HEt-VED119wQSrlyWkDjnik/edit?resourcekey=0-PjZMG4hhTMKuCUoSdqA4eA#heading=h.7mpabhf4c47

https://docs.google.com/document/d/1FSojZpkbOgtm9sraHrhKevsfRMDTgMvyU3dm04kYeeM/edit?resourcekey=0-rJedtMgcx0RlD2GuGaYXZA#heading=h.njjqpwlx2pqv


static/js/chart/ranking_table.tsx
```diff
@@ -0,0 +1,145 @@
...
+  scaling?: number;
+  /**
+   * Total number of points. This is used for calculating ranks if isHighest is false
+   * (ordering from lowest ranks to highest) and ranks are not provided, e.g., n,n-1,n-2,...
+   */
+  numDataPoints?: number;
+  /**
+   * The row with dcid equal to currentDcid will be bolded.
+   */
+  currentDcid?: string;
```
chejennifer, line 64: nit: highlightedDcid

static/js/chart/ranking_table.tsx
```diff
@@ -0,0 +1,145 @@
...
+    }
+    return numberOfTotalDataPoints
+      ? numberOfTotalDataPoints - index
+      : index + 1;
+  }
+
+  return (
+    <div className="ranking-table">
+      <h4>
+        {formatString(props.title, {
```
chejennifer, line 89: why use formatString here? it looks like `props.title` is just going to be "Highest" or "Lowest" so there's nothing to replace in those titles?

static/js/chart/ranking_table.tsx
```diff
@@ -0,0 +1,145 @@
...
+   * The row with dcid equal to currentDcid will be bolded.
+   */
+  currentDcid?: string;
+  /**
+   * If true, only the ranks and labels will be shown and not the values.
+   */
+  notShowValue?: boolean;
+}
+
+export function RankingTable(props: RankingTablePropType): JSX.Element {
```
chejennifer, line 71: The code for this component is very similar to the code for the RankingUnit component.. the two should be consolidated. This can be done in a followup PR, but add a TODO for now.

chejennifer (commented):
Nice job!

static/js/chart/ranking_table.tsx
raebo1017: I tried to create a component that can be used by both the topic page and the place explorer page. This is the code copied from the [ranking unit component](https://github.com/datacommonsorg/website/blob/master/static/js/topic_page/ranking_unit.tsx#L45-L50). It will show the stats var in the title.

static/js/chart/ranking_table.tsx
```diff
@@ -0,0 +1,145 @@
...
+const NUM_FRACTION_DIGITS = 2;
+
+export interface Point {
+  dcid: string;
+  label: string;
+  value: number;
+  /**
+   * If not provided, the component will calculate the rank based on the order of the input points.
+   */
+  rank?: number;
```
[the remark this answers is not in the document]
raebo1017: I tried to create a component that can be used by both the topic page and the place explorer. The API used by the topic page does not return rank. So I make rank optional.
For the redirect link, what I'm thinking about is if there is a situation where the label is not a place and we do not need the link for that label?  
I want to make the ranking table more general. But if it doesn't make sense, I can change it.

static/js/place/chart.tsx
```diff
@@ -288,6 +338,9 @@ class Chart extends React.Component<ChartPropType, ChartStateType> {
   componentDidMount(): void {
     window.addEventListener("resize", this._handleWindowResize);
     this.processData();
+    if (this.props.chartType === chartTypeEnum.TABLE) {
```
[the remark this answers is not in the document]
raebo1017: Ranking table component needs to know whether to show the value based on the SVG container width when the page is rendered for the first time and no resize event happens. [Elemwidth](https://github.com/datacommonsorg/website/blob/master/static/js/place/chart.tsx#L146-L149) is set to be 0 as default. The reason why I call handelWindowResize at initial mounting is to set the elemwidth to be the current offset width and pass that to the ranking table component.

static/js/place/chart.tsx
```diff
@@ -557,10 +651,79 @@ class Chart extends React.Component<ChartPropType, ChartStateType> {
...
+        ? "(" + this.state.rankingChartDataGroup.date + ")"
+        : "";
     } else {
       return this.props.snapshot ? "(" + this.props.snapshot.date + ")" : "";
     }
   }
+
+  private getRankingTableData(): { lowest: Point[]; highest: Point[] } {
+    const lowestAndHighestDataPoints = { lowest: [], highest: [] };
+    const dataPoints: Point[] = this.state.rankingChartDataGroup.data.map(
```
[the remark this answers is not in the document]
raebo1017: I was thinking that the redirect link can be used other than just place. I tried to make the component more general so I add a prop for that.
What's more, I generalized the key name of the Point in the ranking-table component. It is consistent with general [datapoint](https://github.com/datacommonsorg/website/blob/master/static/js/place/chart.tsx#L121-L123) defined in the chart state. However, the key name is not the same as rankingChartDataGroup.data so I do a mapping to change  the key name...

static/js/place/chart_block.tsx
```diff
@@ -444,6 +449,43 @@ class ChartBlock extends React.Component<ChartBlockPropType> {
...
+            id={id}
+            dcid={this.props.dcid}
+            chartType={chartTypeEnum.TABLE}
+            // TODO: Support i18n.
+            title={intl.formatMessage(
+              {
+                defaultMessage: "{variable}: rankings in {placeName}",
+                description:
+                  "Used for chart titles like '{Unemployment rate}: rankings in {USA}'.",
+                id: "chart_clause-variable_in_place",
```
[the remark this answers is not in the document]
raebo1017: I'm not quite sure about how to add an i18n message. There are many languages there. Could you explain the overall process and when to use these commands?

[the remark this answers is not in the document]
chejennifer: so what I believe happens is when you add this piece of code (line 467-470) and then run `./scripts/extract_messages.sh`, a new key value pair will be added to the file /i18n/strings/en/place.json where the key is the `id` specified and the value is an object with defaultMessage and description. Therefore, when you are trying to add a new string, you want to use an id that hasn't already been used. Later, the strings will be sent for translation and once the string has been translated, the rest of the language files will be updated.

static/js/chart/ranking_table.tsx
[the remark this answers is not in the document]
chejennifer: I think for a ranking data point, it would make sense for there to be a rank field. For the ranking in topic pages, each data point is actually generated from the fetched data, so could add an additional rank field to the object: https://github.com/datacommonsorg/website/blob/master/static/js/topic_page/ranking_tile.tsx#L151

I think the redirect link is not necessary because currently both topic page and place page is doing the same redirect. I think when/if the time comes where we want to re-use this component to redirect to a different link, the component can just take in an additional prop such as `getRedirect: (point: Point) => string`

static/js/place/chart.tsx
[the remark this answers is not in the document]
chejennifer: That makes sense. Let's add a comment for this

static/js/place/chart_block.tsx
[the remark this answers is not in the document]
raebo1017: Done

static/js/chart/types.ts
```diff
@@ -137,3 +139,20 @@ export interface MapPoint {
...
+    placeDcid: string;
+    placeName: string;
+  }[];
+  numDataPoints: number;
+  exploreUrl: string;
+  sources: string[];
+}
+
+export interface CachedRankingChartData {
+  statVars: RankingChartDataGroup;
```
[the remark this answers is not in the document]
raebo1017: Done

static/js/chart/types.ts
```diff
@@ -21,6 +21,7 @@ export const chartTypeEnum = {
   CHOROPLETH: "CHOROPLETH",
   GROUP_LINE: "GROUP_LINE",
   HISTOGRAM: "HISTOGRAM",
+  TABLE: "TABLE",
```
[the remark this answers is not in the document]
raebo1017: Done

static/js/place/chart_embed.tsx
```diff
@@ -266,14 +267,16 @@ class ChartEmbed extends React.Component<unknown, ChartEmbedStateType> {
...
+          {this.chartDownloadXml && (
+            <Button color="primary" onClick={this.onDownloadSvg}>
+              {intl.formatMessage({
+                id: "embed_download_chart_link",
+                defaultMessage: "Download Chart Image",
+                description:
+                  "Text for the hyperlink text that will download the chart image.",
+              })}
+            </Button>
+          )}{" "}
```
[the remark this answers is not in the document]
raebo1017: Done

static/js/place/chart_block.tsx
```diff
@@ -444,6 +449,43 @@ class ChartBlock extends React.Component<ChartBlockPropType> {
           ></Chart>
         );
       }
+      if (this.props.data.isRankingChart) {
+        const parentPlaceName: string = this.parentPlaceDcid
+          ? this.props.names[this.parentPlaceDcid].split(",")[0]
+          : "";
+        this.displayDataTitle = this.props.data.title;
```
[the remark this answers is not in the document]
raebo1017: Done

static/js/chart/ranking_table.tsx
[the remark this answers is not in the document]
raebo1017: Done

static/js/chart/ranking_table.tsx
```diff
@@ -0,0 +1,145 @@
...
+   */
+  numDataPoints?: number;
+  /**
+   * The row with dcid equal to currentDcid will be bolded.
+   */
+  currentDcid?: string;
+  /**
+   * If true, only the ranks and labels will be shown and not the values.
+   */
+  notShowValue?: boolean;
```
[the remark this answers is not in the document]
raebo1017: Done

static/js/chart/ranking_table.tsx
raebo1017: Done

static/js/chart/ranking_table.tsx
raebo1017: Done

static/js/chart/ranking_table.tsx
raebo1017: Done. Also, change the rank of the lowest to n,n-1,n-2..on the topic page.

server/i18n/all.pot
```diff
@@ -874,663 +874,3 @@ msgstr "Nitrous oxide emissions (large facilities, non-biogenic)"
 #. Title of a place chart: Water withdrawal rate
 msgid "CHART_TITLE-water_withdrawal_rate"
 msgstr "Water withdrawal rate"
-
```
chejennifer, line 877: hm, curious why these got deleted, @beets?

static/js/place/chart.tsx
```diff
@@ -523,6 +589,35 @@ class Chart extends React.Component<ChartPropType, ChartStateType> {
...
+        if (this.props.rankingChartData) {
+          this.props.rankingChartData
+            .then((rankingChartData) => {
+              if (_.isEmpty(this.props.statsVars)) {
+                this.setState({ display: false });
+                return;
+              }
+              const sv = this.props.statsVars[0];
+              const svData = rankingChartData[sv];
+              for (const data of svData.data) {
```
chejennifer, line 602: should check that svData is not null or undefined before calling `svData.data`.
Maybe can move lines 606-610 before this for loop and just `return` inside the `if` statement after setting display to false.

static/js/place/chart.tsx
```diff
@@ -533,33 +628,83 @@ class Chart extends React.Component<ChartPropType, ChartStateType> {
...
+    const sliceNumber = Math.floor(MAX_RANKING_DATAPOINTS / 2);
+    lowestAndHighestDataPoints.lowest = data.data.slice(-sliceNumber).reverse();
+    lowestAndHighestDataPoints.highest = data.data.slice(0, sliceNumber);
+    return lowestAndHighestDataPoints;
+  }
+
+  private getRankingChartContainerTitle(): string {
+    const placeName = this.props.names[this.props.dcid] || this.props.dcid;
+    const currentPlaceRankAndValue = this.state.rankingChartDataGroup.data.find(
+      ({ placeDcid }) => placeDcid === this.props.dcid
```
chejennifer, line 696: oh, have never seen this syntax before, it works?
usually, we do: `(d) => d.placeDcid === this.props.dcid`

static/js/topic_page/ranking_unit.tsx
```diff
@@ -29,18 +29,54 @@ const NUM_FRACTION_DIGITS = 2;
...
 interface RankingUnitPropType {
   title: string;
   points: Point[];
-  statVarName: string;
+  /**
+   * If true, rows are ordered from highest ranks to lowest, e.g., 1,2,3...
+   * Otherwise, from lowest to highest, e.g., n,n-1,n-2,...
+   */
+  isHighest: boolean;
+  statVarName?: string;
```
chejennifer, line 47: Because we're making this a more generalized component, I think we should remove this optional `statVarName` prop and just have the `title` prop. We can do the `formatString()` in the RankingTile component and pass the fully formed title to this RankingUnit component

chejennifer (commented):
Thanks for the refactor and the updates!

static/js/place/chart.tsx
raebo1017: Yeup, Just learned from the [website ](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find)that destructuring can be used here

static/js/place/chart.tsx
raebo1017: Done

static/js/topic_page/ranking_unit.tsx
raebo1017: Done

server/i18n/all.pot
beets: yes - these should not be deleted. @raebo1017 for now, could you revert all the i18n files?

static/js/chart/types.ts
```diff
@@ -137,3 +141,20 @@ export interface MapPoint {
   latitude: number;
   longitude: number;
 }
+
+export interface RankingChartDataGroup {
```
beets, line 145: please add some documentation for the shared interface

static/js/topic_page/ranking_tile.tsx
```diff
@@ -126,15 +139,15 @@ function fetchData(
         for (const place in statData[item.statVar].stat) {
           const rankingPoint = {
             placeDcid: place,
-            stat: statData[item.statVar].stat[place].value,
+            value: statData[item.statVar].stat[place].value,
           };
-          if (rankingPoint.stat === undefined) {
+          if (typeof rankingPoint.value === "undefined") {
```
beets, line 144: should be sufficient to test `rankingPoint.value === undefined` rather than using typeof

beets (commented):
Hi Rae, the PR looks reasonable to me. There is definitely a bug in the i18n scripts that needs to be fixed. Could you revert the server side i18n files for now?

static/js/topic_page/ranking_tile.tsx
raebo1017: I could not pass the code style check with `rankingPoint.value === undefined`. It told me to avoid using undefined because it can be overwritten and it is not a keyword, so I change to use typeof...

static/js/chart/types.ts
raebo1017: Done

static/js/place/chart_block.tsx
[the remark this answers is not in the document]
raebo1017: Reran the commands after commenting out the python part.

static/js/topic_page/ranking_tile.tsx
chejennifer: maybe we can use the lodash method _.isUndefined()

chejennifer (approved):
Thanks for the updates!

static/js/topic_page/ranking_tile.tsx
raebo1017: Done

PR #1638, opened 2022-06-14
[download] Add source selection

chejennifer:
- add ability to select the source to download data for for each stat var
    - add flask endpoint for getting available sources
    - update flask endpoint for getting csv to get data for specific facets
    - client side updates to SourceSelector component and download tool components
- update flask endpoints used for download tool to use new V1 API

screen-recording-_43_

dev also updated: https://dev.datacommons.org/tools/download

### DIFF
--- a/server/routes/api/place.py
+++ b/server/routes/api/place.py
@@ -106,6 +106,8 @@ STATE_EQUIVALENTS = {"State", "AdministrativeArea1"}
 US_ISO_CODE_PREFIX = 'US'
 ENGLISH_LANG = 'en'
 EARTH_DCID = "Earth"
+PERSON_COUNT_LIMIT = 1000
+POPULATION_DCID = "Count_Person"
 
 # Define blueprint
 bp = Blueprint("api.place", __name__, url_prefix='/api/place')
@@ -800,30 +802,49 @@ def api_ranking_chart(dcid):
         parent_place_dcid = EARTH_DCID
         place_type = "Country"
     else:
-        parent_place_list = get_parent_place(dcid).get(dcid, [])
-        # Get the last parent place returned.
-        if parent_place_list:
-            parent_place_dcid = parent_place_list[-1]["dcid"]
-            place_type = get_place_type(dcid)
+        place_type = get_place_type(dcid)
+        parent_place_list = parent_places(dcid).get(dcid, [])
+        for parent in parent_place_list:
+            parent_type_list = parent.get("types", [])
+            parent_place_dcid = parent.get("dcid", "")
+            if parent_type_list:
+                parent_type = parent_type_list[0]
+                # All wanted place types plus continent except CensusZipCodeTabulationArea.
+                if parent_type == "Continent" or (
+                        parent_type in ALL_WANTED_PLACE_TYPES and
+                        parent_type != "CensusZipCodeTabulationArea"):
+                    break
+        # If break is not encountered, return empty result.
         else:
             return Response(json.dumps(result),
                             200,
                             mimetype='application/json')
+
     configs = get_ranking_chart_configs()
     # Get the first stat var of each config.
     stat_vars, _ = shared_api.get_stat_vars(configs)
+    # Make sure POPULATION_DCID is included in stat vars.
+    if POPULATION_DCID not in stat_vars:
+        stat_vars.add(POPULATION_DCID)
     sv_data = dc.get_stat_set_within_place(parent_place_dcid, place_type,
                                            list(stat_vars), "")
     sv_data_values = sv_data.get("data", {})
+
     if not sv_data or not sv_data_values:
         return Response(json.dumps(result), 200, mimetype='application/json')
-    ## Get all the place names of dcids in the sv_data
+    # Get all the place names of dcids in sv_data.
     place_dcids = set()
     for sv_value in sv_data_values.values():
         sv_place_dcids = sv_value.get("stat", {}).keys()
         place_dcids = place_dcids.union(sv_place_dcids)
-    place_names = get_name(list(place_dcids))
+    place_names = get_i18n_name(list(place_dcids))
     sv_metadata = sv_data.get("metadata", {})
+    # POPULATION_DCID is used to filter out the places with the population less than PERSON_COUNT_LIMIT
+    places_to_rank = set()
+    count_person_data = sv_data_values.get(POPULATION_DCID, {})
+    for place_dcid, place_data in count_person_data.get("stat", {}).items():
+        if place_data.get("value", 0) > PERSON_COUNT_LIMIT:
+            places_to_rank.add(place_dcid)
     # Consider the configs with single sv but ignore denominators.
     for config in configs:
         sv = config["statsVars"][0]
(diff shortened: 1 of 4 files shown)


server/routes/api/stats.py
```diff
@@ -414,65 +414,212 @@ def get_place_stat_date_within_place():
     return Response(json.dumps(response), 200, mimetype='application/json')
 
 
-def get_set_csv_rows(stat_set, sv_list, row_limit=None):
+def get_facets_by_variable_series(series_response):
```
shifucun, line 417: maybe rename to: get_variable_facets_from_series()

server/routes/api/stats.py
```diff
@@ -414,65 +414,212 @@ def get_place_stat_date_within_place():
...
+        facets_by_variable[sv] = {}
+        for place_obs in sv_obs.get("observationsByEntity", []):
+            for facet_obs in place_obs.get("seriesByFacet", []):
+                facet = str(facet_obs.get("facet", ""))
+                if facet and not facet in facets_by_variable[sv]:
+                    facets_by_variable[sv][facet] = facets.get(facet, {})
+    return facets_by_variable
+
+
+def get_facets_by_variable_points(points_response):
```
shifucun, line 451: nit: rename to: get_variable_facets_from_points()

server/routes/api/stats.py
```diff
@@ -414,65 +414,212 @@ def get_place_stat_date_within_place():
...
+        for place_obs in sv_obs.get("observationsByEntity", []):
+            for facet_obs in place_obs.get("pointsByFacet", []):
+                facet = str(facet_obs.get("facet", ""))
+                if not facet in facets_by_variable[sv]:
+                    facets_by_variable[sv][facet] = facets.get(facet, {})
+    return facets_by_variable
+
+
+@bp.route('/api/stats/facets/within-place')
+@cache.cached(timeout=3600 * 24, query_string=True)
```
shifucun, line 486: Since the dc.xxx API call are all cached, here does not really need to use @cached for caching, since there are no computation inside this function.

With that, can change this route to POST, which make will the client code more readable by using req_json instead of building long url string

server/routes/api/stats.py
```diff
@@ -704,3 +846,24 @@ def get_stats_within_place_csv():
...
+        childType: type of places to get data for
+        statVars: list of statistical variables to get data for
+        minDate (optional): earliest date to get data for
+        maxDate (optional): latest date to get data for
+        facetMap (optional): map of statistical variable dcid to the id of the
+            facet to get data from
+        rowLimit (optional): number of csv rows to return
+    """
+    req_str = json.dumps(request.json, sort_keys=True)
+    return get_stats_within_place_csv_wrapper(req_str)
```
shifucun, line 869: Can remove this wrapper now since dc.xxx call is cached. Also no need to cache this route either.

shifucun (commented):
Very clear and structured PR! Some minor comments for readability and caching.

static/js/tools/shared/tool_chart_footer.tsx
```diff
@@ -111,7 +113,8 @@ export function ToolChartFooter(props: ToolChartFooterPropType): JSX.Element {
           )}
           {props.children}
           <SourceSelector
-            svInfoList={props.sourceSelectorSvInfoList}
+            svMetahash={props.svMetahash}
```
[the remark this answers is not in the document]
chejennifer: makes sense, should "Facets" be something surfaced on the UI too? ie. currently, there's the option to "edit sources", should that text become "edit facets"?

server/routes/api/stats.py
chejennifer: get_stats_within_place_csv_wrapper has a decent amount of computation tho .. do we not want to cache that?

shifucun: I am not quite sure on this, for following reasons

1) The request body has a lot of params, it might not be common for a lot of requests to have the same request json (different users download exact the same thing)
2) The computation is mostly data conversion, which is fairly cheap

I feel caching makes sense if it involves API call (network traffic) or the request is very common and could be repeatedly used (browser/place page). Download tool might not be like that...

shifucun: Another reason is we don't want to cache all the stuff, as the memcache is getting full all the time already.. so keys are being evicted. https://pantheon.corp.google.com/memorystore/redis/locations/us-central1/instances/webserver-cache/details?project=datcom-website-prod

chejennifer: ah, ok, that makes sense to me, I will remove the caching for this one!

server/routes/api/stats.py
```diff
@@ -414,65 +414,212 @@ def get_place_stat_date_within_place():
...
-                metadata: {
-                    [metaHash]: {
-                        provenanceUrl: "",
+                ...
+            }
+    """
+    facets_by_variable = {}
+    facets = series_response.get("facets", {})
+    for sv_obs in series_response.get("observationsByVariable", []):
+        sv = sv_obs.get("variable", "")
```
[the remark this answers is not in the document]
chejennifer: done

server/routes/api/stats.py
```diff
@@ -414,65 +414,212 @@ def get_place_stat_date_within_place():
...
                     ...
-                }
+                },
+                ...
             }
+    """
+    facets_by_variable = {}
+    facets = points_response.get("facets", {})
+    for sv_obs in points_response.get("observationsByVariable", []):
+        sv = sv_obs.get("variable", "")
```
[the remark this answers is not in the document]
chejennifer: done

static/js/tools/timeline/chart.tsx
```diff
@@ -209,16 +215,14 @@ class Chart extends Component<ChartPropsType, ChartStateType> {
     this.drawChart();
   }
 
-  private getSourceSelectorSvInfo(statVars: string[]): SourceSelectorSvInfo[] {
+  private getSvSourceInfo(statVars: string[]): SourceSelectorSvSourceInfo[] {
```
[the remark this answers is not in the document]
chejennifer: done

static/js/tools/shared/tool_chart_footer.tsx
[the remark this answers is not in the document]
chejennifer: as per offline chat, in the code, updated "source" to "facet" and "metahash" to "facetId". Did this for code that was directly touched by this PR.

Will update uses of "metahash" in component Contexts in a later PR.

static/js/tools/download/page.tsx
```diff
@@ -265,7 +321,13 @@ export function Page(): JSX.Element {
     setSelectedOptions((prev) => {
       const updatedSV = _.cloneDeep(prev.selectedStatVars);
       updatedSV[sv] = svInfo;
-      return { ...prev, selectedStatVars: updatedSV };
+      const updatedFacets = _.cloneDeep(prev.selectedFacets);
+      updatedFacets[sv] = "";
+      return {
+        ...prev,
+        selectedStatVars: updatedSV,
```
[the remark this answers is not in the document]
chejennifer: done

server/routes/api/stats.py
chejennifer: done

server/routes/api/stats.py
chejennifer: done

server/routes/api/stats.py
chejennifer: done

server/routes/api/stats.py
chejennifer: done

chejennifer:
Thanks for the review!

PR #1639, opened 2022-06-15
Ranking chart API: Restrict parent place type and add population filter 

raebo1017:
1. Restrict the parent place type and keep consistent with the rule of place filtering in [getParentPlaces](https://github.com/datacommonsorg/mixer/blob/master/internal/server/placepage/place_page.go) in mixer.

2. Add a population filter so that API won't return data points of places with populations less than 1000. Keep consistent with the rules of the population filtering of the [ranking page](https://source.corp.google.com/piper///depot/google3/datacommons/prophet/related_places/calculator_svobs.cc;l=309-321).

### DIFF
--- a/server/routes/api/place.py
+++ b/server/routes/api/place.py
@@ -801,9 +801,9 @@ def api_ranking_chart(dcid):
         place_type = "Country"
     else:
         parent_place_list = get_parent_place(dcid).get(dcid, [])
-        # Get the last parent place returned.
+        # Get the first parent place returned.
         if parent_place_list:
-            parent_place_dcid = parent_place_list[-1]["dcid"]
+            parent_place_dcid = parent_place_list[0]["dcid"]
             place_type = get_place_type(dcid)
         else:
             return Response(json.dumps(result),
--- a/server/routes/api/place.py
+++ b/server/routes/api/place.py
@@ -107,6 +107,7 @@ US_ISO_CODE_PREFIX = 'US'
 ENGLISH_LANG = 'en'
 EARTH_DCID = "Earth"
 PERSON_COUNT_LIMIT = 1000
+POPULATION_DCID = "Count_Person"
 
 # Define blueprint
 bp = Blueprint("api.place", __name__, url_prefix='/api/place')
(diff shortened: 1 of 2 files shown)


server/routes/api/place.py
```diff
@@ -800,30 +801,50 @@ def api_ranking_chart(dcid):
...
                             200,
                             mimetype='application/json')
+
     configs = get_ranking_chart_configs()
     # Get the first stat var of each config.
     stat_vars, _ = shared_api.get_stat_vars(configs)
+    # Make sure the ranking chart configs include stat var "Count_Person".
+    if "Count_Person" not in stat_vars:
+        stat_vars.add("Count_Person")
     sv_data = dc.get_stat_set_within_place(parent_place_dcid, place_type,
```
chejennifer, line 828: We are trying to migrate to V1 APIs (from Mixer) now & to help with this process, let's try to use the new mixer APIs wherever possible in new and touched code. I have a [pr](https://github.com/datacommonsorg/website/pull/1638/files) up right now that adds the new function dc.points_within that will eventually replace dc.get_stat_set_within_place. Once that pr is in, let's update this flask endpoint to use the new function. Can do this in a followup PR

server/routes/api/place.py
```diff
@@ -800,30 +801,50 @@ def api_ranking_chart(dcid):
...
     for sv_value in sv_data_values.values():
         sv_place_dcids = sv_value.get("stat", {}).keys()
         place_dcids = place_dcids.union(sv_place_dcids)
-    place_names = get_name(list(place_dcids))
+    place_names = get_i18n_name(list(place_dcids))
     sv_metadata = sv_data.get("metadata", {})
+    # "Count_Person" is used to filter out the places with the population less than PERSON_COUNT_LIMIT
+    places_to_rank = set()
+    count_person_data = sv_data_values.get("Count_Person", {})
+    if count_person_data:
```
chejennifer, line 844: nit: this `if` statement is unnecessary because doing `.get("stat", {})` on an empty object will just return an empty object, so you can do the for-loop on line 845 without checking that count_person_data is empty or not

chejennifer (commented):
nice job catching the bugs

server/routes/api/place.py
```diff
@@ -800,30 +801,50 @@ def api_ranking_chart(dcid):
...
         else:
             return Response(json.dumps(result),
                             200,
                             mimetype='application/json')
+
     configs = get_ranking_chart_configs()
     # Get the first stat var of each config.
     stat_vars, _ = shared_api.get_stat_vars(configs)
+    # Make sure the ranking chart configs include stat var "Count_Person".
+    if "Count_Person" not in stat_vars:
```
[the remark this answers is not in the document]

server/routes/api/place.py
raebo1017: Got you! Done

server/routes/api/place.py
raebo1017: Sure!

PR #1641, opened 2022-06-17
[protein browser] Add depth-1 protein-protein interaction graph

rogershijin:
Features:
- Central protein and top 10 interactors (by confidenceScore) are displayed.
- Nodes colored by distance to central protein.
- Nodes draggable.
- Link width proportional to confidenceScore.
- Nodes/links highlight on hover (currently uses opacity unfortunately which produces transparency effect)

Screenshot:
image


static/js/protein/chart.ts
```diff
@@ -432,6 +460,280 @@ export function drawProteinInteractionChart(
...
+
+  const linkData: InteractionLink[] = nodeData.map((node) => {
+    return {
+      score: node.value,
+      source: centerNodeID,
+      target: node.id,
+    };
+  });
+
+  const height = 400 - MARGIN.top - MARGIN.bottom;
```
chejennifer, line 569: 400 and 700 should be constants

static/js/protein/chart.ts
```diff
@@ -432,6 +460,280 @@ export function drawProteinInteractionChart(
...
+  const width = 700 - MARGIN.left - MARGIN.right;
+
+  const svg = d3
+    .select(`#${elementID}`)
+    .append("svg")
+    .attr("width", width + MARGIN.left + MARGIN.right)
+    .attr("height", height + MARGIN.top + MARGIN.bottom)
+    .attr("viewBox", `${-width / 2} ${-height / 2} ${width} ${height}`)
+    .append("g")
+    .attr("transform", `translate(${MARGIN.left - 150}, ${MARGIN.top - 25})`)
```
chejennifer, line 579: where do the 150 and 25 come from?

static/js/protein/chart.ts
```diff
@@ -432,6 +460,280 @@ export function drawProteinInteractionChart(
...
+    .attr("style", "max-width: 100%; height: auto; height: intrinsic");
```
chejennifer, line 580: these styles should be in the css file

static/js/protein/chart.ts
```diff
@@ -432,6 +460,280 @@ export function drawProteinInteractionChart(
...
+    .attr("fill", NODE_STYLE.labels.font.color)
+    .attr("text-anchor", "middle")
+    .attr("dominant-baseline", "middle")
+    .style(
+      "font",
+      `${NODE_STYLE.labels.font.size} ${NODE_STYLE.labels.font.name}`
+    )
+    .text(({ name }) => `${name}`);
+
+  // floating name labels we can switch in
```
chejennifer, line 644: what do floating name labels look like? are you trying to decide between that and the current labels? maybe we can make the decision and not have to check in both the code

static/js/protein/chart.ts
```diff
@@ -432,6 +460,280 @@ export function drawProteinInteractionChart(
...
+  // const labels = node
+  //   .append("text")
+  //   .attr("fill", "#555")
+  //   .attr("dx", 15)
+  //   .attr("dy", -15)
+  //   .text(({name}) => `${name}`)
+
+  function ticked(): void {
```
chejennifer, line 652: this and `drag()` should be functions that are defined outside of the `drawProteinInteractionGraph` function

static/js/protein/chart.ts
```diff
@@ -432,6 +460,280 @@ export function drawProteinInteractionChart(
...
+
+    return d3
+      .drag()
+      .on("start", dragstarted)
+      .on("drag", dragged)
+      .on("end", dragended);
+  }
+
+  // @reviewers: lightenThis/darkenThis should probably use color.darker() and color.brighter() instead so as not to alter transparency.
+  // However, transparency effect may be useful to view edges entering a node in a dense graph.
```
chejennifer, line 724: if this is trying to show mouse hover state, I think changing the color would be better than opacity. These functions should also be defined outside of `drawProteinInteractionGraph` function.

When you have something you want to bring to the attention of reviewers, please just include a comment in your PR instead of including it in your code. Please remove these comments from the actual code.

static/js/protein/chart.ts
```diff
@@ -432,6 +460,280 @@ export function drawProteinInteractionChart(
...
+
+  function lightenThis(): void {
```
chejennifer, line 726: nit: just `lighten` and `darken` instead of `lightenThis` and `darkenThis`

chejennifer (commented):
Nice job!

static/js/protein/chart.ts
```diff
@@ -31,6 +38,27 @@ export interface ProteinStrData {
   value: string;
 }
 
+// types and interfaces for protein-protein interaction chart
+type NodeID = string | number;
```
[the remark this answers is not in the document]
rogershijin: Sorry! I had copied the line 508 example data from an old implementation I inherited from Sam/Suhana where they ID'd with numbers.  I now id with {protein}_{species} (e.g. P53_HUMAN).  I'll update 508 with the new IDs.  And I think just in case this interface gets reused for some other graph viz in the future perhaps we can leave the numerical ID support in there?

static/js/protein/chart.ts
```diff
@@ -432,6 +460,280 @@ export function drawProteinInteractionChart(
...
+  elementID: string,
+  data: InteractingProteinType[]
+): void {
+  /*
+  References:
+    1) Force-directed layout Observable: https://observablehq.com/@d3/force-directed-graph
+    2) Andrew Chen's force-directed layout with text labels tutorial: https://www.youtube.com/watch?v=1vHjMxe-4kI
+  */
+
+  const MAX_INTERACTIONS = 10;
```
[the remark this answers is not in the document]
rogershijin: thanks for the feedback! will do!

static/js/protein/chart.ts
```diff
@@ -432,6 +460,280 @@ export function drawProteinInteractionChart(
...
+      radius: 15,
+      stroke: {
+        color: "#fff",
+        opacity: 1,
+        width: 1.5,
+      },
+    },
+    labels: {
+      font: {
+        color: "#222",
```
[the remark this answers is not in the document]
rogershijin: sounds good! I'll move these over

static/js/protein/chart.ts
```diff
@@ -432,6 +460,280 @@ export function drawProteinInteractionChart(
...
+    stroke: {
+      color: "#999",
+      linecap: "round",
+      opacity: 0.6,
+      scoreWidthMultiplier: 8,
+    },
+  };
+
+  /*
+  DATA PROCESSING
```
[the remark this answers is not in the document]
rogershijin: Ah yes that would be cleaner! Will do!

static/js/protein/chart.ts
```diff
@@ -432,6 +460,280 @@ export function drawProteinInteractionChart(
...
+
+    links = [
+      { source: 1, target: 2, score: 0.3 },
+      { source: 1, target: 3, score: 0.7 }
+    ]
+  */
+
+  function nodeFromID(protein_speciesID: string, depth: number): ProteinNode {
+    // protein_speciesID: id of form {protein}_{species}, e.g. P53_HUMAN
+    const lastIndex = protein_speciesID.lastIndexOf("_"); // danger: assumes species name does not contain _
```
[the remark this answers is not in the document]
rogershijin: Thanks will do! And will check with Sam about the underscores - I had used `lastIndexOf` because I thought it was less likely for species name to contain an underscore but I could definitely have this backwards.

static/js/protein/chart.ts
rogershijin: Good question! They look something like this:
image

I think when the graph gets more crowded it may be easier to use smaller circles and floating labels.  However it's easier to tell which label is associated with which circle with the inside-circle labels.  Perhaps we can make this an option instead of a comment and later on change between floating and inside based on how crowded the graph is?

static/js/protein/chart.ts
rogershijin: O you caught me! These are from trial and error in trying to position the graph near the center of the space.    Should I try to compute these from height/width? I was also thinking of forcing the central (page) protein to be at the center of the box but for higher depth graphs it might sometimes look better to have the central protein on the left (resp. top left) and {node distance from center} increasing left to right  (resp. top left --> bottom right).

static/js/protein/chart.ts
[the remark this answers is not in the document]
chejennifer: currently the Node interface on line 44 is just used here right? you can just make id type to be string .. I don't think it's necessary to create a separate interface NodeID.

static/js/protein/chart.ts
chejennifer: do you have an example of a crowded graph? I feel like having a lot of floating labels when a graph is crowded could also get confusing ..

static/js/protein/chart.ts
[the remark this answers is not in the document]
rogershijin: sounds good!

static/js/protein/chart.ts
rogershijin: Sadly I don't have such an example yet (will soon when I finish implementing breadth 2).  I'll just take out the floating label comment and we can revisit this question when the graph gets complicated enough to think about this with proper examples.

static/js/protein/chart.ts
```diff
@@ -432,6 +460,280 @@ export function drawProteinInteractionChart(
...
+      depth,
+      id: protein_speciesID,
+      name: protein_speciesID.slice(0, lastIndex),
+      species: protein_speciesID.slice(lastIndex + 1),
+    };
+  }
+
+  // P53_HUMAN is central protein in below examples.
+  // take interaction names of the form P53_HUMAN_ASPP2_HUMAN | ASPP2_HUMAN_P53_HUMAN and parse into ASPP2_HUMAN.
+  const centerNodeID = data[0].parent;
```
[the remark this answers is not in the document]
rogershijin: Nice catch! Added Suhana's emptiness guard to top of function.

static/js/protein/chart.ts
rogershijin: oops yes! replaced with Suhana's new constants GRAPH_HEIGHT_M, GRAPH_WIDTH_M.

static/js/protein/chart.ts
rogershijin: Oh ok I think global then (because I haven't generated enough examples to know what this computation should actually take into account).  Might be something like the center of mass of the graph should be as close to the center of the plot as possible without any node going outside the graph but this sounds like a job for future Roger.

static/js/protein/chart.ts
rogershijin: I spent like 2 hrs trying to figure this one out! Eventually went with [brightness](https://stackoverflow.com/a/69610045) which doesn't require stateful tracking of the original color of object under mouse and also doesn't alter transparency like opacity does.

chejennifer (commented):
Thanks for the updates!

static/js/protein/data_processing_utils.ts
```diff
@@ -162,6 +169,84 @@ export function getProteinInteraction(
   return [];
 }
 
+export function nodeFromID(
```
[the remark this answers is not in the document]
rogershijin: nice catch!

static/js/protein/data_processing_utils.ts
```diff
@@ -162,6 +169,84 @@ export function getProteinInteraction(
...
+          { source: MECOM_HUMAN, target: CTPB1_HUMAN, score: 0.3 },
+          { source: MECOM_HUMAN, target: SUPT16H_HUMAN, score: 0.7 },
+        ],
+
+      }
+  */
+
+  // P53_HUMAN is central protein in below examples.
+  // take interaction names of the form P53_HUMAN_ASPP2_HUMAN | ASPP2_HUMAN_P53_HUMAN and parse into ASPP2_HUMAN.
+  const centerNodeID = data[0].parent;
```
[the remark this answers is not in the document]
rogershijin: oops yes my bad! thanks for finding this!

static/js/protein/chart.ts
```diff
@@ -198,6 +262,89 @@ function addXLabel(
...
+
+  function dragged(nodeDatum: ProteinNode): void {
+    nodeDatum.fx = d3.event.x;
+    nodeDatum.fy = d3.event.y;
+  }
+
+  function dragended(nodeDatum: ProteinNode): void {
+    if (!d3.event.active) {
+      simulation.alphaTarget(0);
+    } // cool down simulation
```
chejennifer, line 336: put comment in the line above the code it refers to?

chejennifer (approved):
Thanks for the updates! Just a few commenting nits

PR #1642, opened 2022-06-22
Ranking Chart API: change to v1 endpoint

raebo1017:
Changed the endpoint `get_stat_set_within_place` to v1 endpoint `points_within`.


server/routes/api/place.py
```diff
@@ -819,65 +819,81 @@ def api_ranking_chart(dcid):
...
-
-    if not sv_data or not sv_data_values:
+    points_response_best = dc.points_within(parent_place_dcid, place_type,
+                                            list(stat_vars), "", False)
+    sv_data = points_response_best.get("observationsByVariable")
+    sv_facets = points_response_best.get("facets")
+    if not points_response_best or not sv_data:
         return Response(json.dumps(result), 200, mimetype='application/json')
+    # Build a dict to map stat vars to dicts mapping place dcids to their data.
+    sv_to_data_points = {}
```
shifucun, line 835: Is this intermediate data structure necessary or all the following logic can just happen while looping through the original response?

raebo1017: Per offline discussion, I replaced the intermediate data structure with a dict mapping config svs to dicts of unit and scaling. Instead of looping through the configs, I loop through the sv in the original response data and check against the map.

chejennifer (commented):
nice job!

server/routes/api/place.py
```diff
@@ -819,65 +819,87 @@ def api_ranking_chart(dcid):
...
                             mimetype='application/json')
-
+    # Read configs and build a dic to map stat vars to dicts of unit and scaling.
+    # Consider the configs with single sv but ignore denominators.
     configs = get_ranking_chart_configs()
+    config_sv_to_info = {}
+    for config in configs:
+        stat_vars = config.get("statsVars")
+        if stat_vars:
+            sv = stat_vars[0]
```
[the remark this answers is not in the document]

PR #1644, opened 2022-06-23
Add protein description and clickable functionality for protein-protein interactions

suhana13:


static/js/protein/chart.ts
```diff
@@ -442,10 +442,10 @@ export function drawProteinInteractionChart(
     .attr("height", y.bandwidth())
     .style("fill", BAR_COLOR)
     //PROTEIN_REDIRECT
-    .on("click", function(d) {
+    .on("click", function (d) {
       const protein_url = "bio/" + d.name + "_" + d.parent;
```
chejennifer, line 446: - variable names should be camelCase
- maybe proteinId instead proteinUrl?

chejennifer (commented):
nice!

static/js/protein/chart.ts
suhana13: Done!

PR #1645, opened 2022-06-24
add legend and fix bar width

suhana13:


static/js/protein/chart.ts
```diff
@@ -219,6 +240,76 @@ function addYLabel(
...
+  // checks if the data is empty or not
+  if (_.isEmpty(data)) {
+    return;
+  }
+  const svg = d3
+    .select("#tissue-score-legend")
+    .append("svg")
+    .attr("width", GRAPH_WIDTH_XL)
+    .attr("height", GRAPH_HEIGHT_XS);
+  const data0 = d3.keys(TISSUE_LEGEND_DICT);
```
chejennifer, line 253: nit: allOrgans or organTypes or something more specific than data

static/js/protein/chart.ts
```diff
@@ -219,6 +240,76 @@ function addYLabel(
...
+  // slicing the dictionary in half to display the legend in two rows
+  const data1 = data0.slice(0, 8);
```
chejennifer, line 255: instead of using 8 and 15, should calculate those indices using the length of data0 instead so in the future, if that object gets updated, you don't have to also update these numbers.

nit: name as dataRowOne and dataRowTwo instead (if name of constant at line 253 is updated, then <name of constant>RowOne and <name of constant>RowTwo)

static/js/protein/chart.ts
```diff
@@ -219,6 +240,76 @@ function addYLabel(
...
+  const data2 = data0.slice(8, 15);
+  // creating circles for the first row in legend
+  svg
+    .selectAll("mydots1")
```
chejennifer, line 259: - since the dots on the first row and the second row should be styled the same way, don't need to distinguish between "mydots1" and "mydots2"
- when naming components and variables, adding possessive adjectives (eg. my) is unnecessary. Here, it can just be "dots" or "legend-dots"

static/js/protein/chart.ts
```diff
@@ -219,6 +240,76 @@ function addYLabel(
...
+    .data(data1)
+    .enter()
+    .append("circle")
+    .attr("cx", (d, i) => {
+      return NUM_DATA_POINTS + i * GRAPH_HEIGHT_XS;
```
chejennifer, line 264: curious why the x coordinate of the dot depends on the graph height? Because shouldn't x coordinate be related to width of the graph? 
Also curious how NUM_DATA_POINTS plays into the calculations

static/js/protein/chart.ts
```diff
@@ -219,6 +240,76 @@ function addYLabel(
...
+    })
+    .attr("cy", 50)
+    .attr("r", 7)
```
chejennifer, line 267: 7 can be a constant or can be added to the css

static/js/protein/chart.ts
```diff
@@ -219,6 +240,76 @@ function addYLabel(
...
+    .style("fill", (d) => {
+      return TISSUE_LEGEND_DICT[d];
+    });
+  svg
+    .selectAll("legend-text1")
```
chejennifer, line 272: since the text on the first row and the second row should be styled the same way, don't need to distinguish between "legend-text1" and "legend-text2", can just be "legend-text"

static/js/protein/chart.ts
```diff
@@ -219,6 +240,76 @@ function addYLabel(
...
+    .data(data1)
+    .enter()
+    .append("text")
+    .attr("x", (d, i) => {
+      return 2 * NUM_DATA_POINTS + i * GRAPH_HEIGHT_XS;
+    })
+    .attr("y", 55)
+    .style("font-size", "10px")
```
chejennifer, line 280: can be added to the css

static/js/protein/chart.ts
```diff
@@ -424,7 +545,11 @@ export function drawProteinInteractionChart(
     .attr("transform", "translate(-10,0)rotate(-45)")
     .style("text-anchor", "end");
   // Adds the x-axis text label
-  addXLabel(width, height, "Confidence Score (IntactMiScore)", svg);
+  if (arrayLength < 10) {
```
chejennifer, line 548: this if statement seems unnecessary since the code in both cases is the exact same?

static/js/protein/chart.ts
```diff
@@ -465,8 +590,16 @@ export function drawDiseaseGeneAssocChart(
   if (_.isEmpty(data)) {
     return;
   }
+  //Finds the length of the object array
+  const arrayLength = Object.keys(data).length;
+  let height = null;
```
chejennifer, line 595: you don't actually have to set height as anything, you can do `let height: number;`

static/js/protein/chart.ts
```diff
@@ -373,20 +469,45 @@ export function drawProteinInteractionChart(
...
-
   reformattedData = reformattedData.filter((entry) => {
     const duplicate = seen.has(entry.name);
     seen.add(entry.name);
     return !duplicate;
   });
-  const height = GRAPH_HEIGHT_M - MARGIN.top - MARGIN.bottom;
+  //Finds the length of the object array
+  let arrayLength = Object.keys(reformattedData).length;
+  //Formats the graph as per number of entries
```
chejennifer, line 479: This feels quite hacky, could you try to accomplish drawing graphs that don't have enough data points by having a MIN_HEIGHT for charts and then calculating the scaleBand paddingInner and paddingOuter?

scaleBand documentation: https://observablehq.com/@d3/d3-scaleband

static/js/protein/chart.ts
```diff
@@ -447,7 +572,7 @@ export function drawProteinInteractionChart(
       window.location.href = `${PROTEIN_REDIRECT}${proteinId}`;
     })
     .on("mouseover", mouseover)
-    .on("mousemove", () => mousemove(TOOL_TIP_LEFT_POSITION, 510))
+    .on("mousemove", () => mousemove(TOOL_TIP_LEFT_POSITION, 650))
```
chejennifer, line 575: Can you try to calculate the topPosition or in mousemove, calculate the y without topPosition. You don't want to have to update this value for all the charts every time there is an update to chart dimensions. 

there are a few different MouseEvent properties that you can try out.
documentation on MouseEvent properties: https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent
documentation on the diff between a few of the properties: https://stackoverflow.com/questions/6073505/what-is-the-difference-between-screenx-y-clientx-y-and-pagex-y

chejennifer (commented):
Thanks for the UI updates!

run_npm.sh
```diff
@@ -14,7 +14,7 @@
 # limitations under the License.
 
 cd static
-export NODE_OPTIONS=--openssl-legacy-provider
+#export NODE_OPTIONS=--openssl-legacy-provider
```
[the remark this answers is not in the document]
suhana13: Sorry about that, didn't mean to push those changes. Fixed it!

static/package-lock.json
```diff
@@ -40,6 +40,7 @@
         "react-collapsible": "2.8.3",
         "react-dom": "^16.13.1",
         "react-intl": "^5.10.1",
+        "react-papaparse": "^4.0.2",
```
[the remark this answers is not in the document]
suhana13: Sorry about that, didn't mean to push those changes. Fixed it!

static/js/protein/chart.ts
```diff
@@ -86,51 +90,69 @@ const TISSUE_COLOR_DICT = {
...
-  ThyroidGland: "snow",
-  Tonsil: "saddlebrown",
+  Thymus: "khaki",
+  ThyroidGland: "bisque",
+  Tonsil: "khaki",
   UrinaryBladder: "sandybrown",
   Vagina: "mistyrose",
 };
+// color dictionary for tissue legend
+const TISSUE_LEGEND_DICT = {
```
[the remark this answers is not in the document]
suhana13: That's an excellent suggestion! Implemented it!!

static/js/protein/chart.ts
```diff
@@ -219,6 +240,76 @@ function addYLabel(
     .text(labelText);
 }
 
+export function drawTissueLegend(id: string, data: ProteinStrData[]): void {
+  // checks if the data is empty or not
+  if (_.isEmpty(data)) {
+    return;
+  }
+  const svg = d3
+    .select("#tissue-score-legend")
```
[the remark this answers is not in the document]
suhana13: correct, added that!

static/js/protein/chart.ts
```diff
@@ -219,6 +240,76 @@ function addYLabel(
     .text(labelText);
 }
 
+export function drawTissueLegend(id: string, data: ProteinStrData[]): void {
```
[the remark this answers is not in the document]
suhana13: Done!

static/js/protein/chart.ts
```diff
@@ -219,6 +240,76 @@ function addYLabel(
...
+  if (_.isEmpty(data)) {
+    return;
+  }
+  const svg = d3
+    .select("#tissue-score-legend")
+    .append("svg")
+    .attr("width", GRAPH_WIDTH_XL)
+    .attr("height", GRAPH_HEIGHT_XS);
+  const data0 = d3.keys(TISSUE_LEGEND_DICT);
+  // slicing the dictionary in half to display the legend in two rows
```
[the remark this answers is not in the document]
suhana13: Yes, I checked that. The legend behaves the same as the graphs would. If I shrink or widen the window, both the graph and legend get cut off or stretched respectively!

static/js/protein/chart.ts
suhana13: Used organTypes!

static/js/protein/chart.ts
suhana13: Done!

static/js/protein/chart.ts
suhana13: Done!

static/js/protein/chart.ts
suhana13: I want to use the values of NUM_DATA_POINTS and GRAPH_HEIGHT_XS, that's why I used those constants. Is it okay to define more constants with the same values?

static/js/protein/chart.ts
suhana13: Added as a constant!

static/js/protein/chart.ts
suhana13: Done!

static/js/protein/chart.ts
suhana13: Should I add another class in the css just for the font size ?

static/js/protein/chart.ts
suhana13: Removed it!

static/js/protein/chart.ts
suhana13: Done!

static/js/protein/chart.ts
suhana13: That really helped! I did the math and came up with an inner bar padding constant with automates as per the data points! It should be all good now

static/js/protein/chart.ts
suhana13: That's a great idea! Used d3.event.pageX and d3.event.pageY instead!

static/js/protein/chart.ts
chejennifer: yes, that's ok to define more constants with the same values

static/js/protein/chart.ts
```diff
@@ -241,8 +345,8 @@ export function drawTissueScoreChart(id: string, data: ProteinStrData[]): void {
 
   //groups the tissues of a similar origin and sorts them in ascending order
   reformattedData.sort((x, y) => {
-    const a = TISSUE_COLOR_DICT[x.name];
-    const b = TISSUE_COLOR_DICT[y.name];
+    const a = TISSUE_ORGAN_DICT[x.name];
```
chejennifer, line 348: do we know that all names in reformattedData will be in TISSUE_ORGAN_DICT? if not, maybe you want to first check that the names are in the DICT and if they're missing, sort them to the end?

static/js/protein/chart.ts
```diff
@@ -313,9 +418,9 @@ export function drawTissueScoreChart(id: string, data: ProteinStrData[]): void {
     .attr("y", (d) => y(d.value))
     .attr("height", (d) => height - y(d.value))
     .attr("width", x.bandwidth())
-    .style("fill", (d) => TISSUE_COLOR_DICT[d.name])
+    .style("fill", (d) => ORGAN_COLOR_DICT[TISSUE_ORGAN_DICT[d.name]])
```
chejennifer, line 421: do we know that all names in reformattedData will be in TISSUE_ORGAN_DICT? maybe we can have a default color for cases where d.name is not in TISSUE_ORGAN_DICT or the organ that d.name is part of isn't in ORGAN_COLOR_DICT?

chejennifer (commented):
Thanks for the updates!

static/package-lock.json
```diff
@@ -32789,4 +32789,4 @@
       "integrity": "sha512-rVksvsnNCdJ/ohGc6xgPwyN8eheCxsiLM8mxuE/t/mOVqJewPuO1miLpTHQiRgTKCLexL4MeAFVagts7HmNZ2Q=="
     }
   }
-}
+}
```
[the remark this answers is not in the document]
suhana13: I just copied over the file from datacommons/master. Would you suggest a better way to do it?

static/js/protein/chart.ts
```diff
@@ -347,6 +452,10 @@ export function drawProteinInteractionChart(
     if (d === "") {
       d = parentProtein;
     }
+    // for the case when a blank entry is added for graph formatting
```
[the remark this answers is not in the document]
suhana13: You're right! Removed it!

static/package-lock.json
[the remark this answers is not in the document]
chejennifer: I think you added an extra space at the end of line 32792, just trying copying that line again

static/js/protein/chart.ts
```diff
@@ -219,6 +244,85 @@ function addYLabel(
...
+    return;
+  }
+  const svg = d3
+    .select("#" + id)
+    .append("svg")
+    .attr("width", GRAPH_WIDTH_XL)
+    .attr("height", GRAPH_HEIGHT_XS);
+  const organTypes = d3.keys(ORGAN_COLOR_DICT);
+  // slicing the dictionary in half to display the legend in two rows
+  const dataRowOne = organTypes.slice(0, (organTypes.length + 1) / 2);
```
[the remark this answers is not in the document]
suhana13: Done!

static/js/protein/chart.ts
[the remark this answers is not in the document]
suhana13: Screen Shot 2022-06-27 at 1 46 25 PM
Here's how it looks like when shrinking the window. You're right that we shouldn't be experiencing this issue but it might be a good idea to fix it in another PR, with mobile screen view?

static/js/protein/chart.ts
suhana13: Done!

static/js/protein/chart.ts
suhana13: Done!

chejennifer (approved):

static/js/protein/chart.ts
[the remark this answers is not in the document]
suhana13: You're right! Added as P0 to the protein to-do list

PR #1646, opened 2022-06-25
Ranking chart webdriver test and configs update.

raebo1017:
1. Add a webdriver test for the ranking chart.
2. Add"isRankingChart" to the configs that have only one stat var and no denominators.


server/webdriver_tests/tests/place_explorer_test.py
```diff
@@ -186,3 +186,82 @@ def test_demographics_redirect_link(self):
...
+            './/table/tbody/tr[1]/td')
+        self.assertEqual(chart_table_highest_row[0].text, FIRST_RANK)
+
+        chart_table_lowest_row = chart_table_lowest.find_elements_by_xpath(
+            './/table/tbody/tr[1]/td')
+        self.assertEqual(chart_table_lowest_row[0].text, LAST_RANK)
+
+    def test_ranking_chart_redirect_link(self):
+        """
+        Test the rediret link can work correctly.
```
chejennifer, line 247: "... the redirect link .."

server/webdriver_tests/tests/place_explorer_test.py
```diff
@@ -186,3 +186,82 @@ def test_demographics_redirect_link(self):
...
+        chart_subtitle = self.driver.find_element_by_xpath(
+            '//*[@id="main-pane"]/section[5]/div/div[5]/div/div/div/h4').text
+        self.assertTrue = ("California ranks" in chart_subtitle)
+
+        chart_table_highest = self.driver.find_element_by_xpath(
+            '//*[@id="main-pane"]/section[5]/div/div[5]/div/div/div/div/div[1]')
+        chart_table_lowest = self.driver.find_element_by_xpath(
+            '//*[@id="main-pane"]/section[5]/div/div[5]/div/div/div/div/div[2]')
+
+        chart_table_highest_title = EC.text_to_be_present_in_element((
```
chejennifer, line 223: I think instead of waiting for the text to be present in the element, it'll be better to wait for the element to be present, then assert that the text in the element is expected. This way, we know if a test fails because an element doesn't show up or if the text is incorrect. Here and chart_table_lowest_title and checking the redirect link in the next test

chejennifer (commented):
nice job!

server/webdriver_tests/tests/place_explorer_test.py
raebo1017: Done

server/webdriver_tests/tests/place_explorer_test.py
raebo1017: Done

chejennifer (approved):
Thanks for the changes!

PR #1647, opened 2022-06-28
feat: add barebones content for disease browser

suhana13:


server/__init__.py
```diff
@@ -47,15 +47,18 @@ def createMiddleWare(app, exporter):
...
                         special_announcement, topic_page)
     app.register_blueprint(dev.bp)
+    app.register_blueprint(disease.bp)
     app.register_blueprint(placelist.bp)
     app.register_blueprint(protein.bp)
     app.register_blueprint(redirects.bp)
     app.register_blueprint(special_announcement.bp)
     app.register_blueprint(topic_page.bp)
     from routes.api import (protein as protein_api)
+    from routes.api import (disease as disease_api)
```
shifucun, line 60: should remove this since there is no "routes.api.disease" file yet. Then make sure the test passed locally by running `./run_test.sh -p` and `./run_test.sh -c`

suhana13: Removed routes.api.disease and ran the tests!

shifucun (approved):

PR #1651, opened 2022-06-29
[map] fix per capita legend bug

chejennifer:
- fixed bug where legend calculation did not take the correct min and max resulting in 0 contrast in the map.

original: 
Screen Shot 2022-06-29 at 4 20 04 PM
fixed:
Screen Shot 2022-06-29 at 4 18 55 PM


shifucun (approved):
Thanks for the quick fix!

PR #1654, opened 2022-07-01
feat: pipe in real protein data

suhana13:
I added the .py script for diseases under the api directory and I also added more frontend framework. The config for diseases is ready too, I just have make a CL request for the config and require more guidance on that @shifucun!
It'll be great if you could review it please!

### DIFF
--- a/static/js/disease/chart.ts
+++ b/static/js/disease/chart.ts
@@ -18,10 +18,13 @@ import _ from "lodash";
 const SVGNS = "http://www.w3.org/2000/svg";
 const XLINKNS = "http://www.w3.org/1999/xlink";
 export interface DiseaseGeneAssociationData {
-    name: string;
-    score: number;
+  name: string;
+  score: number;
 }
 
-export function drawDiseaseGeneAssocChart(id:string, data: DiseaseGeneAssociationData[]): void {
-    return;
+export function drawDiseaseGeneAssocChart(
+  id: string,
+  data: DiseaseGeneAssociationData[]
+): void {
+  return;
 }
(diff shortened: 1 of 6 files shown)


shifucun (approved):

static/js/protein/page.tsx
```diff
@@ -44,12 +44,12 @@ import {
   getVarSigAssoc,
   getVarTypeAssoc,
 } from "./data_processing_utils";
-interface PagePropType {
+export interface PagePropType {
```
chejennifer, line 47: I don't think we should use the same PageProp and PageState for both Protein browser and Disease browser

static/js/disease/chart.ts
```diff
@@ -0,0 +1,23 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import * as d3 from "d3";
+import _ from "lodash";
+const SVGNS = "http://www.w3.org/2000/svg";
+const XLINKNS = "http://www.w3.org/1999/xlink";
+
+export function drawDiseaseGeneAssocChart(id:string, data: {name:string, score: number}): void {
```
chejennifer, line 21: we should make {name: string, score:number} a type

chejennifer (commented):
nice! Just some small things

static/js/protein/page.tsx
suhana13: So, do we use the same PageProp and PageState definitions but define them in different files without exporting them?

static/js/disease/chart.ts
suhana13: Done!

static/js/protein/page.tsx
chejennifer: Yes, because we want to be able to update the prop type and state types of each component without having to update the other component

chejennifer (approved):
thanks for updating!

PR #1655, opened 2022-07-08
[protein browser] Tooltips for protein-protein interaction graph, highlight on hover

rogershijin:
- Adds tooltips to nodes and links of protein-protein interaction graph (in shameless thievery of @suhana13''s code)
- Adds highlight on hover to all elements with tooltips (thanks to @chejennifer's advice to use element IDs instead of `d3.select(this)`)
- Factors both of these behaviors into a single handler `handleMouseEvents` shared by all charts

@chejennifer unfortunately this goes against your advice of not writing a mouse handler interface that involves callbacks (so I may have written not-so-readable code). Since I factored out the entire `.on("mouseover", ...).on("mousemove", ...).on("mouseout", ...)` chain into `handleMouseEvents` it was easier for me to think in terms of callbacks but please let me know how I can clean this up.

image

I couldn't figure out how to screenshot hover behavior so I ended up with a screenshot with the tooltip but without the highlight on hover.


static/js/protein/chart.ts
```diff
@@ -242,34 +258,117 @@ const LINK_STYLE = {
   },
 };
 
-// https://stackoverflow.com/a/69610045
-function brighten(): void {
-  // brighten object under cursor
-  d3.select(this).style("filter", `brightness(${BRIGHTEN_PERCENTAGE})`);
+/**
```
chejennifer, line 261: nit: I think all of these helper functions: brighten, unbrighten, updateToolTipText, updateToolTipPosition, showTooltip and hideTooltip are not necessary because:
1. they're each only used in one place
2. they all just pass parameters to one other function and don't really contain any logic

static/js/protein/chart.ts
```diff
@@ -73,6 +74,13 @@ export interface VarGeneDataPoint {
   upper: number;
 }
 
+type Selectable = Element | EnterElement | Document | Window | SVGLineElement;
```
chejennifer, line 77: Is this necessary? I see it's used when you pass a selection to handleMouseEvents, but aren't all the selections going to be of type `Element`?

static/js/protein/chart.ts
```diff
@@ -73,6 +74,13 @@ export interface VarGeneDataPoint {
...
+type Datum =
+  | ProteinNumData
```
chejennifer, line 79: don't need that firs `|`

static/js/protein/chart.ts
```diff
@@ -242,34 +258,117 @@ const LINK_STYLE = {
...
+      onMouseOver(idFunc(i), toolTipFunc(d), brightenPercentage);
+    })
+    .on("mousemove", onMouseMove)
+    .on("mouseout", (d, i) => onMouseOut(idFunc(i)));
+}
+
+/**
+ * Get function that takes an index and returns an ID containing the index and chart ID.
+ */
+function getBarIDFunc(chartID: string): (index: number) => string {
```
chejennifer, line 368: I think instead of making this specific to bar, this function can just return `${chartID}-${index}` or `{chartID}-datum-${index}` or something general so it can be used for all the charts

chejennifer (commented):
Awesome job with the refactoring!

static/js/protein/chart.ts
rogershijin: Makes sense, sounds good!

static/js/protein/chart.ts
rogershijin: Oh yea I couldn't figure this one out - if I remove any of these I get an opaque error of the form

```
[tsl] ERROR in /home/rogershijin/project/datacommons-website/static/js/protein/chart.ts(838,7)
      TS2345: Argument of type '(selection: Selection<Element, Datum, SVGGElement, unknown>, idFunc: (index: number) => string, toolTipFunc: (datum: Datum) => string, brightenPercentage?: string) => void' is not assignable to parameter of type '(selection: Selection<SVGLineElement | BaseType, InteractionLink, SVGGElement, unknown>, ...args: any[]) => void'.
  Types of parameters 'selection' and 'selection' are incompatible.
    Type 'Selection<SVGLineElement | BaseType, InteractionLink, SVGGElement, unknown>' is not assignable to type 'Selection<Element, Datum, SVGGElement, unknown>'.
      Type 'SVGLineElement | BaseType' is not assignable to type 'Element'.
        Type 'Window' is missing the following properties from type 'Element': attributes, classList, className, clientHeight, and 154 more.
```

The current hack uses the fact that a [union of types only guarantees the intersection of their properties](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#union-types) to deal with this, but perhaps there's a much more straightforward way to do this.  Any tips would be greatly appreciated.

static/js/protein/chart.ts
rogershijin: Oh yes!! This bothers me too but I think the linter puts it there automatically so if we took it out every time somebody runs linter on the codebase it would get reinserted.

static/js/protein/chart.ts
rogershijin: Great suggestion! This takes care of my manual `circleIDFunc` now too!

```const circleIDFunc = getElementIDFunc(id, 'circle');```

File: static/js/protein/chart.ts [hunk 5]
```diff
@@ -242,34 +258,77 @@ const LINK_STYLE = {
...
+ * When mouse first enters element specified by given id, brighten it and update/display the global tooltip.
+ */
+function onMouseOver(
+  elementID: string,
+  toolTipText: string,
+  brightenPercentage: string = DEFAULT_BRIGHTEN_PERCENTAGE
+): void {
+  // brighten element: https://stackoverflow.com/a/69610045
+  d3.select(`#${elementID}`).style(
+    "filter",
```
suhana13, line 271: Thank you so much for adding the brightness component. It really enhances the aesthetics of the chart. I was wondering if we'd wanna add the style elements to the protein.scss file ?

suhana13 (approved):
The code looks more organized and so much cleaner. Thank you very much for the awesome changes @rogershijin!

static/js/protein/chart.ts
chejennifer: usually we do want to add styling to the scss file, however, since the brightenPercentage is an argument that gets passed depending on the chart and you can't pass arguments to scss, I think it's easier to set the style here.

rogershijin: Oh! Good point! I'm not sure about which is more kosher - so on the one hand the brightness percentage is a variable which motivates handling this in `chart.ts`, but on the other hand if the number of brightness values we want to support is small we could just enumerate all of them in a few CSS classes.  @chejennifer thoughts on this?

static/js/protein/chart.ts
rogershijin: Wow this works!!! Brilliant!

I also experimented with the below things which failed - I would've never thought to modify the latter 3 arguments without your example.

```
d3.Selection<SVGElement, Datum, SVGGElement, unknown>
d3.Selection<SVGElement | SVGLineElement, Datum, SVGGElement, unknown>
d3.Selection<Element | SVGElement | SVGLineElement, Datum, SVGGElement, unknown>
```

static/js/protein/chart.ts
rogershijin: Oops sorry for not refreshing/thanks for answering my question before I asked it!

chejennifer (approved):
Thanks for the changes!

PR #1659, opened 2022-07-12
fix broken webdriver test

chejennifer:
- Webdriver test failure was caused by change in stat var name. Make webdriver test more reliable by finding element based on ID.
- Also realized the test is named "check statvar and uncheck" but the test only tested for checking statvar and not unchecking statvar, so added that part


shifucun:
Maybe this is an example of web drive finding real data issue. I am debating weather to still let it catch the error..

chejennifer:
> Maybe this is an example of web drive finding real data issue. I am debating weather to still let it catch the error..

There is a TAP presubmit test that should've caught this data issue so I don't think webdriver should be used for this?

shifucun:
Ok, then we can let this focus on UI behavior. But should make it required...

shifucun (approved):

chejennifer:
> Ok, then we can let this focus on UI behavior. But should make it required...

Should make what required?

PR #1660, opened 2022-07-13
[Protein Browser] Increase PPI graph depth - Part 1: Utils

rogershijin:
This is the first in a series of 3-4 PRs to increase the depth of the protein-protein interaction graph.  I've taken out the PPI graph from the protein browser page temporarily so that these piece-wise changes will not break the page.  All of the important changes are in `data_processing_utils.ts` except one (`GraphDataNested` in `chart.ts`).

- Typing for V1 responses: `V1BaseDatum`, `V1ResponseDatum<T>`, `V1Response<T>`.  @shifucun as the creator of the V1 API, would you mind critiquing these?  (Would greatly appreciate your feedback on the rest of the code too)
- A type `GraphDataNested` which represents a graph as two parallel depthwise nested lists of nodes and links for ease of increasing/decreasing depth (in `chart.ts`).
- General utility functions `objectFromArray` and `zip` inspired by python
- DCID and DCID list manipulation functions `idFromDCID`, `dcidFromID`, `proteinsFromInteractionID`, `sortInteractionDCID`, `deduplicateInteractionDCIDs`, `getInteractionTarget`, `quantityFromDCID`
- Lightweight response data extraction functions `getFromResponse`, `valuesFromResponse`, and `dcidsFromResponse`, which map a response to a list of desired data.
- The heavier response processing function `scoreDataFromResponse` and its helper `symmetricScoreRec`, which together construct a score record mapping interaction IDs to their confidence scores.

Sorry for the extreme slowness on this - I think I find types in typescript more difficult than most people and would greatly appreciate your advice and feedback.  Also if this is still too much code I can split further.

Thanks for the review!


rogershijin:
Will revert mixer soon.

static/js/protein/data_processing_utils.ts
```diff
@@ -592,3 +537,319 @@ export function getProteinDescription(data: GraphNodes): string {
...
+  sourceDCID: string,
+  returnDCID = false
+): string {
+  // note this also works in the case of a self-interaction
+  const interactionID = idFromDCID(interactionDCID);
+  const sourceID = idFromDCID(sourceDCID);
+  const id = interactionID
+    .replace(`${sourceID}_`, "")
+    .replace(`_${sourceID}`, "");
+  if (returnDCID) return dcidFromID(id);
```
rogershijin, line 660: For ifs with 1-line bodies, do you guys prefer 1-line or multiple lines?

static/js/protein/data_processing_utils.ts
```diff
@@ -592,3 +537,319 @@ export function getProteinDescription(data: GraphNodes): string {
...
+  ) as string[];
+}
+
+/**
+ * Given interaction DCID and source DCID, infer and return target ID (or target DCID if returnDCID is true).
+ */
+export function getInteractionTarget(
+  interactionDCID: string,
+  sourceDCID: string,
+  returnDCID = false
```
rogershijin, line 652: Do you guys generally put type annotations on boolean flags? The linter wants this format while codacy wants `returnDCID: boolean = false`.

static/js/protein/chart.ts
```diff
@@ -58,11 +58,21 @@ export interface InteractionLink extends SimulationLinkDatum<ProteinNode> {
...
 export interface InteractionGraphData {
   nodeData: ProteinNode[];
   linkData: InteractionLink[];
 }
 
+// Represents interaction graph depthwise as two parallel nested lists.  For each index i,
+// nodeDataNested[i] will contain the nodes added in i iterations of BFS,
+// and linkDataNested[i] will contain the links added in i iterations of BFS,
+// along with any links between nodesDataNested[i] and existing nodes in the graph.
+export interface InteractionGraphDataNested {
```
rogershijin, line 71: Is `InteractionGraphDataNested` a good name for this? I was also considering `InteractionGraphDataDepthwise`.

static/js/protein/data_processing_utils.ts
```diff
@@ -12,7 +13,39 @@ import { ProteinVarType } from "./page";
 import { InteractingProteinType } from "./page";
 import { DiseaseAssociationType } from "./page";
 
-const MAX_INTERACTIONS = 10; // upper bound on node degree in interaction graph viz's
+// Base type of "values" value of objects stored in <V1 response>.data.data
+type BaseDCDataType = {
```
rogershijin, line 17: @shifucun all of these properties guaranteed in every response?

static/js/protein/data_processing_utils.ts
```diff
@@ -592,3 +537,319 @@ export function getProteinDescription(data: GraphNodes): string {
...
+ */
+export function scoreDataFromResponse(
+  scoreResponse: V1Response<BaseDCDataType>
+): Record<string, number> {
+  const scoreValues = valuesFromResponse(scoreResponse);
+  const interactionDCIDs = dcidsFromResponse(scoreResponse);
+  const scoreList = scoreValues
+    // filter results for IntactMiScore - there should be 0 or 1 match
+    .map((scoreRecList: BaseDCDataType[] | undefined) => {
+      if (scoreRecList !== undefined) {
```
rogershijin, line 842: Codacy doesn't like this check for if `scoreRecList` is `undefined`.  What's the proper way to do things?  Thanks!

static/js/protein/data_processing_utils.ts
beets: multiple lines please, for readability.

static/js/protein/data_processing_utils.ts
beets: prefer linter over codacy.

static/js/protein/data_processing_utils.ts
```diff
@@ -592,3 +537,324 @@ export function getProteinDescription(data: GraphNodes): string {
...
+    .replace(`${sourceID}_`, "")
+    .replace(`_${sourceID}`, "");
+  if (returnDCID) return dcidFromID(id);
+  return id;
+}
+
+/**
+ * Given id of the form {protein id}_{species id} (e.g. P53_HUMAN), parse into and return ProteinNode
+ */
+export function nodeFromID(
```
rogershijin, line 671: Previously checked in (moving to resolve codacy error)

static/js/protein/data_processing_utils.ts
```diff
@@ -592,3 +537,324 @@ export function getProteinDescription(data: GraphNodes): string {
...
+        ],
+
+        linkData : [
+          { source: MECOM_HUMAN, target: CTPB1_HUMAN, score: 0.3 },
+          { source: MECOM_HUMAN, target: SUPT16H_HUMAN, score: 0.7 },
+        ],
+
+      }.
+ */
+export function getProteinInteractionGraphData(
```
rogershijin, line 703: Previously checked in (moving to resolve codacy error)

static/js/protein/data_processing_utils.ts
rogershijin: got it, thanks!

static/js/protein/data_processing_utils.ts
rogershijin: got it, thanks!

static/js/protein/data_processing_utils.ts
```diff
@@ -592,3 +537,324 @@ export function getProteinDescription(data: GraphNodes): string {
...
+
+  const linkData: InteractionLink[] = neighbors.map((node) => {
+    return {
+      score: node.value,
+      source: centerNodeID,
+      target: node.id,
+    };
+  });
+
+  return {
```
rogershijin, line 748: Only big change is return type is now `InteractionGraphDataNested` instead of `InteractionGraphData`

static/js/protein/data_processing_utils.ts
```diff
@@ -12,7 +13,32 @@ import { ProteinVarType } from "./page";
...
 import { DiseaseAssociationType } from "./page";
 
-const MAX_INTERACTIONS = 10; // upper bound on node degree in interaction graph viz's
+// Base type of "values" value of objects stored in <V1 response>.data.data
+type V1BaseDatum = {
+  dcid: string;
+  name: string;
+  provenanceId: string;
+  types: string[];
+};
```
rogershijin, line 22: @shifucun Should some of these properties be optional? Thanks!

static/js/protein/data_processing_utils.ts
```diff
@@ -592,3 +530,328 @@ export function getProteinDescription(data: GraphNodes): string {
...
+ *
+ * Adapted from implementation by Matt Mancini
+ *
+ * Thanks also to Carlos Israel Ortiz Garcia and Jack Davison for the below references:
+ *  - https://ncjamieson.com/typing-zipwith/#typing-zip
+ *  - https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-0.html#variadic-tuple-types
+ */
+// Recursive elementwise type inference on array of arrays
+type Zip<T> = T extends [Array<infer U>, ...infer Z] ? [U, ...Zip<Z>] : [];
+export function zip<T extends unknown[][]>(...arrays: T): Array<Zip<T>> {
```
chejennifer, line 569: i don't think you need to create your own zip method since lodash has one: https://lodash.com/docs/4.17.15#zip

static/js/protein/data_processing_utils.ts
chejennifer: you could just check `!_.isEmpty(scoreRecList)`

static/js/protein/chart.ts
rogershijin: Got it that's a great name! And yeah `InteractionGraphData` is the input type to the draw function.

static/js/protein/data_processing_utils.ts
rogershijin: Thanks! I'll do this

static/js/protein/chart.ts
```diff
@@ -783,103 +793,103 @@ export function drawProteinInteractionChart(
...
-    .attr("fill", (node) => nodeColors(node.depth))
-    .attr("id", (node, i) => nodeIDFunc(i));
-
-  // node labels
-  nodes
-    .append("text")
-    .text(({ name }) => name)
-    .attr("class", "protein-node-label");
-}
+// export function drawProteinInteractionGraph(
```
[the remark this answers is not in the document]
rogershijin: Oh that's right I think I was getting type errors I was too lazy to deal with when I first commented this out but perhaps now that the code is self-consistent again they'll go away.

static/js/protein/data_processing_utils.ts
```diff
@@ -12,7 +13,32 @@ import { ProteinVarType } from "./page";
...
+// Base type of "values" value of objects stored in <V1 response>.data.data
+type V1BaseDatum = {
+  dcid: string;
+  name: string;
+  provenanceId: string;
+  types: string[];
+};
+
+// Generic for objects stored in <V1 response>.data.data
+type V1ResponseDatum<ValueDatum extends V1BaseDatum> = {
```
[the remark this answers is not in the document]
rogershijin: Oh yeah I think I need to ask @shifucun about this.  It might be that everything returns stuff of type `V1BaseDatum` and we don't need a generic.

static/js/protein/data_processing_utils.ts
rogershijin: Oh oops thanks didn't know! Will just delete this and use lodash's then unless I find an example for which this does better type inference than lodash.

static/js/protein/data_processing_utils.ts
```diff
@@ -592,3 +537,324 @@ export function getProteinDescription(data: GraphNodes): string {
...
+
+/**
+ * Given interaction DCID of the form bio/{protein1}_{protein2}, return [protein1, protein2]
+ */
+export function proteinsFromInteractionDCID(interactionDCID: string): string[] {
+  const id = idFromDCID(interactionDCID);
+  // danger: assumes neither {protein id}, {species id} contain an underscore
+  const split = id.split("_");
+  if (split.length !== 4) {
+    throw `Unsupported DCID ${interactionDCID}`;
```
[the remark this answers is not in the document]
rogershijin: Oh yeah failing loudly might not be the best I'll just do a `console.assert`.

static/js/protein/data_processing_utils.ts
```diff
@@ -592,3 +537,324 @@ export function getProteinDescription(data: GraphNodes): string {
...
+ * keep only one.
+ */
+export function deduplicateInteractionDCIDs(
+  interactionDCIDs: string[]
+): string[] {
+  // Can't just use Set here because not guaranteed that A_B in list implies B_A in list,
+  // so we need to keep track of which order is the true DCID corresponding to an entity in the response.
+  // Counterexample: http://datacommons.org/browser/bio/KLOTB_HUMAN
+  return Object.values(
+    objectFromArray(interactionDCIDs, sortInteractionDCID, (dcid) => dcid)
```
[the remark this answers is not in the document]
rogershijin: I think python has brainwashed me into not finding your way simpler but I trust your judgment/will implement this insteaad.

static/js/protein/data_processing_utils.ts
```diff
@@ -592,3 +537,324 @@ export function getProteinDescription(data: GraphNodes): string {
...
+ * Given an array of interaction DCIDs and a parallel array of corresponding scores,
+ * construct and return score record such that for each interaction A_B with corresponding DCID in interaction DCIDs,
+ * both A_B and B_A map to the score of A_B.
+ */
+export function symmetricScoreRec(
+  interactionDCIDs: string[],
+  scores: number[]
+): Record<string, number> {
+  const scoreRec = {};
+  zip(interactionDCIDs, scores).forEach(([dcid, score]) => {
```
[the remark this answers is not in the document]
rogershijin: Sorry yes totally unnecessary! I think this is a bad combo of when I code in js I sometimes forget for loops exist/my python habits making me allergic to array indexing.

Will fix thanks for pointing this out!

static/js/protein/data_processing_utils.ts
```diff
@@ -592,3 +537,328 @@ export function getProteinDescription(data: GraphNodes): string {
...
+    linkDataNested: [[], linkData],
+    nodeDataNested: [[centerDatum], neighbors],
+  };
+}
+
+/**
+ * Given an object mapping interaction IDs to confidence scores and the DCIDs of two proteins, return the score
+ * of the interaction between the two proteins if it appears in the object, or the default score otherwise.
+ */
+export function scoreFromProteinDCIDs(
```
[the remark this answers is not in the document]
rogershijin: Hmm I'm still ironing out the refactor of the downstream things but I currently believe it'll only be used once so it can be taken out.

static/js/protein/data_processing_utils.ts
```diff
@@ -592,3 +537,328 @@ export function getProteinDescription(data: GraphNodes): string {
...
+  T extends BaseDCDataType,
+  K extends keyof V1ResponseDatum<T>
+>(resp: V1Response<T>, key: K): V1ResponseDatum<T>[K][] {
+  if (!("data" in resp.data)) {
+    return [];
+  }
+  return resp.data.data.map((obj) => obj[key]);
+}
+
+// Useful special cases of getFromResponse
```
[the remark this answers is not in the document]
rogershijin: Thanks noted! I'll replace with the calls to the general function.

static/js/protein/data_processing_utils.ts
```diff
@@ -592,3 +530,328 @@ export function getProteinDescription(data: GraphNodes): string {
...
+ * satisfying the property that if A_B: A_B.score is in the record, then B_A: A_B.score is also.
+ */
+export function scoreDataFromResponse(
+  scoreResponse: V1Response<V1BaseDatum>
+): Record<string, number> {
+  const scoreValues = valuesFromResponse(scoreResponse);
+  const interactionDCIDs = dcidsFromResponse(scoreResponse);
+  const scoreList = scoreValues
+    // filter results for IntactMiScore - there should be 0 or 1 match
+    .map((scoreRecList: V1BaseDatum[] | undefined) => {
```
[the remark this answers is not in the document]
rogershijin: Whoa keen eye thanks for catching! Yes I also think it can be done in one.

static/js/protein/data_processing_utils.ts
```diff
@@ -592,3 +530,328 @@ export function getProteinDescription(data: GraphNodes): string {
...
+  return Array.from(Array(minLength).keys()).map((_, i) => {
+    return arrays.map((array) => array[i]) as Zip<T>;
+  });
+}
+
+/**
+ * Convert DCID of form bio/<id> to id.
+ * Note: probably should be doing dcid: `bio/${string}`, but currently causes many type errors
+ */
+export function idFromDCID(dcid: string): string {
```
beets, line 580: Since there are so many ID types floating around in this module -- perhaps specify that this is the `proteinId`? I see places here where this sometimes include species -- specificity will go a long way towards reducing errors.

Also - only export functions that will be used outside the module (applies to everything else in here).

static/js/protein/data_processing_utils.ts
```diff
@@ -592,3 +530,328 @@ export function getProteinDescription(data: GraphNodes): string {
...
+  if (!quantityDCID.includes(quantityName)) {
+    throw `DCID "${quantityDCID}" does not contain quantity name "${quantityName}"`;
+  }
+  return Number(_.last(quantityDCID.split(quantityName)));
+}
+
+/**
+ * Given interaction DCID of the form bio/{protein1}_{protein2}, return [protein1, protein2]
+ */
+export function proteinsFromInteractionDCID(interactionDCID: string): string[] {
```
beets, line 607: Do we get access to the underlying nodes for these? That can avoid the assumptions for the structure of these dcids.

Would be helpful to have an example in the comments here - e.g. bio/2A5D_YEAST_AHA1_YEAST

static/js/protein/data_processing_utils.ts
```diff
@@ -592,3 +530,328 @@ export function getProteinDescription(data: GraphNodes): string {
...
+    throw `Unsupported DCID ${interactionDCID}`;
+  }
+  return [`${split[0]}_${split[1]}`, `${split[2]}_${split[3]}`];
+}
+
+/**
+ * Given interactionDCID of the form {protein 1}_{protein 2}, return ID of the same form
+ * but with protein 1, protein 2 in alpha order.
+ */
+function sortInteractionDCID(interactionDCID: string): string {
```
beets, line 621: Since this is only used for the following function - this could return a `proteins` to avoid the string munging.

beets (commented):
I feel the code will be easier if this was encapsulated in a component.. then you can store some of the intermediate data and avoid so many string operations.

please add tests for the helpers (or please be sure to do so in follow ups, since the code might change after the series of PRs).

thanks for the detailed comments in the code. i also recognize the amount of time you must have spent figuring out all the typing issues and typescript <> python translations. this is part of the reason smaller, earlier pr's are helpful so we can align on coding practices.

static/js/protein/chart.ts
rogershijin: Oh right BFS is just an implementation detail I'll remove it from the description.  And got it I'll collect all the types I use in a single file then.  

Unfortunately I'm doing BFS from the client right now.  I'll PR this soon for you guys to look at so we can decide what should be moved to flask.

static/js/protein/data_processing_utils.ts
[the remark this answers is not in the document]
rogershijin: Sorry about the helpers guys I think I have bad intuitions about what other people find readable.  This is very useful feedback.  

I'll reimplement with a loop and not try to be fancy.

static/js/protein/data_processing_utils.ts
```diff
@@ -170,94 +196,6 @@ export function getProteinInteraction(
   return [];
 }
 
-/**
- * Given id of the form {protein id}_{species id} (e.g. P53_HUMAN), parse into and return ProteinNode
- */
-function nodeFromID(proteinSpeciesID: string, depth: number): ProteinNode {
```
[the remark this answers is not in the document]
rogershijin: yeah these were just to move this function and the one after - sorry didn't know it would result in a confusing PR view

static/js/protein/data_processing_utils.ts
```diff
@@ -592,3 +530,328 @@ export function getProteinDescription(data: GraphNodes): string {
...
+ * Given id, convert to DCID of form bio/<id>
+ */
+export function dcidFromID(id: string): string {
+  return `bio/${id}`;
+}
+
+/**
+ * Given quantity DCID of the form "<quantityName><quantityValue>", extract and return quantityValue.
+ */
+export function quantityFromDCID(
```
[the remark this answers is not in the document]
rogershijin: great name will adopt!

static/js/protein/data_processing_utils.ts
rogershijin: Oh ok thanks for pointing this out - the only two ID types for the data of the interaction graph which are true IDs (i.e. used to ID nodes/links in the graph) are protein ID and interaction ID, and `idFromDCID` and `dcidFromID` apply to both.  However in `chart.ts` there's stuff like `elementID`, `barIDFunc`, `circleIDFunc`.  Perhaps I could make `idFromDCID` and `dcidFromID` static methods in the new interaction graph component?

Also I think I wrote species ID in a comment somewhere without realizing it was confusing - I'll just revise the comments so there's no mention of IDs in the graph part other than protein ID and interaction ID.

static/js/protein/data_processing_utils.ts
rogershijin: I've asked Bo a similar question before and I think this would require another hop in the knowledge graph (which from the clientside would be another request).  Migrating to flask and using the [interactionProtein](https://datacommons.org/browser/bio/1433G_HUMAN_P53_HUMAN) property would be much cleaner than what I'm currently doing.  

This might be overkill but another request might not be too bad because it could be sent in parallel with the request for interaction scores. `interactinType` and `interactionDetectionMethod` might also be useful to pull. (but yes this would all be better in flask)

static/js/protein/data_processing_utils.ts
rogershijin: Oh do you mean something like `[P53_HUMAN, 1433G_HUMAN]`?  If so I think we can delete this function and add a `sorted` flag to `proteinsFromInteractionDCID`.

static/js/protein/data_processing_utils.ts
```diff
@@ -1,8 +1,9 @@
+import { AxiosResponse } from "axios";
```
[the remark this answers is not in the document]
rogershijin: Thank you so much for the review! Yes will do.

static/js/protein/data_processing_utils.ts
rogershijin: I'm sure you've experienced more weird null cases than I have in your lifetime, so I'm gonna add it!

static/js/protein/data_processing_utils.ts
rogershijin: I renamed these functions to `ppiIDFromDCID` and `ppiDCIDFromID` and replaced all mentions of `species ID` with `species name`.  

To make this a bit safer I also added an assertion in `ppiIDFromDCID` to enforce that the input cannot start with `"bio/"`, and a new type `bioDCID = `bio/${string}` to enforce that all variables labeled DCID throughout the code start with `"bio/"`.  (please let me know if you think this is overkill I can revert back to just string type for everything).

static/js/protein/data_processing_utils.ts
rogershijin: I ended up deleting this helper without adding the `sorted` flag because all use cases so far would work with `sorted=False`.

static/js/protein/chart.ts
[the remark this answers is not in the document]
rogershijin: Yup you're right I was able to uncomment with a minor tweak to the input type.

static/js/protein/data_processing_utils.ts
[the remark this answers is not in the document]
rogershijin: Deleted!

beets (commented):
thanks for the updates -- there's a huge improvement in readability. a few more comments / questions below.

static/js/protein/data_processing_utils.ts
```diff
@@ -1,18 +1,44 @@
+/**
+ * Copyright 2021 Google LLC
```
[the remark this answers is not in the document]
rogershijin: Thanks for reviewing again!! And for eagle-eyeing this

static/js/protein/data_processing_utils.ts
```diff
@@ -592,3 +530,263 @@ export function getProteinDescription(data: GraphNodes): string {
...
+  return arr.reduce(
+    (obj, elt) => Object.assign(obj, { [keyFunc(elt)]: valFunc(elt) }),
+    {}
+  );
+}
+
+/**
+ * Convert DCID of form bio/<id> to id. Utility for protein-protein interaction (PPI) graph.
+ */
+export function ppiIDFromDCID(dcid: bioDCID): string {
```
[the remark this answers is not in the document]
rogershijin: got it! will update everywhere!

static/js/protein/data_processing_utils.ts
```diff
@@ -592,3 +530,263 @@ export function getProteinDescription(data: GraphNodes): string {
...
+  return dcid.replace("bio/", "");
+}
+
+/**
+ * Given id, convert to DCID of form bio/<id>.  Utility for protein-protein interaction (PPI) graph.
+ */
+export function ppiDCIDFromID(id: string): bioDCID {
```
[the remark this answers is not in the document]
rogershijin: expunged!

static/js/protein/data_processing_utils.ts
```diff
@@ -592,3 +530,263 @@ export function getProteinDescription(data: GraphNodes): string {
...
+
+/**
+ * Given quantity DCID of the form "<quantityName><quantityValue>", extract and return quantityValue.
+ */
+export function quantityFromDCID(
+  quantityDCID: string,
+  quantityName: string
+): number {
+  if (!quantityDCID.includes(quantityName)) {
+    throw `DCID "${quantityDCID}" does not contain quantity name "${quantityName}"`;
```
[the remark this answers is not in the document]
rogershijin: got it! returned `NaN`!

static/js/protein/data_processing_utils.ts
```diff
@@ -592,3 +530,263 @@ export function getProteinDescription(data: GraphNodes): string {
...
+  if (_.isEmpty(data)) {
+    return;
+  }
+
+  // P53_HUMAN is central protein in below examples.
+  // take interaction names of the form P53_HUMAN_ASPP2_HUMAN | ASPP2_HUMAN_P53_HUMAN and parse into ASPP2_HUMAN.
+  const centerNodeID = data[0].parent;
+  let neighbors = data.map(({ name: interactionID, value }) => {
+    // value is confidenceScore
+    console.log("id", interactionID);
```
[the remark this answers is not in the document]
rogershijin: sorry about this I've become sloppy

static/js/protein/data_processing_utils.ts
```diff
@@ -592,3 +530,263 @@ export function getProteinDescription(data: GraphNodes): string {
...
+  if (!("data" in resp.data)) {
+    return [];
+  }
+  return resp.data.data.map((obj) => obj[key]);
+}
+
+/**
+ * Given an array of interaction DCIDs and a parallel array of corresponding scores,
+ * construct and return score record such that for each interaction A_B with corresponding DCID in interaction DCIDs,
+ * both A_B and B_A map to the score of A_B.
```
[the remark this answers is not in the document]
rogershijin: Oh good question!  I'm not sure this is the right thing to do, but we currently do not graphically distinguish between the A_B and B_A scores, and Sam has said that A_B.score should equal B_A.score.  But in the future perhaps we should make the graph directed to make relationships like "A phosphorylates B" more clear.  I also think there should be assays for which the scores are not symmetric - ex. "knockout A and see what happens to the expression of B".  These might not exist in the graph currently but could very well be imported in the future.

I think the treatment of A_B vs. A_B will require more careful thinking later on, but for now I'll add something which keeps the higher of the two scores.

static/js/protein/types.ts
```diff
@@ -0,0 +1,79 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { AxiosResponse } from "axios";
+import { SimulationLinkDatum, SimulationNodeDatum } from "d3";
+
+export type bioDCID = `bio/${string}`;
```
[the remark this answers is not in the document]
rogershijin: Since this applies to all dcids across biomedical data commons, how about "biomedDcid"? (or if we want something specific to the interaction graph, we could do "bioNodeOrInteractionDcid")

static/js/protein/types.ts
```diff
@@ -0,0 +1,79 @@
...
+
+// interfaces for protein-protein interaction graph
+
+export interface Node {
```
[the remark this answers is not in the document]
rogershijin: Oh this was intended to be a broad interface (`ProteinNode` below extends it).  Some other proposed projects on datacommons (drug browser/Crystal's project) might use parts of this code for graph viz so wanted to eventually factor out the graph stuff to an interface totally independent of protein stuff, and make proteins a special case.  What do you think?

static/js/protein/types.ts
```diff
@@ -0,0 +1,80 @@
...
+  types?: string[];
+  value?: string | number;
+}
+
+export interface V1BioDatum extends V1BaseDatum {
+  dcid: bioDcid;
+  name: string;
+  provenanceId: string;
+  types: string[];
+}
```
rogershijin, line 68: I was really tempted to do 

```
export interface V1BioDatum extends Omit<Required<V1BaseDatum>, "value"> {
  dcid: bioDcid;
}
```

here to avoid copying

```
  name: string;
  provenanceId: string;
  types: string[];
```

but held back for the sake of readability.  Hope the reviewers are proud :p

rogershijin:
ptal! Changes:
- IDs/DCIDs are now properly camelCased to Ids/Dcids
- `symmetricScoreRec` now assigns `max(A_B.score, B_A.score)` to both `A_B` and `B_A`
- type `V1BaseDatum` now has an optional `value` property (thanks @shifucun for comments and [this example](https://autopush.api.datacommons.org/v1/property/out/name/values/country/USA)!)
- `quantityFromDcid` now returns `NaN` when it fails to parse.

static/js/protein/data_processing_utils.ts
```diff
@@ -12,7 +13,32 @@ import { ProteinVarType } from "./page";
...
+
+// Generic for objects stored in <V1 response>.data.data
+type V1ResponseDatum<ValueDatum extends V1BaseDatum> = {
+  entity: string;
+  values: ValueDatum[];
+};
+
+// Generic for V1 response
+// Reference: https://github.com/axios/axios/blob/7d6bddba2d8de29c263feaef4c40daa50cb4b176/index.d.ts#L83
+type V1Response<ValueDatum extends V1BaseDatum> = AxiosResponse<{
```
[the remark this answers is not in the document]
chejennifer: Yes, but you can just make the type 
```
interface V1Response {
data: ...
}
```
and functions that take in the V1 response would have an argument of type `V1Response` and where these functions are called, just pass `resp.data`?

chejennifer (commented):
thanks for the updates

static/js/protein/data_processing_utils.ts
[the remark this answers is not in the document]
beets: don't forget to check for NaN downstream (and please document this as a possible value)

static/js/protein/data_processing_utils.ts
```diff
@@ -592,3 +530,254 @@ export function getProteinDescription(data: GraphNodes): string {
...
+}
+
+/**
+ * Given interaction DCID and source DCID, infer and return target ID (or target DCID if returnDCID is true).
+ */
+export function getInteractionTarget(
+  interactionDcid: bioDcid,
+  sourceDcid: bioDcid,
+  returnDcid = false
+): typeof returnDcid extends true ? bioDcid : string {
```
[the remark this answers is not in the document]
rogershijin: yeah that would be cleaner, thanks!  I also realized I never verified that this return type actually works.

static/js/protein/data_processing_utils.ts
```diff
@@ -592,3 +530,254 @@ export function getProteinDescription(data: GraphNodes): string {
...
+ */
+export function ppiIdFromDcid(dcid: bioDcid): string {
+  return dcid.replace("bio/", "");
+}
+
+/**
+ * Given id, convert to DCID of form bio/<id>.  Utility for protein-protein interaction (PPI) graph.
+ */
+export function ppiDcidFromId(id: string): bioDcid {
+  console.assert(!id.includes("bio/"), `Invalid ID: ${id}`);
```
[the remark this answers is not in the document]
rogershijin: I'll check with Sam and get back to you, thanks for the detailed instructions covering what to do in all the cases!

static/js/protein/data_processing_utils.ts
```diff
@@ -592,3 +530,254 @@ export function getProteinDescription(data: GraphNodes): string {
...
+export function deduplicateInteractionDcids(
+  interactionDcids: bioDcid[]
+): bioDcid[] {
+  const uniqueInteractionDcids: bioDcid[] = [];
+  const interactions = new Set<[string, string]>();
+  interactionDcids.forEach((interactionDcid) => {
+    const proteins = proteinsFromInteractionDcid(interactionDcid);
+    if (!interactions.has(proteins)) {
+      uniqueInteractionDcids.push(interactionDcid);
+
```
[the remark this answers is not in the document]
rogershijin: expunged!

static/js/protein/data_processing_utils.ts
```diff
@@ -592,3 +530,254 @@ export function getProteinDescription(data: GraphNodes): string {
...
+ * Given an array of interaction DCIDs and a parallel array of corresponding scores,
+ * construct and return score record such that for each interaction A_B with corresponding DCID in interaction DCIDs,
+ * both A_B and B_A map to the score of A_B.
+ */
+export function symmetricScoreRec(
+  interactionDcids: bioDcid[],
+  scores: number[]
+): Record<string, number> {
+  const scoreRec = {};
+  console.assert(interactionDcids.length === scores.length);
```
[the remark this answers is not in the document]
rogershijin: ok! will take this out

static/js/protein/data_processing_utils.ts
```diff
@@ -592,3 +530,254 @@ export function getProteinDescription(data: GraphNodes): string {
...
+  for (let i = 0; i < interactionDcids.length; i++) {
+    const [protein1, protein2] = proteinsFromInteractionDcid(
+      interactionDcids[i]
+    );
+    const score12 = _.get(
```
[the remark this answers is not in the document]
rogershijin: yea makes sense!

static/js/protein/data_processing_utils.ts
```diff
@@ -592,3 +530,254 @@ export function getProteinDescription(data: GraphNodes): string {
...
+ * satisfying the property that if A_B: A_B.score is in the record, then B_A: A_B.score is also.
+ */
+export function scoreDataFromResponse(
+  scoreResponse: V1Response<V1BioDatum>
+): Record<string, number> {
+  const scoreValues = getFromResponse(scoreResponse, "values");
+  const interactionDcids = getFromResponse(scoreResponse, "entity");
+  const scoreList = scoreValues
+    // filter results for IntactMiScore - there should be 0 or 1 match
+    .map((bioData: V1BioDatum[] | undefined) => {
```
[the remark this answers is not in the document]
rogershijin: got it! will take out

static/js/protein/types.ts
```diff
@@ -0,0 +1,79 @@
...
+
+export interface V1BioDatum extends V1BaseDatum {
+  dcid: bioDcid;
+}
+
+// Generic for objects stored in <V1 response>.data.data
+export type V1ResponseDatum<ValueDatum extends V1BaseDatum> = {
```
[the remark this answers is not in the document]
rogershijin: thanks for thinking carefully about this! I'll check with Bo.

about the naming thing though - I think I'm calling mixer directly so these types were intended just for the mixer response.  is there a requirement for flask endpoints to follow the same or a similar format?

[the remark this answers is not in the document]
chejennifer: I don't think you should be calling mixer directly..  and what format are you referring to?

static/js/protein/data_processing_utils.ts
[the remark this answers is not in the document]
rogershijin: Oh ok sure this works and I like that it is independent of Axios.  I'll make the change!

static/js/protein/types.ts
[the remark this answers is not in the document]
rogershijin: Oh ok perhaps I'm confused.  The root endpoint is https://autopush.api.datacommons.org/v1/bulk . Is this flask?

Also sorry I was referring to the flask response format.  Supposing that I'm calling mixer directly and the types are then for mixer responses, would they also work for flask responses?

static/js/protein/data_processing_utils.ts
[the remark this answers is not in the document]
rogershijin: Ok Sam said it is important to communicate these errors to the user but ideally in the form of a log file they can download.  But for now we'll log these to the console and the user can download the console logs.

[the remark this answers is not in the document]
rogershijin: But I think this location is not an informative place to do it so I'll delete the assert here.

static/js/protein/types.ts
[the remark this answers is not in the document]
rogershijin: Oh Bo agrees with your advice! I also think that extending an interface with only optional properties is no better than extending `{}`.  Each user would still have to make their own subinterfaces which would repeat subsets of the keys.

static/js/protein/data_processing_utils.ts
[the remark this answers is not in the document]
rogershijin: Ok I think I'm also not doing good enough error handling to be able to tell the users I'm doing something, so I'll just take out all of these asserts

static/js/protein/data_processing_utils.ts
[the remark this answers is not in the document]
rogershijin: Sure will add!

static/js/protein/types.ts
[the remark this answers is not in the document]
rogershijin: Cool thanks!

static/js/protein/data_processing_utils.ts
```diff
@@ -592,3 +530,254 @@ export function getProteinDescription(data: GraphNodes): string {
...
+        (bioDatum) =>
+          !_.isEmpty(bioDatum) &&
+          bioDatum.dcid.includes(INTERACTION_QUANTITY_DCID)
+      );
+      if (_.isEmpty(emptyOrInteractionSingleton)) {
+        return DEFAULT_INTERACTION_SCORE;
+      }
+      // if 1 match, return the retrieved IntactMiScore
+      console.assert(emptyOrInteractionSingleton.length === 1);
+      return quantityFromDcid(
```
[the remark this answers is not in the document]
rogershijin: this turned out much much cleaner than what I had! thanks for the simplification!

static/js/protein/types.ts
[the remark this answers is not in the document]
rogershijin: Ah ok got it thanks.  I do hope to migrate a lot of the chained data retrieval/processing to flask eventually.  I guess I'll keep the V1s in the names and then eventually redo/rename these types when I migrate to flask?

rogershijin:
[ptal!] Thanks @chejennifer and @beets for another review! Changes:
- simplified response types following @chejennifer's design - in particular `V1Response` is no longer generic (and has been renamed to `V1BioResponse`).  
- removed `console.assert`s in all but one place where I check for NaN values in `scoreDataFromResponse`
- Updated comments to reflect your suggestions

static/js/protein/data_processing_utils.ts
```diff
@@ -592,3 +529,248 @@ export function getProteinDescription(data: GraphNodes): string {
...
+ * ex. "knockout A and see what happens to the expression of B". These might not exist in the graph currently
+ * but could very well be imported in the future. Thus, we might have two or more edges between proteins
+ * for each type of interaction they participate in, with different scores for each.
+ */
+export function symmetricScoreRec(
+  interactionDcids: bioDcid[],
+  scores: number[]
+): Record<string, number> {
+  const scoreRec = {};
+  for (let i = 0; i < interactionDcids.length; i++) {
```
chejennifer, line 728: in case the two arrays are different lengths, you want the termination condition to be `i < Math.min(interactionDcids.length, scores.length)`

static/js/protein/data_processing_utils.ts
```diff
@@ -592,3 +529,248 @@ export function getProteinDescription(data: GraphNodes): string {
...
+ * Given score response, construct a record mapping interaction IDs of the form A_B to their confidence scores,
+ * satisfying the property that if A_B: A_B.score is in the record, then B_A: A_B.score is also.
+ */
+export function scoreDataFromResponse(
+  scoreResponse: V1BioResponse
+): Record<string, number> {
+  const scoreValues = getFromResponse(scoreResponse, "values");
+  const interactionDcids = getFromResponse(scoreResponse, "entity");
+  const scoreList = scoreValues
+    // map scoreValues to IntactMiScore if available or default interaction score
```
chejennifer, line 759: ".. if available, otherwise map to default interaction score"

static/js/protein/data_processing_utils.ts
```diff
@@ -592,3 +529,248 @@ export function getProteinDescription(data: GraphNodes): string {
...
+      for (const bioDatum of bioData) {
+        if (_.get(bioDatum, "dcid", "").includes(INTERACTION_QUANTITY_DCID)) {
+          const score = quantityFromDcid(
+            bioDatum.dcid,
+            INTERACTION_QUANTITY_DCID
+          );
+          if (!isNaN(score)) {
+            return score;
+          }
+          console.warn(`Invalid quantity ID ${bioDatum.dcid} -- skipping`);
```
chejennifer, line 770: I'm not sure putting the console.warn here makes the most sense either because you're still assigning a score here (DEFAULT_INTERACTION_SCORE), so for a user, it won't actually look like anything has been skipped? 

Maybe the console.warn can be added where this function is called and we can console.warn all cases where the score is DEFAULT_INTERACTION_SCORE & the message can be something like `"Unable to retrieve score for interaction <interaction dcid> -- default interaction score of <interaction score value> used"`

static/js/protein/data_processing_utils.ts
```diff
@@ -592,3 +529,248 @@ export function getProteinDescription(data: GraphNodes): string {
...
+ * Given interaction DCID of the form bio/{protein1}_{protein2} (e.g. bio/2A5D_YEAST_AHA1_YEAST),
+ * return [protein1, protein2]
+ */
+export function proteinsFromInteractionDcid(
+  interactionDcid: bioDcid
+): [string, string] {
+  const id = ppiIdFromDcid(interactionDcid);
+  // danger: assumes neither {protein name}, {species name} contain an underscore
+  const split = id.split("_");
+  return [`${split[0]}_${split[1]}`, `${split[2]}_${split[3]}`];
```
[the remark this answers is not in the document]
rogershijin: got it! will do!

static/js/protein/data_processing_utils.ts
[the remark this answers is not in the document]
rogershijin: oh ahh sorry! I think I did this in the wrong branch.  definitely gone now!

static/js/protein/data_processing_utils.ts
```diff
@@ -592,3 +529,248 @@ export function getProteinDescription(data: GraphNodes): string {
...
+}
+
+/**
+ * Given response and key, map each response datum to value of key and return map
+ */
+export function getFromResponse<K extends keyof V1BioResponseDatum>(
+  resp: V1BioResponse,
+  key: K
+): V1BioResponseDatum[K][] {
+  if (!("data" in resp.data)) {
```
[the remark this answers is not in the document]
rogershijin: Oh that's right! great catch! done!

static/js/protein/data_processing_utils.ts
rogershijin: ah ok that's right, thanks for catching!

static/js/protein/data_processing_utils.ts
rogershijin: thanks for catching! done!

static/js/protein/data_processing_utils.ts
rogershijin: Ohhh! Yes that's much better. I changed it to

```
scoreValues
    .map((bioData: V1BioDatum[], i) => {
      for (const bioDatum of bioData) {
        ...
      }
      console.warn(`Unable to retrieve score for interaction ${interactionDcids[i]} -- default score of ${DEFAULT_INTERACTION_SCORE} used`);
      return DEFAULT_INTERACTION_SCORE;
    })
```

rogershijin:
[ptal!] Thanks for all the reviews so far! Changes:
- Improve user warnings in `scoreDataFromResponse` according to @chejennifer's comments
- `proteinsFromInteractionDcid` now returns null when interaction dcid does not contain exactly 3 underscores
- `symmetricScoreRec` and `deduplicateInteractionDcids`, which call `proteinsFromInteractionDcid`, now check for nullity and skip offending interaction dcids

static/js/protein/data_processing_utils.ts
```diff
@@ -592,3 +529,272 @@ export function getProteinDescription(data: GraphNodes): string {
...
+/**
+ * Given score response, construct a record mapping interaction IDs of the form A_B to their confidence scores,
+ * satisfying the property that if A_B: A_B.score is in the record, then B_A: A_B.score is also.
+ */
+export function scoreDataFromResponse(
+  scoreResponse: V1BioResponse
+): Record<string, number> {
+  const scoreValues = getFromResponse(scoreResponse, "values");
+  const interactionDcids = getFromResponse(scoreResponse, "entity");
+  if (scoreValues.length !== interactionDcids.length) {
```
chejennifer, line 772: Since we care about the scoreValues and interactionDcids together, I think it will be cleaner to just loop through scoreResponse instead of extracting out scoreValues and interactionDcids. If interactionDcids are unique, I would also suggest passing a record with interactionDcid as key and score as value to the symmetricScoreRec function.

So we would have something like: 
```
const scores = {}
for (const responseDatum of scoreResponse.data) {
  const interactionDcid = responseDatum.entity;
  const values = responseDatum.values;
  for (const value of values) {
     ... try to get score, if there is a score add it to the record and break
   }
  warn about no score and add to the record
}
```

chejennifer (commented):
thanks for the updates, one more suggestion

static/js/protein/data_processing_utils.ts
rogershijin: oh yes this is great I agree this would be a lot cleaner, thanks for the pseudocode!

rogershijin:
[ptal!] Thanks @chejennifer for looking at this so many times! Made the changes in your latest comment!

static/js/protein/data_processing_utils.ts
```diff
@@ -592,3 +529,271 @@ export function getProteinDescription(data: GraphNodes): string {
...
+  scoreResponse: V1BioResponse
+): Record<string, number> {
+  if (_.isEmpty(scoreResponse.data)) {
+    return {};
+  }
+  const scoreRec: Record<string, number> = {};
+  for (const responseDatum of scoreResponse.data) {
+    const interactionDcid = responseDatum.entity;
+    if (!interactionDcid) {
+      console.warn("Undefined or empty interaction DCID -- skipping");
```
chejennifer, line 776: include what the interaction DCID that's being skipped is?

static/js/protein/data_processing_utils.ts
```diff
@@ -592,3 +529,271 @@ export function getProteinDescription(data: GraphNodes): string {
...
+      continue;
+    }
+    const interactionId = ppiIdFromDcid(interactionDcid);
+    const values = responseDatum.values;
+    for (const bioDatum of values || []) {
```
chejennifer, line 781: move `|| []` to when defining `values`

static/js/protein/data_processing_utils.ts
```diff
@@ -592,3 +529,271 @@ export function getProteinDescription(data: GraphNodes): string {
...
+      if (_.get(bioDatum, "dcid", "").includes(INTERACTION_QUANTITY_DCID)) {
+        const score = quantityFromDcid(
+          bioDatum.dcid,
+          INTERACTION_QUANTITY_DCID
+        );
+        if (!isNaN(score)) {
+          scoreRec[interactionId] = score;
+          continue;
```
chejennifer, line 789: we probably want to `break` when a score is found

static/js/protein/data_processing_utils.ts
```diff
@@ -592,3 +529,271 @@ export function getProteinDescription(data: GraphNodes): string {
...
+        }
+      }
+      console.warn(
```
chejennifer, line 792: I think this needs to go outside of the current for loop, we don't want to do this for every bioDatum that isn't a score. Should also first check that no score was found for the interactionId

static/js/protein/data_processing_utils.ts
```diff
@@ -592,3 +529,271 @@ export function getProteinDescription(data: GraphNodes): string {
...
+    const scoreBA = _.get(
+      scoreRec,
+      `${proteinB}_${proteinA}`,
+      DEFAULT_INTERACTION_SCORE
+    );
+    const maxScore = Math.max(scoreAB, scoreBA);
+    scoreRec[`${proteinA}_${proteinB}`] = maxScore;
+    scoreRec[`${proteinB}_${proteinA}`] = maxScore;
+  }
+  return scoreRec;
```
chejennifer, line 759: usually it's better to avoid mutating the object that gets passed in, so here I would create a new record object that holds the new values

static/js/protein/data_processing_utils.ts
rogershijin: done!

static/js/protein/data_processing_utils.ts
rogershijin: done!

static/js/protein/data_processing_utils.ts
rogershijin: oops yes absolutely right!

static/js/protein/data_processing_utils.ts
rogershijin: oh yes my mistake! thank you so much for catching this

static/js/protein/data_processing_utils.ts
rogershijin: oh yea I also like pure functions. done!

beets (approved):
thanks for all the iterations! lgtm, pending jennifer's approval

chejennifer (approved):
thanks for all the changes!

rogershijin:
Thank you so much for the reviews/continued guidance! The code is much cleaner now than before.

PR #1661, opened 2022-07-14
GA custom events

raebo1017:
Create custom events for both place explorer and tools and prevent double counting of the event when the component is re-rendered.

This is the detailed design doc: https://docs.google.com/document/d/1KwZUlDNFLtYfCt3Ee0rFh4Do6zj-OFfBwAMKuN14meo/edit?resourcekey=0-RjBbh49MmMYxy8PIAbxB0g#


server/templates/base.html
```diff
@@ -1,33 +1,33 @@
 {#-
- Copyright 2020 Google LLC
+Copyright 2020 Google LLC
```
chejennifer, line 2: curious why the formatting changes? was it from linting?

static/js/tools/map/stat_var_chooser.tsx
```diff
@@ -108,7 +108,12 @@ export function StatVarChooser(props: StatVarChooserProps): JSX.Element {
       samplePlaces={samplePlaces}
       deselectSVs={deselectSVs}
       selectedSVs={selectedSVs}
-      selectSV={(svDcid) => selectStatVar(statVar, display, placeInfo, svDcid)}
+      selectSV={(svDcid) => {
+        selectStatVar(statVar, display, placeInfo, svDcid);
+        window &&
```
chejennifer, line 113: I noticed this same gtag added as part of the selectSV props for all of our tools. Maybe we can just add the gtag to the StatVarHierarchy component where the props.selectSV is used: https://github.com/datacommonsorg/website/blob/78bc010552434c8d41f0e3d92c1638b24c43da58/static/js/stat_var_hierarchy/stat_var_hierarchy.tsx#L298

chejennifer (commented):
this is cool!

static/js/tools/scatter/chart.tsx
```diff
@@ -141,6 +141,22 @@ export function Chart(props: ChartPropsType): JSX.Element {
...
+    statVar2 = props.facetList[1].dcid;
+  }
+  useEffect(() => {
+    window &&
+      window.gtag &&
+      window.gtag("event", "tool_chart_plot", {
+        place_dcid: props.placeInfo.enclosingPlace.dcid,
+        stat_var: [statVar1, statVar2],
+      });
+  }, [statVar1, statVar2, props.placeInfo.enclosingPlace.dcid]);
```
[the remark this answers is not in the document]
raebo1017: I thought about that too. The problem with the props.facetList is it includes metadata. If the source of a stat var changes, it will be triggered as well. However, for this event, we may want to focus more on what combination of stat vars and place dcids are plotted.
For props.xLabel and props.yLabel, they are different from stat var dcid format. Or do you mean we can look at changes to props.xlabel and props.ylabel and return statvar1 and statvar2 value?

server/templates/base.html
raebo1017: Yes, these are caused by running ./run_test.sh -f

static/js/tools/scatter/chart.tsx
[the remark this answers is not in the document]
chejennifer: ah, I see what you mean, ok, keeping it the way it is makes sense!

server/templates/base.html
```diff
@@ -39,9 +39,10 @@
   <script async src="https://www.googletagmanager.com/gtag/js?id={{ GA_ACCOUNT }}"></script>
   <script>
     window.dataLayer = window.dataLayer || [];
-    function gtag(){dataLayer.push(arguments);}
+    function gtag() { dataLayer.push(arguments); }
     gtag('js', new Date());
-    gtag('config', '{{ GA_ACCOUNT }}', { client_storage: 'none', anonymize_ip: true });
+    gtag('config', '{{ GA_ACCOUNT }}', { client_storage: 'none', anonymize_ip: true, place_category: new URLSearchParams(window.location.search).get('category') || '' });
```
beets, line 44: please use category that's already parsed in flask. you can pass in new arguments to the template (and give it a specific name like `place_category` since this will be used across all templates).

static/js/chart/draw.ts
```diff
@@ -90,7 +90,17 @@ function appendLegendElem(
     .attr("style", (d) => `background: ${color(d.label)}`)
     .append("a")
     .text((d) => d.label)
-    .attr("href", (d) => d.link || null);
+    .attr("href", (d) => d.link || null)
+    // Triggered when stat var legend is clicked and send data to google analytics.
```
beets, line 94:
```suggestion
    // Triggered when stat var legend chip is clicked: sends data to google analytics.
```

static/js/chart/draw.ts
```diff
@@ -90,7 +90,17 @@ function appendLegendElem(
...
+    .on(
+      "click",
+      () =>
+        window &&
+        window.gtag &&
+        window.gtag("event", "place_chart_click", {
+          place_chart_click: "stat var",
```
beets, line 101:
```suggestion
          place_chart_click: "stat var chip",
```

static/js/chart/draw.ts
```diff
@@ -90,7 +90,17 @@ function appendLegendElem(
     .attr("style", (d) => `background: ${color(d.label)}`)
     .append("a")
     .text((d) => d.label)
-    .attr("href", (d) => d.link || null);
+    .attr("href", (d) => d.link || null)
+    // Triggered when stat var legend is clicked and send data to google analytics.
+    .on(
+      "click",
+      () =>
+        window &&
```
beets, line 98: window is a global, so we can safely skip this check. curious if you saw documentation / examples of when it might not be the case?

static/js/place/chart.tsx
```diff
@@ -209,7 +209,18 @@ class Chart extends React.Component<ChartPropType, ChartStateType> {
...
       return (
         <span key={source}>
-          <a href={source}>{domain}</a>
+          <a
+            href={source}
+            onClick={() =>
+              window &&
+              window.gtag &&
+              window.gtag("event", "place_chart_click", {
+                place_chart_click: "data from",
```
beets, line 218:
```suggestion
                place_chart_click: "data source",
```

static/js/place/menu.tsx
```diff
@@ -45,6 +45,14 @@ class MenuCategory extends React.Component<MenuCategoryPropsType> {
           href={hrefString}
           className={`nav-link ${selectCategory === category ? "active" : ""}`}
           text={this.props.categoryDisplayStr}
+          handleClick={() =>
+            window &&
+            window.gtag &&
+            window.gtag("event", "place_category_click", {
+              place_category_click: category,
+              place_category_click_source: "menu",
```
beets, line 53: nit: can we call this `sidebar` to be more specific (we ought to rename the component too, but that's for later).

static/js/tools/map/chart.tsx
```diff
@@ -172,6 +172,16 @@ export function Chart(props: ChartProps): JSX.Element {
...
 
+  // Triggered only when stat vars or places change and send data to google analytics.
+  useEffect(() => {
+    window &&
+      window.gtag &&
+      window.gtag("event", "tool_chart_plot", {
+        place_dcid: props.placeInfo.enclosingPlace.dcid,
+        stat_var: props.statVar.value.dcid,
+      });
+  }, [props.statVar.value.dcid, props.placeInfo.enclosingPlace.dcid]);
```
beets, line 183: have you been able to validate that this does not overtrigger?

static/js/tools/scatter/plot_options.tsx
```diff
@@ -86,7 +86,14 @@ function PlotOptions(): JSX.Element {
                     id="per-capita-y"
                     type="checkbox"
                     checked={y.value.perCapita}
-                    onChange={(e) => y.setPerCapita(e.target.checked)}
+                    onChange={(e) => {
+                      y.setPerCapita(e.target.checked);
+                      if (!y.value.perCapita && window && window.gtag) {
+                        window.gtag("event", "tool_chart_option_click", {
+                          tool_chart_option: "per capita",
```
beets, line 93: It just occured to me - do we need to specify what tool this is click is generated from?

static/js/tools/timeline/chart.tsx
```diff
@@ -173,6 +182,13 @@ class Chart extends Component<ChartPropsType, ChartStateType> {
     this.loadRawData();
     this.resizeObserver = new ResizeObserver(this.handleWindowResize);
     this.resizeObserver.observe(this.svgContainer.current);
+    // Triggered when the component is mounted and send data to google analytics.
+    window &&
+      window.gtag &&
+      window.gtag("event", "tool_chart_plot", {
+        place_dcid: Object.keys(this.props.placeNames),
```
beets, line 189: how does the list get interpreted by GA? will [a, b] be treated separately from [a] and [b] (or would we be able to get stats for 'a' in the first example)?

beets (commented):
thanks - this looks good!

since these events are all specified as strings (and some are repeated across callbacks), can we define these as constants in a new file that's specific to GA events? with an appropriate prefix (eg GA_EVENT_TOOL_OPTION, GA_TOOL_OPTION_PER_CAPITA) . this will help us keep track of event keys and values in use throughout the site.

static/js/chart/draw.ts
raebo1017: When I ran it locally, sometimes the whole page cannot be loaded and the error message is window.gtag is not a function. I took the answers on https://stackoverflow.com/questions/62791299/typeerror-window-gtag-is-not-a-function as a reference.

beets: i'm surprised and not sure if i buy those answers. as you can see in our code, we never check for the existence of window, and window.gtag is defined at the head of the template before any other script is run. when exactly did you see that error happen?

static/js/tools/map/chart.tsx
raebo1017: I've tested it in several ways. e.g., only stat var or place is selected, zoom in or out the map, change sources or select per capita. It appears to only trigger once when both stat var and place are selected or any of them change.

static/js/chart/draw.ts
raebo1017: It occurred randomly...Window should already exist and we can skip checking that if we never check for the existence of window. But we may still need to check window.gtag because this script is async and I did encounter some errors...
https://github.com/datacommonsorg/website/blob/master/server/templates/base.html#L39

static/js/tools/scatter/plot_options.tsx
raebo1017: We have page path as a default parameter for every event. We can see which tool this click is generated from from the page path. e.g. /tools/timeline/

static/js/tools/timeline/chart.tsx
raebo1017: GA will change the list to a comma-separated string. For example,  if we have [geoId/06, geoId/0667000], in the GA report, we will get "geoId/06, geoId/0667000". We could only see a combination of them. But we can see the statistics for every single place from the event "tool_place_add".

static/js/place/menu.tsx
raebo1017: Done

static/js/place/chart.tsx
raebo1017: Done

static/js/chart/draw.ts
raebo1017: Done

static/js/chart/draw.ts
raebo1017: Done

static/js/chart/draw.ts
raebo1017: Done

server/templates/base.html
raebo1017: Done

static/js/tools/timeline/chart.tsx
raebo1017: > 

Done. Define constants in a new file for events, parameters and values.

static/js/tools/map/stat_var_chooser.tsx
raebo1017: Done

static/js/shared/types.ts
```diff
@@ -168,3 +168,14 @@ export interface StatVarSearchResult {
   statVars: NamedNode[];
   matches: string[];
 }
+
+// Set new property gtag on window.
+declare global {
```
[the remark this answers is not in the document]
raebo1017: Done

static/js/chart/draw.ts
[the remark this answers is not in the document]
raebo1017: Done

static/js/shared/types.ts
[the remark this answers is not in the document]
chejennifer: I meant, you can put it in this file: https://github.com/datacommonsorg/website/blob/master/static/index.d.ts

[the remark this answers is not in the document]
raebo1017: Understood. Changed it to file index.d.ts

chejennifer (approved):
Thanks for the updates!

beets (commented):
thanks for the updates! just one small update requested to base.html

static/js/shared/util.ts
```diff
@@ -144,3 +144,17 @@ export function removeSpinner(containerId: string): void {
     }
   }
 }
+
+/**
+ * Triggers events and send data to google analytics.
+ * @param eventName name of the event
+ * @param parameter parameter-value pairs
+ */
+export function triggerGAEvent(
```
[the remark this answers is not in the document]
raebo1017: Done

server/templates/base.html
```diff
@@ -41,7 +41,7 @@
     window.dataLayer = window.dataLayer || [];
     function gtag() { dataLayer.push(arguments); }
     gtag('js', new Date());
-    gtag('config', '{{ GA_ACCOUNT }}', { client_storage: 'none', anonymize_ip: true, place_category: new URLSearchParams(window.location.search).get('category') || '' });
+    gtag('config', '{{ GA_ACCOUNT }}', { client_storage: 'none', anonymize_ip: true, place_category: '{{ place_category }}' });
```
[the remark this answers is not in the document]
raebo1017: Done.

beets (approved):
thanks for the upates!

PR #1667, opened 2022-07-18
[Protein Browser] Increase PPI graph depth - Part 4: Graph search

rogershijin:
- Adds a method `bfsIter` to `ProteinProteinInteractionGraph` which expands the protein graph by 1 layer. Some parts of this method could be moved to `data_processing_utils.ts` but I left it in one block to be slightly easier to follow on a first read. This method is currently extremely convoluted so would appreciate any tips for simplification!
- Updates `componentDidUpdate` to call `bfsIter`.  
- Graph now loads by default at `depth=2`.

Another shoutout to @chejennifer for helping me with React debugging!

View on initial load:
image

View after some dragging:
image


rogershijin:
Closing and redoing

PR #1669, opened 2022-07-19
[Import Wizard] Move import wizard to website repo

chejennifer:
- move the import wizard files over to website repo and set up the routing (/import)


server/routes/import_wizard.py
```diff
@@ -0,0 +1,23 @@
...
+"""Import Wizard routes"""
+
+from flask import Blueprint, render_template
+
+bp = Blueprint('import_wizard', __name__, url_prefix='/import')
+
+
+@bp.route('/')
+def main():
+    return render_template('/import_wizard.html')
```
shifucun, line 23: Should only allow this in autopush for now?

shifucun (approved):
Nice!

server/routes/import_wizard.py
chejennifer: yes, you're right, done

chejennifer:
> Does the existing CI of this repo cover the running of import wizard tests? Asking because [datacommonsorg/tools#128](https://github.com/datacommonsorg/tools/pull/128) removes periodic build coverage for the wizard.

yes

PR #1671, opened 2022-07-20
[import wizard] add unit constant

chejennifer:
- add functionality to specify unit for a column
- always show the constants that are applied to the whole file

units for specific columns:
Screen Shot 2022-07-20 at 3 57 37 PM

unit specified for whole file:
Screen Shot 2022-07-20 at 3 57 57 PM


static/js/import_wizard/types.ts
```diff
@@ -46,6 +46,8 @@ export interface Column {
   header: string;
   // column index (leftmost column will be 0)
   columnIdx: number;
+  // unit of measure set for this column
```
pradh, line 49: This feels a bit odd to me to have some mapped-thing specific field in Column, which is sort of a general structure. It is also odd that UNIT appears as a MappedThing when it is a constant, but not when it is specified in the header.

What do you think about having UNIT always appear as MappedThing, and then in MappingVal, when MappingType is COLUMN_HEADER, we use a new field:

```
export type UnitValue = string;

export interface MappingVal {
   ...
   unitValue?: { [columnIdx: number]: UnitValue };`
}
```

chejennifer: Hm, I'm not sure how I feel about COLUMN_HEADER mappingType needing to read from a different field when the MappedThing is UNIT vs everything else. What about an additional MappingType like COLUMN_CONSTANT and then a field in MappingVal like `columnConstants?: { [columnIdx: number]: string}`?

pradh: +1 I like that

chejennifer: done

static/js/import_wizard/utils/validation.ts
```diff
@@ -104,14 +104,19 @@ export function checkMappings(mappings: Mapping): Array<string> {
...
       if (mthing === MappedThing.PLACE) {
         // Check #8
-        errors.push(mthingName + ": must not be CONSTANT type");
+        errors.push(mthingName + ": must not be FILE_CONSTANT type");
+      }
+    } else if (mval.type === MappingType.COLUMN_CONSTANT) {
+      if (_.isEmpty(mval.columnConstants)) {
+        // Check #1
+        errors.push(mthingName + ": missing value for COLUMN_CONSTANT type");
       }
```
pradh, line 120: To confirm, do we only support UNIT as column_constant? Or is it that the UI supports only unit, but the rest of the code is general?  Accordingly add comments to types.ts.

pradh (approved):
This is very nice!

static/js/import_wizard/utils/validation.ts
chejennifer: the UI supports only unit for column_constant but the rest of the code is general.

Added comment in the types.ts file

PR #1673, opened 2022-07-21
[Protein Browser] User input - Part 1: Depth

rogershijin:
- Adds depth input to protein-protein interaction graph (input element copied from @chejennifer's download tool)
- Supports depths 1-3.

This requires 7 request RTTs: original page request --> [ (interactors --> score) 3 times ], but does not slow down other plots on the same page.  On 131 Mbps wifi, this is 2.2 seconds slower than the other plots.

This depends on [Increase PPI Graph Depth - Part 4](https://github.com/datacommonsorg/website/pull/1667) but I think `render` of `protein_protein_interaction_graph.tsx` can be checked independently.

Thanks for reading!

image


rogershijin:
Closing and redoing

PR #1674, opened 2022-07-21
[import wizard] fix some bugs

chejennifer:
- fix bug where page goes blank when place type of "state" is chosen
- update radio button text
- allow selecting place type and place property when header is a place

Screen Shot 2022-07-21 at 3 10 48 PM


static/js/import_wizard/types.ts
```diff
@@ -53,10 +53,12 @@ export interface MappingVal {
   // Column that holds the mapping values. Should be set if type is
   // MappingType.COLUMN
   column?: Column;
-  // When MappedThing is PLACE, the value corresponds to place property in KG.
-  placeProperty?: DCProperty;
-  // When MappedThing is PLACE, the value corresponds to place type in KG.
-  placeType?: DCType;
+  // Record of column idx to the place property (in KG) associated with that
+  // column or column header. Should be set if MappedThing is PLACE.
+  placeProperty?: Record<number, DCProperty>;
```
pradh, line 58: I was previously puzzled with Record, and ran into issues when using it for "Mapping", since not all keys (MappedThing) may appear in an instance.

Reading https://betterprogramming.pub/typescripts-record-type-explained-691372b1a449, would the Index Signature type be more appropriate here?

Given we're not constraining the "number" values, and it allows a human-readable name for `number`, like `{ [columnIdx: number]: DCProperty }`.

pradh: To confirm, there will be multiple entries here only if MappingType is COLUMN_HEADERS?

static/js/import_wizard/utils/validation.ts
```diff
@@ -60,27 +62,40 @@ export function checkMappings(mappings: Mapping): Array<string> {
...
-    }
     if (mval.type === MappingType.COLUMN) {
       if (mval.column == null || mval.column.id === "") {
         // Check #1
         errors.push(mthingName + ": missing value for COLUMN type ");
       }
+      if (mthing === MappedThing.PLACE) {
+        if (
+          _.isEmpty(mval.placeProperty) ||
+          _.isEmpty(mval.placeProperty[mval.column.columnIdx])
```
pradh, line 73: Also assert placeProperty Record has only one entry?

pradh (commented):
Nicely done!

static/js/import_wizard/components/mapping_column_options.tsx
```diff
@@ -169,10 +93,50 @@ export function MappingColumnOptions(
...
+                .map((thing) => (
+                  <option value={thing} key={thing}>
+                    {MAPPED_THING_NAMES[thing] || thing}
+                  </option>
+                ))}
+            </Input>
+            {props.column.columnMappedThing === MappedThing.PLACE &&
+              getPlaceTypePropertyInputs(MappingType.COLUMN)}
+            {!_.isEmpty(props.column.sampleValues) && (
+              <span>{`(eg. ${props.column.sampleValues
```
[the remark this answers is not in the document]
chejennifer: done

static/js/import_wizard/components/mapping_section.tsx
```diff
@@ -301,7 +307,19 @@ function getColumnInfo(
...
+      headerPlaceType: placeDetector.placeTypes.get(defaultPlaceType),
+      headerPlaceProperty: defaultPlaceProperty,
       headerMappedThing: REQUIRED_MAPPINGS[0],
+      sampleValues: [],
+    });
+  });
+  csvData.rowsForDisplay.forEach((rowVals) => {
+    rowVals.forEach((val, idx) => {
+      const colInfo = columnInfo.get(idx);
+      if (_.isEmpty(colInfo) || _.isEmpty(val)) {
```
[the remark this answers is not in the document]
chejennifer: done

static/js/import_wizard/components/mapping_section.tsx
```diff
@@ -341,6 +361,18 @@ function getColumnInfo(
...
         columnInfo.get(colIdx).headerMappedThing = mappedThing;
+        if (mappedThing === MappedThing.PLACE) {
+          const placeType =
+            colIdx in mappingVal.placeType
+              ? mappingVal.placeType[colIdx]
+              : placeDetector.placeTypes.get(defaultPlaceType);
+          const placeProperty =
+            colIdx in mappingVal.placeProperty
+              ? mappingVal.placeProperty[colIdx]
+              : Array.from(validPlaceTypeProperties[placeType.dcid])[0];
```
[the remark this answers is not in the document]
chejennifer: done

static/js/import_wizard/components/mapping_table_header.tsx
```diff
@@ -170,12 +170,16 @@ function getColumnMappingString(column: ColumnInfo): string {
   if (column.type === MappingType.COLUMN) {
     let mString = `${mThingName}s`;
```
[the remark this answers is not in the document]
chejennifer: done

static/js/import_wizard/types.ts
chejennifer: that makes sense to me, done.

Correct, there should only be multiple entries if MappingType is COLUMN_HEADERS

static/js/import_wizard/utils/validation.ts
chejennifer: I'm not sure this is necessary because when generating tmcf and observations, if the mapping type is COLUMN, we will only read the place property for the specified column. Also, there is only one place to input a place property and type when user has chosen mapping type column, so I don't think it'll be possible for there to be more than one entry

pradh: If you consider the point of this function is to make correctness assertions on mappings based on the spec, without assuming any knowledge of code that updates or uses the mappings (that was me when I wrote it), how would you feel about this assertion? :)

If you're still unsure, feel free to skip...

pradh (approved):
Thanks for the edits!

static/js/import_wizard/utils/validation.ts
chejennifer: ok, I see your point, updated!

pradh: nit: 
```
if (Object.keys(mval.placeProperty).length !== 1) {
   errors.push("Place mapping for a column expected to have one placeProperty");
}
```

PR #1677, opened 2022-07-26
[import wizard] Add value mapping section

chejennifer:
- add section where user can input value mappings or upload a file. (value mapping as in a mapping from cell value in original csv to that cell's value in the cleaned csv)

user input
screen-recording-_46_

file upload
screen-recording-_47_

### DIFF
--- a/static/js/import_wizard/components/value_map_section.tsx
+++ b/static/js/import_wizard/components/value_map_section.tsx
@@ -61,13 +61,14 @@ export function ValueMapSection(props: ValueMapSectionProps): JSX.Element {
     <>
       <div className="section-container">
         <div className="section-header">
-          <h2>Value Mappings</h2>
+          <h2>Update File Values</h2>
           <div className="header-tab-section">
+            {/*TODO: make this into a toggle switch like in scatter tool*/}
             <div
               className={`header-tab${showFileUpload ? "" : "-selected"}`}
               onClick={() => setShowFileUpload(false)}
             >
-              Input mapping
+              Enter values to remap
             </div>
             <span>|</span>
             <div
@@ -99,7 +100,8 @@ export function ValueMapSection(props: ValueMapSectionProps): JSX.Element {
                   </li>
                 </ul>
               </li>
-              <li>Original CSV value is case insensitive</li>
+              <li>Original CSV value is case sensitive</li>
+              <li>If the original CSV value is empty, the mapping will be skipped.</li>
             </ul>
             <input
               type="file"
@@ -206,7 +208,7 @@ export function ValueMapSection(props: ValueMapSectionProps): JSX.Element {
           if (row.length !== 2) {
             continue;
           }
-          valueMap[row[0].toLowerCase()] = row[1];
+          valueMap[row[0]] = row[1];
         }
         props.onValueMapSubmitted(valueMap);
         setShowConfirmationButton(false);
@@ -214,7 +216,7 @@ export function ValueMapSection(props: ValueMapSectionProps): JSX.Element {
       reader.readAsText(uploadedFile);
     } else {
       Object.values(mappingInput).forEach((entity) => {
-        valueMap[entity.originalVal.toLowerCase()] = entity.newVal;
+        valueMap[entity.originalVal] = entity.newVal;
       });
       props.onValueMapSubmitted(valueMap);
       setShowConfirmationButton(false);
--- a/static/js/import_wizard/components/value_map_section.tsx
+++ b/static/js/import_wizard/components/value_map_section.tsx
@@ -1,0 +1,236 @@
+/**
+ * Copyright 2022 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * Component for showing and updating the mapping of cell values.
+ */
+
+import _ from "lodash";
+import Papa from "papaparse";
+import React, { useEffect, useState } from "react";
+import { Button, Input } from "reactstrap";
+
+import { ValueMap } from "../types";
+
+interface ValueMapSectionProps {
+  valueMap: ValueMap;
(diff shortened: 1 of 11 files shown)


beets:
Just a quick comment before taking a look at the code:

- Should we do a case-sensitive search/replace?
- I'm not sure if we should allow mapping from blank (since this is more likely to apply to multiple columns).

static/js/import_wizard/components/value_map_section.tsx
```diff
@@ -0,0 +1,236 @@
...
+  useEffect(() => {
+    props.onValueMapUpdated();
+    setShowConfirmationButton(true);
+  }, [uploadedFile, mappingInput]);
+
+  return (
+    <>
+      <div className="section-container">
+        <div className="section-header">
+          <h2>Value Mappings</h2>
```
beets, line 64: Maybe: "Update file data" or "Update csv file values"

static/js/import_wizard/components/value_map_section.tsx
```diff
@@ -0,0 +1,236 @@
...
+          <div className="header-tab-section">
+            <div
+              className={`header-tab${showFileUpload ? "" : "-selected"}`}
+              onClick={() => setShowFileUpload(false)}
+            >
+              Input mapping
```
beets, line 70: "Enter data to update" or "Enter values to remap"

static/js/import_wizard/components/value_map_section.tsx
```diff
@@ -0,0 +1,236 @@
...
+    setShowConfirmationButton(true);
+  }, [uploadedFile, mappingInput]);
+
+  return (
+    <>
+      <div className="section-container">
+        <div className="section-header">
+          <h2>Value Mappings</h2>
+          <div className="header-tab-section">
+            <div
```
beets, line 66: (for later) We could use a toggle switch buttons here like we do for scatter

static/js/import_wizard/components/value_map_section.tsx
```diff
@@ -0,0 +1,236 @@
...
+
+  function onConfirmationClicked(): void {
+    const valueMap = {};
+    if (showFileUpload) {
+      const reader = new FileReader();
+      reader.onload = (e) => {
+        const rows = Papa.parse(e.target.result as string).data;
+        for (const row of rows) {
+          if (row.length !== 2) {
+            continue;
```
beets, line 207: if it's not a blank line, should we display an error? e.g. Skipping ___ because there was no value specified.

chejennifer:
> Just a quick comment before taking a look at the code:
> 
> * Should we do a case-sensitive search/replace?
> * I'm not sure if we should allow mapping from blank (since this is more likely to apply to multiple columns).

1. case-sensitive makes sense to me
2. I guess mapping blank would make sense per column, but maybe not if it's for the whole file..

static/js/import_wizard/components/value_map_section.tsx
chejennifer: actually, if there's only one value, should we map that value to blank in the cleaned csv?

beets: This would still need to be specified as 
`bad_value,`
with the trailing comma, so i would imagine that papaparse would return that as 2 values

pradh: +1 to having the capability of mapping bad_value to empty

static/js/import_wizard/utils/file_generation.test.ts
```diff
@@ -246,3 +263,47 @@ test("generateCsv", () => {
...
+  const originalCsvFile = new File([originalCsvContent], "original.csv", {
+    type: "text/csv;chartset=utf-8",
+  });
+  const csvData = {
+    orderedColumns: [columnSameHeaderId1, columnDiffHeaderId2, column3],
+    columnValuesSampled: new Map(),
+    rowsForDisplay: new Map(),
+    rawCsvFile: originalCsvFile,
+  };
+  const valueMap = { test: "abc", "test1, test2": "test3, test4" };
```
pradh, line 300: nit: add empty value?

static/js/import_wizard/components/value_map_section.tsx
pradh: also, might be worth having unit-test for this parsing?

static/js/import_wizard/components/value_map_section.tsx
```diff
@@ -0,0 +1,236 @@
...
+          }
+          valueMap[row[0].toLowerCase()] = row[1];
```
pradh, line 209: should we note in the UI text that it is case insensitive?

static/js/import_wizard/utils/obs_generation.ts
```diff
@@ -54,24 +55,35 @@ function hasRequiredProps(obs: Observation): boolean {
...
-  obsGenMaps: ObsGenMaps
+  obsGenMaps: ObsGenMaps,
+  valueMap: ValueMap
 ): Array<Observation> {
   const obsList = new Array<Observation>();
   for (const valColIdx of Array.from(obsGenMaps.valCol2Hdr.keys())) {
     const hdrThing = obsGenMaps.valCol2Hdr.get(valColIdx);
     const obs: Observation = new Map();
+    let cellVal = row[valColIdx] || "";
+    const lowerCaseCellVal = cellVal.toLowerCase();
```
pradh, line 66: Given this happens twice, use a helper `getCellVal(row, valColIdx, valueMap)` that does the empty check, lower-casing and valueMap lookup if needed?

pradh (commented):
Nice!

static/js/import_wizard/components/value_map_section.tsx
chejennifer: done

static/js/import_wizard/components/value_map_section.tsx
chejennifer: done

static/js/import_wizard/components/value_map_section.tsx
chejennifer: done, added TODO

static/js/import_wizard/components/value_map_section.tsx
chejennifer: done

static/js/import_wizard/utils/file_generation.test.ts
chejennifer: done

static/js/import_wizard/utils/obs_generation.ts
chejennifer: done

pradh (approved):
Thanks for the test!

beets (approved):
thanks for the update and adding tests!

chejennifer:
thanks for the reviews!

PR #1682, opened 2022-07-27
Fix broken webdriver test

chejennifer:
- Fix broken webdriver test caused by updated Count_Person data


beets (approved):
thanks for the fix!

PR #1685, opened 2022-07-28
[map] domain set in URL should take priority

chejennifer:
If user has set values for the domain through the URL params, we should use that domain when generating the legend


pradh (approved):
Thanks for the quick fix!

beets (approved):
thanks for the fix!

PR #1686, opened 2022-07-28
datacommons.py property-values function

rogershijin:
Include `property-values` function from [this commit](https://github.com/shifucun/website/blob/79046d4923e7fd1b027cfcf0c24ea92b5249cdc6/server/services/datacommons.py#L156) by @shifucun, updated to use the new V1 endpoint format.

### DIFF
--- a/server/services/datacommons.py
+++ b/server/services/datacommons.py
@@ -328,6 +328,22 @@ def get_property_labels(dcids):
     return results
 
 
+def property_values(entities, prop, dir):
+    resp = post(f'/v1/bulk/property/values/{dir}', {
+        'entities': sorted(entities),
+        'property': prop,
+    })
+    result = {}
+    for item in resp['data']:
+        result[item['entity']] = []
+        for v in item['values']:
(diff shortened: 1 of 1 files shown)


server/services/datacommons.py
```diff
@@ -328,6 +328,22 @@ def get_property_labels(dcids):
...
+        'property': prop,
+    })
+    result = {}
+    for item in resp['data']:
+        result[item['entity']] = []
+        for v in item['values']:
+            if 'dcid' in v:
+                result[item['entity']].append(v['dcid'])
+            else:
+                result[item['entity']].append(v['value'])
```
shifucun, line 343: entity, values = item['entity'], item['values]
result[entity] = []
for v in values:
   if ...
      result[entity].append(v['dcid'])

shifucun (approved):
thanks!

server/services/datacommons.py
rogershijin: thanks that's cleaner! will do

PR #1687, opened 2022-07-28
[Protein Browser] New flask endpoint for protein-protein interaction graph

rogershijin:
Adds new flask endpoint `/ppi/bfs` which retrieves graph data for protein-protein interaction graph, along with helpers

- `id`, `dcid` for id conversions
- `node` for constructing a node object
- `extract_intactmi` for parsing IntactMi interaction confidence score from IntactMi dcid
- `interactors` for retrieving the two interactors from an interaction id
- `target` for retrieving the other interactor given an interaction id and the id of one of the interactors
- `reverse_interaction_id` and `reverse_interaction_dcid` for flipping an interaction id/dcid
- `filter_interaction_dcids` for deduplicating and removing self-interactions from a list of interaction dcids
- `symmetrize_scores` for constructing a version of a interaction_id --> score dict where both `interaction_id` and `reverse(interaction_id)` map to the greater score between the two directions
- `flatten` for flattening a list of lists (1 deep). 

Uncomments the graph in `page.tsx`.  Graph loads at depth 1 with top 5 interactors for page protein and looks like this:
image


server/routes/api/protein.py
```diff
@@ -25,10 +30,293 @@
...
-                                     compress=False,
-                                     post=False,
-                                     has_payload=False)
+                             compress=False,
+                             post=False,
+                             has_payload=False)
     return response
+
+
+def id(id_or_dcid):
```
jehangiramjad, line 42: if this function will only be called from within protein.py then consider naming it with an underscore, e.g. _id(). If you expect this function to be called from outside this file, then the function-level comments should follow the Python style guide. See the example in this section: https://google.github.io/styleguide/pyguide.html#383-functions-and-methods

server/routes/api/protein.py
```diff
@@ -25,10 +30,293 @@
...
+    '''
+    Delete 'bio/' prefix from a biomedical DCID, if it is present.
+    E.g. 'bio/P53_HUMAN' --> 'P53_HUMAN'
+    '''
+    # .replace replaces all instances, but no biomedical DCID will contain more than once instance of 'bio/'
+    return id_or_dcid.replace(BIO_DCID_PREFIX, '')
+
+
+def dcid(id_or_dcid):
```
jehangiramjad, line 51: Same comment as above about _dcid(), if applicable.

server/routes/api/protein.py
```diff
@@ -25,10 +30,293 @@
...
+    '''
+    Add 'bio/' prefix to an id if it does not already start with 'bio/'.
+    E.g. 'P53_HUMAN' --> 'bio/P53_HUMAN'
+    '''
+    if id_or_dcid.startswith(BIO_DCID_PREFIX):
+        return id_or_dcid
+    return BIO_DCID_PREFIX + id_or_dcid
+
+
+def node(protein_id_or_dcid, depth):
```
jehangiramjad, line 61: This also looks like an internal function only so it should be _node(). Otherwise, you should have asserts/checks for the calls to id() and .split() not failing.

server/routes/api/protein.py
```diff
@@ -25,10 +30,293 @@
...
+
+def extract_intactmi(score_dcids):
+    '''
+    Takes list of interaction score DCIDs and numerically parses and returns the first instance
+    of an IntactMi score, or the default interaction score if none are found.
+    
+    TODO: add example
+    '''
+    for score_dcid in score_dcids:
+        if score_dcid.startswith("IntactMiScore"):
```
jehangiramjad, line 84: nit: If you're using the BIO_DCID_PREFIX as a global variable then perhaps do the same for IntactMiScore too? Not a big deal if you keep it this way

server/routes/api/protein.py
```diff
@@ -25,10 +30,293 @@
...
+            return float(score_dcid.replace("IntactMiScore", ""))
```
jehangiramjad, line 85: Can this call to float() fail?

server/routes/api/protein.py
```diff
@@ -25,10 +30,293 @@
...
+
+
+def interactors(interaction_id_or_dcid):
+    '''
+    Parse and return ids of two interactors from interaction DCID of the form 'bio/{protein1 id}_{protein2 id}'
+    E.g. 'bio/P53_HUMAN_CBP_HUMAN' --> 'P53_HUMAN', 'CBP_HUMAN'
+    '''
+    interaction_id = id(interaction_id_or_dcid)
+    protein1, species1, protein2, species2 = interaction_id.split('_')
+    return f'{protein1}_{species1}', f'{protein2}_{species2}'
```
jehangiramjad, line 96: I am not a fan of returning tuples. Could we introduce a data structure with two attributes and set the attributes here and return that data structure?

server/routes/api/protein.py
```diff
@@ -25,10 +30,293 @@
...
+    '''
+    dcids = set()
+    uniques = []
+    for interaction_dcid in interaction_dcids:
+        source_id, target_id = interactors(interaction_dcid)
+        # filter out self-interactions
+        if source_id == target_id:
+            continue
+        reverse_dcid = reverse_interaction_dcid(interaction_dcid)
+        if interaction_dcid in dcids or reverse_dcid in dcids:
```
jehangiramjad, line 144: So the first time you see an interaction_dcid it won't be in the dcids set, right? And at this point, you won't insert it in uniques. So if you never see that again, you won't return it as part of uniques?

server/routes/api/protein.py
```diff
@@ -25,10 +30,293 @@
...
+       partition (either expansion_targets or cross_targets), id1 appears before id2 in this partition too.
+    '''
+    expansion_targets = []
+    cross_targets = []
+    for target_id in target_ids:
+        if target_id in node_set:
+            cross_targets.append(target_id)
+        else:
+            expansion_targets.append(target_id)
+    return expansion_targets, cross_targets
```
jehangiramjad, line 195: Same comment about returning tuples as above. You could also return a map/dict which has the keys corresponding to each of these targets.

server/routes/api/protein.py
```diff
@@ -25,10 +30,293 @@
...
+        'nodeDataNested': [[center protein dict], [...center protein neighbor dicts], ...]
+        'linkDataNested': [[], [...center protein link dicts], ...]
+    }.
+
+    TODO: example
+    '''
+    # adjacency list representation. maps interaction ids to list of target nodes, sorted in descending order by score
+    scores = {}
+
+    center_protein_dcid = escape(request.json['proteinDcid'])
```
jehangiramjad, line 226: what if any of these request params are not present?

jehangiramjad (commented):
Leaving some comments, mostly on style and not the actual logic. For all functions which will only be used by other functions in this file, name them with an underscore, e.g. _id().

server/routes/api/protein.py
```diff
@@ -25,10 +30,292 @@
...
+    '''
+    Parse and return ids of two interactors from interaction DCID of the form 'bio/{protein1 id}_{protein2 id}'
+    E.g. 'bio/P53_HUMAN_CBP_HUMAN' --> 'P53_HUMAN', 'CBP_HUMAN'
+    '''
+    interaction_id = id(interaction_id_or_dcid)
+    protein1, species1, protein2, species2 = interaction_id.split('_')
+    return f'{protein1}_{species1}', f'{protein2}_{species2}'
+
+
+def target(interaction_id_or_dcid, source_id_or_dcid):
```
chejennifer, line 99: looks like this method is only ever called with an interaction_dcid and a source_id so maybe update the arguments

server/routes/api/protein.py
```diff
@@ -25,10 +30,292 @@
...
+def reverse_interaction_id(interaction_id):
+    '''
+    Given interaction ID of the form '{protein1 id}_{protein2 id}', return '{protein2 id}_{protein1 id}'.
+    E.g. 'P53_HUMAN_CBP_HUMAN' --> 'CBP_HUMAN_P53_HUMAN'
+    '''
+    interactor_id1, interactor_id2 = interactors(interaction_id)
+    return f'{interactor_id2}_{interactor_id1}'
+
+
+def reverse_interaction_dcid(interaction_dcid):
```
chejennifer, line 121: I'm wondering if this helper is necessary, are there many cases this is needed?

server/routes/api/protein.py
rogershijin: Thank you so much for the review! Got it will do!

server/routes/api/protein.py
rogershijin: Got it will do!

server/routes/api/protein.py
rogershijin: Thanks for noticing I'll add these checks!

server/routes/api/protein.py
rogershijin: Sounds good I'll add a constant!

server/routes/api/protein.py
rogershijin: I suppose never say never, I'll try/catch!

server/routes/api/protein.py
rogershijin: Yes that's cleaner, will do!

server/routes/api/protein.py
rogershijin: Ahh thank you so much for catching this! Yes I meant `interaction_dcids`, but this was very hard to catch because almost everything is repeated in the forward/reverse directions except for self-interactions which we already filtered out.

server/routes/api/protein.py
rogershijin: Got it will do!

server/routes/api/protein.py
rogershijin: Good point! should we respond with some sort of error here?

server/routes/api/protein.py
```diff
@@ -25,10 +30,300 @@
...
+    cross_targets = []
+    for target_id in target_ids:
+        if target_id in node_set:
+            cross_targets.append(target_id)
+        else:
+            expansion_targets.append(target_id)
+    return expansion_targets, cross_targets
+
+
+@bp.route('/ppi/bfs/', methods=['POST'])
```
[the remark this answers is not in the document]
rogershijin: Got it! I'll go with protein protein interaction for both!

server/routes/api/protein.py
```diff
@@ -25,10 +30,300 @@
...
+    {
+        'nodeDataNested': [[...depth 0 node dicts], [...depth 1 node dicts], ...]
+        'linkDataNested': [[...depth 0 link dicts], [...depth 1 link dicts], ...]
+    }
+    The only depth 0 node is the center protein and since we do not support self-interactions at the moment,
+    there are no depth-0 links, so the above can be simplified to
+    {
+        'nodeDataNested': [[center protein dict], [...center protein neighbor dicts], ...]
+        'linkDataNested': [[], [...center protein link dicts], ...]
+    }.
```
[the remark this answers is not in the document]
rogershijin: Sorry yes your version is better! thanks for catching

server/routes/api/protein.py
```diff
@@ -25,10 +30,300 @@
...
+        'depth': 2,
+        'maxInteractors': 2,
+        'proteinDcid': 'bio/P53_HUMAN',
+        'scoreThreshold': 0.4
+    }.
+    '''
+    # adjacency list representation. maps interaction ids to list of target nodes, sorted in descending order by score
+    scores = {}
+
+    center_protein_dcid = escape(request.json['proteinDcid'])
```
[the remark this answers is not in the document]
rogershijin: Oh I added this because there was a code scanning error about an [XSS vulnerability](https://owasp.org/www-community/attacks/xss/).  I think it's like if an attacker gets ahold of a victim's computer they might be able to make a request to this endpoint with proteinDcid set to some malicious script and then the script could run in the victim's browser if proteinDcid is somehow incorporated into HTML.  Should we just ignore it?

server/routes/api/protein.py
```diff
@@ -25,10 +30,300 @@
...
+    score_threshold = request.json['scoreThreshold']
+    max_interactors = request.json['maxInteractors']
+    max_depth = request.json['depth']
```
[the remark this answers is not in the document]
rogershijin: Sounds good! Thanks for noticing!

server/routes/api/protein.py
```diff
@@ -25,10 +30,292 @@
...
+    node_id_set = set([id(center_protein_dcid)])
+
+    # to become final nested list of returned node dicts
+    nodes = [[node(center_protein_dcid, 0)]]
+    # to become final nested list of returned link dicts
+    links = [[]]
+
+    # used to request the next layer of interactors
+    last_layer_node_dcids = [center_protein_dcid]
+    for depth in range(1, max_depth + 2):
```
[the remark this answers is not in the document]
rogershijin: Yeah the last iteration is purely to compute the edges between the new nodes from the previous iteration and other nodes already in the graph (I call these "cross-links" but there's probably a better word for these).  We wanna show these in the rendering so that if two nodes don't have an edge between them, we can say that they don't interact above threshold.

server/routes/api/protein.py
```diff
@@ -25,10 +30,292 @@
...
+        scores.update(layer_scores)
+        # used to request the next layer of interactors
+        last_layer_node_dcids = []
+        # links from a protein in the graph to some new protein outside the graph
+        expansion_links = []
+        # ids of new proteins not currently in graph
+        new_node_ids = set()
+
+        for source_dcid, interaction_dcids in layer_interactors.items():
+            interaction_dcids_new = [
```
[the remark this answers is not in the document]
rogershijin: You know I love using these filter/map things! Let's do it!

server/routes/api/protein.py
```diff
@@ -25,10 +30,292 @@
...
+            interaction_dcids_filtered = filter_interaction_dcids(
+                interaction_dcids_new)
+
+            # add both interaction dcids and their reversals
+            interaction_dcid_set.update(interaction_dcids_filtered)
+            interaction_dcid_set.update(
+                map(reverse_interaction_dcid, interaction_dcids_filtered))
+
+            source_id = id(source_dcid)
+            target_ids = [
```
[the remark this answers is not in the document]
rogershijin: absolutely will do!

server/routes/api/protein.py
rogershijin: Just two! what's the usual repetition allowance before we define a helper?

server/routes/api/protein.py
rogershijin: I added an assert for in id() for input not having multiple instances of 'bio/', but I feel like the only (easy to implement) constraint for the split is that it's length 2, and `protein, species = node_id.split('_')` will already raise an error in this case.  Should I still add an explicit assert here?

server/routes/api/protein.py
rogershijin: Oops sorry for this I really can't think of better attribute names than "first_interactor" and "second_interactor".  "source", "target" might be an option and we could rename the `target` function to `_other_interactor`?  Although an interaction id specifies an undirected edges so source/target might not be the best either and we should rename `_target` regardless.

server/routes/api/protein.py
rogershijin: sounds good!

server/routes/api/protein.py
rogershijin: Actually I have no idea what I was thinking! Rewrote the function hopefully it makes more sense now

server/routes/api/protein.py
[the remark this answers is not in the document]
rogershijin: Ok I took it out - the markup symbols might also make it harder to write tests but I'm not sure about this.

server/routes/api/protein.py
rogershijin: Oh actually one of these uses is a map which will get kind of ugly if we replace it with a long lambda.  Is it ok to keep?

server/routes/api/protein.py
jehangiramjad: Looks good!

server/routes/api/protein.py
jehangiramjad: first_interactor and second_interaction seems fine to me

server/routes/api/protein.py
jehangiramjad: Great!

jehangiramjad: Ideally, we'd have unit tests for all functions in this file. But this can be a TODO if you have some time

jehangiramjad (approved):
Thanks! Looks much better now. Modulo @chejennifer's comment resolutions, this looks fine to me. I would encourage you to work on adding unit tests corresponding to all functions in this file if you happen to have a spare day or so.

server/routes/api/protein.py
```diff
@@ -13,22 +13,411 @@
...
+    for interaction_dcid in interaction_dcids:
+        try:
+            protein_id1, protein_id2 = _interactors(interaction_dcid).values()
+        # skip malformatted ids
+        except ValueError:
+            continue
+        # filter out self-interactions
+        if protein_id1 == protein_id2:
+            continue
+        reverse_dcid = _reverse_interaction_dcid(interaction_dcid)
```
chejennifer, line 193: I think this should also be wrapped in a try because _reverse_interaction_dcid could throw an error, but I don't think this filter method should end up throwing any errors ..

server/routes/api/protein.py
```diff
@@ -13,22 +13,411 @@
...
+    '''
+    Given a dict mapping interaction IDs to scores, return a new dict D of the same type such that
+    1) For any interaction ID in D, its reversal is also a key of D
+    2) Any interaction ID in D maps to the same value as its reversal
+    3) If an interaction ID is in scores but its reversal is not, D[interaction ID] = scores[interaction ID]
+    4) If both an interaction ID F and its reversal R are in scores, D[F] = D[R] = max(scores[F], scores[R])
+    '''
+    scores_sym = {}
+    for interaction_id, score in scores.items():
+        reverse_id = _reverse_interaction_id(interaction_id)
```
chejennifer, line 210: should this be in a try? since reverse_interaction_id can throw errors

server/routes/api/protein.py
```diff
@@ -13,22 +13,411 @@
...
+            raise TypeError('scoreThreshold must be a number')
+        if type(max_interactors) != int:
+            raise TypeError("maxInteractors must be an integer")
+        if type(max_depth) != int:
+            raise TypeError("maxDepth must be an integer")
+        if type(center_protein_dcid) != str:
+            raise TypeError("proteinDcid must be a string")
+
+    except KeyError as key_error:
+        return f'Missing request parameter {key_error}', BAD_REQUEST_CODE
```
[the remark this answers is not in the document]
chejennifer: please fix this, you don't want the KeyError stacktrace exposed. For each of those KeyErrors, we just want to say something like: "error: must provide a <key> field"

server/routes/api/protein.py
```diff
@@ -13,22 +13,411 @@
...
+
+    except TypeError as type_error:
+        return f'Incorrect request parameter type: {type_error}', BAD_REQUEST_CODE
```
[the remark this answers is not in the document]
chejennifer: please fix this as well, don't want to expose the TypeError stack trace.

server/routes/api/protein.py
```diff
@@ -13,22 +13,398 @@
...
+        max_depth = request.json['maxDepth']
+    except KeyError as key_error:
+        return f'Missing request parameter {key_error}', BAD_REQUEST_CODE
+    for param, limits in LIMITS.items():
+        if not (limits['min'] <= request.json[param] <= limits['max']):
+            return f'{param} of {score_threshold} out of bounds. Please try a value in [{limits["min"]}, {limits["max"]}]', BAD_REQUEST_CODE
+    try:
+        center_protein_node = _node(center_protein_dcid, 0)
+    except ValueError:
+        return f'Invalid protein DCID {center_protein_dcid}'
```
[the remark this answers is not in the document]
chejennifer: Please fix this by updating the message to just "Invalid proteinDcid".

Also, need to add the returned code here

server/routes/api/protein.py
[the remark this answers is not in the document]
chejennifer: Please add a comment for this.

server/routes/api/protein.py
```diff
@@ -13,22 +13,411 @@
...
+        if type(center_protein_dcid) != str:
+            raise TypeError("proteinDcid must be a string")
+
+    except KeyError as key_error:
+        return f'Missing request parameter {key_error}', BAD_REQUEST_CODE
+
+    except TypeError as type_error:
+        return f'Incorrect request parameter type: {type_error}', BAD_REQUEST_CODE
+
+    for param, limits in LIMITS.items():
```
chejennifer, line 314: From the conversation yesterday, doesn't seem like we need these limits anymore?

chejennifer (commented):
thanks for the updates!

server/routes/api/protein.py
```diff
@@ -13,22 +13,411 @@
...
+                except ValueError:
+                    continue
+                new_nodes.append(new_node)
+            nodes.append(new_nodes)
+            node_id_set.update(new_node_ids)
+            links.append(expansion_links)
+
+    graph = {'nodeDataNested': nodes, 'linkDataNested': links}
+
+    return Response(json.dumps(graph))
```
chejennifer, line 423: Also need to return a 200 code: https://github.com/datacommonsorg/website/blob/master/server/routes/api/stats.py#L539

server/routes/api/protein.py
rogershijin: Oh good catch will do!

server/routes/api/protein.py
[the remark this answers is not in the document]
rogershijin: Got it will do!

server/routes/api/protein.py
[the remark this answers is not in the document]
rogershijin: Oh ok will do! I think when I tested this it seemed like client only receives the name of the key in the response payload but perhaps the full stacktrace is there too.

server/routes/api/protein.py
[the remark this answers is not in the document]
rogershijin: Sounds good!

server/routes/api/protein.py
rogershijin: Thanks will do!

server/routes/api/protein.py
rogershijin: thanks for catching will do!

server/routes/api/protein.py
rogershijin: These may have to be reintroduced in the future but I think currently the graph has 26k proteins with interaction data and 77,000 interactions so it would not be terrible for the user to load everything.

chejennifer (approved):
thanks for all the changes!

PR #1688, opened 2022-07-28
[Protein Browser] Factor out PPI graph as child component of Page

rogershijin:
- Creates a new component `ProteinProteinInteractionGraph` in `protein_protein_interaction_graph.tsx` and adds it as a child component in `page.tsx`, commented out until the flask endpoint PR is merged
- Deletes unnecessary utilities from `data_processing_utils.ts` now that data processing is been migrated to the flask endpoint
- Adds `"lib": ["es2019"]"` to `tsconfig.json` so typescript recognizes `array.flat`


server/services/datacommons.py
```diff
@@ -328,6 +328,23 @@ def get_property_labels(dcids):
     return results
 
 
+def property_values(entities, prop, dir):
```
jehangiramjad, line 331: It is a good idea to document what this function returns. Ideally, one shouldn't need to parse the function implementation to understand what's expected :)

jehangiramjad (commented):
Just one minor comment on the Python part of the code. I have not reviewed any of the react code.

server/services/datacommons.py
rogershijin: Got it! Will add

rogershijin: I actually just merged in master and this is no longer a diff.  I'll address this in the backend pr since it fits better there.

static/js/biomedical/protein/protein_protein_interaction_graph.tsx
```diff
@@ -0,0 +1,100 @@
...
+ */
+
+import axios from "axios";
+import _ from "lodash";
+import React from "react";
+
+import { drawProteinInteractionGraph } from "./chart";
+import { bioDcid, MultiLevelInteractionGraphData } from "./types";
+
+interface InteractionGraphProps {
```
chejennifer, line 28: nit: typically we want to name the props and the state: <componentName>Props and <componentName>State

static/js/biomedical/protein/protein_protein_interaction_graph.tsx
```diff
@@ -0,0 +1,100 @@
...
+    this.state = {
+      depth: DEFAULTS.DEPTH,
+      graphData: null,
+      scoreThreshold: DEFAULTS.SCORE_THRESHOLD,
+      numInteractions: DEFAULTS.MAX_INTERACTIONS,
+    };
+  }
+
+  /**
+   * Perform BFS and (re)draw graph
```
chejennifer, line 63: I don't think this performs BFS anymore?

static/js/biomedical/protein/protein_protein_interaction_graph.tsx
```diff
@@ -0,0 +1,100 @@
...
+   */
+  componentDidUpdate(prevProps: InteractionGraphProps): void {
+    // takes two calls to this method to draw the graph.
+    //  1) first call retrieves graph data from flask endpoint
+    //  2) second call draws the graph
+
+    // this branch executes on first call to this method
+    if (prevProps !== this.props) {
+      axios
```
chejennifer, line 72: - This also needs to be done in componentDidMount, otherwise on initial render of this component, data won't be fetched. 
- And since you'll be doing the same data fetch in two spots, should create a new private function in this component to do the data fetch.
- I think you also want to refetch data whenever state (except for graphData) changes? so maybe `if (!_.isEmpty(this.state.graphData) && !_.isEqual(prevState.graphData, this.state.graphData)) {draw and then return}` otherwise fetch data.

static/js/biomedical/protein/protein_protein_interaction_graph.tsx
```diff
@@ -0,0 +1,100 @@
...
+        .then((resp) => {
+          this.setState({ graphData: resp.data });
+        });
+      return;
+    }
+    // this branch executes on second call to this method
+    if (!_.isEmpty(this.state.graphData)) {
+      drawProteinInteractionGraph(CHART_ID, {
+        linkData: this.state.graphData.linkDataNested
+          .slice(0, this.state.depth + 1)
```
chejennifer, line 88: you shouldn't have to do any slicing anymore right (here and for nodeData)? since you're going to return the data for the current state?

chejennifer (commented):
nice!

static/js/biomedical/protein/protein_protein_interaction_graph.tsx
```diff
@@ -0,0 +1,100 @@
...
+/**
+ * Protein-protein interaction graph
+ */
+
+import axios from "axios";
+import _ from "lodash";
+import React from "react";
+
+import { drawProteinInteractionGraph } from "./chart";
+import { bioDcid, MultiLevelInteractionGraphData } from "./types";
```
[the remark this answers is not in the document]
rogershijin: Thanks for the review! Got it will fix!

static/js/biomedical/protein/protein_protein_interaction_graph.tsx
rogershijin: sounds good thanks!

static/js/biomedical/protein/protein_protein_interaction_graph.tsx
rogershijin: Oops that's right! thanks for catching!

static/js/biomedical/protein/protein_protein_interaction_graph.tsx
rogershijin: Ok will do! Currently the graph still shows up initially in a hacky way because the parent component rerenders after the page request comes back --> rerenders the children and calls componentDidUpdate but your way is much more robust.

static/js/biomedical/protein/protein_protein_interaction_graph.tsx
rogershijin: I was thinking it might be somewhat useful to keep one more case for when the user changes the depth while keeping the other inputs the same, and just slicing the list.  And when either of the other inputs change, we reload the graph at depth 3 (the max).  What do you think?

But for the current setup using just default values the slicing doesn't accomplish anything.

chejennifer: Ok sure, slicing if user changes the depth works. When other inputs change, why would you reload the graph at depth 3? it should just reload graph at whatever the current depth is?

static/js/biomedical/protein/protein_protein_interaction_graph.tsx
chejennifer: ah, ok, then maybe also add a check that one of either props or state has changed before running the code in componentDidUpdate

rogershijin: Oh ok will do!

static/js/biomedical/protein/protein_protein_interaction_graph.tsx
rogershijin: I was thinking we can just preload the graph always at the max depth, then we don't have to track the max depth the user has loaded so far.

static/js/biomedical/protein/protein_protein_interaction_graph.tsx
rogershijin: Ok I think fiddled a bit more and I think that for the original code I just got lucky in that I was using `prevProps !== this.props` previously instead of `_.isEqual(prevProps, this.props)`.  

I tried logging the following in the original version of the code:
```
  componentDidUpdate(prevProps: InteractionGraphProps): void {
    console.log('props', _.cloneDeep(prevProps), _.cloneDeep(this.props))
    console.log('1', _.isEqual(prevProps, this.props))
    console.log('2', prevProps === this.props)
    if (prevProps !== this.props) {
        //make request
            .then((resp) => {
              this.setState({ graphData: resp.data });
            });
        return;
    }
    // draw graph
```
and got
```
props {centerProteinDcid: 'bio/P53_HUMAN'} {centerProteinDcid: 'bio/P53_HUMAN'}
1 true
2 false
props {centerProteinDcid: 'bio/P53_HUMAN'} {centerProteinDcid: 'bio/P53_HUMAN'}
1 true
2 true
```
(so 1st call happens when Suhana's page request comes back, I guess a new props object is constructed so `prevProps !== this.props` is true, and then we update the state and get a 2nd call with the same props.)

I think basically we want a way to avoid calling `fetchData` twice (this can happen if we make the request in `componentDidMount` but it doesn't resolve by the time of the 1st call to `componentDidUpdate` from Suhana's page request coming back, so `state.graphData` is empty and we `fetchData` again).  Perhaps we can leave the first `fetchData` to componentDidMount, and in `componentDidUpdate` first check if `_.isEmpty(state.graphData)` and then return, and then check if the user input parts of the state are all the same as previously, and if so draw, and otherwise `fetchData`?

rogershijin: Oops I think another problematic case is that if this component's `fetchData` arrives before Suhana's, then we draw the component twice.

rogershijin: Ok I fixed this by adding a `drawn` property to the state.  So drawn is false initially and also set to false whenever we've fetched new data, and set to true when we call the draw function.

static/js/biomedical/protein/protein_protein_interaction_graph.tsx
rogershijin: Ok for now I've taken out the slicing since it doesn't really make sense without user input.

static/js/biomedical/protein/protein_protein_interaction_graph.tsx
chejennifer: I don't think the drawn flag is needed .. couldn't we just do this in the componentDidUpdate?
```
if (_.isEqual(prevProps, this.props) && _.isEqual(prevState, this.state) {
return
}
(!_.isEmpty(this.state.graphData) && !_.isEqual(prevState.graphData, this.state.graphData)) {
draw and then return
}
fetch data
```

rogershijin: Oh! yes sorry this works thanks for writing it out!

chejennifer (approved):
thanks for the changes!

PR #1689, opened 2022-07-28
User page with minimal authenticating mechanism

shifucun:
Authentication uses google_auth_oauthlib library for oauth2 flow.

This requires configuration of oauth2 client in https://pantheon.corp.google.com/apis/credentials and storing the secret in GCP Secret Manager.


server/__init__.py
```diff
@@ -149,6 +158,30 @@ def create_app():
         app.config['MAPS_API_KEY'] = secret_response.payload.data.decode(
             'UTF-8')
 
+    if cfg.ADMIN:
+        secret_client = secretmanager.SecretManagerServiceClient()
+        secret_name = secret_client.secret_version_path(cfg.SECRET_PROJECT,
```
shifucun, line 163: @Fructokinase this is another resource for website, where the oauth2 client needs to be created first and stored in Secret Manager

server/configmodule.py
```diff
@@ -96,6 +99,7 @@ class MinikubeConfig(Config):
 
 class LocalConfig(Config):
     LOCAL = True
+    ADMIN = True
```
beets, line 102: the local private dc's should have this set back to False since they won't have access to secret manager.

server/routes/user.py
```diff
@@ -0,0 +1,80 @@
+# Copyright 2022 Google LLC
```
beets, line 1: Should we add a /admin subdirectory under routes for all admin / user related pages?

server/routes/user.py
```diff
@@ -0,0 +1,80 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import requests
+from flask import Blueprint, current_app, session, abort, redirect, request
+from google.oauth2 import id_token
+from cachecontrol import CacheControl
+import google.auth.transport.requests
+
+bp = Blueprint('user', __name__)
```
beets, line 21: For these routing modules, I like having a url prefix that matches the file path so it's easy to know which module responds to a particular path.

server/routes/user.py
```diff
@@ -0,0 +1,80 @@
...
+    return redirect('/user/login')
+
+
+def login_is_required(function):
+
+    def wrapper(*args, **kwargs):
+        if 'google_id' not in session:
+            return redirect('/user/login')
+        else:
+            return function()
```
beets, line 67: I'm not too familiar with OAuth - what happens if it falls through to this case?

beets (approved):
this is neat!

server/configmodule.py
shifucun: Updated this with refactored base class for local and private instance now.

server/routes/user.py
shifucun: We don't have subdirectory for pages (only has /api to hold all apis), looks like each file handles routes for one page. Maybe keep this as is to be consistent?

server/routes/user.py
shifucun: Done

server/routes/user.py
shifucun: This is more of a decorator, so falls through would execute the original function to load the requested page.

shifucun (commented):
thanks for review!

PR #1690, opened 2022-07-29
[Protein browser] Add depth input to protein-protein interaction graph

rogershijin:
`protein_protein_interaction_graph.tsx`:
- adds form group containing user input/label for depth (adapted from example provided by @chejennifer )
- user requests now load at the max depth of 3 and render at depth user sets

`chart.ts`:
- adds line to delete children of drawing div to clear previous drawing on multiple successive calls

Graph + input component will look like this when backend PR is merged:
image


static/js/biomedical/protein/protein_protein_interaction_graph.tsx
```diff
@@ -88,13 +98,34 @@ export class ProteinProteinInteractionGraph extends React.Component<
...
+      return <div></div>;
+    }
+    return (
+      <>
+        <div id={CHART_ID}></div>
+        <FormGroup>
+          <Label for={DEPTH_INPUT_ID}>Depth</Label>
+          <Input
+            id={DEPTH_INPUT_ID}
+            className={DEPTH_INPUT_ID}
```
chejennifer, line 111: I don't think it's necessary to set both the id and className here, you could just use the id to select for the element

static/js/biomedical/protein/protein_protein_interaction_graph.tsx
```diff
@@ -88,13 +98,34 @@ export class ProteinProteinInteractionGraph extends React.Component<
   }
 
   render(): JSX.Element {
-    return <div id={CHART_ID}></div>;
+    if (this.state.graphData === null) {
+      return <div></div>;
```
chejennifer, line 102: instead of returning an emtpy div, usually we would just want to return null

chejennifer (commented):
nice job!

static/js/biomedical/protein/protein_protein_interaction_graph.tsx
rogershijin: Thanks for the review! Got it will fix!

chejennifer (approved):
Thanks for the changes!

PR #1692, opened 2022-07-30
[Biomedical] Patch tooltip bug

rogershijin:
Currently tooltips on biomedical pages sometimes do not appear on hover (initially discovered by @suhana13, verified by @beets).  One way to reproduce is to 
1. Hover over a tooltipped object.  A tooltip should appear above and to the left.
3. Leave the object via the bottom right.  Tooltip should disappear somewhere around the top left of the object.
4. Circle back to hover over the region the tooltip last occupied, avoiding all tooltipped objects.  If this region intersects with the object, hovering over the area of intersection will not activate the tooltip.

The problem is that the tooltip is made to disappear by setting its opacity to 0, so it is still present but invisible on mouseout and can block mouse hover over intended targets.

A fix is to move the tooltip on mouseout somewhere out of the way.  @beets suggested a negative coordinate, so here we propose moving the tooltip to -1000px, -1000px on mouseout.


chejennifer (commented):
This feels slightly hacky, can we try:
`display: block` on mouseOver and `display: none` on mouseOut
or 
`visibility: visible` on mouseOver and `visibility: hidden` on mouseOut

rogershijin:
Sorry for the extreme lateness on this! `display: block`/`display: none` worked after some minor adjustments! These were originally being overriden by `display: inline` in the css so I removed this, and I found that the default opacity seemed to be 0 so I added `opacity: 1`.

chejennifer (approved):
Thanks for the changes!

PR #1693, opened 2022-07-31
[Protein Browser] PPI Aesthetics

rogershijin:
- Clamps node x, y coordinates to prevent departure from chart bounding box from either dragging or repulsion
- Reverses color scheme and adjust node radii so page protein is most prominent
- Adds collision detection so nodes don't overlap
- Tunes repulsion strengths so nodes spread out

image


static/js/biomedical/protein/chart.ts
```diff
@@ -205,18 +205,25 @@ const ERROR_POINT_POSITION_Y3 = 50;
...
-  "mistyrose",
-  "peachpuff",
   "lightCoral",
+  "peachpuff",
+  "mistyrose",
   "lightsalmon",
 ];
 
+// decrease node repulsion strength with increasing depth
+const NODE_REPULSION_STRENGTHS = [-500, -150, -30, -20];
```
chejennifer, line 221: curious why the repulsion strength changes aren't even across depths & how did you come up with these numbers?

chejennifer (approved):
nice job!

static/js/biomedical/protein/chart.ts
rogershijin: Thanks for the review! And I guess this is just personal preference but I kind of like having some negative space around the center node and having the higher-depth nodes be clumped together on the outside.  The default repulsion strength is -20 and the -500 I got from [this stackoverflow](https://stackoverflow.com/questions/29639270/space-out-nodes-evenly-around-root-node-in-d3-force-layout) which I'll link.

PR #1694, opened 2022-08-01
[Disease Browser]: Add chemical compound disease contraindication table 

suhana13:
Screen Shot 2022-08-01 at 11 19 21 AM
For the disease of interest, this table displays the associated chemical compounds contraindicated in this disease, along with compound id, link to the graph browser view and drug source name


static/js/biomedical/disease/data_processing_utils.ts
```diff
@@ -232,3 +233,84 @@ export function getCompoundDiseaseTreatment(
   }
   return rawData;
 }
+
+/**
+ * Fetches the chemical compound disease contraindication data
+ * @param data
+ * @returns
+ */
+export function getCompoundDiseaseContraindication(
```
chejennifer, line 242: The logic in this function is exactly the same as the last function `getCompoundDiseaseTreatment`, except for the check for the typeField (line 305) and the property (line 275), could we extract out a shared helper function that takes those two strings as arguments? I know the return type is also slightly different, but the shared function can return the raw value and then the individual functions can map that value to what is needed.

static/js/biomedical/disease/chart.ts
```diff
@@ -63,6 +63,18 @@ export interface CompoundDiseaseTreatmentData {
   // FDA clinical phase for which the compound has been studied
   clinicalPhaseNumber: number;
 }
+
+export interface CompoundDiseaseContraindicationData {
```
chejennifer, line 67: make the types.ts file and put this there. Can defer to next PR, but please make that PR next.

static/js/biomedical/disease/drug_table.tsx
```diff
@@ -1,17 +1,28 @@
...
+export interface CompoundDiseaseTreatmentProps {
   data: CompoundDiseaseTreatmentData[];
 }
 
-export function DrugTable(props: DrugTableProps): JSX.Element {
+export interface CompoundDiseaseContraindicationProps {
+  data: CompoundDiseaseContraindicationData[];
+}
+
+export function CompoundDiseaseTreatmentTable(
```
chejennifer, line 17: Each file should only have one component. Please put the two different components into different files. Although, the two tables are pretty much exactly the same, so I think better, you should make one DrugTable component & use it for both the CompoundDiseaseTreatmentTable and the CompoundDiseaseContraindicationTable

chejennifer (commented):
nice!

static/js/biomedical/disease/data_processing_utils.ts
suhana13: That's a good point. I'm still trying to understand how do we return the raw data because the two return data types/interfaces would be different. And what part of the code would we want in the individual functions?

chejennifer: The return type can be something like (w/ better naming)

```
DataType {
value: string;
id: string;
name: string;
node: string;
}
```

And each individual function would just call the shared function & do a map on the result to convert the value to the form needed.

static/js/biomedical/disease/drug_table.tsx
suhana13: I think having one component makes more sense. But then, how would I change the props datatype and column names for different tables?

static/js/biomedical/disease/chart.ts
suhana13: I'll definitely do it on the next PR!

static/js/biomedical/disease/drug_table.tsx
chejennifer: we could have props take in a list of column headers and an array of objects with keys that match the headers eg.

```
interface DrugTableProps {
columns: string[];
data: { [columnKey: string]: string }[]
}
```
or can also make columns a map where key is a columnId and value is the display string.

Please also add comments for these props

suhana13: As per our discussion, I made column an array of objects and data, just a generic array of objects because it can belong to either the treatment type or contraindication type. Now that the indices are similar for the two objects, I was thinking we could just make column a simple array, instead of doing a name map. The current refactored code works for the two tables, but if there are any more suggestions, I'd be happy to implement them

chejennifer (commented):
thanks for the refactor

static/js/biomedical/disease/page.tsx
```diff
@@ -79,7 +96,21 @@ export class Page extends React.Component<PagePropType, PageStateType> {
         <br></br>
         <div>
           <div id="table"></div>
-          <DrugTable data={chemicalCompoundDiseaseTreatment} />
+          <DrugTreatmentTable
+            // columns = [{id:"node", name:"Parent Node"}, {id:"id", name:"Compound ID"}, {id:"name", name:"Compound Name"}, {id:"clinicalPhaseNumber", name:"FDA Clinical Phase"}]
```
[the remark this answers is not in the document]
suhana13: Done!

static/js/biomedical/disease/page.tsx
```diff
@@ -67,6 +68,22 @@ export class Page extends React.Component<PagePropType, PageStateType> {
...
+      { id: "node", name: "Parent Node" },
+      { id: "id", name: "Compound ID" },
+      { id: "name", name: "Compound Name" },
+      { id: "clinicalPhaseNumber", name: "FDA Clinical Phase" },
+    ];
+    const diseaseContraindicationColumns = [
+      { id: "node", name: "Parent Node" },
+      { id: "id", name: "Compound ID" },
+      { id: "name", name: "Compound Name" },
+      { id: "clinicalPhaseNumber", name: "Drug Source" },
```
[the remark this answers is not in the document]
suhana13: Good catch! Fixed it!

static/js/biomedical/disease/drug_table.tsx
```diff
@@ -1,37 +1,48 @@
...
+} from "./chart";
 
+export interface DrugTableColumnType {
+  id: string;
+  name: string;
+}
 export interface DrugTableProps {
-  data: CompoundDiseaseTreatmentData[];
+  columns: DrugTableColumnType[];
+  data: Record<never, never>[];
```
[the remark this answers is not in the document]
suhana13: Instead of writing an additional function which converts `CompoundDiseaseContraindicationData` and `CompoundDiseaseTreatmentData`, I created another generic interface called `DrugDataType`, which removes the ambiguity with the use of index signature syntax.

static/js/biomedical/disease/data_processing_utils.ts
```diff
@@ -219,16 +228,75 @@ export function getCompoundDiseaseTreatment(
...
+  // checks if the data is empty and for null values
+  if (
+    _.isEmpty(data) ||
+    _.isEmpty(data.nodes) ||
+    _.isEmpty(data.nodes[0].neighbors)
+  ) {
+    return [];
+  }
+  const rawData = getChemicalCompoundData(data);
+  let modifiedData: CompoundDiseaseTreatmentData[] = [];
```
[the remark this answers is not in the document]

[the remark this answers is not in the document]

static/js/biomedical/disease/data_processing_utils.ts
```diff
@@ -219,16 +228,75 @@ export function getCompoundDiseaseTreatment(
...
+  modifiedData = rawData
+    .filter((e) => e.type === "ChemicalCompoundDiseaseTreatment")
```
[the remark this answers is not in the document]
suhana13: Replaced `e` with `element`

static/js/biomedical/disease/data_processing_utils.ts
```diff
@@ -219,16 +228,75 @@ export function getCompoundDiseaseTreatment(
...
+
+/**
+ * Fetches the chemical compound disease treatment data
+ * @param data
+ * @returns
+ */
+export function getCompoundDiseaseTreatment(
+  data: GraphNodes
+): CompoundDiseaseTreatmentData[] {
+  // checks if the data is empty and for null values
```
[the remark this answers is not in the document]
suhana13: You're right! Removed the redundant check!

static/js/biomedical/disease/drug_table.tsx
```diff
@@ -1,37 +1,48 @@
...
-export function DrugTable(props: DrugTableProps): JSX.Element {
+export function DrugTreatmentTable(props: DrugTableProps): JSX.Element {
   // takes the top 10 chemical compound disease associations for greater than 10 values
-  const drugTableData = props.data.slice(0, 10);
-  // sorts the array based on FDA clinical phase number
-  drugTableData.sort((a, b) => b.clinicalPhaseNumber - a.clinicalPhaseNumber);
+  const drugTreatmentTableData = props.data.slice(0, 10);
+  // take the drug treatment table columns
+  const drugTreatmentColumns = props.columns;
+  const nameArray = drugTreatmentColumns.map(function (el) {
```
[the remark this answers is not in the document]
suhana13: yeah, great idea! implemented it!

static/js/biomedical/disease/drug_table.tsx
```diff
@@ -1,37 +1,48 @@
...
-
-export function DrugTable(props: DrugTableProps): JSX.Element {
+export function DrugTreatmentTable(props: DrugTableProps): JSX.Element {
   // takes the top 10 chemical compound disease associations for greater than 10 values
-  const drugTableData = props.data.slice(0, 10);
-  // sorts the array based on FDA clinical phase number
-  drugTableData.sort((a, b) => b.clinicalPhaseNumber - a.clinicalPhaseNumber);
+  const drugTreatmentTableData = props.data.slice(0, 10);
+  // take the drug treatment table columns
+  const drugTreatmentColumns = props.columns;
```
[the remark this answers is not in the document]
suhana13: Right! removed it

static/js/biomedical/disease/drug_table.tsx
```diff
@@ -1,37 +1,48 @@
...
+import {
+  CompoundDiseaseContraindicationData,
+  CompoundDiseaseTreatmentData,
+} from "./chart";
 
+export interface DrugTableColumnType {
+  id: string;
+  name: string;
+}
 export interface DrugTableProps {
```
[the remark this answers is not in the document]
suhana13: Changed it to DrugTreatmentTableProps

static/js/biomedical/disease/drug_table.tsx
```diff
@@ -1,37 +1,48 @@
...
+          <td>{nameArray[0]}</td>
+          <td>{nameArray[1]}</td>
+          <td>{nameArray[2]}</td>
+          <td>{nameArray[3]}</td>
         </tr>
       </thead>
       <tbody>
-        {drugTableData.map((item, idx) => {
+        {drugTreatmentTableData.map((item, idx) => {
+          const items = Object.entries(item);
```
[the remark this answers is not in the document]
suhana13: I implemented your first suggestion to use column.id as the index wouldn't account for the change in order. For the second suggestion, I implemented an if-else condition to account for the hyperlinked node

static/js/biomedical/disease/drug_table.tsx
```diff
@@ -1,37 +1,59 @@
...
+  // compound name
+  name: string;
+  // clinical phase number or drug source
+  [key: string]: any;
+}
+export interface DrugTreatmentTableProps {
+  // stores the column id and column header name
+  columns: DrugTableColumnType[];
+  // stores the data in a generic DrugDataType interface
+  data: DrugDataType[];
```
chejennifer, line 24: Sorry for changing things up on you, but now that I think about it, it might be better to just make the type for data : `CompoundDiseaseContraindicationData[] | CompoundDiseaseTreatmentData[]` and I think everything should still work the same way

static/js/biomedical/disease/drug_table.tsx
```diff
@@ -1,37 +1,59 @@
...
+  node: string;
+  // compound id
+  id: string;
+  // compound name
+  name: string;
+  // clinical phase number or drug source
+  [key: string]: any;
+}
+export interface DrugTreatmentTableProps {
+  // stores the column id and column header name
```
chejennifer, line 21: please also mention that this stores the order of the columns and that column id should match the keys of the objects in the data

static/js/biomedical/disease/drug_table.tsx
```diff
@@ -1,37 +1,59 @@
 import React from "react";
 
 import { GRAPH_BROWSER_REDIRECT } from "../bio_charts_utils";
-import { CompoundDiseaseTreatmentData } from "./chart";
-
-export interface DrugTableProps {
-  data: CompoundDiseaseTreatmentData[];
+export interface DrugTableColumnType {
```
chejennifer, line 4: Let's keep everything consistent and also name this DrugTreatmentTableColumn

chejennifer (commented):
thanks for the changes!

static/js/biomedical/disease/drug_table.tsx
suhana13: Great! Implemented the suggestion:)

static/js/biomedical/disease/drug_table.tsx
suhana13: Done!

static/js/biomedical/disease/drug_table.tsx
suhana13: Done!

static/js/biomedical/disease/drug_table.tsx
```diff
@@ -1,37 +1,54 @@
...
 import { GRAPH_BROWSER_REDIRECT } from "../bio_charts_utils";
-import { CompoundDiseaseTreatmentData } from "./chart";
-
-export interface DrugTableProps {
-  data: CompoundDiseaseTreatmentData[];
+import {
+  CompoundDiseaseContraindicationData,
+  CompoundDiseaseTreatmentData,
+} from "./chart";
+export interface DrugTreatmentTableColumn {
```
chejennifer, line 8: add empty line above this line

static/js/biomedical/disease/drug_table.tsx
```diff
@@ -1,37 +1,54 @@
...
+  // the column id
+  id: string;
+  // the column display name or header
+  name: string;
 }
-
-export function DrugTable(props: DrugTableProps): JSX.Element {
+export interface DrugTreatmentTableProps {
+  // stores the column id and column name
+  // retains the order of the columns and column id matches the key of data
```
chejennifer, line 16: nit: ... and column ids should match the keys of the objects in the data array

static/js/biomedical/disease/drug_table.tsx
```diff
@@ -1,37 +1,54 @@
...
+  columns: DrugTreatmentTableColumn[];
+  // stores the data in a generic DrugDataType interface
```
chejennifer, line 18: please update this comment

static/js/biomedical/disease/drug_table.tsx
```diff
@@ -1,37 +1,54 @@
...
-              <td>{item.id}</td>
-              <td>{item.name}</td>
-              <td>{item.clinicalPhaseNumber}</td>
+              {props.columns.map((column) => {
+                const element = item[column.id];
+                // if column id is node, then display a hyperlinked table element
+                if (column.id === "node") {
+                  return (
+                    <td>
+                      <a href={GRAPH_BROWSER_REDIRECT + element}>Node</a>
```
chejennifer, line 45: I think the value in the table should show the actual node instead of the word "Node"?

static/js/biomedical/disease/drug_table.tsx
```diff
@@ -1,37 +1,54 @@
...
           return (
             <tr key={idx}>
-              <td>
-                <a href={GRAPH_BROWSER_REDIRECT + item.node}>Node</a>
-              </td>
-              <td>{item.id}</td>
-              <td>{item.name}</td>
-              <td>{item.clinicalPhaseNumber}</td>
+              {props.columns.map((column) => {
+                const element = item[column.id];
```
chejennifer, line 40: make `const element = item[column.id] || ""` & you can just use this in line 49 too

chejennifer (commented):
Thanks for the updates!

static/js/biomedical/disease/drug_table.tsx
suhana13: Done!

static/js/biomedical/disease/drug_table.tsx
suhana13: Added that!

static/js/biomedical/disease/drug_table.tsx
suhana13: Done!

static/js/biomedical/disease/drug_table.tsx
suhana13: Nope, we intentionally want the word "Node" to be showed as the actual node values are too long and incomprehensible:(

static/js/biomedical/disease/drug_table.tsx
suhana13: Done!

static/js/biomedical/disease/drug_table.tsx
chejennifer: ah, got it, ok, makes sense

static/js/biomedical/disease/drug_table.tsx
chejennifer: this isn't updated? this isn't a DrugDataType interface anymore

suhana13: My bad! I confused it with the previous comment. It's updated now!

chejennifer:
Please merge master and fix the failing tests. You should be locally running client tests and checking for failures

chejennifer (approved):
thanks for the changes

PR #1698, opened 2022-08-03
[Protein Browser] Add user input for number of interactors and confidence threshold, update button

rogershijin:
- Adds user inputs for number of interactors and confidence threshold
- Adds an update button.  Graph now updates only after clicking this button.
- Adds a `loading` flag to the state which is false when the user is adjusting inputs and set to true upon clicking the update button.  A future PR will add a spinner to the update button animated whenever `loading` is true.
- Inputs are currently subject to both lower and upper limits.  Upper limits will be removed once table view is introduced.

image

I had also tried out this layout which doesn't have the labels hanging over the input components, at the cost of label clarity
image


static/js/biomedical/protein/protein_protein_interaction_graph.tsx
```diff
@@ -104,19 +119,67 @@ export class ProteinProteinInteractionGraph extends React.Component<
...
+                max={MAX_SCORE_THRESHOLD}
+                step={0.1}
+                onChange={(e) => {
+                  this.setState({ scoreThreshold: Number(e.target.value) });
+                }}
+                value={this.state.scoreThreshold}
+              />
+            </FormGroup>
+          </Col>
+          <Col className="form-group align-self-end" md={2}>
```
chejennifer, line 172: I don't think these class names are necessary since your'e not doing anything with them

static/js/biomedical/protein/protein_protein_interaction_graph.tsx
```diff
@@ -63,6 +73,7 @@ export class ProteinProteinInteractionGraph extends React.Component<
     this.state = {
       depth: DEFAULTS.DEPTH,
       graphData: null,
+      loading: true,
```
chejennifer, line 76: when using loading, we usually do:

- set it to false in the constructor
- set it to true right before data fetch is called (so probably first line in the data fetch function)
- set to to false when the promise completes (in the `.then()` and the `.catch()`)

static/js/biomedical/protein/protein_protein_interaction_graph.tsx
rogershijin: Oh I forgot to link a reference but I think these might be handled by reactstrap - before adding these the button was vertically aligning with the labels instead of the text fields as in [https://stackoverflow.com/questions/40878566/align-button-with-input-forms-with-labels](here).  What do you think?

static/js/biomedical/protein/protein_protein_interaction_graph.tsx
rogershijin: Ah ok thanks for letting me know! I think probably I wanna come back to loading in a future PR then when I add the loading spinner.  For the update button behavior I can think of two possibilities then:
1. We rename the current `loading` to `toUpdate` (so later on we would have both a `loading` flag and a `toUpdate` flag)
2. We create 3 new properties in the state for each of the text inputs, and the update button sets the original properties `onClick`:
```
this.state.depth  = this.state.depthInput;
this.state.numInteractions = this.state.numInteractionsInput;
this.state.depth = this.state.scoreThresholdInput;
```

What do you think?

chejennifer: Ah, I see what you're trying to do now. Yes, let's come back to the real loading in a future PR. For here, I think we could:

- on button clicked, call fetchData()
- in this didComponentUpdate function, just only fetch data when `!_.isEqual(prevProps, this.props)`

static/js/biomedical/protein/protein_protein_interaction_graph.tsx
chejennifer: ok, fine to keep then.

rogershijin: thanks a lot! will do

static/js/biomedical/protein/protein_protein_interaction_graph.tsx
rogershijin: wow that's clean!! thanks I'll do this

rogershijin: Oh I realized we can also do `fetchData` in `componentDidMount` and remove it completely from `componentDidUpdate`.

chejennifer: we want to keep it in componentDidUpdate in case the proteinDcid changes but the component is never unmounted (although I'm not sure it'll happen in this case, but good to have it there still)

rogershijin: Ah good point will do!

chejennifer: also, I don't think you need the first check (return if there is no change in props and state) in the componentDidUpdate function anymore

rogershijin: oh! yes another nice catch!

chejennifer (approved):
Thanks for the changes!

PR #1699, opened 2022-08-03
Use mixer API key for local development

shifucun:
We plan to enforce API key for the V1 mixer API. 

This won't affect website deployment in GKE as the the mixer is an internal service within pod, so it does not need API key restriction. However, local website development requires talking to an instance of mixer. For example, regular local dev talks to the autopush mixer instance.

This PR allows a local website instance to fetch the API key from the corresponding mixer project and use it when sending requests.

A prerequisite of this is to creating the API key and store it in the secret manager. I have done this manually for the two mixer projects involved in this PR.

### DIFF
--- a/server/__init__.py
+++ b/server/__init__.py
@@ -182,6 +182,14 @@ def create_app():
     if app.config['LOCAL']:
         os.environ['OAUTHLIB_INSECURE_TRANSPORT'] = '1'
 
+    if app.config['API_PROJECT']:
+        secret_client = secretmanager.SecretManagerServiceClient()
+        secret_name = secret_client.secret_version_path(cfg.API_PROJECT,
+                                                        'mixer-api-key',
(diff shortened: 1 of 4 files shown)


server/__init__.py
```diff
@@ -182,6 +182,14 @@ def create_app():
     if app.config['LOCAL']:
         os.environ['OAUTHLIB_INSECURE_TRANSPORT'] = '1'
 
+    if app.config['API_PROJECT']:
+        secret_client = secretmanager.SecretManagerServiceClient()
+        secret_name = secret_client.secret_version_path(cfg.API_PROJECT,
```
shifucun, line 187: @Fructokinase This is another resource that has to be added to the mixer project.

chejennifer (approved):
this is cool!

PR #1703, opened 2022-08-05
[Download] Add tests

chejennifer:
- add webdriver and client tests for the download tool


shifucun (approved):
Thanks for adding the tests. Very thorough and clean!

PR #1704, opened 2022-08-05
[Protein Browser] Add toggle and download buttons

rogershijin:
- Adds toggle between graph and table views based on example provided by @chejennifer 
- Adds button for downloading CSV of links
- Adds `react-csv` dependency

image
image


chejennifer:
Haven't taken a look at the code yet, but in the second screenshot where the background for "Graph View" is different, at first glance, it kind of looks like a toggle between "Graph View" and CSV. Can we try:
a. choose a color for the buttons that's more of a contract from the background and then increase the space between the two buttons a little bit so that it's clearly 2 different things. Also keep the button color the same for "Graph View" and "Table View" 
b. or move the download csv button to below the graph/chart

rogershijin:
Thanks for the tips! will do!

static/js/biomedical/protein/protein_protein_interaction_graph.tsx
```diff
@@ -111,14 +125,47 @@ export class ProteinProteinInteractionGraph extends React.Component<
...
+        <Row className="justify-content-end">
+          <ButtonGroup>
+            <Button
+              className="btn btn-sm btn-light shadow-none"
+              onClick={() =>
+                this.setState({ showTableView: !this.state.showTableView })
+              }
+            >
+              <i className="material-icons align-middle">
+                {this.state.showTableView ? "hub" : "table_restaurant"}
```
chejennifer, line 142: maybe we can use this icon instead for table: https://fonts.google.com/icons?selected=Material+Symbols+Outlined:table_chart

static/js/biomedical/protein/protein_protein_interaction_graph.tsx
```diff
@@ -111,14 +125,47 @@ export class ProteinProteinInteractionGraph extends React.Component<
...
+              </i>
+              <span>
+                {this.state.showTableView ? " Graph View" : " Table View"}
+              </span>
+            </Button>
+            <CSVLink
+              data={data}
```
chejennifer, line 149: curious what the csv ends up looking like

static/js/biomedical/protein/protein_protein_interaction_graph.tsx
```diff
@@ -83,8 +95,10 @@ export class ProteinProteinInteractionGraph extends React.Component<
 
   componentDidUpdate(prevProps: Props, prevState: State): void {
     if (
+      !this.state.showTableView &&
```
chejennifer, line 98: there seems to be a lot of logic to determine whether or not to draw a graph, maybe make a helper function like `shouldDrawGraph()` that returns a boolean to replace the stuff in the `if`

static/js/biomedical/protein/protein_protein_interaction_graph.tsx
rogershijin: great selection! yes will do!

static/js/biomedical/protein/protein_protein_interaction_graph.tsx
rogershijin: It looks like this:
image

One TODO is to make the score column header say "Confidence" instead.  Would you like me to address this now?

static/js/biomedical/protein/protein_protein_interaction_graph.tsx
rogershijin: ah sounds good will do!

rogershijin:
Thanks for the review! Updated buttons look like this now!  What do you think?
image
image

static/js/biomedical/protein/protein_protein_interaction_graph.tsx
chejennifer: is it an easy change to make the score column header say "Confidence" instead? If it's an easy change, let's do it now, if not, this is ok too

chejennifer (approved):
thanks for the changes!

static/js/biomedical/protein/protein_protein_interaction_graph.tsx
rogershijin: yup just pushed something to make the column headers the same for the displayed table/downloaded table!

rogershijin: The changes are that `CSVLInk` now takes a `headers` prop

```
 <CSVLink
              data={data}
              headers={COLUMNS.map(({ Header, accessor }) => ({
                label: Header,
                key: accessor,
              }))}
              filename={`${this.props.centerProteinDcid
                .replace("bio/", "")
                .toLowerCase()}_links.csv`}
              enclosingCharacter={""}
            >
```
 and that I now export `COLUMNS` form `protein_protein_interaction_table.tsx` and moved the type assertion into `Column<InteractionLink>[]` later

```
= useTable({
    columns: COLUMNS,
    data,
  } as { columns: Column<InteractionLink>[]; data: InteractionLink[] });
```

chejennifer: nice! this looks good, feel free to merge!

PR #1707, opened 2022-08-06
dataset selector

n-h-diaz:
Screen Shot 2022-08-06 at 11 05 42 AM


server/services/datacommons.py
```diff
@@ -531,7 +531,7 @@ def get_statvar_group(stat_var_group, places):
     url = API_ROOT + API_ENDPOINTS['get_statvar_group']
     req_json = {
         'stat_var_group': stat_var_group,
-        'places': places,
+        'entities': places,
```
shifucun, line 534: change the argument "places" to "entities" as well?

static/js/tools/stat_var/page.tsx
```diff
@@ -44,7 +53,9 @@ class Page extends Component<unknown, PageStateType> {
     this.state = {
       description: "",
       displayName: "",
+      entities: [],
```
shifucun, line 56: since you have both "entities" and "sources", should entities be places here?

static/js/tools/stat_var/info.tsx
```diff
@@ -24,12 +24,13 @@ class Info extends Component {
   render(): JSX.Element {
     return (
       <div id="placeholder-container">
-        <h1 className="mb-4">Statistical Variable Explorer</h1>
         <p>
```
shifucun, line 27: Why is the h1 title removed?

server/services/datacommons.py
n-h-diaz: done

static/js/tools/stat_var/page.tsx
n-h-diaz: "entities" refers to the source/dataset that the user selects, while "sources" (updated to sourceMap) refers to the options in the autocomplete - hope this makes more sense, but I can change if not!

static/js/tools/stat_var/info.tsx
n-h-diaz: this was moved to the parent component (page.tsx) because now we always show this header regardless of whether a stat var is selected or not

n-h-diaz:
thanks for the review!

static/js/tools/stat_var/dataset_selector.tsx
```diff
@@ -0,0 +1,231 @@
...
+const MAX_SUGGESTIONS = 5;
+
+interface DatasetSelectorProps {
+  // Filter StatVarHierarchy to stat vars that exist for the entities.
+  filterStatVars: (entities: NamedPlace[]) => void;
+  // Map of source name to dcid.
+  sourceMap: Record<string, string>;
+}
+
+function simplify(s: string): string {
```
chejennifer, line 37: add a comment about what the simplification is?

static/js/tools/stat_var/dataset_selector.tsx
```diff
@@ -0,0 +1,231 @@
...
+ * Component to select a source and dataset.
+ */
+
+import axios from "axios";
+import React, { useEffect, useState } from "react";
+import { Card, Container, CustomInput } from "reactstrap";
+
+import { NamedPlace } from "../../shared/types";
+
+const PREFIX = "dataset-selector";
```
chejennifer, line 27: nit: maybe CSS_PREFIX or CSS_SELECTOR_PREFIX or something mentioning what the prefix is for

static/js/tools/stat_var/dataset_selector.tsx
```diff
@@ -0,0 +1,231 @@
...
+  async function updateDatasets(source: string): Promise<void> {
+    if (!(source in props.sourceMap)) {
+      return;
+    }
+    const dcid = props.sourceMap[source];
+    const datasetsPromise = await axios.get(
+      `/api/browser/propvals/isPartOf/${dcid}`
+    );
+    const currentDatasets = [];
+    for (const dataset of datasetsPromise?.data?.values?.in) {
```
chejennifer, line 75: curious why doing `await` and `datasetsPromise?` instead of the usual 
```
datasetsPromise = axios.get(...)
datasetsPromise.then(...).catch(...)
```

static/js/tools/stat_var/dataset_selector.tsx
```diff
@@ -0,0 +1,231 @@
...
+          rank: simplify(s).indexOf(simplify(input)),
+        };
+      })
+      .sort((a, b): number => {
+        return a.rank - b.rank;
+      })
+      .filter((s) => s.rank > -1)
+      .map((s) => s.name);
+    if (newSuggestions.length > MAX_SUGGESTIONS) {
+      newSuggestions = newSuggestions.slice(0, MAX_SUGGESTIONS);
```
chejennifer, line 103: I think you can just slice it without checking that length is greater than MAX_SUGGESTIONS because if length is shorter, slice will just slice up to length of the array.

pradh:
> Screen Shot 2022-08-06 at 11 05 42 AM

This is exciting!  Some usability questions:

0.  Might be nice to see some screencasts of the e2e usage?
1. Does source text box auto complete?
2. Can user select only the dataset?

static/js/tools/stat_var/dataset_selector.tsx
```diff
@@ -0,0 +1,231 @@
...
+                const name = e.currentTarget.value
+                  ? e.currentTarget.innerText
+                  : userInput;
+                const dcid = e.currentTarget.value
+                  ? e.currentTarget.value
+                  : props.sourceMap[userInput];
+                props.filterStatVars([{ name, dcid }]);
+              }}
+            >
+              <option value="">Optionally select a dataset</option>
```
chejennifer, line 217: are users supposed to do one of search for a source OR select a dataset?

static/js/tools/stat_var/dataset_selector.tsx
```diff
@@ -0,0 +1,231 @@
...
+    Object.keys(props.sourceMap)
+  );
+  const [showSuggestions, setShowSuggestions] = useState(false);
+  const [suggestionsWidth, setSuggestionsWidth] = useState(null);
+  const [userInput, setUserInput] = useState("");
+
+  const handleResize = () => {
+    const width = document.getElementById(`${PREFIX}-ac`).offsetWidth;
+    if (width !== suggestionsWidth) {
+      setSuggestionsWidth(width);
```
chejennifer, line 54: Why does this width needs to be manually calculated and updated - does CSS/flexbox not ensure the suggestions width matches the input width?

chejennifer (commented):
this is cool!

static/js/tools/stat_var/dataset_selector.tsx
n-h-diaz: done

static/js/tools/stat_var/dataset_selector.tsx
n-h-diaz: changed to CSS_PREFIX

static/js/tools/stat_var/dataset_selector.tsx
n-h-diaz: changed to use this pattern instead

static/js/tools/stat_var/dataset_selector.tsx
n-h-diaz: thanks!! updated

static/js/tools/stat_var/dataset_selector.tsx
n-h-diaz: dataset is subset of source, so they can first search for source and then optionally further filter by dataset

static/js/tools/stat_var/dataset_selector.tsx
n-h-diaz: this was mostly to keep the absolute positioning so the autocomplete dropdown would show up above the rest of the search box - I tried getting it to work with the flex instead (which does do the scaling) but wasn't sure how to get it to be visible... however I'm not very familiar with css unfortunately ;; so if you have other suggestions for what I could do, I can definitely try!

n-h-diaz:
@pradh 
thanks for reviewing! to address your questions: 

0. here's a sample video: https://screencast.googleplex.com/cast/NTg1ODEwMzI2NzQyNjMwNHwyNjNiMzAyMi04Mw the dataset dropdown does not show up on screencast for some reason, but it looks like this:
Screen Shot 2022-08-08 at 9 18 54 PM

1. yes it's supposed to!

2. no, the datasets should not be selectable unless a valid source is first entered

static/js/tools/stat_var/dataset_selector.tsx
chejennifer: Do they have to have a source selected first in order to filter by dataset? Because the wording "optionally select a dataset" leads me to believe I can either search for a source or choose a dataset instead .. maybe "select a dataset (optional)"?

static/js/tools/stat_var/dataset_selector.tsx
chejennifer: hm, why does the autocomplete need to show up above the search box? I don't think they would need to overlap right since it goes underneath the search box? or can you take a screenshot/video of the problem you had?

pradh:
> @pradh thanks for reviewing! to address your questions:
> 
> 0. here's a sample video: https://screencast.googleplex.com/cast/NTg1ODEwMzI2NzQyNjMwNHwyNjNiMzAyMi04Mw the dataset dropdown does not show up on screencast for some reason, but it looks like this:
> 
> Screen Shot 2022-08-08 at 9 18 54 PM
> 
> 1. yes it's supposed to!
> 2. no, the datasets should not be selectable unless a valid source is first entered

Thank you!

static/js/tools/stat_var/dataset_selector.tsx
n-h-diaz: oh okay I changed to this instead!

static/js/tools/stat_var/dataset_selector.tsx
n-h-diaz: sorry eventually got it working with the flex!

static/js/tools/stat_var/page.tsx
```diff
@@ -66,21 +93,32 @@ class Page extends Component<unknown, PageStateType> {
...
           openSvHierarchyModalCallback={this.toggleSvHierarchyModal}
           collapsible={false}
           svHierarchyType={StatVarHierarchyType.STAT_VAR}
-          samplePlaces={[]}
+          samplePlaces={this.state.entities}
           deselectSVs={() => this.updateHash("")}
-          selectedSVs={{ [this.state.statVar]: {} }}
+          selectedSVs={svs}
           selectSV={(sv) => this.updateHash(sv)}
+          disableAlert={true}
```
chejennifer, line 113: I'm curious why we want to disable the alert for the stat var page, because wouldn't we want to alert the user when a variable is deselected?

static/js/tools/stat_var/dataset_selector.tsx
```diff
@@ -0,0 +1,218 @@
...
+          },
+        ]);
+      }
+    } else if (e.key === "ArrowUp") {
+      if (activeSuggestion === 0) {
+        return;
+      }
+      setActiveSuggestion(activeSuggestion - 1);
+    } else if (e.key === "ArrowDown") {
+      if (activeSuggestion - 1 === filteredSuggestions.length) {
```
chejennifer, line 139: should it be `activeSuggestion === filteredSuggestions.length - 1` which would also be `activeSuggestion + 1 === filteredSuggestions.length`?

static/js/tools/stat_var/page.tsx
```diff
@@ -58,6 +69,22 @@ class Page extends Component<unknown, PageStateType> {
       this.fetchSummary();
     };
     this.fetchSummary();
+    const sourcePromise = await axios.get(
```
chejennifer, line 72: for both this promise and the sourceNamesPromise, we should add some error handling

chejennifer (commented):
thanks for the updates!

static/js/tools/stat_var/dataset_selector.tsx
n-h-diaz: sorry about that! changed

static/js/tools/stat_var/page.tsx
n-h-diaz: done

static/js/tools/stat_var/page.tsx
n-h-diaz: this use case seemed a little different than the other tools to me, because a user is deliberately filtering the stat vars here (rather than picking a place without knowing whether a stat var exists). so it seemed expected that a stat var the user is currently looking at might not exist for a completely different source (and in fact, given the distribution of stat vars vs sources, this actually happens most of the time). because of this, when I was testing, it seemed a bit inconvenient to have to keep closing the alert, so instead it'll just redirect to the stat var landing page. wdyt?

chejennifer: ok, that makes sense to me!

chejennifer (approved):
Thanks for the changes!

static/js/tools/stat_var/page.tsx
```diff
@@ -162,13 +207,13 @@ class Page extends Component<unknown, PageStateType> {
         : undefined;
     this.setState({
       description:
-        descriptionPromise.data[sv].length > 0
-          ? descriptionPromise.data[sv][0]
+        descriptionPromise?.data[sv].length > 0
```
chejennifer, line 210: sorry, one more thing, we should probably also handle failed promises here

n-h-diaz: done

PR #1710, opened 2022-08-09
npm test speed up

chejennifer:
- update npm cloudbuild to use a machine with 8 CPUs instead of the default 1 CPU. This improved the webpack build  time from over 7mins to about 2mins
- add a plugin that profiles the webpack build time broken down into time spent by each loader and time spent by each plugin:
    - Screen Shot 2022-08-11 at 2 34 52 PM


build/ci/cloudbuild.npm.yaml
```diff
@@ -28,3 +28,6 @@ steps:
 
         # ./run_test.sh -lbc will run lint test, build client packages, and run client tests.
         ./run_test.sh -lbc
+
+options:
```
shifucun, line 32: nice!

shifucun (approved):
Thanks for the fix!

sass loader takes quite some time.. when we have time, should profile and see if we can trim down the packages

chejennifer:
> Thanks for the fix!
> 
> sass loader takes quite some time.. when we have time, should profile and see if we can trim down the packages

Yes, I agree, it takes almost as long as ts-loader but processes much fewer modules in that time

PR #1711, opened 2022-08-09
[devops] Terraform for DC website part2. Add network resources

Fructokinase:
VPC related resources. 

TESTED=terraform apply on project "datacom-alexyfchen-tf-test2"

### DIFF
--- a/deploy/terraform-datacommons-website/modules/vpc/variables.tf
+++ b/deploy/terraform-datacommons-website/modules/vpc/variables.tf
@@ -1,0 +1,73 @@
+/**
+ * Copyright 2022 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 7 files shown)


deploy/terraform-datacommons-website/modules/vpc/variables.tf
```diff
@@ -0,0 +1,73 @@
...
+variable "network_name" {
+  type        = string
+  description = "The name of the network being created, or the name of a existing VPC network."
+  default     = "data-commons-network"
+}
+
+variable "subnet_name" {
+  type        = string
+  description = "The name of the subnet being created to host the cluster in."
+  default     = "data-commons-subnet"
```
shifucun, line 37: Is it able to change the current deployment subnet_name to this? Or would there be discrepancy between the config and existing deployments?

Fructokinase: Terraform doesn't seem to do a good job with changing existing deployment. It will likely try to delete and recreate the VPC, which means deleting and recreating the cluster that depends on it, and there is a lot of risk.

For changing the params of existing infra, I think it might be easier to manually edit it through UI.

shifucun (approved):

PR #1714, opened 2022-08-10
Updates deployment config.

nopper:


server/services/ai.py
```diff
@@ -206,33 +206,11 @@ def _delexicalize_query(query: str,
     return ''.join(output)
 
 
-# This matches the cache builder logic.
-_KEY_NORMALIZATION = {
```
shifucun, line 210: What is the change here? Why are they not needed?

shifucun (commented):
Please revert server/config/topic_page_pb2.py

shifucun (commented):
Can you add some PR descriptions on what has changed in the new model?

shifucun (approved):

PR #1718, opened 2022-08-12
webdriver test fixes

chejennifer:
- fix a test break caused by updated landing page data from mixer
- add more waiting for download tool and browser tests to fix flakiness


shifucun (approved):
thanks for the fix!

PR #1720, opened 2022-08-12
only profile webpack build times when running test

chejennifer:


shifucun (approved):
nice, now we have more advanced webpack too :)

PR #1727, opened 2022-08-22
[Disease Browser] Add disease chart descriptions, common name and integrate interfaces in one file 

suhana13:
In this PR, I've added the description for each chart and table. In addition, I've linked it to the graph browser view and added the disease common name as the first header (instead of the disease ontology ID, with the former being more intuitive and easy to understand). I've also added a types file with all disease browser interfaces added to it, for the sake of better organization of the code. 
Screen Shot 2022-08-22 at 5 05 51 PM

### DIFF
--- a/static/js/biomedical/bio_charts_utils.ts
+++ b/static/js/biomedical/bio_charts_utils.ts
@@ -15,7 +15,7 @@
  */
 import * as d3 from "d3";
 
-import { DiseaseGeneAssociationData } from "./disease/chart";
+import { DiseaseGeneAssociationData } from "./disease/types";
 import { ProteinNumData } from "./protein/chart";
 import { DiseaseAssociationType } from "./protein/page";
 import { InteractionLink, ProteinNode } from "./protein/types";
--- a/static/js/biomedical/disease/data_processing_utils.ts
+++ b/static/js/biomedical/disease/data_processing_utils.ts
@@ -285,3 +285,31 @@ export function getCompoundDiseaseContraindication(
   processedData.sort((a, b) => (a.drugSource > b.drugSource ? 1 : -1));
   return processedData;
 }
+
+/**
+ * Fetches the common name of the disease of interest
+ * @param data
+ * @returns - string with disease common name 
+ */
+ export function getDiseaseCommonName(data: GraphNodes): string {
+  let commonName = null;
+  if (!data) {
+    return;
+  }
+  // check for null values
+  if (_.isEmpty(data.nodes) || _.isEmpty(data.nodes[0].neighbors)) {
+    return;
+  }
+  for (const neighbour of data.nodes[0].neighbors) {
+    if (neighbour.property !== "commonName") {
+      continue;
+    }
+    // check for null or non-existent property values
+    if (_.isEmpty(neighbour.nodes)) {
+      continue;
+    }
+    commonName = neighbour.nodes[0].value;
+    return commonName;
+  }
+}
+
--- a/static/js/biomedical/disease/data_processing_utils.ts
+++ b/static/js/biomedical/disease/data_processing_utils.ts
@@ -22,7 +22,7 @@ import {
   CompoundDiseaseTreatmentData,
   DiseaseGeneAssociationData,
   DiseaseSymptomAssociationData,
-} from "./chart";
+} from "./types";
 /**
  * Fetches the disease-gene association data
  * @param data - the data pertaining to the disease of interest
@@ -89,6 +89,7 @@ export function getDiseaseGeneAssociation(
   }
   return rawData;
 }
+
 /**
  * Fetches the disease-symptom association data
  * @param data
--- a/static/js/biomedical/disease/data_processing_utils.ts
+++ b/static/js/biomedical/disease/data_processing_utils.ts
@@ -289,9 +289,9 @@ export function getCompoundDiseaseContraindication(
 /**
  * Fetches the common name of the disease of interest
  * @param data
- * @returns - string with disease common name 
+ * @returns - string with disease common name
  */
- export function getDiseaseCommonName(data: GraphNodes): string {
+export function getDiseaseCommonName(data: GraphNodes): string {
   let commonName = null;
   if (!data) {
     return;
@@ -309,7 +309,10 @@ export function getCompoundDiseaseContraindication(
       continue;
     }
     commonName = neighbour.nodes[0].value;
-    return commonName;
+    // capitalize the first letter of the disease name
+    const formattedDiseaseName =
+      commonName[1].toUpperCase() + commonName.slice(2);
+    // return formatted disease name with double quotes removed
+    return formattedDiseaseName.replaceAll('"', "");
   }
 }
-
--- a/static/js/biomedical/disease/disease_chart.test.tsx
+++ b/static/js/biomedical/disease/disease_chart.test.tsx
@@ -4,12 +4,6 @@ import Adapter from "enzyme-adapter-react-16";
 Enzyme.configure({ adapter: new Adapter() });
 
 import { GraphNodes } from "../../shared/types";
-import {
-  CompoundDiseaseContraindicationData,
-  CompoundDiseaseTreatmentData,
-  DiseaseGeneAssociationData,
-  DiseaseSymptomAssociationData,
-} from "./chart";
 import dataDOID2403 from "./data_DOID_2403.json";
 import {
   getCompoundDiseaseContraindication,
@@ -17,6 +11,12 @@ import {
   getDiseaseGeneAssociation,
   getDiseaseSymptomAssociation,
 } from "./data_processing_utils";
+import {
+  CompoundDiseaseContraindicationData,
+  CompoundDiseaseTreatmentData,
+  DiseaseGeneAssociationData,
+  DiseaseSymptomAssociationData,
+} from "./types";
 
 test("getDiseaseGeneAssociation", () => {
   const cases: {
--- a/static/js/biomedical/disease/page.tsx
+++ b/static/js/biomedical/disease/page.tsx
@@ -35,6 +35,7 @@ import {
   getDiseaseSymptomAssociation,
 } from "./data_processing_utils";
 import { DrugTreatmentTable } from "./drug_table";
+
 export interface PagePropType {
   dcid: string;
   nodeName: string;
--- a/static/js/biomedical/disease/page.tsx
+++ b/static/js/biomedical/disease/page.tsx
@@ -127,6 +127,10 @@ export class Page extends React.Component<PagePropType, PageStateType> {
         </div>
         <br></br>
         <h5>Chemical Compound Disease Contraindication</h5>
+        <p>
+          The chemical compounds contraindicated with {diseaseName} as reported
+          by the ChEMBL database.
+        </p>
         <br></br>
         <div>
           <div id="table"></div>
--- a/static/js/biomedical/disease/page.tsx
+++ b/static/js/biomedical/disease/page.tsx
@@ -22,6 +22,7 @@ import axios from "axios";
 import React from "react";
 
 import { GraphNodes } from "../../shared/types";
+import { GRAPH_BROWSER_REDIRECT } from "../bio_charts_utils";
 import {
   drawDiseaseGeneAssocChart,
   drawDiseaseSymptomAssociationChart,
@@ -29,6 +30,7 @@ import {
 import {
   getCompoundDiseaseContraindication,
   getCompoundDiseaseTreatment,
+  getDiseaseCommonName,
   getDiseaseGeneAssociation,
   getDiseaseSymptomAssociation,
 } from "./data_processing_utils";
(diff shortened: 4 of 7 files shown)


static/js/biomedical/disease/data_processing_utils.ts
```diff
@@ -22,7 +22,7 @@ import {
   CompoundDiseaseTreatmentData,
   DiseaseGeneAssociationData,
   DiseaseSymptomAssociationData,
-} from "./chart";
+} from "./types";
```
beets, line 25: nit: please add an empty line after imports

static/js/biomedical/disease/data_processing_utils.ts
```diff
@@ -285,3 +286,34 @@ export function getCompoundDiseaseContraindication(
...
+    // check for null or non-existent property values
+    if (_.isEmpty(neighbour.nodes)) {
+      continue;
+    }
+    commonName = neighbour.nodes[0].value;
+    // capitalize the first letter of the disease name
+    const formattedDiseaseName =
+      commonName[1].toUpperCase() + commonName.slice(2);
+    // return formatted disease name with double quotes removed
+    return formattedDiseaseName.replaceAll('"', "");
```
beets, line 317: I might use `trim` here instead, so only completely quoted names are cleared. Could there be instances where we'd want to keep the double quote e.g. `foo "bar"`?

static/js/biomedical/disease/data_processing_utils.ts
```diff
@@ -285,3 +286,34 @@ export function getCompoundDiseaseContraindication(
...
+    if (neighbour.property !== "commonName") {
+      continue;
+    }
+    // check for null or non-existent property values
+    if (_.isEmpty(neighbour.nodes)) {
+      continue;
+    }
+    commonName = neighbour.nodes[0].value;
+    // capitalize the first letter of the disease name
+    const formattedDiseaseName =
```
beets, line 314: Before performing these string operations, first check that there's a valid string and has length > 2.

It's safer to use `charAt` then strict array dereference. And it should start with `.charAt(0)` and `.slice(1)`

static/js/biomedical/disease/data_processing_utils.ts
```diff
@@ -285,3 +286,34 @@ export function getCompoundDiseaseContraindication(
   processedData.sort((a, b) => (a.drugSource > b.drugSource ? 1 : -1));
   return processedData;
 }
+
+/**
+ * Fetches the common name of the disease of interest
+ * @param data
+ * @returns - string with disease common name
+ */
+export function getDiseaseCommonName(data: GraphNodes): string {
```
beets, line 295: Please add a unit test for this exported function

static/js/biomedical/disease/page.tsx
```diff
@@ -65,6 +68,8 @@ export class Page extends React.Component<PagePropType, PageStateType> {
     );
   }
   render(): JSX.Element {
+    const diseaseName = getDiseaseCommonName(this.state.data);
```
beets, line 71: Consider a fallback to the DCID if there is no name available (or there was an error parsing / fixing the name)

static/js/biomedical/disease/page.tsx
```diff
@@ -65,6 +68,8 @@ export class Page extends React.Component<PagePropType, PageStateType> {
...
+    const diseaseLink = `${GRAPH_BROWSER_REDIRECT}${this.props.dcid}`;
```
beets, line 72: Would be less error-prone if instead of a const, graph_browser_redirect was a function that returns a link given a DCID.

static/js/biomedical/disease/page.tsx
```diff
@@ -85,13 +90,33 @@ export class Page extends React.Component<PagePropType, PageStateType> {
...
       <>
-        <h2>Disease Browser</h2>
+        <h2>{diseaseName}</h2>
+        <h6>
+          <a href={diseaseLink}>Graph Browser View</a>
+        </h6>
         <h5>Disease-Gene Association</h5>
+        <p>
+          The association score of {diseaseName} with genes as reported by
+          DISEASES by Jensen Lab. Associations were determined by text mining of
```
beets, line 100: should we add a link for DISEASES, SPOKE and ChEMBL?

beets (commented):
looks nice overall!

static/js/biomedical/disease/data_processing_utils.ts
suhana13: Done!

static/js/biomedical/disease/data_processing_utils.ts
suhana13: That's a great idea! `.trim()` would remove all double quotes from the string and we don't want to keep any double quotes, so this works perfectly well!

static/js/biomedical/disease/data_processing_utils.ts
suhana13: I believe the validity of the string is checked in the `_.isEmpty()` function. I added another check for string length and if it is less than 2, then we use the DCID instead!

static/js/biomedical/disease/data_processing_utils.ts
suhana13: Done!

static/js/biomedical/disease/page.tsx
suhana13: That's a great idea! Added a fallback to the DCID!

static/js/biomedical/disease/page.tsx
suhana13: Great suggestion! Added a function `getEntityLink()` in `bio_charts_utils` as it will be a shared function across diseases and proteins

static/js/biomedical/disease/page.tsx
suhana13: That's a good idea! What do you think @spiekos ?

suhana13: After discussing it with Sam, we are thinking of adding the name and source of the provenance for each of the graphs. But, instead of hardcoding it, we're planning to getting it from the config.

static/js/biomedical/disease/data_processing_utils.ts
```diff
@@ -285,3 +289,43 @@ export function getCompoundDiseaseContraindication(
...
+  let commonName = null;
+  let diseaseDCID = null;
```
chejennifer, line 300: also here, don't need to define diseaseDCID as null, can just define the variable in line 308

static/js/biomedical/disease/disease_chart.test.tsx
```diff
@@ -175,3 +176,26 @@ test("getCompoundDiseaseContraindication", () => {
     }
   }
 });
+
+test("getDiseaseCommonName", () => {
```
chejennifer, line 180: it looks like this file is testing the functions in the file data_processing_utils.ts, so this test file should be named data_processing_utils.test.ts

For a file named abc.ts, the corresponding test file should be named abc.test.ts

static/js/biomedical/disease/disease_chart.test.tsx
```diff
@@ -175,3 +176,26 @@ test("getCompoundDiseaseContraindication", () => {
...
+  const cases: {
+    data: GraphNodes;
+    wantArray: string;
```
chejennifer, line 183: nit: `wantName` instead of `wantArray` since it's not expecting an array?

static/js/biomedical/disease/disease_chart.test.tsx
```diff
@@ -175,3 +176,26 @@ test("getCompoundDiseaseContraindication", () => {
...
+  }[] = [
+    {
```
chejennifer, line 185: we should include the following cases:
- happy case where common name is found
- there is a common name property but empty value so fallback to dcid
- there is no common name property so fallback to dcid

static/js/biomedical/disease/types.ts
```diff
@@ -0,0 +1,87 @@
...
+}
+
+export interface DrugTreatmentTableColumn {
+  // the column id
+  id: string;
+  // the column display name or header
+  name: string;
+}
+
+export interface DrugTreatmentTableProps {
```
chejennifer, line 81: please keep `Props` types in the file that defines the React component using the `Props` type. This is because `Props` are associated with a react component.

chejennifer (commented):
Thanks for the refactoring, a few small things

static/js/biomedical/bio_charts_utils.ts
```diff
@@ -148,3 +148,13 @@ export function addYLabel(
     .attr("dy", "1em")
     .text(labelText);
 }
+
+/**
+ * Creates the page redirection link of the input entity
+ * @param dcid - entity dcid
+ * @returns - redirection link
```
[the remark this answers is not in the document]
suhana13: Thanks for pointing it out! Removed the `-` from all `@returns` statements in the protein and disease directories.

static/js/biomedical/disease/data_processing_utils.ts
```diff
@@ -97,6 +99,7 @@ export function getDiseaseGeneAssociation(
 export function getDiseaseSymptomAssociation(
   data: GraphNodes
 ): DiseaseSymptomAssociationData[] {
+  let diseaseDCID = null;
```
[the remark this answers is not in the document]
suhana13: Yeah, I had initially defined to diseaseDCID to check something but I'm not using it in this function, so removed it:)

static/js/biomedical/disease/data_processing_utils.ts
suhana13: Correct! Fixed that!

static/js/biomedical/disease/disease_chart.test.tsx
suhana13: Thank you for pointing out the file naming rules. I changed the name to `data_processing_utils.test.ts` for both the protein and disease directories.

static/js/biomedical/disease/disease_chart.test.tsx
suhana13: RIght! Changed to `wantName`

static/js/biomedical/disease/types.ts
suhana13: Fixed the location of the Props!

static/js/biomedical/disease/disease_chart.test.tsx
suhana13: Currently in the graph, we just have happy cases i.e., all diseases have a common name (which is due to the fact that they originate from the same data import which has commonName as a mandatory property). However, there are a few diseases where we don't have commonName but the Name instead. We're on the verge of updating the disease data to have name instead of commonName which would just change the property name and we would consistently have the name for all diseases on the graph. The fallback case is just if in the future, we end up adding diseases from other databases which don't have a name, but that seems very unlikely

static/js/biomedical/protein/chart.ts
```diff
@@ -341,7 +341,7 @@ function dragNode(
  * Draws the legend for the tissue score chart if the data exists.
  * @param id - div id for the tissue legend
  * @param data - json formatted protein data
- * @returns - empty for null data
+ * @returns empty for null data
```
chejennifer, line 344: looking at the function definition, this function shouldn't return anything? since the return type is `void`? If nothing is being `return`ed, don't need this line. Here and the rest of the functions in this file

chejennifer (commented):
Thanks for the changes! Sorry, found a few more nits

static/js/biomedical/protein/chart.ts
suhana13: Done!

static/js/biomedical/disease/data_processing_utils.ts
```diff
@@ -285,3 +287,42 @@ export function getCompoundDiseaseContraindication(
...
+
+/**
+ * Fetches the common name of the disease of interest
+ * @param data
+ * @returns string with disease common name
+ */
+export function getDiseaseCommonName(data: GraphNodes): string {
+  let commonName = null;
+  if (!data) {
+    return;
```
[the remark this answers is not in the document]
suhana13: Done!

static/js/biomedical/disease/data_processing_utils.ts
```diff
@@ -13,7 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-import _ from "lodash";
+import _, { trim } from "lodash";
```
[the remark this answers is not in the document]

static/js/biomedical/bio_charts_utils.ts
```diff
@@ -148,3 +148,13 @@ export function addYLabel(
     .attr("dy", "1em")
     .text(labelText);
 }
+
+/**
+ * Creates the page redirection link of the input entity
+ * @param dcid - entity dcid
```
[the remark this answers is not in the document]
suhana13: Done!

static/js/biomedical/disease/data_processing_utils.ts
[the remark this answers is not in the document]
chejennifer: this return isn't updated?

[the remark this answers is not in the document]
chejennifer: this return isn't updated?

chejennifer (approved):
thanks for the changes! one small change missing, but looks good otherwise

PR #1728, opened 2022-08-23
[import wizard] Set up the skeleton for V2

chejennifer:
- added the changes to have V2 appear at /import/new
- types.ts is copied from V1 - static/js/import2 will have all the files needed for V2 (even if they're duplicate files from V1) so that when we do the flip, we can just rename the folder and delete the V1 folder

### DIFF
--- a/server/templates/import_wizard2.html
+++ b/server/templates/import_wizard2.html
@@ -1,0 +1,39 @@
+{#
+  Copyright 2022 Google LLC
+
+  Licensed under the Apache License, Version 2.0 (the "License");
+  you may not use this file except in compliance with the License.
+  You may obtain a copy of the License at
+
+        http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+#}
+
+{%- extends BASE_HTML -%}
+
+{% set subpage_title = "Import Wizard" %}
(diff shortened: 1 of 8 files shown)


server/templates/import_wizard2.html
```diff
@@ -0,0 +1,39 @@
...
+{% set subpage_title = "Import Wizard" %}
+{% set subpage_url = url_for('import_wizard.main_new') %}
+{% set title = "Import Wizard" %}
+{% set main_id = 'main-import-wizard' %}
+{% set page_id = 'page-import-wizard' %}
+
+{% block head %}
+  <link rel="stylesheet" href={{url_for('static', filename='css/import_wizard2.min.css')}} >
+  <link rel="stylesheet" href="https://fonts.googleapis.com/icon?family=Material+Icons">
+  <link rel="stylesheet" href="https://fonts.googleapis.com/icon?family=Material+Icons+Outlined">
```
beets, line 28: nit: pull these in only if required.

static/js/import_wizard2/import_wizard.ts
```diff
@@ -0,0 +1,24 @@
...
+ */
+import React from "react";
+import ReactDOM from "react-dom";
+
+import { Page } from "./components/page";
+
+ReactDOM.render(
+  React.createElement(Page),
+  document.getElementById("main-pane")
+);
```
beets, line 24: this should be set as a callback to `window.onload`

beets (approved):
only small nits!

server/templates/import_wizard2.html
chejennifer: done

static/js/import_wizard2/import_wizard.ts
chejennifer: done

PR #1735, opened 2022-08-26
[map] update legend domain for dc aggregate temp diffs

chejennifer:
- Update legend domain for dc aggregate temp diffs to range from -30 to 30
Screen Shot 2022-08-26 at 3 01 31 PM


n-h-diaz (approved):
thanks!

PR #1736, opened 2022-08-28
[User Portal] Signed in user management database and function to upload csv/tmcf file to GCS

shifucun:
This is with minimal function, workflow check and css change, to get end-to-end working as the first goal.

image

image

image


server/routes/user.py
```diff
@@ -74,7 +78,37 @@ def index():
...
 
 
+@bp.route('/upload', methods=['POST'])
+@login_is_required
+def upload():
+    # TODO: changeg SECRETE_PROJECT to APP_PROJECT
+    # Upload files to GCS
+    user_id = session["google_id"]
+    project = current_app.config['SECRET_PROJECT']
+    bucket_name = project + '-resources'
```
Fructokinase, line 88: Is this a general resource bucket? Maybe a separate function in lib/gcs.py or lib/config.py instead of inline?

server/routes/user.py
```diff
@@ -74,7 +78,37 @@ def index():
     return "Data Commons </br> <a href='/user/auth/login'><button>Login</button></a>"
 
 
+@bp.route('/upload', methods=['POST'])
+@login_is_required
+def upload():
```
Fructokinase, line 83: nit: upload is a bit vague, maybe upload_import?

server/routes/user.py
```diff
@@ -74,7 +78,37 @@ def index():
...
+    # TODO: changeg SECRETE_PROJECT to APP_PROJECT
+    # Upload files to GCS
+    user_id = session["google_id"]
+    project = current_app.config['SECRET_PROJECT']
+    bucket_name = project + '-resources'
+    gcs_client = storage.Client(project=project)
+    bucket = gcs_client.get_bucket(bucket_name)
+    import_name = request.form.get('importName')
+    for f in request.files.getlist('files'):
+        blob_name = f'{user_id}/{import_name}/{f.filename}'
```
Fructokinase, line 93: Does this overwrite existing files? If a file of the same name already exists, should it be a warning? 

We could have a force flag in the request. If force flag is turned on, overwrite it. Otherwise uploading file of the same name returns 400.

server/routes/api/user.py
```diff
@@ -0,0 +1,48 @@
...
+
+import flask
+import json
+
+from flask import current_app
+
+bp = flask.Blueprint('api.user', __name__, url_prefix='/api/user')
+
+
+def get_user_info(google_id):
```
beets, line 24: should we rename google_id to user_id or oauth_user_id?

static/js/user/page.tsx
```diff
@@ -0,0 +1,122 @@
...
+      {props.newUser && (
+        <div>
+          Welcome to Data Commons Import Portal. We have created a profile for
+          you. Please add your imports.
+        </div>
+      )}
+      <div id="imports">
+        <div>All Imports</div>
+        <Button
+          id="add-import-open-modal-button"
```
beets, line 74: to keep this separate from the wizard, should we rename this to "Upload import data" for now?

beets (approved):
this is very cool!

beets:
we should also allow mcf uploads (for additional schema when required). we will also need to support multiple file uploads for a single import group (e.g. sharded csv's), but that can be in a follow up.

server/routes/api/user.py
```diff
@@ -0,0 +1,48 @@
...
+    for doc in docs:
+        if doc.id == google_id:
+            return json.dumps(doc.to_dict())
+    return {}
+
+
+def create_user(google_id, data):
+    """Create a new user."""
+    db = current_app.config['USER_DB']
+    doc_ref = db.collection(u'users').document(google_id)
```
pradh, line 38: do we have to store user email, or can we just store a hash(email) to identify the user?

server/routes/api/user.py
```diff
@@ -0,0 +1,48 @@
...
+    doc_ref.set(data)
+
+
+def add_import(google_id, import_name):
+    """Add a new import for a user"""
+    db = current_app.config['USER_DB']
+    doc_ref = db.collection(u'users').document(google_id)
+    importUpdate = {}
+    importUpdate[f'import.{import_name}.status'] = "imported"
```
pradh, line 47: for my understanding, where is the schema defined?

pradh, line 47: should this be an ENUM rather than string?

server/routes/api/user.py
```diff
@@ -0,0 +1,48 @@
...
+
+from flask import current_app
+
+bp = flask.Blueprint('api.user', __name__, url_prefix='/api/user')
+
+
+def get_user_info(google_id):
+    """Returns data given a user id."""
+    db = current_app.config['USER_DB']
+    users_ref = db.collection(u'users')
```
pradh, line 27: Could we index "users" by hash(google_id) to not loop?

server/routes/api/user.py
```diff
@@ -0,0 +1,48 @@
...
+    docs = users_ref.stream()
+    for doc in docs:
+        if doc.id == google_id:
+            return json.dumps(doc.to_dict())
+    return {}
+
+
+def create_user(google_id, data):
```
pradh, line 35: Describe the parameters?  What is "data"?

server/routes/user.py
```diff
@@ -74,7 +78,37 @@ def index():
...
+@bp.route('/upload', methods=['POST'])
+@login_is_required
+def upload():
+    # TODO: changeg SECRETE_PROJECT to APP_PROJECT
+    # Upload files to GCS
+    user_id = session["google_id"]
+    project = current_app.config['SECRET_PROJECT']
+    bucket_name = project + '-resources'
+    gcs_client = storage.Client(project=project)
+    bucket = gcs_client.get_bucket(bucket_name)
```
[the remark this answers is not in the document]
pradh: +1.  validate inputs too, like importName being empty?

pradh (commented):
Exciting!

server/routes/api/user.py
```diff
@@ -0,0 +1,48 @@
...
+import flask
+import json
+
+from flask import current_app
+
+bp = flask.Blueprint('api.user', __name__, url_prefix='/api/user')
+
+
+def get_user_info(google_id):
+    """Returns data given a user id."""
```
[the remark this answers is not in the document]
shifucun: Added a blob of the database structure on the top and some details of the "user document" here.

server/routes/api/user.py
```diff
@@ -0,0 +1,48 @@
...
+    db = current_app.config['USER_DB']
```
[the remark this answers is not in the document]
shifucun: Using a helper function now.

server/routes/api/user.py
shifucun: Done

server/routes/api/user.py
shifucun: Not saving name in user anymore, so not taking data field and set it as empty {} object.

server/routes/api/user.py
shifucun: Not saving user email here, just use the hash(auth_id) as primary key

server/routes/api/user.py
shifucun: cloud firestore is a NOSQL database and does not have strict schema. I added a blob as document comments about the layout, basically with "user document" and "import subcollection"

shifucun: Done

static/js/user/page.tsx
shifucun: Done

server/routes/user.py
shifucun: Done

server/routes/user.py
shifucun: We would eventually have a user flow design and decide if user should overwrite an import or we save versioned import files. I like the idea of "forced" flag and also show it on UI to callout "overwrite". Will make this simple for now and enrich it in the next round

shifucun:
> we should also allow mcf uploads (for additional schema when required). we will also need to support multiple file uploads for a single import group (e.g. sharded csv's), but that can be in a follow up.

Sounds good, will save these for next few rounds :)

server/routes/user.py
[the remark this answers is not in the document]
shifucun: Added check and early return here

shifucun (commented):
Thank you all for the useful inputs.

Fructokinase (approved):

server/routes/api/user.py
```diff
@@ -16,33 +16,50 @@
...
     """Add a new import for a user"""
-    db = current_app.config['USER_DB']
-    doc_ref = db.collection(u'users').document(google_id)
-    importUpdate = {}
-    importUpdate[f'import.{import_name}.status'] = "imported"
-    doc_ref.update(importUpdate)
+    db = get_user_db()
+    user_ref = db.collection(USER_COLLECTION).document(user_id)
+    import_ref = user_ref.collection(IMPORT_COLLECTION).document(import_name)
+    import_ref.set({'status': ImportStatus.IMPORTED.value})
```
pradh, line 65: `status` is a field within the import collection?  Use ENUM for it?

Maybe not in this PR, but I wonder if its worth strongly typing the in-memory data structures with pytype?

server/routes/user.py
```diff
@@ -87,12 +89,19 @@ def index():
 def upload_import():
     # TODO: changeg SECRETE_PROJECT to APP_PROJECT
```
pradh, line 90: nit: change SECRET_PROJECT ...

server/routes/user.py
```diff
@@ -87,12 +89,19 @@ def index():
...
     bucket_name = get_gcs_bucket(project)
     gcs_client = storage.Client(project=project)
-    bucket = gcs_client.get_bucket(bucket_name)
+    try:
+        bucket = gcs_client.get_bucket(bucket_name)
+    except exceptions.NotFound:
+        return f'Bucket {bucket_name} is not found', 500
+
     import_name = request.form.get('importName')
+    if import_name == '':
```
pradh, line 103: `if not import_name` ?  (resilient to it being `None` too)

static/js/user/page.tsx
```diff
@@ -76,7 +76,7 @@ export function Page(props: PagePropType): JSX.Element {
           color="light"
           onClick={() => setModalOpen(true)}
         >
-          Add Import
+          Upload import data
```
pradh, line 79: nit:  "Upload data for import" or "Upload files for import"

pradh (approved):
Thanks for the changes!

server/routes/api/user.py
pradh: Jehangir's wizard code is a pretty good example of strong typing: https://github.com/datacommonsorg/website/blob/master/server/routes/api/import_detection/detection_types.py

shifucun: Added typing and made Import and User as class

server/routes/user.py
shifucun: Done

server/routes/user.py
shifucun: Done

static/js/user/page.tsx
shifucun: Done

server/routes/api/user.py
```diff
@@ -35,31 +37,45 @@ class ImportStatus(Enum):
     IMPORTED = 1
 
 
-def get_user_db():
+@dataclass
+class Import:
+    """Data type for import entry"""
+    status: int
```
pradh, line 43: nit:  comment this uses ImportStatus enum

pradh (approved):
Thanks!  Looks a lot more readable with typing!

Maybe as this grows we can refactor the types to storage_schema.py or something.

server/routes/api/user.py
shifucun: Thanks! I will make this change in follow up PR

PR #1737, opened 2022-08-30
Use getters for DC API key

beets:
Use getters for accessing DC_API_KEY which is only required in dev instances. At the moment only affects data download tool


shifucun (approved):
thanks for the fix!

PR #1740, opened 2022-09-02
Display import list in user dashboard

shifucun:
image

### DIFF
--- a/server/routes/api/user.py
+++ b/server/routes/api/user.py
@@ -20,7 +20,7 @@ from google.cloud import firestore
 from dataclasses import asdict, dataclass
 from enum import Enum
 from flask import current_app
-from typing import List
+from typing import Dict, List
 
 bp = flask.Blueprint('api.user', __name__, url_prefix='/api/user')
 
@@ -39,10 +39,17 @@ class ImportStatus(Enum):
 
 
 @dataclass
+class GcsFile:
+    """Data type for a file stored in Google Cloud Storage"""
+    bucket: str
+    object: str
+
+
+@dataclass
 class Import:
     """Data type for import entry"""
     status: int  # This is from ImportStatus Enum
-    files: List[str]
+    gcs_files: List[GcsFile]  # The full GCS path of the import files.
 
 
 @dataclass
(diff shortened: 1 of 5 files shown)


static/js/user/user.ts
```diff
@@ -19,14 +19,16 @@ import ReactDOM from "react-dom";
 
 import { Page } from "./page";
 
+const dom = "main-section";
```
beets, line 22: nit: domId?

static/js/user/user.ts
```diff
@@ -19,14 +19,16 @@ import ReactDOM from "react-dom";
...
+
 window.onload = () => {
-  const info = JSON.parse(document.getElementById("user").dataset.info);
-  const newUser = "True" === document.getElementById("user").dataset.new_user;
+  const user = JSON.parse(document.getElementById(dom).dataset.user);
```
beets, line 25: document.getElementById is a somewhat expensive operation which we should avoid calling repeatedly if possible. Could we instead hold a reference to the dataset object?

server/routes/user.py
```diff
@@ -85,7 +86,18 @@ def index():
...
+def get_imports() -> Dict[str, Dict]:
+    user_id = libutil.hash_id(session['oauth_user_id'])
+    user = user_api.get_user(user_id)
+    imports: Dict[str, Dict] = {}
+    for im in user.reference.collection(user_api.IMPORT_COLLECTION).stream():
+        imports[im.id] = im.to_dict()
+    return imports
+
+
+@bp.route('/import/upload', methods=['POST'])
```
beets, line 100: Should we start an API file for this route?

server/routes/api/user.py
```diff
@@ -34,13 +35,14 @@
...
-    IMPORTED = 1
+    UPLOADED = 1
 
 
 @dataclass
 class Import:
     """Data type for import entry"""
-    status: int
+    status: int  # This is from ImportStatus Enum
+    files: List[str]
```
beets, line 45: nit: maybe `file_names` since files could be file handles. 

also would be great to document assumptions for these. looks like these are only file_names from an assumed fixed folder name. it might be worth storing the folder path too as part of the import so we can still recreate a full path to the files even if path assumptions change

server/routes/user.py
```diff
@@ -85,7 +86,18 @@ def index():
...
 
-@bp.route('/upload/import', methods=['POST'])
+@bp.route('/imports')
+@login_is_required
+def get_imports() -> Dict[str, Dict]:
+    user_id = libutil.hash_id(session['oauth_user_id'])
+    user = user_api.get_user(user_id)
+    imports: Dict[str, Dict] = {}
+    for im in user.reference.collection(user_api.IMPORT_COLLECTION).stream():
+        imports[im.id] = im.to_dict()
```
beets, line 96: Should this be in a helper in user.py (corollary to add_import, so all interaction with the db happens in that module).

beets (approved):
nice!

server/routes/api/user.py
shifucun: Change file name to an object with GCS bucket and object name now.

server/routes/user.py
shifucun: This is a user action and now with full URL /user/import/upload. Maybe can leave it here for now and as more endpoints added, we can decide if have a new file for user import actions?

shifucun (commented):
thanks for the review

server/routes/user.py
beets: sounds good, but please add a TODO so we add one when the next endpoint is added.

shifucun: Done

PR #1743, opened 2022-09-13
[wizard2] constant var template updates

beets:
- added more mappings
- hooked up user mapping updates from template

### DIFF
--- a/static/js/import_wizard2/components/mapping_sections/constant_var.tsx
+++ b/static/js/import_wizard2/components/mapping_sections/constant_var.tsx
@@ -17,7 +17,60 @@
 import React from "react";
 
 import { MappingSectionProps } from "../../templates";
+import { MappedThing, MappingVal } from "../../types";
 
 export function ConstantVar(props: MappingSectionProps): JSX.Element {
-  return <></>;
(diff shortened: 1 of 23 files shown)


static/js/import_wizard2/components/mapping_sections/constant_var.tsx
```diff
@@ -14,61 +14,109 @@
...
 
 export function ConstantVar(props: MappingSectionProps): JSX.Element {
-  let placeMapping, dateMapping, unitMapping: MappingVal;
+  let [
+    placeMapping,
+    dateMapping,
+    unitMapping,
+    statVarMapping,
+    valueMapping,
+  ]: MappingVal[] = [];
```
chejennifer, line 31: curious why using this method instead of the original method to define these variables as undefined

chejennifer (approved):
looks good!

static/js/import_wizard2/components/mapping_sections/constant_var.tsx
beets: This ensures all the vars have the type `MappingVal`. In the previous method, the type was only assigned to the last variable; the others all had type `any`.

beets:
thanks for the review!

PR #1745, opened 2022-09-14
Remove unused mixer API wrapper and update V1 API wrapper

shifucun:
- Removed are old API wrappers that are not used anymore
- The /v1/bulk/property/value has a changed signature from "entity" to "node"


beets (approved):
thanks for the cleanup!

PR #1750, opened 2022-09-17
[Import Wizard 2] Add server-side column detection to the mapping page

juliawu:
This PR:
1) Adds column detection to the mapping page via the server-side API
2) Moves client-side detection files from `import_wizard/` to `import_wizard2/`

This does NOT add template as an extra parameter to the API, slating that for a separate commit.

This is my first time writing TypeScript, so extra scrutiny on the implementation would be appreciated!


static/js/import_wizard2/components/mapping_page.tsx
```diff
@@ -55,9 +56,32 @@ export function MappingPage(props: MappingPageProps): JSX.Element {
   }
 
   useEffect(() => {
-    // TODO(beets): Use server-side detection API.
-    const predictedMapping = getPredictions(props.csvData, placeDetector);
-    setPredictedMapping(predictedMapping);
+    /* Build JSON of column data for detection API Request */
+    const column_ids = new Map(
```
chejennifer, line 60: we want to use camelcase for variables in typescript.

static/js/import_wizard2/components/mapping_page.tsx
```diff
@@ -55,9 +56,32 @@ export function MappingPage(props: MappingPageProps): JSX.Element {
...
+      })
+    );
+    const columnParameters = new Map<string, unknown>([
+      ["column_ids", column_ids],
+      ["column_headers", column_headers],
+      ["column_values", props.csvData.columnValuesSampled],
+    ]);
+    /* Use server-side API for column detection */
+    axios
+      .post(`/api/detection/detect_columns`, JSON.stringify(columnParameters))
```
chejennifer, line 77: have you checked that JSON.stringify(columnParameters) will give you the correct string? I remember running into problems with JSON.stringify on the Map type in the past. If Map type doesn't work, I would suggest just using an object for everything

```
const columnIds = {};
props.csvData.orderedColumns.forEach((column) => {
columnIds[column.columnIdx] = column.id;
});
...
const columnParameters = {
column_ids: columnIds,
...
}
```

static/js/import_wizard2/components/mapping_page.tsx
```diff
@@ -55,9 +56,32 @@ export function MappingPage(props: MappingPageProps): JSX.Element {
...
+      props.csvData.orderedColumns.map((column) => {
+        return [column.columnIdx, column.id];
+      })
+    );
+    const column_headers = new Map(
+      props.csvData.orderedColumns.map((column) => {
+        return [column.columnIdx, column.header];
+      })
+    );
+    const columnParameters = new Map<string, unknown>([
```
chejennifer, line 70: usually would prefer not to use the "unknown" type

static/js/import_wizard2/components/mapping_page.tsx
```diff
@@ -55,9 +56,32 @@ export function MappingPage(props: MappingPageProps): JSX.Element {
...
+    ]);
+    /* Use server-side API for column detection */
+    axios
+      .post(`/api/detection/detect_columns`, JSON.stringify(columnParameters))
+      .then((resp) => {
+        const parsedResponse = parseDetectionApiResponse(resp.data);
+        setPredictedMapping(parsedResponse);
+      })
+      .catch((error) => {
+        console.log(error);
```
chejennifer, line 83: maybe set predictedMapping to an empty mapping (i.e., `new Map()`)

chejennifer (commented):
nice job, this looks good!

static/js/import_wizard2/components/mapping_page.tsx
beets: let's also pull this out into a helper function so that we can easily add tests

juliawu: Moved to a helper function!

static/js/import_wizard2/components/mapping_page.tsx
juliawu: Replaced Map with Object

static/js/import_wizard2/components/mapping_page.tsx
juliawu: Switching to object fixed this!

static/js/import_wizard2/components/mapping_page.tsx
juliawu: added!

chejennifer (commented):
thanks for the changes! few small nits

PR #1754, opened 2022-09-20
Use new categorical landing page cache to in the place explorer and compare performance against old cache

pulkit-s:
b/206032102


server/routes/api/landing_page.py
```diff
@@ -41,20 +41,29 @@
...
 
 
 @cache.memoize(timeout=3600 * 24)  # Cache for one day.
-def get_landing_page_data_helper(dcid, stat_vars_string):
+def get_landing_page_data_helper(dcid, stat_vars_string, category,
+                                 new_landing_page_cache_flag):
     data = {'place': dcid}
+    req_url = '/v1/internal/page/place/' + dcid
+    req_url += '?category={}&new_stat_vars={}'.format(
+        category, stat_vars_string.split(','))
```
shifucun, line 56: should be `split('^')`

shifucun, line 56: Can you move line 53 to 58 to the corresponding if, else block

server/routes/api/landing_page.py
```diff
@@ -41,20 +41,29 @@
...
     if stat_vars_string:
         data['newStatVars'] = stat_vars_string.split('^')
-    response = dc_service.fetch_data('/landing-page',
-                                     data,
-                                     compress=False,
-                                     post=True,
-                                     has_payload=False)
+    if new_landing_page_cache_flag == "false":
```
shifucun, line 59: switch the if else order:

`if new_landing_page_cache_flag:` ...

shifucun, line 59: I think we don't need this flag for this PR and can just change all the logic to use the new V1 API here. Having the flag won't help much, since if we decide to rollback, can just use and old version. It also requires removing the flag and clean the code later on.

Without that, you can also remove the cache_prefix, which makes the code simpler.

server/routes/api/landing_page.py
```diff
@@ -402,18 +411,30 @@ def has_data(data):
...
 
 
+def cache_prefix(*args, **kwargs):
+    return request.args.get("category") + "_" + request.args.get(
+        "cache_flag") + "_"
+
+
 @bp.route('/data/<path:dcid>')
-@cache.cached(timeout=3600 * 24, query_string=True)  # Cache for one day.
+@cache.cached(timeout=3600 * 24, query_string=True,
```
shifucun, line 420: With query_string=True, seems don't need cache_prefix? https://github.com/pallets-eco/flask-caching/blob/f11403fa0fb71e95cc73955c10b582911a613f39/flask_caching/__init__.py#L332-L339

Let me know if you see other behavior during testing

server/routes/api/landing_page.py
```diff
@@ -518,21 +545,63 @@ def populate_category_data(category):
...
+                    spec_and_stat[OVERVIEW][category] = []
+                    for topic in spec_and_stat[category]:
+                        spec_and_stat[OVERVIEW][category].extend(
+                            spec_and_stat[category][topic])
+                        total_charts += len(spec_and_stat[category][topic])
+                        if total_charts > MAX_OVERVIEW_CHART_GROUP:
+                            break
+                else:
+                    # if using the new cache, need to make another cache call
+                    # to get data to be borrowed from category page
```
pradh, line 560: For my understanding, what does data to be borrowed mean?  Elaborate?

server/routes/api/landing_page.py
```diff
@@ -455,23 +476,29 @@ def data(dcid):
...
                 spec_and_stat[category][topic] = filtered_charts
-        if not spec_and_stat[category]:
-            del spec_and_stat[category]
+        if new_landing_page_cache_flag == "false":
+            if not spec_and_stat[category]:
+                del spec_and_stat[category]
+        else:
+            # Don't delete a category if it is in the valid categories list.
+            if not spec_and_stat[category] and category not in raw_page_data[
+                    "validCategories"][dcid]["category"]:
```
pradh, line 485: Wonder if it's better to check before accessing maps to not throw exceptions and handle gracefully? (e.g., if code runs against an import-group not having a change)

server/routes/api/landing_page.py
pradh: Sorry I'd suggested that as a way of doing apples-to-apples performance comparison with and without this change on autopush instance.  Do you have a alternate suggestion?

pradh (commented):
Given the fairly non-trivial change without tests, might be good to include some notes on testing/validation done.

server/routes/api/landing_page.py
beets: We could do this on dev - push current HEAD and send a bunch of requests and look at the API's performance stats, then push the new code and send the same requests. That will also take into account caching, will be isolated and simpler code wise. That could even be done without merging the PR.

server/routes/api/landing_page.py
beets: Can we also please pull this out into a helper for testability.

server/routes/api/landing_page.py
shifucun: If this is for testing purpose, then it's fine. I am just a bit worried that removing the flag can introduce issues. 
Other option would be to deploy two instance before and post a a clean PR, but that would require more endprod work.

beets (commented):
Thanks for the change. Left a comment about testing below.
Please add some unit tests as well.

server/routes/api/landing_page.py
pulkit-s: Done

pulkit-s: Done

server/routes/api/landing_page.py
pulkit-s: Done

server/routes/api/landing_page.py
pulkit-s: it works with just the query string too.

server/routes/api/landing_page.py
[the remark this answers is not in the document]
pulkit-s: Done

server/routes/api/landing_page.py
```diff
@@ -402,18 +411,30 @@ def has_data(data):
...
 def data(dcid):
     """
     Get chart spec and stats data of the landing page for a given place.
     """
-    logging.info("Landing Page: cache miss for %s, fetch and process data ...",
-                 dcid)
+    start_time = time.time()
+    logging.info(
+        "Landing Page: cache miss for place:%s and category:%s and "
+        "cache_flag:%s , fetch and process data ...", dcid,
```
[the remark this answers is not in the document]
pulkit-s: Done

server/routes/api/landing_page.py
pulkit-s: There are cases where we have data for a category but none of those stat vars are configured to be shown on overview page (i.e. isOverview = true). For such cases, we borrow a few charts from the category page to show on the overview page.

server/routes/api/landing_page.py
pulkit-s: Have cleaned it up and removed the flag. Also measured run times for the landing page data module for various cases and tabulated results: https://docs.google.com/spreadsheets/d/1NY6t5m2xY2Jc4EQ4V_8yLvrb2-N0cIEV8ItFREfYPsw/edit?usp=sharing

server/routes/api/landing_page.py
pulkit-s: Done

static/js/place/place.ts
```diff
@@ -152,10 +152,13 @@ async function getChoroplethData(
 async function getLandingPageData(
   dcid: string,
   category: string,
-  locale: string
+  locale: string,
+  cache_flag: string
```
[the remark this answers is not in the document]
pulkit-s: removed this file from the change

pulkit-s:
Have cleaned it up and removed the flag. Also measured run times for the landing page data module for various cases and tabulated results: https://docs.google.com/spreadsheets/d/1NY6t5m2xY2Jc4EQ4V_8yLvrb2-N0cIEV8ItFREfYPsw/edit?usp=sharing

server/routes/api/landing_page.py
```diff
@@ -455,23 +461,28 @@ def data(dcid):
                 del spec_and_stat[category][topic]
             else:
                 spec_and_stat[category][topic] = filtered_charts
-        if not spec_and_stat[category]:
-            del spec_and_stat[category]
+        # Don't delete a category if it is in the valid categories list.
+        if (not spec_and_stat[category]) and ("validCategories"
```
chejennifer, line 465: It looks like we're only doing del spec_and_stat[category] if not spec_and_stat[category] and category not in raw_page_data["validCategories"][dcid]["category"]. But what if it's not spec_and_stat[category], but there's no "validCategories" in raw_page_data or dcid is not in raw_page_data["validCategories"]? Should we also del in those cases or are those cases not possible?

server/routes/api/landing_page.py
```diff
@@ -41,20 +41,23 @@
...
     if stat_vars_string:
-        data['newStatVars'] = stat_vars_string.split('^')
-    response = dc_service.fetch_data('/landing-page',
-                                     data,
-                                     compress=False,
-                                     post=True,
-                                     has_payload=False)
+        stat_var_condition = "&new_stat_vars={}".format(
+            stat_vars_string.split('^'))
+    req_url = '/v1/internal/page/place/{}?category={}'.format(
```
shifucun, line 58: Can you add a wrapper function in https://github.com/datacommonsorg/website/blob/master/server/services/datacommons.py#L163-L169.

We try to put all the API access function in datacommons.py

shifucun: Also try to use the POST endpoint https://github.com/datacommonsorg/mixer/pull/931.

With both changes, you can remove the get_landing_page_data_helper() which was used for caching.

shifucun (commented):
Chatted offline to make build_spec() also based on category, so all the logic in this file will be based on single category.

server/routes/api/landing_page.py
pulkit-s: Done

server/routes/api/landing_page.py
pulkit-s: Done

server/routes/api/landing_page.py
```diff
@@ -518,20 +563,22 @@ def populate_category_data(category):
...
-        categories[category] = gettext(f'CHART_TITLE-CHART_CATEGORY-{category}')
+    for category in list(spec_and_stat.keys()) + list(
+            spec_and_stat[OVERVIEW]
+    ) + raw_page_data["validCategories"][dcid]['category']:
+        if category in configured_categories:
+            categories[category] = gettext(
+                f'CHART_TITLE-CHART_CATEGORY-{category}')
+
+    # Sort in the order of chart config.
+    ordered_category_dict = {}
```
chejennifer, line 578: will dicts keep order?

pulkit-s: Yes, dicts are ordered as of python 3.7

server/routes/api/landing_page.py
```diff
@@ -455,23 +461,31 @@ def data(dcid):
                 del spec_and_stat[category][topic]
             else:
                 spec_and_stat[category][topic] = filtered_charts
-        if not spec_and_stat[category]:
-            del spec_and_stat[category]
+        # Don't delete a category if it is in the valid categories list.
+        if (not spec_and_stat[category]) and ("validCategories"
+                                              in raw_page_data):
+            if dcid in raw_page_data[
```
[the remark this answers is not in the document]
pulkit-s: done

server/routes/api/landing_page.py
```diff
@@ -455,23 +461,31 @@ def data(dcid):
                 del spec_and_stat[category][topic]
             else:
                 spec_and_stat[category][topic] = filtered_charts
-        if not spec_and_stat[category]:
-            del spec_and_stat[category]
+        # Don't delete a category if it is in the valid categories list.
+        if (not spec_and_stat[category]) and ("validCategories"
```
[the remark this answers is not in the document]
pulkit-s: done

static/js/place/menu.tsx
```diff
@@ -112,7 +112,9 @@ class Menu extends React.Component<MenuPropsType> {
   render(): JSX.Element {
     const dcid = this.props.dcid;
     const selectCategory = this.props.selectCategory;
-    const categories = Object.keys(this.props.pageChart);
+    const categories = Object.keys(this.props.categories);
+    // const categories = Object.keys(this.props.pageChart);
```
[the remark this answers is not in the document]
pulkit-s: done

server/routes/api/landing_page.py
```diff
@@ -408,12 +403,23 @@ def data(dcid):
...
+        " , fetching and processing data ...", dcid,
+        request.args.get("category"))
     target_category = request.args.get("category")
-    spec_and_stat = build_spec(current_app.config['CHART_CONFIG'])
+    spec_and_stat = build_spec(current_app.config['CHART_CONFIG'],
+                               target_category)
     new_stat_vars = current_app.config['NEW_STAT_VARS']
-    raw_page_data = get_landing_page_data(dcid, new_stat_vars)
+
+    configured_categories = set()
```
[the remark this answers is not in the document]
pulkit-s: done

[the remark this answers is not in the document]
pulkit-s: done

server/routes/api/landing_page.py
```diff
@@ -96,6 +88,9 @@ def build_spec(chart_config, i18n=True):
     # Map: category -> topic -> [config]
     # Within each category, the topics are sorted.
     for conf in chart_config:
+        # skip if the config category does not match target category.
+        if target_category not in [conf['category'], "Overview"]:
```
shifucun, line 92: add comments on why keep "Overview"?

server/routes/api/landing_page.py
```diff
@@ -455,23 +455,28 @@ def data(dcid):
                 del spec_and_stat[category][topic]
             else:
                 spec_and_stat[category][topic] = filtered_charts
-        if not spec_and_stat[category]:
-            del spec_and_stat[category]
+        # Don't delete a category if it is in the valid categories list.
+        if (not spec_and_stat[category]) and ("validCategories"
```
shifucun, line 459: Can we make a meaningful named boolean variable here for this long condition? some formatting like:

if (a and
    b and
    c and
    d):

Can make it easier to view

server/routes/api/landing_page.py
```diff
@@ -518,20 +554,27 @@ def populate_category_data(category):
...
-    for category in list(spec_and_stat.keys()) + list(spec_and_stat[OVERVIEW]):
-        categories[category] = gettext(f'CHART_TITLE-CHART_CATEGORY-{category}')
+    config_categories = set()
+    ordered_categories = []
+    for conf in current_app.config['CHART_CONFIG']:
+        config_categories.add(conf['category'])
+        ordered_categories.append(conf['category'])
+    for category in list(spec_and_stat.keys()) + list(
+            spec_and_stat[OVERVIEW]
+    ) + raw_page_data["validCategories"][dcid]['category']:
```
shifucun, line 568: Better readability to make a variable to hold this concatenated list.

pulkit-s: Done

server/routes/api/landing_page.py
pulkit-s: Done

server/routes/api/landing_page.py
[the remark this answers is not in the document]
chejennifer: I'm slightly confused on the intentions of when we should be deleting the `spec_and_stat[category]`

Are the following cases possible and are they all cases we would want to delete?
1. not spec_and_stat[category] and category is not in the valid categories list
2. not spec_and_stat[category] and there is no valid categories list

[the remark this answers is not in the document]
pulkit-s: so we want to delete category from the spec and stat when spec_and_stat is empty for that category, and the category is not in the valid categories for that place.

So case 1 is basically the case we are looking for. case 2 is a subset of case 1, but that would probably never occur because it essentially means we have no stat var for that place. If that happens, line 420 handles that case by saying that we have no data for that page.

server/routes/api/landing_page.py
```diff
@@ -518,20 +556,28 @@ def populate_category_data(category):
...
-                    total_charts += len(spec_and_stat[category][topic])
-                    if total_charts > MAX_OVERVIEW_CHART_GROUP:
-                        break
+                populate_additional_category_data(category)
 
     # Get chart category name translations
     categories = {}
-    for category in list(spec_and_stat.keys()) + list(spec_and_stat[OVERVIEW]):
-        categories[category] = gettext(f'CHART_TITLE-CHART_CATEGORY-{category}')
+    config_categories = set()
```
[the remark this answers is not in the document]
pulkit-s: Done

chejennifer (approved):
thanks for the updates!

shifucun (approved):
thanks for making all the updates!

PR #1757, opened 2022-09-21
Fix and refactor browser page provenance issue

shifucun:
The name of a provenance triple is changed from url to the import name. Browser fetches all provenance and url, so the logic needs to be updated to fetch the url.

This PR also updates to use V1 API for triples fetch. This is only used in browser page.

### DIFF
--- a/server/routes/api/browser.py
+++ b/server/routes/api/browser.py
@@ -31,8 +31,8 @@ NO_OBSPERIOD_KEY = 'no_obsPeriod'
 @bp.route('/triples/<path:direction>/<path:dcid>')
 def triples(direction, dcid):
     """Returns all the triples given a node dcid."""
-    if direction != "in" or direction != "out":
-        return 400, "Invalid direction, use 'in' or 'out'"
+    if direction != "in" and direction != "out":
+        return f"Invalid direction {direction}, use 'in' or 'out'", 400
     return dc.triples(dcid, direction).get("triples")
 
 
--- a/server/routes/api/browser.py
+++ b/server/routes/api/browser.py
@@ -20,9 +20,7 @@ from cache import cache
 import services.datacommons as dc
 from flask import Response
 from flask import request
-from flask import current_app
 import routes.api.place as place_api
-import logging
 
 bp = flask.Blueprint('api.browser', __name__, url_prefix='/api/browser')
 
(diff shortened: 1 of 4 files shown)


server/routes/api/browser.py
```diff
@@ -20,21 +20,34 @@
...
 
-@cache.memoize(timeout=3600 * 24)  # Cache for one day.
-@bp.route('/triples/<path:dcid>')
-def triple_api(dcid):
+@bp.route('/triples/<path:direction>/<path:dcid>')
+def triples(direction, dcid):
     """Returns all the triples given a node dcid."""
-    return json.dumps(dc.get_triples([dcid]).get(dcid, []))
+    if direction != "in" and direction != "out":
+        return f"Invalid direction {direction}, use 'in' or 'out'", 400
```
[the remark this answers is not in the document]
shifucun: Done

server/routes/api/browser.py
```diff
@@ -20,21 +20,34 @@
...
+    return dc.triples(dcid, direction).get("triples")
+
+
+@bp.route('/provenance')
+def provenance():
+    """Returns all the provenance information."""
+    resp = dc.triples("Provenance", "in")
+    prov_list = resp.get("triples").get("typeOf").get("nodes")
```
beets, line 43: return default objects if the keys aren't in the response?
```suggestion
    prov_list = resp.get("triples", {}).get("typeOf", {}).get("nodes", [])
```

server/routes/api/browser.py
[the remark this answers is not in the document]
shifucun: Done

static/js/browser/browser_page.tsx
```diff
@@ -176,31 +176,21 @@ export class BrowserPage extends React.Component<
...
       .then(([labelsData, ProvenanceData]) => {
         const provDomain = {};
-        for (const prov of ProvenanceData) {
-          if (this.isProvenanceEntity(prov)) {
-            try {
-              provDomain[prov["subjectId"]] = new URL(prov["subjectName"]).host;
-            } catch (err) {
-              console.log("Invalid url in prov: " + prov["subjectName"]);
-            }
+        for (const prov in ProvenanceData) {
```
beets, line 189: nit: while we're here, can we fix the var casing? s/ProvenanceData/provenanceData/

beets (approved):
thanks for the update - just a few access issues..

static/js/browser/browser_page.tsx
shifucun: Done

server/routes/api/browser.py
shifucun: Done. The data here is fixed (for all provenance), but changed to make it more defensive

shifucun (commented):
Thanks for review!

PR #1765, opened 2022-10-06
Separate geojson data fetch into a separate module

shifucun:
This is a start of efforts to simplify and speed up map tool by having multiple data fetch components. 

The rendering will be based on needed data instead of all data ready.

Also removed "showChart" check for <ChartLoader> in app.js. This turns out reducing the network request from around 60 to 30 for the first chart load (duplicated data request). The checking is moved into chart_loader.tsx. This also reduces extra geojson fetch when geo types do not change.

### DIFF
--- a/static/js/tools/map/chart_loader.tsx
+++ b/static/js/tools/map/chart_loader.tsx
@@ -23,11 +23,8 @@ import axios from "axios";
 import _ from "lodash";
 import React, { useContext, useEffect, useState } from "react";
 
-import { GeoJsonData, GeoJsonFeature, MapPoint } from "../../chart/types";
-import {
-  EUROPE_NAMED_TYPED_PLACE,
-  IPCC_PLACE_50_TYPE_DCID,
-} from "../../shared/constants";
+import { GeoJsonData, MapPoint } from "../../chart/types";
+import { EUROPE_NAMED_TYPED_PLACE } from "../../shared/constants";
 import { FacetSelectorFacetInfo } from "../../shared/facet_selector";
 import {
   EntityObservation,
@@ -59,6 +56,11 @@ import {
   StatVar,
   StatVarWrapper,
 } from "./context";
+import {
+  fetchGeoJson,
+  getGeoJsonDataFeatures,
+  MANUAL_GEOJSON_DISTANCES,
+} from "./geojson";
 import { PlaceDetails } from "./place_details";
 import {
   BEST_AVAILABLE_METAHASH,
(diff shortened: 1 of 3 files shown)


static/js/tools/map/chart_loader.tsx
```diff
@@ -182,7 +187,7 @@ export function ChartLoader(): JSX.Element {
         setLegendBoundsPerCapita(rawData, statVar, display);
       }
     }
-  }, [rawData, statVar.value.metahash, statVar.value.perCapita]);
+  }, [rawData, geoJson, statVar.value.metahash, statVar.value.perCapita]);
```
shifucun, line 190: adding "geoJson" here is crucial

static/js/tools/map/chart_loader.tsx
```diff
@@ -739,8 +678,26 @@ function loadChartData(
     return;
   }
   const calculateRatio = statVar.perCapita && statVar.denom ? true : false;
-  // populate mapValues with data value for each geo that we have geoJson data for.
-  for (const geoFeature of rawData.geoJsonData.features) {
+  // populate mapValues with data value for each geo that we have geoJson data
+  // for.
+  if (!geoJsonData) {
```
shifucun, line 683: return early here is crucial

beets: could you add the note in code too (along with why it's crucial for future code updates)?

static/js/tools/map/chart_loader.tsx
```diff
@@ -194,7 +198,7 @@ export function ChartLoader(): JSX.Element {
 
   if (chartData === undefined) {
     return null;
-  } else if (_.isEmpty(chartData.geoJsonData)) {
+  } else if (_.isEmpty(geoJson)) {
```
beets, line 224: should we add a way to differentiate between no geojson available vs geojson still loading?

shifucun: Good point! Right now chartData is defined only when there is geoJson data, so it will always exit in the if statement above. As we do more refactor, will need a different state for this.

static/js/tools/map/chart_loader.tsx
shifucun: This is moved into a separate useEffect and I added a note there. The statement management will need more cleanup as we proceed with the refactor

chejennifer (approved):
thanks for starting on extracting out data fetchers!

PR #1770, opened 2022-10-09
[Memcache Update]: Add redis cache for feeding-america; Add direct cache for observation API

shifucun:


server/configmodule.py
```diff
@@ -36,6 +36,8 @@ class Config:
     # be set for local development. Website deployed to GKE bundles the mixer
     # as a private service accessible via localhost.
     API_PROJECT = ''
+    # Do not use memcache so frequent updated data can be seen.
+    NO_CACHE = False
```
beets, line 40: Nit: it might be easier to configure and understand if it was a positive bool, i.e. USE_MEMCACHE = True (and we override in the various configs)

beets (approved):
nice, thanks bo!

server/configmodule.py
shifucun: Done

PR #1775, opened 2022-10-11
[map refactor] Add data fetcher modules and a chart store

shifucun:
This is a no-op change to existing code and prepares for future updates of chart_loader.

The goal is to remove the dependency of data fetch and processing. Ideally, the tool only needs to fetch the needed data and will render the map when all necessary data is ready. The other data can be loaded on the background or on-demand, which can be controlled independently without touching the component logic.

Some design thoughts:

- Each data load should be in a separate module, which fetches data and sends the raw data to the desired component. These modules are put in /fetcher folder. New data fetcher should be created in the same pattern.

- All the raw data are used in various ways to construct the chart data, the construction depends on context (user input) and is dependent on data availability. Here I created a ChartStore to hold all the raw data and also the context when they are created. This has the following benefits:
1.  Allow each fetcher to independently "dispatch" the data. The data is merged using "useReducer" hook in the UI component.
2. The UI component can check if the data is ready by comparing the context in ChartStore vs the current context.
3. In future change, the async data processing dependency can be resolved by "useEffect()".  By checking the readiness of the data, each useEffect() can exit early if needed.

### DIFF
--- a/static/js/tools/map/chart_loader.tsx
+++ b/static/js/tools/map/chart_loader.tsx
@@ -23,7 +23,13 @@ import axios from "axios";
 import geoblaze from "geoblaze";
 import { GeoRaster } from "georaster-layer-for-leaflet";
 import _ from "lodash";
-import React, { useContext, useEffect, useState } from "react";
+import React, {
+  useCallback,
+  useContext,
+  useEffect,
+  useReducer,
+  useState,
+} from "react";
 
 import { GeoJsonData, MapPoint } from "../../chart/types";
 import { EUROPE_NAMED_TYPED_PLACE } from "../../shared/constants";
@@ -50,6 +56,7 @@ import { setUpBqButton } from "../shared/bq_utils";
 import { getMatchingObservation, getUnit } from "../shared_util";
 import { getNonPcQuery, getPcQuery } from "./bq_query_utils";
 import { Chart, MAP_TYPE } from "./chart";
+import { chartStoreReducer, emptyChartStore } from "./chart_store";
 import {
   Context,
   DisplayOptionsWrapper,
@@ -58,18 +65,16 @@ import {
   StatVar,
   StatVarWrapper,
 } from "./context";
-import {
-  fetchGeoJson,
-  getGeoJsonDataFeatures,
-  MANUAL_GEOJSON_DISTANCES,
-} from "./geojson";
+import { fetchGeoJson } from "./geojson";
 import { PlaceDetails } from "./place_details";
 import {
   BEST_AVAILABLE_METAHASH,
   DataPointMetadata,
+  getGeoJsonDataFeatures,
   getLegendBounds,
   getPlaceChartData,
   getTimeSliderDates,
+  MANUAL_GEOJSON_DISTANCES,
 } from "./util";
 
 interface ChartRawData {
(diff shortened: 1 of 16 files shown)


static/js/tools/map/chart_store.ts
```diff
@@ -0,0 +1,199 @@
...
+  statVar: StatVar
+): boolean => {
+  if (_.isEmpty(dataWrapper)) {
+    return false;
+  }
+  const ctx = dataWrapper.context;
+  if (_.isEmpty(ctx)) {
+    return false;
+  }
+  switch (type) {
```
chejennifer, line 127: nit: I feel like this is a large switch statement with lots of checks for each case making it not the easiest to read. Maybe we can extract out functions for each case and just call the function for each case? or even make a map of type to function & just do something like `return TYPE_TO_READY_FN[type]`

static/js/tools/map/chart_store.ts
```diff
@@ -0,0 +1,199 @@
...
+    case ChartDataType.breadcrumbDenomStat:
+      return (
+        placeInfo.enclosingPlace.dcid === ctx.placeInfo.enclosingPlace.dcid &&
+        statVar.denom === ctx.statVar.denom
+      );
+    case ChartDataType.mapPointStat:
+      return (
+        placeInfo.enclosingPlace.dcid === ctx.placeInfo.enclosingPlace.dcid &&
+        statVar.dcid === ctx.statVar.dcid &&
+        statVar.dcid === ctx.statVar.dcid &&
```
chejennifer, line 168: this is a repeat of the last line?

Also should check that statVar.mapPointSv is the same

static/js/tools/map/chart_store.ts
```diff
@@ -0,0 +1,199 @@
...
+        statVar.date === ctx.statVar.date
+      );
+    case ChartDataType.breadcrumbDenomStat:
+      return (
+        placeInfo.enclosingPlace.dcid === ctx.placeInfo.enclosingPlace.dcid &&
+        statVar.denom === ctx.statVar.denom
+      );
+    case ChartDataType.mapPointStat:
+      return (
+        placeInfo.enclosingPlace.dcid === ctx.placeInfo.enclosingPlace.dcid &&
```
chejennifer, line 166: need to also check that placeInfo.mapPointPlaceType is the same

static/js/tools/map/fetcher/map_point_stat.ts
```diff
@@ -0,0 +1,83 @@
...
+      statVar: {
+        dcid: statVar,
+        date,
+      },
+    },
+  };
+  axios
+    .get<PointApiResponse>("/api/observations/point/within", {
+      params: {
+        child_type: childType,
```
chejennifer, line 60: sorry, this is not relevant to this PR, but we should make flask endpoint params be camelcase. I believe that was the pattern we were using before

static/js/tools/map/fetcher/map_point_stat.ts
```diff
@@ -0,0 +1,83 @@
...
+      if (_.isEmpty(resp.data.data[statVar])) {
+        action.error = "error fetching map point stat data";
+      } else {
+        action.payload = {
+          data: resp.data.data[statVar],
+          facets: resp.data.facets,
+        } as EntityObservationWrapper;
+      }
+      dispatch(action);
+      console.log("map point stat dispatched");
```
chejennifer, line 77: I don't think we want to have console.logs in our pushed code unless necessary. The purpose of console.log should be for debugging code and that shouldn't really be exposed in production.

Applies here and all the other console.logs

static/js/tools/map/fetcher/denom_stat.ts
```diff
@@ -0,0 +1,80 @@
...
+  statVar: string,
+  dispatch: Dispatch<ChartStoreAction>
+): void {
+  const action: ChartStoreAction = {
+    type: ChartDataType.denomStat,
+    error: null,
+    context: {
+      placeInfo: {
+        enclosingPlace: {
+          // The last place is the selected place.
```
chejennifer, line 44: I'm not sure what this comment is talking about?

static/js/tools/map/fetcher/breadcrumb_stat.ts
```diff
@@ -0,0 +1,81 @@
...
+  statVar: string,
+  date: string,
+  dispatch: Dispatch<ChartStoreAction>
+): void {
+  const action: ChartStoreAction = {
+    type: ChartDataType.breadcrumbStat,
+    context: {
+      placeInfo: {
+        enclosingPlace: {
+          dcid: entities[entities.length - 1],
```
chejennifer, line 44: maybe add a note that the entities are in a specific order

static/js/tools/map/fetcher/geojson.ts
```diff
@@ -0,0 +1,114 @@
...
+      dispatch(action);
+      console.log("geojson dispatched");
+    })
+    .catch(() => {
+      action.error = "error fetching geo json data";
+      dispatch(action);
+    });
+}
+
+export function getGeoJsonDataFeatures(
```
chejennifer, line 70: do we still want to keep this function in the fetcher file? It looks like all the other fetcher files are just purely a data fetch function

static/js/tools/map/chart_store.ts
shifucun: I later realized it's better to move this as useCallback() function inside <ChartLoader> and only use the necessary dependent variables instead of StatVar and PlaceInfo struct. That way, the function return value changes only when those variables change, which can eliminate unnecessary re-renders.

With that, I split this into separate useCallback functions.

static/js/tools/map/chart_store.ts
shifucun: Done

static/js/tools/map/chart_store.ts
shifucun: Done

static/js/tools/map/fetcher/breadcrumb_stat.ts
shifucun: Done

static/js/tools/map/fetcher/default_stat.ts
```diff
@@ -0,0 +1,83 @@
...
+
+import {
+  EntityObservationWrapper,
+  PointApiResponse,
+} from "../../../shared/stat_types";
+import { stringifyFn } from "../../../utils/axios";
+import { ChartDataType, ChartStoreAction } from "../chart_store";
+import { getDate } from "../util";
+
+export function fetchDefaultStat(
```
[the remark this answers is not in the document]
shifucun: fetchMapPointStat is kinda of different now, as I missed the map point sv and map point place type.

In general, I feel it's okay for repetition here since the type of data is limited and it allows update independently.

static/js/tools/map/fetcher/denom_stat.ts
shifucun: removed

static/js/tools/map/fetcher/geojson.ts
shifucun: Moved it to util.ts

static/js/tools/map/fetcher/map_point_stat.ts
shifucun: sg. I will make a change in follow up PR

static/js/tools/map/fetcher/map_point_stat.ts
shifucun: Sg. I will make a todo to remove once all the refactor is done.

static/js/tools/map/chart_loader.tsx
```diff
@@ -133,6 +138,140 @@ export function ChartLoader(): JSX.Element {
...
+  // +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
+  // START useReducer
+  const [chartStore, dispatch] = useReducer(chartStoreReducer, emptyChartStore);
+  // END useReducer
+  // -------------------------------------------------------------------------
+
+  // +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
+  // The functions below are memoized and only change when the dependant
+  // variables change.
+  const geoJsonReady = useCallback(() => {
```
chejennifer, line 150: does useCallback have to be in this file or can we separate it out into a different file that just holds all these callbacks?

static/js/tools/map/chart_loader.tsx
```diff
@@ -133,6 +138,140 @@ export function ChartLoader(): JSX.Element {
...
+  }, [
+    chartStore.denomStat.context,
+    placeInfo.value.enclosingPlace.dcid,
+    placeInfo.value.enclosedPlaceType,
+    statVar.value.denom,
+  ]);
+
+  const breadcrumbStatReady = useCallback(() => {
+    const c = chartStore.breadcrumbStat.context;
+    console.log("breadcrumbStatReady");
```
chejennifer, line 215: are these console.logs needed? because I notice some of these callbacks have console logs and some don't

chejennifer (commented):
thanks for the changes!

static/js/tools/map/chart_loader.tsx
shifucun: Looks like can create custom hook for this. moved them into ready_hook.ts now

static/js/tools/map/chart_loader.tsx
shifucun: removed them now

static/js/tools/map/chart_store.ts
```diff
@@ -0,0 +1,199 @@
...
+import { GeoJsonData, MapPoint } from "../../chart/types";
+import {
+  EntityObservationListWrapper,
+  EntityObservationWrapper,
+  EntitySeriesWrapper,
+} from "../../shared/stat_types";
+import { DataContext, PlaceInfo, StatVar } from "./context";
+
+export enum ChartDataType {
+  geoJson,
```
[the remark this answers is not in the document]
shifucun: I rely on the enum string to access the actual field below.. This is a bit hacky but saves a mapping, wdyt?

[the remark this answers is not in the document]
chejennifer: ah, I see, but could you just set the key using the enum? and then access using the enum? If not, this is ok to keep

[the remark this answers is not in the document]
chejennifer: ah, ok, that makes sense, but should we just make the keys using the enum strings then?

chejennifer (approved):
thanks for the changes! just one suggestion for the enums

static/js/tools/map/chart_store.ts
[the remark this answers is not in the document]
shifucun: updated

shifucun:
Thanks for the detailed review!

beets (commented):
Sorry for the late review! In general, please leave more comments & documentation in the code, especially as we're cleaning it up, so we can improve maintainability :)

PR #1776, opened 2022-10-11
Control map tool display state inside each component instead of fetching context from app.js

shifucun:
When context is changed, App() component get re-rendered, hence the children components including <ChartLoader>. In the same time, <ChartLoader> listens to context change and re-render based on that. This creates extra rendering and does not interact well since <ChartLoader> has lots of async data fetch.

### DIFF
--- a/static/js/tools/map/title.tsx
+++ b/static/js/tools/map/title.tsx
@@ -1,5 +1,5 @@
 /**
- * Copyright 2021 Google LLC
+ * Copyright 2022 Google LLC
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/static/js/tools/map/title.tsx
+++ b/static/js/tools/map/title.tsx
@@ -1,0 +1,36 @@
+/**
+ * Copyright 2021 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


shifucun:
The context is not fully removed since I need to keep `isLoading`. Will fully fix this once the chart loading is finalized.

static/js/tools/map/title.tsx
```diff
@@ -0,0 +1,36 @@
+/**
+ * Copyright 2021 Google LLC
```
chejennifer, line 2: nit: 2022

static/js/tools/map/title.tsx
```diff
@@ -0,0 +1,36 @@
...
+/**
+ * Title component before a chart is shown.
+ */
+
+import React, { useContext } from "react";
+
+import { Context } from "./context";
+import { ifShowChart } from "./util";
+
+export function Title(): JSX.Element {
```
chejennifer, line 26: I wonder if instead of having a separate title component and info component, we could just have one LandingPage component that returns null if showChart is true? And then have another ChartPage component that returns null if showChart is false? Then App would render both those components and PlaceSelector can go into both those components.

chejennifer (commented):
Thanks for the change!

static/js/tools/map/title.tsx
shifucun: Done

static/js/tools/map/title.tsx
shifucun: Done

static/js/tools/map/title.tsx
shifucun: Done

static/js/tools/map/title.tsx
chejennifer: ah, ok then

chejennifer (approved):

static/js/tools/map/title.tsx
shifucun: We may come back to this later. I also feel there could be a better way to organize it.

PR #1777, opened 2022-10-11
[map] update geotiff flag so that it's in the url args instead of the url params

chejennifer:
- this allows flask to read the flag and only load the leaflet css if leaflet flag is set

### DIFF
--- a/static/js/tools/map/app.tsx
+++ b/static/js/tools/map/app.tsx
@@ -28,6 +28,7 @@ import { Info } from "./info";
 import { PlaceOptions } from "./place_options";
 import { StatVarChooser } from "./stat_var_chooser";
 import {
+  ALLOW_LEAFLET_URL_ARG,
   applyHashDisplay,
   applyHashPlaceInfo,
   applyHashStatVar,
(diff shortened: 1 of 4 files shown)


static/js/tools/map/app.tsx
```diff
@@ -114,9 +115,16 @@ function updateHash(context: ContextType): void {
...
+  if (context.display.value.allowLeaflet) {
+    args += `?${ALLOW_LEAFLET_URL_ARG}=1`;
+  }
   const newHash = encodeURIComponent(hash);
   const currentHash = location.hash.replace("#", "");
-  if (newHash && newHash !== currentHash) {
-    history.pushState({}, "", `${MAP_REDIRECT_PREFIX}#${newHash}`);
+  const currentArgs = location.search;
+  if (newHash && (newHash !== currentHash || args !== currentArgs)) {
+    history.pushState({}, "", `${MAP_REDIRECT_PREFIX}${args}#${newHash}`);
```
beets, line 128: We should forward along all args (not just leaflet), and then append the hash. Query args are a good place for persistent effects (e.g. allow various features). This works for now, and we can update that in a TODO.

Perhaps also rename `MAP_REDIRECT_PREFIX` to `MAP_URL_PATH` so we don't accidentally add other args / path to the const.

static/js/tools/map/util.ts
```diff
@@ -344,7 +342,11 @@ export function getRedirectLink(
     parentPlaces: [],
     selectedPlace,
   });
-  return `${MAP_REDIRECT_PREFIX}#${encodeURIComponent(hash)}`;
+  let args = "";
+  if (displayOptions.allowLeaflet) {
+    args += `?${ALLOW_LEAFLET_URL_ARG}=1`;
+  }
+  return `${MAP_REDIRECT_PREFIX}${args}#${encodeURIComponent(hash)}`;
```
beets, line 349: please add a manual comment equivalent to IFTT to update both files when either is edited

beets (approved):

static/js/tools/map/app.tsx
chejennifer: added todo & renamed the constant

static/js/tools/map/util.ts
chejennifer: done

PR #1783, opened 2022-10-13
Use fetch hook for map points and european countries

shifucun:
The map point loading is effectively in the background.

This is a show case of how to migrate the current code to be React hook based.

### DIFF
--- a/static/js/tools/map/fetcher/european_countries.ts
+++ b/static/js/tools/map/fetcher/european_countries.ts
@@ -24,7 +24,7 @@ import { EUROPE_NAMED_TYPED_PLACE } from "../../../shared/constants";
 import { NamedPlace } from "../../../shared/types";
 import { getEnclosedPlacesPromise } from "../../../utils/place_utils";
 
-export function useFetchEuropeanCountries() {
+export function useFetchEuropeanCountries(): Array<NamedPlace> {
   const [data, setData] = useState<Array<NamedPlace>>();
   useEffect(() => {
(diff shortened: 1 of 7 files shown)


static/js/tools/map/fetcher/european_countries.ts
```diff
@@ -18,17 +18,21 @@
...
-export function fetchEuropeanCountries(
-  setEuropeanCountries: (data: Array<NamedPlace>) => void
-): void {
-  getEnclosedPlacesPromise(EUROPE_NAMED_TYPED_PLACE.dcid, "Country").then(
-    (resp: Array<NamedPlace>) => {
-      setEuropeanCountries(resp);
-      console.log("european countries loaded");
-    }
-  );
+export function useFetchEuropeanCountries(): Array<NamedPlace> {
```
chejennifer, line 27: Does this need to be a useEffect? since it doesn't look like this is watching for any changes?

chejennifer (commented):
this is very clean! Just one small thing

static/js/tools/map/fetcher/european_countries.ts
shifucun: In line 30, it is an async call, so here needs to "useState" to catch the response from the promise. 

Then needs useEffect() to work together here IIUC. It only fetches once (with [] dependency).

chejennifer: ok, I see what you mean

chejennifer (approved):

PR #1784, opened 2022-10-14
Use context directly in map sub components

shifucun:
All the map children components can directly access the context instead of getting them through the props.

### DIFF
--- a/static/js/tools/map/chart_loader.tsx
+++ b/static/js/tools/map/chart_loader.tsx
@@ -26,7 +26,6 @@ import _ from "lodash";
 import React, { useContext, useEffect, useReducer, useState } from "react";
 
 import { GeoJsonData, MapPoint } from "../../chart/types";
-import { EUROPE_NAMED_TYPED_PLACE } from "../../shared/constants";
 import { FacetSelectorFacetInfo } from "../../shared/facet_selector";
 import {
   EntityObservation,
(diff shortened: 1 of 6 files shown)


static/js/tools/map/chart_loader.tsx
```diff
@@ -69,16 +65,6 @@ import { useFetchGeoJson } from "./fetcher/geojson";
 import { useFetchMapPointCoordinate } from "./fetcher/map_point_coordinate";
 import { useFetchMapPointStat } from "./fetcher/map_point_stat";
 import { PlaceDetails } from "./place_details";
-import {
```
shifucun, line 72: These import are not used yet, so removing for now

static/js/tools/map/d3_map.tsx
```diff
@@ -75,39 +68,133 @@ const DEBOUNCE_INTERVAL_MS = 30;
...
 export function D3Map(props: D3MapProps): JSX.Element {
+  const { placeInfo, statVar, display } = useContext(Context);
+
   const [errorMessage, setErrorMessage] = useState("");
   const [zoomTransformation, setZoomTransformation] = useState(
     DEFAULT_ZOOM_TRANSFORMATION
   );
   const chartContainerRef = useRef<HTMLDivElement>();
 
-  function replot() {
```
shifucun, line 84: Here making draw() into a useCallback() effect, so it updates based on dependency changes, and ease the need for replot().

static/js/tools/map/leaflet_map.tsx
```diff
@@ -64,46 +69,25 @@ export function LeafletMap(props: LeafletMapProps): JSX.Element {
     });
   }, []);
 
-  // Replot when data changes.
-  // TODO: handle resizing of chart area
-  useEffect(() => {
```
shifucun, line 69: This is moved to below.

chejennifer (approved):
nice use of useCallback()!

PR #1788, opened 2022-10-14
[map] geotiff updates

chejennifer:
- add georaster fetcher
- remove per capita option if rendering leaflet map
- dynamically resize leaflet map

### DIFF
--- a/static/js/tools/map/fetcher/georaster.ts
+++ b/static/js/tools/map/fetcher/georaster.ts
@@ -1,0 +1,86 @@
+/**
+ * Copyright 2022 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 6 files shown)


static/js/tools/map/fetcher/georaster.ts
```diff
@@ -0,0 +1,86 @@
...
+import { Context } from "../context";
+
+export function useFetchGeoRaster(dispatch: Dispatch<ChartStoreAction>): void {
+  const { placeInfo, statVar, display } = useContext(Context);
+  useEffect(() => {
+    const contextOk =
+      placeInfo.value.enclosingPlace.dcid &&
+      placeInfo.value.enclosedPlaceType &&
+      statVar.value.dcid &&
+      (!statVar.value.perCapita || statVar.value.denom) &&
```
shifucun, line 35: why checking statVar.value.denom here?

shifucun (commented):
nice adding the fetcher!

static/js/tools/map/fetcher/georaster.ts
chejennifer: ah, I forgot we're not doing per capita for geotiff for now. I was thinking that since we would fetch geotiff with per capita already calculated, we would need to pass the denom. Removed

shifucun (approved):

PR #1790, opened 2022-10-15
Refactor breadcrumb values to use the custom hook

shifucun:
Add a compute hook to compute the breadcrumb values based on breadcrumbStat and breadcrumbDenomStat data.

### DIFF
--- a/static/js/tools/map/chart_loader.tsx
+++ b/static/js/tools/map/chart_loader.tsx
@@ -52,6 +52,7 @@ import {
   chartStoreReducer,
   emptyChartStore,
 } from "./chart_store";
+import { useComputeBreadcrumbValues } from "./compute_hooks";
 import {
   Context,
   DisplayOptionsWrapper,
@@ -60,6 +61,8 @@ import {
   StatVar,
   StatVarWrapper,
 } from "./context";
+import { useFetchBreadcrumbDenomStat } from "./fetcher/breadcrumb_denom_stat";
+import { useFetchBreadcrumbStat } from "./fetcher/breadcrumb_stat";
 import { useFetchEuropeanCountries } from "./fetcher/european_countries";
 import { useFetchGeoJson } from "./fetcher/geojson";
 import { useFetchMapPointCoordinate } from "./fetcher/map_point_coordinate";
@@ -80,7 +83,7 @@ interface ChartRawData {
   allEnclosedPlaceStat: EntityObservationList;
   metadataMap: Record<string, StatMetadata>;
   population: EntitySeries;
-  breadcrumbPlaceStat: EntityObservation;
+  breadcrumbPlaceStat?: EntityObservation;
   dataDate: string;
 
   // Map of metahash to array of ~10 dates for time slider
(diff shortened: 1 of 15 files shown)


static/js/tools/map/fetcher/breadcrumb_denom_stat.ts
```diff
@@ -50,11 +50,12 @@ export function useFetchBreadcrumbDenomStat(
       type: ChartDataType.BREADCRUMB_DENOM_STAT,
       context: {
         placeInfo: {
-          enclosingPlace: {
+          selectedPlace: {
             // entities are the parent places + the enclosing place.
             // so the last element is the enclosing place.
             dcid: placeDcids[placeDcids.length - 1],
             name: "",
+            types: null,
```
chejennifer, line 58: curious why we add types for selectedPlace here but not in the default_stat fetcher or any other fetcher

static/js/tools/map/fetcher/default_stat.ts
```diff
@@ -80,8 +80,8 @@ export function useFetchDefaultStat(
             facets: resp.data.facets,
           } as EntityObservationWrapper;
         }
-        dispatch(action);
         console.log("default stat dispatched");
+        dispatch(action);
```
chejennifer, line 84: curious what the difference in order of console.log and dispatch(action) causes?

static/js/tools/map/compute_hooks.ts
```diff
@@ -67,3 +77,49 @@ export function useUpdateGeoJson(
...
+    const breadcrumbValues = {};
+    for (const place in chartStore.breadcrumbStat.data.data) {
+      const placeChartData = getPlaceChartData(
+        chartStore.breadcrumbStat.data.data,
+        place,
+        ifRatio,
+        chartStore.breadcrumbDenomStat.data
+          ? chartStore.breadcrumbDenomStat.data.data
+          : null,
+        Object.assign(
```
chejennifer, line 104: instead of creating this object for every place in the for loop, can we do it once outside the for loop?

chejennifer (commented):
nice! Just a few small comments

static/js/tools/map/chart_loader.tsx
```diff
@@ -80,7 +83,7 @@ interface ChartRawData {
   allEnclosedPlaceStat: EntityObservationList;
   metadataMap: Record<string, StatMetadata>;
   population: EntitySeries;
-  breadcrumbPlaceStat: EntityObservation;
+  breadcrumbPlaceStat?: EntityObservation;
```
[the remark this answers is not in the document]
shifucun: This is only kept since the time slider sampled data uses it. Keep for now so time slider is not touched.

static/js/tools/map/fetcher/default_stat.ts
shifucun: I observed delay after dispatch, so the log here could be after the reducer action. Move it above so the log sequence is more clear

static/js/tools/map/fetcher/breadcrumb_denom_stat.ts
shifucun: selectedPlace uses the type "NamedTypedPlace" which requires types. This maybe due to some other map logic?

shifucun (commented):
thanks for review!

chejennifer (approved):
thanks for the changes!

PR #1791, opened 2022-10-17
Use V1 API and a custom hook for computing sample dates in time slider

shifucun:


server/routes/api/observation_dates.py
```diff
@@ -0,0 +1,43 @@
...
+bp = Blueprint("observation-dates", __name__)
+
+
+@bp.route('/api/observation-dates')
+def observation_dates():
+    """
+    Given ancestor place, child place type and stat vars, return the dates that
+	have data for each stat var across all child places.
+    """
+    ancestor_place = request.args.get('ancestorPlace')
```
chejennifer, line 28: since within place endpoints all use the params `parentEntity`, `childType`, and `variables`, should we keep params consistent and use those same parameters?

static/js/tools/map/compute_hooks.ts
```diff
@@ -124,3 +126,20 @@ export function useComputeBreadcrumbValues(chartStore: ChartStore): {
...
+  const allDatesReady = useAllDatesReady(chartStore);
+  return useMemo(() => {
+    if (!allDatesReady()) {
+      return {};
+    }
+    const allSampledDates = getTimeSliderDates(
+      chartStore.allDates.data.facets,
+      chartStore.allDates.data.data
+    );
+    console.log(allSampledDates);
```
chejennifer, line 142: did you mean to keep this console.log?

static/js/tools/map/util.ts
```diff
@@ -572,47 +573,55 @@ function getMetahashMap(
  * @param placeStatDateWithinPlace
```
chejennifer, line 573: update the function comment? Also, include that observationDates is arranged in order (if that's correct)?

static/js/tools/map/util.ts
```diff
@@ -572,47 +573,55 @@ function getMetahashMap(
...
+      Math.ceil(entityCount.length / NUM_SAMPLE_DATES)
+    );
+    const selectedDates: Array<string> = [];
+    let weight = 0;
+    for (let i = 0; i < entityCount.length - 1; i += increment) {
+      selectedDates.push(entityCount[i].date);
+      weight += entityCount[i].count;
     }
+    // Include most recent date
+    selectedDates.push(entityCount[entityCount.length - 1].date);
```
chejennifer, line 611: might want to check if most recent date has already been added during the for loop before adding again?

static/js/tools/map/util.ts
```diff
@@ -572,47 +573,55 @@ function getMetahashMap(
...
+    weight += entityCount[entityCount.length - 1].count;
+    sampleDates[facet] = selectedDates;
+    facetWeight[facet] = weight;
   }
+  let bestWeight = 0;
+  let bestAvailable = "";
+  for (const facet of latestFacets) {
+    if (facetWeight[facet] > bestWeight) {
+      bestWeight = facetWeight[facet];
+      bestAvailable = facet;
```
chejennifer, line 621: since sampleDates is a map of facetId to list of selected dates, shouldn't bestAvailable be a list of dates? or am I misunderstanding this?

static/js/tools/map/chart_loader.tsx
```diff
@@ -144,7 +143,9 @@ export function ChartLoader(): JSX.Element {
   useFetchBreadcrumbDenomStat(chartStore, dispatch);
   useFetchMapPointStat(dispatch);
   useFetchGeoRaster(dispatch);
+  useFetchAllDates(dispatch);
 
+  const allSampleDates = useComputeSampledDates(chartStore);
```
chejennifer, line 148: nit: maybe keep the naming consistent with either "sampleDates" or "sampledDates"

static/js/tools/map/chart_loader.tsx
```diff
@@ -581,19 +579,12 @@ function fetchData(
...
         let legendBounds: Record<string, [number, number, number]> = {};
-        if (BEST_AVAILABLE_METAHASH in sampleDates) {
+        if (BEST_AVAILABLE_METAHASH in allSampleDates) {
           // Set dates for "Best Available" to best series
-          const bestAvailableHash = sampleDates[BEST_AVAILABLE_METAHASH][0];
-          sampleDates[BEST_AVAILABLE_METAHASH] =
-            sampleDates[sampleDates[BEST_AVAILABLE_METAHASH][0]];
+          const bestAvailableHash = allSampleDates[BEST_AVAILABLE_METAHASH][0];
+          allSampleDates[BEST_AVAILABLE_METAHASH] =
+            allSampleDates[allSampleDates[BEST_AVAILABLE_METAHASH][0]];
```
chejennifer, line 587: instead of this logic here, can we just have allSampleDates[BEST_AVAILABLE_METAHASH] populated with the correct list of dates in the compute hook? Or is there a reason it needs to be populated here?

chejennifer (commented):
nice! Few small comments, but Natalie might know the nuances of time slider better than me so maybe also wait on her review!

static/js/tools/map/chart_loader.tsx
shifucun: Meant to use useComputeSampleDates(). updated

static/js/tools/map/chart_loader.tsx
shifucun: I have not dig deep into the logic here but just to keep the logic matching. In next step where legend bound is computed, let me clean this up.

static/js/tools/map/util.ts
n-h-diaz: yes that makes more sense!

static/js/tools/map/chart_loader.tsx
n-h-diaz: I think it can be computed in the hook - sorry I'm not sure why I did it like this before

n-h-diaz: okay actually thinking about it more, I think the reason it's like this was because we wanted to know which facet we're choosing as the "best available"
without time slider, the "best available" view could potentially draw from multiple facets and dates at once, but now since we're forcing a date, I think we had decided to just show a single facet (whichever had the best coverage of observations). so this gives us access to which facet to display. this is probably not the neatest way to do this though, so makes sense to clean this up

n-h-diaz (approved):
thanks!! I think this looks okay for making it work as before :)
however, we did want to revisit which dates are selected (this can be done separately, or here if it's straightforward)
for example, for daily time series, it might be good to pick the same date each year for seasonality (currently it just divides evenly across all available dates)
also we should possibly rethink how to pick the "best available" dates (i.e. do we want the results to all come from the same facet as it does now? can we merge across time series? see other comment)

shifucun:
> thanks!! I think this looks okay for making it work as before :) however, we did want to revisit which dates are selected (this can be done separately, or here if it's straightforward) for example, for daily time series, it might be good to pick the same date each year for seasonality (currently it just divides evenly across all available dates) also we should possibly rethink how to pick the "best available" dates (i.e. do we want the results to all come from the same facet as it does now? can we merge across time series? see other comment)

Got it. So we can see how this could be represented when time slider data, operation is put in its own hook. Thanks for checking!

chejennifer (approved):
thanks for the updates!

static/js/tools/map/chart_loader.tsx
chejennifer: Sounds good to keep it as is for this PR and revisit this in a future PR!

shifucun:
thank you both for the reviews

PR #1792, opened 2022-10-17
Add a hook to compute map point values; Add reducer to capture sources and metadata

shifucun:


static/js/tools/map/chart_loader.tsx
```diff
@@ -132,7 +136,26 @@ export function ChartLoader(): JSX.Element {
   //   };
   // }, []);
 
+  function sourcesReducer(
```
chejennifer, line 139: wonder if we should have these reducers in a separate file like with chartStoreReducer?

static/js/tools/map/chart_loader.tsx
```diff
@@ -132,7 +136,26 @@ export function ChartLoader(): JSX.Element {
...
+    return { ...metadata, ...payload };
+  }
+
   // +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
+  const [sources, dispatchSources] = useReducer(
+    sourcesReducer,
+    new Set<string>()
+  );
+  const [metadata, dispatchMetadata] = useReducer(metadataReducer, {});
   const [chartStore, dispatch] = useReducer(chartStoreReducer, emptyChartStore);
```
chejennifer, line 159: with additional dispatch functions, maybe we can call this dispatchChartStore?

chejennifer (commented):
nice job separating out map point values computation!

chejennifer (approved):
thanks for the changes!

PR #1797, opened 2022-10-19
Refactor map value data fetch, compute and time slider logic

shifucun:
* Time Slider: simplify logic by removing the callback functions and let props.current_date to trigger its rendering. The component is responsible for set date context(). With this, data is fetched for each date as if a normal rendering in map tool. This can save the long initial data load, but could result in longer frame rate.
* Make compute hooks for map values and breadcrumb values, sample dates and other chart data.
* Date context: Separate date as its own context so it can be set in time slider with minimal effect in render logic. 
* Loading spinner should be re-visited.


static/js/tools/map/compute/map_value_dates.ts
```diff
@@ -0,0 +1,162 @@
...
+    const sources = new Set<string>();
+    const mapDates = new Set<string>();
+    const metadata = {};
+    const facets = Object.assign(
+      [],
+      chartStore.defaultStat.data.facets,
+      chartStore.denomStat.data ? chartStore.denomStat.data.facets : {},
+      chartStore.allStat.data ? chartStore.allStat.data.facets : {}
+    );
+    for (const geoFeature of chartStore.geoJson.data.features) {
```
chejennifer, line 84: I wonder if instead of looping through chartStore.geoJson.data.features, it would be better to just loop through the stat data. That way map values won't have to depend on geoJson

static/js/tools/map/compute/map_value_dates.ts
```diff
@@ -0,0 +1,162 @@
...
+      const placeDcid = geoFeature.properties.geoDcid;
+      let wantedFacetData = chartStore.defaultStat.data.data;
```
chejennifer, line 86: wantedFaceData can probably be calculated (line 86 -95) outside the for loop because it will be the same for every place?

static/js/tools/map/chart_loader.tsx
```diff
@@ -151,133 +85,84 @@ export function ChartLoader(): JSX.Element {
...
-    if (onPlayCallback) {
-      onPlayCallback();
-    }
-  }, [sampleDatesChartData]);
+  // Rendering logic below ...
+  if (!statVar.value.info) {
+    return null;
+  }
+
+  if (!placeInfo.value.enclosingPlace.dcid || !placeInfo.value.enclosingPlace) {
```
chejennifer, line 154: probably want to swap the order of the two checks?

static/js/tools/map/chart_loader.tsx
```diff
@@ -291,102 +176,96 @@ export function ChartLoader(): JSX.Element {
...
-        if (
-          sampleDatesChartData[metaHash][date].enclosedPlaceStat[place].value
-        ) {
-          placeStatData = true;
-          break;
-        }
-      }
-    }
-    statVar.setDate(date);
+  if (!_.isEqual(chartStore.mapValuesDates.context.statVar, statVar.value)) {
```
chejennifer, line 188: maybe we can have a ready hook for mapValuesDates? or for rendering the chart? Just because there seems to be a lot of logic here about when to render a chart and we might want to extract that out. But maybe in a future PR

static/js/tools/map/chart_loader.tsx
```diff
@@ -291,102 +176,96 @@ export function ChartLoader(): JSX.Element {
...
+  if (
+    !_.isEqual(chartStore.mapValuesDates.context.placeInfo, placeInfo.value)
+  ) {
+    console.log("map value placeInfo do not match");
+    return null;
+  }
+
+  const date = getDate(statVar.value.dcid, dateCtx.value);
+
+  let unit = "";
```
chejennifer, line 201: maybe we can make a helper function for getUnit like we have for getDate and getRankingLink?

static/js/tools/map/chart_loader.tsx
```diff
@@ -291,102 +176,96 @@ export function ChartLoader(): JSX.Element {
...
+
+  const rankingLink = getRankingLink(
+    statVar.value,
+    placeInfo.value.selectedPlace.dcid,
+    placeInfo.value.enclosedPlaceType,
+    date,
+    unit
+  );
+
+  let sampleDates: string[];
```
chejennifer, line 216: maybe also a helper function for getSampleDates

static/js/tools/map/ready_hooks.ts
```diff
@@ -186,5 +203,14 @@ export function useAllDatesReady(chartStore: ChartStore) {
...
     chartStore.allDates.context,
+    chartStore.allDates.error,
   ]);
 }
+
+export function useStatVarSummaryReady(chartStore: ChartStore) {
+  const { statVar } = useContext(Context);
+  return useCallback(() => {
+    const c = chartStore.statVarSummary.context;
+    return !_.isEmpty(c) && statVar.value.dcid === c.statVar.dcid;
```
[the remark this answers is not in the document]
shifucun: Done

static/js/tools/map/time_slider.tsx
```diff
@@ -33,119 +41,103 @@ interface TimeSliderProps {
...
-  // Fetches data for slider dates when play is pressed
-  onPlay(callback: () => void): void;
-
-  // Updates map date to slider date
-  updateDate(date: string): void;
 }
 
 export function TimeSlider(props: TimeSliderProps): JSX.Element {
-  const INTERVAL_MS = 500;
+  const INTERVAL_MS = 1000;
```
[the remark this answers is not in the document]
shifucun: done

static/js/tools/map/time_slider.tsx
```diff
@@ -33,119 +41,103 @@ interface TimeSliderProps {
...
   const firstUpdate = useRef(true);
 
-  useEffect(() => {
-    setLoaded(false);
-    setIndex(
-      props.startEnabled ? getIndex(props.dates, props.currentDate) : -1
-    );
-    setCurrentDate(props.startEnabled ? props.currentDate : "--");
-  }, [props.metahash]);
+  const start = props.dates[0];
```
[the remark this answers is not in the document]
shifucun: This is guarded in chart_loader but I agree it's not ideal. However an early return like 
`if (props.dates.length < 2) return null` before hooks definition is not a good practice. I think the structure of this component requires more thoughts

static/js/tools/map/time_slider.tsx
```diff
@@ -33,119 +41,103 @@ interface TimeSliderProps {
...
-          setInterval(() => {
-            setIndex((index) => index + 1);
-          }, INTERVAL_MS)
-        );
-      });
-    } else {
-      clearInterval(timer);
+        setPlay(false);
+      }
+      dateCtx.set(props.dates[index]);
```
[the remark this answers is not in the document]
shifucun: index of props.dates.length will be set to 0 immediately in line 106, so it's okay?

static/js/tools/map/util.ts
```diff
@@ -151,6 +151,11 @@ export const BEST_AVAILABLE_METAHASH = "Best Available";
 export const USA_PLACE_HIERARCHY = ["Country", "State", "County"];
 export const MAP_URL_PATH = "/tools/map";
 
+export interface SampleDates {
```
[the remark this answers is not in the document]
shifucun: Moved to js/shared/types.ts. We badly need to have a folder structure of js/types/<map>,<scatter>...

server/webdriver_tests/tests/map_test.py
```diff
@@ -108,7 +106,7 @@ def test_charts_from_url(self):
         self.assertEqual(chart_title.text, "Median Age (2020)")
         chart_map = self.driver.find_element_by_id('map-items')
         map_regions = chart_map.find_elements_by_tag_name('path')
-        self.assertEqual(len(map_regions), 52)
+        self.assertEqual(len(map_regions), 58)
```
shifucun, line 109: This is mysterious to me. 58 is the right answer but somehow before it only sees 52.

static/js/tools/map/chart_loader.tsx
shifucun: Moved this to the compute hooks now

static/js/tools/map/chart_loader.tsx
shifucun: Added the ready hooks for computed values.

static/js/tools/map/chart_loader.tsx
shifucun: Moved all of these to a renderReady() checks now. In there these checks are not necessary, as it will check if computed values are ready.

static/js/tools/map/util.ts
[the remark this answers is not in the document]
chejennifer: This is good for now, but I agree we should have a folder structure of js/types/..  and/or we could keep component specific types in a types.ts file within that component's folder

static/js/tools/map/chart_store.ts
```diff
@@ -38,6 +38,9 @@ export enum ChartDataType {
   STAT_VAR_SUMMARY = "statVarSummary",
   ALL_DATES = "allDates",
   GEO_RASTER = "geoRaster",
+  BREADCRUMB_VALUES = "breadcrumbValues",
+  MAP_POINT_VALUES = "mapPointValues",
+  MAP_VALUES_DATES = "mapValuesDates",
```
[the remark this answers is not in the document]
chejennifer: that sounds good to me & separate PR makes sense! Maybe add a TODO to revisit the naming

static/js/tools/map/time_slider.tsx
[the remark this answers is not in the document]
chejennifer: ah, I missed that, yes, that makes sense then

static/js/tools/map/time_slider.tsx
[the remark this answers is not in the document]
chejennifer: Maybe add a comment for this for now

server/webdriver_tests/tests/map_test.py
chejennifer: hm, interesting ..

static/js/tools/map/time_slider.tsx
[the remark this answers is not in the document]
shifucun: Done

static/js/tools/map/chart_store.ts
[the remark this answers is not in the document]
shifucun: Added TODO

static/js/tools/map/fetcher/breadcrumb_stat.ts
```diff
@@ -64,10 +64,11 @@ export function useFetchBreadcrumbStat(
       error: null,
     };
 
+    const date = getDate(statVar.value.dcid, dateCtx.value);
     axios
       .get<PointApiResponse>("/api/observations/point", {
         params: {
-          date: getDate(statVar.value.dcid, dateCtx.value),
+          date: getDate(statVar.value.dcid, date),
```
chejennifer, line 71: should just use `date` here?

server/webdriver_tests/tests/map_test.py
shifucun: This is actually a bug. 58 is only for initial page load

static/js/tools/map/ready_hooks.ts
```diff
@@ -211,6 +211,87 @@ export function useStatVarSummaryReady(chartStore: ChartStore) {
...
+  const { dateCtx, statVar, placeInfo } = useContext(Context);
+  return useCallback(
+    (checkDate: boolean) => {
+      const c = chartStore.mapValuesDates.context;
+      let ready =
+        !_.isEmpty(c) &&
+        _.isEqual(statVar.value, c.statVar) &&
+        _.isEqual(placeInfo.value, c.placeInfo);
+      // If only date changes like in time slider, we consider the data as ready
+      // so the <ChartLoader /> componenet will not be unmount, so the time slider
```
chejennifer, line 236: nit: component

static/js/tools/map/compute/map_value_dates.ts
```diff
@@ -81,18 +71,17 @@ export function useComputeMapValueAndDate(
...
-        if (_.isEmpty(wantedFacetData)) {
-          continue;
-        }
+    let wantedFacetData = chartStore.defaultStat.data.data;
+    if (statVar.value.metahash) {
+      wantedFacetData = filterAllFacetData(
+        chartStore.allStat.data.data,
+        statVar.value.metahash
+      );
+      if (_.isEmpty(wantedFacetData)) {
```
chejennifer, line 80: should this be outside the `if (statVar.value.metahash)`?

chejennifer (commented):
thanks for the changes! just few small things

static/js/tools/map/compute/map_value_dates.ts
shifucun: chartStore.allStat.data.data should not be empty at this time with ready check, but good to move outside.

static/js/tools/map/fetcher/breadcrumb_stat.ts
shifucun: yes!

static/js/tools/map/ready_hooks.ts
shifucun: oops, fixed this and elsewhere..

server/webdriver_tests/tests/map_test.py
shifucun: So the map works fine here. But now the chart does not unmount and mount, so all the html elements are not destroyed. The test sees an old value from laster rendering before the new rendering kicks in. 

I have a pending PR to add spinner. With that, can revisit the logic here to additional check spinner not shown.

shifucun (commented):
thanks for the very detailed review!

chejennifer (approved):
Thanks for all the changes!

PR #1801, opened 2022-10-22
Add spinner for chart loader; More fixes

shifucun:


static/js/tools/map/ready_hooks.ts
```diff
@@ -282,15 +282,18 @@ export function useRenderReady(chartStore: ChartStore) {
   const { display, statVar } = useContext(Context);
   const breadcrumbValueReady = useBreadcrumbValuesReady(chartStore);
   const mapValuesDatesReady = useMapValuesDatesReady(chartStore);
+  const geoJsonReady = useGeoJsonReady(chartStore);
   return useCallback(() => {
     return (
       statVar.value.info &&
+      geoJsonReady() &&
```
chejennifer, line 289: I think we only need to wait for geoJson before rendering if the map type is D3

static/js/tools/map/time_slider.tsx
```diff
@@ -102,10 +102,10 @@ export function TimeSlider(props: TimeSliderProps): JSX.Element {
       index += 1;
       if (index === props.dates.length) {
         index = 0;
-        if (firstUpdate.current) {
-          firstUpdate.current = false;
+        if (!firstUpdate.current) {
+          setPlay(false);
```
chejennifer, line 106: curious why we don't `setPlay(false)` if firstUpdate is true? maybe add some comments?

static/js/tools/map/chart_loader.tsx
```diff
@@ -152,99 +153,126 @@ export function ChartLoader(): JSX.Element {
...
-            chartStore.geoJson.data ? chartStore.geoJson.data.features : []
-          }
-          europeanCountries={europeanCountries}
-        />
-      )}
-      <BqModal getSqlQuery={getSqlQuery} showButton={true} />
-    </div>
+    <>
+      {renderContent()}
+      <div id="chart-loader-screen">
```
chejennifer, line 271: should this use the `CHART_LOADER_SCREEN`constant?

static/js/tools/map/ready_hooks.ts
shifucun: I added a hook for deciding the map type, which implicitly checks data availability too.

static/js/tools/map/time_slider.tsx
shifucun: Done

static/js/tools/map/util.ts
```diff
@@ -494,7 +496,7 @@ export function getPlaceChartData(
   if (calculateRatio) {
     const placePopData =
       placeDcid in populationData ? populationData[placeDcid] : null;
-    if (_.isEmpty(placePopData)) {
+    if (_.isEmpty(placePopData.series)) {
```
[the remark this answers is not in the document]
shifucun: Done

static/js/tools/map/ready_hooks.ts
chejennifer: I'm not sure about doing the check with this hook because we want to default to d3 map & update to leaflet map if we get raster data back before all the d3map information. With the hook you added, it will update the mapType back to d3 once the d3map information is completed. 

Can we go back to the original way of updating mapType & just pass mapType as a parameter to this ready function and then check that `(geoJsonReady() || mapType === MAP_TYPE.LEAFLET)`?

chejennifer (approved):
thanks for the changes!

Issue #1807, opened 2022-10-26
[BUG] map spinner doesn't go away in error state


chejennifer:
when choosing an unavailable stat var is chosen for a certain place, need to remove spinner once alert has shown up

e.g., 

1. start at: https://staging.datacommons.org/tools/map#%26sv%3DDifferenceRelativeToBaseDate2006_Max_Temperature_RCP45%26pc%3D0%26denom%3DCount_Person%26pd%3Dcountry%2FUSA%26ept%3DCounty
2. change place type to state
3. close the alert
4. spinner should disappear

chejennifer:
this is fixed

PR #1809, opened 2022-10-27
fix spinner for map error state

chejennifer:
fixes github issue: https://github.com/datacommonsorg/website/issues/1807


shifucun (approved):
thanks for the fix!

PR #1812, opened 2022-11-02
fix /version field change issue

shifucun:


pradh (approved):

server/main.py
```diff
@@ -93,7 +93,7 @@ def version():
                                  website_hash=os.environ.get("WEBSITE_HASH"),
                                  mixer_hash=mixer_version['gitHash'],
                                  tables=mixer_version['tables'],
-                                 bigquery=mixer_version['bigQuery'])
+                                 bigquery=mixer_version['bigquery'])
```
beets, line 96: should we fix this in the mixer? to maintain parity with gitHash which is camel case..

shifucun: Shanth suggested bigquery should be one word..

pradh: I was suggesting call a field bigquery instead of big_query, but sorry, didn't realize it will cause this churn.

PR #1813, opened 2022-11-02
Add back landing page cache improvement

shifucun:
This involves revert of revert of the initial PR with the recently added mixer change that uses the correct menu categories data.

Verified it fixes the issue of missing "Environment" category in this page:

image


pradh (approved):
Thanks for fixing and rolling forward!

server/routes/api/landing_page.py
```diff
@@ -518,20 +556,22 @@ def populate_category_data(category):
...
+                populate_additional_category_data(category)
 
     # Get chart category name translations
-    categories = {}
-    for category in list(spec_and_stat.keys()) + list(spec_and_stat[OVERVIEW]):
-        categories[category] = gettext(f'CHART_TITLE-CHART_CATEGORY-{category}')
+    ordered_category_dict = {}
+    all_categories = set(
+        list(spec_and_stat.keys()) + list(spec_and_stat[OVERVIEW]) +
+        raw_page_data["validCategories"][dcid]['category'])
```
beets, line 565: sorry i haven't been following this change too closely -- but what is `validCategories`?

server/routes/api/landing_page.py
```diff
@@ -518,20 +556,22 @@ def populate_category_data(category):
...
-                    total_charts += len(spec_and_stat[category][topic])
-                    if total_charts > MAX_OVERVIEW_CHART_GROUP:
-                        break
+                populate_additional_category_data(category)
 
     # Get chart category name translations
-    categories = {}
-    for category in list(spec_and_stat.keys()) + list(spec_and_stat[OVERVIEW]):
-        categories[category] = gettext(f'CHART_TITLE-CHART_CATEGORY-{category}')
+    ordered_category_dict = {}
```
beets, line 562: this is a misnomer (one shouldn't rely on ordering in a dict..). just `category_dict` should do, or revert to the former `categories`..

beets (commented):
I have a few questions (although I realize you're just rolling forward an old PR...)

server/routes/api/landing_page.py
beets: i stand corrected -- saw the response in the original pr

server/routes/api/landing_page.py
shifucun: It holds the categories that has data for a place. It's used to determine what categories to show on the LHS menu

server/routes/api/landing_page.py
```diff
@@ -518,20 +556,22 @@ def populate_category_data(category):
...
+    all_categories = set(
+        list(spec_and_stat.keys()) + list(spec_and_stat[OVERVIEW]) +
+        raw_page_data["validCategories"][dcid]['category'])
+
+    ordered_category_dict[OVERVIEW] = gettext(
+        f'CHART_TITLE-CHART_CATEGORY-{OVERVIEW}')
```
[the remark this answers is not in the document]
shifucun: maybe leave as is since it's using the OVERVIEW variable here ? (though it's the same as the actual string).

server/routes/api/landing_page.py
beets: ok sg, thanks!

beets (approved):
thanks for the explanations bo!

pradh:
> /cc @juliawu let's schedule some testing time before the next push

+1

shifucun:
Thank you all for the check!

PR #1816, opened 2022-11-03
Use new property values API to fetch names

shifucun:
Since dc.property_values() supports cache already, there is no need for a wrapper function to cache names. This is to gradually migrate to use V1 API and simplify existing code.


chejennifer (approved):
thanks for the clean up!

PR #1818, opened 2022-11-04
[private dc] Use a config for tool landing page links

beets:
Create json objects that is passed into the template. The component uses this to render the landing page links. This can be customized for the private dc instances.

Using JSON seemed simplest for configuration purposes, and will be more inline with how other templates are overwritten for private dc (I tried using protos which had a lot more config / creation overhead).

If this looks good, will apply this to other tools.


shifucun (approved):
This is nice! 

For naming, we try to use "custom" instead of "private" now to reduce confusion.. might name the folder "custom_dc"

beets:
thanks! sg, there are other templates there already -- i'll do the rename in a subsequent pr.

chejennifer (approved):
this is nice!

PR #1824, opened 2022-11-10
[disaster dashboard] add flask endpoints

chejennifer:
add flask endpoints for getting date range and getting data

This is one part of the changes for the disaster dashboard MVP here on dev:
https://dev.datacommons.org/dev/disaster-dashboard

### DIFF
--- a/server/__init__.py
+++ b/server/__init__.py
@@ -22,6 +22,7 @@ import urllib.error
 
 from flask import Flask, request, g
 from flask_babel import Babel
+from google.cloud import storage
 
 import firebase_admin
 from firebase_admin import credentials
@@ -36,6 +37,7 @@ from opencensus.trace.samplers import AlwaysOnSampler
 import lib.config as libconfig
 import lib.i18n as i18n
 import lib.util as libutil
+from lib.disaster_dashboard import get_disaster_dashboard_config
 import services.ai as ai
 
 propagator = google_cloud_format.GoogleCloudFormatPropagator()
@@ -97,7 +99,7 @@ def register_routes_common(app):
     from routes.api import (browser as browser_api, choropleth, place as
                             place_api, landing_page, ranking as ranking_api,
                             stats, translator, csv, facets, series, point,
-                            observation_dates)
+                            observation_dates, disaster_dashboard)
     app.register_blueprint(browser_api.bp)
     app.register_blueprint(choropleth.bp)
     app.register_blueprint(factcheck.bp)
@@ -111,6 +113,7 @@ def register_routes_common(app):
     app.register_blueprint(series.bp)
     app.register_blueprint(point.bp)
     app.register_blueprint(observation_dates.bp)
+    app.register_blueprint(disaster_dashboard.bp)
 
 
 def create_app():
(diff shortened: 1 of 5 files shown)


server/configmodule.py
```diff
@@ -54,10 +54,12 @@ class StagingConfig(Config):
 class AutopushConfig(Config):
     GA_ACCOUNT = 'G-Y6ZXZ9JK3H'
     ADMIN = True
+    GCS_BUCKET = 'datcom-website-autopush-resources'
     pass
 
 
 class DevConfig(Config):
+    GCS_BUCKET = 'datcom-website-autopush-resources'
```
shifucun, line 62: ideally this should be in dev? I guess you need some cross gcp IAM permission for robot account if doing this.

server/lib/disaster_dashboard.py
```diff
@@ -0,0 +1,76 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Helper functions for getting disaster dashboard data for the app config"""
+
+from google.cloud import storage
+import json
+
+EVENT_TYPE_FILE_NAME = {
```
shifucun, line 19: might just convert this into a function from CamelEvent to camel_events.json?

server/__init__.py
```diff
@@ -214,6 +217,12 @@ def create_app():
     app.config['BABEL_DEFAULT_LOCALE'] = i18n.DEFAULT_LOCALE
     app.config['BABEL_TRANSLATION_DIRECTORIES'] = 'i18n'
 
+    # load disaster dashboard data frpm GCS
```
shifucun, line 220: nit: from.

should it be `get_disaster_dashboard_data` instead of config?

server/routes/api/disaster_dashboard.py
```diff
@@ -0,0 +1,77 @@
...
+
+    Returns: an object with minDate and maxDate
+        {
+            minDate: string,
+            maxDate: string
+        }
+    """
+    event_type = request.args.get('eventType', '')
+    result = {"minDate": "", "maxDate": ""}
+    if event_type in current_app.config['DISASTER_DASHBOARD_DATA']:
```
shifucun, line 42: define a module level var for `current_app.config['DISASTER_DASHBOARD_DATA']`? this is a bit long..

server/configmodule.py
chejennifer: do you mean we would need cross gcp IAM permission for robot account if putting the files in dev or if putting the files in autopush? Because right now, dev seems to work with this code?

server/routes/api/disaster_dashboard.py
chejennifer: done

server/__init__.py
chejennifer: done

server/lib/disaster_dashboard.py
chejennifer: done

shifucun (approved):
thanks for the update!

chejennifer:
thanks for the review!

PR #1827, opened 2022-11-10
Update sv and svg info fetch to use V1 API

shifucun:
This is to use the new V1 REST API, which has better support (ex, custom DC sv node path).

Note here the routes are changed to match that of the REST API, with prefixes "/variable/" and "/variable-group/".


server/routes/api/variable.py
```diff
@@ -0,0 +1,29 @@
...
+import services.datacommons as dc
+
+bp = Blueprint("variable", __name__, url_prefix='/api/variable')
+
+
+@bp.route('/path')
+def get_variable_path():
+    """Gets the path of a stat var to the root of the stat var hierarchy.
+    """
+    dcid = request.args.get("dcid")
```
beets, line 27: nit: call html.escape on the dcid before further processing? it should not affect valid dcid's

beets (commented):
nit: please add more detail to the pr description (especially with respect to the updates in routes).

server/config/topic_page_pb2.py
```diff
@@ -2,10 +2,9 @@
 # Generated by the protocol buffer compiler.  DO NOT EDIT!
```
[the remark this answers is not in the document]
shifucun: This is from running `protoc` with a newer version (3.21.9). I verified locally the topic page works fine.

server/routes/api/variable.py
shifucun: This is directly passed to another url (instead of using as html text), so this should be fine?

shifucun:
> nit: please add more detail to the pr description (especially with respect to the updates in routes).

Added more background

server/routes/api/variable.py
chejennifer: I think the problem is that the dcid is directly returned as part of the response

server/routes/api/variable_group.py
```diff
@@ -0,0 +1,59 @@
...
+    """
+    dcid = request.args.get("dcid")
+    entities = request.args.getlist("entities")
+    url = "/v1/info/variable-group/{}".format(dcid)
+    if entities:
+        url += "?constrained_entities={}".format(entities[0])
+    if len(entities) > 1:
+        parts = list(
+            map(lambda e: "&constrained_entities={}".format(e), entities[1:]))
+        url += "".join(parts)
```
chejennifer, line 45: nit: could replace lines 40 - 45 with: 
```
if entities:
    url += "?constrained_entities=" + "&constrained_entities=".join(entities)
```

server/tests/api/api_stats_test.py
```diff
@@ -236,12 +236,10 @@ def side_effect(query, places, sv_only):
             assert result == expected_sv_only_result
 
 
-class TestGetStatVarGroup(unittest.TestCase):
```
chejennifer, line 239: to keep things consistent, maybe we can extract these tests out into an "api_variable_test.py" and "api_variable_group_test.py" files?

server/routes/api/variable.py
beets: yep that's the issue. and it gets `eval`ed by the browser so is a risk.

shifucun: Updated now

server/routes/api/variable_group.py
shifucun: this is neat!

server/tests/api/api_stats_test.py
shifucun: done

beets (approved):
thanks for the updates!

PR #1833, opened 2022-11-11
[disaster dashboard] address nits

chejennifer:
- address nits from previously merged pr

### DIFF
--- a/static/js/disaster_dashboard/data_fetcher.ts
+++ b/static/js/disaster_dashboard/data_fetcher.ts
@@ -163,7 +163,7 @@ function fetchEventTypeData(
 }
 
 /**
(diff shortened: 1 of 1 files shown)


static/js/disaster_dashboard/data_fetcher.ts
```diff
@@ -165,8 +165,8 @@ function fetchEventTypeData(
 /**
  * Get promise for all event points for a specific disaster type, place, and date
```
beets, line 166:
```suggestion
 * Get promise for all event points for a specific disaster type, place, and date. Only dates of type YYYY or YYYY-MM are supported.
```

beets (approved):

PR #1834, opened 2022-11-14
feat: add disease prevalence rates chart

suhana13:
In this PR, I'm re-using the `maptile` component from the graph explorer to plot the disease prevalence rates for the disease browser. Here is a screenshot for one of the diseases.

Screen Shot 2022-11-14 at 10 46 14 AM


static/js/biomedical/disease/page.tsx
```diff
@@ -44,6 +46,13 @@ export interface PagePropType {
 export interface PageStateType {
   data: GraphNodes;
 }
+
+export const USA: NamedTypedPlace = {
+  dcid: "country/USA",
+  name: "United States of America",
+  types: ["Country"],
+};
```
beets, line 54: please add this to constants.ts

static/js/topic_page/map_tile.tsx
```diff
@@ -49,6 +49,7 @@ import { getDateRange } from "../utils/string_utils";
 import { ChartTileContainer } from "./chart_tile";
 import { CHART_HEIGHT } from "./constants";
 import { ReplacementStrings } from "./string_utils";
+export const MAP_ID = "map";
```
beets, line 52: The topic page is designed for multiple map tiles to load, which is why the element ID is passed in as a prop, and not hardcoded to a single ID (HTML element IDs must be unique within a page).

Please pass this value in from the disease page component.

static/js/topic_page/map_tile.tsx
```diff
@@ -274,8 +283,9 @@ function draw(
     }
     return place.name + ": " + value;
   };
+  document.getElementById(MAP_ID).innerHTML = "";
```
beets, line 286: This should not be required. Maybe this was added after the map Id's were hardcoded?

static/js/topic_page/map_tile.tsx
```diff
@@ -274,8 +283,9 @@ function draw(
...
   drawD3Map(
-    props.id,
+    MAP_ID,
```
beets, line 288: Please revert.

static/js/biomedical/disease/page.tsx
```diff
@@ -87,7 +97,15 @@ export class Page extends React.Component<PagePropType, PageStateType> {
       { id: "name", name: "Compound Name" },
       { id: "drugSource", name: "Drug Source" },
     ];
-
+    const diseasePrevalenceStatVarValue =
+      "Count_MedicalConditionIncident_Condition" + this.props.dcid;
+    const statVarDisease = {} as StatVarSpec;
+    statVarDisease.statVar = diseasePrevalenceStatVarValue;
+    statVarDisease.denom = "";
```
beets, line 104: Since we are showing "prevalence", should this be Count_Person? That might also negate the changes in map_tile when there is no denominator defined.

static/js/topic_page/map_tile.tsx
```diff
@@ -176,15 +177,23 @@ function fetchData(
...
+      if (!_.isEmpty(populationData.facets)) {
+        Object.assign(metadataMap, populationData.facets);
+      }
+      let population = {};
+      if (
+        !_.isEmpty(populationData.data) &&
+        denomStatVar in populationData.data
+      ) {
+        population = populationData.data[denomStatVar];
+      }
```
beets, line 191: perhaps, this should check if denomStatVar is defined in the first place. checking for `"" in object` works but is error prone.

also see comment above about always using a denominator for this..

beets (commented):
As you make updates to map tile, please ensure that the topic pages still work:
/topic/equity/country/USA

At a high level, also wondering if we need a denominator for these maps since we are calling them prevalence charts.

Thanks!

static/js/biomedical/disease/page.tsx
```diff
@@ -87,7 +92,15 @@ export class Page extends React.Component<PagePropType, PageStateType> {
       { id: "name", name: "Compound Name" },
       { id: "drugSource", name: "Drug Source" },
     ];
-
+    const diseasePrevalenceStatVarValue =
+      "Count_MedicalConditionIncident_Condition" + this.props.dcid;
+    const statVarDisease = {} as StatVarSpec;
```
chejennifer, line 97: instead of defining an empty object and then setting the fields, you could just define the object with all the values like this:

```
const statVarSpec = {
statVar: diseasePrevalenceStatVarValue,
denom: "",
...
}
```

static/js/biomedical/disease/page.tsx
```diff
@@ -87,7 +92,15 @@ export class Page extends React.Component<PagePropType, PageStateType> {
       { id: "name", name: "Compound Name" },
       { id: "drugSource", name: "Drug Source" },
     ];
-
+    const diseasePrevalenceStatVarValue =
```
chejennifer, line 95: nit: since this is the stat var dcid, maybe `diseasePrevalenceStatVarDcid` instead

static/js/shared/constants.ts
```diff
@@ -37,6 +37,11 @@ export const EARTH_NAMED_TYPED_PLACE: NamedTypedPlace = {
   name: "Earth",
   types: ["Planet"],
 };
+export const USA: NamedTypedPlace = {
```
chejennifer, line 40: for consistency with the other NamedTypedPlace constants, let's name this "USA_NAMED_TYPED_PLACE"

static/js/biomedical/disease/page.tsx
chejennifer: +1 - showing just the count of a disease might not be super meaningful because that count can be impacted by the population of a certain area

chejennifer (commented):
nice job getting the map to work!

static/js/topic_page/map_tile.tsx
```diff
@@ -283,6 +283,7 @@ function draw(
     }
     return place.name + ": " + value;
   };
+  document.getElementById(props.id).innerHTML = "";
```
beets, line 286: nit: this shouldn't be required anymore -- please remove.

beets (approved):
one small nit left and feel free to submit after fixing it

static/js/biomedical/disease/page.tsx
suhana13: Resolved!

static/js/topic_page/map_tile.tsx
suhana13: Done!

static/js/topic_page/map_tile.tsx
suhana13: Reverted!

static/js/biomedical/disease/page.tsx
suhana13: Displaying percentages instead of counts! Thanks for the great suggestion!

static/js/topic_page/map_tile.tsx
suhana13: Done!

static/js/biomedical/disease/page.tsx
suhana13: Done!

static/js/biomedical/disease/page.tsx
suhana13: Done!

static/js/shared/constants.ts
suhana13: Changed the name to USA_NAMED_TYPED_PLACE, for consistency!

static/js/topic_page/map_tile.tsx
suhana13: @beets , the reason I added this was because before it, the map was getting duplicated multiple times (every time the state of the data got updated) but with the addition of this line, it's only being plotted once:)

static/js/topic_page/map_tile.tsx
```diff
@@ -275,14 +279,22 @@ function draw(
   const getTooltipHtml = (place: NamedPlace) => {
     let value = "Data Unavailable";
     if (place.dcid in chartData.dataValues) {
-      value = formatNumber(
-        Math.round((chartData.dataValues[place.dcid] + Number.EPSILON) * 100) /
-          100,
-        props.statVarSpec.unit
-      );
+      // shows upto 4 decimal points for very low values
```
chejennifer, line 282: maybe we can use [toPrecision](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toPrecision) & keep 2 significant digits for values whose absolute value is between 0-1.

static/js/topic_page/map_tile.tsx
beets: I see - let's do this within the disease page then (not in map_tile). It looks like we'll have to look at how the page is being redrawn with data refreshes.

chejennifer (commented):
thanks for the changes!

static/js/topic_page/map_tile.tsx
suhana13: Done!

static/js/topic_page/map_tile.tsx
suhana13: Done!

static/js/topic_page/map_tile.tsx
```diff
@@ -275,11 +279,18 @@ function draw(
...
+      if (chartData.dataValues[place.dcid] < 0.01) {
```
chejennifer, line 283: let's do this for all values where the absolute value (in case there are negatives) is between 0 and 1

static/js/topic_page/map_tile.tsx
```diff
@@ -275,11 +279,18 @@ function draw(
...
+        const val = Number(chartData.dataValues[place.dcid]);
```
chejennifer, line 284: nit: don't need the Number() since dataValues values are already numbers
nit: let's be more specific in the naming about what value this variable is referring to because we already have another variable named value.

chejennifer (commented):
thanks for the changes

static/js/topic_page/map_tile.tsx
suhana13: That's a great catch! Changed it to absolute value.

static/js/topic_page/map_tile.tsx
suhana13: Done! Though I still had to do `Number(chartDatavalue.toPrecision(2))` in line 286, as `.Precision()` returned a string and we want a number!

static/js/topic_page/map_tile.tsx
chejennifer: please also change the check so that this case is only for numbers between 0 and 1

chejennifer (approved):
Thanks for all the changes, please fix this nit before merging

PR #1835, opened 2022-11-14
[custom_dc] Rename config / folders to custom_dc from private_dc

beets:
Standardizing our content / files / configs around custom_dc over private_dc

Please take an especially close look at the deployment files.


shifucun (approved):
Thanks for the clean up. The deployment change LGTM.

PR #1836, opened 2022-11-15
Fix a few key/value error issues

shifucun:
* These errors are found from Error Reporting in GCP. We should set up alerts for such errors.
* All Flask endpoint should do "try/catch" to handle any errors from mixer response. This PR updates one such case from error reporting logs.
* Changed "dcid" to "dcids" in name fetch endpoint for better readability and consistency.

### DIFF
--- a/server/routes/api/place.py
+++ b/server/routes/api/place.py
@@ -14,6 +14,7 @@
 
 import collections
 import json
+import logging
 import urllib.parse
 
 from flask import Blueprint, request, Response, url_for, g, current_app
@@ -132,6 +133,7 @@ def api_name():
     try:
         return names(dcids)
     except Exception as e:
+        logging.error(e)
         return 'error fetching names for {}'.format(dcids), 400
 
 
--- a/server/routes/api/place.py
+++ b/server/routes/api/place.py
@@ -128,9 +128,11 @@ def get_place_type(place_dcid):
 @bp.route('/name')
 def api_name():
     """Get place names."""
-    dcids = request.args.getlist('dcid')
-    result = names(dcids)
-    return Response(json.dumps(result), 200, mimetype='application/json')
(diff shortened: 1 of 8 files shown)


server/routes/api/place.py
```diff
@@ -128,9 +128,11 @@
...
 def api_name():
     """Get place names."""
-    dcids = request.args.getlist('dcid')
-    result = names(dcids)
-    return Response(json.dumps(result), 200, mimetype='application/json')
+    dcids = request.args.getlist('dcids')
+    try:
+        return names(dcids)
+    except Exception as e:
+        return '{}'.format(e), 400
```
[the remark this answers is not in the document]
beets: Please fix. We could log the error and return a simplified error message.

static/js/utils/place_utils.ts
```diff
@@ -165,15 +167,16 @@ export function getPlaceNames(
...
+  return axios
+    .get("/api/place/name", {
+      params: {
+        dcids: dcids,
+      },
+      paramsSerializer: stringifyFn,
+    })
+    .then((resp) => {
+      return resp.data;
+    });
```
beets, line 179: Should we add an error handler here, that returns just the dcid's? Could be a TODO as well.

beets (commented):
thanks for these fixes! +1 on alerts

server/routes/api/place.py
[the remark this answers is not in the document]

static/js/utils/place_utils.ts
shifucun: Yes, we should. I will file a bug to fix all the axios errors handling.

beets: sounds good - thanks!

beets (approved):

PR #1837, opened 2022-11-15
Download example

beets:
The customization for this is slightly different than the other tools, since only 2 place examples are required.
For now, do this customization in the python code.

No visible diff to the download page.

### DIFF
--- a/static/js/tools/download/info.tsx
+++ b/static/js/tools/download/info.tsx
@@ -20,7 +20,20 @@
 
 import React from "react";
 
-export function Info(): JSX.Element {
+// Data object used to customize the example rendered by this component.
+// A single example place to display.
+export interface InfoPlace {
+  name: string;
+  dcid: string;
+}
+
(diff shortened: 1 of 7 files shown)


static/js/tools/download/info.tsx
```diff
@@ -20,7 +20,20 @@
...
+// A single example place to display.
+export interface InfoPlace {
+  name: string;
+  dcid: string;
+}
+
+// Data object used to customize the example rendered by this component.
+interface InfoPropType {
+  // NOTE: Exactly 2 InfoPlace objects are expected.
+  infoPlaces: InfoPlace[];
```
shifucun, line 33: Wonder can do [InfoPlace, InfoPlace] here to indicate 2 elements?

shifucun (approved):

static/js/tools/download/info.tsx
beets: Ah, that works! Thanks for the suggestion!

PR #1838, opened 2022-11-15
[custom_dc] fold in IITM updates

beets:
Ports most functional changes from #1529, #1758 and [iitm branch](https://github.com/datacommonsorg/website/tree/iitm) (iitm branch is now behind this PR).

Includes:
- All html templates updated to use custom base template
- Data download tool added to header
- Primary button color overwritten
- Customized landing pages for place, browser, map, scatter, timeline, download tools
- Loads JS libraries from CDN's where possible (except particle.js)
- Load BQ SQL button

There are some minor differences from datacommons.iitm.ac.in
- Fonts don't look as crisp anymore (especially bold text on the tool info pages). Not sure if this is because of font-awesome
- Slight spacing updates

Before (left, prod), after (right, localhost):

image
image
image
image
image
image
image
image
image
image
image
image
image
image
image
image


SenthamizhanV (approved):
Pulled this branch and deployed the website locally. Everything works fine. I could not view the /place page since I do not have the DC API key. All other pages look good.

The difference in fonts can be ignored. One place I noticed there is a difference in the text below 'India Data Commons' in the homepage banner. The font-weight is different - this PR has value of 600 but the prod version has 500. But we can leave it at that.

server/config/topic_page_pb2.py
```diff
@@ -2,9 +2,10 @@
 # Generated by the protocol buffer compiler.  DO NOT EDIT!
```
shifucun, line 2: nit: can you check the protoc version with `protoc --version`. The current file is generated using 3.21.9 and I suspect this change is from using an older version.

server/configmodule.py
```diff
@@ -87,9 +87,10 @@ class TidalConfig(CustomConfig):
 
 
 class IitmConfig(Config):
-    IITM = True
     NAME = "IITM"
```
shifucun, line 90: nice!

server/routes/static.py
```diff
@@ -42,13 +40,20 @@ def about():
     env_name = current_app.config.get('ENV_NAME', None)
```
shifucun, line 40: move `env_name = current_app.config.get('ENV_NAME', None)` to the top of this module and only declare once?

server/routes/tools.py
```diff
@@ -76,7 +104,10 @@ def stat_var():
 
 @bp.route('/download')
 def download():
+    env_name = current_app.config.get('ENV_NAME', None)
```
shifucun, line 107: same here, moved to the top of the module

server/config/topic_page_pb2.py
beets: Thanks for pointing that out -- reverted for now.

server/routes/static.py
beets: It requires an application context to load the value, so it won't work at the top of the module. Instead, I've added it to the request global. ptal and lmk what you think

shifucun (approved):

shifucun:
Glad to see this is checked in!

server/routes/static.py
shifucun: This is concise!

beets:
> Pulled this branch and deployed the website locally. Everything works fine. I could not view the /place page since I do not have the DC API key. All other pages look good.
> 
> The difference in fonts can be ignored. One place I noticed there is a difference in the text below 'India Data Commons' in the homepage banner. The font-weight is different - this PR has value of 600 but the prod version has 500. But we can leave it at that.

I just noticed that requesting 'Public Sans' was dropped from the header, so added it back and the text looks much better now. I've also updated the font weight for just the homepage.

beets:
Thanks for the reviews!

@SenthamizhanV If there are other diffs you notice, we can now continue fixing this on the main branch :)

PR #1842, opened 2022-11-16
Fix a timeline x axis domain issue

shifucun:
The X domain should be computed from all the time points, instead of sampled series.

After fix:
image

Before fix:
image


beets (approved):
thanks for the fix!!

PR #1848, opened 2022-11-28
[Disaster Pages] Initial Page Setup

juliawu:
Sets up skeleton for dedicated event pages which will eventually contain information, maps, news, etc about each natural disaster event in the disaster dashboard. Current PR is for a blank page. Future PRs will fill in with info.


server/routes/dev.py
```diff
@@ -49,3 +49,9 @@ def disaster_dashboard():
       dc.get_places_in(["europe"], "Country").get("europe", []))
   return flask.render_template('dev/disaster_dashboard.html',
                                european_countries=european_countries)
+
+@bp.route('/event-page')
```
beets, line 53: nit: "event" rather than "event-page" since they're all pages. drop the "-page" for the route and filenames too

static/js/event_page/event_page.ts
```diff
@@ -0,0 +1,31 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * disaster dashboard
```
beets, line 18: Nit: Disaster events.

static/js/event_page/types.ts
```diff
@@ -0,0 +1,30 @@
...
+ */
+
+import { DisasterType } from "../disaster_dashboard/constants";
+import { NamedTypedPlace } from "../shared/types";
+
+/**
+ * Types specific to Event Page
+ */
+
+export interface EventInfo {
```
beets, line 24: in general, please add documentation for shared interfaces. 

can we add this type as it's needed?

server/routes/dev.py
juliawu: Renamed all `event-page` instances to `event`

static/js/event_page/event_page.ts
juliawu: Fixed!

static/js/event_page/types.ts
juliawu: I meant it as a placeholder, removed the interface. Will add types as necessary in future PRs.

beets (approved):
hooray!

PR #1851, opened 2022-11-29
Be resilient to empty place dcids

### LINKED ISSUE #1849: [BUG] Timeline tool generates invalid API request
Timeline tool generated this request:
https://datacommons.org/api/place/name?dcids=&dcids=geoId%2F06&dcids=geoId%2F06085

Currently results in a failure:
"error fetching names for the given places"

We should
a) handle this on the server
b) fix the request point

beets:
Fixes #1849 

Could not find the root cause of the issue, but a user could also easily play with the URL and we would still need handle it in the API handler.

### DIFF
--- a/server/routes/api/place.py
+++ b/server/routes/api/place.py
@@ -130,6 +130,7 @@ def get_place_type(place_dcid):
 def api_name():
   """Get place names."""
   dcids = request.args.getlist('dcids')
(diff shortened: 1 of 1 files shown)


juliawu (approved):
Great, thanks for the fix!

PR #1852, opened 2022-12-01
[Event Pages] Add properties to the event pages

juliawu:
This PR adds code to put up property values on event pages (/dev/event/<dcid>).

Formatting and beautifying will come in later PRs.

Sample page screenshot:
Screen Shot 2022-12-01 at 11 11 39 AM


static/js/event/page.tsx
```diff
@@ -20,6 +20,43 @@
 
 import React from "react";
 
+import { Property } from "./types";
+
```
beets, line 24: please add top-level comment per component.

server/routes/dev.py
```diff
@@ -56,3 +57,17 @@ def event():
...
   return flask.render_template('dev/event.html')
+
+
+@bp.route('/event/<path:dcid>')
+def event_node(dcid):
+  if not os.environ.get('FLASK_ENV') in ['autopush', 'local', 'dev']:
+    flask.abort(404)
+  node_name = dcid
+  try:
+    api_name = shared_api.names([dcid]).get(dcid)
```
beets, line 68: please check return value of shared_api.names before calling .get

server/routes/dev.py
```diff
@@ -56,3 +57,17 @@ def event():
   if not os.environ.get('FLASK_ENV') in ['autopush', 'local', 'dev']:
     flask.abort(404)
   return flask.render_template('dev/event.html')
+
+
+@bp.route('/event/<path:dcid>')
+def event_node(dcid):
+  if not os.environ.get('FLASK_ENV') in ['autopush', 'local', 'dev']:
+    flask.abort(404)
+  node_name = dcid
```
chejennifer, line 66: could probably replace lines 66-72 with just
```
node_name = shared_api.names([dcid]).get(dcid, dcid)
```

static/js/event/event.ts
```diff
@@ -18,14 +18,49 @@
...
 
+import axios from "axios";
 import React from "react";
 import ReactDOM from "react-dom";
 
-import { Page } from "./page";
+import { EventPage, Page } from "./page";
+import { Property } from "./types";
+
+function parseApiResponse(resp): Array<Property> {
```
chejennifer, line 28: when defining functions, we want to include types for the arguments. So we want `function parseApiResponse(resp: <Type for resp>) ...`

nit: parseTriplesApiResponse because you might end up using more apis and we want to know which api response this function is parsing.

static/js/event/event.ts
```diff
@@ -18,14 +18,49 @@
...
+  // Parses response from triples API for later rendering
```
chejennifer, line 29: when adding comments for what a function does, we want to add function comments above the function like: 

```
/**
 * Parses response from ...
 */
function parseApiResponse ...

File: static/js/event/event.ts [hunk 6]
```diff
@@ -18,14 +18,49 @@
...
+        React.createElement(EventPage, {
+          dcid: dcid,
+          name: nodeName,
+          properties: parsedResponse,
+        }),
+        document.getElementById("node")
+      );
+    })
+    .catch(() => {
+      ReactDOM.render(
```
chejennifer, line 61: I don't think we should have a separate component for when there is an error retrieving triples data. We can just render the EventPage component but pass in an empty array for the properties prop. To inform users that there was a problem, could add an alert like `window.alert("Sorry, there was a problem retrieving data about this disaster")`

static/js/event/page.tsx
```diff
@@ -20,6 +20,43 @@
...
 export function Page(): JSX.Element {
```
chejennifer, line 25: as mentioned in previous comment, we can remove this Page component

static/js/event/event.ts
```diff
@@ -18,14 +18,49 @@
...
-import { Page } from "./page";
+import { EventPage, Page } from "./page";
+import { Property } from "./types";
+
+function parseApiResponse(resp): Array<Property> {
+  // Parses response from triples API for later rendering
+  const parsedResponse = [];
+  for (const [property, connectedNodes] of Object.entries(resp)) {
+    for (const nodes of Object.values(connectedNodes)) {
+      parsedResponse.push({ name: property, values: nodes });
```
chejennifer, line 33: I'm curious what the intended behavior here is. For parsedResponse, did you want:
a. one list item per property
b. one list item per list of nodes in connectedNodes

I think currently you are doing (b).

static/js/event/event.ts
```diff
@@ -18,14 +18,49 @@
...
-    React.createElement(Page),
-    document.getElementById("main-pane")
-  );
+  const dcid = document.getElementById("node").dataset.dcid;
+  const nodeName = document.getElementById("node").dataset.nn;
+  const propertiesPromise = axios
+    .get(`/api/node/triples/out/${dcid}`)
+    .then((resp) => resp.data)
+    .catch(() => {
+      return 0;
```
chejennifer, line 46: since the error case is already being handled in line 60, I don't think we need this catch block.

Sidenote, we probably wouldn't have wanted to return 0. In the`.catch`, we would want to return something that is of the same type as what is returned in the `.then`

static/js/event/page.tsx
```diff
@@ -20,6 +20,43 @@
...
+export function EventPage(props: EventPagePropsType): JSX.Element {
+  return (
+    <div>
+      <h1>{props.name}</h1>
+      <h3>dcid: {props.dcid}</h3>
+      <br />
+      <h3>Properties:</h3>
+      {props.properties.map((property) => {
+        return (
+          <div key={property.name}>
```
chejennifer, line 44: nit: name works, but I would prefer to use dcid in the key because dcid is guaranteed to be unique

chejennifer (commented):
nice job figuring this stuff out!

static/js/event/event.ts
juliawu: Done!

static/js/event/event.ts
juliawu: Intended behavior is (b). Updated the logic to make it more clear what's happening.

static/js/event/event.ts
juliawu: Thanks for the tip! The triples API call is now being done in flask, so this section has been removed.

static/js/event/event.ts
juliawu: removed separate component

static/js/event/page.tsx
juliawu: Removed!

static/js/event/page.tsx
juliawu: property.name actually referred to dcid. Updated the naming to make this clear.

juliawu (commented):
Thanks for the thorough reviews! I've responded with updates.

static/js/event/page.tsx
```diff
@@ -20,6 +20,45 @@
...
+      <br />
+      <h3>Properties:</h3>
+      {props.properties.map((property) => {
+        return (
+          <div key={property.dcid}>
+            <b>{property.dcid}:</b>
+            <ul>
+              {property.values.map((value) => {
+                return (
+                  <li key={JSON.stringify(value)}>
```
chejennifer, line 51: nit: for the key, maybe we can use: {\`${property.dcid}-${value.dcid}\`} so that uniqueness is guaranteed?

static/tsconfig.json
```diff
@@ -5,7 +5,7 @@
         "name": "typescript-tslint-plugin",
       }
     ],
-    "lib": ["es2019"],
+    "lib": ["es2021"],
```
chejennifer, line 8: curious what the reason for this change is?

chejennifer (commented):
thanks for the update

static/tsconfig.json
juliawu: `str.replaceAll()` requires upgrade to es2021. I use it in `parseTriplesApiResponse()` to replace single quotes with double quotes. 

If we don't like this change, another option is to do the replaceAll in the server, but splitting the post-processing between the server and client is less clean to me.

static/js/event/page.tsx
juliawu: Done!

static/js/event/event.ts
```diff
@@ -21,11 +21,35 @@
...
 
-import { Page } from "./page";
+import { EventPage } from "./page";
+import { Property } from "./types";
+
+/**
+ * Parses response from triples API for later rendering.
+ */
+function parseTriplesApiResponse(resp: string): Array<Property> {
+  const respJson = JSON.parse(resp.replaceAll("'", '"')); //JSON.parse requires double quotes
```
beets, line 31: let's do this in flask rather than react (use our compute, vs the clients)

static/js/event/types.ts
```diff
@@ -17,3 +17,21 @@
 /**
  * Types specific to event pages
  */
+
+export interface PropertyValue {
```
beets, line 21: nit: this is a good candidate for a shared interface (since all callers of the triples api would need this). for future reference

static/tsconfig.json
beets: upgrading lgtm, though see my earlier comment about moving the processing..

server/routes/dev.py
```diff
@@ -56,3 +58,22 @@ def event():
   if not os.environ.get('FLASK_ENV') in ['autopush', 'local', 'dev']:
     flask.abort(404)
   return flask.render_template('dev/event.html')
+
+
+@bp.route('/event/<path:dcid>')
+def event_node(dcid):
+  if not os.environ.get('FLASK_ENV') in ['autopush', 'local', 'dev']:
```
beets, line 65: add stanford to this list too.

juliawu: Added!

static/js/event/types.ts
juliawu: Moved to shared interface!

server/routes/dev.py
```diff
@@ -43,6 +45,49 @@ def screenshot(folder):
 
 @bp.route('/event')
 def event():
-  if not os.environ.get('FLASK_ENV') in ['autopush', 'local', 'dev']:
+  if not os.environ.get('FLASK_ENV') in [
```
chejennifer, line 48: maybe we should also move event pages out of dev to stanford dc .. what do you think @beets?

Maybe for the disasters routes we can have:
`/disasters/dashboard/v0`
`/disasters/dashboard/<placeDcid>`
`/disasters/event/<eventDcid>`

This can be done in a future PR.

static/js/event/event.ts
```diff
@@ -21,11 +21,18 @@
 import React from "react";
 import ReactDOM from "react-dom";
 
-import { Page } from "./page";
+import { EventPage } from "./page";
 
 window.onload = () => {
```
chejennifer, line 26: When creating **new** pages, types, etc, please follow the file structure outlined [here](https://docs.google.com/document/d/1eYvx8sYVNxJ5tWQ229gHb3lPCdepwBt5Z2TYKRdvSNQ/edit?resourcekey=0-_Sz3NjUjbJ621ZtVB9PD6Q#heading=h.95wucwdioq2) (trying to gradually update the whole repo to this structure). An example that uses this structure is the new [disaster dashboard](https://github.com/datacommonsorg/website/tree/master/static/js/apps/disaster_dashboard).

server/routes/dev.py
```diff
@@ -43,6 +45,49 @@ def screenshot(folder):
...
+      'autopush', 'local', 'dev', 'stanford', 'local-stanford',
+      'stanford-staging'
+  ]:
     flask.abort(404)
   return flask.render_template('dev/event.html')
+
+
+def parseTriplesResponse(response):
```
chejennifer, line 56: nit: please use snake_case when defining python functions

server/routes/dev.py
```diff
@@ -43,6 +45,49 @@ def screenshot(folder):
...
+
+
+@bp.route('/event/<path:dcid>')
+def event_node(dcid):
+  if not os.environ.get('FLASK_ENV') in [
+      'autopush', 'local', 'dev', 'stanford', 'local-stanford',
+      'stanford-staging'
+  ]:
+    flask.abort(404)
+  node_name = dcid
```
chejennifer, line 80: escape dcid before using it as the default value for node_name

server/routes/dev.py
```diff
@@ -43,6 +45,49 @@ def screenshot(folder):
...
+  properties = "{}"
+  try:
+    name_results = shared_api.names([dcid])
+    if dcid in name_results.keys():
+      node_name = name_results.get(dcid)
+    properties = node_api.triples('out', dcid)
```
chejennifer, line 86: nit: let's package the api call with the parsing so we just have one function like get_properties(dcid) that will return properties that are properly formatted.

chejennifer (commented):
thanks for the updates!

server/routes/dev.py
beets: since we are adding these now, let's move it here too.

server/routes/dev.py
juliawu: Done

server/routes/dev.py
juliawu: Done

server/routes/dev.py
juliawu: Done

static/js/event/event.ts
juliawu: Moved to follow the new structure

server/routes/dev.py
juliawu: Moved

server/routes/disasters.py
```diff
@@ -39,6 +43,10 @@ def disaster_dashboard_v0():
 @bp.route('/')
 @bp.route('/<path:place_dcid>', strict_slashes=False)
 def disaster_dashboard(place_dcid=DEFAULT_PLACE_DCID):
+  if place_dcid == "event":
+    # Access '/event' route instead
+    event_node()
+    return
```
beets, line 49: Sorry for the miscommunication - my preference would be for this to be in a separate route (/event/event_id rather than /disasters/event/event_id). Then we can avoid this, and the URL space is cleaner.

static/tsconfig.json
beets: this isn't required anymore, since we moved to server-side string replace.

beets (commented):
thanks for the updates! mostly please move out of /disasters into a new event route.

server/routes/disasters.py
juliawu: Thanks for the clarification - done!

server/__init__.py
```diff
@@ -87,10 +87,11 @@ def register_routes_custom_dc(app):
 
 def register_routes_stanford_dc(app, is_test):
   # Install blueprints specific to Stanford DC
-  from routes import (disasters)
+  from routes import (disasters, event)
```
beets, line 90: super nit: always alphabetize these lists.

beets (approved):
thanks!

PR #1854, opened 2022-12-02
[protein] fix disease gene association chart bug

chejennifer:
- fix bug in the disease gene association chart caused by null values in the disease name which was caused by a change in the property name for the name value

Screen Shot 2022-12-02 at 11 20 51 AM


static/js/biomedical/protein/data_processing_utils.ts
```diff
@@ -236,14 +236,14 @@ export function getDiseaseGeneAssoc(
                 }
                 diseaseID = n3.value;
                 for (const n4 of n3.neighbors) {
-                  if (n4.property !== "commonName") {
+                  if (n4.property !== "name") {
```
shifucun, line 239: nit for later: put all these constants terms in a separate files, since they get changed often.

shifucun (approved):

suhana13 (approved):
Thank you very much! I actually had a branch locally to fix this but I'm glad you made the PR. We're also missing data in the protein interaction chart. Not sure if that's something we want to resolve in this PR.

static/js/biomedical/protein/data_processing_utils.ts
suhana13: That's a great idea! Will implement in the upcoming PR!

beets:
thanks for the fix!

PR #1870, opened 2022-12-07
[Stanford] Move disasters dashboard to Stanford DC

juliawu:


server/__init__.py
```diff
@@ -134,6 +140,7 @@ def register_routes_common(app):
   app.register_blueprint(browser_api.bp)
   app.register_blueprint(choropleth.bp)
   app.register_blueprint(csv.bp)
+  # app.register_blueprint(disasters.bp)
```
chejennifer, line 143: remove this?

server/__init__.py
```diff
@@ -248,7 +257,9 @@ def create_app():
   app.config['BABEL_TRANSLATION_DIRECTORIES'] = 'i18n'
 
   # load disaster dashboard data from GCS
-  if os.environ.get('FLASK_ENV') in ['autopush', 'local', 'dev']:
+  if os.environ.get('FLASK_ENV') in [
+      'autopush', 'local', 'dev', 'local-stanford'
```
chejennifer, line 261: should we also be including 'stanford'?

server/routes/api/disasters.py
```diff
@@ -0,0 +1,37 @@
...
+# Define blueprint
+bp = Blueprint("disasters", __name__, url_prefix='/disasters')
+
+EARTH_DCID = "Earth"
+
+
+@bp.route('/v0')
+def disaster_dashboard():
+  if not os.environ.get('FLASK_ENV') in [
+      'autopush', 'local', 'dev', 'local-stanford'
```
chejennifer, line 31: should we include 'stanford'?

server/routes/api/disasters.py
```diff
@@ -0,0 +1,37 @@
...
+from flask import Blueprint
+import services.datacommons as dc
+import json
+import os
+import flask
+
+# Define blueprint
+bp = Blueprint("disasters", __name__, url_prefix='/disasters')
+
+EARTH_DCID = "Earth"
```
chejennifer, line 25: curious what this constant is being used for?

chejennifer (commented):
thanks for moving the dashboard!

server/routes/api/disasters.py
juliawu: This file is a copy of the original disasters.py in dev. The constant was in the original file, so I didn't touch it. Let me check to see if it can be removed.

server/__init__.py
juliawu: Based on https://github.com/datacommonsorg/website/blob/b943707aeb967bba4a2377e54e736caca36fe25f/run_server.sh#L55, there is no 'stanford' FLASK_ENV.

server/__init__.py
juliawu: Thanks for the catch! Done!

server/routes/api/disasters.py
chejennifer: ah, ok, so what you were looking at was the api file & what you're creating is the route file, so some things that need to be done:

- move this file to server/routes
- remove the copied code from the dev file
- nice to have: rename the bp in disaster_dashboard.py to "disaster-api" and update the naming where the bp is registered in the __init__.py

juliawu: Update: removed! Thanks for the catch!

juliawu: Further update: Made the changes!

* disasters.py is now in server/routes
* no more unused constant
* Renamed disaster_dashboard bp to `disaster_api`
* Updated the registry in __init__

server/__init__.py
chejennifer: I believe this is the full list of flask environments: https://github.com/datacommonsorg/website/blob/master/server/lib/config.py#L29

server/routes/disasters.py
```diff
@@ -0,0 +1,35 @@
...
+import services.datacommons as dc
+import json
+import os
+import flask
+
+# Define blueprint
+bp = Blueprint("disasters", __name__, url_prefix='/disasters')
+
+
+@bp.route('/v0')
```
chejennifer, line 26: since we've moved the route here, should we delete it from dev? https://github.com/datacommonsorg/website/blob/master/server/routes/dev.py#L44

chejennifer (commented):
thanks for the updates!

server/routes/disasters.py
juliawu: Oops, missed this in your original comment. Deleted from dev!

server/__init__.py
juliawu: Thanks for linking! Added stanford

server/__init__.py
```diff
@@ -88,6 +88,12 @@ def register_routes_custom_dc(app):
   pass
 
 
+def register_routes_stanford_dc(app):
+  # Apply blueprints specific to Stanford DC
```
beets, line 92:
```suggestion
  # Install blueprints specific to Stanford DC
```

server/__init__.py
```diff
@@ -134,7 +140,7 @@ def register_routes_common(app):
   app.register_blueprint(browser_api.bp)
   app.register_blueprint(choropleth.bp)
   app.register_blueprint(csv.bp)
-  app.register_blueprint(disaster_dashboard.bp)
+  app.register_blueprint(disaster_api.bp)
```
beets, line 143: This should move to the "stanford" instances too (and we would need to move the data json to the right gcp instance)

server/__init__.py
```diff
@@ -248,7 +256,9 @@ def create_app():
   app.config['BABEL_TRANSLATION_DIRECTORIES'] = 'i18n'
 
   # load disaster dashboard data from GCS
-  if os.environ.get('FLASK_ENV') in ['autopush', 'local', 'dev']:
+  if os.environ.get('FLASK_ENV') in [
+      'autopush', 'local', 'dev', 'local-stanford', 'stanford'
+  ]:
```
beets, line 261: we should also move this into the "register_stanford_routes" function (though this is mostly temporary... so mostly a note for future that we want the flask app to only contain the smallest amount of required resources for the pages it serves)

server/__init__.py
juliawu: Done!

server/__init__.py
juliawu: Done!

server/__init__.py
chejennifer: I think this move may be causing tests for the disaster apis to fail because the blueprint isn't registered for the environment in which the tests are being run anymore

beets: ah good catch. we can also install it if `cfg.TEST`

juliawu: Done!

server/__init__.py
```diff
@@ -89,9 +89,19 @@ def register_routes_custom_dc(app):
...
-  # Apply blueprints specific to Stanford DC
+  # Install blueprints specific to Stanford DC
   from routes import (disasters)
+  from routes.api import (disaster_api)
   app.register_blueprint(disasters.bp)
+  app.register_blueprint(disaster_api.bp)
+
+  # load disaster dashboard data from GCS
+  if os.environ.get('FLASK_ENV') in [
+      'autopush', 'local', 'dev', 'local-stanford', 'stanford'
```
beets, line 100: Since this is part of register_routes_stanford_dc, we don't need this additional `if`.

server/__init__.py
```diff
@@ -88,6 +88,22 @@ def register_routes_custom_dc(app):
...
 
+def register_routes_stanford_dc(app):
+  # Install blueprints specific to Stanford DC
+  from routes import (disasters)
+  from routes.api import (disaster_api)
+  app.register_blueprint(disasters.bp)
+  app.register_blueprint(disaster_api.bp)
+
+  # load disaster dashboard data from GCS
+  if os.environ.get('FLASK_ENV') in [
```
chejennifer, line 99: nit: we don't need this if statement because if we're registering stanford_dc_routes, we need to load the disaster dashboard data

server/routes/disasters.py
```diff
@@ -0,0 +1,35 @@
...
+def disaster_dashboard():
+  if not os.environ.get('FLASK_ENV') in [
```
chejennifer, line 28: nit: since we're only registering these routes for stanford dc, we don't need this if anymore

server/__init__.py
```diff
@@ -180,6 +194,12 @@ def create_app():
   register_routes_common(app)
   if cfg.CUSTOM:
     register_routes_custom_dc(app)
+  if cfg.ENV_NAME == 'STANFORD':
+    register_routes_stanford_dc(app)
```
chejennifer, line 198: wonder if we should also register_routes_stanford_dc for local or autopush (like the original gating we had on the APIs)? what do you think @beets?

chejennifer (commented):
thanks for the updates!

server/__init__.py
beets: yes, autopush for now would be nice, though the data might deviate at some point.

juliawu: Added autopush.

server/routes/disasters.py
juliawu: Fixed!

server/__init__.py
juliawu: Fixed!

server/__init__.py
juliawu: Fixed!

chejennifer (approved):
Thanks for all the changes!

PR #1873, opened 2022-12-09
Add terraform GKE module

Fructokinase:
For simplicity this is a port over of the relevant parts from existing scripts. It allows GKE clusters to be created as part of the set up flow.

Shell script will be gradually replaced by the resources that the gcloud/kubectl command creates.

TESTED=https://pantheon.corp.google.com/kubernetes/clusters/details/us-central1/datacommons3/details?mods=-monitoring_api_staging&project=datcom-website-dev


deploy/terraform-datacommons-website/modules/gke/README.md
```diff
@@ -0,0 +1,23 @@
...
+```
+
+Note: Change the source path if the caller module is not from a particular example in the examples folder.
+
+## Inputs
+
+| Name         | Description                              | Type     | Default       | Required |
+| ------------ | ---------------------------------------- | -------- | ------------- | :------: |
+| project_id   | GCP project id to deploy the cluster to. | `string` | n/a           |   yes    |
+| cluster_name | Name of the GKE cluster to create.       | `string` | `datacommons` |    no    |
```
shifucun, line 21: Since we have multi-cluster setup. Should the cluster_name have region name, by default "us-central1"?

deploy/terraform-datacommons-website/modules/gke/configure_cluster.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+
+gcloud container clusters get-credentials $CLUSTER_NAME --region $REGION
```
shifucun, line 3: Will this file be changed to terraform resources?

deploy/terraform-datacommons-website/modules/gke/configure_cluster.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+
```
shifucun, line 2: nit: add license header and else where.

deploy/terraform-datacommons-website/modules/gke/README.md
Fructokinase: Good point. Changed cluster_name to cluster_prefix and will use the region as suffix

deploy/terraform-datacommons-website/modules/gke/configure_cluster.sh
Fructokinase: Ideally yes, piece by piece. Difference between this and terraform resources is that by converting to Terraform resources, Terraform knows what it means to upgrade and destroy the infra. For the initial use case of automating the setup though, this may be less important.

shifucun: that's good. For one time setup, a script is sufficient.

deploy/terraform-datacommons-website/modules/gke/configure_cluster.sh
Fructokinase: done.

Fructokinase:
for the newer change,

TESTED=https://pantheon.corp.google.com/kubernetes/clusters/details/us-central1/datacommons-us-central1/details?mods=-monitoring_api_staging&project=datcom-website-dev

shifucun (approved):

PR #1874, opened 2022-12-09
[disaster] Flask changes for updating topic page proto

chejennifer:
For the new disaster dashboard, we will be making it config driven & doing this by utilizing what we had for the topic pages. We would also like to make the config style page something that can be re-used in the future by other dashboards/apps. To differentiate between the topic page app and the shared config style pages, we will name shared proto/components/etc SubjectPage.

To make changes easier to view when reviewing, I'm splitting up the change into multiple PRs. This PR has:
- proto changes: naming change + change to the structure of Tile and Block
- updates to topic page configs, flask code & React code to be compatible with the new proto & allow current topic pages to render the same as on prod


static/css/topic_page.scss
```diff
@@ -60,10 +60,12 @@ $chart-container-top-margin: 20px;
   display: flex;
 }
 
-.left-tiles,
-.right-tiles {
+.block-column {
   display: flex;
   flex-direction: column;
+  padding: 0.5rem;
```
beets, line 66: Can we add this as a `gap` instead, to avoid an additional .5rem padding on the outer edges.

https://developer.mozilla.org/en-US/docs/Web/CSS/gap#flex_layout

beets (approved):

static/css/topic_page.scss
chejennifer: as discussed offline: row adds negative padding, so adding padding to the outer edges to get net 0 padding.

PR #1875, opened 2022-12-09
update landing page maps with additional us regions

n-h-diaz:
new maps:
Screenshot 2022-12-09 at 2 26 53 PM
Screenshot 2022-12-09 at 2 27 00 PM


beets (commented):
thanks for the updated charts!!

please shrink the image files by at least half of the width / height (the browser will have to do the image shrinking, and this gets expensive if the original file is very large)

beets (approved):

PR #1878, opened 2022-12-12
[Disease Browser Patch]: Fix Property Name

suhana13:
Issue: For the disease-symptom association charts, the name of the property `descriptorName` has been changed to simply `name` in the latest Disease Ontology data import. I have updated the name accordingly to fix the graph.
Screen Shot 2022-12-12 at 2 22 48 PM


static/js/biomedical/disease/data_processing_utils.ts
```diff
@@ -261,6 +261,7 @@ export function getCompoundDiseaseTreatment(
       clinicalPhaseNumber: Number(element.clinicalPhaseNumber),
     }));
   processedData.sort((a, b) => b.clinicalPhaseNumber - a.clinicalPhaseNumber);
+  console.log(processedData);
```
chejennifer, line 264: remove this console log?

static/js/biomedical/disease/data_processing_utils.ts
```diff
@@ -285,6 +286,7 @@ export function getCompoundDiseaseContraindication(
       drugSource: element.drugSource.toLowerCase(),
     }));
   processedData.sort((a, b) => (a.drugSource > b.drugSource ? 1 : -1));
+  console.log(processedData);
```
chejennifer, line 289: remove this console log?

static/js/biomedical/disease/data_processing_utils.test.tsx
```diff
@@ -184,7 +190,7 @@ test("getDiseaseCommonName", () => {
   }[] = [
     {
       data: dataDOID2403 as GraphNodes,
-      wantName: "Hepatitis B",
+      wantName: undefined,
```
chejennifer, line 193: Should the name still be undefined with this fix ..?

chejennifer (commented):
thanks for the fix

static/js/biomedical/disease/data_processing_utils.ts
suhana13: So sorry about the console statements. Removed em!

static/js/biomedical/disease/data_processing_utils.test.tsx
suhana13: So the fix is just changing the `descriptorName` property to `name` property within the disease-symptom association graph because there we're alluding to symptom names instead of disease names. In this case, for the disease name, we need to make a property addition in the config because some diseases have a `commonName` while some have just `name`. I haven't requested the name change in the config yet because I predict we would need to add some properties so it might be nice to submit a CL for all of them together. In the meantime, this fixes the disease-symptom association chart which is nice!

static/js/biomedical/disease/data_processing_utils.ts

chejennifer (approved):

PR #1880, opened 2022-12-12
[disaster] extract out shared SubjectPage component

chejennifer:
- extract out shared SubjectPage component to be ready to be used by disaster dashboard in a later PR
- do the extraction according to the new folder structure proposed in: [Website Folder Structure](https://docs.google.com/document/d/1eYvx8sYVNxJ5tWQ229gHb3lPCdepwBt5Z2TYKRdvSNQ/edit?resourcekey=0-_Sz3NjUjbJ621ZtVB9PD6Q)


static/js/components/subject_page/page_selector.tsx
```diff
@@ -23,14 +23,14 @@ import _ from "lodash";
 import React, { useEffect, useState } from "react";
```
beets, line 23: note: this should stay specific to topic pages.

static/js/chart/types.ts
```diff
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-import { Point } from "../topic_page/ranking_unit";
+import { Point } from "../types/ranking_unit_types";
```
beets, line 17: While we're here... rename Point to RankingPoint?  (we have so many types of points...)

static/js/types/subject_page_types.ts
```diff
@@ -0,0 +1,24 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * Types used by subject page components.
+ */
+
+export interface SubjectPageConfigSummary {
```
beets, line 21: Can we start to add more docs for these? This is used by the PageSelector to be able to create links to all known topic pages.

static/js/utils/tile_utils.ts
```diff
@@ -14,6 +14,10 @@
  * limitations under the License.
  */
 
+/**
+ * util functions used by tile components.
```
beets, line 18:
```suggestion
 * Util functions used by tile components.
```

beets (approved):
thanks for the refactor!

We would want each "user" of the SubjectPage configs to be able to create separate "universes" of pages. So all topic pages that would stay under /topic, then only disaster ones under /disaster. We would also want to allow for some variations. At the moment, I think that variation is with PageSelector, but there might be others (some might want to drop the sidebar, for example).

static/js/components/subject_page/main_pane.tsx
```diff
@@ -37,11 +42,11 @@ interface MainPanePropType {
...
    * Config of the page
    */
-  pageConfig: TopicPageConfig;
+  pageConfig: SubjectPageConfig;
   /**
-   * Summary for all topic page configs
+   * Summary of all available page configs
    */
-  topicsSummary: TopicsSummary;
+  allConfigsSummary: SubjectPageConfigSummary;
```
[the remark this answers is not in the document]
chejennifer: Updated it so that instead of one SubjectPage component, there are SubjectPageMainPane and SubjectPageSidebar  components that users of the SubjectPage config can use to create their page. This should allow for more variation in the pages that use SubjectPage config.

static/js/components/subject_page/page_selector.tsx
chejennifer: done

static/js/types/subject_page_types.ts
chejennifer: done

static/js/utils/tile_utils.ts
chejennifer: done

beets (approved):
thanks for the updates!

chejennifer:
Thanks for the review!

PR #1883, opened 2022-12-13
Custom DC: make resource suffix optional in setup flow

Fructokinase:
Tested with terraform apply command. See DNS zone for example.

https://pantheon.corp.google.com/net-services/dns/zones/datacommons/details?mods=-monitoring_api_staging&project=datcom-website-dev


shifucun (approved):
thanks!

PR #1888, opened 2022-12-13
Custom DC: Wrap terraform commands in a script

Fructokinase:
Wrap setup and installation of DC terraform commands inside one script.


scripts/get-dc.sh
```diff
@@ -0,0 +1,55 @@
...
+echo "Installing Custom Datacommons web application in $PROJECT_ID."
+
+# Clone DC website repo and mixer submodule.
+git clone https://github.com/datacommonsorg/website
+git submodule foreach git pull origin master
+git submodule update --init --recursive
+
+cd website/deploy/terraform-datacommons-website/examples/setup
+
+terraform init && terraform apply \
```
shifucun, line 40: is terraform in available out of the box?

shifucun (approved):
very nice!!

scripts/get-dc.sh
Fructokinase: It is on cloud shell (one of the benefits), but I should not assume that. Added a checker for gcloud and terraform commands.

shifucun: That's nice so this can run everywhere

PR #1889, opened 2022-12-14
get-dc.sh -> install_custom_dc.sh , make messages more clear

Fructokinase:
Here's an example of how it will look.

image


scripts/install_custom_dc.sh
```diff
@@ -83,8 +83,29 @@ terraform init && terraform apply \
...
-echo "You should expect the instance to be accessible via https://$PROJECT_ID-datacommons.com"
-echo "within 30 minutes or so after custom-datacommons-support@google.com responds."
+_success_msg="
+###############################################################################
+# Status: Successfully launched the installer in $PROJECT_ID.
+###############################################################################
+
+###############################################################################
+# Action required:
+###############################################################################
```
shifucun, line 93: move line93 after line95?

move line99 after line102

and make one extra line between the two blocks?

Fructokinase: How does it look now?

shifucun (approved):
thanks for update!

scripts/install_custom_dc.sh
shifucun: LG!

scripts/install_custom_dc.sh
```diff
@@ -83,8 +83,32 @@ terraform init && terraform apply \
...
+Please don't forget to email support+custom@datacommons.org with your
+GCP project id for data access.
+###############################################################################
+
+
+###############################################################################
+# Action required:
+Please also make sure to click on the activation email for $DOMAIN
+If the contact email has previously been used to verify domains,
+then $DOMAIN will be already active without needing activation emails.
```
beets, line 103: Is this the same as saying: "This step is only required for new and unverified domains"?
The "Action required" heading is also optional, in this case.

PR #1892, opened 2022-12-15
[stanford] add gcs bucket to stanford config

chejennifer:
On server start up, if the ENV is stanford, local-stanford, or autopush, we need to read disaster data files from the GCS bucket

### DIFF
--- a/server/configmodule.py
+++ b/server/configmodule.py
@@ -80,6 +80,7 @@ class StanfordConfig(CustomConfig):
   ENV_NAME = 'STANFORD'
   ENABLE_BLOCKLIST = True
   BASE_HTML_PATH = 'custom_dc/stanford/base.html'
+  GCS_BUCKET = 'datcom-stanford-resources'
 
 
 class StanfordStagingConfig(StanfordConfig):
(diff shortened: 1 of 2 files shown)


server/configmodule.py
```diff
@@ -119,7 +120,6 @@ class LocalBaseConfig(Config):
   AI_CONFIG_PATH = os.path.abspath(
       os.path.join(os.path.curdir, '..', 'deploy/overlays/local/ai.yaml'))
   SCHEME = 'http'
-  GCS_BUCKET = 'datcom-website-autopush-resources'
```
shifucun, line 122: autopush also sets "GCS_BUCKET", is that needed?

jehangiramjad (approved):

server/configmodule.py
chejennifer: yes, we are also allowing disaster dashboard on autopush, so autopush will also be reading disaster data files from GCS bucket on server start up

shifucun: I see, so the files are saved in different buckets? If this is for development purpose and all instance are synced, I wonder if it's easy to save the data in a public bucket and accessible to any instance with dashboard.

chejennifer: I've saved the same files in both buckets .. but this is also temporary & will go away once we move to using the event cache API

beets (approved):

PR #1895, opened 2022-12-16
Custom DC: Add a BT cache module

Fructokinase:
This Terraform module defines resources for 
- BT instance
- GCF source code
- GCF

The GCF source code is copied over using curl during terraform apply.

TESTED=with a main.tf, cloud function created below.
https://pantheon.corp.google.com/functions/details/us-central1/prophet-cache-trigger-datcom-website-dev-alex20221215?env=gen1&mods=-monitoring_api_staging&project=datcom-website-dev

### DIFF
--- a/deploy/terraform-datacommons-website/modules/bt_cache/main.tf
+++ b/deploy/terraform-datacommons-website/modules/bt_cache/main.tf
@@ -1,0 +1,105 @@
+/**
+ * Copyright 2022 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+locals {
+    resource_suffix = (
+        var.resource_suffix != null ?
+        format("-%s", var.resource_suffix) : ""
+    )
+}
+
+# BigTable instance used to store cache tables.
+resource "google_bigtable_instance" "bt_cache" {
+  name           = format("dc-graph%s", local.resource_suffix)
+  project        = var.project_id
+
+  cluster {
+    # There will be one cluster. Constant seems appropriate for now.
+    cluster_id   = "dc-graph-c1"
+    zone         = var.bt_instance_zone
+
+    autoscaling_config {
+      min_nodes  = var.bt_autoscaling_min_nodes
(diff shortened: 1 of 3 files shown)


deploy/terraform-datacommons-website/modules/bt_cache/variables.tf
```diff
@@ -0,0 +1,58 @@
...
+
+variable "dc_resource_bucket" {
+  type        = string
+  description = "Name of the GCS bucket that stores DC resources. Do not include the scheme(gs://)."
+}
+
+variable "csv2bt_template_path" {
+  type        = string
+  description = "Full GCS path, including scheme, to the Dataflow template to load BT cache."
+  default     = "gs://datcom-dataflow-templates/templates/flex/csv_to_bt.json"
```
Fructokinase, line 40: This will be created from the commands in this PR: https://github.com/datacommonsorg/tools/pull/163.

deploy/terraform-datacommons-website/modules/bt_cache/variables.tf
```diff
@@ -0,0 +1,58 @@
...
+  description = "Optional suffix to add to GCP resources for uniqueness."
+  default     = null
+}
+
+variable "project_id" {
+  type        = string
+  description = "This is the same GCP project id from the setup step."
+}
+
+variable "region" {
```
Fructokinase, line 27: Rather than defining the default at the module level, the default will be defined in the instance (examples folder) level and passed down. Same for other variables that are shared.

shifucun:
Why need to copy over the BT automation code? Can we use the go module directly?

Fructokinase:
> Why need to copy over the BT automation code? Can we use the go module directly?

Added a script to curl the go module directly as spoken offline.

deploy/terraform-datacommons-website/modules/bt_cache/main.tf
```diff
@@ -0,0 +1,105 @@
...
+resource "google_bigtable_instance" "bt_cache" {
+  name           = format("dc-graph%s", local.resource_suffix)
+  project        = var.project_id
+
+  cluster {
+    # There will be one cluster. Constant seems appropriate for now.
+    cluster_id   = "dc-graph-c1"
+    zone         = var.bt_instance_zone
+
+    autoscaling_config {
```
Fructokinase, line 33: @shifucun I just realized that we don't actually autoscale our BT instance right now (scaling up and down is signaled in GCF). For custom DC, do we want to keep existing behaviour(manual scaling)?

shifucun: For custom dc we should do autoscaling, can we update that?

Fructokinase: Current PR assumes autoscaling. I will update the gcf function with a follow up PR to remove the manual scaling.

Fructokinase:
As spoken offline, will move this PR to tools repo. Keeping it here in the meantime for reference.

Fructokinase:
Moved to https://github.com/datacommonsorg/tools/pull/166

PR #1898, opened 2022-12-16
Stanford Staging: Do not inherit Stanford prod config

Fructokinase:
It is easy to make mistakes if the staging config is inherited from prod config. For example, the staging instance is currently trying to read the prod GCS bucket because the GCS bucket is not set explicitly in config.

image


jehangiramjad (approved):
Thanks!!

PR #1902, opened 2022-12-21
Add structure for NL interface: Model, API, Page

shifucun:
Verified e2e with data fetched and page rendering. Need more work to add in the data processing and building chart config.


server/services/nl.py
```diff
@@ -0,0 +1,82 @@
...
+    for build in BUILDS:
+      logging.info('Loading build {}'.format(build))
+      ds = load_dataset('csv', data_files=f'/tmp/embeddings_{build}.csv')
+      df = ds["train"].to_pandas()
+      self.dcid_maps[build] = df['dcid'].values.tolist()
+      df = df.drop('dcid', axis=1)
+      self.dataset_embeddings_maps[build] = torch.from_numpy(df.to_numpy()).to(
+          torch.float)
+
+  def download_embeddings(self):
```
pradh, line 50: nit: internal function, _download_embeddings?

server/services/nl.py
```diff
@@ -0,0 +1,82 @@
...
+
+  def __init__(self):
+    self.model = SentenceTransformer('all-MiniLM-L6-v2')
+    self.dataset_embeddings_maps = {}
+    self.download_embeddings()
+    self.dcid_maps = {}
+    for build in BUILDS:
+      logging.info('Loading build {}'.format(build))
+      ds = load_dataset('csv', data_files=f'/tmp/embeddings_{build}.csv')
+      df = ds["train"].to_pandas()
```
pradh, line 44: To confirm, this is using pandas, is that okay?

Also, I hope numpy is okay to include because many of the libs might depend on it

server/services/nl.py
```diff
@@ -0,0 +1,82 @@
...
+  """Holds clients for the language model"""
+
+  def __init__(self):
+    self.model = SentenceTransformer('all-MiniLM-L6-v2')
+    self.dataset_embeddings_maps = {}
+    self.download_embeddings()
+    self.dcid_maps = {}
+    for build in BUILDS:
+      logging.info('Loading build {}'.format(build))
+      ds = load_dataset('csv', data_files=f'/tmp/embeddings_{build}.csv')
```
pradh, line 43: os.path.join(TEMP_DIR, f'embeddings_{build}.csv')

server/services/nl.py
```diff
@@ -0,0 +1,82 @@
...
+      self.dataset_embeddings_maps[build] = torch.from_numpy(df.to_numpy()).to(
+          torch.float)
+
+  def download_embeddings(self):
+    storage_client = storage.Client()
+    bucket = storage_client.bucket(bucket_name=GCS_BUCKET)
+    blobs = bucket.list_blobs(prefix=EMBEDDINGS)  # Get list of files
+    for blob in blobs:
+      _, filename = os.path.split(blob.name)
+      blob.download_to_filename(TEMP_DIR + filename)  # Download
```
pradh, line 56: nit: use os.path.join?

server/services/nl.py
```diff
@@ -0,0 +1,82 @@
...
+from google.cloud import storage
+from sentence_transformers import SentenceTransformer
+from sentence_transformers.util import semantic_search
+
+import os
+import torch
+from datasets import load_dataset
+import logging
+
+BUILDS = ['demographics300', 'uncurated3000']
```
pradh, line 25: As a note, looks like we are only using BUILD, so could load only that or leave TODO.

server/services/nl.py
```diff
@@ -0,0 +1,82 @@
...
+BUILD = 'uncurated3000'  #@param ['demographics300', 'uncurated3000']
+QUERY = "people who cannot see"  #@param {type:"string"}
```
pradh, line 27: QUERY can be deleted?

pradh (approved):
Thanks!

server/services/nl.py
```diff
@@ -0,0 +1,82 @@
...
+    blobs = bucket.list_blobs(prefix=EMBEDDINGS)  # Get list of files
+    for blob in blobs:
+      _, filename = os.path.split(blob.name)
+      blob.download_to_filename(TEMP_DIR + filename)  # Download
+
+  def search(self, query):
+    query_embeddings = self.model.encode([query])
+    hits = semantic_search(query_embeddings,
+                           self.dataset_embeddings_maps[BUILD],
+                           top_k=15)
```
jehangiramjad, line 62: nit: should this be a config?

server/services/nl.py
jehangiramjad: Also, should we set BUILD as a config?

server/services/nl.py
```diff
@@ -0,0 +1,82 @@
...
+GCS_BUCKET = 'datcom-csv'
+EMBEDDINGS = 'embeddings/'
+TEMP_DIR = '/tmp/'
+
+
+class Model:
+  """Holds clients for the language model"""
+
+  def __init__(self):
+    self.model = SentenceTransformer('all-MiniLM-L6-v2')
```
jehangiramjad, line 37: nit: this should also be set as either a global const or a config. We could switch to different models at some point.

server/services/nl.py
jehangiramjad: Also, use the EMBEDDINGS constant

server/routes/nl_interface.py
```diff
@@ -0,0 +1,41 @@
...
+    flask.abort(404)
+  model = current_app.config['NL_MODEL']
+  scores, svs = model.search('people who cannot see')
+  print(scores)
+  print(svs)
+  place_page_data = dc.get_landing_page_data(dcid, 'Overview', [])
+  svg_info = dc.get_variable_group_info("Count_Person", [])
+  # Process data as needed
+  print(place_page_data)
+  print(svg_info)
```
chejennifer, line 39: not sure if you meant to keep all the print statements?

server/services/nl.py
```diff
@@ -0,0 +1,82 @@
...
+    query_embeddings = self.model.encode([query])
+    hits = semantic_search(query_embeddings,
+                           self.dataset_embeddings_maps[BUILD],
+                           top_k=15)
+
+    # Note: multiple results may map to the same DCID. As well, the same string may
+    # map to multiple DCIDs with the same score.
+    sv2score = {}
+    score2svs = {}
+    for e in hits[0]:
```
chejennifer, line 68: is hits guaranteed to have at least one element? otherwise should add a check that it's not empty before accessing the first element 

nit: instead of single letter variables, maybe we can use variable names that tell us a little more about what the variable is?

server/services/nl.py
```diff
@@ -0,0 +1,82 @@
...
+      for d in self.dcid_maps[BUILD][e['corpus_id']].split(','):
```
chejennifer, line 69: accessing a few fields here, are they all guaranteed to be present? otherwise, might want to use `.get()` instead

server/routes/nl_interface.py
```diff
@@ -0,0 +1,41 @@
...
+from flask import current_app
+import flask
+from flask import Blueprint, render_template
+
+import services.datacommons as dc
+
+bp = Blueprint('nl', __name__, url_prefix='/nl')
+
+
+@bp.route('/<path:dcid>')
```
jehangiramjad, line 27: what dcid is this? as far as i know, the landing page will have a search box where the user specifies a query/sentence. We then infer the SV dcid and place(s) dcid. So, shouldn't this load without the dcid? I see that it is being used to get place data ( place_page_data = dc.get_landing_page_data(dcid, 'Overview', []) ), but at this point we don't know what place, right?

jehangiramjad (commented):
Thanks! Some comments/clarifications

server/services/nl.py
```diff
@@ -0,0 +1,82 @@
...
+        s = e['score']
+        # Prefer the top score.
+        if d not in sv2score:
+          sv2score[d] = s
```
chejennifer, line 73: correct me if i'm wrong, it looks like sv2score is just a map used to keep track of `d` that has been seen and the value isn't actually used? so maybe a set would be better for this

chejennifer (approved):
nice! few nits and suggestions

server/services/nl.py
pradh: scores will be used in future :)

server/services/nl.py
shifucun: Done

server/services/nl.py
shifucun: I am copying the original code, so maybe keep it as is and update in later rounds?

server/services/nl.py
shifucun: same here

PR #1906, opened 2022-12-22
Quick fix for broken subject page tiles

juliawu:
Many tile types on the subject/topic pages fail to render due to not having a `disasterEventMapTileSpec` specified. This change moves logic specific to the disaster event map to the dedicated disaster_event_map case.

Current behavior:
Screenshot 2022-12-21 at 4 39 30 PM

Proposed:
Screenshot 2022-12-21 at 4 39 55 PM


chejennifer (approved):
nice catch, thanks for the fix!

PR #1923, opened 2022-12-26
Reduce image size by removing pip cache and nvidia lib

Fructokinase:
This PR removes things that can be removed without changing functionality. 

Debate of where to put the language model and the torch library aside, there are immediate improvements that can be made to reduce the compressed image size from 5.5G to 2.1G (over 60% reduction). This saves both build and deploy time. Cloud build now runs in ~9min, which is an improvement from ~14min. It is further observed that 8CPU machine and 32CPU machine doesn't seem to make a difference.

Pip cache takes 2.7G: 
Screen Shot 2022-12-23 at 3 47 19 PM

Nvidia takes 1.4G:
Screen Shot 2022-12-23 at 3 38 03 PM

Note that this is still not ideal as our image size is still > 3times larger than before and this has a noticeable difference in both build and deploy times. Next steps would be to discuss whether we can move nlp features to a separate container, and launched together like mixer and esp.

Tested with scripts/push_image.sh: [Cloud run took 8min 6sec](https://pantheon.corp.google.com/cloud-build/builds;region=global/34368def-85df-4b6c-97a3-d2056e9ef963;tab=detail?mods=-monitoring_api_staging&project=datcom-ci), [image is 2.1G](https://pantheon.corp.google.com/gcr/images/datcom-ci/global/datacommons-website@sha256:7c2f207630f7ecbc979f296bf741536151f9d35a7111aa28e7ba89016d56145d/details?mods=-monitoring_api_staging&project=datcom-ci)

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -20,7 +20,9 @@ ENV ENV=${ENV}
 WORKDIR /workspace
 
 COPY server/requirements.txt /workspace/server/requirements.txt
-RUN pip3 install -r /workspace/server/requirements.txt
+# --no-cache-dir removes ~/.cache files, which can be a few GB.
+RUN pip3 install --no-cache-dir \
+    -r /workspace/server/requirements.txt
 
 COPY static/package.json /workspace/static/package.json
 COPY static/package-lock.json /workspace/static/package-lock.json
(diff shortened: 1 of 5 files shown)


jehangiramjad (approved):
Thanks

PR #1925, opened 2022-12-27
[disaster] add severity filters

chejennifer:
- add ability to filter the events that show up by specified props

screen-recording-_55_ (1)

### DIFF
--- a/server/config/disaster_dashboard/Earth.textproto
+++ b/server/config/disaster_dashboard/Earth.textproto
@@ -29,6 +29,7 @@ metadata {
       name: "Earthquake"
       event_type_dcids: "EarthquakeEvent"
       color: "#930000"
+      severity_props: "magnitude"
     }
   }
   event_type_spec {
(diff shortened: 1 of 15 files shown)


server/config/disaster_dashboard/Earth.textproto
```diff
@@ -29,6 +29,7 @@ metadata {
       name: "Earthquake"
       event_type_dcids: "EarthquakeEvent"
       color: "#930000"
+      severity_props: "magnitude"
```
juliawu, line 32: For my own understanding: Are earthquakes the only event type with magnitudes? The severity filter will eventually be applied to things like droughts and hurricanes, right?

chejennifer: I believe Earthquakes are the only event type with magnitudes right now. Yes, the severity filter will be applied for any event types that we set severity_props for in the config

juliawu (approved):
Changes look good to me!

chejennifer:
thanks for the review!

PR #1930, opened 2023-01-03
Various fixes for NL interface

shifucun:
* Correctly set search box text from url query string
* Fix duplicate react component key issue
* Check containedPlaceTypes is not undefined


jehangiramjad (approved):
Lovely, thanks!

PR #1931, opened 2023-01-03
Add cookies support in NL interface; Clean up page data

shifucun:
* The cookie is used within the /nl page only.
* The html data is not needed since the data is fetched from the client after page loads now.

### DIFF
--- a/static/js/apps/nl_interface/app.tsx
+++ b/static/js/apps/nl_interface/app.tsx
@@ -20,6 +20,7 @@
 
 import axios from "axios";
 import React, { useEffect, useState } from "react";
+import { useCookies } from "react-cookie";
 import { Col, Container, Row } from "reactstrap";
 
 import { SubjectPageMainPane } from "../../components/subject_page/main_pane";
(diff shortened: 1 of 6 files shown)


jehangiramjad (approved):
Thanks!

static/js/apps/nl_interface/app.tsx
```diff
@@ -162,6 +164,9 @@ export function App(): JSX.Element {
                       null,
                       `/nl?q=${q}&build=${selectedBuild}`
                     );
+                    const queries: string[] = cookies["q"] || [];
+                    queries.push(q);
+                    setCookie("q", queries);
```
beets, line 169: we should set an expiration for the cookie (and some limit of queries for context?)

we should also set a path to limit this to /nl pages.

shifucun: From the documentation https://github.com/reactivestack/cookies/tree/master/packages/react-cookie#setcookiename-value-options, if the path is not set, the cookie should be restricted to the CookieProvider component of this app only.

For query limit, do you suggest we only keep the last N queries?

beets: yes, a limit for number of queries (maybe 10?), and also set the maxAge on the cookie. of the two, maxAge is probably most important, since we don't want to retain the context for too long..

hm, it was not obvious to me from the documentation that the cookie would be restricted to the app / path. but we can easily verify that.

PR #1933, opened 2023-01-03
[disaster] Minor fixes/updates

chejennifer:
- add FireEvent event type to list of eventType dcids for Fire
- check that date is valid before using for the dropdown

### DIFF
--- a/static/js/utils/string_utils.ts
+++ b/static/js/utils/string_utils.ts
@@ -89,6 +89,10 @@ export function getCommonPrefix(words: string[]): string {
  * Given a date string, check that it is in the form YYYY-MM-DD or YYYY-MM or YYYY
  */
 export function isValidDate(date: string): boolean {
+  const dateObj = new Date(date);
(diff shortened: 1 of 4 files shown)


static/js/utils/string_utils.ts
```diff
@@ -89,6 +89,10 @@ export function getCommonPrefix(words: string[]): string {
  * Given a date string, check that it is in the form YYYY-MM-DD or YYYY-MM or YYYY
  */
 export function isValidDate(date: string): boolean {
+  const dateObj = new Date(date);
+  if (dateObj.toString() === "Invalid Date") {
```
beets, line 93: This could be platform dependent. Could we instead use:
```
Date.parse(date) !== NaN
```

beets (approved):

static/js/utils/string_utils.ts
beets: Add a quick test for this too? Since it's a util..

chejennifer: done

PR #1939, opened 2023-01-04
Move imports for NL models into flag guard

beets:
Server start up was slow even for the imports. Moving it all behind the flag guard.

Tested the site with/out -m flag and the NL pages continue to work as expected.


beets:
Looks like @jehangiramjad beat me to the punch :)

PR #1944, opened 2023-01-05
[disasters] fix severity prop name for earthquake magnitude

chejennifer:


server/config/subject_page_pb2.py
```diff
@@ -2,9 +2,10 @@
 # Generated by the protocol buffer compiler.  DO NOT EDIT!
 # source: subject_page.proto
 """Generated protocol buffer code."""
-from google.protobuf.internal import builder as _builder
```
shifucun, line 5: Curious what protoc version do you use? The format of this file has been changing through the PRs

beets: i've also been explicitly skipping submitting the file for most prs.

we could add the protos to be generated at build time, and not check it in.. wdyt?

beets (approved):
thanks for the fix!

we could revert the compiled proto file for this pr..

server/config/subject_page_pb2.py
chejennifer: I'm using 3.19.4 

but yes, will revert this file!

shifucun: Not checking this in might work, but we need to ensure `protoc` runs in local/webdriver/cloudbuild etc.. For now, I can add a check for protoc version in `run_server.sh`? If it's not desired version, protoc won't run...

shifucun: https://github.com/datacommonsorg/website/pull/1945 sent out a fix for this

PR #1945, opened 2023-01-05
Use a specific protoc version

shifucun:


chejennifer (approved):
thanks for adding the check!

PR #1949, opened 2023-01-06
[disasters] update URL when user clicks a region or a breadcrumb

chejennifer:
screen-recording-_57_

### DIFF
--- a/static/js/components/tiles/disaster_event_map_tile.tsx
+++ b/static/js/components/tiles/disaster_event_map_tile.tsx
@@ -58,6 +58,7 @@ import {
   getEnclosedPlacesPromise,
   getParentPlacesPromise,
 } from "../../utils/place_utils";
+import { getPlaceNames } from "../../utils/place_utils";
 import { ReplacementStrings } from "../../utils/tile_utils";
 import { ChartTileContainer } from "./chart_tile";
 import { DisasterEventMapFilters } from "./disaster_event_map_filters";
@@ -68,6 +69,9 @@ const ZOOM_OUT_BUTTON_ID = "zoom-out-button";
 const CONTENT_SPINNER_ID = "content-spinner-screen";
 const CSS_SELECTOR_PREFIX = "disaster-event-map";
 const DATE_SUBSTRING_IDX = 10;
+const BREADCRUMB_PARAM_KEY = "bc";
+const PARAM_SEPARATOR = "__";
+const REDIRECT_URL_PREFIX = "/disasters/";
 
 interface DisasterEventMapTilePropType {
   // Id for this tile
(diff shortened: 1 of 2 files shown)


static/js/components/tiles/disaster_event_map_tile.tsx
```diff
@@ -315,6 +312,41 @@ export function DisasterEventMapTile(
       });
   }
 
+  /**
+   * Updates breadcrumbs using URL params and current place.
+   */
+  function updateBreadcrumbs(selectedPlace: NamedTypedPlace): void {
```
beets, line 318: we could technically compute this from parent places. add a TODO, so we won't rely on a URL param?

static/js/components/tiles/disaster_event_map_tile.tsx
```diff
@@ -68,6 +69,9 @@ const ZOOM_OUT_BUTTON_ID = "zoom-out-button";
 const CONTENT_SPINNER_ID = "content-spinner-screen";
 const CSS_SELECTOR_PREFIX = "disaster-event-map";
 const DATE_SUBSTRING_IDX = 10;
+const BREADCRUMB_PARAM_KEY = "bc";
+const PARAM_SEPARATOR = "__";
+const REDIRECT_URL_PREFIX = "/disasters/";
```
beets, line 74: Add a TODO to make this config driven?

beets (approved):
thanks!

static/js/components/tiles/disaster_event_map_tile.tsx
chejennifer: added a TODO so we can revisit this, but I didn't want to rely on parent places because there may be parent places that weren't seen by the user. 

E.g., parent places for USA are: Earth, North America. However, if the user clicked on USA from the Earth page, we wouldn't want North America to show up in the breadcrumbs.

static/js/components/tiles/disaster_event_map_tile.tsx
chejennifer: good call, added a TODO

static/js/components/tiles/disaster_event_map_tile.tsx
beets: i've always seen breadcrumbs as a way to learn the navigation (so if USA was the first page you clicked on, you would learn that there is also an Earth page).  just like how we have it on docsite. less so of a navigation history tool.

chejennifer: I see, if it's more of a navigation tool, then using parent places makes sense. Although, we might want to hide continents

PR #1953, opened 2023-01-06
[disasters] update severity filtering

chejennifer:
- update severity filtering to be done using the mixer api update

screen-recording-_58_

### DIFF
--- a/server/routes/api/disaster_api.py
+++ b/server/routes/api/disaster_api.py
@@ -77,14 +77,9 @@ def event_data():
     return "error: must provide a place field", 400
   filter_prop = request.args.get('filterProp', '')
   filter_unit = request.args.get('filterUnit', '')
-  filter_upper_limit = request.args.get('filterUpperLimit', None)
-  filter_lower_limit = request.args.get('filterLowerLimit', None)
-  # if filtering values, need all of filter prop, filter upper limit, and filter
-  # lower limit. Filter unit is optional.
-  if filter_prop and filter_upper_limit and filter_lower_limit:
-    result = dc.get_event_collection(event_type, place, date, filter_prop,
-                                     filter_unit, float(filter_upper_limit),
-                                     float(filter_lower_limit))
-  else:
-    result = dc.get_event_collection(event_type, place, date)
+  filter_upper_limit = float(request.args.get('filterUpperLimit', '0'))
+  filter_lower_limit = float(request.args.get('filterLowerLimit', '0'))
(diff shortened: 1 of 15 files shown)


shifucun:
Curious how is the initial page loading time?

chejennifer:
> Curious how is the initial page loading time?

Looking at the network calls, the longest took about 6.5s

server/routes/api/disaster_api.py
```diff
@@ -74,5 +75,16 @@ def event_data():
...
+  filter_upper_limit = request.args.get('filterUpperLimit', None)
+  filter_lower_limit = request.args.get('filterLowerLimit', None)
+  # if filtering values, need all of filter prop, filter upper limit, and filter
+  # lower limit. Filter unit is optional.
+  if filter_prop and filter_upper_limit and filter_lower_limit:
+    result = dc.get_event_collection(event_type, place, date, filter_prop,
+                                     filter_unit, float(filter_upper_limit),
+                                     float(filter_lower_limit))
+  else:
+    result = dc.get_event_collection(event_type, place, date)
```
shifucun, line 89: is this else condition necessary? If the filter params are not set (and take 0 value), there will be no filtering in mixer.

shifucun:
> > Curious how is the initial page loading time?
> 
> Looking at the network calls, the longest took about 6.5s

Wonder if can make more aggressive filtering on fire, so they don't look dense in the map, and the loading could be much faster, say under 2 sec?

chejennifer:
> > > Curious how is the initial page loading time?
> > 
> > 
> > Looking at the network calls, the longest took about 6.5s
> 
> Wonder if can make more aggressive filtering on fire, so they don't look dense in the map, and the loading could be much faster, say under 2 sec?

Yes, this is just using a range that includes all the fire events. I thought we still needed to figure out the defaults to use? Once we know the default range that we want to use, we can just update the config

shifucun (approved):

server/routes/api/disaster_api.py
chejennifer: if we get an empty filterUpperLimit or empty filterLowerLimit from the client, converting a None value to a float will throw an error so I thought it would be easier to handle it this way

shifucun: Maybe do `filter_upper_limit = request.args.get('filterUpperLimit') or 0`, then can just keep line 85-87 with full args

chejennifer: done

PR #1956, opened 2023-01-07
NL: cache model in disk for local dev

shifucun:
This would save a ton of time for local development. Right now each change would download and build the model from scratch.

### DIFF
--- a/server/__init__.py
+++ b/server/__init__.py
@@ -42,7 +42,7 @@ propagator = google_cloud_format.GoogleCloudFormatPropagator()
 
 nl_model_cache_key = 'nl_model'
 nl_model_cache_path = '~/.datacommons/'
-nl_model_cache_expire = 3600  # Cache for 1 hour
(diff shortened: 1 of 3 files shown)


server/__init__.py
```diff
@@ -40,6 +40,10 @@
 
 propagator = google_cloud_format.GoogleCloudFormatPropagator()
 
+nl_model_cache_key = 'nl_model'
+nl_model_cache_path = '~/.datacommons/'
+nl_model_cache_expire = 3600  # Cache for 1 hour
```
jehangiramjad, line 45: Could we make this last longer? Like 24*3600? The model inputs don't change often. And if they do, one can always empty the cache.

jehangiramjad (approved):
Very nice! Thank you :)

server/__init__.py
shifucun: Done

PR #1961, opened 2023-01-07
[nl] add exporting charts for nl

chejennifer:
- allow downloading svg and csv data for each chart in a tile

some bug fixes along the way:
- fix data fetch for bivariate tile (param was missing a plural)
- fix "download chart image" only showing up the second time the export button is clicked (here and place charts)

ezgif com-gif-maker

### DIFF
--- a/static/js/components/tiles/bivariate_tile.tsx
+++ b/static/js/components/tiles/bivariate_tile.tsx
@@ -37,6 +37,7 @@ import {
   shouldShowMapBoundaries,
 } from "../../tools/shared_util";
 import { stringifyFn } from "../../utils/axios";
+import { scatterDataToCsv } from "../../utils/chart_csv_utils";
 import { getStringOrNA } from "../../utils/number_utils";
 import { getPlaceScatterData } from "../../utils/scatter_data_utils";
 import { getStatVarName, ReplacementStrings } from "../../utils/tile_utils";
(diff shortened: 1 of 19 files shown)


static/css/bq_modal.scss
```diff
@@ -18,6 +18,14 @@
 
 @import "shared/modal.scss";
 
+// TODO (chejennifer): move this back to shared modal scss file.
```
beets, line 21: super nit - could you add context for what's required to to this?

fwiw, i think it's ok for the modals to go very wide at the smaller screen sizes, even if the content is then surrounded by lots of padding.

static/js/components/tiles/bivariate_tile.tsx
```diff
@@ -146,7 +157,7 @@ function getPopulationPromise(
         params: {
           parent_entity: placeDcid,
           child_type: enclosedPlaceType,
-          variable: variables,
+          variables: variables,
```
beets, line 160: thanks for the fix!

static/js/components/tiles/chart_tile.tsx
```diff
@@ -28,29 +30,78 @@ interface ChartTileContainerProp {
...
+                event.preventDefault();
+                handleEmbed();
+              }}
+            >
+              Export
+            </a>
+          )}
+        </div>
+      </footer>
+      <ChartEmbed ref={embedModalElement} />
```
beets, line 75: should this also be wrapped with props.allowEmbed?

beets (approved):
very nice - thanks for the fixes and the tests!

beets:
will merge this in first -- we can fix in followups!

static/css/bq_modal.scss
chejennifer: this was a piece of code that was originally in the shared modal stylesheet and I think you had added it so I don't actually know the context of what it does. The reason I had to move it out of the shared modal stylesheet was that when I was trying to "@forward" and "@use" the different stylesheets, it was complaining about this piece of code ..

static/js/components/tiles/chart_tile.tsx
chejennifer: yes, good call, will make a follow up PR for this

PR #1967, opened 2023-01-08
Add chart spec to debug info

shifucun:


jehangiramjad (approved):
Thanks! Do we really want all of charts config to be shown in the debug?

beets (approved):
maybe hide the chart spec behind another toggle?

shifucun:
For the sake of time, I will update the UI later..

PR #1969, opened 2023-01-09
Add a simple place type detection in query

Fructokinase:


server/services/nl.py
```diff
@@ -160,13 +161,28 @@ def _temporal_classification(self, prediction) -> Union[NLClassifier, None]:
...
 
   def _containedin_classification(self,
-                                  prediction) -> Union[NLClassifier, None]:
+                                  prediction, query: str) -> Union[NLClassifier, None]:
     if prediction != "Contained In":
       return None
 
+    contained_in_place_type = ContainedInPlaceType.PLACE
+    place_type_to_enum = OrderedDict({
+      "county": ContainedInPlaceType.COUNTY,
```
Fructokinase, line 170: @jehangiramjad  let me know if the orders of detection are correct (ex: county before city)

jehangiramjad: Good question. Given that we have a lot more data about counties vs cities, having that as top is good. How about:

1. County
2. State
3. Country
4. City

The rest are fine. as they are. 

But also, please also look for plurals or some sort of regex so that plurals can also be matched?

jehangiramjad (approved):
Thanks! Approving to unblock. But please also update the string here: 

Add the detected type to this string here: https://github.com/datacommonsorg/website/blob/f922866b52d19c4b5cde831b4a9239c7ea59433f/server/routes/nl_interface.py#L414
This string is displayed on the webpage under the debug section so it just needs to be a string appended.

beets:
@Fructokinase I'm compiling a dict of placetype -> plurals if that's helpful. you could use it in a follow-on pr

PR #1970, opened 2023-01-09
[NL] smaller charts

beets:
shrink charts down to 160px tall, and 3 charts wide (also removed legends, margins, etc)
also fixes chart / block titles with plurals

image


beets:
i'll merge and can fix comments in a follow up!

PR #1982, opened 2023-01-10
[NL] hide charts if there are more than 6

beets:


static/css/nl_interface.scss
```diff
@@ -113,20 +113,29 @@ $answer-min-height: 100vh;
...
+  .line-chart .legend,
+  .block-title {
     display: none;
   }
 
-  .block-title {
-    display: none;
+  .tile-hidden {
+    position: absolute;
+    top: -1000px;
```
chejennifer, line 123: nit: could this just be `display: none`?

static/js/components/subject_page/block.tsx
```diff
@@ -52,10 +53,16 @@ export interface BlockPropType {
...
+const NUM_TILES_SHOWN = 6; // for NL: show 2 rows (3 tiles per row).
+const HIDE_TILE_CLASS = "tile-hidden";
+
 export function Block(props: BlockPropType): JSX.Element {
   const columnWidth = props.columns
     ? `${(100 / props.columns.length).toFixed(2)}%`
     : "0";
+  // HACK for NL. Assumes all charts are in a single column.
+  const showExpando = isNlInterface();
+  const hideIfIndexGreater = showExpando ? NUM_TILES_SHOWN : -1;
```
chejennifer, line 65: nit: minIdxToHide or something like that? 'hideIfIndexGreater' sounds like it should be a boolean of whether or not to hide.

static/js/components/subject_page/block.tsx
```diff
@@ -81,13 +95,33 @@ export function Block(props: BlockPropType): JSX.Element {
...
+  const selfEl = e.target as HTMLAnchorElement;
+  const parentEl = selfEl.parentElement;
+  const tiles = parentEl.getElementsByClassName(
+    HIDE_TILE_CLASS
+  ) as HTMLCollectionOf<HTMLElement>;
+  for (let i = 0; i < tiles.length; i++) {
+    tiles[i].classList.remove(HIDE_TILE_CLASS);
+  }
+  selfEl.hidden = true;
+  e.preventDefault;
```
chejennifer, line 110: e.preventDefault()

chejennifer (approved):
thanks, small things!

static/css/nl_interface.scss
beets: with display: none, all the calls to getBBox() during chart drawing receives 0, so the charts never render. this way, they are rendered but out of the way

static/js/components/subject_page/block.tsx
beets: good call - done

static/js/components/subject_page/block.tsx
beets: thanks for the catch! done

PR #1990, opened 2023-01-10
[NL] set context history using callbacks rather than cookie

beets:


static/js/apps/nl_interface/app.tsx
```diff
@@ -44,8 +45,29 @@ export function App(): JSX.Element {
...
+      return;
+    }
+    const newList = [...contextHistory];
+    newList.push(context);
+    setContextList(newList);
+  }
+
   const queryResults = queries.map((q, i) => (
-    <QueryResult key={i} query={q}></QueryResult>
+    <QueryResult
```
shifucun, line 64: from testing, do you see the error case. Or if the component here can always invoke the callback in order?

shifucun (approved):

static/js/apps/nl_interface/app.tsx
beets: i haven't seen the error case. but wanted to prevent it from getting wildly out of sync..

beets:
thx!

static/js/apps/nl_interface/app.tsx
beets: i take it back. it happens when you change the embeddings build. in which case we shouldn't update the context?

shifucun: I think this works for now. We can revisit if errors do occur.

PR #1992, opened 2023-01-10
Use context to fetch previous stat var

shifucun:


server/lib/nl_page_config.py
```diff
@@ -107,61 +107,67 @@ def _single_place_multiple_var_timeline_block(svs, sv2name):
...
     category.blocks.append(block)
     for sv_key, spec in stat_var_spec_map.items():
       category.stat_var_spec[sv_key].CopyFrom(spec)
 
     # The siblings for the primary stat var
-    usable_sibiling_svs = [
-        x for x in data_spec.extended_sv_map[primary_sv]
-        if x in data_spec.main.svs
-    ]
-    if usable_sibiling_svs:
```
pradh, line 147: If I'm reading this right, is main_place_spec.svs not relied on?

pradh (approved):

server/lib/nl_page_config.py
shifucun: Right, in the bottom of server/lib/nl_data_spec.py, I noted the primary_sv does not depend on the place.

This is definitely not handled well now. For two queries with two places, they could have different svs with data, so the renderable sv list depend on the place. This is a super simplification now.

pradh: I see. Thanks.

PR #1993, opened 2023-01-10
Add a class to store the stat var entries used to compute page config

shifucun:


pradh (approved):
Could you add a meaningful title please :)

server/lib/nl_variable.py
```diff
@@ -13,11 +13,34 @@
...
 
 
+class EntryType(Enum):
+  SINGLE = 1
+  PEER = 2
+
+
+@dataclass
+class Entry:
+  # Only one of single and peers should be set.
```
pradh: s/single/source/?

server/lib/nl_variable.py
```diff
@@ -13,11 +13,36 @@
...
+
+
+@dataclass
+class SimpleEntry(Entry):
+  dcid: str
+
+
+@dataclass
+class PeerEntry(Entry):
+  # sources contains the sv/svg from the model. They are used to deduce peer
```
pradh, line 36: sv/svg/topic ?

pradh (approved):
Thanks!

server/lib/nl_variable.py
shifucun: I will update in next PR

PR #1995, opened 2023-01-10
[NL] Fix missing chart spec

juliawu:
Adds back json under "Chart Spec" in debugging info


beets (approved):
while you're in the tsx file, rename the label in the debug table to data spec too?

PR #1998, opened 2023-01-10
[nl] add search history

chejennifer:
screen-recording-_65_


beets (approved):
instead of re-issuing a query, i think we just want to scroll to the previous query. but i see where you are coming from, since this is interacting with the search box. let's get this in

chejennifer:
Ah, I see, ok, I can try that in a follow up PR

PR #2000, opened 2023-01-10
[NL] Improve NER place detection

juliawu:
Workaround for the BERT-NER model missing place names when the user doesn't use proper capitalization.


beets (approved):

juliawu:
Per offline discussion, going to expand NER changes in a separate PR

PR #2006, opened 2023-01-10
Ranking and Per capita ranking charts

shifucun:


rvguha (approved):

server/lib/nl_data_spec.py
```diff
@@ -87,7 +87,9 @@ def _highlight_svs(sv_df):
 
 
 def _sample_child_place(main_place_dcid, contained_place_type):
-  # Find child place, this is
+  # Find a sampled child place
+  if contained_place_type == "City":
```
jehangiramjad, line 91: why hardcoding this?

server/lib/nl_page_config.py
```diff
@@ -220,29 +220,63 @@ def build_page_config(detection: Detection, data_spec: DataSpec,
...
+            tile.ranking_tile_spec.show_lowest = True
+
+        tile.title = sv2name[
+            primary_sv] + ': rankings within ' + main_place_spec.name
+      else:
+        tile.type = subject_page_pb2.Tile.TileType.MAP
+        tile.title = "Per Capita " + sv2name[primary_sv] + ' (${date})'
+      category.stat_var_spec[sv_key].stat_var = primary_sv
+      category.stat_var_spec[sv_key].name = sv2name[primary_sv]
+      category.stat_var_spec[sv_key].denom = "Count_Person"
```
jehangiramjad, line 278: this will guaranteed work?

server/lib/nl_data_spec.py
jehangiramjad: Please add a comment in case this is just use for checks

jehangiramjad (approved):
Thanks. A couple of nits

PR #2007, opened 2023-01-10
Add a script to manually verify the flow of NLP features

Fructokinase:
Since the flow is fixed and we have fast merging PRs, it would be nice to have a way to verify the flow end to end with a script. This PR only verifies Palo Alto is correctly detected. The rest of the flow will be followed.

The goal is to test: query -> nl/data API response, with contexts in between.

### DIFF
--- a/server/services/nl.py
+++ b/server/services/nl.py
@@ -37,7 +37,7 @@ import re
 import string
 
 BUILDS = [
-    'demographics300',  #'uncurated3000', 
+    'demographics300',  #'uncurated3000',
     'demographics300-withpalmalternatives',
     'curatedJan2022',
     'us_filtered',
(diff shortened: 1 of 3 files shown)


server/services/nl.py
```diff
@@ -462,7 +462,7 @@ def query_classification(self, type_string: str,
   def detect_svs(self, query, embeddings_build):
     query_embeddings = self.model.encode([query])
     if embeddings_build not in self.dataset_embeddings_maps:
-      return ValueError(f'Embeddings Build: {embeddings_build} was not found.')
+      raise ValueError(f'Embeddings Build: {embeddings_build} was not found.')
```
Fructokinase, line 465: Some small fixes

jehangiramjad (approved):
Nice!

PR #2013, opened 2023-01-11
Pass along context in data_spec

shifucun:


server/lib/nl_data_spec.py
```diff
@@ -83,7 +86,7 @@ def _related_places(dcid):
 def _highlight_svs(sv_df):
   if sv_df.empty:
     return []
-  return sv_df[sv_df['CosineScore'] > 0.4]['SV'].values.tolist()
+  return sv_df[sv_df['CosineScore'] > 0.5]['SV'].values.tolist()
```
shifucun, line 89: @jehangiramjad I need to change this to 0.5 to make "Which cities are the worst in California" work. Currently it detects Count_Person_Urban with score > 0.4

pradh: We should probably validate it doesn't impact any of the working queries?

jehangiramjad: I can check that offline and let you know how it looks

beets (approved):

server/lib/nl_data_spec.py
shifucun: Thanks. If not working, we can revert this back. But maybe tune the model to let this query not returning high score vars.

jehangiramjad: Sure, sounds good. You can merge then. We can play with it. We an always put this condition on the output of the model's matches, if you want?

PR #2015, opened 2023-01-11
[nl] Add more test cases for Palo Alto flow

Fructokinase:
Things to verify:

Q3: "what about Boston" -> currently returns some random svs in the column tiles. 
Q4: "what are the worst cities in California" -> is the query correct? What statvars are supposed to be there?


server/test_nlp_manual.py
```diff
@@ -31,25 +39,162 @@ def _encoded_q(query: str, embedding: str = 'us_filtered'):
...
+      # Q1.
+      QueryTestCase(query='tell me about Palo Alto',
+                    config_expected=Config(metadata=Metadata(
+                        place_dcid=[PALO_ALTO_DCID]))),
+      # Q2.
+      QueryTestCase(
+          query='what about auto theft',
+          config_expected=Config(
+              categories=Categories(tiles=[
+                  Tile(stat_var_key='Count_CriminalActivities_MotorVehicleTheft'
```
shifucun, line 159: I feel testing the key here would be very subject to impl details in server. Any change would require updating this test. Let's see how well it works and adjust.

Fructokinase: > Any change would require updating this test
Right, the downside of this approach. Curious to see how well this works as well.

server/test_nlp_manual.py
```diff
@@ -31,25 +39,162 @@ def _encoded_q(query: str, embedding: str = 'us_filtered'):
...
+  metadata: Optional[Metadata] = None
+
+
+def _verify_categories(categories: Categories, got: Dict):
+  if categories.tiles:
+    # Check if all tile specs in this this has stat var keys
+    assert all(t.stat_var_key for t in categories.tiles)
+
+    stat_var_keys_got = set()
+    for tile in got[0]['blocks'][0]['columns'][0]['tiles']:
```
jehangiramjad, line 76: since this is being invoked a couple of times, could this be in a helper function? This may also end up being fragile due to the the fairly complex path to get tiles

server/test_nlp_manual.py
```diff
@@ -31,25 +40,136 @@ def _encoded_q(query: str, embedding: str = 'us_filtered'):
...
+                  'Count_CriminalActivities_MotorVehicleTheft',
+                  'Count_CriminalActivities_MotorVehicleTheft_pc'
+              ]),
+              # Check if the place is still Palo Alto.
+              metadata=Metadata(place_dcid=[PALO_ALTO_DCID]))),
+      # Q3.
+      QueryTestCase(
+          query='what about Boston',
+          config_expected=Config(
+              # TODO: Fix vars below, currently it returns 2 random svs.
```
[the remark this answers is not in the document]
jehangiramjad: This should be working now, right?

jehangiramjad (approved):
This is nice, thanks! I have a feeling we will need to have separate tests for the SV matches being returned by the server and then separate ones for testing the logic of how the places and SVs found are then transformed to charts. This particular end-to-end flow is useful! I think we'll need to build out the tests for the individual components before adding more tests here? What do you think?

server/test_nlp_manual.py
Fructokinase: oops, I forgot I had auto merge on. Will iterate on improving the logic here. Completely agree that this is very fragile.

PR #2018, opened 2023-01-11
Add scatterplot as response to correlation query

juliawu:
Screenshot 2023-01-10 at 8 49 03 PM


server/lib/nl_page_config.py
```diff
@@ -354,6 +354,38 @@ def build_page_config(detection: Detection, data_spec: DataSpec,
       category.stat_var_spec[sv_key].name = sv2name[primary_sv]
       category.stat_var_spec[sv_key].denom = "Count_Person"
 
+  # Render scatter plot if query asks for a correlation
+  # IMPORTANT: Right now this is fragile, as it will only
+  # use SVs from the most recent context.
+  # Also assumes that
+  #    - previous context had a primary_sv that was not "".
+  #    - data_spec.selected_svs is not empty
```
pradh, line 362: nit: could add this as a check? (for future)

server/lib/nl_page_config.py
```diff
@@ -354,6 +354,38 @@ def build_page_config(detection: Detection, data_spec: DataSpec,
...
+  elif classificationType == ClassificationType.CORRELATION:
+
+    # get first stat var from previous context
+    latest_context = context_history[-1]
+    sv_1 = latest_context['debug']['primary_sv']
+    sv_1_key = sv_1 + "_scatter"
+    sv_1_name = get_sv_name([sv_1])[sv_1]
```
pradh, line 369: nit: Might be good to call get_sv_name once with both SVs, for API batching

jehangiramjad (approved):

pradh (approved):
Nice!  Couple of nits, but lets get this in first.

server/lib/nl_page_config.py
```diff
@@ -350,6 +350,39 @@ def build_page_config(detection: Detection, data_spec: DataSpec):
       category.stat_var_spec[sv_key].name = sv2name[primary_sv]
       category.stat_var_spec[sv_key].denom = "Count_Person"
 
+  # Render scatter plot if query asks for a correlation
+  # IMPORTANT: Right now this is fragile, as it will only
+  # use SVs from the most recent context.
+  # Also assumes that
+  #    - previous context had a primary_sv that was not "".
```
beets, line 357: if previous context does not have a primary_sv, we should walk up the history chain to find one. there could be null contexts in between. for later also.

server/lib/nl_page_config.py
```diff
@@ -350,6 +350,39 @@ def build_page_config(detection: Detection, data_spec: DataSpec):
...
+  elif classificationType == ClassificationType.CORRELATION:
+
+    # get first stat var from previous context
+    sv_1 = data_spec.context_sv
+    # get second stat var from current
+    sv_2 = data_spec.selected_svs[0]
+    sv_names = get_sv_name([sv_1, sv_2])
+
+    sv_1_key = sv_1 + "_scatter"
+    sv_1_name = sv_names[sv_1]
```
beets, line 368: also for future - the key's don't need a prefix. the keys can be reused across tiles.

server/lib/nl_page_config.py
```diff
@@ -366,6 +366,52 @@ def build_page_config(detection: Detection, data_spec: DataSpec,
...
       category.stat_var_spec[sv_key].denom = "Count_Person"
 
+  # Render scatter plot if query asks for a correlation
+  # IMPORTANT: assumes that data_spec.selected_svs is not empty
+  # This might be fragile
+  # TODO: add check for data_spec.selected_svs 
+  elif classificationType == ClassificationType.CORRELATION:
+
+    # get first stat var from current data spec
+    sv_1 = data_spec.selected_svs[0]
```
beets, line 376: we should use highlight_svs so it passes the score filter

beets (approved):

chejennifer (approved):

chejennifer:
please run `./run_test.sh -f` to fix the lint errors

PR #2019, opened 2023-01-11
Fix context logic

shifucun:


server/lib/nl_data_spec.py
```diff
@@ -116,6 +117,14 @@ def compute(query_detection: Detection, context):
   main_place_name = places_detected.main_place.name
   main_place_type = places_detected.main_place.place_type
 
+  context_sv = None
+  context_place = None
+  if context and context['debug'] and context['debug']['data_spec']:
+    context_sv = context['debug']['data_spec']['primary_sv']
```
jehangiramjad, line 123: Shouldn't we check for existence of 'primary_sv' key?

server/lib/nl_data_spec.py
```diff
@@ -243,12 +222,10 @@ def compute(query_detection: Detection, context):
...
   # Get stat var from context
-  if not selected_svs:
-    if context and context['debug'] and context['debug']['data_spec']:
-      data_spec.primary_sv = context['debug']['data_spec']['primary_sv']
-      data_spec.use_context_sv = True
+  if data_spec.use_context_sv:
+    data_spec.primary_sv = context_sv
   else:
-    data_spec.primary_sv = selected_svs[0]
+    data_spec.primary_sv = data_spec.main_place_spec.svs[0]
```
jehangiramjad, line 228: what is there is nothing in svs list?

jehangiramjad (approved):
Thanks! A couple of quick comments

server/lib/nl_data_spec.py
shifucun: Done

server/lib/nl_data_spec.py
```diff
@@ -149,43 +158,17 @@ def compute(query_detection: Detection, context):
...
-
-  # This is a new try to extend svs to siblingins. This is to extend the
-  # stat vars "a little bit"
-
-  context_sv = None
-  if context and context['debug'] and context['debug']['data_spec']:
-    context_sv = context['debug']['data_spec']['primary_sv']
+  use_context_sv = False
+  if not selected_svs and context_sv:
+    selected_svs = [context_sv]
```
beets, line 164: should this be highlight_svs, so we use context when there are only poor matches?

beets (approved):

server/lib/nl_data_spec.py
shifucun: added check, otherwise not set it. I think the function of "primary_sv" especially in context is very confusing now..

shifucun (commented):
thanks!

server/lib/nl_data_spec.py
shifucun: from here downward, only selected_svs are used? so here would directly use context_sv for further processing.

PR #2028, opened 2023-01-11
Add more check for sv variables

shifucun:


beets (approved):
@juliawu has a similar fix - you'll need to update at merge time

PR #2029, opened 2023-01-11
[NL] filter out pc charts for temperature and precipitation sv's

beets:
image

### DIFF
--- a/server/lib/nl_page_config.py
+++ b/server/lib/nl_page_config.py
@@ -86,6 +86,7 @@ _SV_KEYWORDS_NO_PC = [
     "Snowfall",
     "Visibili",
     "WindSpeed",
+    "ConsecutiveDryDays",
 ]
 
 
--- a/server/lib/nl_page_config.py
+++ b/server/lib/nl_page_config.py
@@ -76,11 +76,18 @@ def get_sv_name(svs):
   return sv_name_map
 
 
+def _should_add_percapita(sv_dcid: str) -> bool:
+  if 'Temperature' in sv_dcid or 'Precipitation' in sv_dcid:
+    return False
+  return True
+
+
 def _single_place_single_var_timeline_block(sv_dcid, sv2name):
   """A column with two charts, main stat var and per capita"""
   block = subject_page_pb2.Block(title=sv2name[sv_dcid],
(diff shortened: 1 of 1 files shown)


server/lib/nl_page_config.py
```diff
@@ -169,8 +184,8 @@ def _topic_sv_blocks(category: subject_page_pb2.Category,
       sub_svs = extended_sv_map[sv]
       if not sub_svs:
         continue
-      sub_svs_exist = filter(lambda x: x in sv_exists_list, sub_svs)
-      if not sub_svs_exist:
+      sub_svs_exist = list(filter(lambda x: x in sv_exists_list, sub_svs))
+      if len(sub_svs_exist) == 0:
```
jehangiramjad, line 188: nit: can you do if sub_svs_list:

jehangiramjad (approved):

server/lib/nl_page_config.py
beets: done

server/lib/nl_page_config.py
```diff
@@ -341,37 +371,39 @@ def build_page_config(detection: Detection, data_spec: DataSpec,
...
-          # first check if "best" or "worst"
-          if RankingType.BEST in classifier.attributes.ranking_type:
-            tile.ranking_tile_spec.show_lowest = True
-          elif RankingType.WORST in classifier.attributes.ranking_type:
-            tile.ranking_tile_spec.show_highest = True
+      if _should_add_percapita(primary_sv):
+        tile = column.tiles.add()
+        sv_key = primary_sv + "_pc"
+        tile.stat_var_key.append(sv_key)
+        if classifier.type == ClassificationType.RANKING:
```
juliawu, line 378: I think moving `if classifier.type == ClassificationType.RANKING:` under `if _should_add_percapita(primary_sv):` means we won't ever show ranking charts for the weather data. Is this intended behavior?

beets: not for per capita. there's another block above for non-percapita

juliawu: gotcha. thanks for clarifying!

juliawu (approved):

PR #2031, opened 2023-01-11
[NL] ensure both scatter axes are comparable

beets:
specifically, if one is percentage / prevalence, do the same for the other 

image


shifucun (approved):
very nit: can we make the scatter plot squarish?

juliawu (approved):

beets:
harder to do without changing all chart heights (we use a constant for that.. sigh)

beets:
thanks for reviews!

PR #2032, opened 2023-01-11
Add error message to scatter plots when data is missing.

juliawu:


chejennifer (approved):

static/js/components/tiles/scatter_tile.tsx
```diff
@@ -78,19 +79,20 @@ export function ScatterTile(props: ScatterTilePropType): JSX.Element {
       props.statVarSpec,
       setRawData
     );
+    setErrorMsg("");
```
beets, line 82: this isn't needed unless we are refetching data

beets (approved):
please test that this doesn't lead to re-renders..

static/js/components/tiles/scatter_tile.tsx
juliawu: leaving this to catch the case if we ever refetch data.

PR #2033, opened 2023-01-11
[NL] double height of NL scatter tiles

beets:


static/js/components/subject_page/block.tsx
```diff
@@ -32,7 +32,7 @@ import {
   EventTypeSpec,
   TileConfig,
 } from "../../types/subject_page_proto_types";
-import { isNlInterface } from "../../utils/nl_interface_utils";
+import { isNlInterface, isNlInterface } from "../../utils/nl_interface_utils";
```
juliawu, line 35: why the double import?

shifucun, line 35: dup here?

shifucun (approved):
thx for update!

chejennifer (approved):

static/js/components/subject_page/block.tsx
beets: v weird! fixed

PR #2040, opened 2023-01-12
Check in vscode settings with debug instructions

shifucun:


shifucun:
Can you confirm if the .vscode settings are compatible with what you have locally?

chejennifer (approved):
this is very useful!

beets (approved):
thanks! i have very few settings in my vscode:

```
{
  "files.exclude": {
    "**/node_modules": true
  },
  "python.linting.pylintEnabled": true,
  "python.linting.enabled": true,
  "python.pythonPath": "/Users/cyin/google/website/.env/bin/python3",
  "go.inferGopath": false
}
```

perhaps we can skip adding settings.json? i'm afraid everyone's various settings will keep clobbering each other

.gitignore
```diff
@@ -12,7 +12,6 @@ server/dist/**
 **/node_modules/**
 
 # Local env.
-.vscode
```
beets, line 15: we could perhaps limit this only allow launch.json only (there are other vscode files which we don't want to check in yet..)

shifucun: only adding launch.json now

PR #2043, opened 2023-01-13
Organize tests director to match the main directory; Add missing tests back

shifucun:
* We should enforce the directory matching when adding new test files.
* A bunch of tests were missing because the `pytest` file glob only picks up files in direct subdirectories.
* Fixed all failed tests as they are added back now.


server/tests/lib/subject_page_config_content_test.py
```diff
@@ -47,7 +47,8 @@ def verify_tile(self, tile, stat_vars, msg, event_type_specs):
 
     # Non-chart tiles should have a title
     if not (tile.type == TileType.RANKING or tile.type == TileType.HIGHLIGHT or
-            tile.type == TileType.DESCRIPTION):
+            tile.type == TileType.DESCRIPTION or
+            tile.type == TileType.DISASTER_EVENT_MAP):
```
shifucun, line 51: @chejennifer can you confirm this

server/tests/services/discovery_test.py
```diff
@@ -27,45 +27,43 @@ def test_default_config(self):
...
-      assert get_service_url(
-          endpoint_path) == f'https://api-root{endpoint_path}'
+      assert get_service_url(endpoint_path) == f'http://api-root{endpoint_path}'
 
   def test_configure_endpoints_from_ingress_1(self):
     """Tests simple ingress configuration."""
     configure_endpoints_from_ingress('tests/test_data/ingress/test1.yaml')
 
-    assert get_service_url('/query') == 'https://query-host:8080/query'
+    assert get_service_url('/query') == 'http://query-host:8080/query'
```
shifucun, line 36: @Fructokinase can you confirm this

server/tests/lib/subject_page_config_content_test.py
chejennifer: yes, this is correct

chejennifer (approved):
thanks for the clean up!

server/tests/services/discovery_test.py
Fructokinase: yes, that's correct. I forgot to fix the tests. Thanks!

Fructokinase (approved):

jehangiramjad (approved):
Thanks!

PR #2045, opened 2023-01-13
[disasters] save information in url hash

chejennifer:
- update url hash with severity filter and date information

Added TODOs in the code for:
- if there are multiple disaster event map tiles on one page, need a way to specify in the url params which tile the param is for
- putting the whole severity filter object into the url makes the hash very long and ugly, might want a better way to save the information


static/js/utils/disaster_event_map_utils.tsx
```diff
@@ -391,3 +395,52 @@ export function onPointClicked(
...
+/**
+ * Gets the severity filters for a disaster event map using url params.
+ * @param eventTypeSpec the event type spec for the disaster event map
+ */
+export function getSeverityFilters(
+  eventTypeSpec: Record<string, EventTypeSpec>
+): Record<string, SeverityFilter> {
+  const urlParams = new URLSearchParams(window.location.hash.split("#")[1]);
+  let urlSeverityFilters = {};
+  if (urlParams.get(URL_HASH_PARAM_KEYS.SEVERITY_FILTER)) {
```
beets, line 430: nit: add the return value  to a local var, then test that var? instead of calling `.get` twice.

beets (approved):

PR #2046, opened 2023-01-14
[disasters] Shrink points on map

beets:
Hardcoded it for now -- couldn't get a good set of values that worked for all levels without disrupting the rest of our existing maps too much.

Also shrunk the severity filters, and tuned the upper and lower bounds after looking at our data.

image


chejennifer (approved):
thanks for the updates!

PR #2054, opened 2023-01-18
[NL] Display extended stat vars as comparison charts

shifucun:
* Remove unused codes: related places, other commented codes
* Use "extended_sv_maps" as how topic branch works. The major change is to find extended svs robustly by inspecting direct siblings or remote siblings.
* Simplify context by using `data_spec` directly

### DIFF
--- a/server/lib/nl/nl_variable.py
+++ b/server/lib/nl/nl_variable.py
@@ -148,6 +148,8 @@ def extend_svs(svs: Dict[str, List[str]]):
             continue
           if curr_sv_obj.pt != sv_obj.pt:
             continue
+          if curr_sv_obj.md != sv_obj.md:
+            continue
           if len(curr_sv_obj.pvs) != len(sv_obj.pvs):
             continue
           res[sv].append(sv_info['id'])
--- a/server/lib/nl/nl_variable.py
+++ b/server/lib/nl/nl_variable.py
@@ -14,11 +14,25 @@
 """Module for NL page variable"""
 
 from abc import ABC
-from dataclasses import dataclass
-from enum import Enum
-from typing import List
+from dataclasses import dataclass, field
+from typing import Dict, List
 
 import services.datacommons as dc
+
+
+@dataclass
+class SV:
+  mp: str = ''
+  st: str = ''
+  pt: str = ''
+  pvs: Dict[str, str] = field(default_factory=dict)
+
+
+@dataclass
+class SVG:
+  pt: str = ''
+  pvs: Dict[str, str] = field(default_factory=dict)
+  p: str = ''
 
 
 class Entry(ABC):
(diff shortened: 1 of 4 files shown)


server/lib/nl/nl_variable.py
```diff
@@ -43,38 +57,82 @@ class VariableStore:
...
-  return result
-
-
-def extend_svs(svs: List[str]):
-  """Extend all svs with siblings svs under the same svg.
+      res.pvs[k] = v
+  return res
+
+
+def parse_svg(svg_dcid: str) -> SVG:
```
pradh, line 76: please note that SVG DCID does not directly relate to the schema. We should probably set definitions for SVGs too in the cache?

server/lib/nl/nl_variable.py
```diff
@@ -43,38 +57,82 @@ class VariableStore:
...
+    if "-" in part:
+      p, v = part.split("-")
+      p_lower = p.replace(p[0], p[0].lower(), 1)
+      res.pvs[p_lower] = v
+    else:
+      res.p = part.replace(part[0], part[0].lower(), 1)
+  return res
+
+
+def extend_svs(svs: Dict[str, List[str]]):
```
pradh, line 93: Could you add some comments on the logic here?

server/lib/nl/nl_variable.py
```diff
@@ -43,38 +57,82 @@ class VariableStore:
...
+        for sv_info in item['info'].get('childStatVars', []):
+          curr_sv_obj = parse_sv(sv_info['definition'])
+          if curr_sv_obj.mp != sv_obj.mp:
+            continue
+          if curr_sv_obj.st != sv_obj.st:
+            continue
+          if curr_sv_obj.pt != sv_obj.pt:
+            continue
+          if len(curr_sv_obj.pvs) != len(sv_obj.pvs):
+            continue
```
pradh, line 133: should we consider measurementDenominator too?  I think measurementQualifier being different maybe ok for peers.

server/lib/nl/nl_variable.py
shifucun: have definition for SVG is very helpful!

server/lib/nl/nl_variable.py
shifucun: Done

shifucun (commented):
thanks for review!

server/lib/nl/nl_variable.py
pradh: Do we need to handle md in comparison in extend_svs() ?

pradh (approved):
Thanks!

server/lib/nl/nl_variable.py
shifucun: Done

PR #2056, opened 2023-01-18
[disasters] render AA2 maps for AA1 places

chejennifer:
- for the maps that are rendered, render AA2 for AA1 places
- in the breadcrumbs, prefer to show "State" or "County" over "AdministrativeArea" as the place type


beets (approved):
thx!

PR #2057, opened 2023-01-18
[disasters] add child places to sidebar

beets:
Rewrote the child-places handler to use the new V1 api -- the old handler was not used after we moved to adding more data to the landing page cache.

Also added some side-padding to the entire page (to match topic and place explorer)

image


server/routes/api/place.py
```diff
@@ -287,43 +287,49 @@ def child(dcid):
...
+    for t in types:
+      if t in wanted_types:
+        wanted_dcids.add(dcid)
         continue
-      if place_type in wanted_types and place_pop > 0:
+  wanted_types = list(wanted_types)
+
+  # Fetch population of child places
+  pop = {}
+  obs = dc.obs_point(wanted_dcids, ['Count_Person'])
```
shifucun, line 314: nit: Make 'Count_Person' as constant

shifucun (approved):
looks very nice!

server/routes/api/place.py
beets: done (there was already a constant to use!)

beets:
thanks for the review! i missed having the responses from the API be maps. having arrays of objects with the keys embedded meant i needed to build my own maps with the data.

static/js/components/disaster_dashboard/child_places.tsx
```diff
@@ -0,0 +1,69 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * Parent breadcrumbs for disaster dashboard.
```
chejennifer, line 18: update this comment?

static/js/utils/place_utils.ts
```diff
@@ -102,6 +106,23 @@ export function getParentPlacesPromise(
     .catch(() => []);
 }
 
+/**
+ * Used to get child places (filtered by wanted place type list.
```
chejennifer, line 110: add ")" at the end?

static/js/utils/place_utils.ts
```diff
@@ -102,6 +106,23 @@ export function getParentPlacesPromise(
...
+export function getChildPlacesPromise(
+  placeDcid: string
+): Promise<ChildPlacesByType> {
+  return axios
+    .get(`/api/place/child/${placeDcid}`)
+    .then((resp) => {
+      return resp.data;
+    })
+    .catch(() => {
+      return undefined;
```
chejennifer, line 122: maybe return `{}` so that return type is the same in the then and catch?

chejennifer (approved):
nice! just small comments

server/routes/api/place.py
```diff
@@ -287,43 +287,50 @@ def child(dcid):
...
+
+  place_types = dc.property_values(place_dcids, 'typeOf', True)
+  wanted_dcids = set()
+  for dcid, types in place_types.items():
+    for t in types:
+      if t in wanted_types:
+        wanted_dcids.add(dcid)
         continue
-      if place_type in wanted_types and place_pop > 0:
+  wanted_types = list(wanted_types)
```
chejennifer, line 310: wanted_types is already a list I believe

server/routes/api/place.py
```diff
@@ -287,43 +287,50 @@ def child(dcid):
...
+
+  # Fetch population of child places
+  pop = {}
+  obs = dc.obs_point(wanted_dcids, [POPULATION_DCID])
```
chejennifer, line 314: I believe obs_point takes a list of entities, not a set

server/routes/api/place.py
```diff
@@ -287,43 +287,50 @@ def child(dcid):
...
+  obs = obs.get('observationsByVariable', [])
+  if obs and obs[0]['variable'] == 'Count_Person':
```
chejennifer, line 316: nit: also use POPULATION_DCID here?

server/routes/api/place.py
```diff
@@ -287,43 +287,50 @@ def child(dcid):
...
+    obs = obs[0].get('observationsByEntity', [])
+    for p in obs:
+      v = p.get('pointsByFacet', [])
+      if v and len(v) > 0:
```
chejennifer, line 320: I don't think len(v) > 0 is necessary because empty array will be falsey

server/routes/api/place.py
```diff
@@ -287,43 +287,50 @@ def child(dcid):
...
+  result = collections.defaultdict(list)
+  for place_dcid in wanted_dcids:
+    for place_type in place_types[place_dcid]:
+      place_pop = pop.get(place_dcid, 0)
+      if place_pop > 0:
+        place_name = place_names.get(place_dcid, place_dcid)
         result[place_type].append({
-            'name': place.get('name', place['dcid']),
-            'dcid': place['dcid'],
+            'name': place_name[0] if len(place_name) > 0 else place_name,
```
chejennifer, line 332: looking at this line, does that mean place_name could either be a list or a string? probably don't want one variable to hold list and string?

chejennifer (commented):
Sorry, forgot to review the python file earlier

server/routes/api/place.py
beets: done. the wanted_types above was meant to convert wanted_dcids :(

server/routes/api/place.py
beets: yes, the return value from the name fetch sometimes returns a list of names or a single string :( so this normalizes this for the return value

static/js/utils/place_utils.ts
beets: done

static/js/utils/place_utils.ts
beets: ah good point. was fighting lint a little here. done

beets (commented):
thanks for the reviews!

server/routes/api/place.py
beets: done

server/routes/api/place.py
beets: done

server/routes/api/place.py
beets: updated to wanted_dcids

PR #2061, opened 2023-01-18
Update deployment config for enableModel

shifucun:


deploy/overlays/stanford-staging/kustomization.yaml
```diff
@@ -19,29 +19,29 @@ nameSuffix: -stanford-staging
 namespace: website
 
 resources:
-- ../../base
```
Fructokinase, line 22: is this intended?

Fructokinase: For my own learning, are these the same?

```
resources:
- A


resources:
  - A
```

Fructokinase (approved):

deploy/overlays/stanford-staging/kustomization.yaml
shifucun: Yes, which we used in other files. I get this formatting from vscode linter tool.

shifucun (commented):
thanks!

PR #2066, opened 2023-01-19
[disasters] read disaster event map data from saved jsons

chejennifer:
- add new flask endpoint that reads disaster event map data from saved jsons
- add url hash param that will read data from cache if "&cache=1" is set

### DIFF
--- a/server/configmodule.py
+++ b/server/configmodule.py
@@ -124,6 +124,7 @@ class LocalBaseConfig(Config):
   AI_CONFIG_PATH = os.path.abspath(
       os.path.join(os.path.curdir, '..', 'deploy/overlays/local/ai.yaml'))
   SCHEME = 'http'
+  GCS_BUCKET = 'datcom-website-autopush-resources'
 
 
 class LocalConfig(LocalBaseConfig):
--- a/server/lib/disaster_dashboard.py
+++ b/server/lib/disaster_dashboard.py
@@ -16,6 +16,7 @@
 from google.cloud import storage
 import json
 import re
+import logging
 
 EVENT_TYPES = [
     "FireEvent", "WildlandFireEvent", "WildfireEvent", "CycloneEvent",
@@ -58,6 +59,7 @@ def get_disaster_dashboard_data(gcs_bucket):
     file_name = re.sub('(?!^)([A-Z]+)', r'_\1', event_type).lower() + ".json"
     blob = bucket.get_blob(DISASTER_DATA_FOLDER + file_name)
     if not blob:
+      logging.info(f'file for {event_type} not found, skipping.')
       continue
     events_data = json.loads(blob.download_as_bytes())
     events_by_date = {}
--- a/server/lib/disaster_dashboard.py
+++ b/server/lib/disaster_dashboard.py
@@ -1,0 +1,73 @@
+# Copyright 2022 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Helper functions for getting disaster dashboard data for the app config"""
+
+from google.cloud import storage
+import json
+import re
+
+EVENT_TYPES = [
+    "FireEvent", "WildlandFireEvent", "WildfireEvent", "CycloneEvent",
+    "HurricaneTyphoonEvent", "HurricaneEvent", "TornadoEvent", "FloodEvent",
+    "DroughtEvent"
+]
+DISASTER_DATA_FOLDER = "disaster_dashboard/"
+
+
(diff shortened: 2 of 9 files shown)


server/lib/disaster_dashboard.py
```diff
@@ -0,0 +1,73 @@
...
+      }
+  """
+  result = {}
+  for event_type in EVENT_TYPES:
+    storage_client = storage.Client()
+    bucket = storage_client.get_bucket(gcs_bucket)
+    file_name = re.sub('(?!^)([A-Z]+)', r'_\1', event_type).lower() + ".json"
+    blob = bucket.get_blob(DISASTER_DATA_FOLDER + file_name)
+    if not blob:
+      continue
```
beets, line 61: nit: log a comment if there's a missing bucket?

server/routes/api/disaster_api.py
```diff
@@ -49,21 +50,134 @@ def event_date_range():
...
+  if place != EARTH_DCID and not place in event.get("affectedPlaces", []):
+    # event does not affect the place of interest
+    return False
+  if not filter_prop or not filter_prop in event:
+    # no props to filter by or event doesn't contain the prop information
+    return True
+  prop_val = event[filter_prop]
+  if len(prop_val) < len(filter_unit):
+    # value for the property to filter by is empty
+    return True
```
beets, line 67: we should return false here?

server/routes/api/disaster_api.py
```diff
@@ -49,21 +50,134 @@ def event_date_range():
...
+  val_string = prop_val[len(filter_unit):].strip()
+  try:
+    val = float(val_string)
```
beets, line 70: might need to `trim()` the string before float conversion?

server/routes/api/disaster_api.py
```diff
@@ -49,21 +50,134 @@ def event_date_range():
...
+    return val <= filter_upper_limit and val >= filter_lower_limit
+  except:
+    # can't parse the value to filter by into a float
+    return True
```
beets, line 74: Could we also log here so we can debug these as they come up?

server/routes/api/disaster_api.py
```diff
@@ -49,21 +50,134 @@ def event_date_range():
...
+  date = request.args.get('date', '')
+  if not date:
+    return "error: must provide a date field", 400
+  place = request.args.get('place', '')
+  if not place:
+    return "error: must provide a place field", 400
+  filter_prop = request.args.get('filterProp', '')
+  filter_unit = request.args.get('filterUnit', '')
+  filter_upper_limit = float(request.args.get('filterUpperLimit', '0'))
+  filter_lower_limit = float(request.args.get('filterLowerLimit', '0'))
```
beets, line 121: should we default to sys.float_info.max or  float("inf") and -float("inf")

server/routes/api/disaster_api.py
```diff
@@ -49,21 +50,134 @@ def event_date_range():
...
+      }
+      for eventKey in event.keys():
+        if eventKey in EVENT_POINT_KEYS:
+          continue
+        event_formatted["propVals"][eventKey] = {"vals": [event[eventKey]]}
+      event_points.append(event_formatted)
+  result = {}
+  if event_points:
+    result = {"eventCollection": {"events": event_points, "provenanceInfo": {}}}
+  return json.dumps(result), 200
```
beets, line 148: suggest returning this as Response(json.dumps(result), 200, "application/json") so you can format these in the inspector.

beets (commented):
feel free to fix these as follow ups in the interest of time.

server/configmodule.py
```diff
@@ -124,6 +124,7 @@ class LocalBaseConfig(Config):
   AI_CONFIG_PATH = os.path.abspath(
       os.path.join(os.path.curdir, '..', 'deploy/overlays/local/ai.yaml'))
   SCHEME = 'http'
+  GCS_BUCKET = 'datcom-website-autopush-resources'
```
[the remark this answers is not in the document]
chejennifer: the stanford instances already have their own GCS_BUCKET

server/routes/api/disaster_api.py
chejennifer: in mixer, if the prop isn't found or can't be parsed, we keep the event .. not sure if we want this logic to be different from mixer logic?

server/routes/api/disaster_api.py
chejennifer: I used `.strip()`, I think that's the same thing?

server/routes/api/disaster_api.py
chejennifer: we can, but defaulting to 0 makes the logic easier & the only time we will have to use default is if we get a filter_prop but no upper and lower limits which shouldn't happen .. I can make the change if you feel strongly

server/lib/disaster_dashboard.py
chejennifer: done

server/routes/api/disaster_api.py
chejennifer: done

server/routes/api/disaster_api.py
chejennifer: done

chejennifer:
thanks for the review! Made some of the suggested changes & left comments for the rest. If more changes are needed, I can address them in a followup

beets (approved):

beets:
thanks - sorry for late approval!

PR #2067, opened 2023-01-20
[Disaster Dashboard] Binning histogram to latest year

juliawu:
This PR:
* Adds a server endpoint that filters observation series to data for a specific year.
* Use that endpoint to have histogram tiles only show data from 2022.

This allows us to pre-bin the data used in the histogram tile to a specific year, as a quick workaround to the rendering issues seen when trying to show histograms for long historical timeseries.

Allowing for multiple types of binning (by decade, year, month) and either allowing this to be set in a config or dynamically chosen based on the data can come in a future PR.

Current:
Screenshot 2023-01-19 at 4 58 24 PM

Proposed:
Screenshot 2023-01-19 at 4 58 15 PM

### DIFF
--- a/server/routes/api/series.py
+++ b/server/routes/api/series.py
@@ -15,6 +15,7 @@
 from flask import Blueprint, request
 from cache import cache
 import services.datacommons as dc
+import logging
 
 # Define blueprint
 bp = Blueprint("series", __name__, url_prefix='/api/observations/series')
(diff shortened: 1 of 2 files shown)


server/routes/api/series.py
```diff
@@ -58,6 +59,58 @@ def series_within_core(parent_entity, child_type, variables, all_facets):
...
+    the specified year.
+  """
+  # Get raw series from mixer
+  data = series_core(entities, variables, False)
+
+  for stat_var in variables:
+    for location in data['data'][stat_var].keys():
+
+      # filter series to just observations from the selected year
+      series = data['data'][stat_var][location]['series']
```
beets, line 92: nit: use `.get` with defaults for most of these dereferences to be resilient to server errors or missing data.

server/routes/api/series.py
```diff
@@ -58,6 +59,58 @@ def series_within_core(parent_entity, child_type, variables, all_facets):
...
+  Note: This assumes the dates of the series are in 'YYYY-MM' format.
+
+  Args:
+    entities (list): DCIDs of entities to query
+    variables (list): DCIDs of variables to query
+    year (str): year in "YYYY" format to get observations for
+  
+  Returns:
+    JSON response from server, with series containing only observations from 
+    the specified year.
```
beets, line 83: describe the shape of the json?

server/routes/api/series.py
```diff
@@ -58,6 +59,58 @@ def series_within_core(parent_entity, child_type, variables, all_facets):
...
+        last_month = int(dates_with_data[-1][-2:])
+        months_to_fill = [str(mm).zfill(2) for mm in range(1, last_month + 1)]
+        for month in months_to_fill:
+          date = f"{year}-{month}"
+          if date not in dates_with_data:
+            pruned_series.append({'date': date, 'value': 0})
+        pruned_series.sort(key=lambda x: x['date'])
+
+      data['data'][stat_var][location]['series'] = pruned_series
+  return data
```
beets, line 111: return Response(data, 200, 'application/json'). No need to json.encode?

server/routes/api/series.py
```diff
@@ -58,6 +59,58 @@ def series_within_core(parent_entity, child_type, variables, all_facets):
...
+      # filter series to just observations from the selected year
+      series = data['data'][stat_var][location]['series']
+      pruned_series = []
+      dates_with_data = []
+      for obs in series:
+        if obs['date'][:4] == year:
+          pruned_series.append(obs)
+          dates_with_data.append(obs['date'])
+
+      if len(pruned_series) > 0:
```
beets, line 100: if there is no pruned series -- shouldn't we return an empty series instead of the raw mixer response?

server/routes/api/series.py
```diff
@@ -122,3 +175,27 @@ def series_within_all():
   if not variables:
     return 'error: must provide a `variables` field', 400
   return series_within_core(parent_entity, child_type, variables, True)
+
+
+@bp.route('/binned')
+@bp.route('/binned/<path:year>')
```
beets, line 181: please add a big todo here that we will not forget to update!

for the event maps, we are doing current date - 1 year for the "last year" filter.

beets (approved):
feel free to address these in follow ups so we can see how this code works out in the meantime.

juliawu:
Merging for now to see results on autopush, will address comments in followup PR

PR #2069, opened 2023-01-20
[Disasters] map updates

beets:
- Update lower thresholds for floods and fires to 25 km2 (brings us down from 80k to 40k events plotted for Earth)
- Move date filters down to legend

image


chejennifer (approved):
thanks for the updates

PR #2072, opened 2023-01-20
[Disasters] small updates to colors and titles

beets:
image


shifucun:
what do you think to use red for fire and yellow for drought. Blue for flood is nice now.

beets:
wdyt?
image

beets:
in some sense, droughts and fires co-occur, so the colors needs to have more contrast / farther apart on the color scale

beets:
i might switch back to the originals, if you don't mind?

shifucun:
> i might switch back to the originals, if you don't mind?

Nope, this is very subjective. As we launch, we could gather more thoughts too.

shifucun (approved):

PR #2073, opened 2023-01-20
add debugger config without NL models

beets:
for faster server starts..


shifucun (approved):
nice!

beets:
using these configs a lot this week! thanks for setting it up!

PR #2076, opened 2023-01-20
[Disaster Dashboard] Improve info card titles

juliawu:
Adds preprocessing to the event names to add the event type and handle non-user-friendly names.

Now instead of `DroughtEvent that started on 2022-08 within grid_1/37_-121`, the info card title is `Drought: Unnamed`.

And instead of `Ian`, the info card title is `Storm: Ian`

Also makes some quick fixes:
* Updates the "more info" link in info cards to point to a graph browser page, instead of the unfinished event pages.
* Remove extra `<p>` in category descriptions

### DIFF
--- a/static/js/components/tiles/disaster_event_map_info_card.tsx
+++ b/static/js/components/tiles/disaster_event_map_info_card.tsx
@@ -55,7 +55,7 @@ export function DisasterEventMapInfoCard(
             );
           })}
         <span>
-          <a href={`/event/${props.eventData.placeDcid}`}>More info</a>
+          <a href={`/browser/${props.eventData.placeDcid}`}>More info</a>
         </span>
       </div>
     </div>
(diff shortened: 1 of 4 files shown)


static/js/components/tiles/disaster_event_map_info_card.tsx
```diff
@@ -55,7 +55,7 @@ export function DisasterEventMapInfoCard(
             );
           })}
         <span>
-          <a href={`/event/${props.eventData.placeDcid}`}>More info</a>
+          <a href={`/browser/${props.eventData.placeDcid}`}>More info</a>
```
beets, line 58: I'd prefer we clean up the event pages, but this is ok for now

static/js/utils/disaster_event_map_utils.tsx
```diff
@@ -142,6 +143,36 @@ export function fetchDateList(
   });
 }
 
+/**
+ * Get a user-friendly info card title to use for each event.
+ *
+ * Titles follow the format `<EventType>: <Name>`.
+ *            Ex: "Storm: Ian"
+ * This also converts names that are "<EventType> that started on <Date> at
+ * <S2Cell>" into "Unnamed" for the title.
```
beets, line 152: Not a fan of using Unnamed. We should include place name or something better. If we go with this, all events of certain types will be unnamed.

Do we also need the event type in the title? We can add the type as an annotation in the card.

This should go in a event --> name util that we can use across tiles (or pages) so the events show up consistently.

beets (commented):
some comments while in development..

PR #2078, opened 2023-01-21
[disasters] add event ranking tile

beets:
Many todos:
- add more properties
- add embed
- create better names
- hook up with url params for dates and cache usage

Also removed the SV maps - waiting on event pt maps for each disaster.

image


beets:
This also doesn't remove dups -- I'm seeing a lot of events with the same dcid returned multiple times. Might be in the json generation script @chejennifer 

image

static/js/components/tiles/top_event_tile.tsx
```diff
@@ -0,0 +1,166 @@
...
+      ref={chartContainer}
+    >
+      {!_.isEmpty(topEvents) && (
+        <div className="ranking-unit-container">
+          <div className="ranking-list">
+            <h4>{props.title}</h4>
+            <table>
+              <tbody>
+                {topEvents.map((event, i) => {
+                  console.log(event);
```
chejennifer, line 79: remove this console.log?

chejennifer (approved):
thanks! just nits, I'll merge this in so I can work on todos for this tile

chejennifer:
> This also doesn't remove dups -- I'm seeing a lot of events with the same dcid returned multiple times. Might be in the json generation script @chejennifer
> 
> image

Good catch, I updated the jsons, should be better now

PR #2081, opened 2023-01-21
[disasters] various updates

chejennifer:
- event ranking tile embed
Screenshot 2023-01-20 at 11 19 01 PM

- event ranking tile use url params
- updates to allow multiple disaster event maps on one page (config changes to show the multiple event maps will be in a separate PR in case we don't want to include those changes because it slows down the page)
Screenshot 2023-01-20 at 11 20 08 PM


beets (approved):
Thanks for adding in the missing features in ranking the too!

PR #2083, opened 2023-01-23
[disasters] add delay to severity filter inputs

chejennifer:
screen-recording-_67_


beets (approved):
thanks - this is usually also done onBlur (user is done with the input, so cancel the timer and send the request), but it seems reactstrap input might not support it. this is a good workaround that works for all cases.

chejennifer:
Thanks for the review! Updated so that the filter also updates when user clicks enter or clicks away from the input box

beets (approved):
thanks for the update!

PR #2084, opened 2023-01-23
[disaster] small fixes

beets:
- Update link from map info cards to browser
- Move maps above histograms for each disaster section
- Remove unused drought sv's


chejennifer (approved):
thanks for the fixes!

PR #2085, opened 2023-01-23
[disaster] add category for wet bulb; drop histograms

beets:
also drops histograms while we fix data display consistency issues

image

image


chejennifer (approved):
thanks for the change!

PR #2091, opened 2023-01-24
[NL Next] Frontend changes (1/5)

pradh:
Notes:

1. Calls a different data API if this is an NL Next URL
2. Changes to support a slightly different response structure
3. Context history is handled by the server, and client just saves/returns what server sent.


shifucun (approved):
very precise change!

static/js/apps/nl_interface/app.tsx
```diff
@@ -58,12 +64,20 @@ export function App(): JSX.Element {
     setContextList(newList);
   }
 
+  function getContextHistory(i: number) {
```
chejennifer, line 67: nit: we usually want to specify the type that a function is returning

so `function getContextHistory(i: number): <return type> {}`

static/js/apps/nl_interface/query_result.tsx
```diff
@@ -66,8 +67,13 @@ export const QueryResult = memo(function QueryResult(
   function fetchData(query: string, build: string): void {
     setIsLoading(true);
     console.log("context:", props.query, props.contextHistory);
+    const is_nl_next = isNLInterfaceNext();
+
+    // The API endpoint is different NL Next version.
```
chejennifer, line 72: nit: ...is different for NL Next version

chejennifer (approved):
looks good, just small nits

static/js/apps/nl_interface/app.tsx
```diff
@@ -44,6 +45,11 @@ export function App(): JSX.Element {
   }, [queries]);
 
   function addContext(context: any, idx: number) {
+    if (isNLInterfaceNext()) {
+      // The context list is handled entirely by the server.
```
beets, line 49: earlier queries could be resent if the app refreshes (sometimes happens when it's low on memory etc) which is why it was always sliced to just the available context when the query was first sent. is the server also handling the index of each query?

beets (approved):

static/js/apps/nl_interface/app.tsx
beets: also, to preserve each "question/answer" component so it can be memoized, the context for each pair should be fixed. i think the append should stay the same here, but for `getContextHistory`, it should be a `get` instead of `slice`.

PR #2092, opened 2023-01-24
[NL Next] Utterance data structure (2/5)

pradh:
Introduces the main Utterance data structure ([link](https://docs.google.com/document/d/10yCcs2YJuFrh1oL3Zdw02CVAv4QgG7JaiSqQbH67F6w/edit)).

Also includes Utterance <-> Dict transformation for saving/restoring context from client. 

TODO: add tests


server/lib/nl/nl_utterance.py
```diff
@@ -0,0 +1,226 @@
...
+@dataclass
+class Utterance:
+  # Linked list of past utterances
+  prev_utterance: Utterance
+  # Unmodified user-issued query
+  query: str
+  # Result of classification
+  detection: Detection
+  # A characterization of the query
+  query_type: ClassificationType
```
shifucun, line 78: is the query_type relevant to the context? For example "tell me about san jose" could be an "overview" query or "query stat var from context"?

server/lib/nl/nl_utterance.py
```diff
@@ -0,0 +1,226 @@
...
+  # TODO: Fill this up
+  answerPlaces: List[str]
+
+
+#
+# Helper functions for serializing/deserializing Utterance
+#
+
+
+def _save_places(places: List[Place]) -> List[Dict]:
```
shifucun, line 100: should these helper functions in other util files (as they are independent of the utterance class)

shifucun, line 100: why not operate directly on the class but convert them to dict?

server/lib/nl/nl_utterance.py
```diff
@@ -0,0 +1,226 @@
...
+  for c in classifications:
+    cdict = {}
+    cdict['type'] = c.type
+
+    if isinstance(c.attributes, ContainedInClassificationAttributes):
+      cip = c.attributes.contained_in_place_type
+      if isinstance(cip, ContainedInPlaceType):
+        cdict['contained_in_place_type'] = cip.value
+      else:
+        cdict['contained_in_place_type'] = cip
```
beets, line 132: nit: curious what else this could be if not a ContainedinPlaceType instance?

server/lib/nl/nl_utterance.py
```diff
@@ -0,0 +1,226 @@
...
+      attributes = RankingClassificationAttributes(
+          ranking_type=[RankingType(r) for r in cdict['ranking_type']],
+          ranking_trigger_words=[])
+    classifications.append(
+        NLClassifier(type=ClassificationType(cdict['type']),
+                     attributes=attributes))
+  return classifications
+
+
+def _save_charts(charts: List[ChartSpec]) -> List[Dict]:
```
beets, line 159: instead of load/save (which is odd to see in a request handler) -- perhaps we can spell out the conversion types? something like chart_spec_to_dict

server/lib/nl/nl_utterance.py
```diff
@@ -0,0 +1,226 @@
...
+    uttr_dicts.append(udict)
+    u = u.prev_utterance
+    cnt += 1
+
+  return uttr_dicts
+
+
+# Given a list of dicts previously saved by `save_utterance()`, loads
+# them into Utterance structures and returns the latest one.
+def load_utterance(uttr_dicts: List[Dict]) -> Utterance:
```
beets, line 206: i see - this could be a constructor for Utterance from a dict (if Utterance were a regular class, rather than just a dataclass). add a todo for this?

beets (approved):
nice, though perhaps this should be a regular class?

pradh (commented):
Thanks for the quick reviews!

PR #2093, opened 2023-01-24
[NL Next] Logic to construct Utterance (3/5)

pradh:
This builds on the [Utterance data structure PR](https://github.com/datacommonsorg/website/pull/2092).

Given classification results and past Utterance history, constructs a new Utterance, which includes a list of chart specs.

Most query types (SIMPLE, CONTAINED-IN, RANKING) follow a common processing function `populate_charts()`, with custom callbacks.   Only CORRELATION uses a different handler.

TODO: add unit tests

PR #2094, opened 2023-01-24
[NL Next] Logic to generate chart config proto from Utterance (4/5)

pradh:
This builds on the [Utterance data structure PR](https://github.com/datacommonsorg/website/pull/2092).

Given an Utterance data structure (containing ranked chart specs), this PR generates the chart config proto.  It is largely based on `nl_page_config.py`.

The changes in `nl_utils.py` are shared between this PR and [Utterance building logic](https://github.com/datacommonsorg/website/pull/2093).

TODO: Add unit tests


beets (approved):
cool!

some nits:
- would be nice to have utils for the chart titles (eg "  - Per Capita") so we can easily update when we tweak the titles
- use a const for Count_Person?

PR #2096, opened 2023-01-24
[disaster] make heat and cold event data available

chejennifer:
- made heat and cold event data available to be used, but NOT added to config
- don't read disaster data from GCS bucket if running lite server

### DIFF
--- a/server/lib/disaster_dashboard.py
+++ b/server/lib/disaster_dashboard.py
@@ -21,7 +21,7 @@ import logging
 EVENT_TYPES = [
     "FireEvent", "WildlandFireEvent", "WildfireEvent", "CycloneEvent",
     "HurricaneTyphoonEvent", "HurricaneEvent", "TornadoEvent", "FloodEvent",
-    "DroughtEvent", "WetBulbTemperatureEvent"
+    "DroughtEvent", "WetBulbTemperatureEvent", "ColdEvent", "HeatEvent"
 ]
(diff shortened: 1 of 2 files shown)


shifucun (approved):

server/lib/disaster_dashboard.py
```diff
@@ -21,7 +21,7 @@
 EVENT_TYPES = [
     "FireEvent", "WildlandFireEvent", "WildfireEvent", "CycloneEvent",
     "HurricaneTyphoonEvent", "HurricaneEvent", "TornadoEvent", "FloodEvent",
-    "DroughtEvent", "WetBulbTemperatureEvent"
+    "DroughtEvent", "WetBulbTemperatureEvent", "ColdEvent", "HeatEvent"
```
beets, line 24: I believe these should be "HeatTemperatureEvent" and "ColdTemperatureEvent", per latest email from  @jehangiramjad 

https://github.com/datacommonsorg/schema/blob/main/core/natural_phenomena.mcf#L425-L435

jehangiramjad: Yes please :)

jehangiramjad: Also @ajaits as FYI

chejennifer: PR to update this: https://github.com/datacommonsorg/website/pull/2098

PR #2097, opened 2023-01-24
Add new map component to browser pages

juliawu:
This PR adds a map component to the browser pages. 

A map is shown:

(A)When there is a "dcid: latLong/..." entry in the table, the lat/long is shown on a map in the cell. 

(B) The node has both a latitude and longitude property.

This should help make debugging locations easier :)

Examples:
Screenshot 2023-01-27 at 8 49 27 AM

Screenshot 2023-01-27 at 8 49 06 AM

This new component replaces the old component used by the place pages as well, but should not change the place pages' appearance:

Screenshot 2023-01-27 at 8 51 59 AM

### DIFF
--- a/static/js/components/google_map.tsx
+++ b/static/js/components/google_map.tsx
@@ -57,11 +57,11 @@ interface GoogleMapPropType {
 }
 
 interface GoogleMapStateType {
-  // parsed response of propvals api
+  // coordinates where marker should be drawn on the map.
   markerLocation: GoogleMapCoordinates;
-  // parsed response of mapinfo api
+  // response of mapinfo api with KML coordinates to draw.
   mapInfo: MapInfoResponse;
-  // Whether there is data for map to render
+  // Whether there is data for map to render.
   shouldShowMap: boolean;
 }
 


juliawu:
Question for the group: Should I be placing the new map component somewhere else?

I've left this component under the browser folder for now, because it's currently only used on the browser pages. I know @chejennifer wrote up a doc for restructuring the website repo and eventually this component could be used in the event pages, so it might be worth extracting out?

chejennifer:
> Question for the group: Should I be placing the new map component somewhere else?
> 
> I've left this component under the browser folder for now, because it's currently only used on the browser pages. I know @chejennifer wrote up a doc for restructuring the website repo and eventually this component could be used in the event pages, so it might be worth extracting out?

I think ideally, we could:
-  reuse this map component instead of creating a new one: https://github.com/datacommonsorg/website/blob/master/static/js/place/map.tsx
- move the component to the /components folder
- add the ability to put a marker on the map just with logic. Maybe in the props either take in a place dcid or a latlong dcid & depending on what props the component receives, draw the map accordingly

wonder what everyone else thinks?

juliawu:
> I think ideally, we could:
> -  reuse this map component instead of creating a new one: https://github.com/datacommonsorg/website/blob/master/static/js/place/map.tsx
> - move the component to the [shared components folder](https://github.com/datacommonsorg/website/tree/master/static/js/components) 
> - add the ability to put a marker on the map just with logic. Maybe in the props either take in a place dcid or a latlong dcid & depending on what props the component receives, draw the map accordingly
> 
> wonder what everyone else thinks?

I like the idea of extracting out and reusing the map component from the place pages! Let me make those changes.

beets:
sharing sgtm - don't forget that we should also expect multiple of these on a page (so don't use `div id`, use `div class`)

beets:
hi @juliawu - is this ready for review? it's still in draft but looks like you've already made a new component

juliawu:
> hi @juliawu - is this ready for review? it's still in draft but looks like you've already made a new component

Yes. Was having issues restarting the server yesterday, and wanted to hold off until I could properly test all the changes. Ready for review now!

static/js/place/overview.tsx
```diff
@@ -36,7 +36,7 @@ class Overview extends React.Component<OverviewPropType> {
       <section className="factoid col-12">
         <div className="row">
           <div className="col-12 col-md-4">
-            <Map dcid={this.props.dcid}></Map>
```
beets, line 39: should we also delete the Map component, if it's no longer used?

static/css/place/place.scss
```diff
@@ -198,11 +198,6 @@ h3 {
   padding: 0.25rem;
 }
 
-#map-container {
```
beets, line 201: does this need to be converted to .map-container instead of deleting?

static/js/browser/browser_page.tsx
```diff
@@ -113,6 +114,7 @@ export class BrowserPage extends React.Component<
             </h2>
           </>
         )}
+        <GoogleMap dcid={this.props.dcid}></GoogleMap>
```
beets, line 117: this needs a bit more top-padding or margin (should match what's below it on the page)

static/js/components/google_map.tsx
```diff
@@ -0,0 +1,257 @@
...
+  /**
+   * DCID of the place/event to show a map for.
+   */
+  dcid: string;
+}
+
+interface GoogleMapStateType {
+  /**
+   * Stores responses from api calls.
+   */
```
beets, line 40: since this refers to the entire type (and not the first prop), goes above the interface.

static/js/components/google_map.tsx
```diff
@@ -0,0 +1,257 @@
...
+
+const DEFAULT_MAP_ZOOM = 4;
+
+// number of digits after decimal point in LatLong/<lat>_<long> dcids
+const LAT_LONG_PRECISION = 5;
+
+interface GoogleMapPropType {
+  /**
+   * DCID of the place/event to show a map for.
+   */
```
beets, line 33: simple comment format sufficient here for properties (`//`, for brevity)

static/js/components/google_map.tsx
```diff
@@ -0,0 +1,257 @@
...
+   * Stores responses from api calls.
+   */
+  markerLocation: GoogleMapCoordinates;
+  mapInfo: MapInfoResponse;
+}
+
+interface GoogleMapCoordinates {
+  /**
+   * Lat/Long coordinates, stored in a form the Google maps API expects.
+   */
```
beets, line 48: ditto above.

static/js/components/google_map.tsx
```diff
@@ -0,0 +1,257 @@
...
+  lat: number;
+  lng: number;
+}
+
+interface MapInfoResponse {
+  /**
+   * Response format of /api/place/mapinfo/.
+   */
```
beets, line 56: ditto

static/js/components/google_map.tsx
```diff
@@ -0,0 +1,257 @@
...
+ * @param dcid latLong DCID to convert
+ * @returns a {lat: latitude, lng: longitude} object for the google maps API
+ */
+function convertLatLongDcid(dcid: string): GoogleMapCoordinates {
+  if (dcid.startsWith("latLong/") && dcid.includes("_")) {
+    const coordinateStrings = dcid.slice("latLong/".length).split("_");
+    const coords = coordinateStrings.map((coordString) =>
+      // Add decimal point to parsed latitude and longitude numbers
+      parseFloat(
+        coordString.slice(0, -LAT_LONG_PRECISION) +
```
beets, line 76: please add lots of comments for these (and best if we can reference the backend equivalent which needs to stay in sync). or vice versa, depending on where the code lives.

static/js/components/google_map.tsx
```diff
@@ -0,0 +1,257 @@
...
+  coordinateSequenceSet: google.maps.LatLng[][];
+}
+
+/**
+ * Converts a DCID of the form latLong/<latitude>_<longitude> to
+ * lat/long coordinates used by the google maps API.
+ * @param dcid latLong DCID to convert
+ * @returns a {lat: latitude, lng: longitude} object for the google maps API
+ */
+function convertLatLongDcid(dcid: string): GoogleMapCoordinates {
```
beets, line 70: this function is ripe for a unit test :)

static/js/components/google_map.tsx
```diff
@@ -0,0 +1,257 @@
...
+  }
+
+  private fetchData(): void {
+    // Skip API calls if we can get lat/long from DCID itself
+    if (
+      this.props.dcid.startsWith("latLong/") &&
+      this.props.dcid.includes("_")
+    ) {
+      this.setState({
+        markerLocation: convertLatLongDcid(this.props.dcid),
```
beets, line 207: check return value first, since it could be null

static/js/components/google_map.tsx
```diff
@@ -0,0 +1,257 @@
...
+    return <div className="map-container" ref={this.div}></div>;
+  }
+
+  componentDidMount(): void {
+    this.fetchData();
+  }
+
+  componentDidUpdate(): void {
+    // Initialize map
+    const map = initMap(this.div.current);
```
beets, line 188: only call this if the value is not null?

static/js/components/google_map.tsx
```diff
@@ -0,0 +1,257 @@
...
+}
+
+/**
+ * Initialize a google map widget.
+ * @param container the element to render the map in
+ */
+function initMap(container: HTMLDivElement): google.maps.Map {
+  // set height for map to be visible
+  container.style.height = "100%";
+  container.style.minHeight = "160px";
```
beets, line 93: can we do this with css?

beets (commented):
thanks! some comments below

static/js/place/overview.tsx
chejennifer: agree with deleting Map component if it's no longer used!

static/js/components/google_map.tsx
```diff
@@ -0,0 +1,257 @@
...
+
+interface GoogleMapStateType {
+  /**
+   * Stores responses from api calls.
+   */
+  markerLocation: GoogleMapCoordinates;
+  mapInfo: MapInfoResponse;
+}
+
+interface GoogleMapCoordinates {
```
chejennifer, line 45: nit: define an interface before it is used

& also MapInfoResponse

static/js/components/google_map.tsx
chejennifer: Instead of having to make sure the client code stays in sync with the backend code, should we just always read the property values of "latitude" and "longitude" .. I don't think fetching that data would significantly impact speed or memory usage?

static/js/components/google_map.tsx
```diff
@@ -0,0 +1,257 @@
...
+  return new google.maps.Map(container, mapOptions);
+}
+
+/**
+ * Draw polygons on map given KML coordinates for each polygon.
+ * @param state current component state with KML Coordinates stored in it
+ * @param map map to draw polygon in
+ */
+function drawKmlCoordinates(
+  state: GoogleMapStateType,
```
chejennifer, line 113: nit: instead of taking the whole state as a param, just need `mapInfo: MapInfoResponse`

static/js/components/google_map.tsx
```diff
@@ -0,0 +1,257 @@
...
+        .get(`/api/browser/propvals/latitude/${this.props.dcid}`)
+        .then((resp) => resp.data);
+      const longitudePromise = axios
+        .get(`/api/browser/propvals/longitude/${this.props.dcid}`)
+        .then((resp) => resp.data);
+
+      Promise.all([polygonPromise, latitudePromise, longitudePromise]).then(
+        ([mapInfo, latitudes, longitudes]) => {
+          // Update state with mapInfo
+          if (mapInfo && Object.keys(mapInfo).length > 0) {
```
chejennifer, line 226: can just check `!_.isEmpty(mapInfo)`

static/js/components/google_map.tsx
```diff
@@ -0,0 +1,257 @@
...
+            const latValues = latitudes.values.out.map(
+              (latValue) => latValue.value
+            );
+            const longValues = longitudes.values.out.map(
+              (longValue) => longValue.value
+            );
+            //TODO (juliawu): Update logic to use highest precision lat/long if
+            //                multiple values are provided
+            const coordinates = {
+              lat: parseFloat(latValues[0]),
```
chejennifer, line 248: don't need to create latValues and longValues constants - can just use `latitudes.values.out[0].value` and `longitudes.values.out[0].value`?

static/js/components/google_map.tsx
```diff
@@ -0,0 +1,257 @@
...
+
+      // Get lat/long from properties
+      const latitudePromise = axios
+        .get(`/api/browser/propvals/latitude/${this.props.dcid}`)
+        .then((resp) => resp.data);
+      const longitudePromise = axios
+        .get(`/api/browser/propvals/longitude/${this.props.dcid}`)
+        .then((resp) => resp.data);
+
+      Promise.all([polygonPromise, latitudePromise, longitudePromise]).then(
```
chejennifer, line 223: please include a `.catch()` when using Promises to handle errors

chejennifer (commented):
this is nice!

beets:
if this is ready for review, please reply to comments

static/js/components/google_map.tsx
```diff
@@ -0,0 +1,226 @@
...
+    if (this.state.shouldShowMap) {
+      // initialize Map
+      const map = initMap(this.div.current);
+
+      // plot KML Coordinates
+      if (Object.values(this.state.mapInfo).every((val) => val !== null)) {
+        drawKmlCoordinates(this.state.mapInfo, map);
+      }
+      // plot marker
+      else if (this.state.markerLocation.lat && this.state.markerLocation.lng) {
```
beets, line 172: else always starts after the previous closing parens

static/js/components/google_map.tsx
```diff
@@ -0,0 +1,226 @@
...
+        }
+
+        // Update state with lat/long if both are present
+        if (
+          latitudes.values &&
+          longitudes.values &&
+          !_.isEmpty(latitudes.values.out) &&
+          !_.isEmpty(longitudes.values.out)
+        ) {
+          //TODO (juliawu): Update logic to use highest precision lat/long if
```
beets, line 212: super nit: always a space after `//`. please see style guide for commenting format (e.g following line starts from after `// `.

static/js/components/google_map.tsx
```diff
@@ -0,0 +1,226 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * Component for embedding a map from Google Maps
```
beets, line 18: please document here that kml coordinates outlines from /mapinfo are preferred over lat/lng pins. that was quite unclear from the outset.

static/js/components/google_map.tsx
juliawu: Fixed!

static/js/components/google_map.tsx
juliawu: Fixed!

static/js/components/google_map.tsx
juliawu: Good point! Added more documentation, and also clarified in the inline comments.

static/js/place/overview.tsx
juliawu: Update: deleted the old map component.

static/css/place/place.scss
juliawu: Updated to class

static/js/browser/browser_page.tsx
juliawu: Added more padding via CSS

static/js/components/google_map.tsx
juliawu: Updated comment formats.

static/js/components/google_map.tsx
juliawu: Thanks for the pointer! Updated the comment styling. (This interface got implemented, switched to inline comments about each item)

static/js/components/google_map.tsx
juliawu: Reordered the interfaces!

static/js/components/google_map.tsx
juliawu: done.

static/js/components/google_map.tsx
juliawu: done.

static/js/components/google_map.tsx
juliawu: Went with chejennifer's suggestion of always reading lat/long for property values. It feels cleaner, and avoids potential mismatch with backend code. This section was removed.

static/js/components/google_map.tsx
juliawu: Good pointer! I ended up removing this though (see discussion about staying in sync with backend)

static/js/components/google_map.tsx
juliawu: Moved to CSS!

static/js/components/google_map.tsx
juliawu: Good catch! Updated.

static/js/components/google_map.tsx
juliawu: Good catch. Ended up removing this section. (See discussion on sync with backend)

static/js/components/google_map.tsx
juliawu: Thanks for the pointer! Updated.

static/js/components/google_map.tsx
juliawu: Good catch! That's much cleaner. Updated.

static/js/components/google_map.tsx
```diff
@@ -0,0 +1,236 @@
...
+  lng: number;
+}
+
+interface GoogleMapPropType {
+  // DCID of the place/event to show a map for.
+  dcid: string;
+}
+
+interface GoogleMapStateType {
+  // parsed response of propvals api
```
chejennifer, line 60: nit: it is more useful to indicate what the field actually is/how it is used. i.e., this is the coordinates where a marker should be drawn on the map.

chejennifer (approved):
thanks for the updates!

static/js/components/google_map.tsx
juliawu: thanks for the pointer! Done.

PR #2099, opened 2023-01-25
[disasters] update top event tile when hash updates

chejennifer:
- update top event tile to listen to hash and update if there are changes to the options used to get data
- move getting date range from individual tile to part of the data fetch util function because we should be getting the same date range for both disaster event map tile and top event tile

screen-recording-_68_


static/js/components/tiles/disaster_event_map_tile.tsx
```diff
@@ -367,16 +357,12 @@ export function DisasterEventMapTile(
...
-      Object.values(eventTypeSpec),
-      placeInfo.selectedPlace.dcid,
-      dateRange,
-      severityFilters,
-      useCache
+      dataOptions.eventTypeSpecs,
+      dataOptions.place,
+      dataOptions.selectedDate,
+      dataOptions.severityFilters,
+      dataOptions.useCache
```
beets, line 365: should we pass in the dataOptions object instead?

beets (approved):
thanks for the fix!

static/js/components/tiles/disaster_event_map_tile.tsx
chejennifer: that makes sense, updated!

PR #2102, opened 2023-01-25
[disasters] remove v0 code

chejennifer:
- remove disaster dashboard v0 code
- moved components that are specific to disaster dashboard into the disaster dashboard app folder


beets (approved):
thanks for the cleanup!

PR #2103, opened 2023-01-26
[disasters] render parent map for places with no enclosedPlaceType

chejennifer:
- for disaster event map tile, when a place doesn't have a specified contained place type (e.g., counties), render the parent map and highlight the selected place (like map explorer)

Screenshot 2023-01-25 at 4 25 29 PM

Screenshot 2023-01-25 at 4 25 41 PM


beets (approved):
nice!

PR #2104, opened 2023-01-26
[disasters] by default, use cache data

chejennifer:
- use param "json=1" to use json data, otherwise, use cache data

### DIFF
--- a/static/js/utils/disaster_event_map_utils.tsx
+++ b/static/js/utils/disaster_event_map_utils.tsx
@@ -544,7 +544,8 @@ export function getSeverityFilters(
 }
 
 /**
- * Gets whether or not to use cache for data fetch.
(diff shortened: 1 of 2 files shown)


static/js/utils/disaster_event_map_utils.tsx
```diff
@@ -523,12 +523,12 @@ export function getSeverityFilters(
 }
 
 /**
- * Gets whether or not to use cache from url params.
+ * Gets whether or not to use cache for data fetch.
```
beets, line 526: nit: maybe mention that the cache here is JSON, to link the two.

beets (approved):
thanks - please let the group know when it's in autopush

static/js/utils/disaster_event_map_utils.tsx
chejennifer: updated the comment

PR #2106, opened 2023-01-26
Add a Flask server for NL

shifucun:
This will be another container deployed together with website and mixer in the same Pod.

All the NL related code and endpoints should be moved to it.


server/services/datacommons.py
```diff
@@ -327,6 +327,13 @@ def get_event_collection_date(event_type, affected_place):
   })
 
 
+def nl_helloworld():
+  """fetch nl hello world data.
+  """
+  url = cfg.NL_ROOT
```
shifucun, line 333: @Fructokinase what would be a good way to put this in discovery.py? Also I wonder if we should complete the work to have mixer service (and now with nl service)

nl_server/routes.py
```diff
@@ -0,0 +1,24 @@
...
+# limitations under the License.
+
+import json
+
+from flask import Blueprint
+
+bp = Blueprint('main', __name__, url_prefix='/')
+
+
+@bp.route('/')
```
shifucun, line 22: @jehangiramjad You can update this function, rename this file or add new files for the NL endpoints

jehangiramjad: got it, thanks!

nl_server/requirements.txt
```diff
@@ -0,0 +1,3 @@
+Flask==1.1.4
```
jehangiramjad, line 1: We will certainly need this (I can also add them later):

torch==1.12.1
pandas==1.3.5
spacy==3.4.4
en_core_web_md @ https://github.com/explosion/spacy-models/releases/download/en_core_web_md-3.4.1/en_core_web_md-3.4.1-py3-none-any.whl

jehangiramjad (approved):
Many thanks! I'll let @Fructokinase reply to the other comment

nl_server/requirements.txt
shifucun: ok, sg, we can add those as you migrate them

nl_server/routes.py
beets: my understanding is that this is strictly an API server to host bigger embeddings? in that case, we should perhaps set up with a /api endpoint.

beets (approved):

nl_server/routes.py
shifucun: Yep. @jehangiramjad will need to update those and we can make sure they are /api/...

shifucun (commented):
thanks for review!

PR #2109, opened 2023-01-27
[disaster] add extreme heat and cold

beets:
@jehangiramjad @ajaits looks like the ColdTemperatureEvent values come without a unit, unlike the HeatTemperatureEvents.
Please also let me know if those are the right thresholds to use (especially the upper limits).


server/config/disaster_dashboard/Earth.textproto
```diff
@@ -127,6 +127,35 @@ metadata {
...
+    key: "heat"
+    value {
+      id: "heat"
+      name: "Extreme Heat"
+      event_type_dcids: "HeatTemperatureEvent"
+      color: "#f01212"
+      default_severity_filter: {
+        prop: "differenceTemperature",
+        unit: "Celsius"
+        upper_limit: 25,
```
jehangiramjad, line 140: Why do we need to have an upper limit? Is the assumption that this would indicate a bug?

server/config/disaster_dashboard/Earth.textproto
```diff
@@ -394,3 +425,99 @@ categories {
...
+      # tiles {
+      #   type: HISTOGRAM
+      #   title: "Total area affected by extreme heat events last year"
+      #   stat_var_key: "heat_area"
+      # }
+    }
+    columns {
+      tiles {
+        type: TOP_EVENT
+        title: "Most severe heat differences in the last year"
```
jehangiramjad, line 453: nit: suggestion for title string:

"Most severe heat events in the last year" OR "Greatest differences in maximum temperatures compared to historical monthly averages in the last year"

server/config/disaster_dashboard/Earth.textproto
```diff
@@ -394,3 +425,99 @@ categories {
...
+      # tiles {
+      #   type: HISTOGRAM
+      #   title: "Total area affected by extreme cold events last year"
+      #   stat_var_key: "cold_area"
+      # }
+    }
+    columns {
+      tiles {
+        type: TOP_EVENT
+        title: "Most severe cold differences in the last year"
```
jehangiramjad, line 501: nit: similar suggestion for title of these cold events as made about for heat events

jehangiramjad (approved):
Thanks. Just a few questions/suggestions.

server/config/disaster_dashboard/Earth.textproto
beets: I think it's mostly because we are not handling the case where it's unspecified well. I'll add as a TODO for us to handle it.

jehangiramjad: That sounds Ok. Could we increase this limit so that we aren't excluding any amazingly severe events? I suggest an upper_limit of 35? Ditto for the cold events?

chejennifer (approved):
thanks!

server/config/disaster_dashboard/Earth.textproto
beets: thanks, used the shorter title, and used the longer title as description text.

server/config/disaster_dashboard/Earth.textproto
beets: done

server/config/disaster_dashboard/Earth.textproto
beets: done

beets:
thanks for the review!

PR #2113, opened 2023-01-27
[disasters] proto changes for creating new block type

chejennifer:


server/config/subject_page.proto
```diff
@@ -103,20 +103,33 @@ message TopEventTileSpec {
...
     SCATTER = 4;
+    // allowed in BlockTypes: TYPE_NONE
     BIVARIATE = 5;
+    // allowed in BlockTypes: TYPE_NONE
     RANKING = 6;
+    // allowed in BlockTypes: TYPE_NONE
     HIGHLIGHT = 7;
+    // allowed in BlockTypes: TYPE_NONE
     DESCRIPTION = 8;
+    // allowed in BlockTypes: DISASTER_EVENT
```
beets, line 126: let's group the ones for DISASTER_EVENT together at the bottom? they needn't be arranged by value.

beets (approved):
cool!

server/config/subject_page.proto
chejennifer: done

PR #2115, opened 2023-01-27
[disasters] client changes to add new disaster block type

chejennifer:
- new components: DisasterEventBlock, BlockContainer, Column
- extracted functions that are shared between block types into subject_page_utils
- move date selection & severity filtering from DisasterEventMapTile to DisasterEventBlock
- move disaster event data fetch out of DisasterEventMapTile, TopEventTile and into DisasterEventBlock

screen-recording-_70_

TODO: update some component/file names & move files into different folders. I will do this in a followup to hopefully make it easier to review


chejennifer:
thanks for the review!

PR #2117, opened 2023-01-28
Add version config for embeddings file

shifucun:
Instead of hard coding the embedding as a constant, it is set in a config file now.

Also performed some code clean up, including logging, autopush handling


deploy/base/deployment.yaml
```diff
@@ -48,6 +48,9 @@ spec:
         - name: ai-config
           configMap:
             name: ai-config
+        - name: model-config
```
jehangiramjad, line 51: In this case, we are associating the word "model" with just the "embeddings". That can be a bit confusing. There are many other "models" that are or could be used. For example, we already have the Named-Entity Recognition "model" for place recognition. So to avoid confusion, I propose we take EITHER of the two suggestions below:

1. Keep this called "model-config" but update the "file" in model.yaml to be called "embeddings_file".
2. Update "model-config" and "model.yaml" to be called "embeddings-config" and "embeddings.yaml".

I think (1) makes most sense because then we can put any/all model related settings/constants in that file and it will be understood with embeddings are just one type of model.

nl_server/loader.py
```diff
@@ -0,0 +1,47 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import logging
+import os
+
+GCS_BUCKET = 'datcom-nl-models'
```
jehangiramjad, line 18: Shouldn't this also be provided as a config (because what if we decide to change this folder or rename it some day)? Could this be set in the model.yaml?

nl_server/lib/nl_embeddings.py
```diff
@@ -63,12 +64,9 @@ def __init__(self) -> None:
   def _download_embeddings(self):
     storage_client = storage.Client()
     bucket = storage_client.bucket(bucket_name=GCS_BUCKET)
-    try:
-      blob = bucket.get_blob(FOLDER + EMBEDDINGS_FILENAME)
-      # Download
-      blob.download_to_filename(os.path.join(TEMP_DIR, EMBEDDINGS_FILENAME))
-    except Exception as e:
-      logging.info(e)
+    blob = bucket.get_blob(self.gcs_folder + self.embeddings_file)
```
jehangiramjad, line 67: Are we sure that this will never ever fail? For example, what if the file on the server is somehow deleted? Won't it make sense to keep this under the try/catch?

deploy/base/model.yaml
```diff
@@ -0,0 +1 @@
+file: embeddings_us_filtered_2023_1_26_22_15_41.csv
```
jehangiramjad, line 1: Who updates this file to the latest version?

jehangiramjad (commented):
Thanks!

deploy/base/deployment.yaml
shifucun: Very nice suggestions, updated based on 1

deploy/base/model.yaml
shifucun: This is like the bigtable version, so oncall would send a PR to update it as a release step.

nl_server/lib/nl_embeddings.py
shifucun: Previously, it logs the failure but the program would continue. Since this is used when server starts, it is better to let it throw instead of silently proceed? Since if no model is loaded, the server should not start.

nl_server/loader.py
shifucun: The config holds values that changes often. Since this would be fairly stable (if not changed). We can keep it as a constant here.

shifucun (commented):
thanks for the review!

jehangiramjad (approved):
Thanks!

nl_server/loader.py
jehangiramjad: ok, sgtm

nl_server/lib/nl_embeddings.py
jehangiramjad: i see, ok sounds good.

deploy/base/model.yaml
jehangiramjad: ok

PR #2118, opened 2023-01-29
[NL Server] simplify folder and files

shifucun:
This is a light weighted server, so we can put all files under root folder to simplify the structure.

Also remove "nl_" prefix in files as they are in scope of "nl_server"


jehangiramjad (approved):
LGTM

PR #2123, opened 2023-01-30
[disaster] add unit back for cold events

beets:
also updated api handling to handle cases where upper or lower limits are unspecified.

### DIFF
--- a/server/routes/api/disaster_api.py
+++ b/server/routes/api/disaster_api.py
@@ -204,8 +204,10 @@ def event_data():
     return "error: must provide a place field", 400
   filter_prop = request.args.get('filterProp', '')
   filter_unit = request.args.get('filterUnit', '')
-  filter_upper_limit = float(request.args.get('filterUpperLimit', None))
-  filter_lower_limit = float(request.args.get('filterLowerLimit', None))
+  req_upper = request.args.get('filterUpperLimit', None)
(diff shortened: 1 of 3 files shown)


server/routes/api/disaster_api.py
```diff
@@ -130,8 +130,8 @@ def json_event_data():
     return "error: must provide a place field", 400
   filter_prop = request.args.get('filterProp', '')
   filter_unit = request.args.get('filterUnit', '')
-  filter_upper_limit = float(request.args.get('filterUpperLimit', '0'))
-  filter_lower_limit = float(request.args.get('filterLowerLimit', '0'))
+  filter_upper_limit = float(request.args.get('filterUpperLimit', float("inf")))
+  filter_lower_limit = float(request.args.get('filterLowerLimit', -float("inf")))
```
chejennifer, line 134: not sure if there would be problems with `float(-float("inf"))`

server/tests/routes/api/disaster_test.py
```diff
@@ -301,3 +301,19 @@ def test_with_filter(self):
           }
       }
       assert json.loads(response.data) == expected
+
+  def test_with_filter_no_upper_limit(self):
```
chejennifer, line 305: nit: ...no_lower_limit

chejennifer (approved):
thanks

PR #2127, opened 2023-01-31
Initial version of COMPARISON fulfillment

pradh:
This also re-uses the populate_charts() framework, but extends it to support multiple places.

Imminent TODO (hopefully tonight):  unit tests for nl_data_spec_next.py

Example: [auto theft in sunnyvale] -> [compare with palo alto]

Screenshot 2023-01-30 at 4 18 56 PM

### DIFF
--- a/server/lib/nl/nl_data_spec_next.py
+++ b/server/lib/nl/nl_data_spec_next.py
@@ -498,6 +498,8 @@ def _places_from_context(uttr: Utterance) -> List[Place]:
   return ans
 
 
+# Returns a list of place lists.  Each inner list is a candidate for place comparison.
+# The outer list is ordered from most prefered candidates to least.
 def _places_for_comparison_from_context(uttr: Utterance) -> List[List[Place]]:
   ans = []
   first_uttr_count = 0
--- a/server/lib/nl/nl_data_spec_next.py
+++ b/server/lib/nl/nl_data_spec_next.py
@@ -19,8 +19,8 @@ import logging
 
 from lib.nl.nl_detection import ClassificationType, Detection, \
   ContainedInPlaceType, ContainedInClassificationAttributes, \
-    RankingType, RankingClassificationAttributes, Place, \
-      ClassificationAttributes
+    RankingType, RankingClassificationAttributes, ComparisonClassificationAttributes, \
+    Place, ClassificationAttributes
 from lib.nl import nl_variable, nl_topic
 from lib.nl.nl_utils import is_sv, is_topic, \
   sv_existence_for_places, get_sample_child_places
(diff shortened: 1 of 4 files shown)


server/lib/nl/nl_data_spec_next.py
```diff
@@ -449,6 +498,30 @@ def _places_from_context(uttr: Utterance) -> List[Place]:
...
+      second_uttr_count = first_uttr_count + 1
+      while (second and second_uttr_count < CTX_LOOKBACK_LIMIT + 1):
+        # For the first place, add all combinations of the second.
+        if second.places:
+          ans.append(second.places + first.places)
+        second = second.prev_utterance
+        second_uttr_count = second_uttr_count + 1
+    elif len(first.places) > 1:
+      # Found multiple places in a single utterance.  Use this.
+      ans.append(first.places)
```
shifucun, line 519: do you want to break here?

shifucun (approved):
extending to multiple places is cool!

server/lib/nl/nl_data_spec_next.py
pradh: We are returning multiple lists so the caller will try each related-places-list in order until successful.  So we don't want to break here?

shifucun: Can you add some comments or example for the choice here. For example for a query sequence of "tell me about san jose", "how is auto theft in fremont", "compare with palo alto". Will. the last one use all the places above (fremont, san jose)?

pradh: Clarified, please take a look.

server/lib/nl/nl_data_spec_next.py
```diff
@@ -498,6 +498,26 @@ def _places_from_context(uttr: Utterance) -> List[Place]:
...
+# 1. If an utterance has multiple places, then that is a candidate list.
+# 2. If an utterance has a single place, then we walk up prior utterances until
+#    we find an utterance with at least one place.  The candidate list is
+#    the former single place plus the latter from prior utterance.
+#
+# For example, suppose the query sequence is as follows:
+#   [tell me about san jose] -> [how is auto theft in fremont] -> [compare with palo alto]
+#
+# In this case, the ordered candidates will be:
+#   [[palo alto, fremont], [palo alto, san jose], [fremont, san jose]]
```
shifucun, line 514: So we will show comparison for all the 3 combinations? Not sure if that will be too much, esp if there are more context places before.

shifucun:
Thanks for adding examples!

server/lib/nl/nl_data_spec_next.py
pradh: No, the caller will loop through each combo, and stop when the first one generated any charts.  (This is part of the general pattern of trying to show something when data doesn't exist for certain places)

shifucun: Ok, so here will fill as much as possible, now I see the pattern..

PR #2129, opened 2023-01-31
Add startupProbe for NL server; Update all the Kubernetes probers

shifucun:
NL server restarts a few times now due to the lack of startup probe.

Removes initialDelaySeconds=10 as it is default value.

### DIFF
--- a/deploy/base/deployment.yaml
+++ b/deploy/base/deployment.yaml
@@ -66,13 +66,15 @@ spec:
               path: /healthz
               port: 8080
             failureThreshold: 30
-            periodSeconds: 10
+
           readinessProbe:
             httpGet:
               path: /healthz
               port: 8080
-            failureThreshold: 1
-            periodSeconds: 10
+          livenessProbe:
+            httpGet:
+              path: /healthz
+              port: 8080
           resources:
             limits:
               memory: "3G"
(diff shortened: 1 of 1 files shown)


deploy/base/deployment.yaml
```diff
@@ -66,13 +66,15 @@ spec:
               path: /healthz
               port: 8080
             failureThreshold: 30
-            periodSeconds: 10
```
Fructokinase, line 69: Is this deleted because the default is 10?

deploy/base/deployment.yaml
```diff
@@ -190,17 +189,19 @@ spec:
           volumeMounts:
             - name: model-config
               mountPath: /datacommons/model
+          startupProbe:
+            httpGet:
+              path: /healthz
```
Fructokinase, line 194: Just double checking, I'm assuming /healthz exists in the nl server.

Fructokinase (approved):

deploy/base/deployment.yaml
shifucun: Yes, default is 10

deploy/base/deployment.yaml
shifucun: Yes, it exists already

PR #2130, opened 2023-01-31
Change large SV comparisons to use bar charts instead of timeline

pradh:
Screenshot 2023-01-30 at 5 30 38 PM

### DIFF
--- a/server/lib/nl/nl_data_spec_next.py
+++ b/server/lib/nl/nl_data_spec_next.py
@@ -210,11 +210,14 @@ def _populate_simple_cb(state: PopulateState, chart_vars: ChartVars,
                         places: List[Place],
                         chart_origin: ChartOriginType) -> bool:
   if len(chart_vars.svs) <= _MAX_VARS_PER_CHART:
-    _add_chart_to_utterance(ChartType.TIMELINE_CHART, state, chart_vars, places,
-                            chart_origin)
+    # For fewer SVs, comparing trends over time is nicer.
+    chart_type = ChartType.TIMELINE_CHART
   else:
-    _add_chart_to_utterance(ChartType.BAR_CHART, state, chart_vars, places,
-                            chart_origin)
+    # When there are too many, comparing latest values is better
+    # (than, say, breaking it into multiple timeline charts)
+    chart_type = ChartType.BAR_CHART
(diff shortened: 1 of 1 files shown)


server/lib/nl/nl_data_spec_next.py
```diff
@@ -209,8 +209,12 @@ def populate_simple(uttr: Utterance) -> bool:
 def _populate_simple_cb(state: PopulateState, chart_vars: ChartVars,
                         places: List[Place],
                         chart_origin: ChartOriginType) -> bool:
-  return _add_chart_to_utterance(ChartType.TIMELINE_CHART, state, chart_vars,
-                                 places, chart_origin)
+  if len(chart_vars.svs) <= _MAX_VARS_PER_CHART:
```
shifucun, line 212: nit: use if condition to set the chart_type? That makes it clear the only difference is chart type

server/lib/nl/nl_data_spec_next.py
```diff
@@ -638,25 +642,12 @@ def _add_chart_to_utterance(chart_type: ChartType, state: PopulateState,
...
-  start_index = 0
-  while start_index < len(chart_vars.svs):
-    l = min(len(chart_vars.svs) - start_index, _MAX_VARS_PER_CHART)
-    ch = ChartSpec(chart_type=chart_type,
-                   svs=chart_vars.svs[start_index:start_index + l],
-                   places=places,
-                   utterance=state.uttr,
-                   attr=attr)
-    start_index += l
-    state.uttr.chartCandidates.append(ch)
```
shifucun, line 659: To confirm, before there could be several charts, now they are all merged into one?

shifucun (approved):

server/lib/nl/nl_data_spec_next.py
pradh: Good idea.

server/lib/nl/nl_data_spec_next.py
pradh: Correct.  Yes, before we breakup SVs across multiple time-line charts. Guha wanted them all in one long Bar chart.

pradh (commented):
Thanks!

PR #2132, opened 2023-01-31
[NL Next] Break down fulfillment into modules by classification type

shifucun:
Tested locally for demo scripts, matching what is now in autopush.


server/lib/nl/fulfillment/comparison.py
```diff
@@ -0,0 +1,59 @@
...
+  state = PopulateState(uttr=uttr,
+                        main_cb=_populate_comparison_cb,
+                        fallback_cb=_fallback_comparison_cb)
+  for places_to_compare in context.places_for_comparison_from_context(uttr):
+    if populate_charts_for_places(state, places_to_compare):
+      return True
+  return False
+
+
+def _populate_comparison_cb(state: PopulateState, chart_vars: ChartVars,
```
pradh, line 43: super nit: can these just be _populate_cb and _fallback_cb everywhere?

pradh (approved):
Very nice refactor!  Thanks!

server/lib/nl/fulfillment/comparison.py
shifucun: good idea!

PR #2135, opened 2023-01-31
[NL server] Updating the embeddings

jehangiramjad:


deploy/base/model.yaml
```diff
@@ -1 +1 @@
-embeddings_file: embeddings_us_filtered_2023_1_26_22_15_41.csv
+embeddings_file: embeddings_us_filtered_2023_1_31_17_20_20.csv
```
shifucun, line 1: nit: could change the format to 2023_01_...

PR #2136, opened 2023-01-31
[NL Next] Tests for page config generation from Utterance

pradh:


server/lib/nl/nl_page_config_next.py
```diff
@@ -18,7 +18,7 @@
   SubjectPageConfig, Tile
 from lib.nl.nl_utterance import Utterance, ChartType, ChartSpec
 from lib.nl.nl_detection import Place, RankingType
-from lib.nl.nl_utils import get_sv_name
+from lib.nl import nl_utils
```
shifucun, line 21: Shall we remove all the "nl_" file prefix under nl folder?

server/tests/lib/nl/nl_page_config_next_test.py
```diff
@@ -0,0 +1,487 @@
...
+from parameterized import parameterized
+from typing import Dict
+from unittest.mock import patch
+
+from lib.nl import nl_utils, nl_utterance, nl_page_config_next
+from config.subject_page_pb2 import SubjectPageConfig
+from tests.lib.nl.test_utterance import PLACE_ONLY_UTTR, SIMPLE_UTTR, SIMPLE_WITH_SV_EXT_UTTR, \
+  SIMPLE_WITH_TOPIC_UTTR, COMPARISON_UTTR, CONTAINED_IN_UTTR, CORRELATION_UTTR, RANKING_UTTR
+
+PLACE_ONLY_CONFIG = """
```
shifucun, line 27: not necessary for this PR, is it better to have individual textproto test file for each of these?

shifucun (approved):
thanks for adding tests!

server/lib/nl/nl_page_config_next.py
pradh: Let me do in follow-up PR?  I want to rename nl_data_spec to fulfillment too.

server/tests/lib/nl/nl_page_config_next_test.py
pradh: that would be nice, yes

PR #2138, opened 2023-01-31
[NL] Remove "nl_" prefix from a bunch of files

pradh:
I tested both nl and nlnext using the two Demo Scripts.


shifucun (approved):
nice!!

PR #2139, opened 2023-01-31
[disasters] add flag to enable json loading

beets:
to load json configs - use `./run_server.sh -d`

we'll need to add this to our deployment configs if we want to load the json's elsewhere (or we can only check for the flag in local mode). thoughts?


beets:
Made the change so the flag is only usable locally (and otherwise is loaded by default on other servers).

shifucun (approved):
thanks for adding this!

chejennifer (approved):
thanks!

PR #2140, opened 2023-02-01
Fix python import sorting and add lint checker (use google style)

shifucun:


pradh (approved):
Nice!  Will this run when we run `run_server.sh -f` ?

shifucun:
> Nice! Will this run when we run `run_server.sh -f` ?

Yes, it will run.

beets (approved):
nice - better warn the team of this new feature.

shifucun:
> nice - better warn the team of this new feature.

thanks, I will ping in chat group once this is merged

PR #2141, opened 2023-02-01
[disasters] one geojson call per subject page

chejennifer:
- save geojson data to subject page context & read the context when rendering a disaster event map tile, allowing us to only have to fetch geojson data once


static/js/components/subject_page/data_context.ts
```diff
@@ -0,0 +1,31 @@
...
+ * Data context to hold data that is used across multiple components for the
+ * whole page.
+ */
+
+export interface DataContextType {
+  // TODO (chejennifer): use this in map tile
+  geoJsonData: GeoJsonData;
+}
+
+export const DataContext = createContext({} as DataContextType);
```
beets, line 31: nice

static/js/components/subject_page/data_context.ts
```diff
@@ -0,0 +1,31 @@
...
+
+import { GeoJsonData } from "../../chart/types";
+
+/**
+ * Data context to hold data that is used across multiple components for the
+ * whole page.
+ */
+
+export interface DataContextType {
+  // TODO (chejennifer): use this in map tile
```
beets, line 27: update todo to a comment?

beets (approved):
v cool!

static/js/components/subject_page/data_context.ts
chejennifer: this is already a comment?

PR #2145, opened 2023-02-01
[NL Next] Implement ranking across SVs.  Rendered in bar chart.

pradh:
The approach is described in the comments in ranking.py.

Examples below.

Screenshot 2023-02-01 at 1 07 53 AM

Screenshot 2023-02-01 at 1 08 07 AM


server/lib/nl/fulfillment/context.py
```diff
@@ -110,15 +110,21 @@ def classifications_of_type_from_context(
...
     prev = prev.prev_utterance
     prev_uttr_count = prev_uttr_count + 1
   return result
+
+
+def classifications_of_type_from_utterance(
+    uttr: Utterance,
+    ctype: ClassificationType) -> List[ClassificationAttributes]:
+  result = []
+  for cl in uttr.classifications:
```
shifucun, line 127: nit: return [cl for cl in uttr.classifications if x.type == type]

server/lib/nl/fulfillment/ranking.py
```diff
@@ -14,53 +14,73 @@
...
 from lib.nl.fulfillment.base import ChartVars
 from lib.nl.fulfillment.base import populate_charts
 from lib.nl.fulfillment.base import PopulateState
-from lib.nl.fulfillment.context import classifications_of_type_from_context
 from lib.nl.utterance import ChartOriginType
 from lib.nl.utterance import ChartType
 from lib.nl.utterance import Utterance
 
 
+#
```
shifucun, line 34: I wonder if we should have two rankings fulfillment, which can eliminates the condition check and make it straightforward?

server/lib/util.py
```diff
@@ -120,3 +120,30 @@ def parse_date(date_string):
     return datetime.strptime(date_string, "%Y-%m-%d")
   else:
     raise ValueError("Invalid date: %s", date_string)
+
+
+# Returns a compact version of observation point API results
+def compact_point(point_resp, all_facets):
```
shifucun, line 126: why need to move the function here?

server/lib/nl/fulfillment/ranking.py
pradh: I'm a bit divided on that.  Because the fork between ranking for places vs. ranking for svs is subtle and doing that upstream might mean we add more logic in common code.  In any case, Guha wants us to do multi-query-type handling, and as part of that we can reconsider?

server/lib/nl/fulfillment/ranking.py
```diff
@@ -14,53 +14,73 @@
...
+# This handles both ranking across places and ranking across SVS.
+#
+# * For ranking across places, we should detect a ranking and contained-in
+#   classification in the current utterance.  For example, [counties with most rainfall],
+#   assuming california is in the context.
+# * For ranking across SVS, we should detect a ranking, but not contained-in
+#   classification in the current utterance.  In the callback, we will also
+#   check that the SVs are part of a peer group (only those are comparable!).
+#   For example, [most grown agricultural things], again assuming california
+#   is in the context.
```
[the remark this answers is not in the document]
pradh: So, the current logic relies on peer groups (SVPG) defined within topics.  So it implicitly assumes these have been curated well to have comparable units.

I suppose in future, we might expand to "computed peers" too (from SV expansion), and in that scenario we'll certainly need unit checks.  I've added todo.

server/lib/nl/fulfillment/ranking.py
shifucun: sgtm

server/lib/util.py
pradh: Because this PR uses it from server/lib/nl/utils.py

server/lib/nl/fulfillment/context.py
pradh: Good idea!

pradh (commented):
Thanks for the review!

shifucun (approved):

PR #2153, opened 2023-02-02
[disasters] fix european maps

chejennifer:
- update European places to use Euronuts as contained place types instead of admin areas. Fixes map problems for a lot of the european places


beets (approved):
thansk for the fix!

PR #2154, opened 2023-02-02
Make peer variable checks more resilient to custom DC SVG hierarchy

pradh:
Sanity checked with [male population in california] which pulls in peer vars.


n-h-diaz (approved):
thank you 😃

PR #2155, opened 2023-02-02
[NL Next] Improve comparison classification

juliawu:
This PR:

* Adds more heuristics to the comparison classifier
* Adds comparison classification results to the debug output

Example screenshot:
Screenshot 2023-02-01 at 6 04 17 PM

### DIFF
--- a/server/lib/nl/constants.py
+++ b/server/lib/nl/constants.py
@@ -192,6 +192,14 @@ QUERY_CLASSIFICATION_HEURISTICS: Dict[str, Union[List[str], Dict[
             "Best": ["best",],
             "Worst": ["worst",],
         },
+        "Comparison": [
+            "compare(s|d)?",
+            "comparison",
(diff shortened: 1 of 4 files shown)


server/lib/nl/constants.py
```diff
@@ -192,6 +192,14 @@
             "Best": ["best",],
             "Worst": ["worst",],
         },
+        "Comparison": [
+            "compare(s|d)?",
+            "comparison",
+            "(is|has|have)( a| the)? \w+er",
+            # WARNING: These will conflate with Correlation
```
shifucun, line 199: do we return both comparison and correlation?

shifucun (approved):

server/lib/nl/constants.py
pradh: Can you please validate that the Jan scripts work with this change?

juliawu: Done! Both scripts still work.

PR #2156, opened 2023-02-02
add topic for solar potential

n-h-diaz:


server/lib/nl/topic.py
```diff
@@ -29,6 +29,7 @@
         "dc/15lrzqkb6n0y7",
         "dc/svpg/AmountOfFarmInventoryByType",
     ],
+    "dc/topic/SolarPotential": ["dc/svpg/SolarPotentialDetails"]
 }
```
pradh, line 33: Maybe add a comment on top that dc/topic/SolarPotential is a custom-dc only (not in KG) thing used for demo, with a TODO to remove it?

We could try adding this as MCF to the custom DC and if that works, remove it.

pradh (approved):

server/lib/nl/topic.py
n-h-diaz: added todo

n-h-diaz:
thanks for reviewing (and for all the help with this) 😃

PR #2157, opened 2023-02-02
Fork ranking into by place and by var. Also improve fallback

pradh:
Bo had the right intuition earlier today to break up ranking into by places and vars.  With the addition of "time delta" for ranking across vars, this is better.

Also, often times when the ranking, correlation and contained-in fails on a SV a better choice is to show the place rather than attempt some poor fallback of population in county, which invariably also fails.

### DIFF
--- a/server/lib/nl/fulfillment/base.py
+++ b/server/lib/nl/fulfillment/base.py
@@ -123,6 +123,8 @@ def _add_charts(state: PopulateState, places: List[Place],
     # REQUIRES: len(places) == 1
     places_to_check = utils.get_sample_child_places(places[0].dcid,
                                                     state.place_type.value)
+  if not places_to_check:
+    return False
 
   # Map of main SV -> peer SVs
   sv2extensions = variable.extend_svs(utils.get_only_svs(svs))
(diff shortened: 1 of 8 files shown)


server/lib/nl/fulfillment/base.py
```diff
@@ -175,6 +177,17 @@ def _add_charts(state: PopulateState, places: List[Place],
...
 
 
+def overview_fallback(state: PopulateState, places: List[Place],
+                      chart_origin: ChartOriginType) -> bool:
+  # Overview chart is a safe fallback.
+  state.block_id += 1
+  chart_vars = ChartVars(svs=[],
+                         block_id=state.block_id,
+                         include_percapita=False)
+  return add_chart_to_utterance(ChartType.PLACE_OVERVIEW, state, chart_vars,
```
shifucun, line 187: is this the overview for the first place in the list?

shifucun (approved):

server/lib/nl/fulfillment/base.py
pradh: Yes, good point.  As of now, there is only one place here.  When we handle multiple places from query PLACE_OVERVIEW handling in page_config_next.py might need improvement.

PR #2159, opened 2023-02-02
[Custom DC] improvements for re-runs

Fructokinase:
This PR adds the following improvements

- Website, mixer, and tools repo now are fetched from a tagged version of the code
- Docker images now match the version of the code with the install script.
- api key secret: global resource -> regional resource
- For re-runs, the script is no longer stuck trying to fetch non existent tmp files. We force a file fetch.
- Small fixes: BIGQUERY -> BIG_QUERY, always force mixer to be fetched for re-runs.

Tested using a [test tag](https://github.com/Fructokinase/website/releases/tag/test-custom-dc-v0.1.1). The current tag the script references to (custom-dc-v0.1.0) does not exist, and will be created in website repo and tools repo after this PR is committed.

Result: https://custom-dc-tag-based-launch-datacommons.com/

### DIFF
--- a/deploy/terraform-datacommons-website/modules/apikeys/main.tf
+++ b/deploy/terraform-datacommons-website/modules/apikeys/main.tf
@@ -62,6 +62,19 @@ gcloud alpha services api-keys create \
 --api-target=service=maps-backend.googleapis.com \
 --api-target=service=places-backend.googleapis.com
 
+EOT
+  }
+}
+
+resource "null_resource" "maps_api_key_fetch" {
+  # Regardless of the state, we always want to fetch the API key to a tmp file so
+  # the api key can be found in /tmp even in re-runs.
+  triggers = {
+    always_run = "${timestamp()}"
+  }
+
+  provisioner "local-exec" {
+    command = <<EOT
 touch /tmp/dc-website-api-key
 
 API_KEY_NAME=$(gcloud alpha services api-keys list --project=${var.project_id} --filter='displayName=maps-api-key${var.resource_suffix}' --format='value(name)' | head -n 1)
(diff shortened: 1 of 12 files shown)


deploy/terraform-datacommons-website/modules/apikeys/main.tf
```diff
@@ -62,20 +62,35 @@ gcloud alpha services api-keys create \
...
+    command = <<EOT
 touch /tmp/dc-website-api-key
 
 API_KEY_NAME=$(gcloud alpha services api-keys list --project=${var.project_id} --filter='displayName=maps-api-key${var.resource_suffix}' --format='value(name)' | head -n 1)
 gcloud alpha services api-keys get-key-string $API_KEY_NAME --format='value(keyString)' >> /tmp/dc-website-api-key
 
 EOT
   }
+
+  depends_on = [null_resource.maps_api_key]
```
shifucun, line 86: what is a "null_resource"?

scripts/install_custom_dc.sh
```diff
@@ -63,17 +65,22 @@ ROOT=$PWD
...
-  git clone https://github.com/datacommonsorg/website
+  git clone https://github.com/datacommonsorg/website --branch $CUSTOM_DC_RELEASE_TAG --single-branch
 fi
 
 cd website
-if [[ ! -d "mixer" ]]; then
-  git submodule foreach git pull origin master
-  git submodule update --init --recursive
-fi
+WEBSITE_GITHASH=$(git rev-parse --short=7 HEAD)
```
shifucun, line 72: Can we have a version file to store a fixed version file? And we can update it regularly.

deploy/terraform-datacommons-website/modules/apikeys/main.tf
Fructokinase: Added a readme!

https://github.com/datacommonsorg/website/blob/041f5b1f24b94b6b524ff8d8a00d88730f855e1c/deploy/terraform-datacommons-website/README.md#notes

scripts/install_custom_dc.sh
Fructokinase: I agree, but I will leave that as a separate work because it is a bit involving (ex: we need to make sure it works with the version files in the separate source repo).

shifucun (approved):
thanks for the update!

PR #2160, opened 2023-02-02
Implement time-delta fulfillment (making jobs queries work)

pradh:
The growth rate computation considers last 10 years and ensures it is resilient to seasonality for monthly data.

Also, include jobs and medical conditions topic SVs under an SVPG.

With this PR, the two jobs queries and health query should be handled, as shown below.

Screenshot 2023-02-02 at 12 10 04 AM

Screenshot 2023-02-02 at 3 29 07 AM

Screenshot 2023-02-02 at 12 09 12 AM

Screenshot 2023-02-02 at 12 09 44 AM

TESTED: all Jan demo scripts

### DIFF
--- a/server/lib/nl/fulfillment/ranking_across_vars.py
+++ b/server/lib/nl/fulfillment/ranking_across_vars.py
@@ -1,0 +1,81 @@
+# Copyright 2023 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from typing import List
+
+from lib.nl import utils
+from lib.nl.detection import ClassificationType
+from lib.nl.detection import Place
+from lib.nl.detection import RankingClassificationAttributes
+from lib.nl.fulfillment import context
+from lib.nl.fulfillment.base import add_chart_to_utterance
+from lib.nl.fulfillment.base import ChartVars
+from lib.nl.fulfillment.base import overview_fallback
+from lib.nl.fulfillment.base import populate_charts
+from lib.nl.fulfillment.base import PopulateState
+from lib.nl.utterance import ChartOriginType
+from lib.nl.utterance import ChartType
+from lib.nl.utterance import Utterance
+
+
+#
+# For ranking across vars, we should have detected a ranking, but not contained-in
+# classification in the current utterance.  In the callback, we will also
+# check that the SVs are part of a peer group (only those are comparable!).
+# For example, [most grown agricultural things], again assuming california
+# is in the context.
+# TODO: consider checking for common units, especially when we rely on
+#       auto-expanded peer groups of SVs.
+#
+def populate(uttr: Utterance):
+  # Get the RANKING classifications from the current utterance. That is what
+  # let us infer this is ranking query-type.
+  current_ranking_classification = context.classifications_of_type_from_utterance(
+      uttr, ClassificationType.RANKING)
+
+  if (current_ranking_classification and
+      isinstance(current_ranking_classification[0].attributes,
+                 RankingClassificationAttributes) and
+      current_ranking_classification[0].attributes.ranking_type):
+    ranking_types = current_ranking_classification[0].attributes.ranking_type
+
+    # Ranking among stat-vars.
+    if populate_charts(
+        PopulateState(uttr=uttr,
+                      main_cb=_populate_cb,
+                      fallback_cb=overview_fallback,
+                      ranking_types=ranking_types)):
+      return True
+
+  return False
+
+
+def _populate_cb(state: PopulateState, chart_vars: ChartVars,
+                 places: List[Place], chart_origin: ChartOriginType) -> bool:
+  if not state.ranking_types:
+    return False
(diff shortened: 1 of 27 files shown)


server/lib/nl/fulfillment/ranking_across_vars.py
```diff
@@ -75,7 +75,8 @@ def _populate_cb(state: PopulateState, chart_vars: ChartVars,
     return False
 
   # Ranking among peer group of SVs.
-  chart_vars.svs = utils.ranked_svs_for_place(places[0].dcid, chart_vars.svs,
-                                              state.ranking_types[0])
+  chart_vars.svs = utils.rank_svs_by_latest_value(places[0].dcid,
```
shifucun, line 78: does it matter if there are multiple places here? Maybe add a comment on implication of selecting the first one.

server/lib/nl/fulfillment/time_delta.py
```diff
@@ -0,0 +1,94 @@
...
+                     TimeDeltaClassificationAttributes) or
+      not time_delta_classification[0].attributes.time_delta_types):
+    return False
+  time_delta = time_delta_classification[0].attributes.time_delta_types
+
+  # Get potential ranking type
+  current_ranking_classification = context.classifications_of_type_from_utterance(
+      uttr, ClassificationType.RANKING)
+  ranking_types = []
+  if (current_ranking_classification and
```
shifucun, line 54: this if condition seems repeated in other places, consider a helper function? (or in later PR)

server/lib/nl/utils.py
```diff
@@ -167,6 +170,116 @@ def ranked_svs_for_place(place: str, svs: List[str],
...
+
+    if net_growth_rate > 0 and growth_direction != detection.TimeDeltaType.INCREASE:
+      continue
+    if net_growth_rate < 0 and growth_direction != detection.TimeDeltaType.DECREASE:
+      continue
+
+    svs_with_vals.append((sv, net_growth_rate))
+
+  if not rank_order:
+    reverse = True if growth_direction == detection.TimeDeltaType.INCREASE else False
```
beets, line 204: should this be the opposite (i.e. follow the growth direction if rank order is not specified?)

server/lib/nl/utils.py
```diff
@@ -167,6 +170,116 @@ def ranked_svs_for_place(place: str, svs: List[str],
...
+  else:
+    reverse = True  # Descending order
+    if growth_direction == detection.TimeDeltaType.INCREASE and rank_order == detection.RankingType.LOW:
+      # lowest growing
+      reverse = False
+    elif growth_direction == detection.TimeDeltaType.DECREASE and rank_order == detection.RankingType.HIGH:
+      # highest shrinking
+      reverse = False
+
+  logging.info(f'Rank order: {reverse}')
```
beets, line 214: we should add an easier way to pipe these down to debug info.

server/lib/nl/utils.py
```diff
@@ -167,6 +170,116 @@ def ranked_svs_for_place(place: str, svs: List[str],
...
+      continue
+
+    if net_growth_rate > 0 and growth_direction != detection.TimeDeltaType.INCREASE:
+      continue
+    if net_growth_rate < 0 and growth_direction != detection.TimeDeltaType.DECREASE:
+      continue
+
+    svs_with_vals.append((sv, net_growth_rate))
+
+  if not rank_order:
```
shifucun, line 203: I wonder if you can make a map from (rank_order, growth_direction) -> bool and just use that? Maybe easier to check the condition.

server/lib/nl/utils.py
```diff
@@ -167,6 +170,116 @@ def ranked_svs_for_place(place: str, svs: List[str],
...
+                         reverse=reverse)
+  logging.info(svs_with_vals)
+  return [sv for sv, _ in svs_with_vals]
+
+
+# Computes net growth-rate for a time-series including only recent (since 2012) observations.
+def compute_growth_rate(series: List[Dict]) -> float:
+  latest = None
+  earliest = None
+  for s in series:
```
shifucun, line 226: series should already be sorted. So can directly use the first and last item?

beets (approved):
nice!

shifucun (approved):

server/lib/nl/fulfillment/time_delta.py
pradh: SG for later PR.

server/lib/nl/utils.py
beets: chatted offline and we do want to reverse sort to show largest changes first.

server/lib/nl/utils.py
pradh: Ah interesting.  I didn't find it in https://docs.datacommons.org/api/rest/v1/bulk/observations/series, but could have missed.

But still need to loop to handle _THRESHOLD_DATE_FOR_GROWTH_RATE, so I'll just keep this for now.

server/lib/nl/fulfillment/ranking_across_vars.py
pradh: The API has multiple places (because it handles place comparison).

server/lib/nl/utils.py
pradh: Very nice idea! It reads better now.

server/lib/nl/utils.py
pradh: Sorry left over from debugging.

server/lib/nl/utils.py
pradh: I've added TODO to revisit.

pradh (commented):
Thanks for the reviews!

pradh:
Thanks both for the quick review!

server/lib/nl/utils.py
shifucun: Good point, this needs to be added in the documentation. This was one of the improvement in v1 API

PR #2161, opened 2023-02-02
Implement event fulfillment by leveraging disaster dashboard configs

pradh:
This approach is convenient, but adds config dependencies on disaster dashboard that can potentially impact NL.  We will eventually fork the config out to be its own thing.

Screenshot 2023-02-02 at 3 16 30 AM

Screenshot 2023-02-02 at 3 17 57 AM

Screenshot 2023-02-02 at 3 18 33 AM

Screenshot 2023-02-02 at 3 19 03 AM

Screenshot 2023-02-02 at 3 19 27 AM

Screenshot 2023-02-02 at 3 19 56 AM

### DIFF
--- a/server/lib/nl/fulfillment/event.py
+++ b/server/lib/nl/fulfillment/event.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 import logging
+from typing import List
 
 from lib.nl import detection
 from lib.nl import utils
@@ -23,18 +24,6 @@ import lib.nl.utterance as nl_uttr
 #
 # Handler for Event queries.
 #
-
-# NOTE: This relies on disaster config's event_type_spec IDs.
-_EVENT_TYPE_TO_CONFIG_KEY = {
-    detection.EventType.COLD: "cold",
-    detection.EventType.CYCLONE: "storm",
-    detection.EventType.DROUGHT: "drought",
-    detection.EventType.EARTHQUAKE: "earthquake",
-    detection.EventType.FIRE: "fire",
-    detection.EventType.FLOOD: "flood",
-    detection.EventType.HEAT: "heat",
-    detection.EventType.WETBULB: "wetbulb",
-}
 
 
 def populate(uttr: nl_uttr.Utterance) -> bool:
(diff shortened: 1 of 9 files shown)


beets (commented):
we could make a more complete textproto to reference with all known event types for NL usage (to decouple from dashboard updates)

pradh:
> we could make a more complete textproto to reference with all known event types for NL usage (to decouple from dashboard updates)

+1.

Do you think we can do that in Monday timeframe?  Or should I go forward with this PR (adding tests), and revisit next week?

beets:
sorry i missed this. let's go with this for now and we can refactor later. feel free to put todos for me to unroll that.

pradh:
> sorry i missed this. let's go with this for now and we can refactor later. feel free to put todos for me to unroll that.

Sounds good.  I've added tests and its now ready for review.

server/lib/nl/page_config_next.py
```diff
@@ -21,17 +21,44 @@
...
 from config.subject_page_pb2 import Tile
 from lib.nl import utils
+from lib.nl.detection import EventType
 from lib.nl.detection import Place
 from lib.nl.detection import RankingType
 from lib.nl.utterance import ChartSpec
 from lib.nl.utterance import ChartType
 from lib.nl.utterance import Utterance
 
+# NOTE: This relies on disaster config's event_type_spec IDs.
```
shifucun, line 31: Maybe both the textproto and here should use enum defined in proto

server/tests/lib/nl/page_config_next_test.py
```diff
@@ -548,6 +549,93 @@
  }
 """
 
+EVENT_CONFIG = """
```
shifucun, line 552: add a TODO to save these textproto into test_data/*.textproto?

shifucun (approved):

beets (approved):
thanks for adding the tests!

server/lib/nl/page_config_next.py
pradh: Which enum in proto, could you point to it?

shifucun: I am suggesting to create an Enum in the proto, as now they are free formed key?

server/lib/nl/fulfillment/event.py
```diff
@@ -0,0 +1,78 @@
...
+#
+
+
+def populate(uttr: nl_uttr.Utterance) -> bool:
+  event_classification = ctx.classifications_of_type_from_utterance(
+      uttr, detection.ClassificationType.EVENT)
+  if (not event_classification or
+      not isinstance(event_classification[0].attributes,
+                     detection.EventClassificationAttributes) or
+      not event_classification[0].attributes.event_types):
```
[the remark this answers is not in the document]
pradh: Done (in base.py so its common).  Lets defer refactors to after Monday?

server/tests/lib/nl/page_config_next_test.py
pradh: Done.

pradh (commented):
Thanks!  I'll submit this and handle followups in different PR.

server/lib/nl/page_config_next.py
pradh: Oh, I thought you were suggesting a pre-existing enum.  That is a good idea, I'll add a TODO.

@chejennifer or @beets for inputs/scheduling

pradh (commented):
Thanks for the reviews!

server/lib/nl/page_config_next.py
chejennifer: I don't think having an enum in the proto for event type ids makes the most sense because the event_type_spec was designed to be created by whoever is creating that page's config. I would think of it like the stat_var_spec where the caller adds an event_type_spec with the id "x" & then includes "x" in the list of event_type_keys for tiles where that event type should be shown. 

I would actually suggest not relying on disaster dashboard's config, but instead, maybe have a constant dict somewhere where key is EventType and value is all the field values for the event_type_spec we want for that EventType. Then when adding an event chart block, use this constant to add required event_type_specs to the page metadata.

pradh: I actually started with the latter approach, and found myself duplicating proto config in py code, and it looked ugly :)

I slightly like the plan that @beets suggests better, of using textproto, but forking it away from disasters.

chejennifer: ah, I see, so a .textproto file with all the event_type_specs specific for NL? that makes sense to me!

pradh: yup!

PR #2162, opened 2023-02-02
update solar potential topic

n-h-diaz:


server/lib/nl/topic.py
```diff
@@ -54,17 +61,30 @@
         "Amount_FarmInventory_WheatForGrain",
         "Amout_FarmInventory_CornForGrain",
     ],
-    "dc/svpg/SolarPotentialDetails": [
-        "Count_Building_SuitableForSolar",
-        "Percent_Building_SuitableForSolar_ProjectSunroof",
+    "dc/svpg/SolarEnergyGenerationPotential": [
         "Amount_SolarPotential",
```
pradh, line 65: Do you want to put the total in the main list and have a peer-group for break down only?

pradh (approved):

server/lib/nl/topic.py
n-h-diaz: good idea! done (thanks for the quick review :) )

PR #2164, opened 2023-02-02
[subject page] fix re-renders

chejennifer:


shifucun (approved):
wonder why this was not an issue before?

chejennifer:
> wonder why this was not an issue before?

this was because we added a geojson data fetch at the page level. Even though the data gets saved to the context, the state still updated causing the page to re-render and the children components to also re-render.

beets (approved):
nice find. was it because we moved the filtering & data fetch out?

chejennifer:
> nice find. was it because we moved the filtering & data fetch out?

no, it was from moving the geojson data fetch out

shifucun:
> > wonder why this was not an issue before?
> 
> this was because we added a geojson data fetch at the page level. Even though the data gets saved to the context, the state still updated causing the page to re-render and the children components to also re-render.

This is good for now but I wonder if we should have precise useEffect dependency or other stuff. If data is not available, then page should not render, so this indicates some dependency issues. Just a little worried of adding memo to components (and seems this could be done for any components?)

shifucun:
I have seen similar issues when refactor map tools. Dependency can be tricky and important.

chejennifer:
> > > wonder why this was not an issue before?
> > 
> > 
> > this was because we added a geojson data fetch at the page level. Even though the data gets saved to the context, the state still updated causing the page to re-render and the children components to also re-render.
> 
> This is good for now but I wonder if we should have precise useEffect dependency or other stuff. If data is not available, then page should not render, so this indicates some dependency issues. Just a little worried of adding memo to components (and seems this could be done for any components?)

Yes, I think this is something that should be re-visited, I'll create an issue

PR #2167, opened 2023-02-02
[NL Next] A few fulfillment fixes

pradh:
1. Downgrade single point time-series to bar charts
2. Open up topics in correlation, and include multiple SVs from current utterance and past context
3. Show contained-in maps for peer-group SVs
4. Sprinkle more debug logs

This improves on Jan script, in these ways:

(1) [agriculture in california] contained single-dot timeline charts
Screenshot 2023-02-02 at 12 11 47 PM

(2) There used to be only one correlation chart
Screenshot 2023-02-02 at 12 12 22 PM

### DIFF
--- a/server/lib/nl/fulfillment/base.py
+++ b/server/lib/nl/fulfillment/base.py
@@ -178,6 +178,8 @@ def _add_charts(state: PopulateState, places: List[Place],
       logging.info('Existence check failed for %s - %s',
                    ', '.join(places_to_check), ', '.join(extended_svs))
 
+  logging.info("Add chart %s %s returning %r" %
+               (', '.join(_get_place_names(places)), svs, found))
   return found
 
 
--- a/server/lib/nl/fulfillment/base.py
+++ b/server/lib/nl/fulfillment/base.py
@@ -115,7 +115,7 @@ def populate_charts_for_places(state: PopulateState,
 # TODO: Batch existence check calls
 def _add_charts(state: PopulateState, places: List[Place],
                 svs: List[str]) -> bool:
-  print("Add chart %s %s" % (', '.join(_get_place_names(places)), svs))
+  logging.info("Add chart %s %s" % (', '.join(_get_place_names(places)), svs))
 
   # If there is a child place_type, get child place samples for existence check.
   places_to_check = _get_place_dcids(places)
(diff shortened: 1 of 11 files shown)


server/lib/nl/fulfillment/base.py
```diff
@@ -249,3 +253,38 @@ def _svg_or_topic_to_svs(state: PopulateState, sv: str,
     return charts
 
   return []
+
+
+def open_topics_ordered(svs: List[str]) -> List[str]:
+  opened_svs = []
+  sv_set = set()
+  for rank, sv in enumerate(svs):
```
jehangiramjad, line 261: is `svs` expected to be ordered in some way to have `rank` make sense? If so, please add a comment to that effect.

server/lib/nl/fulfillment/base.py
```diff
@@ -249,3 +253,38 @@ def _svg_or_topic_to_svs(state: PopulateState, sv: str,
...
+    for sv in _open_topic_in_var(sv, rank):
```
jehangiramjad, line 262: both the inner and outer loops are using the same var name `sv`?

server/lib/nl/fulfillment/base.py
```diff
@@ -249,3 +253,38 @@ def _svg_or_topic_to_svs(state: PopulateState, sv: str,
...
+      if sv not in sv_set:
```
jehangiramjad, line 263: why use both a set and a list? could we not just use the set and return list(sv_set)?

server/lib/nl/fulfillment/base.py
```diff
@@ -249,3 +253,38 @@ def _svg_or_topic_to_svs(state: PopulateState, sv: str,
...
+        opened_svs.append(sv)
+        sv_set.add(sv)
+  return opened_svs
+
+
+def _open_topic_in_var(sv: str, rank: int) -> List[str]:
+  if utils.is_sv(sv):
+    return [sv]
+  if utils.is_topic(sv):
```
jehangiramjad, line 272: Why is nothing being returned from this conditional path?

server/lib/nl/utils.py
```diff
@@ -170,6 +170,21 @@ def rank_svs_by_latest_value(place: str, svs: List[str],
...
+                                 all_facets=False)
+  for _, place_data in series_data['data'].items():
+    if place not in place_data:
+      continue
+    series = place_data[place]['series']
+    if len(series) < 2:
+      logging.info('Found single data point series in %s - %s', place,
+                   ', '.join(svs))
+      return True
+  return False
```
jehangiramjad, line 185: Just checking the logic: a series with no points is also false because it does NOT have a single data point?

jehangiramjad (commented):
Thanks. Some questions. Haven't reviewed the tests carefully though

jehangiramjad (approved):
Approving to unblock

server/lib/nl/fulfillment/base.py
pradh: good catch!

server/lib/nl/fulfillment/base.py
pradh: Because we want it ordered.

server/lib/nl/fulfillment/base.py
pradh: Great eyes!  I just fixed the bug.

server/lib/nl/fulfillment/base.py
pradh: Added comments

PR #2169, opened 2023-02-02
[NL] Add a URL param to put in a series of prompts 2s apart

beets:
to make testing a sequence of prompts easier on our hands.
e.g. using single query param to enter each spaced 2s apart. each query should be separated by semi-colon.

http://localhost:8080/nlnext/#q=tell%20me%20about%20palo%20alto;what%20about%20auto%20theft;compare%20with%20Berkeley

### DIFF
--- a/static/js/apps/nl_interface/app.tsx
+++ b/static/js/apps/nl_interface/app.tsx
@@ -30,6 +30,7 @@ export function App(): JSX.Element {
   const [contextList, setContextList] = useState<any[]>([]);
 
   // Executes a single query through the search box.
+  // TODO: Do this by going through state/props instead of directly manipulating the DOM.
   function executePrompt(prompt: string) {
     if (prompt) {
       (
--- a/static/js/apps/nl_interface/app.tsx
+++ b/static/js/apps/nl_interface/app.tsx
@@ -20,6 +20,7 @@
 
 import React, { useEffect, useState } from "react";
 
+import { getUrlToken } from "../../tools/stat_var/util";
 import { isNLInterfaceNext } from "../../utils/nl_interface_utils";
 import { QueryResult } from "./query_result";
 import { QuerySearch } from "./query_search";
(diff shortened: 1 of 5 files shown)


pradh:
This is quite cool!  As we expose this to more people, we can perhaps use this to sprinkle some links to example flows on the main page, so they aren't presented with a blank screen, but have an idea for the queries that work and results that show.

static/js/apps/nl_interface/app.tsx
```diff
@@ -28,6 +29,38 @@ export function App(): JSX.Element {
...
+    }
+  }
+
+  useEffect(() => {
+    // Runs each prompt (';' separated) 2s apart.
+    const urlPrompts = getUrlToken("q");
+    if (urlPrompts) {
+      const prompts = urlPrompts.split(";");
+      executePrompt(prompts.shift());
+      if (prompts.length) {
```
chejennifer, line 50: hm, I think this would run max 2 prompts, did you want a while loop instead?

static/js/apps/nl_interface/app.tsx
```diff
@@ -28,6 +29,38 @@ export function App(): JSX.Element {
...
 
+  // Executes a single query through the search box.
+  function executePrompt(prompt: string) {
+    if (prompt) {
+      (
+        document.getElementById("query-search-input") as HTMLInputElement
+      ).value = prompt;
+      (
+        document.getElementById("rich-search-button") as HTMLButtonElement
+      ).click();
```
chejennifer, line 40: this feels hacky, but fine for now!

Maybe there is some way to make use of the initialValue prop in TextSearchBar instead of having to find the element and set the value

chejennifer (approved):
this is very cool!

static/js/apps/nl_interface/app.tsx
beets: I tried that but was unsuccessful... I also didn't want to mess with the state too much. This somewhat simulates what a user would do :) but I will add a TODO to do this in a less hacky way.

static/js/apps/nl_interface/app.tsx
beets: The `prompts` array gets gradually `shift`ed in the interval until the array is empty. I also wanted to execute the first immediately.

beets (commented):
thanks for the reviews!

PR #2170, opened 2023-02-02
[NL Next] Improve Time-Delta classifier and add Unit Testing

juliawu:
Adds more heuristics to the time-delta classifier to make it more robust. More wordings are now supported.

Tested by:
(1) Running new classifier on 50+ sample queries (both manual and GPT-3 generated)
(2) Running new classifier on all previous sample queries to check for false positives
(3) Running through both test scripts, confirmed no change in output correctness.
(4) Added unit-tests

Examples of new queries supported:
Screenshot 2023-02-02 at 2 47 04 PM
Screenshot 2023-02-02 at 2 50 34 PM


juliawu:
Update: also added unit tests following discussion in PR #2171.

server/lib/nl/constants.py
```diff
@@ -238,12 +238,26 @@
         },
         "TimeDelta": {
             "Increase": [
-                "grow(n|th)",
+                "grow(n|th)?",
+                "grew",
+                "gain",
                 "increased?",
```
shifucun, line 244: how about increasing?

shifucun (approved):

PR #2175, opened 2023-02-03
[NL] add labels on scatter for most correlated places 

chejennifer:
Screenshot 2023-02-03 at 9 28 45 AM

when there are points close together
Screenshot 2023-02-03 at 11 13 40 AM

when there are not many points in the quadrant of interest
Screenshot 2023-02-03 at 9 29 10 AM


pradh:
@chejennifer : @rvguha asks "is there something we can do to so that we print fewer so that whatever is printed is legible?"

chejennifer:
> @chejennifer : @rvguha asks "is there something we can do to so that we print fewer so that whatever is printed is legible?"

hm, so if there's overlapping labels, then do we remove labels for every dot in that section? So in this example, do we want to remove the labels for everything that is in the larger box but not in the smaller box?

static/js/chart/draw_scatter.ts
```diff
@@ -550,6 +558,151 @@ function addRegressionLine(
...
+  points: Point[],
+  xScale: ScatterScale,
+  yScale: ScatterScale,
+  quadrants: ChartQuadrant[]
+): Record<ChartQuadrant, Point[]> {
+  const numSectionsPerAxis = HIGHLIGHT_QUADRANT_SECTIONS * 2;
+  const xSectionLength =
+    (xScale.range()[1] - xScale.range()[0]) / numSectionsPerAxis;
+  const ySectionLength =
+    (yScale.range()[0] - yScale.range()[1]) / numSectionsPerAxis;
```
beets, line 596: nit: we should only call `.range()` once since it could get expensive with the number of points

static/js/chart/draw_scatter.ts
```diff
@@ -550,6 +558,151 @@ function addRegressionLine(
...
+        for (let y = startingY - 1; y >= 0; y--) {
+          if (sectionPoints[startingX] && sectionPoints[startingX][y]) {
+            points.push(...sectionPoints[startingX][y]);
+          }
+        }
+      }
+      layer++;
+    }
+    // filter out points that are too close to other points
+    result[quadrant] = points.filter((point, idx) => {
```
beets, line 664: hm, i wonder if we should reverse `points` before calling filter. If filter goes 0-n, then we might overfilter points that we'd want to keep (further towards the corners of the chart).

static/js/chart/draw_scatter.ts
```diff
@@ -550,6 +558,151 @@ function addRegressionLine(
...
+    .attr("x", (point) => xScale(point.xVal) + 7)
+    .attr("y", (point) => yScale(point.yVal))
+    .text((point) => point.place.name);
+}
+
+/**
+ * Gets at least MIN_HIGHLIGHT_POINTS points to highlight in each quadrant for a
+ * list of quadrants
+ */
+function getQuadrantHighlightPoints(
```
beets, line 586: we could split this up more into smaller fns but this is good for now!

beets (approved):
v nice!

static/js/chart/draw_scatter.ts
chejennifer: actually, the points are in order of furthest towards the corners of the chart -> closer to the center of the chart!

static/js/chart/draw_scatter.ts
chejennifer: done

static/js/chart/draw_scatter.ts
chejennifer: ok, will add a TODO

PR #2176, opened 2023-02-03
[NL Next] Add support for 'major' in rankings

juliawu:
Add support for 'major' triggering a high ranking classification, including test cases. This allows for queries like "which were the major earthquakes in california?"

Similar additions like "large", "big", "small", "minor" were tested, but prone to false positives, so leaving those out for now.


pradh (approved):
Thanks.

PR #2177, opened 2023-02-03
[NL Next] Add overview classifier

juliawu:
This PR:
* Adds an "overview" classifier to look for "tell me about <place>" and "more about <place>" queries
* Hooks the classifier up to debug info
* Adds unit-tests for the new classifier

Example:
image

### DIFF
--- a/server/lib/nl/constants.py
+++ b/server/lib/nl/constants.py
@@ -263,7 +263,7 @@ QUERY_CLASSIFICATION_HEURISTICS: Dict[str, Union[List[str], Dict[
                 "loss",
             ],
         },
-        "Overview": ["(tell me|more) about",],
+        "Overview": ["tell me about",],
     }
 
 PLACE_TYPE_TO_PLURALS: Dict[str, str] = {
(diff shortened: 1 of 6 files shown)


server/lib/nl/constants.py
```diff
@@ -263,6 +263,7 @@
                 "loss",
             ],
         },
+        "Overview": ["(tell me|more) about",],
```
pradh, line 266: lets just stick to "tell me about" per guha's last comment?

server/lib/nl/detection.py
```diff
@@ -231,6 +238,7 @@ class ClassificationType(IntEnum):
     ClassificationType.CORRELATION,
     ClassificationType.TIME_DELTA,
     ClassificationType.EVENT,
+    ClassificationType.OVERVIEW,
```
pradh, line 241: Can you remove from this?  The ranking here I think could result in not fulfilling the query when we run with this change

pradh: (might be worth testing what happens with the query)

pradh (commented):
Thanks for the quick PR!

server/lib/nl/constants.py
juliawu: Done!

server/lib/nl/detection.py
juliawu: Removed! Tested and "tell me about tulare county" and "tell me about palo alto" work after removal.

pradh (approved):

PR #2178, opened 2023-02-03
Add counters to fulfillment flows

pradh:
One dedupe bug fix in server/lib/nl/fulfillment/context.py.

I've tested it with the sample queries.

### DIFF
--- a/server/lib/nl/constants.py
+++ b/server/lib/nl/constants.py
@@ -174,6 +174,7 @@ QUERY_CLASSIFICATION_HEURISTICS: Dict[str, Union[List[str], Dict[
                 "sickest",
                 "illest",
                 "oldest",
+                "major",  # as in 'major storms'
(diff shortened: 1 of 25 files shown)


shifucun (approved):
Very nice! For adding future counters, how to find / reuse a counter key?

pradh:
> Very nice! For adding future counters, how to find / reuse a counter key?

By looking at all references to `update_counter`?  We could define a constant for every counter, but I was trading off ease of adding it as we write code to reuse.

pradh:
Thanks for the review!

shifucun:
> > Very nice! For adding future counters, how to find / reuse a counter key?
> 
> By looking at all references to `update_counter`? We could define a constant for every counter, but I was trading off ease of adding it as we write code to reuse.

Yea, constant might be annoying.

PR #2182, opened 2023-02-03
[NL] Top Event tile updates

chejennifer:
- remove "unnamed event" if no events have names
- change "unnamed event" to "<event type> event in <place>" or "<event type> event at latlong: <lat, long>" if some events in the list have names
- add a <place type> column to show the place each event took place in
- Update NL page config generation for event block to add a contained place type

Screenshot 2023-02-03 at 3 52 28 PM

Additional cases:
1. no event names
Screenshot 2023-02-03 at 3 53 04 PM

2. less than 40% of the event latitude/longitude could resolve to a place of the specified type
Screenshot 2023-02-03 at 3 53 37 PM

3. some event latitude/longitude could not resolve to a place
Screenshot 2023-02-03 at 3 54 42 PM


static/js/components/tiles/top_event_tile.tsx
```diff
@@ -107,13 +125,20 @@ export function TopEventTile(props: TopEventTilePropType): JSX.Element {
...
+                        <td>
+                          <a href={`/browser/${event.placeDcid}`}>
+                            {getEventName(event)}
+                          </a>
+                        </td>
+                      )}
+                      {showPlaceColumn && (
+                        <td>
+                          {eventPlaces[event.placeDcid]
+                            ? eventPlaces[event.placeDcid].name
```
beets, line 138: please also add a link to the event here

beets (approved):
thanks!!

static/js/components/tiles/top_event_tile.tsx
chejennifer: should we always link the event here or only when there's no event name?

chejennifer: updated to add a link only when there's no event name

PR #2184, opened 2023-02-04
[NL] add more legibility to bar chart legend

beets:
image

(tried black text, but the result is worse imo, since we have such strong colors for the background. changing the background would effect place explorer et al, so this feels safer for now).
image


chejennifer (approved):
nice! 

another idea could be to try a low opacity dark background behind the text?

pradh (approved):

beets:
> nice!
> 
> another idea could be to try a low opacity dark background behind the text?

Yea, was trying to do that somewhat with the text shadow without altering the dom structure.. we'll see how well this works, if not we can go modify how we build the legend (maybe switch to disaster style legend).

PR #2187, opened 2023-02-04
Miscellaneous NL fixes for script

pradh:
(1) Add topic for Projected Climate Extremes (ensure temperature hack doesn't kick in)
(2) Add per-capita for ranking across vars and multi-var chart (peer group)
(3) For growth rate computation, consider the whole series instead of last 10 years
(4) For disasters show both top-events and map
(5) Drop a stale Jobs SV and add a few more missing 2-digit NAICS
(6) For ranking, plot all the SVs in the peer-group (we were previously not handling peer-group in ranking)


chejennifer (approved):
nice!

PR #2188, opened 2023-02-04
[NL] show end date on top event ranking tile

chejennifer:
Screenshot 2023-02-03 at 6 40 45 PM


pradh (approved):
Nice!

beets (commented):
[for later] Hm, I thought the "dates" list was supposed to have all dates associated with the event, so the latest date is the end date (instead of specifying these props). @Spaceenter

pradh:
hi, please validate the full demo sequence for any regression.

PR #2189, opened 2023-02-04
Fix google map zoom out bug

juliawu:
Adds a listener to map state, to set map bounds after the map has finished loading. This fixes the issue where the map zooms out to earth intermittently.

Before:
Screenshot 2023-02-03 at 7 34 15 PM

After:

Screenshot 2023-02-03 at 7 26 38 PM

More examples:

Screenshot 2023-02-03 at 7 26 59 PM

Screenshot 2023-02-03 at 7 27 13 PM

Also verified this change doesn't break browser or place pages:

Screenshot 2023-02-03 at 7 27 39 PM
Screenshot 2023-02-03 at 7 27 49 PM


pradh (approved):
nice find!

PR #2190, opened 2023-02-04
Miscellaneous fixes for new script

pradh:
(1) Support fulfillment of ACROSS contained-in type by resolving to a place-type based on the place considered
(2) Support fulfillment of EXTREME ranking type by deciding whether to show lowest based on SV name
(3) Prune some SV names per Guha request (a hack for now)
(4) Modify the RANKING chart display to always show an adjoining map on the same row
(5) Drop a weird jobs SV from topic
(6) Make and use helper functions for getting {ranking-type, contained-in, time-delta-types} from utterance
(7) Fix the temperature hack properly to avoid pinned max dates

Sorry that (6) is in this PR.  I have tested both the old and new scripts with this though.

For the main UI change of ranking-across-places results, here is an example screenshot.

Screenshot 2023-02-03 at 10 06 25 PM

### DIFF
--- a/server/lib/nl/utils.py
+++ b/server/lib/nl/utils.py
@@ -23,6 +23,8 @@ from typing import Dict, List, Set, Union
 
 import lib.nl.constants as constants
 import lib.nl.detection as detection
+import lib.nl.fulfillment.context as ctx
+import lib.nl.utterance as nl_uttr
 import lib.util as util
 import services.datacommons as dc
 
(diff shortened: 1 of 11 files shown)


rvguha:
approve

rvguha (approved):
looks good

pradh:
@beets @shifucun - please have a look when you find time.  submitting to review script on autopush.

server/lib/nl/utils.py
```diff
@@ -371,13 +373,34 @@ def get_sv_name(all_svs: List[str]) -> Dict:
...
+      sv_name_map[sv] = clean_sv_name(title_by_sv_dcid[sv])
     else:
-      sv_name_map[sv] = uncurated_names[sv]
+      sv_name_map[sv] = clean_sv_name(uncurated_names[sv])
 
   return sv_name_map
 
 
+# TODO: Remove this hack by fixing the name in schema and config.
+def clean_sv_name(name: str) -> str:
```
beets, line 384: there's a similar one in https://github.com/datacommonsorg/website/blob/master/server/lib/nl/descriptions.py#L37

do you also want to add "Percentage of"?

beets (commented):
nice, just a small comment

PR #2194, opened 2023-02-05
Add Climate Change Topic with mix of events and stat-vars in a topic

pradh:
(1) Includes support for including events in topics, mixing with other SVs
(2) Also add existence check for events


rvguha (approved):
approved

PR #2196, opened 2023-02-05
Deploy latest binary/data to any environment

shifucun:
This makes it possible to deploy latest version to dev and other demo instances


pradh (approved):
Thanks!  Could you update README too, if its not documented already?

shifucun:
> Thanks! Could you update README too, if its not documented already?

Done

PR #2201, opened 2023-02-06
[NL] small ui fixes

beets:
- made maps fill up height if possible
- made show debug info less prominent
- update "replay" so text shows in query for 2s before submitting, replay queries 5s apart instead of 10s

image


shifucun (approved):

beets (commented):
thanks for the review!

PR #2203, opened 2023-02-06
[NL] update automatic query search

chejennifer:
- type query 1 character at a time w/ 0.05s in between, enter after 1sec after entire query entered, start typing next query after 2.5sec after last query entered

screen-recording-_71_


beets (approved):
v cool - thanks!

PR #2204, opened 2023-02-06
[NL] add sv footnote to ranking queries

beets:
image


server/lib/nl/utils.py
```diff
@@ -469,6 +484,21 @@ def clean_sv_name(name: str) -> str:
   return name
 
 
+def get_sv_footnote(all_svs: List[str]) -> Dict:
+  sv2footnote_raw = dc.property_values(all_svs, 'footnote')
```
chejennifer, line 488: curious - is footnote a PV?

chejennifer (approved):
nice!

server/lib/nl/utils.py
beets: it will be..

PR #2205, opened 2023-02-06
[NL] update chart titles

chejennifer:
- add unit to temperature charts
- remove "min" and "max" from titles

Screenshot 2023-02-06 at 11 26 42 AM


beets (approved):
thanks

PR #2212, opened 2023-02-07
Add place names and child place types to all relevant charts

pradh:
I think the test diffs give an idea of the title change.  I've sanity checked the demo scripts.

Pasting just a couple of screenshots (there are too many).

Screenshot 2023-02-06 at 6 34 23 PM
Screenshot 2023-02-06 at 6 34 37 PM


server/lib/nl/page_config_next.py
```diff
@@ -264,10 +267,13 @@ def _multiple_place_bar_block(column, places: List[Place], svs: List[str],
     title = attr['title']
   elif len(svs) > 1:
     # This suggests we are comparing against SV peers from SV extension
-    title = 'Compare with Other Variables'
+    title = 'Comparison with Other Variables'
```
beets, line 270: How about "Compared with"? Otherwise should be "In comparison with"

beets (commented):
Might I suggest a `_decorate_title` that takes in a dictionary of possible flags? That will give further flexibility with the titles, since  now they are dependent on the call order.

pradh:
> Might I suggest a `_decorate_title` that takes in a dictionary of possible flags? That will give further flexibility with the titles, since now they are dependent on the call order.

Good idea!  I added the function.  But it could not take in an attr dictionary (because for e.g., the same attr we want with-pc and without-pc charts).

PTAL.

server/lib/nl/page_config_next.py
pradh: Done.

pradh:
Thanks for the review!

beets (approved):

PR #2213, opened 2023-02-07
[NL] Integration test

shifucun:
Added an integration test where web and nl Flasks server are brought together to check data response.

The test can be used to 1) generate golden data or 2) test against golden data.


pradh:
This is very cool!  Thanks for adding it!

PR #2214, opened 2023-02-07
[subject page] Add support for multiple pages

chejennifer:
- get unique id for each component when there are multiple subject pages
- update draw_d3_map to take a ref element instead of an element id

small css fix:
- update font color for severity filters (for disaster event tiles) on nl interface to white


beets (approved):
thanks!

PR #2215, opened 2023-02-07
[Flask Server] A bunch of improvement clean up to get read for integration test

shifucun:
* Use absolute path for configs
* Add pytest config file so it's easier to see log output
* Move all flask app code inside create_app() function
* extract backend check into util functions
* disable "ADMIN" in configmodule as that is not used in any instances
* Only use "POST" for /nlnext/data as the code does not work for "GET"

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,4 +1,4 @@
 [pytest]
 log_cli=true
 log_level=INFO
-log_file=log.txt
+log_file=pytest_log.txt
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,0 +1,4 @@
+[pytest]
+log_cli=true
+log_level=INFO
+log_file=log.txt
--- a/server/lib/util.py
+++ b/server/lib/util.py
@@ -15,7 +15,11 @@
 from datetime import datetime
 import hashlib
 import json
+import logging
 import os
+import time
+from typing import List
+import urllib
 
 from config import subject_page_pb2
 from google.protobuf import text_format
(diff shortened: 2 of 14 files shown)


server/lib/util.py
```diff
@@ -41,11 +45,14 @@
     'poverty': ['USA', 'India'],
 }
 
+_root_dir = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
```
pradh, line 48: should we put this in a more general util and use it?

pradh (commented):
Nice, LG!  @beets is probably a better reviewer for the changes.

server/lib/util.py
shifucun: Good idea. Made a function, but still can not share with nl_server for now.

pytest.ini
```diff
@@ -0,0 +1,4 @@
+[pytest]
+log_cli=true
+log_level=INFO
+log_file=log.txt
```
beets, line 4: maybe pytest_log.txt?

server/lib/util.py
```diff
@@ -196,4 +208,37 @@ def _compact_series(series_resp, all_facets):
...
+    # Codacity still calls out the error so disable the check.
+    # https://bandit.readthedocs.io/en/latest/blacklists/blacklist_calls.html#b310-urllib-urlopen
+    urllib.request.urlopen(url)  # nosec B310
+    return True
+  except urllib.error.URLError:
+    return False
+
+
+def check_backend_ready(urls: List[str]):
+  timeout = 120  # seconds
```
beets, line 229: use a constant for 120?

beets (approved):

server/lib/util.py
shifucun: Done

pytest.ini
shifucun: Done

shifucun (commented):
Thanks for review!

PR #2216, opened 2023-02-07
[Devops] ESP 2G -> 1G memory to fit another pod in a node

Fructokinase:
Decreasing this changes memory requirement from 15G -> 14G per pod. Nodes currently have ~29G of memory and this allows 2 pods to fit in a node as opposed to 1.


shifucun (approved):
Very nice find! Given that we know the exact node memory, shall we make this 1.5G if that works.

PR #2219, opened 2023-02-08
[maps] Clean up map drawing functions

chejennifer:
Clean up map drawing functions to make it easier to add additional geojson layers on top of the base map:
- extracted legend generation out of base map draw function
- extracted projection generation out of base map draw function
- updated legend generation to take a container element ref instead of container id
- updated map point filtering to filter out NaN


static/css/place/place.scss
```diff
@@ -347,4 +347,12 @@ h3 {
...
   text-align: center;
   margin-top: 0rem;
+}
+
+.map-container {
+  display: flex;
+
+  .map {
+    flex-grow: 1;
+  }
```
beets, line 357: curious what this css fixes on place explorer?

static/js/chart/draw_d3_map.ts
```diff
@@ -511,7 +489,12 @@ export function addMapPoints(
     });
   const minDotSize = minDotRadius || Math.max(minRegionDiagonal * 0.02, 1.1);
   const filteredMapPoints = mapPoints.filter((point) => {
-    return !_.isNull(projection([point.longitude, point.latitude]));
+    const projectedPoint = projection([point.longitude, point.latitude]);
+    return (
+      projectedPoint &&
+      !_.isNaN(projectedPoint[0]) &&
+      !_.isNaN(projectedPoint[1])
```
beets, line 496: did you dig into which data points were returning NaN here? there might be some underlying data issues we should fix.

static/css/place/place.scss
beets: also curious there's no css update to tools/map

beets (approved):
thanks for the refactor! please test all our map surfaces again before merging..

static/css/place/place.scss
chejennifer: since I extracted the legend drawing functionality out of the original map drawing function, I also need to have one div to draw the map in and one to draw the legend in. This does the formatting for the two divs. The tools/map was originally already following this pattern so no need for css changes.

static/js/chart/draw_d3_map.ts
chejennifer: dug into this and looked like it was because mixer was returning undefined for 0 values. have a pr up to fix this: https://github.com/datacommonsorg/mixer/pull/1024

PR #2220, opened 2023-02-09
[NL] Add demo script flow to integration tests

pradh:
Also, make the test run on 3.8+ python on MacOS.


shifucun (approved):
very nice! future change will have much more confidence!

beets (approved):
super nice - thanks for adding this!

PR #2223, opened 2023-02-09
[disaster] proto changes for adding geojson layers to disaster event map

chejennifer:
- update DisasterEventTypeSpec to have the fields `repeated string point_event_type_key`, `repeated string path_event_type_key`, and `repeated string polygon_event_type_key`
- update EventTypeSpec to have a new field `string geo_json_prop`


server/config/subject_page.proto
```diff
@@ -53,6 +53,8 @@ message EventTypeSpec {
   repeated DisplayProp display_prop = 6;
   // List of props that could be used to hold the end date for this event type
   repeated string end_date_prop = 7;
+  // The prop to use to get the geojson for each event
+  string geo_json_prop = 8;
```
beets, line 57: since we are separating polygons and paths - perhaps do that for the property too.

beets (approved):

server/config/subject_page.proto
chejennifer: good idea, done

PR #2225, opened 2023-02-10
[NL] Implement query-type fallback

pradh:
In fulfillment, loop over a sequence of query_types until we are able to produce some charts.

The two key functions are:
* `first_query_type`: Given a bunch of classification results in an utterance, we pick the first query_type to process
* `next_query_type`: Given a list of past query_types we tried to fulfill and failed, we pick the next query_type to process

Given fallbacks at query_type level, drop per-fulfillment-type fallbacks.

Also, separate out the ClassificationType from a new QueryType, since the latter may be more fine-grained.


server/lib/nl/fulfillment_next.py
```diff
@@ -28,11 +30,75 @@
...
+        QueryType.SIMPLE,
+    ClassificationType.OTHER:
+        QueryType.SIMPLE,
+    ClassificationType.TEMPORAL:
+        QueryType.SIMPLE,
+    ClassificationType.CLUSTERING:
+        QueryType.SIMPLE,
+}
+
+DIRECT_QUERY_TYPE_FALLBACK = {
```
beets, line 91: add a comment here that some have more complex fallback logic?

beets (approved):
the dictionaries will get really hard to manage.. could we encode the logic in querytype classes instead (with fields that must be set, eg. handler, classification, fallback etc)

server/lib/nl/fulfillment_next.py
```diff
@@ -28,11 +30,75 @@
...
+RANKED_QUERY_TYPES = [
+    QueryType.SIMPLE,
+    QueryType.COMPARISON,
+    QueryType.CONTAINED_IN,
+    QueryType.RANKING_ACROSS_VARS,
+    QueryType.RANKING_ACROSS_PLACES,
+    QueryType.CORRELATION,
+    QueryType.TIME_DELTA,
+    QueryType.EVENT,
+    QueryType.OVERVIEW,
```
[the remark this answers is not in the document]
pradh: This is the order of query_types to first try out.  Currently, if we classified a query as OVERVIEW, Guha had wanted it to trump everything else.

In terms of fallback, you're right that it is after SIMPLE.

server/lib/nl/fulfillment_next.py
pradh: Added (but this is now in a new file)

pradh:
> the dictionaries will get really hard to manage.. could we encode the logic in querytype classes instead (with fields that must be set, eg. handler, classification, fallback etc)

I was in half-mind, so your nudge was helpful :).  I've added a dataclass with all the fields, and have a single map to set them.

pradh (commented):
Thanks for the review!

PR #2227, opened 2023-02-10
[disasters] remove extreme temperature

beets:
Temporarily remove this from the page while we improve data quality.


jehangiramjad (approved):

pradh (commented):
Is there some way to see this data as we iterate on data improvements?

pradh (approved):

chejennifer (approved):

beets:
We could iterate locally by modifying the config. Or use the NL config fork for it, since it would need to include all event types

pradh:
> We could iterate locally by modifying the config. Or use the NL config fork for it, since it would need to include all event types

Fork and use NL sounds great!

PR #2228, opened 2023-02-13
Show full date range in histogram tile

juliawu:
Removes logic that truncates histogram to the last date we have data for. This makes the histograms show bars for every month in the range. As a nice side effect, all histogram bars will have the same width if the same time setting is used.

Before ("last year" setting):
Screenshot 2023-02-13 at 10 22 30 AM
Screenshot 2023-02-13 at 10 22 39 AM

After ("last year" setting):
Screenshot 2023-02-13 at 10 23 40 AM
Screenshot 2023-02-13 at 10 23 49 AM


beets (approved):
thanks for the quick fix!

PR #2229, opened 2023-02-13
[disasters] draw geojsons for drought area and storm path

chejennifer:
- add new flask endpoint for getting geojsons for a list of entities and a specific geojson prop
- fix and add new choropleth flask endpoint tests (some tests were not being run and those tests had some bugs)
- add new drawPolygonLayer and drawPathLayer functions & refactor some of the map drawing functions

storm
drought


static/js/chart/draw_map_utils.ts
```diff
@@ -254,5 +254,5 @@ export function getPlacePathId(placeDcid: string): string {
   if (_.isEmpty(placeDcid)) {
     return "";
   }
-  return placeDcid.replace("/", "-");
+  return placeDcid.replace(/\//g, "-");
```
beets, line 257: use `.replaceAll` instead of a regex (a slightly cheaper operation)

static/js/components/tiles/disaster_event_map_tile.tsx
```diff
@@ -104,15 +141,22 @@ export function DisasterEventMapTile(
...
   useEffect(() => {
-    // re-draw map if placeInfo, geoJsonData, or disasterEventData changes
-    if (
-      shouldShowMap &&
-      !_.isEqual(props.disasterEventData, prevDisasterEventData.current)
-    ) {
-      prevDisasterEventData.current = props.disasterEventData;
-      draw(placeInfo, geoJsonData, props.disasterEventData);
+    if (shouldShowMap) {
+      draw(
```
beets, line 145: we would ideally draw things in order of
polygons -> paths -> points
and avoid redrawing the map unnecessarily

we could do this by calling raise after all the relevant draw functions, or draw only after all geojson data has been fetched. otherwise mouse overs etc would be hard to do.

beets (commented):
this looks great - thanks for the test fixes too

had a question on draw ordering in the tile. we'll need to get the layering right so the map is still interactive on the Earth page.

static/js/components/tiles/disaster_event_map_tile.tsx
chejennifer: we currently already draw the map only after all geojson data has been fetched.

I don't think raise is needed because svg elements are drawn in the order they appear in the html (so last elements are drawn on top) - I've shifted the layer adding to add polygons, then add paths, then add points.

static/js/chart/draw_map_utils.ts
chejennifer: done

static/js/components/tiles/disaster_event_map_tile.tsx
beets: great, missed the shouldShowMap update.

static/tsconfig.json
```diff
@@ -5,7 +5,7 @@
         "name": "typescript-tslint-plugin",
       }
     ],
-    "lib": ["es2019"],
+    "lib": ["ES2021"],
```
beets, line 8: was this an intended update? why was it required

beets (approved):
thanks for the updates!

static/tsconfig.json
beets: oh, it's for replaceAll.

chejennifer: yes, intended for the replaceAll!

chejennifer:
thanks for the review!

PR #2230, opened 2023-02-13
[disaster] small ui fixes

beets:
top event tile:
- add links to dates if place & name are not available
- don't show NaN

event map info card:
- add box shadow
- only show 3 significant digits
- don't show NaN
- TODO: bring in event spec so we can use display name & units

image
image

### DIFF
--- a/static/js/components/tiles/disaster_event_map_info_card.tsx
+++ b/static/js/components/tiles/disaster_event_map_info_card.tsx
@@ -16,7 +16,7 @@
 
 /**
  * Component for the content of the info card.
- * TODO: Use display properties from spec.
+ * TODO: Use display properties & unit from spec.
  */
 import React from "react";
 
--- a/static/js/components/tiles/disaster_event_map_info_card.tsx
+++ b/static/js/components/tiles/disaster_event_map_info_card.tsx
@@ -33,7 +33,6 @@ interface DisasterEventMapInfoCardPropType {
 export function DisasterEventMapInfoCard(
   props: DisasterEventMapInfoCardPropType
 ): JSX.Element {
-  debugger;
   return (
     <div className="disaster-event-map-info-card-content">
       <div className="disaster-event-map-info-card-header">
--- a/static/js/components/tiles/disaster_event_map_info_card.tsx
+++ b/static/js/components/tiles/disaster_event_map_info_card.tsx
@@ -20,8 +20,8 @@
  */
 import React from "react";
 
-import { formatNumber } from "../../utils/string_utils";
 import { DisasterEventPoint } from "../../types/disaster_event_map_types";
+import { formatNumber } from "../../utils/string_utils";
 
 interface DisasterEventMapInfoCardPropType {
   // The event data to show info about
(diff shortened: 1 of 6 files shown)


static/js/components/tiles/disaster_event_map_info_card.tsx
```diff
@@ -46,17 +48,25 @@ export function DisasterEventMapInfoCard(
         {props.eventData.endDate && (
           <span>End Date: {props.eventData.endDate}</span>
         )}
+        {props.eventData.severity &&
```
chejennifer, line 51: should be props.eventData.displayProps?

static/js/components/tiles/disaster_event_map_info_card.tsx
```diff
@@ -46,17 +48,25 @@ export function DisasterEventMapInfoCard(
...
+          Object.keys(props.eventData.displayProps).map((prop) => {
+            return (
+              <span key={prop}>
+                {prop}: {formatNumber(props.eventData.severity[prop])}
```
chejennifer, line 55: should be props.eventData.displayProps[prop]?

static/js/components/tiles/top_event_tile.tsx
```diff
@@ -323,4 +310,22 @@ export function TopEventTile(props: TopEventTilePropType): JSX.Element {
...
+  function addEventLink(eventId: string, displayStr: string): JSX.Element {
+    return <a href={`/browser/${eventId}`}>{displayStr}</a>;
+  }
+
+  function getDisplayDate(event: DisasterEventPoint): string {
+    let ret = "";
+    if (props.topEventMetadata.showStartDate && event.startDate) {
+      ret += formatDateString(event.startDate);
+      if (props.topEventMetadata.showEndDate && event.endDate) {
+        ret += " - ";
```
chejennifer, line 323: Should we use something besides " - " because it could look a little confusing since the date strings use "-" as well? so it might become something like "2020-01-01-2022-01-01"

chejennifer (commented):
thanks!

static/js/components/tiles/top_event_tile.tsx
chejennifer: > Should we use something besides " - " because it could look a little confusing since the date strings use "-" as well (could end up with something like "2020-01-01-2022-01-01")

static/js/components/tiles/disaster_event_map_info_card.tsx
beets: thanks, fixed

static/js/components/tiles/top_event_tile.tsx
beets: changed to an emdash. wanted something visually simpler
image

beets:
updated - ptal, thanks!

static/js/components/tiles/top_event_tile.tsx
chejennifer: nice, emdash looks good!

chejennifer (approved):
thanks for the changes!

juliawu (approved):
LGTM!

PR #2234, opened 2023-02-13
[disasters] add z-index to tooltips

beets:
fixes tooltips which could sometimes go outside the bounds of the map

image


chejennifer (approved):
thanks for the fix!

PR #2237, opened 2023-02-14
[map] fix bangladesh AA1 map

chejennifer:
- one of the AA1 regions were not being drawn because we have multiple property values for it.


beets (approved):
thanks for the fix!

PR #2239, opened 2023-02-14
[disaster] display tweaks

beets:
requested changes:
- add end date for droughts
- remove end date for fires (temporarily)

also added paths and polygons to main map

image


chejennifer (approved):
thanks!

PR #2240, opened 2023-02-14
[disasters] small page optimizations

beets:
- add gzip compression to /geojson and /event-data
- add caching to /event-data
- increase lower limit of fire and flood severity

(third bar from top is floods, bottom bar is geojson. both for earth)

without compression:
image

after compression:

compression level: 1 (flood - 2.85s waiting, wildfire - 2.09s, drought - 2.03s)
image

compression level: 3 (flood - 2.74s waiting. wildfire - 2.10s, drought - 2.35s)
image

caching without compression: download time 171.44ms
image

caching with compression (level 3): download time 121.96ms
image


server/routes/api/choropleth.py
```diff
@@ -241,7 +241,7 @@ def geojson():
           "current_geo": place_dcid
       }
   }
-  return Response(json.dumps(result), 200, mimetype='application/json')
+  return lib_util.gzip_compress_response(result, is_json=True)
```
chejennifer, line 244: maybe we should also use gzip for entity-geojson?

chejennifer (approved):
this is cool! Just curious, are there no client side changes needed to read the compressed data?

server/routes/api/choropleth.py
beets: it didn't seem as necessary yet. it would be more effective to add caching, but it's now a POST. we can do that as a follow up

beets:
> this is cool! Just curious, are there no client side changes needed to read the compressed data?

there isn't! it's part of the http spec

PR #2241, opened 2023-02-15
[disaster] update getting places for events in top event tile

chejennifer:
- move the propvals flask endpoint from /stats (which is stat var tool specific) to /nodes so it can be shared
- update getting the place for events in top event tile to look at the event's affectedPlaces first, if not found, then use latitude/longitude to try to figure out the affected place

This fixes some of the N/A places:

autopush:

autopush

with the change:

local


server/routes/api/node.py
```diff
@@ -34,3 +36,12 @@ def triples(direction, dcid):
   if direction != "in" and direction != "out":
     return "Invalid direction provided, please use 'in' or 'out'", 400
   return dc.triples(dcid, direction).get("triples", {})
+
+
+@bp.route('/propvals')
```
beets, line 41: please add caching

static/js/components/tiles/top_event_tile.tsx
```diff
@@ -235,10 +241,74 @@ export function TopEventTile(props: TopEventTilePropType): JSX.Element {
...
       .catch(() => {
-        setEventPlaces({});
+        return {};
+      });
+  }
+
+  // Given a list of events, update eventPlaces with a map of event id to place
+  // that the event occurs within. Places will be of the same type as
+  // props.enclosedPlaceType.
+  function updateEventPlaces(events: DisasterEventPoint[]): void {
```
beets, line 254: ideally we'd do this fallback in flask so this can be done in a single request, or better yet, in the initial call to fetch event-data. could you add a todo here for now?

beets (approved):

server/routes/api/node.py
chejennifer: done

static/js/components/tiles/top_event_tile.tsx
chejennifer: done

PR #2242, opened 2023-02-15
Drop related diseases for "blood pressure" query

pradh:
Handle that by skipping related SVs unless it is for query_type SIMPLE.

Ranking:

image

Simple:

image


server/integration_tests/test_data/demo_feb2023/query_2/chart_config.json
```diff
@@ -11,7 +11,6 @@
                     "title": "Most severe fires in California",
                     "topEventTileSpec": {
                       "eventTypeKey": "fire",
```
shifucun, line 13: why "showEndDate" is deleted?

shifucun (approved):

server/integration_tests/test_data/demo_feb2023/query_2/chart_config.json
beets: we had really long events, and this was deleted in #2239. but the tests didn't catch that diff..

pradh:
Thanks!

PR #2243, opened 2023-02-15
Fix nlnext redirects with trailing / and query

pradh:
Locally, with this fix http://localhost:8080/nlnext/#q=query+string redirects to http://localhost:8080/nl/#q=query+string


shifucun (commented):
Does /nlnext (without trailing slash) still work? I wonder if we have shared link both w/o trailing slash.

beets:
we could keep both? with and without slash..

pradh:
> Does /nlnext (without trailing slash) still work? I wonder if we have shared link both w/o trailing slash.

Yes, it works with and without trailing slash.

pradh:
> we could keep both? with and without slash..

Thanks, I didn't realize we can have multiple.  I did test it works with and without slash, but added another one anyway to make it extra obvious.

shifucun (approved):

PR #2245, opened 2023-02-16
Fix google maps scrolling issue

juliawu:
After PR #2189, google map tiles with KML coordinates drawn don't let the user scroll or pan.

This PR removes the listener added as a quick-fix, and instead fixes a CSS bug that caused map heights to sometimes be set to 0.

Before:

https://user-images.githubusercontent.com/4034366/219223889-0125501a-70d1-42c0-b91a-a4750ed525bb.mov

After:

https://user-images.githubusercontent.com/4034366/219223926-39c46b17-2ead-406e-a728-f0d7d05d4121.mov

Also tested the changes on the following pages:

Place pages:
* San Francisco City
* California
* USA (no maps expected)
* India (no maps expected)

Browser pages:
* Mountain View
* Santa Clara County
* Alaska
* India

NL interface:
* "Tell me about palo alto"
* "Tell me about Santa Clara County"
* "Tell me about Nebraska"
* "Tell me about USA"

### DIFF
--- a/static/css/shared/_tiles.scss
+++ b/static/css/shared/_tiles.scss
@@ -342,7 +342,7 @@ $box-shadow-8dp: 0 8px 10px 1px rgba(0, 0, 0, .14), 0 3px 14px 2px rgba(0, 0, 0,
   }
 }
 
-#map-container {
+.map-container {
   height: 100%;
   min-height: 200px;
 }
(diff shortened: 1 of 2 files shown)


chejennifer (approved):
thanks for the fix!

PR #2250, opened 2023-02-16
[Disasters] reduce fires

beets:
Increase fire lower limit further for all places (and override to 500 sq km for Earth)
Update color of droughts

before:
image

after:
image


pradh (approved):
Thanks!

PR #2252, opened 2023-02-16
[disasters] use own place geojson if children are missing

chejennifer:
- If a place doesn't have geojson for its children, use its own geojson as the base map

autopush: 

australiaAutopush

with the change:

austrailaLocal

### DIFF
--- a/static/js/components/subject_page/main_pane.tsx
+++ b/static/js/components/subject_page/main_pane.tsx
@@ -25,7 +25,7 @@ import { CATEGORY_ID_PREFIX } from "../../constants/subject_page_constants";
 import { SVG_CHART_HEIGHT } from "../../constants/tile_constants";
 import { NamedPlace, NamedTypedPlace } from "../../shared/types";
 import { SubjectPageConfig } from "../../types/subject_page_proto_types";
-import { fetchEntityGeoJson } from "../../utils/geojson_utils";
+import { fetchNodeGeoJson } from "../../utils/geojson_utils";
 import { fetchGeoJsonData, getId } from "../../utils/subject_page_utils";
 import { ErrorBoundary } from "../error_boundary";
 import { Category } from "./category";
@@ -88,7 +88,7 @@ export const SubjectPageMainPane = memo(function SubjectPageMainPane(
         break;
       }
     }
-    const placeGeoJsonPromise = fetchEntityGeoJson(
+    const placeGeoJsonPromise = fetchNodeGeoJson(
       [props.place.dcid],
       placeGeoJsonProp
     );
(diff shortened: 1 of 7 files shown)


juliawu (approved):
Yay maps!

static/js/components/subject_page/main_pane.tsx
```diff
@@ -63,9 +76,25 @@ export const SubjectPageMainPane = memo(function SubjectPageMainPane(
...
+      props.parentPlaces
+    );
+    let placeGeoJsonProp = "";
+    for (const type of props.place.types) {
+      if (type in PLACE_TYPE_GEOJSON_PROP) {
+        placeGeoJsonProp = PLACE_TYPE_GEOJSON_PROP[type];
+        break;
+      }
+    }
+    const placeGeoJsonPromise = fetchEntityGeoJson(
```
beets, line 91: we might want to rename the function (and route)

static/js/components/tiles/disaster_event_map_tile.tsx
```diff
@@ -341,10 +351,14 @@ export function DisasterEventMapTile(
       (geoDcid: GeoJsonFeatureProperties) =>
         redirectAction(geoDcid.geoDcid) /* redirectAction */,
       (place: NamedPlace) => place.name || place.dcid /* getTooltipHtml */,
-      () => true /* canClickRegion: allow all regions to be clickable */,
+      () =>
+        !isPlaceBaseMap /* canClickRegion: allow all regions to be clickable if not using place base map */,
```
beets, line 355: as well as if it's the same place (on leaf maps, we currently allow clicks to self which resets the view, etc. we should disallow clicks).

static/js/components/tiles/disaster_event_map_tile.tsx
```diff
@@ -70,6 +71,8 @@ const CSS_SELECTOR_PREFIX = "disaster-event-map";
 const REDIRECT_URL_PREFIX = "/disasters/";
 const MAP_POINTS_MIN_RADIUS = 1.5;
 const MAP_POINTS_MIN_RADIUS_EARTH = 0.8;
+// Set of dcids of places that should use the selected place as the base geojson
+const PLACE_MAP_PLACES = new Set(["country/AUS"]);
```
beets, line 75: do we need this map if we have the fallback in place?

beets (commented):
thanks!

static/js/components/subject_page/main_pane.tsx
chejennifer: since this function/route fetches geojson for places or events or any list of "entities", I went with a general name. What would you suggest?

static/js/components/tiles/disaster_event_map_tile.tsx
chejennifer: yes, we need this because we actually do get a few places when getting children for country/AUS ..

static/js/components/subject_page/main_pane.tsx
chejennifer: done - updated to NodeGeoJson

PR #2254, opened 2023-02-16
[ranking tile] add state code for USA place name

chejennifer:
add state code when states differ across the list of places:

stateCode

don't add state code if all places are from the same state:

stateCode2

### DIFF
--- a/static/js/utils/place_utils.ts
+++ b/static/js/utils/place_utils.ts
@@ -21,6 +21,7 @@ import {
   EARTH_NAMED_TYPED_PLACE,
   IPCC_PLACE_50_TYPE_DCID,
 } from "../shared/constants";
+import { DisplayNameApiResponse } from "../shared/stat_types";
 import {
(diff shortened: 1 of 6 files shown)


beets (approved):
cool, thanks

static/js/utils/place_utils.ts
```diff
@@ -196,6 +197,26 @@ export function getPlaceNames(
     });
 }
 
+/**
+ * Given a list of place dcids, returns a promise with a map of dcids to place
+ * display names (adds state code to the end of the name if available)
```
juliawu, line 202: Please clarify the difference between `getPlaceNames()` and `getPlaceDisplayNames()`

juliawu (approved):
Nice!

static/js/utils/place_utils.ts
chejennifer: done

PR #2268, opened 2023-02-17
Fix a ranking-tile issue

pradh:
Could be causing: https://screenshot.googleplex.com/9dTkAw9iMeypSEi

I'm not sure since still trying to get local-env setup with custom DC


beets (approved):
nice catch

PR #2270, opened 2023-02-17
[ranking] add support for multi-column layout

beets:
Top is new multi-col format, bottom is current (still works)
Ranks by the last specified SV

Added a new 'dev' topic for tile spec development / testing.

TODO: fulfillment updates

image


static/js/components/ranking_unit.tsx
```diff
@@ -48,6 +49,8 @@ interface RankingUnitPropType {
    * If true, only the ranks and labels will be shown and not the values.
    */
   hideValue?: boolean;
+  // For multi-column, these are the display strings for the value columns, in order.
+  svNames?: string[];
```
chejennifer, line 53: instead of having 3 separate arrays for unit, scaling, and svNames where the values at each index has to match up, might be cleaner to have a single `columns?: { unit: string, scaling: number, name: string }[]`.

To avoid relying on array index, could even include an "svkey" for each column object & have RankingPoint.values be a map of svKey to value. 

These might be bigger changes so maybe add a todo if you agree that this is worthwhile

static/js/components/ranking_unit.tsx
```diff
@@ -93,24 +109,50 @@ export function RankingUnit(props: RankingUnitPropType): JSX.Element {
...
                     <span
                       className={`num-value ${
                         point.placeDcid === props.highlightedDcid ? "bold" : ""
                       }`}
                     >
                       {formatNumber(
-                        props.scaling
-                          ? point.value * props.scaling
+                        props.scaling &&
+                          props.scaling.length &&
```
chejennifer, line 121: nit: instead of checking props.scaling && props.scaling.length, can just check !_.isEmpty(props.scaling)? Same with props.unit

static/js/components/tiles/ranking_tile.tsx
```diff
@@ -269,6 +243,84 @@ function fetchData(
...
+    const placeToVal = {};
+    const points = rankingData[sv].points;
+    for (const p of points) {
+      placeToVal[p.placeDcid] = p.value;
+    }
+    return placeToVal;
+  });
+  for (const p of sortedPlacePoints) {
+    p.value = undefined;
+    p.values = svs.map((_, i) => svsToDict[i][p.placeDcid]);
```
chejennifer, line 264: could just do `svsToDict.map(svDict => svDict[p.placeDcid])`?

static/js/components/tiles/ranking_tile.tsx
```diff
@@ -269,6 +243,84 @@ function fetchData(
...
+  }
+  for (const sv of svs) {
```
chejennifer, line 266: could we just do:

```
rankingData[sortSv].unit = statVarSpecs.map((spec) => spec.unit);
rankingData[sortSv].scaling = statVarSpecs.map((spec) => spec.scaling);
return { [sortSv]: rankingData[sortSv] };
```

chejennifer (approved):
this is cool! Just a few suggestions

static/js/components/ranking_unit.tsx
beets: i'll add a TODO

static/js/components/ranking_unit.tsx
beets: done

static/js/components/tiles/ranking_tile.tsx
beets: we still need the index (since svsToDict is a list also, not a dict).

static/js/components/tiles/ranking_tile.tsx
beets: done

beets:
thanks for the review!

pradh (approved):
Thanks for this!

static/js/components/tiles/ranking_tile.tsx
chejennifer: not sure why the index is needed? if i understand correctly, svsToDict is just a list of { [placeDcid]: value } where svsToDict[i] corresponds to svs[i] .. so we can just map each item in svsToDict to the value in that dict for p.placeDcid?

PR #2271, opened 2023-02-17
[NL] Display fixes to overviews with map only

juliawu:
Fixes the emtpy whitespace on overview tiles in the NL interface when no ranking data is available.

## Before:
Screenshot 2023-02-17 at 2 04 25 PM

## After:
Screenshot 2023-02-17 at 2 04 39 PM

## Validations
Also verified changes don't break other NL queries, the place pages, or demo scripts:

Screenshot 2023-02-17 at 2 05 29 PM
Screenshot 2023-02-17 at 2 05 59 PM
Screenshot 2023-02-17 at 2 06 13 PM
Screenshot 2023-02-17 at 2 06 28 PM
(no map expected for USA)

### DIFF
--- a/static/css/nl_interface.scss
+++ b/static/css/nl_interface.scss
@@ -193,7 +193,7 @@ $answer-min-height: 100vh;
       padding: 0.25rem;
     }
 
-    .map-with-margin {
+    .map-with-padding {
       padding-right: 15px !important;
     }
 
--- a/static/css/nl_interface.scss
+++ b/static/css/nl_interface.scss
@@ -194,8 +194,7 @@ $answer-min-height: 100vh;
     }
 
     .map-with-margin {
-      margin-right: 15px;
-      height: 100%;
+      padding-right: 15px !important;
     }
 
     .map-container {
--- a/static/css/nl_interface.scss
+++ b/static/css/nl_interface.scss
@@ -193,9 +193,13 @@ $answer-min-height: 100vh;
       padding: 0.25rem;
     }
 
+    .map-with-margin {
+      margin-right: 15px;
+      height: 100%;
+    }
+
     .map-container {
       height: 100%;
-      margin-right: 15px;
     }
   }
 }
--- a/static/css/place/place.scss
+++ b/static/css/place/place.scss
@@ -102,10 +102,6 @@ h3 {
   text-decoration: underline;
 }
 
-.map-with-margin {
-  height: 100%;
-}
-
 .map-container {
   height: 100%;
   min-height: 160px;
--- a/static/css/place/place.scss
+++ b/static/css/place/place.scss
@@ -102,6 +102,10 @@ h3 {
   text-decoration: underline;
 }
 
+.map-with-margin {
+  height: 100%;
+}
+
 .map-container {
   height: 100%;
   min-height: 160px;
--- a/static/js/place/overview.tsx
+++ b/static/js/place/overview.tsx
@@ -53,13 +53,15 @@ class Overview extends React.Component<OverviewPropType, OverviewStateType> {
     return (
       <section className="factoid col-12">
         <div className="row">
-          {!_.isEmpty(this.state.rankingData.label) ? (
+          <div
+            className={`col-12 ${this.shouldShowRanking() ? "col-md-4" : ""}`}
+          >
+            <div className={this.shouldShowRanking() ? "map-with-margin" : ""}>
+              <GoogleMap dcid={this.props.dcid}></GoogleMap>
+            </div>
+          </div>
+          {this.shouldShowRanking() ? (
             <>
-              <div className="col-12 col-md-4">
-                <div className="map-with-margin">
-                  <GoogleMap dcid={this.props.dcid}></GoogleMap>
-                </div>
-              </div>
               <div className="col-12 col-md-8">
                 <Ranking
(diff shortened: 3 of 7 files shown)


static/js/place/overview.tsx
```diff
@@ -30,24 +32,66 @@ interface OverviewPropType {
...
-              locale={this.props.locale}
-            ></Ranking>
+          <div
+            className={`col-12 ${this.shouldShowRanking() ? "col-md-4" : ""}`}
+          >
+            <div className={this.shouldShowRanking() ? "map-with-margin" : ""}>
+              <GoogleMap dcid={this.props.dcid}></GoogleMap>
+            </div>
           </div>
+          {this.shouldShowRanking() ? (
```
chejennifer, line 63: can make this check once at the top of the render function & save it in a constant : `const shouldShowRanking = this.shouldShowRanking()`

static/js/place/overview.tsx
```diff
@@ -30,24 +32,66 @@ interface OverviewPropType {
...
+      .then((resp) => {
+        this.setState({ rankingData: resp.data });
+      })
+      .catch((error) => {
+        console.log(error);
+      });
+  }
+
+  shouldShowRanking(): boolean {
+    return !_.isEmpty(this.state.rankingData.label);
```
chejennifer, line 93: nit: this doesn't need its own function

static/js/place/overview.tsx
```diff
@@ -30,24 +32,66 @@ interface OverviewPropType {
...
+              <div className="col-12 col-md-8">
+                <Ranking
+                  dcid={this.props.dcid}
+                  locale={this.props.locale}
+                  data={this.state.rankingData}
+                ></Ranking>
+              </div>
+            </>
+          ) : (
+            <></>
```
chejennifer, line 74: this doesn't need to be an if else check, can just do `shouldShowRanking && <div ...><Ranking .../></div> ` which says only render the ranking div if shouldShowRanking

static/css/nl_interface.scss
```diff
@@ -193,9 +193,13 @@ $answer-min-height: 100vh;
       padding: 0.25rem;
     }
 
+    .map-with-margin {
+      margin-right: 15px;
+      height: 100%;
```
chejennifer, line 198: do we need this `height:100%?` isn't height always going to be 100% with or without the ranking section

static/js/place/overview.tsx
```diff
@@ -30,24 +32,66 @@ interface OverviewPropType {
...
-          </div>
-          <div className="col-12 col-md-8">
-            <Ranking
-              dcid={this.props.dcid}
-              locale={this.props.locale}
-            ></Ranking>
+          <div
+            className={`col-12 ${this.shouldShowRanking() ? "col-md-4" : ""}`}
+          >
+            <div className={this.shouldShowRanking() ? "map-with-margin" : ""}>
```
chejennifer, line 59: do we need this extra div? can we just do `col-12 ${this.shouldShowRanking() ? "col-md-4 map-with-margin" : ""}` in line 57

chejennifer (commented):
thanks!

static/js/place/overview.tsx
juliawu: When I tried to combine map-with-margin in the same div as col-md-4, the rankings would show up under the map instead
image

static/css/nl_interface.scss
juliawu: When height isn't set here, the maps on the place pages don't take the full height (instead, they're stuck at 160px min height).

static/js/place/overview.tsx
juliawu: good catch! Switched to using &&

static/js/place/overview.tsx
juliawu: Ah, this is a much better idea than the helper function I tried to use instead! Switched to saving a constant at the top.

static/js/place/overview.tsx
juliawu: removed the function

static/js/place/overview.tsx
juliawu: Played around with the CSS more and was able to remove the extra div!

static/css/nl_interface.scss
juliawu: Played around with this more: now that I removed the extra div, height is no longer set here.

static/css/place/place.scss
```diff
@@ -102,6 +102,10 @@ h3 {
   text-decoration: underline;
 }
 
+.map-with-margin {
+  height: 100%;
```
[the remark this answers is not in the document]
juliawu: CSS has been cleaned up (and renamed to map-with-padding because padding is now used instead.

juliawu:
Some updated screenshots after CSS changes. I saw no changes to place pages in my testing:

Screenshot 2023-02-17 at 7 01 45 PM
Screenshot 2023-02-17 at 7 02 00 PM

beets:
@pradh - fyi: this also updates "Learn more about" to "Learn about" due to the way components are laid out. we can merge if this becomes an issue (or hold off to avoid code changes). otherwise should be ready for merging.

juliawu:
These changes superseded by PR #2281

PR #2272, opened 2023-02-17
[nl] add multi-col ranking table for temperature query

beets:
also added markdown to footer. footnotes now expect newline separators.

image


pradh (approved):
Nice!

Could you run `./run_tests.sh -g` then `./run_tests.sh -i` to update the integration test?

chejennifer (approved):
nice!

PR #2273, opened 2023-02-18
[NL] add units for percent svs

chejennifer:
sv2unit

### DIFF
--- a/server/lib/nl/page_config_builder.py
+++ b/server/lib/nl/page_config_builder.py
@@ -100,6 +100,7 @@ def build_page_config(
     all_svs.update(cspec.svs)
   all_svs = list(all_svs)
   sv2name = utils.get_sv_name(all_svs)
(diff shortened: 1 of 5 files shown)


server/lib/nl/page_config_builder.py
```diff
@@ -100,6 +100,7 @@ def build_page_config(
     all_svs.update(cspec.svs)
   all_svs = list(all_svs)
   sv2name = utils.get_sv_name(all_svs)
+  sv2unit = utils.get_sv_unit(all_svs)
```
pradh, line 103: Add a TODO to extend to map and ranking too?

pradh (approved):
Very nice!

server/lib/nl/page_config_builder.py
chejennifer: done

PR #2276, opened 2023-02-18
[nl] add support for overriding places on any tile **

beets:
** except disaster tiles
** found out too late that there's getNamedTypedPlace, but that should be updated to use this new endpoint. will merge later since that touches a lot of other stuff

image


pradh (approved):
Ooh nice!

Are we also adding the highlight to every line chart?  Or only when there's a single line?

beets:
The highlight is a separate tile that we can add if we'd like. Was just using it for testing :) They all support the place override now.

PR #2277, opened 2023-02-18
Implement a time-delta across places fulfiller

pradh:
This has a similar approach as the previous TIME_DELTA_ACROSS_VARS.  It leverages https://github.com/datacommonsorg/website/pull/2276.

Examples:

image

image


shifucun (approved):
Nice!!

PR #2278, opened 2023-02-18
Add more per-capita exclusions

pradh:


juliawu (approved):
Thanks!

pradh:
> Thanks!

Thanks.  Made one more tweak to match full DCIDs for something like Count_Person, and updated golden.

PR #2280, opened 2023-02-21
In the absence of both SV and place, avoid reading from context

pradh:
This translate to:
- For SIMPLE case, either SV or place has to be present in the current query.
- For OVERVIEW case, place has to be present in the current query.


juliawu (approved):
This looks good! There might be some edge cases where we might want inheriting from context ("what else can you tell me?") but this is good for now.

pradh:
> This looks good! There might be some edge cases where we might want inheriting from context ("what else can you tell me?") but this is good for now.

Thanks.  Maybe we might special case that in future, or ask to be more specific "what else can you tell me about palo alto"

PR #2281, opened 2023-02-21
[NL] Fix blank place overview for non-US places.

juliawu:
Fix blank space in the place overview tile when querying for non-US places. This PR fixes the same issue as, and supersedes #2271. It uses simpler logic which should be more robust. In particular we:
* Only show ranking tables for places in the US (geoId/ dcids)
* Updates CSS to fix spacing issues when ranking tables are missing
* Changes flavor text to "Learn about" instead of "Learn more about" per offline discussion.

Before:
image

After:
Screenshot 2023-02-21 at 10 03 01 AM

Other Screenshots:
Screenshot 2023-02-21 at 10 03 16 AM
Screenshot 2023-02-21 at 10 03 30 AM
Screenshot 2023-02-21 at 10 03 48 AM
3d-453b-850c-e0438343e9e7.png)

Also confir
Screenshot 2023-02-21 at 9 47 57 AM
med changes don't break place pages:
Screenshot 2023-02-21 at 9 48 09 AM
Screenshot 2023-02-21 at 9 48 28 AM
Screenshot 2023-02-21 at 9 48 48 AM
Screenshot 2023-02-21 at 9 49 12 AM


static/js/place/overview.tsx
```diff
@@ -32,18 +32,26 @@ interface OverviewPropType {
...
   render(): JSX.Element {
+    // only show ranking for places in the US, but not the US itself.
+    const shouldShowRanking = this.props.dcid.startsWith("geoId/");
     return (
       <section className="factoid col-12">
         <div className="row">
-          <div className="col-12 col-md-4">
+          <div
+            className={`col-12 ${
+              shouldShowRanking && "col-md-4 map-with-padding"
```
chejennifer, line 42: nit: I'm not sure "map-with-padding" is the best class to add here because this class will be added for maps on place overview page and wherever else Overview is rendered. However, only the nl interface will actually have padding for maps with this class. 

Maybe instead, we can add a "overview-with-ranking" class to line 38 & in the nl_interface.scss, we can do `.overview-with-ranking .map-container { margin-right: 15px }`

chejennifer (commented):
thanks for the changes!

static/js/place/overview.tsx
juliawu: Switched to using overview-with-ranking as you suggested! Didn't see any changes to the screenshots in my testing.

chejennifer (approved):
thanks for the changes!

static/js/components/tiles/place_overview_tile.tsx
```diff
@@ -61,7 +61,7 @@ export function PlaceOverviewTile(
       </div>
       {!_.isEmpty(subtopics) && (
         <div className="subtopics-section">
-          <h3>Learn more about {props.place.name}:</h3>
+          <h3>Learn about {props.place.name}:</h3>
```
beets, line 64: possible to keep the "more" if it's a US place?

static/js/place/overview.tsx
```diff
@@ -32,18 +32,26 @@ interface OverviewPropType {
 
 class Overview extends React.Component<OverviewPropType> {
   render(): JSX.Element {
+    // only show ranking for places in the US, but not the US itself.
+    const shouldShowRanking = this.props.dcid.startsWith("geoId/");
```
beets, line 36: how about passing this in from the tile (add a prop to this component, and let the test for geoId/ stay in the tile).

beets (commented):
thanks for the update - small nits

static/js/place/overview.tsx
juliawu: Made the change to pass this in from the tile.

static/js/components/tiles/place_overview_tile.tsx
juliawu: Now keeps "more" if it's a US place.

juliawu:
Updated screenshots of the NL interface. No visible changes to the place pages.

Non-US place:
Screenshot 2023-02-22 at 3 18 08 PM


US place:
Screenshot 2023-02-22 at 3 18 25 PM

static/js/place/main_pane.tsx
```diff
@@ -132,7 +132,11 @@ class MainPane extends React.Component<MainPanePropType> {
     return (
       <RawIntlProvider value={intl}>
         {this.showOverview() && (
-          <Overview dcid={this.props.dcid} locale={this.props.locale} />
+          <Overview
+            dcid={this.props.dcid}
+            showRanking={this.props.dcid.startsWith("geoId/")}
```
chejennifer, line 137: maybe just set this as true to keep original behavior of place page

juliawu: Done!

beets (approved):
thanks for the updates

PR #2282, opened 2023-02-21
[Ranking TIles] add source to ranking tiles

chejennifer:
singleSVRanking
multiSVRanking


pradh:
Thank you for the quick change!

Could you try the 2 demo scripts locally once (except the first 2 queries in that require Magic Eye specific data)?

pradh (approved):

chejennifer:
> Thank you for the quick change!
> 
> Could you try the 2 demo scripts locally once (except the first 2 queries in that require Magic Eye specific data)?

Done, the 2 demo scripts work!

PR #2284, opened 2023-02-21
[nl] Add mode where enter has to be clicked to move on to next url prompt

chejennifer:
- update running prompts from url to only automatically run if there is `&a=true` in the url. Otherwise, wait for enter to be pressed after each prompt

Note: This is kind of hacky and not very clean, added TODO to clean up the code


pradh (approved):
Great job pulling off a hairly looking change in short notice!  Lets validate on autopush :)

PR #2286, opened 2023-02-22
[Custom DC] Terraform api key fix, save bt automation status in tf state

Fructokinase:
Currently API keys are duplicated on re-runs. Fix this by converting shell scripted "null_resource" into actual Terraform resource. Original reason Terraform resource was commented out was because of a bug, which is now fixed.

Also save BT automation terraform status to tf state. 

TESTED=https://buoyant-country-377016-datacommons.com/place/geoId/0667000


shifucun (approved):
Nice! Curious what is the original bug?

Fructokinase:
> Nice! Curious what is the original bug?

The original bug for Terraform api key was in the provider code. The original bug for our api key issue was that we pipe into a file in tmp, but during re-runs, if the file was not deleted we piped into the tmp file with existing api key without overwriting it, hence the invalid key issue.

PR #2287, opened 2023-02-22
[nl] Some cleanup for the url prompt handling

chejennifer:
- clean up code so that autorun and non autorun shares the logic for inputting queries from the url into the search box

screen-recording-_72_


static/js/apps/nl_interface/app.tsx
```diff
@@ -34,6 +34,9 @@ export function App(): JSX.Element {
   const [contextList, setContextList] = useState<any[]>([]);
   const urlPrompts = useRef(getUrlPrompts());
   const autoRun = useRef(getUrlToken("a"));
+  const inputIntervalTimer = useRef(null);
+  const nextPromptDelayTimer = useRef(null);
+  const searchDelayTimer = useRef(null);
```
beets, line 39: some documentation for these?

static/js/apps/nl_interface/app.tsx
```diff
@@ -34,6 +34,9 @@ export function App(): JSX.Element {
   const [contextList, setContextList] = useState<any[]>([]);
   const urlPrompts = useRef(getUrlPrompts());
   const autoRun = useRef(getUrlToken("a"));
```
beets, line 36: curious why autorun needs to be a ref? the value shouldn't change once the page has loaded (or rather, we won't need to listen to changes..)

beets (approved):
thanks - this is much cleaner!

static/js/apps/nl_interface/app.tsx
chejennifer: done

static/js/apps/nl_interface/app.tsx
chejennifer: oh yeah, this doesn't need to be a ref anymore, removed the ref part!

PR #2288, opened 2023-02-22
[Disaster Dashboard] Fix date bug that blocks rendering in Safari

juliawu:
This PR fixes the bug that prevented the disaster dashboard from rendering correctly in Safari.

The culprit was the date parsing for the histograms. Safari does not support using "Z" in a datestring for Zulu Time. Since the histogram tile only ever uses dates and doesn't need timezone information specified, this PR switches "Z" to using `Date.parse()` instead, which according to [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/parse) will interpret date-only formatted dates as UTC.

Before:
Screenshot 2023-02-21 at 4 59 34 PM

After:
Screenshot 2023-02-21 at 4 37 13 PM
Screenshot 2023-02-21 at 4 37 32 PM
Screenshot 2023-02-21 at 4 37 40 PM
Screenshot 2023-02-21 at 4 37 47 PM
Screenshot 2023-02-21 at 4 37 52 PM
Screenshot 2023-02-21 at 4 37 58 PM


chejennifer (commented):
nice find! Probably also need to update these, I'm guessing the date drop down on Safari also doesn't render properly because of the same problem: https://github.com/datacommonsorg/website/blob/master/static/js/utils/disaster_event_map_utils.tsx#L116-L117

beets (approved):
nice find!

beets:
> nice find! Probably also need to update these, I'm guessing the date drop down on Safari also doesn't render properly because of the same problem: https://github.com/datacommonsorg/website/blob/master/static/js/utils/disaster_event_map_utils.tsx#L116-L117

yes - let's do that in a follow up

PR #2289, opened 2023-02-22
Add a flag to explicitly log query

shifucun:


beets (approved):
should we add this to magic-eye too?

server/configmodule.py
```diff
@@ -40,6 +40,10 @@ class Config:
   API_PROJECT = ''
   # Set this to False if the deployment has frequently updated data.
   USE_MEMCACHE = True
+  # Whether to log the query (and make them avaiable in query history interface)
+  # Eanble this to "True" requires adding "bigtable/user" acccess for the
```
pradh, line 44: nit: Enabling

shifucun: Done

shifucun:
> should we add this to magic-eye too?

For custom DC, will need to use the Bigtable in the custom project. I will do that in follow up changes.

PR #2292, opened 2023-02-22
[nl] Add feedback link

chejennifer:
screen-recording-_74_

- also fixed an error about `<tr>` can not be a child of a `<td>`

### DIFF
--- a/static/js/apps/nl_interface/debug_info.tsx
+++ b/static/js/apps/nl_interface/debug_info.tsx
@@ -24,6 +24,9 @@ import { Col, Row } from "reactstrap";
 
 import { DebugInfo, SVScores } from "../../types/app/nl_interface_types";
 
+const FEEDBACK_LINK =
+  "https://docs.google.com/forms/d/e/1FAIpQLSe9SG0hOfrK7UBiOkQbK0ieC0yP5v-8gnQKU3mSIyzqdv6WaQ/viewform?usp=pp_url&entry.2003307180=";
+
 const svToSentences = (
   svScores: SVScores,
   svSentences: Map<string, Array<string>>
@@ -122,6 +125,8 @@ export function DebugInfo(props: DebugInfoProps): JSX.Element {
   const toggleShowDebug = () => {
     setShowDebug(!showDebug);
   };
+  // Google forms prefilled link uses + for spaces.
+  const feedbackLinkQuery = debugInfo.originalQuery.replaceAll(" ", "+");
 
   return (
     <>
@@ -135,6 +140,11 @@ export function DebugInfo(props: DebugInfoProps): JSX.Element {
           <a className="debug-info-toggle hide" onClick={toggleShowDebug}>
             X
           </a>
+          <Row className="feedback-link">
+            <a href={`${FEEDBACK_LINK}${feedbackLinkQuery}`} target="_blank">
+              Share Feedback
+            </a>
+          </Row>
           <Row>
             <b>DEBUGGING OPTIONS/INFO: </b>
             <br></br>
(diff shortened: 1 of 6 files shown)


static/js/apps/nl_interface/debug_info.tsx
```diff
@@ -18,12 +18,20 @@
...
 import { Col, Row } from "reactstrap";
 
 import { DebugInfo, SVScores } from "../../types/app/nl_interface_types";
 
+const FEEDBACK_LINK =
+  "https://docs.google.com/forms/d/e/1FAIpQLSfqndIayVhN1bN5oeZT0Te-MhhBMBR1hn97Lgr77QTOpga8Iw/viewform?usp=pp_url";
+const QUERY_PARAM_PREFIX = "&entry.1322830239=";
+const SOURCE_PARAM_PREFIX = "&entry.1070482700=";
+const VERSION_PARAM_PREFIX = "&entry.1420739572=";
+const QUERY_CHAIN_PARAM_PREFIX = "&entry.1836374054=";
```
pradh, line 33: Document where these numbers come from?

pradh (commented):
nice!

shifucun:
Is the feedback for normal users? In that case, should it be under the chart box like it is in Google Search?

static/js/apps/nl_interface/debug_info.tsx
chejennifer: done

chejennifer:
> Is the feedback for normal users? In that case, should it be under the chart box like it is in Google Search?

I think first it will just be for the team, but I assume it will be for normal users as well. In the thread (on the core chat) about this, both options of in the debug panel and under the charts were suggested & debug panel seemed to be the one chosen.

static/js/apps/nl_interface/debug_info.tsx
```diff
@@ -18,12 +18,22 @@
...
 
+import { debug } from "console";
 import _ from "lodash";
 import React, { useState } from "react";
 import { Col, Row } from "reactstrap";
 
 import { DebugInfo, SVScores } from "../../types/app/nl_interface_types";
 
+const FEEDBACK_LINK =
+  "https://docs.google.com/forms/d/e/1FAIpQLSfqndIayVhN1bN5oeZT0Te-MhhBMBR1hn97Lgr77QTOpga8Iw/viewform?usp=pp_url";
```
juliawu, line 29: Suggestion: can we also add a `JSON.stringify(DebugInfo)` pre-filled answer to the form?

static/js/apps/nl_interface/debug_info.tsx
```diff
@@ -86,6 +98,26 @@ const matchScoresElement = (svScores: SVScores): JSX.Element => {
...
+    [QUERY_CHAIN_PARAM_PREFIX]: JSON.stringify(queryChain),
+  };
+  let link = FEEDBACK_LINK;
+  Object.keys(paramMap).forEach((prefix) => {
+    const value = paramMap[prefix];
+    if (!value) {
+      return;
+    }
+    // when prefilling google forms, need to replace space with "+"
+    link += `${prefix}${value.replaceAll(" ", "+")}`;
```
beets, line 116: consider calling encodeURI instead, which would also take care of other html special chars

beets (approved):
cool!

static/js/apps/nl_interface/debug_info.tsx
chejennifer: I originally wanted to include debugInfo, but unfortunately it is too many characters to pass in the URL

pradh: debug-info might be too big.  its perhaps better suited for the BT log Bo has set up?

static/js/apps/nl_interface/debug_info.tsx
chejennifer: done

PR #2296, opened 2023-02-23
Track query sessions and log session information to BT

pradh:
Return the full query list in the session if the first query was OK.  The frontend still does not show all the queries.

Tested/validated locally.


server/lib/nl/fulfillment/context.py
```diff
@@ -132,3 +133,22 @@ def classifications_of_type_from_utterance(
     uttr: Utterance,
     ctype: ClassificationType) -> List[ClassificationAttributes]:
   return [cl for cl in uttr.classifications if cl.type == ctype]
+
+
+def get_session_info(context_history: List[Dict]) -> Dict:
```
shifucun, line 138: could context history contains context for past queries (not in this session)?

server/lib/nl/utterance.py
```diff
@@ -263,16 +267,17 @@ def load_utterance(uttr_dicts: List[Dict]) -> Utterance:
...
+        query=udict['query'],
+        query_type=QueryType(udict['query_type']),
+        svs=udict['svs'],
+        places=_dict_to_place(udict['places']),
+        classifications=_dict_to_classification(udict['classifications']),
+        rankedCharts=_dict_to_chart_spec(udict['ranked_charts']),
+        detection=None,
+        chartCandidates=None,
+        answerPlaces=None,
+        session_id=udict.get('session_id', constants.TEST_SESSION_ID))
```
shifucun, line 281: why use test session id here?

shifucun (approved):

pradh (commented):
Thanks!

PR #2298, opened 2023-02-23
Increase context/session length

pradh:
The context length also determines a session.  Make it longer than our current demo sequences.


shifucun (approved):
I might not fully understand session..., is it for all the queries after a page load? Do we know only show the first one of those in the history?

pradh:
> I might not fully understand session..., is it for all the queries after a page load? Do we know only show the first one of those in the history?

Yes, it is all queries after a page-load (limited by this number)

PR #2299, opened 2023-02-23
Fix bug with query history dedupe

pradh:


shifucun (approved):

beets (approved):

static/js/apps/nl_interface/query_history.tsx
```diff
@@ -47,7 +47,7 @@ export function QueryHistory(): JSX.Element {
     axios.get(`/nl/history`).then((resp) => {
       const result = [];
       for (const item of resp.data) {
-        if (result.includes(item["query_list"])) {
+        if (result.includes(item["query_list"][0])) {
```
chejennifer, line 50: will `item["query_list"]` ever be empty? If so, we should check before accessing the first element

chejennifer (approved):

static/js/apps/nl_interface/query_history.tsx
pradh: Good call, done.

PR #2301, opened 2023-02-23
[NL] change to light background

beets:
image
image


static/css/nl_interface.scss
```diff
@@ -206,6 +225,15 @@ $answer-min-height: 100vh;
   }
 }
 
+.nl-query h2 {
+  color: $question-text-color;
+}
+
+.form-control:focus {
+  // Override the red border which looks like an error.
+  border-color: var(--dark) !important;
```
chejennifer, line 234: maybe also remove the box-shadow/outline because it still looks a little red? https://github.com/datacommonsorg/website/blob/master/static/css/import_wizard2.scss#L38-L43

chejennifer (approved):
thanks for the change!

static/css/nl_interface.scss
beets: how about adding it as an override to base.scss?

beets: (kept the box shadow but altered the color - it's a nice visual feedback for where the focus is)

image
image

chejennifer: yes, that looks good!

chejennifer (approved):
thanks for updating!

PR #2302, opened 2023-02-23
[nl] search box change

chejennifer:
start search box in the center of the screen and move it to the bottom after first query searched

screen-recording-_75_


beets (approved):
cool! 
another way would have been to use a decorator class, with judicious use of `!important`, which is totally fine for cases like this.

pradh:
nit:  From the gif, is it the case that the textbox moves a bit after "Recent queries" gets rendered?  If so, and if there are more queries will that look a bit unsettling for it to move after its rendered?

beets:
That's a good catch! I missed it -- we should fix the position of the
search box so it doesn't jump up.

chejennifer:
good catch, updated:

screen-recording-_76_

static/css/nl_interface.scss
```diff
@@ -48,19 +49,19 @@ $answer-min-height: 100vh;
...
-  transform: translate(0, 50%);
+  position: absolute;
+  top: calc(50% - #{$search-section-bottom-height});
 }
 
 #search-section-container-bottom {
   position: fixed;
-  bottom: 0;
-  transition: bottom 225ms cubic-bezier(0, 0, 0.2, 1) 0ms;
+  top: calc(100% - #{$search-section-bottom-height});
```
beets, line 62: does it still work to set bottom (since there isn't anything else setting it's top anymore). setting `bottom: 0` is much cleaner than the calculations

beets (approved):
thanks for the update!

static/css/nl_interface.scss
chejennifer: setting bottom won't work with the transition because bottom won't be set in the original state ..

PR #2304, opened 2023-02-23
Fix query history component render check

shifucun:


static/js/apps/nl_interface/query_history.tsx
```diff
@@ -32,7 +32,7 @@ export function QueryHistory(): JSX.Element {
 
   return (
     <>
-      {history && (
+      {history && history.length > 0 && (
```
chejennifer, line 35: can just check `!_.isEmpty(history)`?

chejennifer (approved):
good catch!

static/js/apps/nl_interface/query_history.tsx
shifucun: Done

PR #2306, opened 2023-02-23
Use flask env name to look for tools landing page config json

shifucun:
In the future, this allows using /custom_dc/... files for pure custom DC without touching the code in routes/static.py


beets (approved):
very cool!

do you think it's worth wrapping the routes in flask caching now that we're doing extra io reads?

shifucun:
> very cool!
> 
> do you think it's worth wrapping the routes in flask caching now that we're doing extra io reads?

Flask cache will use Redis (for prod) which is network io, and the json read here is local io, which could be faster? Also it should be same as reading it from Jinja template with "% include"?

PR #2308, opened 2023-02-23
[nl] fix font color for place overview tile

chejennifer:
on autopush:
autopushFont

with the change:
localFont

### DIFF
--- a/static/css/nl_interface.scss
+++ b/static/css/nl_interface.scss
@@ -237,6 +237,10 @@ $search-section-bottom-height: 7rem;
       height: 100%;
     }
   }
+
+  .subtopics-section a {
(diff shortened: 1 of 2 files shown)


static/css/nl_interface.scss
```diff
@@ -237,6 +237,10 @@ $search-section-bottom-height: 7rem;
       height: 100%;
     }
   }
+
+  .subtopics-section a {
```
beets, line 241: could you add this to the group in line 256

beets (approved):
thanks for the fix!

static/css/nl_interface.scss
chejennifer: done

PR #2310, opened 2023-02-24
Generalize html template path finding based on flask env name

shifucun:
Moved examples.json file out of custom_dc/default since they belong to base templates.


server/routes/browser.py
```diff
@@ -28,8 +29,10 @@
 
 @bp.route('/')
 def browser_main():
-  if g.env == 'iitm':
-    return render_template('custom_dc/iitm/browser_landing.html')
+  template_file = os.path.join('custom_dc', g.env, 'browser_landing.html')
+  if os.path.exists(
+      os.path.join(current_app.root_path, 'templates', template_file)):
+    return render_template(template_file)
```
beets, line 35: since this is a repeated pattern, perhaps move this into a new utility in lib_customdc?

beets (approved):

PR #2313, opened 2023-02-24
Add default places for certain child types; fix min-temp ranking bug

pradh:
Default Place:

image

Min Temp fix:

image


beets (approved):

pradh:
Thanks!

PR #2314, opened 2023-02-24
[SV hierarchy] Only get sv/svgs with data for a minimum number of places

chejennifer:
- for map and scatter tools, only show sv/svgs in the stat var selector that have 10 or more places with data


beets (approved):
nice!

PR #2316, opened 2023-02-24
[NL] add ability to click on full query history

beets:
will trigger the full stream of queries (on auto=true) if there was more than 1 query in the history


static/js/apps/nl_interface/query_history.tsx
```diff
@@ -41,14 +41,16 @@ export function QueryHistory(props: QueryHistoryProps): JSX.Element {
...
           <h3>Or try one of these recent queries:</h3>
-          {history.map((query, i) => {
+          {history.map((queries, i) => {
             return (
               <div
                 className="history-item"
                 key={i}
-                onClick={() => props.onItemClick(query)}
+                onClick={() => props.onItemClick(history[i])}
+                title={queries.join(", ")}
```
pradh, line 50: Mostly out of curiosity, where the `;` join happening?

pradh (approved):
Thanks!

static/js/apps/nl_interface/query_history.tsx
beets: this is just for display purposes. so a long hover over the text will show you the full query (with comma separation).

`onItemClick(history[i])` sends the full array of queries for that "session" to the parent which does the query-by-query handling.

pradh: thanks!

PR #2321, opened 2023-02-24
[tiles] add units for tiles

chejennifer:
- get units from data fetch for the tiles
Screenshot 2023-02-24 at 3 49 53 PM
Screenshot 2023-02-24 at 3 50 11 PM
Screenshot 2023-02-24 at 3 50 24 PM
Screenshot 2023-02-24 at 3 50 45 PM

### DIFF
--- a/static/js/components/tiles/bar_tile.tsx
+++ b/static/js/components/tiles/bar_tile.tsx
@@ -55,6 +55,7 @@ interface BarTilePropType {
 interface BarChartData {
   dataGroup: DataGroup[];
   sources: Set<string>;
+  unit: string;
 }
 
 export function BarTile(props: BarTilePropType): JSX.Element {
@@ -167,6 +168,7 @@ function processData(
   // Fetch the place names
   getPlaceNames(Array.from(popPoints).map((x) => x.placeDcid)).then(
     (placeNames) => {
+      let unit = "";
       for (const point of popPoints) {
         const placeDcid = point.placeDcid;
         const dataPoints: DataPoint[] = [];
@@ -183,6 +185,11 @@ function processData(
           };
           if (raw.facets[stat.facet]) {
             sources.add(raw.facets[stat.facet].provenanceUrl);
+            let svUnit = raw.facets[stat.facet].unit;
+            if (svUnit && spec.denom) {
+              svUnit += "per person";
+            }
+            unit = unit || svUnit;
           }
           if (spec.denom && spec.denom in raw.data) {
             const denomStat = raw.data[spec.denom][placeDcid];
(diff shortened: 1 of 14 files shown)


static/js/components/tiles/bar_tile.tsx
```diff
@@ -183,6 +185,11 @@ function processData(
           };
           if (raw.facets[stat.facet]) {
             sources.add(raw.facets[stat.facet].provenanceUrl);
+            let svUnit = raw.facets[stat.facet].unit;
+            if (svUnit && spec.denom) {
```
beets, line 189: for these, might want to check that "_Person" is in the denominator (we now also have area denominators)

static/js/components/tiles/line_tile.tsx
```diff
@@ -188,14 +190,23 @@ function rawToChart(
           dataPoints
         )
       );
+      let svUnit = raw.facets[series.facet].unit;
+      if (svUnit && spec.denom) {
```
beets, line 194: ditto

static/js/components/tiles/map_tile.tsx
```diff
@@ -263,6 +266,10 @@ function processData(
   if (_.isEmpty(dataValues)) {
     return;
   }
+  let unit = getUnit(Object.values(rawData.placeStat), rawData.metadataMap);
+  if (unit && isPerCapita) {
```
beets, line 270: ditto

beets (approved):
ok for today, but we'd need to fast follow and add a check for a "_Person" denominator before adding per person. might be worth pulling out into a util.

and we should add MTCO2e to [this util](https://github.com/datacommonsorg/website/blob/master/static/js/utils/string_utils.ts#L131)

beets:
also MetricTon --> t

pradh (approved):
If there are short-form overrides to apply, based on the unit DCID, like `MetricTonCO2e` -> `MTCO2e`, rather than the SV... should we have some way to do it?

pradh:
Very useful change!

pradh:
The per-capita charts still show `%`?

beets:
> If there are short-form overrides to apply, based on the unit DCID, like `MetricTonCO2e` -> `MTCO2e`, rather than the SV... should we have some way to do it?

yep - we do some unit rewrites in the frontend, though it's split for place explorer (internationalized) and the rest of the site...

chejennifer:
> The per-capita charts still show `%`?

The per capita charts had their units set when building the page config .. do we want to change that?

pradh:
> > The per-capita charts still show `%`?
> 
> The per capita charts had their units set when building the page config .. do we want to change that?

I see. We can fix as a follow on.

static/js/components/tiles/bar_tile.tsx
chejennifer: done, created a util function

PR #2322, opened 2023-02-25
Support both absolute and relative rankings for time-delta queries

pradh:
Also, reduce the number of tiles before "show more" from 6 to 3.

image

image


server/lib/nl/fulfillment/time_delta_across_places.py
```diff
@@ -88,26 +89,36 @@ def _populate_cb(state: PopulateState, chart_vars: ChartVars,
...
-  ranked_child_places = []
-  for d in ranked_child_dcids:
-    ranked_child_places.append(dcid2place[d])
+  utils.update_counter(
+      state.uttr.counters, 'time-delta_reranked_places', {
+          'orig': dcids,
+          'ranked_abs': ranked_children.abs,
+          'ranked_pct': ranked_children.pct,
+      })
+  block_id = chart_vars.block_id + 10
```
beets, line 105: curious why we need to add 10 for the block_id?

beets (approved):
nice!

server/lib/nl/fulfillment/time_delta_across_places.py
pradh: Mainly to avoid some other chart from reusing the same block_id.  Needed only for the 2nd block being introduced here.

chejennifer (approved):

PR #2324, opened 2023-02-25
Abstract a render function for static pages; Set up custom DC template folder.

shifucun:
All the files for "custom" env should live in <root>/custom_dc and are copied over to /server and /static by script and webpack.


pradh (approved):

shifucun:
Thanks!

run_server.sh
```diff
@@ -70,6 +71,10 @@ if [[ $FLASK_ENV != "local" ]]; then
 fi
 echo "Starting localhost with FLASK_ENV='$FLASK_ENV' on port='$PORT'"
 
+# Copy over custom DC files
+find ./server/templates/custom_dc/custom ! -name '.empty' -type f -exec rm -f {} +
+cp -rf custom_dc/*.html server/templates/custom_dc/custom/
```
beets, line 76: we should add a note in documentation that hot reloads won't work and they would need to rerun this script on each change.

server/lib/render.py
```diff
@@ -0,0 +1,31 @@
...
+# limitations under the License.
+
+import os
+
+from flask import current_app
+from flask import g
+from flask import render_template
+
+
+def render_page(default_html, custom_dc_html, **context):
```
beets, line 22: could you add more documentation on **context?

static/webpack.config.js
```diff
@@ -159,6 +159,7 @@ const config = {
         { from: "custom_dc/**/*" },
         { from: "*favicon.ico" },
         { from: "robots.txt" },
+        { from: "../custom_dc/", to: "custom_dc/custom" },
```
beets, line 162: is there a way to exclude the html templates from being copied over too?

beets (commented):
sorry for the late review..

instead of custom_dc/custom, maybe custom_dc/deploy?

PR #2325, opened 2023-02-25
allow different formatNumber functions to be used when drawing charts

chejennifer:
- Pass in the formatNumber function to use when drawing charts so that different formatNumber functions can be used between tools that need i18n and those that don't.
- Use the formatNumber function in string_utils for tiles. 

Screenshot 2023-02-24 at 8 11 40 PM
Screenshot 2023-02-24 at 8 11 48 PM


pradh (approved):
Thanks a lot for the fix!

pradh:
I'd like to enable auto-merge to get this in and test stuff out end-to-end.  Hope that's okay, and you'll do follow on PRs on any review comments.

beets:
@chejennifer: curious if this change was only for the different units available in the i18n version, or if there were other features required?

chejennifer:
> @chejennifer: curious if this change was only for the different units available in the i18n version, or if there were other features required?

it was to be able to use different displays strings (from i18n) for units & also to still return the unit string even if it doesn't have a specified display string

beets:
Got it. Then perhaps we should try to use i18n everywhere instead and only
have one to reduce complexity? Fwiw, the other was meant to be lightweight
and isn't used widely, so we could have brought that to parity as well.
Given that we were using the i18n formatNumber in our draw functions, it
wouldn't work for unit conversions without initializing the intl module so
was not working as intended on non-place pages anyway. So we could try
using i18n everywhere.

chejennifer:
hm, so do you propose we update all our tools to initialize the intl module? One other difference between the string_utils formatNumber and the i18n formatNumber is that i18n will skip adding the unit if the unit isn't one of the switch cases .. which I feel like for tools where internationalization doesn't matter, we should still be able to show the unit?

beets:
Yup, that's what I'm thinking. The differences between the two are subtle
and we should have done it earlier. We could update place explorer behavior
to always have units (since we rarely add charts there. when we do, we can
ensure that units are added).

chejennifer:
Ok that sounds good to me, will make that change

beets:
Thanks! Not high priority, but we should do it soon (and can split the
work).

PR #2326, opened 2023-02-25
Support per-capita for ranking charts. Also refactor time-delta chart gen.

pradh:
This is to support per-capita for ranking queries in the climate trace demo script. Also tweak the SV names a bit.

image


server/lib/nl/fulfillment/base.py
```diff
@@ -99,8 +103,10 @@ def add_chart_to_utterance(chart_type: ChartType, state: PopulateState,
   }
   if chart_vars.skip_map_for_ranking:
     attr['skip_map_for_ranking'] = True
-  if chart_vars.set_place_override_for_line:
-    attr['set_place_override_for_line'] = True
+  if chart_vars.growth_direction != None:
```
shifucun, line 106: nit: can just do "if chart_vars.growth_direction:"

shifucun (approved):

server/lib/nl/fulfillment/base.py
pradh: I'm trying to not be misled by growth_direction enum being 0 and thus not being recognized (ran into such issues)

pradh (commented):
Thanks!

PR #2332, opened 2023-02-26
Update deployment config and script so they can be used for custom DC

shifucun:


deploy/overlays/custom_kustomization.yaml.tpl
```diff
@@ -0,0 +1,54 @@
...
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+nameSuffix: -custom
+namespace: website
+
+resources:
+  - ../base
+
+configMapGenerator:
```
Fructokinase, line 25: Shall we disable config map suffix from now on? 

```
generatorOptions:
 disableNameSuffixHash: true
```

deploy/overlays/kustomization.yaml
```diff
@@ -0,0 +1,64 @@
+# Copyright 2022 Google LLC
```
Fructokinase, line 1: Was this meant to be under deploy/overlays/buoyant-country-377016/kustomization.yaml?

scripts/push_image.sh
```diff
@@ -17,12 +17,14 @@
 
 set -e
 
+PROJECT_ID=$1
```
Fructokinase, line 20: nit: add a check here?

deploy/overlays/custom_kustomization.yaml.tpl
shifucun: Done

deploy/overlays/kustomization.yaml
shifucun: This should not be checked in. Removed it.

scripts/push_image.sh
shifucun: Done

PR #2335, opened 2023-02-27
[Custom DC] Fix helm/kustomize diff: remove kustomize config suffix

Fructokinase:
Kustomize currently adds a suffix to all config maps deployed. It seems like there are some rationales for doing this (link below) but so far I'm not sure if it helped us. Since it is difficult for helm to match this, let's get rid of the configmap suffix?

https://pauldally.medium.com/why-is-kustomize-adding-a-bunch-of-characters-to-my-secret-configmap-names-b1572741c846

### DIFF
--- a/deploy/base/kustomization.yaml
+++ b/deploy/base/kustomization.yaml
@@ -26,6 +26,9 @@ resources:
   - deployment.yaml
   - namespace.yaml
 
+generatorOptions:
+ disableNameSuffixHash: true
+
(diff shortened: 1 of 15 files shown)


shifucun:
To confirm, is this change to remove this suffix: https://github.com/datacommonsorg/website/blob/master/deploy/overlays/autopush/kustomization.yaml#L23, or the random hash string for each kustomize build?

PR #2336, opened 2023-02-27
Fix a bug from formatNumber change

chejennifer:
- Missed a place where formatNumber function needed to be passed to drawHistogram


juliawu (approved):
Thanks for the fix!

PR #2338, opened 2023-02-27
Updates and fixes for custom DC UI/deployment

shifucun:
deploy_gke.sh is updated with named args, instead of positional args now. All references should be updated.

Added more html template and place holder file for change

### DIFF
--- a/scripts/deploy_gke.sh
+++ b/scripts/deploy_gke.sh
@@ -96,8 +96,8 @@ fi
 CLUSTER_NAME=$CLUSTER_PREFIX-$REGION
 
 # Deploy to GKE
-kustomize edit set image gcr.io/$IMAGE_PROJECT/datacommons-website=gcr.io/$IMAGE_PROJECT/datacommons-website:$WEBSITE_HASH
-kustomize edit set image gcr.io/$IMAGE_PROJECT/datacommons-nl=gcr.io/$IMAGE_PROJECT/datacommons-nl:$WEBSITE_HASH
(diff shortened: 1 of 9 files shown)


scripts/deploy_gke.sh
```diff
@@ -96,8 +96,8 @@ fi
 CLUSTER_NAME=$CLUSTER_PREFIX-$REGION
 
 # Deploy to GKE
-kustomize edit set image gcr.io/$IMAGE_PROJECT/datacommons-website=gcr.io/$IMAGE_PROJECT/datacommons-website:$WEBSITE_HASH
-kustomize edit set image gcr.io/$IMAGE_PROJECT/datacommons-nl=gcr.io/$IMAGE_PROJECT/datacommons-nl:$WEBSITE_HASH
+kustomize edit set image gcr.io/datcom-ci/datacommons-website=gcr.io/$IMAGE_PROJECT/datacommons-website:$WEBSITE_HASH
+kustomize edit set image gcr.io/datcom-ci/datacommons-nl=gcr.io/$IMAGE_PROJECT/datacommons-nl:$WEBSITE_HASH
```
beets, line 100: hm, so do we no longer need the IMAGE_PROJECT var set above?

beets (approved):
could you add more description to the PR around how this fixes autopush? thanks!

scripts/deploy_gke.sh
shifucun: It's still used, but only in the replaced string. `kustomize edit set image <default> <new>` where the default on is always datcom-ci.

shifucun:
Thanks for review!

PR #2339, opened 2023-02-27
move feedback link outside of debug panel

chejennifer:
- move the feedback link outside of the debug panel and style it like the one on the place page

Screenshot 2023-02-27 at 2 21 04 PM

### DIFF
--- a/static/css/nl_interface.scss
+++ b/static/css/nl_interface.scss
@@ -132,8 +132,11 @@ $search-section-bottom-height: 7rem;
   }
 }
 
-.nl-query-error,
 .nl-query {
+ padding-top: 2rem; 
+}
+
+.nl-query-error {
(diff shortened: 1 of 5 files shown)


static/css/nl_interface.scss
```diff
@@ -168,6 +171,21 @@ $search-section-bottom-height: 7rem;
...
   }
 
+  .container.feedback-link {
+    background-color: transparent;
+    display: flex;
+    justify-content: flex-end;
+    min-height: 1rem;
+
+    a {
+      color: #70757a;
```
beets, line 181: could you add a var for this color?

beets (approved):
cool, thanks!

static/css/nl_interface.scss
chejennifer: done

PR #2340, opened 2023-02-27
[nl] fix small window size css

chejennifer:
fix nl interface layout on small window sizes (width < 992px). The problem was that nl interface css depends on the styling for the containing divs from here: https://github.com/datacommonsorg/website/blob/master/static/css/tools/base_tools.scss#L50-L66, but this styling is only applied when page size is lg which is when min-width = 992px

autopush:
Screenshot 2023-02-27 at 3 17 35 PM

with the change:
Screenshot 2023-02-27 at 3 13 27 PM

### DIFF
--- a/static/css/nl_interface.scss
+++ b/static/css/nl_interface.scss
@@ -35,10 +35,19 @@ $answer-min-height: 100vh;
 $text-color: inherit;
 $search-section-bottom-height: 7rem;
 
+#main {
+  display: flex;
+  flex-direction: column;
+  height: 100vh;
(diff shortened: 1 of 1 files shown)


beets (approved):
thanks for the fix!

beets:
btw, is there another bug on your localhost from the ranking table (with very many significant digits)

chejennifer:
> btw, is there another bug on your localhost from the ranking table (with very many significant digits)

ah yes, for some reason the string_utils.ts file wasn't updated on my localhost

PR #2341, opened 2023-02-28
Track SV strings in github and add index building code

pradh:
As part of updating the index, devs show now also send CSV diffs so we can know what exactly is changing.

The `build_embeddings.py` is a straightforward port from the [notebook here](https://colab.sandbox.google.com/drive/1s2qYkbjZ6ZXoDfYTWWkk5SMgEVQFBAtK#scrollTo=lzmyZYF9C5wv).

Note that the canonical version is still the spreadsheet.  I originally wanted the checked-in file to be that, but the current sheet has some funky string concat across the descriptions that should rationalize first.

We might also need more stuff here as we scale to 10K SVs, so this is just an incremental improvement.


tools/nl/embeddings/build_embeddings.py
```diff
@@ -0,0 +1,187 @@
...
+
+    print("Building embeddings")
+    embeddings = build_embeddings(ctx, texts, dcids)
+
+    print("Saving locally to " + csv_out_path)
+    embeddings.to_csv(csv_out_path, index=False)
+
+    # Also upload to the NL embeddings server's GCS bucket
+    print("Attempting to write to GCS bucket:")
+    print(f"\t file: gs://{FLAGS.bucket_name}/{csv_out_name_date}")
```
jehangiramjad, line 155: nit: since this is the last line that gets printed and the README asks to capture the version number, could we add some text that explicitly says something like:

"Embeddings Filename: <filename>"

And this <filename> should just be {csv_out_name_date} and not the full gcs path.

jehangiramjad (approved):
Thanks!

beets (approved):
nice - thanks for adding the run script too

tools/nl/embeddings/build_embeddings.py
pradh: Good call, done.

pradh:
There is a change in the golden file for [What countries in Africa had the greatest increase in life expectancy].  This comes about due to https://github.com/datacommonsorg/website/pull/2255.  I'm going ahead with the submission to unblock this.  

I've requested @jehangiramjad know to double check the change.

New:
image

Old:
image

pradh:
[Palo Alto] is broken on autopush/dev/magiceye/climate-trace/everywhere now.  Filed https://github.com/datacommonsorg/website/issues/2348.  It is not really a regression in our code per se (as far as I can tell), but Maps API started returning a different place ID seemingly.

So I'm replacing it with SF in our one integration test query.

jehangiramjad:
Regarding the comment/question about the SV matching order change: great catch and the current behavior should be considered an improvement. With #2255 we improved stop words removal which was previously buggy (due to reasons captured in #2255 ). Here are the specifics:

**Query**: "What countries in Africa had the greatest increase in life expectancy"

**Old Behavior:** the SV search would use the following string after stop words removal: "increase life expectancy". Notice that the word "increase" should ideally have been removed but wasn't. Using "increase life expectancy" the top three SVs matched are: 1: Life_Expectancy, 2: LifeExpectancy_Person_Female, 3: LifeExpectancy_Person_Male. 
You can test this on the local servers by using this query: http://127.0.0.1:6060/api/search_sv/?q=increase%20life%20expectancy

**New Behavior:** the SV search would use the following string after stop words removal: "life expectancy". Notice that the word "increase" has now been removed (correctly) after #2255. Using "life expectancy" the top three SVs matched are: 1: Life_Expectancy, 2: LifeExpectancy_Person_Male, 3: LifeExpectancy_Person_Female. 
You can test this on the local servers by using this query: http://127.0.0.1:6060/api/search_sv/?q=life%20expectancy

**Conclusion:** this update to the goldens is correct and should be considered an improvement as part of changes brought about in #2255 .

PR #2342, opened 2023-02-28
Make helm resource naming consistent with existing kustomize naming

Fructokinase:


shifucun (approved):
Looks nice! There are some other variables which can potentially be made static as I went through the configs. Can try to improve them later.

PR #2343, opened 2023-02-28
deploy script fix

beets:
- typos in ./deploy_gke.sh
- update stanford-staging deploy folder to match ENV name, as required by script


shifucun (approved):
thanks for the fix!

PR #2344, opened 2023-02-28
[NL Interface] Detect and Resolve multiple places

jehangiramjad:
In this PR we have the following:

1. The PlaceDetection object should never be None. The main_place or places_found can be None or empty but the containing object should not be None. This was a TODO from before and needed to be fixed because in case the DCID/Maps ID resolution failed, even if the place detection on the query found the right place string, the debugInfo would show place detection failure (which could have been a Maps lookup failure, for example).

2. Ensure all place strings detected are now resolved (if possible) to get DCIDs. The PlaceDetection object still retains the main_place which is the first detected+resolved place in the query. However, places_found is now is a list of Place objects which can be used in correlation and other queries.

3. DebugInfo now shows a Places Resolved field which shows all the resolved places (name and dcid). The previous Places Found field corresponds to all the place strings found in the query (and should be set even if the place resolution fails).

Tested on many local queries + using one of the recent demo queries. All work as before. There should be no functional changes observed due to this PR. However, the debugInfo should show the new information. See examples below:

Screenshot 2023-02-28 at 1 43 34 PM

Screenshot 2023-02-28 at 1 46 43 PM

Screenshot 2023-02-28 at 1 44 58 PM

Screenshot 2023-02-28 at 1 44 00 PM

Screenshot 2023-02-28 at 1 47 23 PM


static/js/apps/nl_interface/debug_info.tsx
```diff
@@ -155,6 +156,9 @@ export function DebugInfo(props: DebugInfoProps): JSX.Element {
           <Row>
             <Col>Places Detected: {debugInfo.placesDetected.join(", ")}</Col>
           </Row>
+          <Row>
+            <Col>Places Resolved: {debugInfo.placesResolved}</Col>
+          </Row>
           <Row>
             <Col>
               Main Place Inferred: {debugInfo.mainPlaceName} (dcid:{" "}
```
pradh, line 164: nit: just "Main Place"?

server/routes/nl.py
```diff
@@ -110,23 +110,33 @@ def _remove_places(query, places_found):
...
   # Iterate over all the places until a valid place DCID is found.
-  for p_str in places_found:
+  for p_str in places_str_found:
+    place_dcid = ""
     # If this is a special place, return the known DCID.
     if p_str.lower() in constants.OVERRIDE_PLACE_TO_DICD_FOR_MAPS_API:
       place_dcid = constants.OVERRIDE_PLACE_TO_DICD_FOR_MAPS_API[p_str.lower()]
       logging.info(
-          f"{p_str} was found in OVERRIDE_PLACE_TO_DICD_FOR_MAPS_API. Returning its DICD {place_dcid} without querying Maps API."
+          f"{p_str} was found in OVERRIDE_PLACE_TO_DICD_FOR_MAPS_API. Recording its DICD {place_dcid} without querying Maps API."
```
pradh, line 136: DCID

server/routes/nl.py
```diff
@@ -229,23 +239,29 @@ def _result_with_debug_info(data_dict: Dict, status: str,
       'counters': debug_counters,
       'data_spec': uttr_history,
   }
-  if query_detection.places_detected:
+
+  places_found_formatted = ""
+  for place in query_detection.places_detected.places_found:
+    places_found_formatted += f" (name: {place.name}, dcid: {place.dcid}); "
```
pradh, line 245: nit: remove `;` there's `()` and space?

server/routes/nl.py
```diff
@@ -255,38 +271,41 @@ def _detection(orig_query, cleaned_query) -> Detection:
...
+  # Look to find place DCIDs.
+  if places_str_found:
+    place_dcids = _infer_place_dcids(places_str_found)
+    logging.info(f"Found {len(place_dcids)} place dcids: {place_dcids}.")
 
+  for place_dcid in place_dcids:
+    # Set the Place objects (dcid, name, type).
+    resolved_places.append(_get_place_from_dcid(place_dcid))
+    logging.info(
+        f"Resolved {len(resolved_places)} place dcids: {resolved_places}.")
```
pradh, line 295: Might be nice to pass all DCIDs in the DC calls?  (so sort of cut serial RPCs / BT round-trips)

server/routes/nl.py
```diff
@@ -255,38 +271,41 @@ def _detection(orig_query, cleaned_query) -> Detection:
...
-                                     main_place=Place(
-                                         dcid=place_dcid,
-                                         name=main_place_name,
-                                         place_type=main_place_type))
-  else:
-    query = cleaned_query
-    # TODO: even if no place_dcid was found, debugInfo should be able to display
-    # the places_found (which can be valid even if no dcid was found).
-    place_detection = None
+    query = _remove_places(cleaned_query.lower(), places_str_found)
```
pradh, line 299: If places_str_found is valid, but we failed to resolve, should we still do _remove_places() ?

server/routes/nl.py
```diff
@@ -255,38 +271,41 @@ def _detection(orig_query, cleaned_query) -> Detection:
...
+    main_place = resolved_places[0]
+    logging.info(f"Using main_place as: {main_place}")
+
+  # Set PlaceDetection.
+  place_detection = PlaceDetection(query_original=orig_query,
```
pradh, line 304: Wondering about tests for the multi-place thing... should we add multiple places in our integration test maybe?  Say: https://github.com/datacommonsorg/website/blob/master/server/integration_tests/nlnext_test.py#L143

And can update the corresponding debug_info...

pradh (commented):
Thanks!

server/lib/nl/detection.py
```diff
@@ -33,7 +33,8 @@ class PlaceDetection:
   """Various attributes of place detection."""
   query_original: str
   query_without_place_substr: str
-  places_found: List[str]
+  query_places_mentioned: List[str]
+  places_found: List[Place]
```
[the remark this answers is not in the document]
jehangiramjad: done

server/routes/nl.py
jehangiramjad: oops. done

static/js/apps/nl_interface/debug_info.tsx
jehangiramjad: done

server/routes/nl.py
jehangiramjad: i removed one extra space but kept the ; because in case we want to do a split later somewhere to show these nicely

server/routes/nl.py
jehangiramjad: Done. PTAL

server/routes/nl.py
jehangiramjad: Yes we should. Otherwise if there is any failure with this resolution, we won't remove those places and without that removal the SV search will become more unpredictable when matching to the embeddings..

server/routes/nl.py
jehangiramjad: Added some more richness to the tests. PTAL

server/routes/nl.py
```diff
@@ -110,12 +110,27 @@ def _remove_places(query, places_str_found: List[str]):
...
-  place_types = dc.property_values([place_dcid], 'typeOf')[place_dcid]
-  place_type = _get_preferred_type(place_types)
-  place_name = dc.property_values([place_dcid], 'name')[place_dcid][0]
+def _get_place_from_dcids(place_dcids: List[str]) -> List[Place]:
+  places = []
+  place_types_dict = dc.property_values(place_dcids, 'typeOf')
+  place_names_dict = dc.property_values(place_dcids, 'name')
+
+  for p_dcid, p_types in place_types_dict.items():
+    p_types = place_types_dict[p_dcid]
```
pradh, line 119: nit: don't need to lookup map again?

server/routes/nl.py
```diff
@@ -110,12 +110,27 @@ def _remove_places(query, places_str_found: List[str]):
...
+    p_type = _get_preferred_type(p_types)
+    p_name = ""
+    if p_dcid in place_names_dict:
+      p_name = place_names_dict[p_dcid][0]
+
+    if not p_name:
+      # This should not really be happening. But don't want to stop execution, so logging it.
+      logging.info(
+          f"Place DCID ({p_dcid}) did not correspond to a place name (using empty name)."
+      )
```
pradh, line 129: And continue?  To avoid name-less places on FE

server/routes/nl.py
pradh: So then, should _remove_places() be called outside the if condition?

server/routes/nl.py
pradh: LG, thanks.

pradh (approved):
Thanks for the changes!  One last issue around removing place names, otherwise LG

server/routes/nl.py
jehangiramjad: Good point. But also, upon some testing I realized that the order in which the propert_values API returns the results does not correspond to the input place_dcids. So now I am iterating over the place_dcids first and then checking for existence for both the type and name. The order  is important to prevent the main_place assignment to be somewhat deterministic so that the first place mentioned (and resolved) is the main place.

server/routes/nl.py
jehangiramjad: Good point. I have moved it to under a different if condition (which corresponds to places_str_found to be non-empty).

PR #2345, opened 2023-02-28
Fix a bunch of deployment issue

shifucun:
- clear ENV and PROJECT_ID before hand just in case they are set locally
- change feeding_america to feedingamerica, which was a regression from previous config refactor


Fructokinase (approved):

scripts/deploy_gke.sh
```diff
@@ -32,6 +32,9 @@ function help {
   exit 1
 }
 
+PROJECT_ID=""
+ENV=""
```
beets, line 36: maybe cleaner if we set new local variables instead of modifying the env?

https://tldp.org/LDP/abs/html/localvar.html

shifucun: Did some experiment with this. So ENV is local to this script and won't affect the environment variable external to this script (however, if not reset here, it will pick up the ENV value from external environment variable via "export").

shifucun: wrapping the option parsing will make the variable assignment within the scope of the parse function too, and I think the options parse should be in the script scope?

PR #2349, opened 2023-03-01
[Custom DC] Various setup improvements

Fructokinase:
Cluster
- Migrate the cluster management from shell script to Terraform module. This allows Terraform to also delete the GKE cluster, 1 step closer for E2E test
- Allow zonal cluster instead of enforcing regional clusters, which should cut the resource by 2/3. Location specifies whether a zonal or a regional cluster is created. ex: "us-cental1" -> regional, "us-central1-a" -> zonal. The difference is 99.95% SLA vs 99.5% SLA. I think most POCs/dev projects makes sense to be zonal.

Helm chart
- Allow gcr repo project to be specified, this is a follow up for @shifucun's support for building images in custom dc project
- Move helm out of Terraform management. In the past few weeks, issue 1 is that it was easy to get into a circular situation where the state needs to be read at the beginning (reading what resource a cluster has requires kube config), but the kube config is yet to be fetched because it happens in the part after reading the states. May be cleaner to separate GCP resource level and k8s resource levels.
- This does not affect GCP level resource lifecycle, as even if pods and services are in a cluster, the cluster can still be deleted.
- Taking the k8s resources out of the Terraform lifecycle may make things easier when integrating with Argo, as Argo changes will not touch Terraform at all.


Fructokinase:
I used this to deploy to RFF instance, but I'm not confident enough without E2E testing so the version is not bumped yet.

Fructokinase:
@juliawu Also adding you as the reviewer, hope you can familiarize yourself with some of the changes happening in custom DC!

deploy/helm_charts/dc_website/templates/ingress.yaml
```diff
@@ -29,10 +29,9 @@ kind: Ingress
 metadata:
   name: {{ .Values.ingress.name }}
   namespace: {{ .Values.namespace.name }}
-  {{- with .Values.ingress.annotations }}
   annotations:
```
shifucun, line 32: this is good, would be nice to have more static resources like this (so it's easier to know what it is)

deploy/helm_charts/dc_website/values.yaml
```diff
@@ -55,7 +56,10 @@ mixer:
   githash:
 
   image:
-    repository: gcr.io/datcom-ci/datacommons-mixer
+    registry: "gcr.io"
```
shifucun, line 59: registry and repository is not changed, maybe make them constant?

deploy/terraform-datacommons-website/modules/apikeys/main.tf
```diff
@@ -45,7 +45,8 @@ resource "google_secret_manager_secret" "maps_api_key_secret" {
   replication {
     user_managed {
       replicas {
-        location = var.location
+        # location needs to be a region here, so if var.location is a zone, make it a region.
+        location = length(split(var.location, "-")) == 2 ? var.location : join("-", slice(split( "-", var.location), 0, 2))
```
shifucun, line 49: wonder if there are tf test to ensure these expressions can be changed later?

deploy/terraform-datacommons-website/modules/gke/main.tf
```diff
@@ -13,24 +13,47 @@
...
 }
 
+resource "google_container_node_pool" "gke_node_pools" {
+  name       = local.cluster_name
+  location   = var.location
+  cluster    = google_container_cluster.primary.name
+  node_count = var.num_nodes
+
+  node_config {
+    machine_type = "e2-highmem-4"
```
shifucun, line 47: can you check if we can use a weaker machine_type?

shifucun (commented):
separate out helm and terraform is a good idea

deploy/helm_charts/dc_website/templates/ingress.yaml
Fructokinase: Agreed, will follow this style from now on!

deploy/helm_charts/dc_website/values.yaml
Fructokinase: done.

deploy/terraform-datacommons-website/modules/gke/main.tf
Fructokinase: Following our kustomize template, I'm changing the memory from website container 8G -> 3G, ESP 2G -> 1G. Which will be something like 12.75G occupied of 15.X G available in a e2-standard-4. 

However, the cost saving is only $30 or so (from ~$132 to ~$98). If there's anything that we shouldn't cheap out  on imo it's the machine type, because k8s under pressure will give all sorts of weird errors. We may also need more resources for new services. I think keeping this machine type is appropriate but will leave the final desicion to you.

deploy/terraform-datacommons-website/modules/apikeys/main.tf
Fructokinase: The closest thing I found was the Terraform console, for interactive expression testing. As for changing the expression, we'll likely have to test it on a live resource (which we will have to do anyways when we release)

https://developer.hashicorp.com/terraform/cli/commands/console

deploy/helm_charts/dc_website/templates/deployment.yaml
```diff
@@ -75,9 +75,9 @@ spec:
             periodSeconds: 10
           resources:
             limits:
-              memory: "8G"
+              memory: "3G"
```
juliawu, line 78: For my own understanding, why are we lowering memory limits, both here and in line 204&206?

deploy/terraform-datacommons-website/modules/gke/configure_cluster.sh
```diff
@@ -12,8 +12,13 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+if [[ $LOCATION =~ ^[a-z]+-[a-z0-9]+$ ]]; then
```
juliawu, line 15: nit: add a quick comment on what the regex is looking for

juliawu (commented):
Thanks for adding me, LGTM, just had some minor comments. Deferring approval to @shifucun.

deploy/terraform-datacommons-website/modules/apikeys/main.tf
shifucun: Then an e2e tests would be important

deploy/terraform-datacommons-website/modules/gke/main.tf
shifucun: Then sounds good to keep this. In that case, can revert the mem change above?

shifucun (approved):

deploy/terraform-datacommons-website/modules/gke/configure_cluster.sh
Fructokinase: done. thanks!

deploy/helm_charts/dc_website/templates/deployment.yaml
Fructokinase: We have two way of generating these yamls. Kustomize and helm (different tools). We deploy to autopush/staging/prod/stanford using kustomize, and custom DC uses helm. 

Since the [kustomize template](https://github.com/datacommonsorg/website/blob/master/deploy/base/deployment.yaml) uses 3G and 1G respectively, it means that it should be enough for custom DC as well. Since we're trying to save cost for custom DC users, we only have 1 machine for the entire GKE. I would like to conserve resources in case we need to deploy more things.

deploy/terraform-datacommons-website/modules/gke/main.tf
Fructokinase: Keeping the mem changes for now to conserve resources for other potential pods (like argo). Please see the response to Julia's comment.

deploy/helm_charts/dc_website/templates/deployment.yaml
shifucun: That makes sense, the flask server won't use too much memory, only the svg mixer needs high memory.

PR #2350, opened 2023-03-01
[tiles] add source for top event tiles

chejennifer:
Screenshot 2023-03-01 at 9 04 15 AM
Screenshot 2023-03-01 at 9 04 31 AM

- Also extract out a shared footer component


static/js/components/tiles/chart_footer.tsx
```diff
@@ -0,0 +1,85 @@
...
+
+import _ from "lodash";
+import React from "react";
+
+import { NL_SOURCE_REPLACEMENTS } from "../../constants/app/nl_interface_constants";
+import { urlToDomain } from "../../shared/util";
+import { isNlInterface } from "../../utils/nl_interface_utils";
+
+interface ChartFooterPropType {
+  sources: Set<string>;
```
beets, line 29: clarify with a comment here that these should be full urls?

beets (approved):
cool - thanks for the refactor!

static/js/components/tiles/chart_footer.tsx
chejennifer: done

PR #2353, opened 2023-03-01
[browser] hide fireperimeter and add flag to show hidden props

beets:
append `?all=1` to the url to show all hidden properties like fireperimeter, geojsoncoordinates, etc

/cc @ajaits


static/js/browser/arc_table_row.tsx
```diff
@@ -123,14 +123,16 @@ export class ArcTableRow extends React.Component<
...
+        {this.props.provenanceId && this.props.src ? (
           <td className="provenance-column">
             {this.props.provenanceId && (
               <a href={HREF_PREFIX + this.props.provenanceId}>
                 {this.props.src}
               </a>
             )}
           </td>
+        ) : (
+          <td></td>
```
beets, line 135: this fixes a DOM validation warning

static/js/browser/out_arc_section.tsx
```diff
@@ -64,6 +66,9 @@ export class OutArcSection extends React.Component<
   OutArcSectionPropType,
   OutArcSectionStateType
 > {
+  // Whether to show all properties
+  showAll: boolean;
```
chejennifer, line 70: I don't think this is used?

chejennifer (approved):
nice, this would be useful!

static/js/browser/out_arc_section.tsx
beets: ah, thanks for catching that!

PR #2354, opened 2023-03-02
Update Stanford DC with new tables; Fix local mixer api root issue

shifucun:


deploy/overlays/stanford/custom_bigtable_info.yaml
```diff
@@ -1,4 +1,7 @@
 project: datcom-stanford
 instance: dc-graph
 tables:
-  - private_2022_11_09_19_28_52
+  - smokepm_2023_03_01_15_05_49
+  - aqicn_2023_02_24_11_39_31
```
Fructokinase, line 5: This can be be updated to aqicn_2023_03_01_19_03_31

Fructokinase (approved):
Thanks! 

nit: would it help to have an extra option in run_server.sh that also sets USE_LOCAL_MIXER?

deploy/overlays/stanford/custom_bigtable_info.yaml
shifucun: Done

shifucun:
> Thanks!
> 
> nit: would it help to have an extra option in run_server.sh that also sets USE_LOCAL_MIXER?

Right, now there is "-l" to set this var

PR #2355, opened 2023-03-02
Use V1 API for bio page data fetch

shifucun:
Tested locally with charts matching autopush instance.


suhana13 (approved):
Thank you so much for the migration to V1, Bo! The changes look good to me!

PR #2358, opened 2023-03-03
Improve fallback behavior, add more place defaults, tweak time-delta patterns

pradh:
Misc smallish changes motivated by feedback queries.

(1) Avoid unexpected results that return variables/places from the context, if the current utterance has specified variables/places respectively.  For example, [counties with most black population in california] -> [obesity in california], will return black population in california.  There's an integration test for this.

(2) Add more patterns for time-delta queries.  [counties that have seen the most declines in median income].

(3) Add USA as a default for city, state and event fulfillment.

(4) Drop per-capita and handle "worst" / "best" ranking for unemployment-rate.

2-4 are illustrated below.

image

image

image


jehangiramjad (approved):
Thanks, this LG!

PR #2360, opened 2023-03-03
[disasters] make legend entries clickable on disaster map

chejennifer:
screen-recording-_77_


static/css/shared/_tiles.scss
```diff
@@ -194,13 +194,16 @@ $box-shadow-8dp: 0 8px 10px 1px rgba(0, 0, 0, .14), 0 3px 14px 2px rgba(0, 0, 0,
     align-items: center;
     padding: 0 1rem 0 0;
     font-size: .9rem;
+    cursor: default;
```
beets, line 197: did you mean `cursor: pointer`?

beets (approved):
nice thanks!

static/css/shared/_tiles.scss
chejennifer: I read in some articles that the pointer is supposed to be used for links that go to another page .. but I can make this a pointer if you think that's better

chejennifer: as discussed offline, added the pointer to be consistent with the rest of the website & added a background on hover

Screenshot 2023-03-06 at 10 38 02 AM

PR #2372, opened 2023-03-03
[Disaster dashboard] Add area histograms

juliawu:
Add histograms for "Area affected by [eventType]".

This PR:
* Adds ability for histograms to support aggregation of other property values, like area.
* Adds a 'prop' field in proto for Histogram tiles to specify which property to aggregate.
* Adds "Total area affected by [eventType]" chart for droughts, fires, and floods
* Adjusts addYAxis in draw to address a y-axis label issue that came up in development. Setting the font-family seems to cause the label to increase its size after rendering, leading to the labels becoming too big for the space. Discussed with @chejennifer, and left a TODO to find out why this happens and fix the root cause.

Tested on the disaster dashboard, and verified charts on place, browser, and statvar explorer were not affected by changes to addYAxis.

Screenshots: 
Screenshot 2023-03-03 at 11 50 27 AM

Screenshot 2023-03-03 at 11 50 38 AM

Screenshot 2023-03-03 at 11 50 54 AM


static/js/components/tiles/histogram_tile.tsx
```diff
@@ -18,7 +18,7 @@
  * Component for rendering a histogram type tile.
  */
 
-import _ from "lodash";
+import _, { property } from "lodash";
```
chejennifer, line 21: i don't think this `property` is being used?

static/js/components/tiles/histogram_tile.tsx
```diff
@@ -170,10 +174,20 @@ function binData(
     // Get start time in the temporal granularity desired
     const eventDate = event.startDate.slice(0, format.length);
 
-    // Increment count in corresponding bin if event has at least
+    // Get value of property to aggregate
+    let eventValue = 0;
```
chejennifer, line 178: could just default to 1 and remove the else statement?

static/js/components/tiles/histogram_tile.tsx
```diff
@@ -170,10 +174,20 @@ function binData(
...
+    if (property) {
+      // default to 0 if property can't be found in display props
+      eventValue = event.displayProps[property] ?? 0;
```
chejennifer, line 181: maybe use `||` to catch other false cases

chejennifer (commented):
nice!

server/config/subject_page.proto
```diff
@@ -115,7 +115,10 @@ message DisasterEventMapTileSpec {
 }
 
 message HistogramTileSpec {
+  // Key for event type specs defined in the PageMetadata that sohuld be displayed
```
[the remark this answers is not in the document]
juliawu: good catch! fixed.

static/js/chart/draw.ts
```diff
@@ -535,6 +535,7 @@ function addYAxis(
   chartWidth: number,
   yScale: d3.ScaleLinear<any, any>,
   formatNumberFn: (value: number, unit?: string) => string,
+  textFontFamily?: string,
```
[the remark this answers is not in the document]
juliawu: Thanks for pointing that out. Updated function docstring.

static/js/components/tiles/histogram_tile.tsx
juliawu: Thanks for the catch, that was auto-added by my IDE by accident. Removed.

static/js/components/tiles/histogram_tile.tsx
juliawu: Changed to ||.

static/js/components/tiles/histogram_tile.tsx
juliawu: I wrote it like this to handle the case where property="area" is set, but the corresponding displayProp was not found for the event type. In that case, all values are 0 and we won't show a broken histogram. If you think it's better to handle that case in a different way, let me know and I can switch to defaulting to 1 here.

beets (commented):
i am surprised by the fix to remove the font family for histograms - did you see if there are other differences with the drawHistogram function vs others? let's make sure to try fixing it and rolling back the change in draw.ts.

static/js/components/tiles/histogram_tile.tsx
juliawu: Gotcha, made the change!

static/js/chart/draw.ts
```diff
@@ -719,7 +736,14 @@ function drawStackBarChart(
     .nice()
     .rangeRound([chartHeight - MARGIN.bottom, MARGIN.top]);
 
-  const leftWidth = addYAxis(tempYAxis, chartWidth, y, formatNumberFn, unit);
+  const leftWidth = addYAxis(
+    tempYAxis,
+    chartWidth,
+    y,
+    formatNumberFn,
+    TEXT_FONT_FAMILY,
```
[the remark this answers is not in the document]
juliawu: I see it defined in line 65.

static/js/components/tiles/histogram_tile.tsx
```diff
@@ -269,20 +285,31 @@ export function HistogramTile(props: HistogramTilePropType): JSX.Element {
   function processData(
     disasterEventPoints: DisasterEventPoint[],
     dateSetting: string,
-    setHistogramData: (data: DataPoint[]) => void
+    setHistogramData: (data: DataPoint[]) => void,
+    property?: string
```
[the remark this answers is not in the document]
juliawu: Good idea. Done.

static/js/chart/draw.ts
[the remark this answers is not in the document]
beets: yes i found it too but didn't remove the comment..

beets (approved):
thanks - but lets continue to work on that histogram axis issue..

static/js/components/tiles/histogram_tile.tsx
chejennifer: yes that's what I had meant!

chejennifer (approved):
thanks for the changes!

PR #2373, opened 2023-03-03
[NL Interface] Place detection integration tests + "US states" fix

jehangiramjad:
In this PR we have two things:

1. Adding some place detection integration tests which test the Maps API + dc recon and NER at the same time.
2. Fix for NER incorrectly thinking "US states" is a place instead of just the "US". The solution is fairly inelegant but there are few elegant fixes for this IMHO.


server/routes/nl.py
```diff
@@ -281,6 +281,19 @@ def _result_with_debug_info(data_dict: Dict, status: str,
...
+  ):
+    index = cleaned_query.lower().find(text)
+    n = len(text)
+    while index >= 0:
+      cleaned_query = cleaned_query[0:index] + replacement + cleaned_query[
+          index + n:]
+      index = cleaned_query.lower().find(text)
+
+  logging.info(f"original query for detection: {orig_query}")
+  logging.info(f"query after text replacement for detection: {cleaned_query}")
```
pradh, line 295: Why not do this in detect_place()?

server/lib/nl/constants.py
```diff
@@ -154,6 +154,11 @@
     'across',
 }
 
+# Replace the text in the query with this alternate text.
+QUERY_TEXT_REPLACEMENT_FOR_DETECTION: Dict[str, str] = {
+    'us states': 'states in the United States'
+}
```
pradh, line 160: us states, states us, usa states, states usa ?

And, should we call this after stop-word replacement?

server/routes/nl.py
jehangiramjad: Good suggestion. Done!

server/lib/nl/constants.py
jehangiramjad: Ah yes, true. This revealed a few more things, e.g. sometimes NER (and I noticed the Cloud API also) detect "states" on their own as a place. So also removed those.

server/integration_tests/nlnext_test.py
```diff
@@ -124,8 +124,8 @@ def run_sequence(self,
             self.assertEqual(dbg["main_place_name"],
                              expected["main_place_name"])
 
-  def test_textbox_sample(self):
-    # This is the sample advertised in our textbox
+  # def test_textbox_sample(self):
```
pradh, line 127: Why are we commenting this out?

server/lib/nl/constants.py
```diff
@@ -186,8 +181,19 @@
     'sf east bay',
     # US
     'united states',  # need this because the word "states" gets replaced.
+    'usa',
 ])
 
+# Replace the detected place text with this alternate (shorter) place text.
+# The replacement text (value) should be contained within the detected text (key).
+# This ensures that the place string removal from queries will still remove valid place strings.
```
pradh, line 189: I don't understand this constraint.  Maybe an example?

server/lib/nl/utils.py
```diff
@@ -650,6 +650,18 @@ def place_detection_with_heuristics(query_fn, query: str) -> List[str]:
...
         if "the " in p:
           p = p.replace("the ", "")
+
+        # If the detected place string needs to be replaced with shorter text,
+        # then do that here.
+        if p.lower() in constants.SHORTEN_PLACE_DETECTION_STRING:
+          p = constants.SHORTEN_PLACE_DETECTION_STRING[p.lower()]
+
+        # Also remove place text detected which is lexactly equal to some place types
+        # e.g. "states" etc. This is a shortcoming of place entity recognitiion libraries.
```
pradh, line 660: Please include an example misbehavior this addresses?

pradh (approved):
Thanks!  Please undo the commented test before submission.

server/integration_tests/nlnext_test.py
jehangiramjad: Uncommented

server/lib/nl/utils.py
jehangiramjad: Provided an example

server/lib/nl/constants.py
jehangiramjad: Added more comment text

PR #2383, opened 2023-03-07
setup redis for stanford

chejennifer:
- add the files needed for redis cache for stanford instance
- update create_redis.sh script and make relevant instruction updates in redis.md
- add earth geojson endpoint to warmup requests


docs/redis.md
```diff
@@ -6,11 +6,11 @@ If you would like to use [Cloud Memorystore](https://cloud.google.com/memorystor
 
 ```bash
 cd gke
-./create_redis.sh <REGION>
+./create_redis.sh <INSTANCE> <REGION>
```
shifucun, line 9: maybe use <ENV>  instead of <INSTANCE> as we use it in other places?

Update below that <ENV> is the name of an instance, which can be found as deploy/gke/<ENV>.yaml

web_app.py
```diff
@@ -38,6 +38,7 @@
     "/api/choropleth/geojson?placeDcid=country/USA&placeType=County",
     "/api/place/parent/country/USA",
     "/api/place/places-in-names?dcid=country/USA&placeType=County",
+    "/api/choropleth/geojson?placeDcid=Earth&placeType=Country",
```
shifucun, line 41: nit, move this above (so sort all endpoints here)

docs/redis.md
chejennifer: done

web_app.py
chejennifer: done

shifucun (approved):

PR #2388, opened 2023-03-08
NL: Fix a few bugs related to guessing place types and per-capita

pradh:
1. Make default child types for "Earth" work right
2. Make contained-in ACROSS work right (regressed with map_type check)
3. Drop per-capita for a couple of more cases
 
image

image

image

### DIFF
--- a/server/lib/nl/utils.py
+++ b/server/lib/nl/utils.py
@@ -834,6 +834,8 @@ _SV_PARTIAL_DCID_NO_PC = [
     "AsFractionOf",
     "AsAFractionOfCount",
     "UnemploymentRate_",
+    "Mean_Income_",
+    "GenderIncomeInequality_",
 ]
 
 _SV_FULL_DCID_NO_PC = ["Count_Person"]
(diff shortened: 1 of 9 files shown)


server/lib/nl/utils.py
```diff
@@ -847,3 +849,9 @@ def is_percapita_relevant(sv_dcid: str) -> bool:
     if skip_sv == sv_dcid:
       return False
   return True
+
+
+def get_default_child_place_type(place: detection.Place) -> str:
+  if place.dcid == constants.EARTH_DCID:
+    return 'Country'
+  return constants.CHILD_PLACES_TYPES.get(place.place_type, 'County')
```
jehangiramjad, line 857: nit: it would be useful to document the behavior here and why might that be the case? Why is "County" the default, for example? What if the place is a non-US country?

jehangiramjad (approved):
Thanks

server/lib/nl/utils.py
pradh: Done, thanks.

PR #2390, opened 2023-03-08
[Custom DC] Argo CD install script and a simple example

Fructokinase:
This PR includes changes for installing Argo CD. For start, it will live in its own script "install.sh". Once we do more internal testing, we can include it in install-custom-dc.sh

https://custom-dc-test.alex-datacommons.dev/ was deployed based on manifest/custom-dc-test.yaml through the following command. The command below abstracts away the user having to call helm manually. It creates helm and its params inside the argo system, which watches for changes in  https://github.com/Fructokinase/website.

```
argocd app create -f custom-dc-test.yaml  --port-forward   --port-forward-namespace website
```

some other changes in this PR: _TAG_NAME -> TAG_NAME, _PROJECT_ID->PROJECT_ID. This way triggers by tag creation uses the tag value from the webhook.


deploy/argo/manifest/custom-dc-test.yaml
```diff
@@ -0,0 +1,77 @@
...
+spec:
+  project: default
+
+  syncPolicy:
+    automated:
+      # Delete resources that are no longer in the helm chart
+      prune: true
+      # When cluster is different from git, apply Git change to the cluster.
+      selfHeal: true
+
```
shifucun, line 28: remove extra newline here and below?

deploy/argo/manifest/custom-dc-test.yaml
```diff
@@ -0,0 +1,77 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+apiVersion: argoproj.io/v1alpha1
+kind: Application
+metadata:
+  name: custom-dc-test
```
shifucun, line 17: to confirm, every environment would have one such file? Does this replace overlays/<ENV>/*?

deploy/argo/manifest/custom-dc-test.yaml
```diff
@@ -0,0 +1,77 @@
...
+
+    helm:
+      releaseName: dc-website
+
+      values : |
+        namespace:
+          name: website
+
+        website:
+          githash: "1421f9a"
```
shifucun, line 43: since githash is based on github tag and is always changing, do you need to check it in here? (It may give the confusion that this version is used)

deploy/argo/manifest/custom-dc-test.yaml
```diff
@@ -0,0 +1,77 @@
...
+
+        mixer:
+          githash: "1aa6137"
+          image:
+            project: "datcom-ci"
+            tag: "1aa6137"
+          gcpProjectID: "custom-dc-test"
+          serviceName: "website-esp.endpoints.custom-dc-test.cloud.goog"
+
+        ingress:
```
shifucun, line 59: Where are flask env, number of pods etc configured?

PR #2391, opened 2023-03-08
Implement place fallback and optimize existence checks

pradh:
1. We may fallback to parent places... eg, jobs in sunnyvale -> SC county
2. We may fallback to parent contained-in types... eg, cities with highest agricultural production in california -> counties of CA
3. Collect SVs and batch existence check... this is done using a tracker class

Fallback is implemented in common code so it works for all query-types (except correlation and comparison, which is v custom rn).

TODO: Providing user feedback about the fallback

image

image

### DIFF
--- a/server/lib/nl/constants.py
+++ b/server/lib/nl/constants.py
@@ -446,18 +446,13 @@ EVENT_TYPE_TO_DC_TYPES = {
     EventType.WETBULB: ["WetBulbTemperatureEvent"],
 }
 
-# Given the original place type, this is the order of related place types.
-FALLBACK_CONTAINED_PLACE_TYPES = {
-    "City": ["County", "State", "Country"],
-    "County": ["State", "Country"],
-    "State": ["Country"],
+CHILD_PLACE_TYPES = {
+    ContainedInPlaceType.COUNTRY: ContainedInPlaceType.STATE,
+    ContainedInPlaceType.STATE: ContainedInPlaceType.COUNTY,
+    ContainedInPlaceType.COUNTY: ContainedInPlaceType.CITY,
 }
 
-CHILD_PLACES_TYPES = {
-    "Country": "State",
-    "State": "County",
-    "County": "City",
-}
+PARENT_PLACE_TYPES = {v: k for k, v in CHILD_PLACE_TYPES.items()}
 
 DEFAULT_PARENT_PLACES = {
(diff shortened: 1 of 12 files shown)


server/lib/nl/constants.py
```diff
@@ -445,6 +445,13 @@
     EventType.WETBULB: ["WetBulbTemperatureEvent"],
 }
 
+# Given the original place type, this is the order of related place types.
+FALLBACK_CONTAINED_PLACE_TYPES = {
+    "City": ["County", "State", "Country"],
+    "County": ["State", "Country"],
+    "State": ["Country"],
+}
+
```
jehangiramjad, line 454: Would it help to use the enums defined for various places (to prevent typo errors): https://github.com/datacommonsorg/website/blob/fa1603e74313bd300a3dd7d547c3a299dd8e63f2/server/lib/nl/detection.py#L95-L113 ?

server/lib/nl/fulfillment/base.py
```diff
@@ -177,8 +177,155 @@ def populate_charts_for_places(state: PopulateState,
   return False
 
 
+def _add_charts_with_place_fallback(state: PopulateState, places: List[Place],
```
jehangiramjad, line 180: nit: some comments or overall docstring might be helpful for others who might have to fiddle with this code in the future :)

server/lib/nl/fulfillment/base.py
```diff
@@ -177,8 +177,155 @@ def populate_charts_for_places(state: PopulateState,
...
+  if _add_charts(state, places, svs):
+    return True
+  if len(places) > 1:
+    return False
+
+  pt = state.place_type if state.place_type else places[0].place_type
+  if isinstance(pt, ContainedInPlaceType):
+    pt = pt.value
+
+  place = places[0]
```
jehangiramjad, line 191: is it guaranteed for places to not be empty? If so, please note why?

jehangiramjad (approved):
Thanks! If you could document the functions (esp the new ones) with more comments, it might be very helpful for others who might be looking in to this in the future :) But approving to unblock in case this is urgent.

beets (commented):
had a couple small comments..

pradh:
> Thanks! If you could document the functions (esp the new ones) with more comments, it might be very helpful for others who might be looking in to this in the future :) But approving to unblock in case this is urgent.

Documented now.

beets (approved):
thanks for the updates!

PR #2395, opened 2023-03-08
[disasters] optimizations part 1

chejennifer:
- remove unused european places fetch
- don't make node geojson fetch if geojson is already in the event point's prop vals
- only fetch parent places if not already passed as a prop to subject page

bug fixes along the way:
- fix info card properties displayed so that there aren't any duplicates
- use post to get propvals in top event tile (some get requests were getting too long because of too many dcids)


beets (approved):
nice!

juliawu (approved):
Thanks for the changes!

PR #2396, opened 2023-03-08
Kustomize changes for making Argo auto deploy to stanford staging

Fructokinase:
I configured Argo to watch for stanford staging changes under deploy/overlay/stanford_staging. This PR contains some changes to support that.

Some noticeable changes 
- We're brining all BT versions into stanford_staging subdirectory. I think it make it clearer which cache versions stanford staging instance is using
- Git versions (hence image tags) are no longer tied to the PR. Meaning, just by looking at the kustomization files, we know exactly which versions of mixer and website is deployed. Also, each instance can have a different version of images. 

For a view of what is deployed, please see:
https://deploy.datacommons.dev/applications/website/stanford-staging?view=tree&resource=

Note that currently my repo is tracked, but will be switched to main repo once this PR is committed.

### DIFF
--- a/deploy/overlays/stanford_staging/base_bigtable_info.yaml
+++ b/deploy/overlays/stanford_staging/base_bigtable_info.yaml
@@ -1,0 +1,8 @@
+project: datcom-store
+instance: prophet-cache
+tables:
+  - frequent_2023_03_07_19_58_29
+  - ipcc_2023_03_01_03_46_42
+  - disaster_2023_03_05_23_44_18
+  - infrequent_2023_03_06_09_05_14
+  - place_2023_03_06_15_27_01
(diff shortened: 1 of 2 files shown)


Fructokinase:
@shifucun After this PR, I will bring all the other instances under Argo too.

deploy/overlays/stanford_staging/base_bigtable_info.yaml
```diff
@@ -0,0 +1,8 @@
+project: datcom-store
```
shifucun, line 1: Not sure if there is an immediate benefit of multiple base_bigtable_info.yaml? Mean, we will almost always need to update multiple (10+) such files during a release.  For now, all dc managed custom instance should always track DC base tables

PR #2402, opened 2023-03-09
[disasters] fix storm paths

chejennifer:
- when storm path geojsons got added to the event data propvals, it was slightly different from the polygon geojsons so the original client side geojson processing did not work with it


static/js/utils/disaster_event_map_utils.tsx
```diff
@@ -173,12 +174,17 @@ function getGeoJsonFeature(
 ) {
   let geoJson = null;
   if (propVals[geoJsonProp] && !_.isEmpty(propVals[geoJsonProp].vals)) {
-    const geoJsonGeo = JSON.parse(propVals[geoJsonProp].vals[0]);
+    // remove any extra backslashes
```
beets, line 177: could you add detail here about the source of the extra backslashes?

beets (approved):

static/js/utils/disaster_event_map_utils.tsx
chejennifer: done

PR #2407, opened 2023-03-09
[disaster] cache a special geojson for countries on earth

chejennifer:
use an extra simplified geojson for countries on Earth

Screenshot 2023-03-09 at 2 47 10 PM

profiling localhost w/ network throttled as fast 3G:

dp3 (what we currently use):
size: 1.1MB
time: 13.3s

dp13 (this new one):
size: 462kB
time: 9.5s


server/lib/util.py
```diff
@@ -110,6 +114,20 @@ def get_nl_disaster_config():
   return get_subject_page_config(filepath)
 
 
+# Returns dict of place dcid to place type to geojson object
```
beets, line 117: could you add to the comment if it's right-hand rule (or not), and if it's a complete geojson (we usually have feature collections only)?

beets (approved):
v cool!

was there a noticeable rendering improvement?

server/lib/util.py
chejennifer: done

chejennifer:
I think it made the page feel slightly less heavy

PR #2408, opened 2023-03-10
[event] small header updates, update links to event pages

beets:
- update links from dashboard to event pages
- update header
   -  small styling updates
   - add place breadcrumbs (back to disaster dashboard)
   - fetch full provenance information

image


server/routes/event.py
```diff
@@ -79,6 +79,36 @@ def get_property_value(dcid: str, prop: str) -> str:
...
+
+  Args:
+    properties: All properties of the event.
+
+  Returns:
+    A list of { url, sourceName } objects to display provenances.
+  """
+  ret = []
+
+  provenance = properties['provenance']
```
chejennifer, line 94: should use properties.get('provenance') instead of accessing directly using the key incase the key doesn't exist

static/js/apps/event/app.tsx
```diff
@@ -91,20 +96,41 @@ export function App(props: AppPropsType): JSX.Element {
...
-          <h3>
-            dcid: <a href={`/browser/${props.dcid}`}>{props.dcid}</a>
-          </h3>
           <h3>
-            source:{" "}
-            <a href={`/browser/${provenance[0].dcid}`}>{provenance[0].name}</a>
+            <span>{_.startCase(typeOf[0].name)}</span>
+            {placeBreadcrumbsJsx && <> in {placeBreadcrumbsJsx}</>}
           </h3>
+          <h4 className="clearfix">
```
chejennifer, line 128: nit: curious why this classname

chejennifer (approved):
nice!

server/routes/event.py
beets: good point - done

static/js/apps/event/app.tsx
beets: there's a float on the browser link in the element, so the clearfix "resets" the floats 

https://getbootstrap.com/docs/4.0/utilities/clearfix/

beets:
thanks for the review!

PR #2409, opened 2023-03-10
NL: Improve debug info, counters

pradh:
1. Add page config proto to Debug Info (instead of utterance) and stop logging it
2. Track timings of RPC-intensive sections
3. Distinguish between info / error / timing counters
4. Refactor counters into its own module

TESTED: all integration tests pass; validated locally


server/lib/nl/counters.py
```diff
@@ -0,0 +1,70 @@
...
+
+  def __init__(self):
+    self._warn: Dict = {}
+    self._info: Dict = {}
+    self._timing: Dict = {}
+
+  #
+  # Adds/updates an info counter.
+  #
+  # For a given counter, caller should always pass the same type
```
jehangiramjad, line 32: if there are two different arg types for `value` then why not support two different functions? Or at least have two different versions of _update(), one for the numeric case and one for the list?

server/lib/nl/counters.py
```diff
@@ -0,0 +1,70 @@
...
+  # accounts it to the given counter.
+  #
+  def timeit(self, counter: str, start: float):
+    duration = round(time.time() - start, 2)
+    self._timing[counter] = self._timing.get(counter, 0) + duration
+
+  #
+  # Returns a dict with all counters.
+  #
+  def get(self) -> Dict:
```
jehangiramjad, line 58: nit: may be have an enum to define the keys? Also, the first key here should be 'WARN' and not 'ERROR', right?

server/lib/nl/counters.py
```diff
@@ -0,0 +1,70 @@
...
+    return {'ERROR': self._warn, 'INFO': self._info, 'TIMING': self._timing}
+
+  def _update(self, counters: Dict, counter: str, value: any):
+    should_add = counter not in counters
+    if isinstance(value, int) or isinstance(value, float):
+      if should_add:
+        counters[counter] = 0
+      counters[counter] += value
+    else:
```
jehangiramjad, line 67: nit: what if value is a set and not a list? There is no check that prevents this right now so perhaps log something here which notes if an unexpected type was found?

server/lib/nl/counters.py
```diff
@@ -0,0 +1,70 @@
...
+    self._timing: Dict = {}
+
+  #
+  # Adds/updates an info counter.
+  #
+  # For a given counter, caller should always pass the same type
+  # for value.  If value is numeric, then its a single updated
+  # counter, otherwise, counter is a list of values.
+  #
+  def info(self, counter: str, value: any):
```
jehangiramjad, line 36: reading the code below it seems that this is expected to either be an int or float or a List of values, right? In that case, a more useful type hint would be:

`def info(self, counter: str, value: Union[int, float, List]):` or even better if the List is expected to have either floats or ints too:

`def info(self, counter: str, value: Union[int, float, List[Union[int, float]]]):`

jehangiramjad, line 36: This is updating the counter so why not have the function name reflect that? info() seems to imply you're getting some information instead of updating an info counter, IMHO. I might even suggest `update_info()` or `add_info()`? If you agree, then the same would apply to `warn()` and `timeit()` below.

jehangiramjad (approved):
Thanks. Left a few comments which can all be ignored and put to a future TODO, if you'd like. Approving to not block any of this.

server/lib/nl/counters.py
pradh: This is called from a bunch of different places, and I was striving for something that is very low barrier to call (i.e., just know whether its info or error and pass "anything" to it really) :)

server/lib/nl/counters.py
pradh: I have some opinions about use of Union type in py :), and I admit it is a taste thing.

Imo the hard-to-read/long Union structure somehow doesn't seem to justify its presence, especially when python won't enforce it either.  I view  typing as a useful hint to know what type it can take (for IDE esp), and that's not too helpful when its Union which can be one of many.

And finally, the type can be more than beyond just int/float... it can be str, enums (ContainedInPlaceType), etc.

pradh: As you see this gets called from all over the fulfillment code, so I was trying to use as few chars as possible :)

In most places, we already have `state.uttr.counters.info(...)` which is already a mounthful?

Tried to model like `logging.info()`, rather than `logging.add_info()`.

server/lib/nl/counters.py
pradh: Good catch.  Sorry, I was lazy.  This really felt like warning more than error, but then (last minute) I wanted it to be ordered before INFO and TIMING in the Debug Info for ease of spotting issues (map keys are ordered alphabetically).

So maybe I should make it `err` after all?

server/lib/nl/counters.py
pradh: Actually "any" type is okay as long as it is JSON serializable.  But you make a good point, if we find an unexpected (JSON unserializable) type, we should error out.  I will make that change to `get()`.

pradh (commented):
Thanks for your comments Jehangir.  I'll make a couple of changes in a follow up PR.

server/lib/nl/counters.py
pradh: Also regarding enums, since this is the only place really the value is used, perhaps sticking to constants is okay (imo the LOC to defining an enum for a single-use case isn't worth it)?

And what also makes it funky is the need to make the enum JSON serializable (i.e., make it a string and inherit from `str`)

pradh:
Sent out the fixes I mentioned in https://github.com/datacommonsorg/website/pull/2413/files.

PR #2413, opened 2023-03-11
NL: minor counter fixes and fix debug info regression

pradh:
Fix regression introduced in https://github.com/datacommonsorg/website/pull/2409, where if page-config is empty, we don't show up debug info.

Also, Jehangir's review comments from https://github.com/datacommonsorg/website/pull/2409.


beets (approved):

pradh:
Thanks for the review!

PR #2416, opened 2023-03-13
[Custom DC] initial commit for e2e setup testing

Fructokinase:
First iteration of integration testing for Terraform. This PR adds a test for the setup of DC website e2e. 

Specifically, it does the following.
1. Apply Terraform from scratch for resources under examples/website_v1
2. Deploys Helm chart
3. Wait up to 30min for the main page to return 200
4. Tear down all the resources (Helm, GKE cluster, IAM)

Note that the test does not include setup module and BT automation resources, but is a good start for testing cluster creation, helm deployment, ingress, etc (where most of the issues happen).

The test is currently manually triggered by the following, and should run in about 30min.
```
go test -v -timeout 60m .
```

Next steps
- Create different test cases (domain registration, regional cluster)
- Also include setup and bt automation modules in the integration testing cycle
- Add a periodic trigger for this test


deploy/helm_charts/dc_website/templates/ingress.yaml
```diff
@@ -13,17 +13,6 @@
 # limitations under the License.
 
 {{- if .Values.ingress.enabled }}
-apiVersion: networking.gke.io/v1
```
shifucun, line 16: why is this removed?

deploy/terraform-datacommons-website/test/go.mod
```diff
@@ -0,0 +1,89 @@
+module github.com/datacommonsorg/website/test
+
+go 1.19
+
+require (
```
shifucun, line 5: can you try `go mod tidy` to clean the modules here

deploy/terraform-datacommons-website/test/website_v1_test.go
```diff
@@ -0,0 +1,190 @@
...
+			"mixer.githash":              RELEASE_MIXER_GITHASH,
+			"website.gcpProjectID":       PROJECT_ID,
+			"website.domain":             WEBSITE_DOMAIN,
+			"website.secretGCPProjectID": PROJECT_ID,
+			"mixer.gcpProjectID":         PROJECT_ID,
+			"mixer.serviceName":          fmt.Sprintf("website-esp.endpoints.%s.cloud.goog", PROJECT_ID),
+			"ingress.enabled":            "true",
+		},
+		SetFiles: map[string]string{
+			"mixer.schemaConfigs.\"base\\.mcf\"":   mixerPathTo("deploy/mapping/base.mcf"),
```
shifucun, line 127: you can use `` to avoid escape "

shifucun:
Can you make a cloudbuild for this test only and add it as a 1h period build in https://github.com/datacommonsorg/tools/tree/master/periodic_builds

deploy/helm_charts/dc_website/templates/ingress.yaml
Fructokinase: Currently certificate is created in Terraform, this one was never used (nowherre in our code references the name).

deploy/terraform-datacommons-website/test/go.mod
Fructokinase: This is the result from go mod tidy. Do you mean that you think there are more indirect dependencies then there should be?

deploy/terraform-datacommons-website/test/website_v1_test.go
Fructokinase: I'm afraid that the consumer of this is the k8s server in go, which only takes double quotes.

Fructokinase:
Will add it to the periodic tests

deploy/terraform-datacommons-website/test/go.mod
shifucun: Yea, but maybe terraform does use a lot of libraries.

deploy/terraform-datacommons-website/test/website_v1_test.go
shifucun: shouldn't they be complied and used?

shifucun:
Maybe can add a cloudbuild for the actual test in this PR too?

shifucun (approved):

PR #2419, opened 2023-03-14
[event] Add more tiles

beets:
- Updated page config trimming logic to iterate through the config once while looking up existence in a dict, and filter more types.
- Add more charts
- Add support for block title placeholder strings

image


static/js/components/subject_page/block_container.tsx
```diff
@@ -40,13 +44,18 @@ export function BlockContainer(props: BlockContainerPropType): JSX.Element {
       })
       .join("\n");
   }
+  const rs: ReplacementStrings = {
+    place: props.place.name,
```
chejennifer, line 48: since place is optional (from the prop type), we should check that `props.place` is defined before trying to access `props.place.name`

server/tests/lib/subject_page_config_test.py
```diff
@@ -42,58 +43,113 @@ def test(self):
...
+                  }
+              },
+          }
+      }
+
+    mock_observation_existence.side_effect = obs_side_effect
+    config = lib_util.get_subject_page_config(
+        "server/tests/test_data/existence.textproto")
+    result = lib_subject_page_config.remove_empty_charts(config, 'place_id')
+    print("result:")
```
chejennifer, line 151: did you mean to keep these print statements?

chejennifer (approved):
nice!

static/js/components/subject_page/block_container.tsx
beets: good call - done.

server/tests/lib/subject_page_config_test.py
beets: no i didn't - thanks, removed.

beets:
thanks for the review!

PR #2422, opened 2023-03-15
Store mixer-api-key in the website GCP project instead of in the mixer project

shifucun:
This aligns with the regular practice that api key sits in a separate (client) project. With this, custom DC users can store the API key in the custom project.

### DIFF
--- a/server/__init__.py
+++ b/server/__init__.py
@@ -275,9 +275,9 @@ def create_app():
   if app.config['LOCAL']:
     os.environ['OAUTHLIB_INSECURE_TRANSPORT'] = '1'
 
-  if app.config['API_PROJECT']:
+  if cfg.SECRET_PROJECT:
     secret_client = secretmanager.SecretManagerServiceClient()
-    secret_name = secret_client.secret_version_path(cfg.API_PROJECT,
+    secret_name = secret_client.secret_version_path(cfg.SECRET_PROJECT,
                                                     'mixer-api-key', 'latest')
     secret_response = secret_client.access_secret_version(name=secret_name)
     app.config['DC_API_KEY'] = secret_response.payload.data.decode('UTF-8')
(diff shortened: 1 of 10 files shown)


server/__init__.py
```diff
@@ -275,9 +275,9 @@ def create_app():
   if app.config['LOCAL']:
     os.environ['OAUTHLIB_INSECURE_TRANSPORT'] = '1'
 
-  if app.config['API_PROJECT']:
+  if cfg.SECRET_PROJECT:
```
Fructokinase, line 278: Will all secrets (currently there's maps api key and oauth client id) always be stored in the same project? oauth client and maps secrets will always be local to the custom dc, so meaning mixer api keys will always be in custom dc project as well? Just making sure

server/app_env/custom.py
```diff
@@ -24,4 +24,11 @@ class Config(_base.Config):
 
 
 class LocalConfig(Config, local.Config):
-  LITE = True
\ No newline at end of file
+  LITE = True
+  API_ROOT = 'https://autopush.api.datacommons.org'
+  # NEED TO DO THE FOLLOWING CHANGES:
+  # 1. Get an API key from Data Commons team
```
Fructokinase, line 30: nit: include contact email here?

server/app_env/integration_test.py
```diff
@@ -18,6 +18,5 @@
 class Config(_base.Config):
   INTEGRATION = True
   API_ROOT = 'https://autopush.api.datacommons.org'
-  API_PROJECT = 'datcom-mixer-autopush'
-  SCHEME = 'http'
-  SECRET_PROJECT = 'datcom-website-dev'
\ No newline at end of file
+  SECRET_PROJECT = 'datcom-website-dev'
```
Fructokinase, line 21: Thanks for creating the secrets in the new projects!

Fructokinase (approved):

server/__init__.py
shifucun: Yes, for custom DC, both api keys and the secret manager is within the custom project.

server/app_env/custom.py
shifucun: Done

PR #2426, opened 2023-03-15
[disaster] some more optimizations

chejennifer:
- remove child place fetch
- remove unnecessary blocking calculation when adding map points
- remove unnecessary node geojson call

### DIFF
--- a/server/routes/disasters.py
+++ b/server/routes/disasters.py
@@ -25,6 +25,7 @@ from google.protobuf.json_format import MessageToJson
 
 import server.lib.subject_page_config as lib_subject_page_config
 import server.lib.util
+import server.routes.api.place as place_api
 
 EARTH_FIRE_SEVERITY_MIN = 500
 FIRE_EVENT_TYPE_SPEC = "fire"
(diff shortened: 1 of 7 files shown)


server/routes/disasters.py
```diff
@@ -66,6 +67,12 @@ def disaster_dashboard(place_dcid=None):
     dashboard_config.metadata.contained_place_types.update(
         place_metadata.contained_place_types_override)
 
+  child_places = place_api.child_fetch(place_dcid)
```
beets, line 70: we could move this to a helper so it's returned in lib_subject_page_config.place_metadata (or I can do it, if this goes in first).

static/js/chart/draw_d3_map.ts
```diff
@@ -516,20 +516,23 @@ export function addMapPoints(
...
-      const pathClientRect = (
-        paths[idx] as SVGPathElement
-      ).getBoundingClientRect();
-      minRegionDiagonal = Math.sqrt(
-        Math.pow(pathClientRect.height, 2) + Math.pow(pathClientRect.width, 2)
-      );
-    });
-  const minDotSize = minDotRadius || Math.max(minRegionDiagonal * 0.02, 1.1);
+  let minDotSize = minDotRadius;
+  if (!minDotSize) {
```
beets, line 520: could you add a comment for why this is important here so we don't undo this in the future

chejennifer: done

server/routes/disasters.py
chejennifer: since I think this will be going in first, I added a TODO for you to move this to a helper

PR #2429, opened 2023-03-16
Clean up SECRET_PROJECT setup

shifucun:
In GKE, do not setup SECRET_PROJECT in yaml since it is always the current GCP project and can be retrieved from GCP environment.

In addition, only need to fetch mixer-api-key for local dev.


Fructokinase (approved):
Thanks for deleting from both the helm charts and kustomize files!

PR #2432, opened 2023-03-16
[Sustainability] fix bugs introduced in #2421

beets:
- Fix page titles of disaster dashboard pages
- Fix links for child places

### DIFF
--- a/static/js/shared/child_places.tsx
+++ b/static/js/shared/child_places.tsx
@@ -23,10 +23,13 @@ import React from "react";
 
 import { displayNameForPlaceType } from "../place/util";
 import { ChildPlacesByType, NamedTypedPlace } from "../shared/types";
+import { formatString, ReplacementStrings } from "../utils/tile_utils";
 
 interface ChildPlacesPropType {
   parentPlace: NamedTypedPlace;
   childPlaces: ChildPlacesByType;
+  // Format string must include ${dcid} for child place URLs.
+  urlFormatString: string;
 }
 
 export function ChildPlaces(props: ChildPlacesPropType): JSX.Element {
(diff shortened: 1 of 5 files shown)


static/js/shared/child_places.tsx
```diff
@@ -42,16 +45,23 @@ export function ChildPlaces(props: ChildPlacesPropType): JSX.Element {
...
-              className="child-place-link"
-              key={`child-place-${i}`}
-            >
-              {place.name || place.dcid}
-              {i < props.childPlaces[placeType].length - 1 ? "," : ""}
-            </a>
-          ))}
+          {props.childPlaces[placeType].map((place, i) => {
+            const rs: ReplacementStrings = {
+              place: "",
```
chejennifer, line 50: maybe we should update place and date to also be optional since there seems to be a lot of cases where it's just empty string passed for those fields? Can just add a TODO

static/js/utils/tile_utils.tsx
```diff
@@ -27,6 +27,7 @@ export interface ReplacementStrings {
   statVar?: string;
   xDate?: string;
   yDate?: string;
+  place_dcid?: string;
```
chejennifer, line 30: nit: placeDcid

Maybe also update place to placeName or add a comment to say it's the name

chejennifer (approved):
thanks for the fixes

static/js/shared/child_places.tsx
beets: i'll add a TODO and update in a followup

static/js/utils/tile_utils.tsx
beets: done. will add a todo to rename place and update it in a followup

beets:
thanks for the review!

PR #2433, opened 2023-03-16
update redis host for stanford

chejennifer:
- created a new redis cache for stanford, update the host


shifucun (approved):
Does clear cache work for this one?

chejennifer:
> Does clear cache work for this one?

does not seem to work still ..

PR #2435, opened 2023-03-16
[subject page] Replacement string updates

beets:
ReplacementString updates, follow up from #2432
- rename place -> placeName
- make all keys optional


chejennifer (approved):
thanks for the update!

PR #2436, opened 2023-03-17
[nl] remove query history

beets:
Removing query history from the NL page, especially while we push the boundaries of the app in testing

### DIFF
--- a/static/js/apps/nl_interface/app.tsx
+++ b/static/js/apps/nl_interface/app.tsx
@@ -21,7 +21,6 @@
 import React, { useEffect, useRef, useState } from "react";
 
 import { getUrlToken } from "../../tools/stat_var/util";
-import { QueryHistory } from "./query_history";
 import { QueryResult } from "./query_result";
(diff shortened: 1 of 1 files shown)


static/js/apps/nl_interface/app.tsx
```diff
@@ -184,7 +177,6 @@ export function App(): JSX.Element {
             }
           }}
         />
-        {isStartState && <QueryHistory onItemClick={onHistoryItemClick} />}
```
chejennifer, line 187: wondering if this is something we're just temporarily getting rid of & eventually want back? 

because if we will want to add this back, we could just replace `isStartState` with `false` instead of deleting everything?

chejennifer (approved):
thanks!

static/js/apps/nl_interface/app.tsx
beets: yes, it seems temporary - sending you a follow up (sorry for premature auto-merge)

PR #2438, opened 2023-03-20
Fix `sed` on linux

RealOrangeOne:
`-i`'s suffix shouldn't have a space.


shifucun:
Thanks for the fix, can you sign the CLA which is a required pre-submit of this PR

shifucun (approved):

shifucun:
Can you update the branch so we can merge this in. Thanks for the fix!

PR #2440, opened 2023-03-20
[event] fix pages

beets:
Inadvertently used a feature of 3.9 that isn't available on 3.7 (reversing while iterating a dictionary): https://stackoverflow.com/questions/55911745/python-reverse-dictionary-items-order


chejennifer (approved):
thanks for the fix!

PR #2442, opened 2023-03-21
NL: Fulfillment of international sub-national geos

pradh:
Most of the logic is for mapping from user's reference to a colloquial admin-area type ("province", "parish", etc.) to the canonical admin-area type (using `ADMIN_DIVISION_EQUIVALENTS` map), and then map from the canonical type to the place type for which we have stats/geojsons for that country (using `*_PLACE_TYPE_REMAP` maps).

In future, we may add more colloquial types (like governorate), and also make the colloquial type to canonical type also be country dependent (e.g., "department" is AA1 in some countries while AA2 in others).

Screenshots of queries added to the integration test:

image

image

image

image

image

image


server/lib/nl/fulfillment/base.py
```diff
@@ -317,10 +321,12 @@ def perform_existence_check(self):
     else:
       logging.info('Existence check failed for %s - %s', ', '.join(self.places),
                    ', '.join(self.all_svs))
-      self.state.uttr.counters.err('failed_existence_check', {
-          'places': self.places,
-          'svs': list(self.all_svs),
-      })
+      self.state.uttr.counters.err(
+          'failed_existence_check', {
+              'places': self.places[:3],
```
beets, line 326: i don't understand the slicing here and below. use a constant (or lower the limit of child places used for the existence check?)

beets (approved):
cool!

pradh (commented):
Thanks!

server/services/datacommons.py
```diff
@@ -221,6 +221,14 @@ def property_values(nodes, prop, out=True):
   return result
 
 
+def get_place_info(dcids: List[str]) -> Dict:
+  """Retrieves Place Info given a list of DCIDs."""
+  url = get_service_url('/v1/bulk/info/place')
+  return post(f'{url}', {
+      'nodes': sorted(set(dcids)),
```
shifucun, line 228: sort is not necessary here?

shifucun (commented):
looking good!

PR #2448, opened 2023-03-21
[sustainability] add most charts for sustainability pages

beets:
- added most charts
- added support for placeholders in category / block descriptions and titles
- updated filtering logic to account for:
   - charts that plot child places
   - bar charts with comparison place overrides. these can include "self" which will be interpreted as the main place dcid.
   - maps - check that child places have geojson's
- added cache for the html responses for this and events

pending: tests for geojson existence for maps, and a main disaster chart

localhost_8080_sustainability_geoId_06


server/tests/lib/subject_page_config_test.py
```diff
@@ -110,44 +112,88 @@ def test_exist_keys_category(self):
...
+    mock_geojson_properties.side_effect = properties_side_effect
+
     config = lib_util.get_subject_page_config(
         "server/tests/test_data/existence.textproto")
-    result = lib_subject_page_config.remove_empty_charts(config, 'place_id')
+    result = lib_subject_page_config.remove_empty_charts(
+        config, 'place_id', 'child_type')
     expect = lib_util.get_subject_page_config(
         "server/tests/test_data/existence_expect.textproto")
+    print(expect)
```
chejennifer, line 198: remove this print statement?

server/routes/event.py
```diff
@@ -199,8 +201,13 @@ def event_node(dcid=DEFAULT_EVENT_DCID):
       subject_config.metadata.contained_place_types.update(
           DEFAULT_CONTAINED_PLACE_TYPES)
 
+    place_type = None
+    for pt in place_metadata.place_types:
+      if pt in DEFAULT_CONTAINED_PLACE_TYPES:
```
chejennifer, line 206: since the contained_place_types in the config can either be `DEFAULT_CONTAINED_PLACE_TYPES` OR `place_metadata.contained_place_types_override`, shouldn't we use whichever we set to the config? Also in line 210

Also for disasters.py and sustainability.py

chejennifer (commented):
nice, this config trimming is cool!

server/tests/lib/subject_page_config_test.py
beets: done

server/lib/subject_page_config.py
```diff
@@ -32,23 +35,29 @@
 }
 
 # Tile types to filter with existence checks.
-FILTER_TILE_TYPES = [
+PLACE_FILTER_TILE_TYPES = [
+    subject_page_pb2.Tile.TileType.TYPE_NONE,
     subject_page_pb2.Tile.TileType.HIGHLIGHT,
     subject_page_pb2.Tile.TileType.RANKING,
```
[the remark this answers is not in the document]
beets: good catch - done

server/routes/event.py
beets: yes - good point! done.

beets:
thanks for the review! ptal.

chejennifer (approved):
thanks for the updates!

PR #2455, opened 2023-03-22
[disasters] fix bug with leaf pages

beets:
We currently throw an error if there aren't any child place types to fetch (if we are on a leaf page according to the config), introduced in #2448 

Also fixed a couple of strange sounding chart titles

### DIFF
--- a/server/routes/sustainability.py
+++ b/server/routes/sustainability.py
@@ -71,8 +71,8 @@ def sustainability_explorer(place_dcid=None):
     if pt in config_place_types:
       place_type = pt
       break
-  contained_place_type = config_place_types[
-      place_type] if place_type != None else None
+  contained_place_type = config_place_types.get(
+      place_type, None) if place_type != None else None
   subject_config = lib_subject_page_config.remove_empty_charts(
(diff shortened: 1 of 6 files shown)


server/routes/sustainability.py
```diff
@@ -71,8 +71,10 @@ def sustainability_explorer(place_dcid=None):
     if pt in config_place_types:
       place_type = pt
       break
+  contained_place_type = config_place_types[
+      place_type] if place_type != None else None
   subject_config = lib_subject_page_config.remove_empty_charts(
-      subject_config, place_dcid, config_place_types[place_type])
+      subject_config, place_dcid, contained_place_type)
```
chejennifer, line 77: nit: could just do `contained_place_type.get(place_type, None)`?

chejennifer (approved):
thanks for fixing!

server/routes/sustainability.py
beets: hm, looks like i need both (since place_type could also be None, and .get throws an error then).

beets:
thanks for the review!

PR #2460, opened 2023-03-22
[i18n] Update i18n instructions

beets:


docs/localization.md
```diff
@@ -26,18 +26,20 @@ Prepare mesages for translation by extracting from both python and react with:
...
 ```
 
+Send a PR with the changes from the script then follow the instructions in google3.
+
+
 ## Compile translation files for use
 
 In order to use the extracted strings (regardless of translation status),
-compile the messages to be used by both libraries.
+compile the messages to be used by both libraries after following the steps in google3.
```
chejennifer, line 35: can we add a link to the google3 readme?

chejennifer (approved):
thanks!

docs/localization.md
beets: good idea - done

PR #2461, opened 2023-03-23
[sustainability] add more logging

beets:
- Added more logging to debug flaky sustainability/Earth page (logging lines are quite verbose so these are temporary)
- Fix a small bug

### DIFF
--- a/server/lib/subject_page_config.py
+++ b/server/lib/subject_page_config.py
@@ -51,6 +51,9 @@ CHILD_FILTER_TILE_TYPES = [
 # Placeholder allowed in config that should be interpreted as the main place dcid.
 SELF_PLACE_DCID_PLACEHOLDER = "self"
 
+# Only keep every third sample place.
+SAMPLE_PLACE_STEP = 3
+
 
 @dataclass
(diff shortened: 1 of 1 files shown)


server/lib/subject_page_config.py
```diff
@@ -165,6 +170,8 @@ def remove_empty_charts(page_config, place_dcid, contained_place_type=None):
                                             stat_vars_existence, place_dcid)
     child_exist_keys = _exist_keys_category(sample_child_places, category,
                                             stat_vars_existence, place_dcid)
+    logging.info('_remove_empty_charts place_exist_keys: %s', place_exist_keys)
```
pradh, line 173: nit: place_exist_keys looks to be a dict, will %s work?

```
logging.info(f'_remove_empty_charts place_exist_keys: {place_exist_keys}')
```

is another option

pradh (approved):

server/lib/subject_page_config.py
beets: yes it seems to work.

beets:
thanks for the quick review!

PR #2462, opened 2023-03-23
[i18n] Update translations

beets:
Importing strings from cl/518703420 (note the new strings are still in english)

image


chejennifer (approved):
thanks!

PR #2468, opened 2023-03-23
[disasters] small fixes

beets:
- add contained_place_types overrides for more countries
- filter child places which are unsupported
- remove logging statements from #2461 (the issue didn't reproduce anymore, leading me to think that we were originally sampling too many places and timing out occasionally)
- update place type for the page from list to str

### DIFF
--- a/server/lib/subject_page_config.py
+++ b/server/lib/subject_page_config.py
@@ -269,7 +269,7 @@ def place_metadata(place_dcid, get_child_places=True) -> PlaceMetadata:
       child_places[place_type] = child_places[place_type][:place_api.
                                                           CHILD_PLACE_LIMIT]
 
-    # Filter child places for unsupported place types
+    # Filter out unsupported place types
     supported_place_types = set(contained_place_types.keys())
     supported_place_types.update(contained_place_types.values())
     for ptype, places in child_places.items():
--- a/server/lib/subject_page_config.py
+++ b/server/lib/subject_page_config.py
@@ -26,12 +26,29 @@ import server.services.datacommons as dc
 
 DEFAULT_PLACE_DCID = "Earth"
 DEFAULT_PLACE_TYPE = "Planet"
-EUROPE_DCID = "europe"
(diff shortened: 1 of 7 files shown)


server/routes/event.py
```diff
@@ -192,29 +192,19 @@ def event_node(dcid=DEFAULT_EVENT_DCID):
...
+        place_metadata.contained_place_types)
+    contained_place_type = place_metadata.contained_place_types.get(
+        place_metadata.place_type, None)
     subject_config = lib_subject_page_config.remove_empty_charts(
         subject_config, place_dcid, contained_place_type)
 
     # TODO: If not enough charts from the current place, add from the next place up and so on.
     subject_page_args = {
-        "place_type": json.dumps(place_metadata.place_types),
+        "place_type": json.dumps(place_metadata.place_type),
```
chejennifer, line 207: nit: looks like place_type is just a string now right? so don't need to json.dumps

server/lib/subject_page_config.py
```diff
@@ -236,27 +247,45 @@ def place_metadata(place_dcid) -> PlaceMetadata:
...
+
+  filtered_child_places = {}
+  if get_child_places:
+    child_places = place_api.child_fetch(place_dcid)
+    for place_type in child_places:
+      child_places[place_type].sort(key=lambda x: x['pop'], reverse=True)
+      child_places[place_type] = child_places[place_type][:place_api.
+                                                          CHILD_PLACE_LIMIT]
+
+    # Filter child places for unsupported place types
```
chejennifer, line 272: nit: ...for supported place types?

chejennifer (approved):
nice, thanks for the fix!

server/routes/event.py
beets: thanks for catching that

server/lib/subject_page_config.py
beets: done

beets:
thanks for the review!

Issue #2386, opened 2023-03-08
[NL] Make unit names in charts shorter


pradh:
Things like InflationAdjustedUSD_CurrentYear/USDollar ("family earnings in california"), Hundredweight ("agricultural output in california"), etc

pradh:
`shortDisplayName` in KG SG.  And FE will make an API call to use it?

pradh:
Units in DC: https://docs.google.com/spreadsheets/d/1dVm8SK3a3xvXpNzENNgh6KIx8xvf9epSaKEMqdHGcDU/edit#gid=2015142530

lucy-kind:
Update: Out for internal review

lucy-kind:
Internal change submitted. Assigning to shanth for front end

pradh:
@chejennifer could you help make the change to use `shortDisplayName` in FE for unit names?  Then we can hopefully get rid of the static maps we have.

PR #2469, opened 2023-03-23
[units] use shortDisplayName for units

chejennifer:
- update flask endpoints to get shortDisplayName for units and add it to the facet information
- refactor frontend functions so that there is a single getUnit function (originally had 3 different functions) that will use the unitDisplayName if available

Screenshot 2023-03-23 at 11 09 50 AM

fixes https://github.com/datacommonsorg/website/issues/2386


beets (approved):
thanks for the refactor too!

PR #2470, opened 2023-03-23
cleanup formatNumber

chejennifer:
- cleanup the code so that there is only the i18n formatNumber function & loadLocaleData for additional apps to be able to use the i18n formatNumber


beets (approved):
yay - thank you for this!

PR #2471, opened 2023-03-23
[disasters] handle empty geojson

chejennifer:
- add geojson drawing for cities
- handle when no geojson found

Screenshot 2023-03-23 at 11 18 31 AM


beets (approved):
thanks for the fix!

PR #2476, opened 2023-03-24
[disaster] event filter updates

beets:
- added support for place type overrides for severity filters
- added overrides for fires and floods at various levels
- updated position of severity filter UI (easiest to review if you ignore whitespace)
- also updated top events to show 20 rows to fill up whitespace (and since we have so many events)

after:
image

image

compare to before:
image

(in this case, sometimes could be out of the viewport)
image


static/js/components/tiles/top_event_tile.tsx
```diff
@@ -40,7 +40,7 @@ import { rankingPointsToCsv } from "../../utils/chart_csv_utils";
 import { getPlaceNames } from "../../utils/place_utils";
 import { ChartFooter } from "./chart_footer";
 
-const RANKING_COUNT = 10;
+const RANKING_COUNT = 20;
```
chejennifer, line 43: did you check how it looks on NL interface?

server/config/subject_page.proto
```diff
@@ -44,6 +44,8 @@ message EventTypeSpec {
   string color = 4;
   // Default severity filter to use for this event type.
   SeverityFilter default_severity_filter = 5;
+  // Severity filter overrides, keyed by place type.
+  map<string, SeverityFilter> place_type_severity_filter = 10;
```
chejennifer, line 48: I wonder if instead of adding this to the proto, we could just have a map somewhere for this .. because this proto is what is getting sent to the client side and it doesn't look like we're actually sending this information?

beets: we used to have that override in code and i feel it's messy and easy to forget. this way all the definitions related to it is in one place and easy to update.

static/js/components/tiles/top_event_tile.tsx
beets: as discussed offline, made this an optional property of top event tile spec

beets:
@chejennifer ptal! :)

server/config/subject_page.proto
chejennifer: hm, ok, that makes sense. Maybe add a comment that this is just used to override default_severity_filter and that it's not actually sent to the client

chejennifer (approved):

beets:
thanks for the review!

Issue #2368, opened 2023-03-03
[NL] Found weird or long chart names


pradh:
[what are the biggest cities in mexico] - "Females Born in Central American Countries Other Than Mexico"

pradh:
Open to track the weird / long names...

pradh:
Typo in the name: 

![image](https://user-images.githubusercontent.com/4375037/225968860-dc84c4ae-ec9d-401b-a4c3-7db29d322586.png)

pradh:
Some of titles gender + race seen wrong.

![image](https://user-images.githubusercontent.com/4375037/226737773-c6a93e80-6f48-4353-bd15-021f21db034c.png)

https://github.com/datacommonsorg/website/blob/19f87b250a17b75a7c52fc31436ea4ecdec89613/server/config/nl_page/chart_titles_by_sv.json#L649-L678

pradh:
One more:

```
  "GiniIndex_EconomicActivity": "Gini Index of Economic Activity of a Population",
```

PR #2480, opened 2023-03-24
add chart titles

lucy-kind:
Add names for wet bulb temperature related SVs and fix titles for male population titles from https://github.com/datacommonsorg/website/issues/2368


jehangiramjad:
I notice the only changes here are in the integration tests. Is that expected?

jehangiramjad (approved):
Thanks!

PR #2483, opened 2023-03-24
[disasters] fix autocomplete searchbox

beets:
the search box should be initialized only after rendering (different from place explorer which includes the searchbox in the html template).


chejennifer (approved):
thanks for the fix!

Issue #2481, opened 2023-03-24
SV hierarchy: Consider showing tool-tip for internal nodes


pradh:
SDG indicators might have longish names:

![image](https://user-images.githubusercontent.com/4375037/227606247-3f5fadeb-e702-4639-90b2-2bf675750dd4.png)

PR #2487, opened 2023-03-24
[sv widget] add tooltip for internal svg nodes

chejennifer:
screen-recording-_83_

fixes: https://github.com/datacommonsorg/website/issues/2481


static/js/stat_var_hierarchy/node_header.tsx
```diff
@@ -80,5 +88,15 @@ export class StatVarHierarchyNodeHeader extends React.Component<StatVarHierarchy
       </div>
     );
   }
+
+  private mouseMoveAction = (e) => {
```
shifucun, line 92: Is there a way to copy the text from the tooltip? We have been wanting to copy the dcid of the svg from hierarchy.

static/js/stat_var_hierarchy/node_header.tsx
```diff
@@ -80,5 +88,15 @@ export class StatVarHierarchyNodeHeader extends React.Component<StatVarHierarchy
...
+    const left = e.pageX;
+    const containerY = (
+      d3.select(`#${SV_HIERARCHY_SECTION_ID}`).node() as HTMLElement
+    ).getBoundingClientRect().y;
+    const top = e.pageY - containerY + TOOLTIP_TOP_OFFSET;
+    const tooltipHtml = `<b>${this.props.title}</b>`;
```
shifucun, line 98: Can we should the dcid as well? This will be very useful.

static/js/stat_var_hierarchy/node_header.tsx
chejennifer: that will require some thought about the design because the tooltip stays to the right of the mouse as the mouse moves, but to copy we would have to highlight the tooltip ..  maybe we can open a github issue for this?

shifucun (approved):

PR #2495, opened 2023-03-25
Add SDG topic page and config generation

shifucun:


server/tests/lib/subject_page_config_content_test.py
```diff
@@ -108,6 +108,8 @@ def test_required_fields(self):
         "sustainability": [libutil.get_disaster_sustainability_config()],
     })
     for id, configs in all_configs.items():
+      if id == 'sdg':
```
chejennifer, line 111: curious why you're skipping sdg config because it should also be verified in the same way?

server/routes/topic_page.py
```diff
@@ -67,12 +71,24 @@ def topic_page(topic_id=None, place_dcid=None):
...
-  for config in topic_configs:
-    if place_dcid in config.metadata.place_dcid:
-      topic_place_config = config
-      break
-  if not topic_place_config:
-    return "Error: no config found"
+  if topic_id == _SDG_TOPIC:
+    topic_place_config = topic_configs[0]
+    topic_place_config.metadata.place_dcid.append(place_dcid)
+    for category in topic_place_config.categories:
```
chejennifer, line 77: maybe add comments that this is to populate all the tiles with comparison places

chejennifer (commented):
nice!

server/config/topic_page/sdg/sdg.textproto
```diff
@@ -0,0 +1,13794 @@
+  metadata {
+    topic_id: "sdg"
+    topic_name: "SDG"
```
pradh, line 3: Should we expand SDG?

tools/sdg/main.py
```diff
@@ -0,0 +1,95 @@
...
+
+
+def build_config():
+  resp = post(API_PATH_SVG_INFO, {'nodes': [_SDG_ROOT]})
+  l0 = resp['data'][0]['info']
+  page_config = SubjectPageConfig()
+  page_config.metadata.topic_id = 'sdg'
+  page_config.metadata.topic_name = 'SDG'
+  for svg in l0['childStatVarGroups']:
+    print(svg['id'])
```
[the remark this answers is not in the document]
shifucun: Yea, it's used to view the progress

server/config/topic_page/sdg/sdg.textproto
```diff
@@ -0,0 +1,13794 @@
...
+  }
+  categories {
+    title: "01: No Poverty"
```
pradh, line 6: Does this need to be 01 to get it sorted, or will the order be decided by the ordering here?

If order can be chosen, can we remove "0" before 1?  Natalie noted how that is a bit odd.

tools/sdg/main.py
```diff
@@ -0,0 +1,98 @@
...
+import os
+
+from google.protobuf import text_format
+import requests
+from subject_page_pb2 import SubjectPageConfig
+from subject_page_pb2 import Tile
+
+logging.getLogger().setLevel(logging.INFO)
+
+_SDG_ROOT = "sdg/g/SDG"
```
pradh, line 25: @n-h-diaz  we should change this when SDG hierarchy is regenerated

server/config/topic_page/sdg/sdg.textproto
pradh: Ok, ignore comments about titles.  I think those should come from the KG given the number of SVs involved

pradh (approved):
Nice!  We might need to see this and tweak...

server/tests/lib/subject_page_config_content_test.py
chejennifer: I think that will be a bigger change because right now topic pages for the same topic will have different charts for different places. So maybe we should revisit how we match places to configs. 

Maybe for now, we just ignore the place verification for sdg (which I see you've added), but can you add a TODO to revisit places?

chejennifer (approved):
thanks for the updates!

PR #2500, opened 2023-03-27
Add url for addtional places in topic page

shifucun:


server/routes/topic_page.py
```diff
@@ -30,6 +31,8 @@
 _SDG_COMPARISON_PLACES = [
     'country/USA', 'country/CHN', 'country/JPN', 'country/IND'
 ]
+# Max number of places; Choose 6 to match the bar_tile bar limits.s
```
chejennifer, line 34: limits.

server/routes/topic_page.py
```diff
@@ -75,12 +78,20 @@ def topic_page(topic_id=None, place_dcid=None):
...
     topic_place_config.metadata.place_dcid.append(place_dcid)
     # Populuate comparison places (fixed places) for all tiles
+    comparison_places = [place_dcid]
+    more_places = request.args.getlist('places')
+    if more_places:
+      comparison_places.extend(more_places)
+    for p in _SDG_COMPARISON_PLACES:
+      if p not in comparison_places:
+        comparison_places.append(p)
+        if len(comparison_places) == _MAX_NUM_PLACES:
```
chejennifer, line 88: nit: check this first in each step of the for-loop (before `if p not in comparison_places:`) because comparison_places could already have a length of 6 when we enter the for loop?

server/routes/topic_page.py
```diff
@@ -75,12 +78,20 @@ def topic_page(topic_id=None, place_dcid=None):
     topic_place_config = topic_configs[0]
     topic_place_config.metadata.place_dcid.append(place_dcid)
     # Populuate comparison places (fixed places) for all tiles
+    comparison_places = [place_dcid]
```
chejennifer, line 81: nit: create a private helper function for getting the topic_place_config for sdg (lines 78-96) because this if case has a lot of specific logic just for sdg

chejennifer (approved):
nice!

server/routes/topic_page.py
shifucun: Done

server/routes/topic_page.py
shifucun: Done

server/routes/topic_page.py
shifucun: good point

PR #2501, opened 2023-03-27
[bar tile] handle x-axis label click for bar charts

chejennifer:
- allow subject page config to specify a root url for the x-axis labels to use when clicked. If not specified, clicking the x axis labels on bar tiles will take user to place explorer page.

sdg page (which has specified root url):
screen-recording-_84_

poverty page (no specified root url)
screen-recording-_85_


shifucun (approved):
thanks for adding this!

PR #2502, opened 2023-03-28
Show all place names from url params

shifucun:
This is not the most cleanest way to do this in topic page. If we want to support multiple places for all topic pages, might need to re-structure the data fetch.

### DIFF
--- a/server/templates/topic_page.html
+++ b/server/templates/topic_page.html
@@ -31,7 +31,7 @@
 
 {% block content %}
 <div id="body" class="container-fluid"></div>
-<div id="metadata" class="d-none" data-place-dcid="{{ place_dcid }}" data-topic-id="{{ topic_id }}">
+<div id="metadata" class="d-none" data-place-dcid="{{ place_dcid }}" data-more-places="{{ more_places }}" data-topic-id="{{ topic_id }}">
   <div id="place-name" data-pn="{{ place_name }}"></div>
   <div id="place-type" data-pt="{{ place_type }}"></div>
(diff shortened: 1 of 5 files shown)


pradh (approved):

server/templates/topic_page.html
```diff
@@ -31,7 +31,7 @@
 
 {% block content %}
 <div id="body" class="container-fluid"></div>
-<div id="metadata" class="d-none" data-place-dcid="{{ place_dcid }}" data-topic-id="{{ topic_id }}">
+<div id="metadata" class="d-none" data-place-dcid="{{ place_dcid }}" data-more-places="{{ more_places }}" data-topic-id="{{ topic_id }}">
```
chejennifer, line 34: maybe we can just get the names of these places already and pass a list of NamedPlace objects

chejennifer (commented):
Sorry for the late review, just a small suggestion that can be updated in followup PR

server/templates/topic_page.html
shifucun: That way, name fetch blocks the page rendering? 

I see the PageSelector had an UI (but commented out now) to pick places and fetch names. I wonder eventually these can be merged together.

chejennifer: I think the rest of the apps that use subject pages blocks on getting names of places so blocking should be ok, but maybe this will also need some more thinking 

Yeah, I remember that, not sure if we're ever going to re-instate that UI though

PR #2507, opened 2023-03-28
Fix topic page more places bug and increase max bars to 7

shifucun:


chejennifer (approved):
thanks for the fix! Maybe check that bar charts on NL still look ok with the number of bars updated to 7

PR #2509, opened 2023-03-29
Support modification of custom BT tables in deploy_gke.sh

Fructokinase:
This script allows something like this:
./scripts/deploy_gke.sh -p iqss-gdc  -t import_group1_test_2023_03_24_12_00_04 -h f730596

https://gdc.dataverse.org/tools/statvar#sv=Count_MedicalConditionIncident_Contract_COVID

### DIFF
--- a/scripts/deploy_gke.sh
+++ b/scripts/deploy_gke.sh
@@ -124,14 +124,14 @@ gcloud config set project $PROJECT_ID
 gcloud container clusters get-credentials $CLUSTER_NAME --region $REGION
 kubectl apply -f kustomize-build.yaml
 
-Deploy Cloud Endpoints
+# Deploy Cloud Endpoints
 export SERVICE_NAME="website-esp.endpoints.$PROJECT_ID.cloud.goog"
 export API_TITLE=$SERVICE_NAME
 cp $ROOT/gke/endpoints.yaml.tpl endpoints.yaml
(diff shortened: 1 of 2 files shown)


scripts/deploy_gke.sh
```diff
@@ -87,6 +91,18 @@ if [[ $PROJECT_ID != "" ]]; then
   cd $ROOT/deploy/overlays
   cp custom_kustomization.yaml.tpl kustomization.yaml
   sed -i '' "s/<PROJECT_ID>/$PROJECT_ID/g" kustomization.yaml
+  export PROJECT_ID=$PROJECT_ID
+  yq eval -i '.project = env(PROJECT_ID)' ../base/custom_bigtable_info.yaml
+  yq eval -i '.instance = "dc-graph"' ../base/custom_bigtable_info.yaml
```
shifucun, line 96: no need to set instance?

shifucun (approved):

scripts/deploy_gke.sh
Fructokinase: It's an empty file right now. Not sure if the emptiness was intended so it was set dynamically.

shifucun: Ok, then should set the instance.

Fructokinase: done.

Issue #2397, opened 2023-03-08
NL: Detect multiple SVs in a query


pradh:
To support queries like [prevalence of asthma vs. temperature rise across california]

PR #2510, opened 2023-03-29
Implement multi-sv detection

pradh:
Roughly as described in [this doc](https://docs.google.com/document/d/1BBQabxodKdDZON7b8Sc_C49in-AZ2LTUkPzXoHZLQjs/edit?pli=1&resourcekey=0-AYbBt44Dh7S1OQOm8ztYNA#heading=h.v6k65row53ow).  Not hooked up to fulfillment yet.  Propagated to debug info (screenshot below).

This PR moves the stop-word removal done for single-sv detection from website server to nl server.  This is so that multi-sv detection can be done with the stop-words present.

BUG=https://github.com/datacommonsorg/website/issues/2397

image

### DIFF
--- a/static/css/shared/_subject_page.scss
+++ b/static/css/shared/_subject_page.scss
@@ -165,19 +165,26 @@ article.category {
 #subject-page-sidebar {
   padding-top: 100px;
   padding-right: 30px;
+  margin-left: -0.5rem;
 
-  .nav-item a {
+  .nav-item {
     font-weight: 300;
     color: var(--dc-gray);
     display: inline-block;
     font-size: 0.9rem;
     line-height: 1.2rem;
-    padding: 0.3rem 0;
+    padding: 0.3rem 0.5rem;
     text-decoration: none;
+    cursor: pointer;
   }
 
-  .nav-item.category a {
+  .nav-item:hover {
+    background-color: rgba(0, 0, 0, 0.05);
+    border-radius: 0.5rem;
+  }
+
+  .nav-item.category {
     font-weight: 600;
-    padding-top: 1em;
+    margin-top: 1em;
   }
 }
--- a/static/css/topic_page.scss
+++ b/static/css/topic_page.scss
@@ -32,11 +32,6 @@
   }
 }
 
-#dc-topic-page a {
-  color: inherit;
-  text-decoration: underline;
-}
-
 #dc-topic-page p {
   font-size: .95rem;
 }
--- a/static/js/chart/draw_scatter.ts
+++ b/static/js/chart/draw_scatter.ts
@@ -757,13 +757,14 @@ export function drawScatter(
     yPerCapita: boolean
   ) => JSX.Element
 ): void {
+  const container = d3.select(svgContainerRef.current);
+  container.selectAll("*").remove();
   const svgContainerWidth = svgContainerRef.current.offsetWidth;
   const svgXTranslation =
     properties.width < svgContainerWidth
       ? (svgContainerWidth - properties.width) / 2
       : 0;
-  const svg = d3
-    .select(svgContainerRef.current)
+  const svg = container
     .append("svg")
     .attr("id", "scatterplot")
     .attr("width", properties.width)
--- a/static/js/components/subject_page/sidebar.tsx
+++ b/static/js/components/subject_page/sidebar.tsx
@@ -74,8 +74,9 @@ function renderItem(
     <li
       key={randDomId()}
       className={`nav-item ${isCategory ? "category" : ""}`}
+      onClick={() => window.open(`#${redirectItemId}`, "_self")}
     >
-      <a href={`#${redirectItemId}`}>{title}</a>
+      {title}
     </li>
   );
 }
--- a/static/js/components/tiles/bar_tile.tsx
+++ b/static/js/components/tiles/bar_tile.tsx
@@ -20,7 +20,7 @@
 
 import axios from "axios";
 import _ from "lodash";
-import React, { useEffect, useState } from "react";
+import React, { useCallback, useEffect, useRef, useState } from "react";
 
 import { DataGroup, DataPoint } from "../../chart/base";
 import { drawGroupBarChart } from "../../chart/draw";
@@ -36,6 +36,7 @@ import { getUnit } from "../../utils/stat_metadata_utils";
 import { getDateRange } from "../../utils/string_utils";
 import { getStatVarName, ReplacementStrings } from "../../utils/tile_utils";
 import { ChartTileContainer } from "./chart_tile";
+import { useDrawOnResize } from "./use_draw_on_resize";
 
 const NUM_PLACES = 7;
 
@@ -67,6 +68,7 @@ interface BarChartData {
 }
 
 export function BarTile(props: BarTilePropType): JSX.Element {
+  const chartContainerRef = useRef(null);
   const [rawData, setRawData] = useState<PointApiResponse | undefined>(null);
   const [barChartData, setBarChartData] = useState<BarChartData | undefined>(
     null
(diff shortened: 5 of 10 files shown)


nl_server/embeddings.py
```diff
@@ -22,11 +24,60 @@
...
 
+from nl_server import query_util
 import nl_server.gcs as gcs
+from server.lib.nl import utils
 
 TEMP_DIR = '/tmp/'
 MODEL_NAME = 'all-MiniLM-L6-v2'
 
+# A value higher than the highest score.
+_INIT_SCORE = 1.1
```
shifucun, line 35: Maybe have _HIGHEST_SCORE and use (_HIGHEST_SCORE + 0.1) for _INIT_SCORE so the purpose is more apparent?

nl_server/embeddings.py
```diff
@@ -22,11 +24,60 @@
...
+_NUM_CANDIDATES_PER_NSPLIT = 3
+
+
+# List of SV candidates, along with scores.
+@dataclass
+class VarCandidates:
+  # The below are sorted and parallel lists.
+  svs: List[str]
+  scores: List[float]
+  sv2sentences: Dict[str, List[str]]
```
shifucun, line 53: Might not worth a change for now, but why not having a list of {sv, score, sentences} struct?

nl_server/embeddings.py
```diff
@@ -22,11 +24,60 @@
...
+  svs: List[str]
+  scores: List[float]
+
+
+# One multi-var candidate containing multiple parts.
+@dataclass
+class MultiVarCandidate:
+  parts: List[MultiVarCandidatePart]
+  # Aggregate score
+  score: float
```
shifucun, line 70: use "aggregate_score" explicitly?

nl_server/embeddings.py
```diff
@@ -22,11 +24,60 @@
...
+  # Is this candidate based on a split computed from delimiters?
+  delim_based: bool
+
+
+# List of multi-var candidates.
+@dataclass
+class MultiVarCandidates:
+  candidates: List[MultiVarCandidate]
+  sv2sentences: Dict[str, List[str]]
```
shifucun, line 79: why sv2sentences is not in MultiVarCandidate? what is they key (one stat var) mean in a multiVarCandidates scenario.

nl_server/embeddings.py
```diff
@@ -80,52 +131,196 @@ def get_embedding_at_index(self, index: int) -> List[float]:
...
-
-    # Sort by scores
-    sv2score_sorted = sorted(sv2score.items(),
-                             key=lambda item: item[1],
-                             reverse=True)
-    svs_sorted = [k for (k, _) in sv2score_sorted]
-    scores_sorted = [v for (_, v) in sv2score_sorted]
-
-    sv_index_sorted = [sv2index[k] for (k, _) in sv2score_sorted]
+    query2sv2sentence2score: Dict[str, Dict[str, Dict[str, float]]] = {}
```
shifucun, line 142: have a comment with an example of query2sv2sentence2score?

nl_server/embeddings.py
```diff
@@ -80,52 +131,196 @@ def get_embedding_at_index(self, index: int) -> List[float]:
...
+            query2sv2sentence2score[q][dcid] = {}
+
+          if ent['corpus_id'] >= len(self.sentences):
+            continue
+          sentence = self.sentences[ent['corpus_id']]
+          query2sv2sentence2score[q][dcid][sentence] = score
+
+    query2result: Dict[str, VarCandidates] = {}
+
+    for q, sv2score in query2sv2score.items():
```
shifucun, line 163: Have a comment for this and the next large for loop on what it processes?

nl_server/embeddings.py
```diff
@@ -80,52 +131,196 @@ def get_embedding_at_index(self, index: int) -> List[float]:
...
+    result_multivar = self._detect_multiple_svs(orig_query)
 
     return {
-        'SV': svs_sorted,
-        'CosineScore': scores_sorted,
-        'EmbeddingIndex': sv_index_sorted,
-        'SV_to_Sentences': all_svs_sentences,
+        'SV': result_monovar.svs,
+        'CosineScore': result_monovar.scores,
+        'SV_to_Sentences': result_monovar.sv2sentences,
```
shifucun, line 203: Use CamelCase too?

nl_server/embeddings.py
```diff
@@ -80,52 +131,196 @@ def get_embedding_at_index(self, index: int) -> List[float]:
...
+    query2result = self._search_embeddings(list(all_queries))
+
+    #
+    # We take the average score from the top SV from the query-parts in a
+    # queryset (ignoring any queryset with a score below threshold). Then
+    # sort all candidates by that score.
+    #
+    # TODO: Come up with a better ranking function.
+    #
+    for qs in querysets:
```
shifucun, line 237: Some examples of "querysets" is helpful, since its fields are very deep

pradh: Done, PTAL.

pradh (commented):
Thanks for the review Bo!

shifucun (approved):

nl_server/embeddings.py
```diff
@@ -80,52 +131,201 @@ def get_embedding_at_index(self, index: int) -> List[float]:
   def get_embedding(self, query: str) -> List[float]:
     return self.model.encode(query).tolist()
 
-  def detect_svs(self, query: str) -> Dict[str, Union[Dict, List]]:
-    query_embeddings = self.model.encode([query])
+  #
+  # Given a list of queries, searches the in-memory embeddings index
+  # and returns a map of candidates keyed by input queries.
+  #
+  def _search_embeddings(self, queries: List[str]) -> Dict[str, VarCandidates]:
```
jehangiramjad, line 138: Question: why do the refactor of this function in the same PR as the introduction of a new function for detecting multi svs? Is there some functionality that was missing which this refactor addresses?

nl_server/query_util.py
```diff
@@ -0,0 +1,148 @@
...
+class QuerySet:
+  # The number of splits
+  nsplits: int
+  # Was it based on delimiter?
+  delim_based: bool
+  # Different combinations of a query
+  combinations: List[QuerySplit]
+
+
+#
```
jehangiramjad, line 62: It would be great to document this with some examples. With the recursive calls inside this function, highly recommended extensive documentation because it is fairly hairy to figure out what's going on. I also recommend more comments to help step line by line.

nl_server/embeddings.py
```diff
@@ -80,52 +131,201 @@ def get_embedding_at_index(self, index: int) -> List[float]:
...
+
+  #
+  # Detects one or more SVs from the query.
+  # TODO: Fix the query upstream to ensure the punctuations aren't stripped.
+  #
+  def _detect_multiple_svs(self, query: str) -> MultiVarCandidates:
+    #
+    # Prepare a combination of query-sets.
+    #
+    querysets = query_util.prepare_multivar_querysets(query)
```
jehangiramjad, line 222: Reading the comments below about a queryset, wouldn't it be better to use this combinatorial API/function built in to python? https://www.geeksforgeeks.org/itertools-combinations-module-python-print-possible-combinations/

jehangiramjad (commented):
Apologies for the delay. Some comments which might be good to discuss

nl_server/embeddings.py
pradh: good idea.  doing in the follow on PR.

nl_server/embeddings.py
pradh: The refactor takes multiple queries and returns results, which was needed for the multi SV case.

PR #2513, opened 2023-03-29
[topic pages] some UI updates

chejennifer:
- remove underline in the legend when it is not clickable
- handle window resize
- add UI treatment so that sidebar looks clickable

screen-recording-_87_


static/js/chart/draw_scatter.ts
```diff
@@ -757,13 +757,14 @@ export function drawScatter(
     yPerCapita: boolean
   ) => JSX.Element
 ): void {
+  const container = d3.select(svgContainerRef.current);
+  container.selectAll("*").remove();
```
shifucun, line 761: what is this change for?

static/js/components/tiles/bar_tile.tsx
```diff
@@ -83,9 +85,18 @@ export function BarTile(props: BarTilePropType): JSX.Element {
...
     }
+    const debouncedHandler = _.debounce(() => {
+      draw(props, barChartData);
+    }, RESIZE_DEBOUNCE_INTERVAL_MS);
+    const resizeObserver = new ResizeObserver(debouncedHandler);
+    resizeObserver.observe(chartContainerRef.current);
+    return () => {
+      resizeObserver.unobserve(chartContainerRef.current);
+      debouncedHandler.cancel();
+    };
```
shifucun, line 99: Wonder if this can be made as a custom React effect as it's used in many places

shifucun (approved):

static/js/components/tiles/bar_tile.tsx
shifucun: Something like this? https://usehooks.com/useDebounce/

static/js/chart/draw_scatter.ts
chejennifer: since we redraw when the window resizes, we need to remove the last drawn chart first. We do this for all our other charts too

static/js/components/tiles/bar_tile.tsx
chejennifer: done

shifucun (approved):
Thanks for the update! Very nice!

chejennifer:
thanks for the review!

PR #2519, opened 2023-03-30
Factor out common code between server and nl_server

pradh:
This moves out code from server.lib.nl.utils and server.lib.nl.constants into shared.lib.

There is no logic change in this PR.

TODO: Update docker script


shifucun (commented):
Can you copy "shared" folder in the Dockerfile too?

pradh:
> Can you copy "shared" folder in the Dockerfile too?

Done

PR #2528, opened 2023-03-31
[NL Interface] Updating embeddings by removing the dc/g SVGs

jehangiramjad:


pradh (commented):
Thanks!

I'm unable to see the diffs.  To confirm, the only change to the spreadsheet is removal of rows with values that start with "dc/g/" in "Id" column?

I see 3 remaining IDs in the current spreadsheet, is that intentional?

dc/g/Farm_FarmType-Cropland
dc/g/Person_Age_MedicalCondition-AllTeethLoss
dc/g/Person_Age-Years65Onwards_MedicalCondition-AllTeethLoss

jehangiramjad:
> Thanks!
> 
> I'm unable to see the diffs. To confirm, the only change to the spreadsheet is removal of rows with values that start with "dc/g/" in "Id" column?
> 
> I see 3 remaining IDs in the current spreadsheet, is that intentional?
> 
> dc/g/Farm_FarmType-Cropland dc/g/Person_Age_MedicalCondition-AllTeethLoss dc/g/Person_Age-Years65Onwards_MedicalCondition-AllTeethLoss

This for noticing this. They were left over and are now removed. Yes, that's the only change in the sheet.

pradh (approved):

pradh:
Unfortunately website-pull-request-py and website-pull-request-nl doesn't run when we update this.  Looks like this breaks the former because the SV results/scores shift.

PR #2531, opened 2023-04-03
[sv hierarchy] remove blocklisted svgs

chejennifer:
- remove svgs that have been cleaned up from the list of blocklisted svgs. Also remove updating num descendents for svgs because only uncategorized is blocklisted now which is a top level svg so skipping that category won't affect any numbers


shifucun (approved):

shifucun:
thanks!

Issue #2527, opened 2023-03-31
NL: ContainedIn places at the same level as the containing place should not trigger a contained in chart


jehangiramjad:
Example query: how does the GDP in the US compare to other countries

Expected behavior(s): it should either just keep showing the US GDP or go one level higher for the containing place (world/earth?) to show comparison across countries. Given the current implementation limitations, just ignoring the contained in classification is probably best when the containing place + contained in type are both the same (Country).

pradh:
Fixed in #2538

Issue #2474, opened 2023-03-24
NL: Make sure topic/dc/Temperature does not conflict with demo results


lucy-kind:
"What are the projected temperature extremes across California" will pick up topic/dc/temperature due to the title "Temperature" receiving a high embedding rank. Disable this topic from occurring with this demo query. Then add topic to embeddings once it does not conflict with the demo anymore

lucy-kind:
Issue when adding topic:

![image](https://user-images.githubusercontent.com/63422135/227393391-b6a3284f-500b-45bb-a21e-6678a7b849ef.png)
![image](https://user-images.githubusercontent.com/63422135/227393401-76c639ba-417a-47bd-b3fa-26af373fce43.png)

Issue #2467, opened 2023-03-23
NL: Expand containedIn logic to handle more ambiguous place types like "places in" and "regions in"


juliawu:
We should support queries that ask for contained in places, but don't specify an exact admin area level. For example "US regions with the most unbalanced gender ratio" should show a ranking of states/counties/cities in the US by gender ratio, picking state/county/city depending on which admin area level we have data for.

Example queries:
* least religious place in USA
* US regions with the most unbalanced gender ratio

pradh:
Fixed in #2538 

eg [US regions with the most unbalanced gender ratio] shows states in US for some (somewhat unrelated) SVs.

PR #2538, opened 2023-04-04
NL: Fix 3 fulfillment bugs

pradh:
This includes fixes for the following three bugs:
* https://github.com/datacommonsorg/website/issues/2474: validated by integration tests, a unit test and screenshot below
* https://github.com/datacommonsorg/website/issues/2527: validated by a new integration test
* https://github.com/datacommonsorg/website/issues/2467: see screenshots below

### Enable Temperature

image

### country in US

image

### "regions in" "places in"

image

image


server/lib/nl/detection.py
```diff
@@ -133,9 +133,8 @@ class ContainedInPlaceType(str, Enum):
   # Typically corresponds to county equivalent
   EU_NUTS_3 = "EurostatNUTS3"
 
-  # Across is a generic containedInPlaceType which determines if the
-  # query is using the word "across".
-  ACROSS = "Across"
+  # Indicates that the fulfiller should guess the child-type.
```
jehangiramjad, line 136: nit: while this works, it feel a little "hacky", no? "Guess" is an action and not a containedin place type. Wouldn't it be better to map to some OTHER type which then triggers the guessing action?

server/lib/nl/fulfiller.py
```diff
@@ -102,12 +103,26 @@ def rank_charts(utterance: Utterance):
...
+  blocked_vars = set()
   while (i < len(sv_list)):
     if (sv_score[i] >= _SV_THRESHOLD):
-      ans.append(sv_list[i])
-    i = i + 1
+      var = sv_list[i]
+
+      # Check if an earlier var blocks this var.
+      if var in blocked_vars:
+        counters.info("blocked_vars", var)
```
jehangiramjad, line 118: nit: perhaps have a const for this string?

jehangiramjad (approved):
Thanks. A couple of nitty observations

server/lib/nl/detection.py
pradh: OTHER doesn't give any further info, but GUESS says something downstream should guess the type?  If we want it changed DEFAULT_CHILD_TYPE could be an option.

Let me know if you prefer that, and I can change in a follow on.

pradh:
Thanks for the change!  I'll submit and handle followup comments in future PR.

PR #2539, opened 2023-04-04
Set versions of Flask to 2.0.0 in server and nl_server

pradh:
The 2.2.x version does not handle trailing slashes well (e.g., https://datacommons.org/place/geoId/06/).


nl_server/requirements.txt
```diff
@@ -1,16 +1,17 @@
 datasets==2.8.0
 diskcache==5.4.0
-Flask==1.1.4
+Flask==2.2.3
```
shifucun, line 3: I have tested that 2.0.0 (Flask and Werkzeug) would work for all the tests. Can you update that and this PR should be good to go

pradh: done.

shifucun (approved):

PR #2542, opened 2023-04-05
Use Flask 2.1.3 and Werkzeug 2.1.2 to fix trailing slash issue

shifucun:
Latest Flask (2.2.3) does not handle /path/ as /path well https://github.com/pallets/werkzeug/issues/2533. Use an older version before a good fix is available.

### DIFF
--- a/nl_server/requirements.txt
+++ b/nl_server/requirements.txt
@@ -1,17 +1,18 @@
 datasets==2.8.0
 diskcache==5.4.0
-Flask==1.1.4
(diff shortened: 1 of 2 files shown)


pradh:
Thanks!

It doesn't look like py and nl tests have run, could you try running them explicitly?

image

pradh:
When I tried 2.2.x, the nl test failed, and I had to do the __init__.py change in https://github.com/datacommonsorg/website/pull/2539/files

PR #2544, opened 2023-04-05
update embeddings based on mar 21 triage

lucy-kind:
Updates on the following:
topic of income
topic of temperature
income inequality
lawyers and legal service workers

image
image
image
image


run_server.sh
```diff
@@ -55,12 +55,12 @@ while getopts ":e:p?m?d?l" OPTION; do
...
-  echo "ERROR: Please use protoc version: ${PROTOC_VERSION}" 1>&2
-  exit 1
-fi
+# if [[ "$(protoc --version)" != "libprotoc ${PROTOC_VERSION}" ]]; then
+#   echo "ERROR: Please use protoc version: ${PROTOC_VERSION}" 1>&2
+#   exit 1
+# fi
 
-export GOOGLE_CLOUD_PROJECT=datcom-website-dev
+# export GOOGLE_CLOUD_PROJECT=datcom-website-dev
```
pradh, line 63: Please revert this change?

deploy/base/model.yaml
```diff
@@ -1 +1 @@
-embeddings_file: embeddings_us_filtered_2023_03_31_19_43_56.csv
+embeddings_file: embeddings_us_filtered_2023_04_04_14_51_24.csv
```
pradh, line 1: should there be an associated update to the CSV file?

lucy-kind: hm I updated it from the online sheet. i can try to run it again

lucy-kind: ok great, now the csv shows up and I checked the screenshot prompts: same results :)

run_server.sh
lucy-kind: done!

tools/nl/embeddings/sheets/embeddings_us_filtered.csv
```diff
@@ -1244,3 +1244,4 @@ Number,Name,Id,Description Strings,PaLM Generated Alternatives,ChatGPT descripti
...
 ,,dc/topic/Temperature,,,what's the temperature like; what's the climate here; tell me about the weather;,Temperature,,what's the temperature like; what's the climate here; tell me about the weather;;Temperature;,
+,,dc/235cmp60ywm46,number of lawyers; people working in law; how many legal service workers are there; population working in legal services; population of legal service workers,,,Population Working in Legal Services,,number of lawyers; people working in law; how many legal service workers are there; population working in legal services; population of legal service workers;Population Working in Legal Services;,
```
pradh, line 1247: "number of lawyers" may not be right because these are people working in legal services.?

(Another is dc/2w4xphkt6vs62 which is offices of lawyers, but I think legal services is probably broader)

pradh (commented):
Thanks

lucy-kind:
I started with the offices of lawyers, but that feels like it would include
workers not in the purview of the triage query (eg cleaners, facility
workers etc). what would be the best place for "number of lawyers", or
should we leave that out and then just have embeddings direct to here as
closest related sv for now and note "missing dataset" for number of lawyers
specifically?

lucy-kind:
@pradh removed number of lawyers from alt descriptions

pradh (approved):

Issue #2515, opened 2023-03-30
NL: Multi-SV correlation


jehangiramjad:
Query: how does fertility rate correlate with GDP across countries in the world
Expected result: correlation plot between fertility rate and GDP for all counties.

pradh:
This is implemented now: https://screenshot.googleplex.com/BAoQndQkPTW9HsU

PR #2547, opened 2023-04-06
NL: Implement multi-sv fulfillment for correlation

pradh:
Generally, try to be conservative in fulfilling multi-sv, so we fulfill multi-sv to show correlations only when:
1. There's a COMPARISON or CORRELATION attribute detected (words like `compare`, `vs`, etc.)
2. The aggregate multi-sv score exceeds the max single-sv score

Additional changes:
* Promote a SIMPLE query with multiple places to COMPARISON_ACROSS_PLACES.
* If a correlation query doesn't have a place-type, use DEFAULT_TYPE
* In COMPARISON_ACROSS_PLACES, if we find the current utterance has multiple places, don't bother continuing into the context
* Add unit and integration tests

Related bugs: https://github.com/datacommonsorg/website/issues/2397, https://github.com/datacommonsorg/website/issues/2515

Screenshot:

image

### DIFF
--- a/server/lib/nl/fulfillment/correlation.py
+++ b/server/lib/nl/fulfillment/correlation.py
@@ -102,6 +102,8 @@ def _populate_correlation_for_place(state: PopulateState, place: Place) -> bool:
     # Counter updated in get_sample_child_places
     return False
 
+  # When multi-sv has a higher score than single-sv prefer that.
+  # See discussion in _route_comparison_or_correlation().
   if utils.is_multi_sv(state.uttr):
     lhs_svs, rhs_svs = _handle_multi_sv_in_uttr(state.uttr, places_to_check)
   else:
--- a/server/lib/nl/fulfillment/correlation.py
+++ b/server/lib/nl/fulfillment/correlation.py
@@ -13,10 +13,13 @@
 # limitations under the License.
 
 import logging
+from typing import List
 
 from server.lib.nl import utils
 from server.lib.nl.detection import ClassificationType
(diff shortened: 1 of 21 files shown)


server/lib/nl/fulfillment/correlation.py
```diff
@@ -93,48 +102,91 @@ def _populate_correlation_for_place(state: PopulateState, place: Place) -> bool:
     # Counter updated in get_sample_child_places
     return False
 
+  if utils.is_multi_sv(state.uttr):
+    lhs_svs, rhs_svs = _handle_multi_sv_in_uttr(state.uttr, places_to_check)
+  else:
+    lhs_svs, rhs_svs = _handle_svs_from_context(state.uttr, places_to_check)
```
jehangiramjad, line 108: Clarifying question: so we give preference to multi_sv case over the context-based case when it comes to correlations? Any intuition for why that is the preference order?

jehangiramjad (approved):
Thanks! A lot going on in the PR but I think the follow :)

server/lib/nl/fulfillment/correlation.py
pradh: `is_multi_sv` being true is already a strong indication of the presence of multiple SVs (because that suggests the aggregate score of multi-sv exceeds the best single-sv).  but we are being conservative in acting on such a higher multi-sv score, by also requiring a comparison/correlation.

I'll add a note to refer to comments in `_route_comparison_or_correlation` where there's a discussion of this.

pradh (commented):
Thanks Jehangir!

PR #2551, opened 2023-04-06
NL: Use a score different threshold for multi-sv

pradh:
This is to handle cases where multi-sv is behind very slightly from the single-sv score.  Also don't truncate multi-sv scores.

Before:

image

After:

image


server/lib/nl/utils.py
```diff
@@ -753,6 +753,14 @@ def is_multi_sv(uttr: nl_uttr.Utterance) -> bool:
...
   top_sv_score = uttr.detection.svs_detected.sv_scores[0]
   top_multi_sv_score = uttr.multi_svs.candidates[0].aggregate_score
-  return top_multi_sv_score > top_sv_score
+
+  # Prefer multi-sv when the scores are higher or up to a score differential.
+  if top_multi_sv_score > top_sv_score:
+    return True
+  if (top_sv_score - top_multi_sv_score <=
+      shared_constants.MULTI_SV_SCORE_DIFFERENTIAL):
+    return True
```
jehangiramjad, line 765: nit: may be combine both these conditions in one statement?

nl_server/embeddings.py
```diff
@@ -38,10 +39,6 @@
 # Scores below this are ignored.
 _SV_SCORE_THRESHOLD = 0.5
 
-# If the difference between successive scores exceeds this threshold, then SVs at
-# the lower score and below are ignored.
-_MULTI_SV_SCORE_DIFFERENTIAL = 0.05
-
 _NUM_CANDIDATES_PER_NSPLIT = 3
```
jehangiramjad, line 42: nit: given that the other constant was moved to the constants file, why not also this?

jehangiramjad (approved):
Nice!

server/lib/nl/utils.py
pradh: Done/

nl_server/embeddings.py
pradh: That is a constants file in shared lib, so only having stuff shared between nl_server and server there.

pradh (commented):
Thanks for the review!

PR #2553, opened 2023-04-06
Reorder chart titles for median age of female population svs

lucy-kind:
Somehow these got scrambled... reordered them to match the correct SV.


pradh (approved):
Thanks for finding/fixing.

PR #2556, opened 2023-04-07
NL: Implement numeric quantity parser for use in filter queries

pradh:
Use a bunch of heuristics for matching ==, >, <, >=, <=.  Currently only supports digits, with an optional multiplier.

image


pradh:
This was merged as part of https://github.com/datacommonsorg/website/pull/2559, so closing.

PR #2557, opened 2023-04-07
[import wizard] add multiVarCol mapping functionality

chejennifer:
- adds the mapping functionality for a multi var col template

Notes:
- does not include making the UI pretty or layout changes from the mock

screen-recording-_90_

downloaded tmcf:
Screenshot 2023-04-06 at 10 25 17 PM


shifucun (commented):
This is very cool!

I wonder can you add some kind of test from UI change to the TMCF text? This will be very helpful for future changes.

chejennifer:
> This is very cool!
> 
> I wonder can you add some kind of test from UI change to the TMCF text? This will be very helpful for future changes.

We actually have tests for all the util functions (the functions that generate the TMCF, example observations, validation, etc): https://github.com/datacommonsorg/website/tree/master/static/js/import_wizard2/utils

And because the UI is going to be updated, maybe better to add tests for interacting with UI later?

shifucun (approved):

shifucun:
> > This is very cool!
> > I wonder can you add some kind of test from UI change to the TMCF text? This will be very helpful for future changes.
> 
> We actually have tests for all the util functions (the functions that generate the TMCF, example observations, validation, etc): https://github.com/datacommonsorg/website/tree/master/static/js/import_wizard2/utils
> 
> And because the UI is going to be updated, maybe better to add tests for interacting with UI later?

That SG. I feel eventually we want to have some golden .tmcf files for testing. Even UI changes, but they would always generate the same TMCF.

chejennifer:
ah yes, golden tmcf files makes sense!

Thanks for the review!

PR #2559, opened 2023-04-08
NL: Implement fulfillment for filter queries

pradh:
Note: This is a followup to https://github.com/datacommonsorg/website/pull/2556 (at the moment includes those diffs too).

Implements fulfillment for filter queries in two scenarios:
1) We identified a quantity and multi-sv is not dominant.  In this case, we treat it as a contained-in fulfillment with the quantity filter.
2) We identified a quantity and dual-sv is dominant. In this case, we make a simplifying assumption that the SVs detected in the later part of the query is for filtering and the SVs detected in the earlier part of the query is for selection (display).

Below are two example screenshots for each case.

### Single Var

image

image

### Dual Vars

image

image


server/lib/nl/fulfillment/filter_with_dual_vars.py
```diff
@@ -0,0 +1,157 @@
...
+from server.lib.nl.utterance import ChartOriginType
+from server.lib.nl.utterance import ChartType
+from server.lib.nl.utterance import Utterance
+
+# TODO: Support per-capita
+# TODO: Increase this after frontend handles comparedPlaces better.
+_MAX_PLACES_TO_RETURN = 7
+
+
+def _get_dual_sv_part(uttr: Utterance, idx: int) -> List[str]:
```
jehangiramjad, line 36: nit: since this is picking the dual part and that's always the second one, why not fix the idx = 1 here instead of relying on the caller to specify it? If the caller specifies idx = 0, then that means this won't be a "dual" part extractor?

server/lib/nl/fulfillment/filter_with_dual_vars.py
```diff
@@ -0,0 +1,157 @@
...
+  found = populate_charts(
+      PopulateState(uttr=uttr,
+                    main_cb=_populate_cb,
+                    place_type=place_type,
+                    quantity=quantity))
+  uttr.svs = orig_svs
+  return found
+
+
+def _populate_cb(state: PopulateState, chart_vars: ChartVars,
```
jehangiramjad, line 74: nit: The _populate_cb() function here has overlap with the similar function in the filter_with_single_var.py file. Is there room to move things in to utils etc?

server/lib/nl/quantity_parser.py
```diff
@@ -0,0 +1,188 @@
...
+     _digits(
+         r'(?:<=|less than or equal to|less than or equals|lesser than or equal to|lesser than or equals|upto|up to)'
+     )),
+    (QCmpType.GT,
+     _digits(
+         r'(?:>|over|above|exceed|exceeds|more than|greater than|higher than|upwards of)'
+     )),
+    (QCmpType.LT,
+     _digits(r'(?:<|under|below|less than|lesser than|lower than)')),
+    (QCmpType.EQ, _digits(r'(?:==?|is|equal to|equals|has|have|as much as)'))
```
jehangiramjad, line 79: nit: extra space between "equal" and "to"?

pradh:
> Thanks! Very cool. But it was a big PR so I am not sure my review is super thorough

Oh, half of the PR is from https://github.com/datacommonsorg/website/pull/2556.  Github doesn't allow diffbase like Critique (at least I'm not aware of how to).

server/lib/nl/fulfillment/filter_with_dual_vars.py
pradh: This function needed comments (added now). It is not always picking the 2nd one, rather it finds the dual-sv candidate (from the multi-sv results), and for that candidate returns either the first or second part, called from different places.

server/lib/nl/fulfillment/filter_with_dual_vars.py
pradh: Good point.  Added TODO.

server/lib/nl/quantity_parser.py
pradh: I think there was only 1 space, and validated in this unit test check - https://github.com/datacommonsorg/website/blob/1b1a045721bf72401ff4854bdb5a9f8f5e5a245a/server/tests/lib/nl/quantity_parser_test.py#L32

pradh (commented):
Thanks for the review Jehangir!

PR #2560, opened 2023-04-10
[Custom DC] Cleanup permission Terraform module

Fructokinase:
This is being replaced by internal scripts.


shifucun (approved):
Just to confirm, these are not used since they are in one time setup?

Fructokinase:
> Just to confirm, these are not used since they are in one time setup?

Not from the customer side!

PR #2561, opened 2023-04-11
[scatter] Updated scatter app test

dwnoble:
This PR fixes some issues in the scatter plot "all functionalities" test case. Previously, a bug in the test's promise/callback structure was preventing it from actually enforcing some of the nested assertions.

Also:
* Flattened nested promises in "all functionalities" test
* Added SVGElement stubs for methods that only exist in the browser
* Updated axios mocks to reflect current call chain workflow

### DIFF
--- a/static/js/tools/scatter/app.test.tsx
+++ b/static/js/tools/scatter/app.test.tsx
@@ -85,17 +85,6 @@ function mockAxios(): void {
       },
     });
 
-  // Statvar data
-  when(axios.get)
-    .calledWith("/api/variable-group/info")
-    .mockResolvedValue({
-      data: {
-        absoluteName: "Data Commons Variables",
-        childStatVarGroups: [],
-        descendentStatVarCount: 0,
-      },
-    });
-
   // Population and statvar data
   const facets = {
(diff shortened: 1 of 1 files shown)


shifucun:
Thanks for the change Dan! Please run `./run_test.sh -f` to fix the lint error

static/js/tools/scatter/app.test.tsx
```diff
@@ -69,8 +85,18 @@ function mockAxios(): void {
       },
     });
 
-  // Population and statvar data
+  // Statvar data
+  when(axios.get)
```
chejennifer, line 89: curious why this mocked call is necessary because I believe `/api/variable-group/info` should always be called with the dcid of the variable group we are trying to get info for

shifucun:
> 

That works. thanks!

chejennifer (approved):
Thanks for updating the test!

static/js/tools/scatter/app.test.tsx
dwnoble: Good catch- verified that the test passes without this.

PR #2563, opened 2023-04-11
Only load bigtable module when LOG_QUERY is enabled

shifucun:
Now server/services/bigtable.py is always imported hence require bigtable client is always created. This gives GCP auth error for 3P developers that do not have corresponding access.


pradh (approved):
Thanks!

PR #2564, opened 2023-04-11
[scatter] Add ability to size scatter plot points based on population

dwnoble:
This PR adds the ability to size scatter plot points based on available population data. Includes options for linear and logarithmic scales, as well as pulling population values from either the x or y axis.

Demo:
dc-resize-points-by-population

Details:

* New URL parameters to control point sizing 
  * [optional] Enable sizing by population
     * Param: `pp`
     * Type: `boolean` (accepts `1` or `0`)
     * Default: Disabled (`0`)
  * [optional] Size dots using a log or linear scale
     * Param: `pl`
     * Type: `boolean` (accepts `1` or `0`)
     * Default: Linear (`0`)
     * Set to 1 to use a log scale
  * [optional] Use populations from x-axis or y-axis observations
     * Param: `px`
     * Type: `boolean` (accepts `1` or `0`)
     * Default: Y-Axis (`0`)
     * Set to `1` to use population values from X-Axis
* URL parameters can be controlled through UI components

Implementation notes
* Updated `scatter_data_utils.ts:getPlaceAxisChartData` to always include population data (previously population data was only relevant and included for per-capita axis scaling)
* By default, points are sized based on the total population of the respective geo (using the “Count_Person” variable). If population scale denominators are enabled (e.g, when per-capita axes are turned on), point sizing will align with the variable used in the denominator scale (using the dx/dy from the denominator url parameters, which is by default is also Count_Person)

Possible follow-on work
* Add population values to point pop-up/hover window 
* Add legend describing population ranges at each point size


chejennifer:
I have not taken a deeper look into the code changes yet, but some questions I had about this change:
- When population sizing is enabled, is the density also updated? I think we would want the density to take into account both the number of points in an area AND the population of each of those points in the area when population sizing is enabled
- I think the UI for the population sizing selection is a little confusing with multiple sets of selection inputs in one row (the row essentially has the controls for: turning on/off population sizing, choosing whether the sizing should be from x or y axis, and choosing whether we want log scale or not). 
    - Something like this might be a little more clear?
Screenshot 2023-04-11 at 3 45 47 PM

dwnoble:
@chejennifer Updated the UI based on your suggestions.

I also took a stab at incorporating population into the point density colors when scaling by population is enabled. it's not a huge difference in color- what do you think?

dc-3

dwnoble:
Forgot to mention- with the last change, i updated the URL parameters, removing the `pl` parameter entirely, and updating `pp` from a boolean to a string. In summary:

* New-new URL parameters to control point sizing:
  * [optional] Enable sizing by population
     * Param: `pp`
     * Type: `string` (accepts `"linear"`, `"log"`, or no value `""`)
     * Default: Disabled (`""`)
  * [optional] Use populations from x-axis or y-axis observations
     * Param: `px`
     * Type: `boolean` (accepts `1` or `0`)
     * Default: Y-Axis (`0`)
     * Set to `1` to use population values from X-Axis

chejennifer (commented):
Thanks for the updates!

Maybe when population is selected, we should also include it in the tooltip, but ok to do this in a follow up PR.

dwnoble:
That sounds reasonable- it took a little while to understand the x/y population myself. I like the idea of using the latest population among the x/y points. I'll update to that approach

static/js/tools/scatter/context.ts
```diff
@@ -112,13 +117,17 @@ interface DisplayOptionsWrapper {
   showLabels: boolean;
   chartType: ScatterChartType;
   showDensity: boolean;
+  showPopulation: PointScaleState;
+  showPopulationX: boolean;
```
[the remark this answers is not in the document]
dwnoble: Removed this state entirely

static/js/tools/scatter/util.ts
```diff
@@ -246,6 +220,19 @@ export function applyHashBoolean(
   return val === "1";
 }
 
+export function applyHashPopulation(
```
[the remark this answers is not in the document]
dwnoble: Done

static/js/chart/draw_scatter.ts
```diff
@@ -368,6 +388,8 @@ function addDensityLegend(
  * @param chartWidth the width of the chart area
  * @param chartHeight the height of the chart area
  * @param marginTop margin between top of the chart area and the top of the container
+ * @param showPopulationX factor population into density calculation
+ * @param pointSizeScale d3 scale for
```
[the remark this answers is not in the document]
dwnoble: Good catch. Done.

static/js/chart/draw_scatter.ts
```diff
@@ -368,6 +388,8 @@ function addDensityLegend(
  * @param chartWidth the width of the chart area
  * @param chartHeight the height of the chart area
  * @param marginTop margin between top of the chart area and the top of the container
+ * @param showPopulationX factor population into density calculation
```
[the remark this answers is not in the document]
dwnoble: Good catch. Ended up removing this param entirely.

static/js/chart/draw_scatter.ts
```diff
@@ -422,6 +451,45 @@ function addDensity(
     .attr("stroke-width", 0);
 }
 
+/**
+ * Gets a d3 scale to size a list of points based on population
+ * @param points
+ * @param showPopulationX
+ * @param logScale
```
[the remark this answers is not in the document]
dwnoble: Done

dwnoble:
OK, now we're down to one net-new URL parameter, and the interface is looking cleaner:

Screenshot 2023-04-18 at 9 25 06 PM

Single new URL parameter:
* [optional] Enable sizing by population
   * Param: `pp`
   * Type: `string` (accepts `linear`, `log`, or no value `` )
   * Default: Disabled (``)

static/js/shared/ga_events.test.tsx
```diff
@@ -342,11 +344,14 @@ const SCATTER_CONTEXT = {
...
     chartType: ScatterChartType.SCATTER,
     showDensity: false,
+    showPopulation: SHOW_POPULATION_OFF,
     showRegression: false,
     setQuadrants: () => null,
     setLabels: () => null,
     setChartType: () => null,
     setDensity: () => null,
+    setPopulation: () => null,
+    setPopulationX: () => null,
```
chejennifer, line 354: remove this?

static/js/tools/scatter/util.ts
```diff
@@ -246,6 +220,25 @@ export function applyHashBoolean(
...
 
+/**
+ * Fetches a population point scale state from the given url query params
+ * @param params url query parameters
+ * @param key query parameter key containing population point scale state
+ * @returns PointScaleState value
+ */
+export function applyHashPopulation(
+  params: URLSearchParams,
+  key: string
```
chejennifer, line 231: nit: don't need to pass in the key because will always use the same key (`FieldToAbbreviation.showPopulation`) to get the population from the hash

static/js/chart/draw_scatter.ts
```diff
@@ -358,6 +361,31 @@ function addDensityLegend(
...
+ * Calculates scatter plot point size
+ * @param point scatter plot Point object
+ * @param pointSizeScale d3 scale for sizing point based on population
+ */
+function calculatePointSize(
+  point: Point,
+  pointSizeScale: ScatterScale
+): number {
+  const population = getPointPopulation(point);
+  const pointSize = pointSizeScale(population);
```
chejennifer, line 385: since population could potentially return undefined, curious what `pointSizeScale(undefined)` will return?

chejennifer (commented):
thanks for the changes!

static/js/chart/draw_scatter.ts
dwnoble: Good callout- i added a comment to the function to clarify this.

If the population is undefined, the result from the scale will also be undefined. That is accounted for in the next line:

```
return pointSize || DEFAULT_POINT_SIZE;
```

Example
```
$ node
> d3 = require('d3')
> scale = d3.scaleLinear().domain([5,10]).range([10,20])
> scale(5)
10
> scale(undefined)
undefined
```

static/js/tools/scatter/util.ts
dwnoble: done

static/js/shared/ga_events.test.tsx
dwnoble: done

chejennifer (approved):
thanks for all the updates, this looks good to me, but let's hold off merging for now!

PR #2566, opened 2023-04-12
[import wizard] Update mapping page

chejennifer:
- update mapping page layout to follow the mocks
- extract shared code from the mapping section of each template type either into the parent mapping page or the individual input components

Screenshot 2023-04-12 at 9 12 25 AM


static/js/import_wizard2/components/mapping_templates/multi_var_mulit_date_col.tsx
```diff
@@ -34,106 +33,44 @@ import { MappingHeaderInput } from "../shared/mapping_header_input";
...
+      />
+      <MappingColumnInput
+        mappedThing={MappedThing.UNIT}
+        mappingVal={props.userMapping.get(MappedThing.UNIT)}
+        onMappingValUpdate={(mappingVal) =>
+          props.onMappingValUpdated(MappedThing.UNIT, mappingVal)
+        }
+        orderedColumns={props.csvData.orderedColumns}
+        isRequired={false}
+      />
```
shifucun, line 73: These 4 components pattern is repeated too. Can you have another component to wrap and parameterize them?

shifucun (commented):
nice simplification!

static/js/import_wizard2/components/mapping_templates/multi_var_mulit_date_col.tsx
chejennifer: from offline discussion, since different templates will use a different combination of inputs, will keep this as is

shifucun (approved):

PR #2567, opened 2023-04-12
[Custom DC] Separate mixer and website Part 1

Fructokinase:


deploy/helm_charts/envs/dev.yaml
```diff
@@ -0,0 +1,61 @@
+# Helm config
```
shifucun, line 1: enableModel=true should be set here?

Fructokinase: done.

shifucun (approved):

deploy/helm_charts/envs/dev.yaml
Fructokinase: done.

PR #2590, opened 2023-04-20
Update and simplify website service group helm config

shifucun:


deploy/helm_charts/envs/dev.yaml
```diff
@@ -44,35 +44,15 @@ ingress:
 
 serviceAccount:
   name: website-ksa
-  gsaName: website-reobot
+  gsaName: website-robot
```
shifucun, line 47: @Fructokinase is gsaName used? This typo indicates not?

Fructokinase: Not used, thanks for finding out.

Fructokinase (approved):

deploy/helm_charts/envs/dev.yaml
shifucun: Removed it.

PR #2594, opened 2023-04-20
[import wizard] add input validation

chejennifer:
- validates that date is in ISO-8601 format
- validates that variable doesn't have any whitespace

screen-recording-_93_

### DIFF
--- a/static/js/import_wizard2/utils/validation.ts
+++ b/static/js/import_wizard2/utils/validation.ts
@@ -154,3 +154,12 @@ export function checkMappings(mappings: Mapping): Array<string> {
 
   return errors;
 }
+
+/**
+ * Returns whether a variable is valid.
+ * @param variable the variable to check
+ */
+export function isValidVariable(variable: string): boolean {
+  // checks that there are no spaces within the string.
+  return !/\s/.test(variable.trim());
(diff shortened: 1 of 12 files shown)


static/js/import_wizard2/utils/validation.ts
```diff
@@ -154,3 +154,12 @@ export function checkMappings(mappings: Mapping): Array<string> {
...
   return errors;
 }
+
+/**
+ * Returns whether a variable is valid.
+ * @param variable the variable to check
+ */
+export function isValidVariable(variable: string): boolean {
+  // checks that there are no spaces within the string.
+  return !/\s/.test(variable.trim());
```
shifucun, line 164: not sure if trim handles all the while spaces /t, /n etc

shifucun (approved):

static/js/import_wizard2/utils/validation.ts
chejennifer: looking at the documentation, looks like trim will remove all the whitespace from either end of the string
Screenshot 2023-04-20 at 10 51 54 AM

chejennifer:
thanks for the review!

PR #2595, opened 2023-04-21
Fix deploy_gke_helm for custom instances

Fructokinase:
- Do not enable mixer ingress
- Specify custom project id for image


scripts/deploy_gke_helm.sh
```diff
@@ -126,6 +126,7 @@ function deploy_mixer() {
   --timeout 10m \
   --force  \
   -f "deploy/helm_charts/envs/$ENV.yaml" \
```
shifucun, line 128: how about $ENV here, would "custom" just work?

shifucun (approved):

PR #2596, opened 2023-04-21
[timeline] fix unit bug

chejennifer:


static/js/tools/timeline/chart.tsx
```diff
@@ -333,8 +333,12 @@ class Chart extends Component<ChartPropsType, ChartStateType> {
     for (const statVar in statData.data) {
       for (const place in statData.data[statVar]) {
         const series = statData.data[statVar][place];
-        if (series && series.facet && statData.facets[series.facet].unit) {
-          units.add(statData.facets[series.facet].unit);
+        if (!series || !series.facet || !statData.facets[series.facet]) {
```
shifucun, line 336: what is the actual statData looks like in this case?

shifucun (approved):

static/js/tools/timeline/chart.tsx
chejennifer: looked into it and there was a problem with updating the statData when there was a mix of modelStatVars and non modelStatVars where only the facet information for modelStatVars were included. Updated this PR with a fix for this.

shifucun (approved):
thanks!

chejennifer:
thanks for the review!

PR #2606, opened 2023-04-24
[download] only get facet information if inputs are valid

chejennifer:


shifucun (approved):
thanks for the fix!

chejennifer:
thanks for the review!

PR #2610, opened 2023-04-26
Use V2 properties API; Fix flask cache warnings

shifucun:


chejennifer (approved):
nice! thanks for fixing the warnings

shifucun:
thanks for review!

PR #2611, opened 2023-04-26
[V2 API] Migrate property values API call

shifucun:
The functions in datacommons.py should be a direct wrapper of REST api result with no modification.

If the response needs processing, the function should live in server/lib/util.py.


server/lib/util.py
```diff
@@ -420,3 +420,16 @@ def gzip_compress_response(raw_content, is_json):
   if is_json:
     response.headers['Content-Type'] = 'application/json'
   return response
+
+
+def property_values(nodes, prop, out=True):
```
chejennifer, line 425: maybe add a function comment

server/lib/nl/utils.py
```diff
@@ -525,28 +525,26 @@ def get_immediate_parent_places(
     main_place_dcid: str, parent_place_type: str,
     counters: ctr.Counters) -> List[detection.Place]:
   start = time.time()
-  resp = dc.property_values_v1([main_place_dcid], 'containedInPlace')
+  resp = util.property_values([main_place_dcid], 'containedInPlace')
   counters.timeit('get_immediate_parent_places', start)
   results = []
-  for item in resp.get('data', []):
-    if item.get('node', '') != main_place_dcid:
+  for value in resp['data'][main_place_dcid].get('arcs', {}).get(
```
chejennifer, line 531: does resp have a 'data' field & an 'arcs' field? Because doesn't util.property_values just return a dict of dcid to property values?

shifucun: Good catch! This one should use the raw response as it fetches name and type. This also explains the NL test failure.

chejennifer (approved):
thanks for the update!

server/lib/nl/utils.py
```diff
@@ -525,28 +525,26 @@ def get_immediate_parent_places(
     main_place_dcid: str, parent_place_type: str,
     counters: ctr.Counters) -> List[detection.Place]:
   start = time.time()
-  resp = dc.property_values_v1([main_place_dcid], 'containedInPlace')
+  resp = dc.property_values([main_place_dcid], 'containedInPlace')
   counters.timeit('get_immediate_parent_places', start)
   results = []
-  for item in resp.get('data', []):
-    if item.get('node', '') != main_place_dcid:
+  for value in resp['data'][main_place_dcid].get('arcs', {}).get(
```
pradh, line 531: nit: to avoid key access should we do:  `resp.get('data', {}).get(main_place_dcid, {}).get(...)` ?

pradh (approved):

server/lib/nl/utils.py
shifucun: 'data' and <place> always exist in the response, but changed to get() to make it more robust

PR #2612, opened 2023-04-26
Remove rich search (experimental work with Pragma team)

shifucun:
DC is building new NL models and interfaces now. Removing this feature as it's not in dev or maintenance state anymore.


nopper (approved):

shifucun:
Thanks @nopper

PR #2619, opened 2023-04-27
[scatter] fix population bug in scatter tile

chejennifer:
- scatter tiles sometimes will have null populationData causing an error when trying to access populationData during the data processing

### DIFF
--- a/static/js/utils/scatter_data_utils.ts
+++ b/static/js/utils/scatter_data_utils.ts
@@ -32,7 +32,7 @@ interface PlaceAxisChartData {
   value: number;
   statDate: string;
   sources: string[];
-  pop?: number;
+  popValue?: number;
   popDate?: string;
   unit?: string;
 }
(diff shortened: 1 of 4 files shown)


static/js/utils/scatter_data_utils.ts
```diff
@@ -78,23 +78,27 @@ function getPlaceAxisChartData(
...
-  let pop = null;
-  let popDate = null;
-  if (popSeries && popSeries.series) {
-    const popObs = getMatchingObservation(popSeries.series, obs.date);
-    if (
-      popBounds &&
-      (!popObs || !isBetween(popObs.value, popBounds[0], popBounds[1]))
-    ) {
-      return null;
+  let pop = denomValue;
```
shifucun, line 81: how about popValue to be specific

shifucun (approved):

static/js/utils/scatter_data_utils.ts
chejennifer: done

PR #2620, opened 2023-04-27
Organize routes (html, api) by pages

shifucun:
As I do the V2 API client migration. I found quite some similar web server API endpoints spreading across UI page specific pages and shared API util modules. Also with the increased number of UI pages, it's hard to track which endpoints/functions are used by which UI feature.

This PR re-organizes the modules under server/routes, so there is a sub-folder for each UI page, containing "html.py" and "api.py". All other api endpoints are put under /server/routes/shared_api by the functional names.


chejennifer (approved):
this is very clean, thanks for the change!

shifucun:
thanks!

PR #2623, opened 2023-04-28
[V2 API] Migrate place stat var union/existence and place name related APIs

shifucun:
* Use /info/place API to get parent places instead of using get_prop_value a few times.
* Get rid of "cached" version of some data fetch functions. The dc.Fn() call result is already cached, and/or there is cache for flask endpoints response.

### DIFF
--- a/server/routes/place/api.py
+++ b/server/routes/place/api.py
@@ -373,8 +373,8 @@ def get_i18n_all_child_places(raw_page_data):
   for place_type in list(all_child_places.keys()):
     for place in all_child_places[place_type]['places']:
       all_dcids.append(place.get('dcid', ''))
-  i18n_names = place_api.get_i18n_name(all_dcids,
-                                       False)  # Don't resolve en-only names
+  # Don't resolve en-only names
+  i18n_names = place_api.get_i18n_name(all_dcids, False)
   for place_type in list(all_child_places.keys()):
     for place in all_child_places[place_type]['places']:
       dcid = place.get('dcid')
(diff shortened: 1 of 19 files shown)


server/routes/place/api.py
```diff
@@ -566,7 +566,7 @@ def populate_additional_category_data(category):
   all_places = [dcid]
   for t in BAR_CHART_TYPES:
     all_places.extend(raw_page_data.get(t + 'Places', []))
-  names = place_api.get_display_name('^'.join(sorted(all_places)), g.locale)
```
pradh, line 569: This is intentionally dropped?

server/routes/shared_api/place.py
```diff
@@ -853,17 +747,15 @@ def api_ranking_chart(dcid):
...
-      parent_type_list = parent.get("types", [])
       parent_place_dcid = parent.get("dcid", "")
-      if parent_type_list:
-        parent_type = parent_type_list[0]
-        # All wanted place types plus continent except CensusZipCodeTabulationArea.
-        if parent_type == "Continent" or (
-            parent_type in ALL_WANTED_PLACE_TYPES and
-            parent_type != "CensusZipCodeTabulationArea"):
-          break
+      parent_type = parent.get('type', '')
```
pradh, line 753: super nit: only this line as different quote style, be consistent?

pradh (approved):
Not super familiar with the code.  LGTM to unblock.

Maybe Jennifer could also look on Monday?

server/routes/place/api.py
shifucun: Yes, the function is simplified by only taking a list of dcids.

server/routes/shared_api/place.py
shifucun: fixed

PR #2627, opened 2023-05-01
[stat var hierarchy] separate blocklist svgs for each environment

chejennifer:
- Get blocklisted svgs from environment file
- set blocklisted svgs file in each helm chart

### DIFF
--- a/deploy/helm_charts/envs/custom.yaml
+++ b/deploy/helm_charts/envs/custom.yaml
@@ -54,3 +54,6 @@ nl:
 
 customDC:
   enabled: true
+
+svg:
+  blocklistFile: ["dc/g/Uncategorized", "dc/g/SDG", "sdg/g/SDG"]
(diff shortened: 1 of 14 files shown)


deploy/helm_charts/envs/custom.yaml
```diff
@@ -54,3 +54,6 @@ nl:
 
 customDC:
   enabled: true
+
+svg:
+  blocklistFile: ["dc/g/Uncategorized", "dc/g/SDG", "sdg/g/SDG"]
```
shifucun, line 59: So this blocks 3 svg from all the custom DC (including Stanford). I wonder what is the plan to update this later on?

chejennifer: "dc/g/Uncategorized" was always blocked in all staging/production instances of the website & in terms of the sdg groups being blocklisted, this change was requested in order to block those from showing up on the hierarchy for Stanford DC .. so not sure what the future plan is?

shifucun (approved):

shifucun (commented):
Can you sync the mixer submodule, otherwise this and future commits before sync would not work.

PR #2633, opened 2023-05-02
[sv hierarchy] Fix configmap problem

chejennifer:
- There were deployment problems when both website and mixer had configmaps of the same name (blocklist-svg)


chejennifer:
Upon offline chat, decided to fix deployment by only having the website deploy a blocklist svg configmap. Closing this.

PR #2635, opened 2023-05-02
[V2] Migrate places in API and consolidate /v2/node service functions

shifucun:


server/lib/fetch.py
```diff
@@ -310,10 +312,25 @@ def triples(nodes, out=True):
...
     for prop, val in arcs.get('arcs', {}).items():
       result[node][prop] = val.get('nodes', [])
-  return result
\ No newline at end of file
+  return result
+
+
+def descendent_places(nodes, descendent_type):
+  return property_values(nodes,
+                         'containedInPlace+',
```
chejennifer, line 326: wonder if it would be better to handle the "+" in the property_values function instead so that the caller just has to pass the proper properties and constraints without needing to know how to build the query string? Same with raw_property_values

chejennifer (commented):
nice!

server/routes/shared_api/place.py
```diff
@@ -610,35 +609,29 @@ def api_display_name():
...
 
-@bp.route('/places-in-names')
+@bp.route('/descendent/name')
 @cache.cached(timeout=3600 * 24, query_string=True)  # Cache for one day.
-def get_places_in_names():
+def descendent_names():
   """Gets names of places of a certain type contained in a place.
 
   Returns:
       Dicts keyed by child place DCIDs with their names as values.
```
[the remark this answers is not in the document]
shifucun: Done

server/lib/fetch.py
shifucun: + is part of the composite property. If making it an argument of property_values(), the caller would still need to set it.

This is the reason to have this wrapper function descendent_places(), and all actual callers should just call this function, and would never need to know about "+"

chejennifer: by handle it in property_values, I meant property_values would still take the same arguments but add a "+" if there are constraints (because from what I understand, we add a "+" if there are constraints?)

shifucun: In the V2 syntax, "+" is like regex, meaning 1 or more chained prop, so in this case it is:
<-conainedInPlace<-conainedInPlace<-...<-containedInPlace.

The constraints {p:v} is independent from that.. meaning any node in the chain that meets the pv constraints.

chejennifer: ohh I see, ok, that makes sense!

chejennifer (approved):

PR #2636, opened 2023-05-02
Add a tool to diff SV index

pradh:
Please see details in the README.md

Example: https://shanth-dc-test.tiiny.site/

And to confirm basic test sanity, running with the same version as both base and test gives no diffs.


tools/nl/svindex_differ/differ.py
```diff
@@ -0,0 +1,106 @@
...
+  with open(query_file) as f:
+    for row in csv.reader(f):
+      if not row:
+        continue
+      query = row[0].strip()
+      if not query or query.startswith('#') or query.startswith('//'):
+        continue
+      assert ';' not in query, 'Multiple query not yet supported'
+      base_result = _prune(base.detect_svs(query))
+      test_result = _prune(test.detect_svs(query))
```
jehangiramjad, line 87: One assumption here is that the model being used for to compute the embeddings is the same (for the base and test but also that the same model is being used to for the "search" over embeddings as was used to build the embeddings in the first place). I think this should be very clearly documented in the README as well.

One other suggestion is that we should add a TODO in the embeddings generation script to append the model name to the embeddings file? That way, we can always check that the model being used is the same?

tools/nl/svindex_differ/differ.py
```diff
@@ -0,0 +1,106 @@
...
+      if base_result != test_result:
+        diffs.append((query, base_result, test_result))
+
+  with open(output_file, 'w') as f:
+    f.write(template.render(diff_table=_diff_table, diffs=diffs))
```
jehangiramjad, line 92: nit: do we also somehow want to propagate the scores even if we don't use them to compare the embeddings? might be nice to observe

jehangiramjad (approved):
This is very very cool!

tools/nl/svindex_differ/differ.py
pradh: Good point.  Clarified in README and added TODO.

tools/nl/svindex_differ/differ.py
pradh: Maybe, in next iteration?  I've added TODO.

pradh (commented):
Thanks for the review!

PR #2643, opened 2023-05-03
Upgrade to Python 3.11

shifucun:
A few things updated together:

* Use light-the-torch to install cpu only torch
* Install en_core_web_lg with "-f" option, so no need to do separate check for en_core_web_lg
* Disable googlecloudprofiler as it has issue with the latest wheel setup. This is not used now.


server/__init__.py
```diff
@@ -219,18 +219,18 @@ def register_routes_common(app):
...
-
-    # Profiler initialization. It starts a daemon thread which continuously
-    # collects and uploads profiles. Best done as early as possible.
-    try:
-      # service and service_version can be automatically inferred when
-      # running on GCP.
-      googlecloudprofiler.start(verbose=3)
-    except (ValueError, NotImplementedError) as exc:
-      logging.error(exc)
+  # if os.environ.get('FLASK_ENV') in ['production', 'staging', 'autopush']:
```
chejennifer, line 222: should we just delete this if cloudprofiler isn't being used?

server/routes/place/html.py
```diff
@@ -30,7 +30,7 @@
 
 
 @bp.route('', strict_slashes=False)
-@bp.route('/<path:place_dcid>', strict_slashes=False)
+@bp.route('/<path:place_dcid>/', strict_slashes=False)
```
chejennifer, line 33: curious why we need the extra slash

chejennifer (approved):
thanks for upgrading!

server/__init__.py
shifucun: Removed

server/routes/place/html.py
shifucun: The latest Flask can not handle the trailing slash without specifying it explicitly.

pradh (approved):
Thanks a lot for doing this!

Very exciting!

jehangiramjad (approved):
LGTM! Many thanks :)

jehangiramjad (approved):
LGTM! Many thanks :)

PR #2646, opened 2023-05-04
Add a snapshot endpoint that uses selenium to return charts as JSON

pradh:
This takes about 10-15 seconds.

TODO: Add caching support


chejennifer (approved):
this is cool!

PR #2647, opened 2023-05-04
Move node server under static/

shifucun:


static/js/shared/place_search_bar.tsx
```diff
@@ -143,7 +143,7 @@ class PlaceSearchBar extends Component<PlaceSearchBarPropType> {
             </span>
           </div>
         );
-        container.prepend(ReactDOM.render(result, container));
+        ReactDOM.render(result, container);
```
chejennifer, line 146: curious if removing container.prepend changes anything on the UI?

chejennifer (approved):

static/js/shared/place_search_bar.tsx
shifucun: No, it was a misuse, ReactDOM.render does not return anything

PR #2648, opened 2023-05-04
Bumped datasets to 2.11

dwnoble:
Old datasets version was causing the error:

```
    return pd.read_csv(xopen(filepath_or_buffer, "rb", use_auth_token=use_auth_token), **kwargs)
TypeError: read_csv() got an unexpected keyword argument 'mangle_dupe_cols'
```

Reference: https://github.com/huggingface/datasets/issues/5744


shifucun:
This can be closed since we the version in HEAD is updated to 2.12.0 now.

PR #2649, opened 2023-05-04
line tile async fetch data

shifucun:
Refactor so the data fetch can be reused by server side rendering too


chejennifer (approved):
this looks clean!

PR #2650, opened 2023-05-04
Added ranking tile chart export from html to svg/image

dwnoble:
Pulls html table and puts it in an svg `<foreignObject>` tag.

Styling on the final SVG download needs work

dc-export-html-table

### DIFF
--- a/static/js/components/tiles/ranking_tile.tsx
+++ b/static/js/components/tiles/ranking_tile.tsx
@@ -20,7 +20,13 @@
 
 import axios from "axios";
 import _ from "lodash";
-import React, { useEffect, useRef, useState } from "react";
+import React, {
+  MutableRefObject,
+  RefObject,
+  useEffect,
+  useRef,
+  useState,
+} from "react";
 
 import { INITAL_LOADING_CLASS } from "../../constants/tile_constants";
 import { formatNumber } from "../../i18n/i18n";
@@ -71,6 +77,8 @@ export function RankingTile(props: RankingTilePropType): JSX.Element {
   const [rankingData, setRankingData] = useState<RankingData | undefined>(null);
   const embedModalElement = useRef<ChartEmbed>(null);
   const chartContainer = useRef(null);
+  const rankingUnitHighestContainer = useRef<HTMLDivElement>(null);
+  const rankingUnitLowestContainer = useRef<HTMLDivElement>(null);
 
   useEffect(() => {
     fetchData(props, setRankingData);
@@ -123,6 +131,7 @@ export function RankingTile(props: RankingTilePropType): JSX.Element {
                   <RankingUnit
                     key={`${statVar}-highest`}
                     unit={unit}
+                    forwardRef={rankingUnitHighestContainer}
                     scaling={scaling}
                     title={formatString(
                       props.title ||
@@ -142,7 +151,7 @@ export function RankingTile(props: RankingTilePropType): JSX.Element {
                   />
                   <ChartFooter
                     sources={sources}
-                    handleEmbed={() => handleEmbed(points.reverse())}
+                    handleEmbed={() => handleEmbed(points.reverse(), sources)}
                   />
                 </div>
               )}
@@ -151,6 +160,7 @@ export function RankingTile(props: RankingTilePropType): JSX.Element {
                   <RankingUnit
                     key={`${statVar}-lowest`}
                     unit={unit}
+                    forwardRef={rankingUnitLowestContainer}
                     scaling={scaling}
                     title={formatString(
                       props.title ||
@@ -171,7 +181,7 @@ export function RankingTile(props: RankingTilePropType): JSX.Element {
                   />
                   <ChartFooter
                     sources={sources}
-                    handleEmbed={() => handleEmbed(points)}
+                    handleEmbed={() => handleEmbed(points, sources)}
                   />
                 </div>
               )}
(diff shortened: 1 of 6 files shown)


static/js/components/tiles/ranking_tile.tsx
```diff
@@ -20,7 +20,13 @@
 
 import axios from "axios";
 import _ from "lodash";
-import React, { useEffect, useRef, useState } from "react";
+import React, {
+  MutableRefObject,
```
chejennifer, line 24: nit: i don't think these new imports are being used?

static/js/components/tiles/ranking_tile.tsx
```diff
@@ -71,6 +77,8 @@ export function RankingTile(props: RankingTilePropType): JSX.Element {
   const [rankingData, setRankingData] = useState<RankingData | undefined>(null);
   const embedModalElement = useRef<ChartEmbed>(null);
   const chartContainer = useRef(null);
+  const rankingUnitHighestContainer = useRef<HTMLDivElement>(null);
```
chejennifer, line 80: a ranking tile can have mulitple ranking units for highest & multiple ranking units for lowest (e.g., https://screenshot.googleplex.com/BbnFKQFz6LYN3FZ) .. we will probably need a ref for each ranking unit?

static/js/components/tiles/ranking_tile.tsx
```diff
@@ -182,15 +192,28 @@ export function RankingTile(props: RankingTilePropType): JSX.Element {
...
   );
 
-  function handleEmbed(rankingPoints: RankingPoint[]): void {
+  function handleEmbed(
+    rankingPoints: RankingPoint[],
+    sources: Set<string>
+  ): void {
+    let chartHtml = "";
+    let chartHeight = 0;
+    if (rankingUnitHighestContainer.current) {
```
chejennifer, line 201: i htink just checking highest and lowest probably is not sufficient? because user can be trying to export the lowest ranking unit, but there also is a highest ranking unit present. We probably want an id associated with each ranking unit so we can specify which one is being exported

static/js/place/chart_embed.tsx
```diff
@@ -92,22 +94,110 @@ class ChartEmbed extends React.Component<unknown, ChartEmbedStateType> {
...
+      sources,
     });
   }
 
+  /**
+   * Decorates chartHtml with title and provenance information embeded in an enclosing
+   * SVG node. Returns the SVG contents as a string.
+   */
+  private decorateChartHtml(): string {
+    const container = this.svgContainerElement.current;
```
chejennifer, line 120: can we make a helper for all the shared functionality between decorateChartHtml and decorateSvgChart?

chejennifer (commented):
this is cool!

static/js/components/tiles/ranking_tile.tsx
dwnoble: fixed

pradh (commented):
Just to note this is at odds with the scraper in https://github.com/datacommonsorg/website/pull/2646.  So can I get that in first so we can let those folks know?

static/js/components/tiles/ranking_tile.tsx
dwnoble: done

dwnoble:
Fixed multi-chart use case

dc-ranking-multi


Still need to refactor `decorateChartHtml`

dwnoble:
> > Just to note this is at odds with the scraper in #2646. So can I get that in first so we can let those folks know?
> 
> Thanks. Now that that change is merged, instead of extracting rows [here](https://github.com/datacommonsorg/website/blob/master/server/lib/nl/scraper.py#L144-L162), should we change to use the SVG?

That works for me

static/js/components/tiles/ranking_tile.tsx
```diff
@@ -54,6 +54,145 @@ interface RankingGroup {
...
+  svName: string;
+  svNames?: string[];
+  title?: string;
+  unit: string[];
+}
+
+/**
+ * Ranking unit tables for a stat var.
+ */
+function SvRankingUnits(props: SvRankingUnitsProps): JSX.Element {
```
chejennifer, line 86: Please only have one component per file

static/js/components/tiles/ranking_tile.tsx
```diff
@@ -54,6 +54,145 @@ interface RankingGroup {
   dateRange: string;
 }
 
+interface SvRankingUnitsProps {
```
chejennifer, line 57: since a lot of these props can be derived from RankingData and RankingTileSpec, maybe lets just pass those in so there aren't so many props to keep track of ..

static/js/components/tiles/ranking_tile.tsx
```diff
@@ -54,6 +54,145 @@ interface RankingGroup {
...
+  numDataPoints: number;
+  highestTitle?: string;
+  isMultiColumn?: boolean;
+  lowestTitle?: string;
+  points: RankingPoint[];
+  scaling: number[];
+  showHighest: boolean;
+  showLowest: boolean;
+  rankingCount: number;
+  showChart: (
```
chejennifer, line 68: nit: showChartEmbed or something like that? showChart makes me think this is just to show the chart in the tile

chejennifer (commented):
thanks for the update

static/js/components/tiles/ranking_tile.tsx
dwnoble: fixed

static/js/components/tiles/ranking_tile.tsx
dwnoble: done

static/js/components/tiles/ranking_tile.tsx
dwnoble: done

chejennifer (approved):
Thanks for the changes! Can you also add a TODO for the change that you didn't get to in chart_embed before merging

static/js/components/tiles/sv_ranking_units.tsx
```diff
@@ -0,0 +1,157 @@
+/**
+ * Copyright 2022 Google LLC
```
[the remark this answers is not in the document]
dwnoble: fixed

pradh:
> > > Just to note this is at odds with the scraper in #2646. So can I get that in first so we can let those folks know?
> > 
> > 
> > Thanks. Now that that change is merged, instead of extracting rows [here](https://github.com/datacommonsorg/website/blob/master/server/lib/nl/scraper.py#L144-L162), should we change to use the SVG?
> 
> That works for me

Do we want to make that change in this PR?

pradh:
> > > > Just to note this is at odds with the scraper in #2646. So can I get that in first so we can let those folks know?
> > > 
> > > 
> > > Thanks. Now that that change is merged, instead of extracting rows [here](https://github.com/datacommonsorg/website/blob/master/server/lib/nl/scraper.py#L144-L162), should we change to use the SVG?
> > 
> > 
> > That works for me
> 
> Do we want to make that change in this PR?

Follow on is fine too...

dwnoble:
> > > > > Just to note this is at odds with the scraper in #2646. So can I get that in first so we can let those folks know?
> > > > 
> > > > 
> > > > Thanks. Now that that change is merged, instead of extracting rows [here](https://github.com/datacommonsorg/website/blob/master/server/lib/nl/scraper.py#L144-L162), should we change to use the SVG?
> > > 
> > > 
> > > That works for me
> > 
> > 
> > Do we want to make that change in this PR?
> 
> Follow on is fine too...

Let's do it in a follow-up.

PR #2651, opened 2023-05-04
[tiles] update fetch functions for some tiles

chejennifer:
- updated fetch functions for bar, bivariate, highlight, map, scatter


shifucun (approved):
thanks!

PR #2653, opened 2023-05-04
[Node Server] Render line tile on server side and return svg

shifucun:
This does not fully work due to getBBox() function not available on the server side.

Check this in so we have an e2e flow to work on.

### DIFF
--- a/static/js/components/tiles/line_tile.tsx
+++ b/static/js/components/tiles/line_tile.tsx
@@ -134,7 +134,7 @@ export function draw(
   elem.innerHTML = "";
   const isCompleteLine = drawLineChart(
     props.id,
-    elem.offsetWidth || props.svgChartWidth,
+    props.svgChartWidth || elem.offsetWidth,
     props.svgChartHeight,
     chartData.dataGroup,
     false,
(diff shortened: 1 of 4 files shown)


static/js/components/tiles/line_tile.tsx
```diff
@@ -116,17 +124,17 @@ export const fetchData = async (props: LineTilePropType) => {
...
-  svgContainer: React.RefObject<HTMLElement>
+  svgContainer: HTMLElement
 ): void {
   const elem = document.getElementById(props.id);
   // TODO: Remove all cases of setting innerHTML directly.
   elem.innerHTML = "";
   const isCompleteLine = drawLineChart(
     props.id,
-    elem.offsetWidth,
+    elem.offsetWidth || props.svgChartWidth,
```
chejennifer, line 137: hm, did you want props.svgChartWidth to take priority or elem.offsetWidth? I would guess props.svgChartWidth because if you pass a width, maybe would expect it to be respected?

chejennifer (approved):
nice!

static/js/components/tiles/line_tile.tsx
shifucun: Good point. I am using it as a supplement since offsetWidth is 0 in jsdom. But good to swap.

Also I think for all the svg width, height, they should be explicitly passed in the draw function. Now it's a mix.

PR #2657, opened 2023-05-05
[tiles] merge all svgs in chart tile container and export

chejennifer:
- When a chart tile has mulitple svgs (e.g., map which has the map and the legend), in the export functionality, package the svgs together into 1 svg to be exported

Screenshot 2023-05-05 at 2 52 33 PM

- also fix a bug in the bivariate tile where multiple legends were showing up

### DIFF
--- a/static/js/chart/draw_bivariate.ts
+++ b/static/js/chart/draw_bivariate.ts
@@ -97,10 +97,9 @@ function drawMapLegend(
   xUnit?: string,
   yUnit?: string
 ): void {
-  const legendSvg = d3
-    .select(legendRef.current)
-    .append("svg")
-    .attr("width", svgWidth);
+  const container = d3.select(legendRef.current)
+  container.selectAll("*").remove();
+  const legendSvg = container.append("svg").attr("width", svgWidth);
   const legendContainer = legendSvg.append("g");
   const legendLabels = legendContainer
     .append("text")
--- a/static/js/chart/draw_bivariate.ts
+++ b/static/js/chart/draw_bivariate.ts
@@ -97,7 +97,7 @@ function drawMapLegend(
   xUnit?: string,
   yUnit?: string
 ): void {
-  const container = d3.select(legendRef.current)
+  const container = d3.select(legendRef.current);
   container.selectAll("*").remove();
   const legendSvg = container.append("svg").attr("width", svgWidth);
   const legendContainer = legendSvg.append("g");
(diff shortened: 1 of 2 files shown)


static/js/chart/draw_bivariate.ts
```diff
@@ -97,10 +97,9 @@ function drawMapLegend(
   xUnit?: string,
   yUnit?: string
 ): void {
-  const legendSvg = d3
-    .select(legendRef.current)
-    .append("svg")
-    .attr("width", svgWidth);
+  const container = d3.select(legendRef.current)
+  container.selectAll("*").remove();
```
shifucun, line 101: what's the effect of the change here?

static/js/components/tiles/chart_tile.tsx
```diff
@@ -81,22 +81,37 @@ export function ChartTileContainer(props: ChartTileContainerProp): JSX.Element {
...
+    // Set embedSvgHeight as the max height of all svgs
+    Array.from(svgElemList).forEach((svg) => {
+      embedSvgHeight = Math.max(svg.getBBox().height, embedSvgHeight);
+    })
+    // Add each svg from svgElemList to embedSvg
+    for (const svg of svgElemList) {
+      const svgBBox = svg.getBBox();
+      const clonedSvg = svg.cloneNode(true) as SVGElement;
+      // Set height of current svg to the height of the embedSvg
+      clonedSvg.setAttribute("height", `${embedSvgHeight}`);
```
shifucun, line 99: would this reshape the original svg? Say two svg are placed vertically, would this still work?

static/js/chart/draw_bivariate.ts
chejennifer: this fixes a bug in the bivariate tile where multiple legends were showing up because old legend was not getting removed

static/js/components/tiles/chart_tile.tsx
chejennifer: If the original chart had more than 1 svg, the svgs would be reshaped to be placed next to each other horizontally & stretched to the full height

shifucun: Once the svg is created, the path/line values are fixed, if just changing the height, would they still look good?

chejennifer: Yes, because
- if an svg has a viewbox attribute, changing the height will cause the svg to shift according to the aspect ratio
- if no viewbox set & height increases (embedSvgHeight is the max height of all the svgs), there won't be any change to the path/line values

shifucun (approved):

PR #2662, opened 2023-05-08
Remove kustomization overlays that have been moved to helm

shifucun:


chejennifer (approved):
thanks for the clean up!

PR #2666, opened 2023-05-09
[nl] update how scraper waits for charts to be loaded

chejennifer:
- instead of always waiting 5 seconds for charts to load, wait for loading placeholders to disappear, then wait for svgs to be drawn

comparing the time spent waiting between the old and new methods (look for time_taken & the time is in seconds):
old method: https://paste.googleplex.com/5697023377408000
new method: https://paste.googleplex.com/5627238262243328

### DIFF
--- a/server/lib/nl/scraper.py
+++ b/server/lib/nl/scraper.py
@@ -246,7 +246,7 @@ def scrape(query, driver):
   driver.get(_URL + query)
 
   # Wait until the page has loaded.
-  wait = WebDriverWait(driver, 30)
+  wait = WebDriverWait(driver, timeout=30, poll_frequency=0.25)
   wait.until(EC.presence_of_element_located((By.CLASS_NAME, 'chart-container')))
   wait.until(lambda d: _check_svgs_drawn(driver))
 
--- a/server/lib/nl/scraper.py
+++ b/server/lib/nl/scraper.py
@@ -218,19 +218,19 @@ def _to_svg(svg):
 
 
 def _wait_for_svgs_drawn(driver):
-  '''Wait (max _MAX_SVG_WAIT_TRIES * 0.5 seconds) for all svgs on a subject page
-  to finish drawing'''
+  """Wait (max _MAX_SVG_WAIT_TRIES * 0.5 seconds) for all svgs on a subject page
+  to finish drawing"""
   svgs_loaded = False
   num_tries = 0
   while not svgs_loaded and num_tries < _MAX_SVG_WAIT_TRIES:
-    subject_page = driver.find_element(By.ID, "subject-page-main-pane")
+    subject_page = driver.find_element(By.ID, 'subject-page-main-pane')
     has_incomplete_svg = False
-    for block in subject_page.find_elements(By.CLASS_NAME, "block"):
+    for block in subject_page.find_elements(By.CLASS_NAME, 'block'):
       # Check each chart in the block for any incomplete svgs
       for chart_container in block.find_elements(By.CLASS_NAME,
-                                                 "chart-container"):
+                                                 'chart-container'):
         svg_tag_to_check = ""
-        for chart_type_class in chart_container.get_attribute("class").split():
+        for chart_type_class in chart_container.get_attribute('class').split():
           if chart_type_class in _CHART_TYPE_TO_DRAWN_SVG_TAG:
             svg_tag_to_check = _CHART_TYPE_TO_DRAWN_SVG_TAG[chart_type_class]
             break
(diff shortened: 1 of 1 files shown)


server/lib/nl/scraper.py
```diff
@@ -205,19 +217,59 @@ def _to_svg(svg):
...
   driver.get(_URL + query)
 
-  # Wait until the test_class_name has loaded.
+  # Wait until the page has loaded.
   wait = WebDriverWait(driver, 30)
-  wait.until(EC.presence_of_element_located((By.CLASS_NAME, 'chart-container')))
   wait.until(
       EC.invisibility_of_element_located((By.CLASS_NAME, 'dot-loading-stage')))
-  time.sleep(5)
+  wait.until(EC.presence_of_element_located((By.CLASS_NAME, 'chart-container')))
```
pradh, line 268: To confirm, do we need the dot-loading-stage and chart-container still?

server/lib/nl/scraper.py
```diff
@@ -205,19 +217,59 @@ def _to_svg(svg):
...
+        if not drawn_objects:
+          has_incomplete_svg = True
+          break
+      # If any block finds an incomplete chart, break out of the loop to wait
+      # and re-check the charts
+      if has_incomplete_svg:
+        break
+    num_tries += 1
+    svgs_loaded = not has_incomplete_svg
+    if not svgs_loaded:
```
pradh, line 253: This sleep still adds delay.

Given that this is going to be in launched soon (beyond a demo), I feel it might be worth thinking if we can do better?  For example, is there a chance of setting a single global tag after everything is loaded?  Or is there at least a chance of doing something like:
for i in range(num_charts):
  wait.until(.presence_of_element_located(By.ID, f'chart-loaded-{i}')))

server/lib/nl/scraper.py
```diff
@@ -190,6 +190,18 @@
     }]
 }
 
+# Max tries to wait for all svgs to be drawn on a subject page
+_MAX_SVG_WAIT_TRIES = 10
```
pradh, line 194: How much e2e delay does this translate to?

pradh (commented):
Thanks!

server/lib/nl/scraper.py
chejennifer: actually, can probably get rid of the dot-loading-stage, but should keep the chart-container

server/lib/nl/scraper.py
chejennifer: Yes, it will add max 0.5s of delay because we only sleep if svgs are not all done drawing. I can try doing the second one where we check every svg is drawn

server/lib/nl/scraper.py
chejennifer: by delay, do you mean extra waiting time even though everything is already drawn? In that case, the max delay would be 0.5s

server/lib/nl/scraper.py
chejennifer: updated

server/lib/nl/scraper.py
pradh: I mean e2e time for all charts to render.  5s?

server/lib/nl/scraper.py
```diff
@@ -205,19 +214,41 @@ def _to_svg(svg):
...
   driver.get(_URL + query)
 
-  # Wait until the test_class_name has loaded.
+  # Wait until the page has loaded.
   wait = WebDriverWait(driver, 30)
   wait.until(EC.presence_of_element_located((By.CLASS_NAME, 'chart-container')))
-  wait.until(
-      EC.invisibility_of_element_located((By.CLASS_NAME, 'dot-loading-stage')))
-  time.sleep(5)
+  wait.until(lambda d: _check_svgs_drawn(driver))
```
pradh, line 251: Nicer!

Maybe pass a 0.25 poll frequency to WebDriverWait(driver=driver, timeout=30, poll_frequency=0.25) ?

https://github.com/SeleniumHQ/selenium/blob/635f86c4c271b686285a6837a7234184af7982e3/py/selenium/webdriver/support/wait.py#LL34C36-L34C36

pradh (approved):
Thanks!

server/lib/nl/scraper.py
chejennifer: e2e time wasn't affected by this _MAX_SVG_WAIT_TRIES constant .. but either way, the constant is gone now & e2e time for all charts to render depends on the query but most ended up taking around 5s.

server/lib/nl/scraper.py
chejennifer: done

chejennifer:
thanks for the review!

PR #2672, opened 2023-05-09
[V2 API] Consolidate place variables (count) fetch

shifucun:


server/integration_tests/test_data/demo_feb2023/query_9/chart_config.json
```diff
@@ -12,10 +12,10 @@
                       "highlightTopRight": true
                     },
                     "statVarKey": [
-                      "Median_Income_Person_scatter",
+                      "Mean_Income_Household_scatter",
```
shifucun, line 15: @pradh It's mysterious this golden file changed. Can you see why?

server/lib/fetch.py
```diff
@@ -258,6 +258,21 @@ def observation_existence(variables, entities):
   return result
 
 
+def entity_variables(entities):
+  """Gets the statistical variables that have observations for given entities.
+
```
chejennifer, line 263: Maybe include in this comment the structure of the return type. If I understand correctly, it is a dict where key is variable dcid and value is a dict where key is entity dcid and value is a list of observations?

server/routes/shared_api/place.py
```diff
@@ -233,22 +233,44 @@ def get_named_typed_place():
...
+  if not dcids:
+    dcids = request.json['dcids']
+  resp = fetch.entity_variables(dcids)
   result = []
-  entities = sorted(request.json.get('dcids', []))
-  resp = dc.entity_variables(entities)
-  for var, entity_obs in resp['byVariable'].items():
-    if len(entity_obs.get('byEntity'), {}) == len(entities):
+  for var, entity_obs in resp.items():
+    if len(entity_obs) == len(dcids):
```
chejennifer, line 249: sorry I didn't catch this in the earlier PR, but for stat var union, the stat var just has to exist for at least one of the dcids: https://github.com/datacommonsorg/mixer/blob/master/internal/server/statvar/place_stat_vars.go#L146

chejennifer (commented):
nice!

server/lib/fetch.py
shifucun: Added comment. The inner level value is an empty struct. This is due to the unified observation API response.

server/routes/shared_api/place.py
shifucun: Updated.

shifucun:
Thanks for review!

PR #2676, opened 2023-05-12
[tiles] various bug fixes

chejennifer:
- Add back state code for counties in the US for ranking tiles
- Fix getting csv data for ranking charts with multiple columns
- Use library to properly format csv string

https://screenshot.googleplex.com/B3aNNo8wbzfAeWF


static/js/utils/tests/chart_csv_utils.test.ts
```diff
@@ -310,17 +312,46 @@ test("rankingPointsToCsv", () => {
...
         "2,placeBId,2",
         "3,placeCName,3",
       ].join("\n"),
     },
+    {
+      name: "multi column",
+      rankingPoints: [
+        {
+          placeDcid: "placeAId",
+          placeName: "placeAName",
```
pradh, line 328: Maybe add some space and `,` to the name, and we should expect the cell to be wrapped in double quotes?

chejennifer: done

pradh (approved):

PR #2677, opened 2023-05-12
Only fetch place page ranking data when needed

shifucun:
Currently the ranking data are fetched in a batch fashion for ALL ranking stat vars and it's done on the server side. This has a bad effect of:

* Blocking page load
* Fetching a huge amount of data for some type of places, causing OOM issue in mixer.

This PR fetches the ranking data for each chart on-demand, much like how topic tile works.


static/js/place/chart.tsx
```diff
@@ -652,33 +649,54 @@ class Chart extends React.Component<ChartPropType, ChartStateType> {
...
-              }
-              for (const data of svData.data) {
-                data.value = data.value * scaling;
-              }
-              svData.rankingData = this.getRankingChartData(svData);
-              this.setState({ rankingChartDataGroup: svData });
-            })
-            .catch(() => {
+        // Fields like unit, denom, scaling and etc. are not used for data
+        // fetch, hencing setting a dummy value here.
```
chejennifer, line 653: nit: hence

static/js/place/chart.tsx
```diff
@@ -652,33 +649,54 @@ class Chart extends React.Component<ChartPropType, ChartStateType> {
...
+              log: false,
+              scaling: 1,
+              statVar: x,
+              unit: "",
+            };
+          }),
+          title: "",
+        })
+          .then((rankingChartData) => {
+            if (_.isEmpty(this.props.statsVars)) {
```
chejennifer, line 676: we don't even need to fetch data if this.props.statsVars is empty right? so could move this before the fetchData

chejennifer (approved):
nice!

static/js/place/chart.tsx
shifucun: Done

static/js/place/chart.tsx
shifucun: Done

PR #2680, opened 2023-05-13
[Place Page] Choropleth data fetch on a per chart basis

shifucun:
Right now the data fetch is for all the stat vars. This should be changed to per chart basis. Switching to map tile seems too much work given the svg difference and data format difference.

### DIFF
--- a/static/js/place/place.ts
+++ b/static/js/place/place.ts
@@ -19,7 +19,12 @@ import _ from "lodash";
 import React from "react";
 import ReactDOM from "react-dom";
 
-import { CachedChoroplethData, GeoJsonData, PageData } from "../chart/types";
+import {
+  ChartBlockData,
+  ChoroplethDataGroup,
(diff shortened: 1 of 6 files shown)


static/js/place/place.ts
```diff
@@ -109,36 +114,31 @@ function adjustMenuPosition(): void {
 /**
  * Get the geo json info for choropleth charts.
  */
-async function getGeoJsonData(
+export async function getGeoJsonData(
```
chejennifer, line 117: for cleanup, probably want to put these exported functions in a utils/fetcher file .. since they're not used by place page initial load

chejennifer (approved):
nice!

static/js/place/place.ts
shifucun: Good idea. Created fetch.ts to hold them

shifucun:
Thanks for review!

PR #2681, opened 2023-05-13
Make url args camel case

shifucun:
Url args should be camel case as we are doing in most cases. This PR cleans up some outliers.

Note there is not change for the json files except for changing current_geo to currentGeo. These files got formatted (now they are easier to read)


chejennifer (approved):
thanks for the cleanup!

PR #2684, opened 2023-05-15
[tiles] add merged svg in data attribute for map tile

chejennifer:
- add the merged (map + legend) svg to the data attribute of map tile if the map is a data tile

Screenshot 2023-05-15 at 1 53 34 PM


pradh (commented):
Thanks for the change!

pradh (approved):

PR #2688, opened 2023-05-16
A set of clean up for javascript build and bundling

shifucun:
* Move production npm packages under "dependencies" and only install them in the docker image (prod release).
* Exclude tests from the build bundles. This reduces the bundle size substantially.


static/webpack.config.js
```diff
@@ -118,13 +116,6 @@ const config = {
   },
   module: {
     rules: [
-      {
-        test: /\.(js|jsx|ts|tsx)$/,
-        exclude: /node_modules/,
```
chejennifer, line 123: do we not need this rule anymore?

chejennifer (approved):
thanks for the clean up!

static/webpack.config.js
shifucun: nope, looks like babel is not needed anymore

PR #2690, opened 2023-05-16
enable disaster dashboard for prod and staging

chejennifer:
- enable disaster dashboard and event page (because disaster dashboard links to event pages)
- add disaster dashboard to explore dropdown
- add webdriver test for event page

### DIFF
--- a/server/__init__.py
+++ b/server/__init__.py
@@ -98,19 +98,25 @@ def register_routes_custom_dc(app):
   pass
 
 
-def register_routes_disasters(app):
-  # Install blueprints specific to Stanford DC
+def register_routes_disasters(app, include_sustainability):
+  # Install blueprints specific to disasters
   from server.routes.disaster import html as disaster_html
   app.register_blueprint(disaster_html.bp)
 
   from server.routes.event import html as event_html
   app.register_blueprint(event_html.bp)
 
-  from server.routes.sustainability import html as sustainability_html
-  app.register_blueprint(sustainability_html.bp)
-
   from server.routes.disaster import api as disaster_api
   app.register_blueprint(disaster_api.bp)
+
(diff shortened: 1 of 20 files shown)


server/app_env/production.py
```diff
@@ -16,4 +16,5 @@
 
 
 class Config(_base.Config):
-  GA_ACCOUNT = 'UA-117119267-1'
\ No newline at end of file
+  GA_ACCOUNT = 'UA-117119267-1'
+  SHOW_DISASTER = True
```
shifucun, line 20: Do we want to make it "True" by default?

server/__init__.py
```diff
@@ -98,30 +98,33 @@ def register_routes_custom_dc(app):
   pass
 
 
-def register_routes_disasters(app):
-  # Install blueprints specific to Stanford DC
+def register_routes_disasters(app, include_sustainability):
```
shifucun, line 101: why have "include_sustainability" here? I feel register_xxx() should be a simple function. All the condition check should be outside and, if needed, have another register_xxx() to hold other related logic groups.

shifucun: if sustainability is different from disaster. then can use SHOW_SUSTAINABILITY flag in the env py files. We should try to remove the if/else checks here as much as possible..

server/app_env/production.py
chejennifer: done

shifucun (approved):
Is "ENABLE_DISASTER_JSON" still needed in local? If not, maybe remove it.

chejennifer:
> Is "ENABLE_DISASTER_JSON" still needed in local? If not, maybe remove it.

we still have the ability to use the disaster json for the disaster dashboard, so we should probably keep that flag until we decide to remove that option.. And probably want to do in separate PR

PR #2705, opened 2023-05-17
NL-SVG: Add a minor loadtest and fix a couple of bugs

pradh:


tools/nl/loadtest/loadtest_results.csv
```diff
@@ -0,0 +1,43 @@
+Result,Query,NumCharts,ChartTypes,Details
```
shifucun, line 1: Should save this as a json (it does not feel like a csv...)

shifucun (approved):
nice!

tools/nl/loadtest/loadtest.py
```diff
@@ -0,0 +1,73 @@
...
+              'SUCCESS', query,
+              str(len(data['charts'])), chart_types,
+              json.dumps(data['debug']['timing'])
+          ])
+        else:
+          print("Request failed with status code:", response.status_code)
+          writer.writerow(
+              ['FAILURE', query, "0", "",
+               str(response.status_code)])
+  print('')
```
chejennifer, line 62: why print the extra empty strings?

tools/nl/loadtest/loadtest_results.csv
pradh: Ok, let me open that up...

PR #2707, opened 2023-05-17
Drop old embeddings and rename python file

pradh:


tools/nl/embeddings/build_embeddings.py
```diff
@@ -11,11 +11,17 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+"""Build the embeddings index by concatenating various inputs."""
+
+# TODO: This file should be renamed build_embeddings.py once it is ready to
+# replace the existing build_embeddings.py file.
+# TODO: Consider adding the model name to the embeddings file for downstream
+# validation.
```
jehangiramjad, line 20: This comment should now become obsolete, right?

jehangiramjad (approved):
Thanks!

tools/nl/embeddings/build_embeddings.py
pradh: Fixed it in follow up PR.

PR #2710, opened 2023-05-17
[nl] update determining if sv is per capita

chejennifer:
- some stat var names no longer contain key words that indicate per capita, so should also check the dcid to tell if a stat var is per capita or not. 

e.g.,
sv dcid: Percent_Person_WithHighBloodPressure
sv name: High Blood Pressure


pradh (approved):
Thanks for the fix!

PR #2714, opened 2023-05-18
Guard disaster/sustainability dashboard

shifucun:


server/__init__.py
```diff
@@ -253,7 +253,7 @@ def create_app():
     register_routes_custom_dc(app)
 
   register_routes_base_dc(app)
-  if cfg.SHOW_DISASTER or os.environ.get('ENABLE_MODEL') == 'true':
+  if cfg.SHOW_DISASTER:
```
chejennifer, line 256: could there be a case where SHOW_DISASTER is false even though ENABLE_MODEL is true? because we need disaster routes to be registered for nl interface

chejennifer (commented):
thanks!

pradh (approved):

server/__init__.py
```diff
@@ -253,7 +253,7 @@ def create_app():
...
     register_routes_disasters(app)
```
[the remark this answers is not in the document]
chejennifer: No, I believe NL interface still needs disaster routes to be registered

server/__init__.py
shifucun: I wonder which instance have that requirement? And for such cases, we should set SHOW_DISASTER=true explicitly?

pradh:
We have NL enabled on: autopush, dev, ME, ClimateTrace (maybe this one we
don’t want disaster)… and might want on Stanford too at some point?

I’m okay with any change that won’t stop NL from working on these instances
:)

server/__init__.py
shifucun: Revert this but we definitely need to separate the two configs

chejennifer (approved):
thanks!

server/__init__.py
shifucun: actually moved the disaster api to base route now and remove the ENABLE_MODEL check

shifucun:
> We have NL enabled on: autopush, dev, ME, ClimateTrace (maybe this one we don’t want disaster)… and might want on Stanford too at some point? I’m okay with any change that won’t stop NL from working on these instances :)

I moved the api routes to base, so all instance should work.

chejennifer (approved):
thanks!

PR #2715, opened 2023-05-18
Use bash script to clear redis cache

shifucun:
The old requires cloud functions and vpc connection to the redis instance, which has issues.

The new approach does everything with gcloud and kubectl commands.


chejennifer (approved):
awesome, thanks for updating this!

PR #2721, opened 2023-05-19
NL: Add script to generate a CSV with name+description

pradh:
This PR takes code originally from @jehangiramjad and makes some minor tweaks.

It copies a saved version of DDD titles in GCS.


jehangiramjad (approved):
Thanks! Only nitty comment is that the comments in the script refer to '14k' and '101k' SVs and that will be updated whenever the numbers differ from those. We might want to use other generic name but that's something we can do later

tools/nl/embeddings/prep/gen_svs.py
```diff
@@ -0,0 +1,130 @@
...
+  # Get all the existing SVs!
+  existing_dfs = []
+  for svf in sorted(glob.glob(FLAGS.existing_svs_filepattern)):
+    # Get the 1300 curated svs (Demo_US_SVs)
+    existing_dfs.append(pd.read_csv(svf).fillna(""))
+  existing_df = pd.concat(existing_dfs)
+
+  print(len(input_df))
+  print(len(ddd_df))
+  print(len(existing_df))
```
shifucun, line 85: remove or make this full print sentence?

tools/nl/embeddings/prep/run.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+# Copyright 2020 Google LLC
```
shifucun, line 2: 2023

shifucun (approved):

tools/nl/embeddings/prep/gen_svs.py
pradh: Removed

pradh:
> Thanks! Only nitty comment is that the comments in the script refer to '14k' and '101k' SVs and that will be updated whenever the numbers differ from those. We might want to use other generic name but that's something we can do later

Done.

PR #2729, opened 2023-05-23
Avoid mutating args in V2 obs API

pradh:
This fixes regression introduced in [this PR](https://github.com/datacommonsorg/website/pull/2672/files#diff-be4c9d19b450d78184e8477cd2184bbf978a95d94cf68c359cd08deb7940ff63), wherein the ordered variables passed into existence check API gets mutated (sorted), and that affects subsequent caller code.

This impacted correlation queries and dual-variable-filter queries.  Thus the diffs in correlation related queries with many SVs.


server/services/datacommons.py
```diff
@@ -172,24 +172,21 @@ def obs_series_within(parent_entity, child_type, variables):
       })
 
 
-def v2observation(select, entity, variable):
```
shifucun, line 175: There are use cases to pass in the expression. All the functions in datacommons.py should be in the exact form of V2 API arguments. Wrapping can be done in fetch.py.

server/services/datacommons.py
```diff
@@ -172,24 +172,21 @@ def obs_series_within(parent_entity, child_type, variables):
...
   Args:
     select: A list of select props.
-    entity: A dict in the form of {'dcids':, 'expression':}
-    variable: A dict in the form of {'dcids':, 'expression':}
+    entities: A list of entities.
+    variables: A list of variables.
 
   """
-  if 'dcids' in entity:
-    entity['dcids'].sort()
```
shifucun, line 184: if mutating is an issue, can do entity['dcids'] = sorted(entity['dcids'])

pradh: that still mutates entity?  see if this works.

server/services/datacommons.py
pradh: I see, reverted.

pradh (commented):
Thanks.

server/services/datacommons.py
```diff
@@ -177,19 +177,16 @@ def v2observation(select, entity, variable):
...
     select: A list of select props.
     entity: A dict in the form of {'dcids':, 'expression':}
-    variable: A dict in the form of {'dcids':, 'expression':}
+    variable: A dict in the form of {'dcids':, 'expression':}    
 
   """
+  req = {'select': select}
   if 'dcids' in entity:
-    entity['dcids'].sort()
+    req['entity'] = {'dcids': sorted(entity['dcids'])}
```
shifucun, line 185: 'entity' would not be set if it is {'expression': xxx}. Can just do:

entity['dcids'] = sorted(entity['dcids'])

?

pradh: good point.

shifucun (approved):
nice! Glad you figure out the flakiness.

PR #2732, opened 2023-05-23
remove earth column from rankings

juliawu:
This PR is a quick fix for the incorrect "X of Y in Earth" rankings we currently show in place overview tiles (the ranking numbers are just a copy of the US rankings). The fix removes the Earth column in the ranking table. We should remove these changes once the Earth rankings are fixed.

Before:
Screenshot 2023-05-23 at 8 40 59 AM

After:
Screenshot 2023-05-23 at 8 41 13 AM


juliawu:
Already fixed in https://github.com/datacommonsorg/website/pull/2731! Closing this PR.

PR #2733, opened 2023-05-23
Check latest bigtable version is not empty for autopush deployment

shifucun:
The table could be empty for newly created import group.


scripts/deploy_latest.sh
```diff
@@ -46,7 +46,9 @@ yq eval -i '.tables = []' deploy/storage/base_bigtable_info.yaml
 for src in $(gsutil ls gs://datcom-control/autopush/*_latest_base_cache_version.txt); do
   echo "Copying $src"
   export TABLE="$(gsutil cat "$src")"
-  yq eval -i '.tables += [env(TABLE)]' deploy/storage/base_bigtable_info.yaml
+  if [[ $TABLE != '' ]]; then
```
pradh, line 49: "" ? for bash

shifucun:
thanks!

PR #2734, opened 2023-05-23
Use line tile in place page

shifucun:


static/js/place/chart.tsx
```diff
@@ -556,44 +602,10 @@ class Chart extends React.Component<ChartPropType, ChartStateType> {
 
   private processData(): void {
     const dataGroups: DataGroup[] = [];
-    const allDates = new Set<string>();
     // TODO(datcom): handle i18n for scaled numbers
     const scaling = this.props.scaling ? this.props.scaling : 1;
     const sv = !_.isEmpty(this.props.statsVars) ? this.props.statsVars[0] : "";
     switch (this.props.chartType) {
-      case chartTypeEnum.LINE:
```
chejennifer, line 564: I think you also need to update the drawChart function & componentDidUpdate?

chejennifer (commented):
this is cool!

shifucun (commented):
Thanks for review!

static/js/components/tiles/line_tile.tsx
```diff
@@ -152,8 +169,7 @@ export function draw(
     chartData.unit
   );
   if (!isCompleteLine) {
-    svgContainer.querySelectorAll(".dotted-warning")[0].className +=
-      " d-inline";
+    elem.querySelectorAll(".dotted-warning")[0].className += " d-inline";
```
chejennifer, line 172: maybe should check that querySelectorAll returns a non-empty list before accessing the first element? & maybe should also iterate through the list and add the class name to all selected elements?

static/js/components/tiles/line_tile.tsx
```diff
@@ -51,6 +52,14 @@ interface LineTilePropType {
...
   // Whether or not to render the data version of this tile
   isDataTile?: boolean;
+  // A map keyed by stat var dcid, with value being the link for that stat var.
+  statVarLink?: Record<string, string>;
+  // The explore more component
+  exploreJsx?: JSX.Element;
+  // Optional callback function when source link is clicked.
+  clickFn?: ClickFnType;
+  // Optional callback function to hide the tile when it only has single dot.
+  hideTileFn?: () => void;
```
chejennifer, line 62: nit: if I'm understanding this correctly, it looks like this callback is really a callback to handle when there is a single dot. Place page handles this by hiding the tile, but other callers of this tile could technically handle it differently. Maybe update the name of the prop and the description?

static/js/components/tiles/chart_footer.tsx
```diff
@@ -55,32 +56,53 @@ function getSourcesJsx(sources: Set<string>): JSX.Element[] {
...
   });
   return sourcesJsx;
 }
 
+interface ChartFooterPropType {
+  // set of full urls of sources of the data in the chart
+  sources: Set<string>;
+  handleEmbed?: () => void;
+  exploreJsx?: JSX.Element;
+  // Optional callback function when source link is clicked.
```
chejennifer, line 74: not just for source link right? also used for export?

static/js/components/tiles/chart_footer.tsx
```diff
@@ -25,20 +25,21 @@ import { NL_SOURCE_REPLACEMENTS } from "../../constants/app/nl_interface_constan
 import { urlToDomain } from "../../shared/util";
 import { isNlInterface } from "../../utils/nl_interface_utils";
 
-interface ChartFooterPropType {
-  // set of full urls of sources of the data in the chart
-  sources: Set<string>;
-  handleEmbed?: () => void;
+export interface ClickFnType {
```
chejennifer, line 28: nit: since this is an exported interface, can we make the name more informative. Maybe something about ChartFooter like ChartFooterClickFns or something of that sort.

Also add a comment describing this type?

chejennifer (commented):
thanks for the updates!

PR #2736, opened 2023-05-24
NL: Improve performance of `variable in place` queries

pradh:
## Context
The `variable in place` queries today can take 10+ seconds for a few reasons:
1. The `extend_svs` call is very slow (does multiple serial calls).
2. For every chart, we perform a `has_series_with_single_datapoint` to promote SVs with a single datapoint from timeline charts to bar charts.
3. We have no limits on the number of charts we return.

Also, we sometimes duplicate charts across SV in topic and SV in index result.

## Fixes

This PR includes the following fixes:
1. Limit the number of charts we attempt to add to 15 (assuming 3 charts per row, 5 rows might be sufficient)
2. Get rid of the `has_series_with_single_datapoint` call by getting the "length of time-series" as part of existence check itself.  But don't always do this, only for SIMPLE query (since I see series_core is slower than existence_check).
3. Instead of doing SV expansion always, we only do it after processing the primary SVs, and then if we've not exceeded the max chart limit, and that too for not more than 5 SVs.
4. Optimize `extend_svs` slightly to avoid redundant API calls (e.g., if we expand SV1 + SV2, and they are in the expansion list of each other, we will do 2 calls).
5. Fixes duplicate charts (changes in `ExistenceCheckStateTracker.__init__`).

Owing to (3+4), we slightly change the layout of charts returned for SIMPLE query from:

```
SV1
peers-of-SV1
SV2
peers-of-SV2
```

to 

```
SV1
SV2
peers-of-SV1
peers-of-SV2
```

As a result, fix the peer chart titles to include SV name (previously they just had a generic "Compared with other variables").

## Results

The "Before" numbers were on autopush, and "After" on local.

### [Income in california]

**Before: 12s**

```
  "TIMING": {
    "build_page_config": 0.3,
    "extend_svs": 4.29,
    "fulfillment": 11.98,
    "get_sv_details": 0.3,
    "has_series_with_single_datapoint": 7.290000000000003,
    "query_detection": 0.39,
    "sv_existence_for_places": 0.27,
    "topic_calls": 0.11
  }
```

**After: 0.7s**

```
  "TIMING": {
    "build_page_config": 0.31,
    "fulfillment": 0.68,
    "get_sv_details": 0.3,
    "query_detection": 0.71,
    "sv_existence_for_places_check_single_point": 0.42,
    "topic_calls": 0.25
  }
```

### [Male population in california]

**Before: 15s**

```
  "TIMING": {
    "build_page_config": 0.57,
    "extend_svs": 10.43,
    "fulfillment": 15.04,
    "get_sv_details": 0.56,
    "has_series_with_single_datapoint": 4.350000000000001,
    "query_detection": 0.75,
    "sv_existence_for_places": 0.25
  }
```

**After: 0.3s**

```
  "TIMING": {
    "build_page_config": 0.36,
    "fulfillment": 0.3,
    "get_sv_details": 0.36,
    "query_detection": 0.26,
    "sv_existence_for_places_check_single_point": 0.29
  }
```


server/lib/nl/fulfillment/base.py
```diff
@@ -36,10 +36,14 @@
 from server.lib.nl.utterance import QueryType
 from server.lib.nl.utterance import Utterance
 
-# TODO: Factor classification processing functions into a common place.
-
 _EVENT_PREFIX = 'event/'
 
+# Limit the number of charts.  With 3 per row max, allow up to 5 rows.
+_MAX_NUM_CHARTS = 15
```
shifucun, line 42: are 3 charts per row fixed? Could this change depending page layout?

server/lib/nl/fulfillment/base.py
```diff
@@ -259,73 +270,91 @@ def _add_charts_with_place_fallback(state: PopulateState, places: List[Place],
...
-            state.uttr.counters.err(
-                'failed_event_existence_check', {
-                    'places': places[:constants.DBG_LIST_LIMIT],
-                    'event': chart_vars.event
-                })
-        else:
+                                          extended_vars=None)
+
+      if not sv2extensions:
+        chart_vars_list = _build_chart_vars(state, sv, rank)
```
shifucun, line 310: (maybe for later) have more functions to wrap the logic in this file, or split this into multiple classes, files. As someone new to the code, it's a bit hard to follow the logic now..

shifucun (approved):

server/lib/nl/fulfillment/base.py
pradh: Oh this is just the rationale for a rough limit.

server/lib/nl/fulfillment/base.py
```diff
@@ -36,10 +36,14 @@
...
 from server.lib.nl.utterance import Utterance
 
-# TODO: Factor classification processing functions into a common place.
-
 _EVENT_PREFIX = 'event/'
 
+# Limit the number of charts.  With 3 per row max, allow up to 5 rows.
+_MAX_NUM_CHARTS = 15
+
+# Do do extension API calls for more than these many SVs
```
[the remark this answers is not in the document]
pradh: Fixed.

server/lib/nl/fulfillment/base.py
pradh: Splitting into multiple classes/files, SG.  Added TODO

pradh (commented):
Thanks!

PR #2737, opened 2023-05-24
[Maps] Add support for zip codes

juliawu:
This PR:
* Adds support for selecting zip codes (ZCTAs) as a place type in the map tool
* Allows for map tiles to use ZCTAs as an enclosed place type
* Allows for map tiles to use cities as an enclosed place type.

Sample map tool screenshots:
Screenshot 2023-05-24 at 10 30 20 AM
Screenshot 2023-05-24 at 10 30 51 AM

Sample map tiles:
Screenshot 2023-05-24 at 11 11 01 AM

Screenshot 2023-05-24 at 11 12 00 AM

Note that ZCTAs and cities currently render poorly -- borders and some rescaling would improve readability. These improvements will come in future PRs.

Testing:
Confirmed these changes don't break the NL interface, place pages, disaster dashboard, and sustainability explorer.

### DIFF
--- a/server/routes/shared_api/choropleth.py
+++ b/server/routes/shared_api/choropleth.py
@@ -72,6 +72,8 @@ CHOROPLETH_GEOJSON_PROPERTY_MAP = {
     "EurostatNUTS2": "geoJsonCoordinatesDP2",
     "EurostatNUTS3": "geoJsonCoordinatesDP1",
     "IPCCPlace_50": "geoJsonCoordinates",
+    "City": "geoJsonCoordinates",
+    "CensusBlockGroup": "geoJsonCoordinates",
     "CensusTract": "geoJsonCoordinates",
     "CensusZipCodeTabulationArea": "geoJsonCoordinates",
 }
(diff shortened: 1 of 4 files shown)


juliawu:
Tagging @pradh for visibility.

chejennifer (approved):
nice

server/routes/shared_api/choropleth.py
```diff
@@ -72,6 +72,8 @@
     "EurostatNUTS2": "geoJsonCoordinatesDP2",
     "EurostatNUTS3": "geoJsonCoordinatesDP1",
     "IPCCPlace_50": "geoJsonCoordinates",
+    "City": "geoJsonCoordinates",
```
pradh, line 75: Are we sure we want this not DP1/DP2?

pradh (approved):
Nice!

server/routes/shared_api/choropleth.py
juliawu: Based on a quick BQ search, it looks like there are some cities that don't have a DP geojsons available. I can add logic to select D1/DP2 if available in a future PR?

pradh: We might need to compute it then.

dwnoble (approved):
Thanks Julia!

One thing to look at in the future: the border is too large when hovering over small geos, and the area gets obstructed:

See: fl-sm-geo

Vs: fl-med-geo

juliawu:
> One thing to look at in the future: the border is too large when hovering over small geos, and the area gets obstructed:

Thanks for flagging! I'll merge this for now, but fix the border sizing issues in a followup PR.

shifucun (approved):

PR #2739, opened 2023-05-24
[library] Added line chart library export

dwnoble:
Screenshot:

dc-standalone-line-chart

### DIFF
--- a/experimental/index.html
+++ b/experimental/index.html
@@ -7,6 +7,12 @@
       .flex-child {
         flex: 1; /* This makes each child div take up an equal amount of space */
       }
+      .flex-column {
+        flex-direction: column;
+      }
+      .flex-row {
+        flex-direction: row;
+      }
       #my-component-1 {
         background-color: #ddd;
       }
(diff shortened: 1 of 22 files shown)


experimental/index.html
```diff
@@ -41,17 +47,74 @@
         };
         datacommons.drawRanking(element1, props1);
         datacommons.drawRanking(element2, props2);
+
+        var lineChartElement1 = document.getElementById("my-line-chart-1");
+        var lineChartElement2 = document.getElementById("my-line-chart-2");
+        var lineChartProps1 = {
+          id: "my-line-chart-content-1",
```
shifucun, line 54: I thought props.id should be the same as the html element id, but looks like it works here.

shifucun (approved):
Very nice!

experimental/index.html
dwnoble: It looks like this `id` gets applied to the component's inner div that contains the chart svg 

Screenshot 2023-05-25 at 1 08 17 PM

PR #2741, opened 2023-05-25
NL: Add support to build "medium" index

pradh:
Also, instead of checking there are no dup names, resolve the duplicate and log those separately, so we don't block on fixing those for building index.

Follow up PR will add the small and medium index to NL server.


tools/nl/embeddings/README.md
```diff
@@ -4,36 +4,46 @@ This directory contains the data sheets (containing StatVar DCID and
...
-* `Override_Alternatives`: if this column has any value in a row, all other columns are ignored. Multiple strings are acceptable for this column. If multiple strings are provided, they must be semi-colon delimited.
-* `Curated_Alternatives`: a semi-colon delimited list of strings which can serve as alternative ways of referring to the StatVar.
+   * `dcid`: the StatVar DCID.
+   * `Name`: the name of the StatVar
+   * `Description`: the description(s) of the StatVar. Multiple descriptions are acceptable. If multiple description strings are provided, they must be semi-colon delimited. This column can be expected to be used as part of any automated alternative generation processed, e.g. using an LLM.
+   * `Override_Alternatives`: if this column has any value in a row, all other columns are ignored. Multiple strings are acceptable for this column. If multiple strings are provided, they must be semi-colon delimited.
+   * `Curated_Alternatives`: a semi-colon delimited list of strings which can serve as alternative ways of referring to the StatVar.
+
+1. Add any auto-generated input CSVs to `data/autogen_input/<index-size>/`.
```
jehangiramjad, line 29: It took me a while to figure out that this is the file (or process) through which the small vs medium are distinguished. If, theoretically, the same file was provided under `data/autogen_input/small/` and `data/autogen_input/medium/` then the embeddings would be the same size, right? So we should make it clear in this README.md that this is the file/process through which the number/size of the SVs/Embeddings is primarily impacted. The rest of the inputs are all common across the embeddings sizes.

jehangiramjad (approved):
Thanks! Just one comment on the README

tools/nl/embeddings/README.md
pradh: thanks, clarified.

PR #2743, opened 2023-05-26
A bunch of change for custom DC using remote mixer

shifucun:


server/templates/version.html
```diff
@@ -39,6 +39,12 @@
             <a href="https://github.com/datacommonsorg/mixer/commit/{{mixer_hash}}">{{mixer_hash}}</a>
           </td>
         </tr>
+        <tr>
```
dwnoble, line 42: Will `remote_mixer_domain` always be set? should we hide this block if it's not set?

pradh (approved):

server/templates/version.html
shifucun: Good point! This will be empty if not set. I feel this maybe good to explicitly showing that the remote mixer is not used.

shifucun:
Thanks for review!

PR #2749, opened 2023-05-27
Migrate all deployment config from kustomize to helm

shifucun:
This prepares for deployment change of remote mixer in all custom DC


server/app_env/stanford_staging.py
```diff
@@ -1,22 +0,0 @@
...
-# Unless required by applicable law or agreed to in writing, software
-# distributed under the License is distributed on an "AS IS" BASIS,
-# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-# See the License for the specific language governing permissions and
-# limitations under the License.
-
-from server.app_env import _base
-
-
-class Config(_base.Config):
```
chejennifer, line 18: curious why we're deleting this, because won't this code look for this file? https://github.com/datacommonsorg/website/blob/master/server/lib/config.py#L24

chejennifer (commented):
thanks for migrating!

docs/new_deployment.md
```diff
@@ -1,120 +0,0 @@
-# Add a New Deployment Instance
-
-This doc details the steps involved to bring up a new Data Commons instance
```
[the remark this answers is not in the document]
shifucun: Yes, new instance should use the one time setup script from Alex and we would need more docs for that.

server/app_env/stanford_staging.py
shifucun: In stanford_staging.yaml, the flaskEnv is now "stanford" because it should use the same html template folder (now it can not find /stanford_staging/..., thus uses the default custom html template).

chejennifer (approved):

PR #2751, opened 2023-05-30
[google map] handle when geojson is not a properly formatted json string

chejennifer:
- instead of having the page break due to improperly formatted json string, just hide the map

autopush fire event page which has an improperly formatted json string as the geojson string:
Screenshot 2023-05-30 at 11 39 50 AM

with the fix:
Screenshot 2023-05-30 at 11 39 33 AM


shifucun (approved):
thanks for the fix!

juliawu (approved):
Thanks for the fix!

PR #2752, opened 2023-05-31
NL: Add the medium index to nl_server enabled by idx= param

pradh:
This change introduces the `medium` index added in https://github.com/datacommonsorg/website/pull/2741 to the nl_server. The current index is tagged as `small`.

The index is selected via a url param `idx=small` or `idx=medium`.

Add an integration test for `medium` (the number of vehicles per family SVs are not in `small`).

NOTE: The `medium` index is 140MB (added to nl_server memory), relative to `small` which is 40MB.


jehangiramjad (approved):
Thanks! 

One general question: we'd be loading both the small and medium embeddings in memory at the same time at startup, right? Is that the model going forward or do we want to choose one or the other at server start time so that the memory footprint is only as needed?

pradh:
Thanks for the review!

shifucun (approved):
very neat!

PR #2753, opened 2023-05-31
[map] draw map points without data values

chejennifer:
- allow drawing map points even if they don't have associated data values

e.g., map point place type is SuperfundSite, no map point sv, and there is no data on median income for superfund sites
Screenshot 2023-05-30 at 10 00 07 PM

### DIFF
--- a/static/js/tools/map/d3_map.tsx
+++ b/static/js/tools/map/d3_map.tsx
@@ -184,9 +184,14 @@ export function D3Map(props: D3MapProps): JSX.Element {
             ? statVar.value.info[statVar.value.mapPointSv].title
             : "";
       }
+      const filteredMapPoints = props.mapPointValues
+        ? props.mapPoints.filter((point) => {
+            return point.placeDcid in props.mapPointValues;
+          })
+        : props.mapPoints;
       addMapPoints(
         mapContainerRef.current,
-        props.mapPoints,
+        filteredMapPoints,
         props.mapPointValues || {},
         projection,
         props.mapPointValues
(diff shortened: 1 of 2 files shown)


static/js/tools/map/d3_map.tsx
```diff
@@ -176,23 +176,23 @@ export function D3Map(props: D3MapProps): JSX.Element {
...
       let mapPointSvTitle = "";
       if (statVar.value.mapPointSv !== statVar.value.dcid) {
         mapPointSvTitle =
           statVar.value.mapPointSv in statVar.value.info
             ? statVar.value.info[statVar.value.mapPointSv].title
             : "";
       }
-      const filteredMapPoints = props.mapPoints.filter((point) => {
-        return point.placeDcid in props.mapPointValues;
-      });
```
pradh, line 189: When mapPointValues exist (i.e., mapsv) should we still filter?

pradh (approved):
Thanks!

static/js/tools/map/d3_map.tsx
chejennifer: sure, updated!

shifucun (changes requested):
In general I feel there should be a finer control on what type of points to show here? I don't think we should show all points for unrelated stat var on a map.

static/js/tools/map/d3_map.tsx
chejennifer: chatted with Bo about this & we think for consistency, it'll make more sense to show all the points & just show "data unavailable" for points missing data. Also when mapPointValues exist, there aren't going to be a lot of points without data, so it doesn't look too different from before.

pradh: ok, sg

shifucun (approved):
thanks for the update!

PR #2762, opened 2023-06-01
[events] handle missing data

chejennifer:
- handle when server passes empty data fields 

on autopush:
Screenshot 2023-06-01 at 12 10 08 PM

with this fix:
Screenshot 2023-06-01 at 12 10 35 PM


pradh:
Thanks for the fix Jennifer!  Does this address the other empty event type page too (I forget floods was it)?

shifucun (approved):

PR #2765, opened 2023-06-02
[map] fix dot size

chejennifer:
- Fixed a bug where map points size kept changing. This was because we were trying to get the min region diagonal without actually calling Math.min
- Reverted the dot size logic to use minDotSize * 2 when there is no dot size scale & increased minimum dot size from 1.1 to 1.25
- Additional bug fixed: infinite loading when there are no map points. This is because of a check for non empty map points data even though there just may not be map points for the chosen place type

Screenshot 2023-06-02 at 11 02 17 AM

### DIFF
--- a/server/routes/shared_api/choropleth.py
+++ b/server/routes/shared_api/choropleth.py
@@ -450,7 +450,7 @@ def get_map_points():
   geos = []
   geos = fetch.descendent_places([place_dcid], place_type).get(place_dcid, [])
   if not geos:
-    return Response(json.dumps({}), 200, mimetype='application/json')
+    return Response(json.dumps([]), 200, mimetype='application/json')
   names_by_geo = place_api.get_display_name(geos)
   # For some places, lat long is attached to the place node, but for other
   # places, the lat long is attached to the location value of the place node.
--- a/static/js/chart/draw_d3_map.ts
+++ b/static/js/chart/draw_d3_map.ts
@@ -69,6 +69,7 @@ const MAP_PATH_LAYER_CLASS = "map-path-layer";
 const MAP_PATH_HIGHLIGHT_CLASS = "map-path-highlight";
 const MAP_PATH_STROKE_WIDTH = "1.5px";
 const MAP_PATH_OPACITY = "0.5";
+const DEFAULT_MIN_DOT_SIZE = 1.25;
 
 /**
  * From https://bl.ocks.org/HarryStevens/0e440b73fbd88df7c6538417481c9065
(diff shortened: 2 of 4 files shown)


dwnoble:
Thanks Jennifer!

Some comments that might be beyond the scope of this fix:
* If the url param mp=1 is set, but no ppt= value is set, the page hangs (the spinner doesn't go away)
* Should we scale the points when zooming in? right now the points get bigger as you zoom
dc-points-zoomed


(might be achievable by adding something like this in the zoom handler)
```
// change DEFAULT_MIN_DOT_SIZE to originally configured point size
map
  .selectAll("circle")
  .attr("r", (DEFAULT_MIN_DOT_SIZE * 2) / d3.event.transform.k);
```

* When ppt/mp are set, changing the location in the UI doesn't update the point layer
Visiting 
http://localhost:8080/tools/map#%26sv%3DCount_Person%26pc%3D0%26denom%3DCount_Person%26pd%3DgeoId%2F42%26ept%3DCounty%26ppt%3DEpaReportingFacility%26mp%3D1
shows:
dc-points-pa

But changing the location to United States shows:
dc-no-point-layer-refresh

Refreshing this page shows the entire point layer:
dc-usa-all-points

dwnoble (approved):

chejennifer:
> Thanks Jennifer!
> 
> Some comments that might be beyond the scope of this fix:
> 
> * If the url param mp=1 is set, but no ppt= value is set, the page hangs (the spinner doesn't go away)
> * Should we scale the points when zooming in? right now the points get bigger as you zoom
>   dc-points-zoomed
> 
> (might be achievable by adding something like this in the zoom handler)
> 
> ```
> // change DEFAULT_MIN_DOT_SIZE to originally configured point size
> map
>   .selectAll("circle")
>   .attr("r", (DEFAULT_MIN_DOT_SIZE * 2) / d3.event.transform.k);
> ```
> 
> * When ppt/mp are set, changing the location in the UI doesn't update the point layer
>   Visiting
>   http://localhost:8080/tools/map#%26sv%3DCount_Person%26pc%3D0%26denom%3DCount_Person%26pd%3DgeoId%2F42%26ept%3DCounty%26ppt%3DEpaReportingFacility%26mp%3D1
>   shows:
>   dc-points-pa
> 
> But changing the location to United States shows: dc-no-point-layer-refresh
> 
> Refreshing this page shows the entire point layer: dc-usa-all-points

Thanks for the review! I agree that these are beyond the scope of this PR, so I've opened up github issues for them instead: https://github.com/datacommonsorg/website/issues/2768, https://github.com/datacommonsorg/website/issues/2767, https://github.com/datacommonsorg/website/issues/2766

PR #2770, opened 2023-06-03
[Event Pages] Fix "observation date" value rendering.

juliawu:
Fix a bug where the "observation date" entry on the property table in the event pages has an incorrectly formatted value. 

Before:
Screenshot 2023-06-02 at 5 08 24 PM

After:
Screenshot 2023-06-02 at 5 07 56 PM


chejennifer (approved):
nice, thanks for the fix!

PR #2774, opened 2023-06-05
[Event Pages] Fix negative numbers bug

juliawu:
Fixes a bug where the negative sign in a property value shows up as a unit instead of being in front of the digits.

Before (see SPI value in table):
Screenshot 2023-06-05 at 10 28 23 AM

After:
Screenshot 2023-06-05 at 10 28 39 AM


chejennifer (approved):
thanks for the quick fix

PR #2775, opened 2023-06-05
[Disaster Dashboard] Increase default max severity filter for floods

juliawu:
Increases the default value for the severity filter's upper limit for floods from 1,000 sq km to 10,000 sq km. This change allows for the 2022 large (>1k sq km) floods in Pakistan to display using default settings.

Before:
Screenshot 2023-06-05 at 1 10 26 PM

After:
Screenshot 2023-06-05 at 1 09 10 PM


chejennifer (approved):
thanks!

PR #2776, opened 2023-06-05
[Disaster Dashboard] Fix zoom buttons on disaster specific maps

juliawu:
Fixes a bug where the (+)/(-) zoom buttons on the individual disaster maps don't do anything on click. This PR makes the ids of each button div unique, which fixes the issue.

https://github.com/datacommonsorg/website/assets/4034366/e9cc526f-29a8-4e98-b1e0-34a4a9244aea

### DIFF
--- a/static/js/components/tiles/disaster_event_map_tile.tsx
+++ b/static/js/components/tiles/disaster_event_map_tile.tsx
@@ -28,6 +28,7 @@ import {
   addPolygonLayer,
   drawD3Map,
   getProjection,
+  MapZoomParams,
 } from "../../chart/draw_d3_map";
 import {
   GeoJsonData,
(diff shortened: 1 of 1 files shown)


static/js/components/tiles/disaster_event_map_tile.tsx
```diff
@@ -67,6 +67,12 @@ const MAP_POINTS_MIN_RADIUS_EARTH = 0.8;
 // Set of dcids of places that should use the selected place as the base geojson
 const PLACE_MAP_PLACES = new Set(["country/AUS", "country/BRA"]);
 
+// Stores IDs for the zoom buttons on the map
+interface ZoomParams {
```
chejennifer, line 71: don't need to re-define this: https://github.com/datacommonsorg/website/blob/master/static/js/chart/draw_d3_map.ts#L46

chejennifer (approved):
thanks for the fix!

PR #2777, opened 2023-06-05
[disaster] remove NA from top event tiles

chejennifer:
Screenshot 2023-06-05 at 2 18 12 PM

### DIFF
--- a/static/js/components/tiles/top_event_tile.tsx
+++ b/static/js/components/tiles/top_event_tile.tsx
@@ -132,7 +132,13 @@ export const TopEventTile = memo(function TopEventTile(
                 <tr>
                   <td></td>
                   {showNameColumn && <td>Name</td>}
-                  {showPlaceColumn && <td>{props.enclosedPlaceType}</td>}
+                  {showPlaceColumn && (
+                    <td>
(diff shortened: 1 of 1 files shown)


static/js/components/tiles/top_event_tile.tsx
```diff
@@ -132,7 +132,13 @@ export const TopEventTile = memo(function TopEventTile(
                 <tr>
                   <td></td>
                   {showNameColumn && <td>Name</td>}
-                  {showPlaceColumn && <td>{props.enclosedPlaceType}</td>}
+                  {showPlaceColumn && (
+                    <td>
+                      {Object.keys(eventPlaces).length < topEvents.length
```
juliawu, line 137: For my own understanding, why compare number of event places with number of top events?

juliawu (approved):
Thanks for the changes!

static/js/components/tiles/top_event_tile.tsx
chejennifer: eventPlaces has a key value pair for every event that has a place found for it. Therefore, if there are less eventPlaces entries than number of events, that means at least one event is missing the place

chejennifer:
thanks for the review!

PR #2778, opened 2023-06-05
[Disaster Dashboard] Add units to info cards

juliawu:
This PR enables property value units to be displayed on the disaster dashboard.

More specifically, the following changes are made:
* Extract out the util functions in `event/app.tsx` to a dedicated utils file, that way these general functions can be reused.
* Edit `displayProps` interface to map to strings instead of numbers, to keep the unit stored.
* Remove stripping the unit in the main `disaster_event_map_utils`.
* Pass that unit to the info card so it can also be displayed.
* Update property value processing in `histogram_tile` and `top_event_tile` which expect numbers instead of a number with unit.
* Update `disaster_event_map_utils_test` to reflect changes

Before:
Screenshot 2023-06-05 at 4 16 33 PM

After:
Screenshot 2023-06-05 at 4 21 25 PM

Dashboard page itself remains unchanged:
Screenshot 2023-06-05 at 4 25 19 PM
Screenshot 2023-06-05 at 4 25 28 PM
Screenshot 2023-06-05 at 4 25 37 PM
Screenshot 2023-06-05 at 4 29 31 PM
Screenshot 2023-06-05 at 4 29 40 PM
Screenshot 2023-06-05 at 4 29 48 PM

### DIFF
--- a/static/js/components/tiles/histogram_tile.tsx
+++ b/static/js/components/tiles/histogram_tile.tsx
@@ -178,10 +178,12 @@ function binData(
 
     // Get value of property to aggregate
     let eventValue = 1;
-    if (property) {
+    if (property && event.displayProps) {
       // default to 0 if property can't be found in display props
-      eventValue =
-        stripUnitFromPropertyValue(event.displayProps[property]) || 0;
+      const propertyValueString = event.displayProps[property];
+      eventValue = propertyValueString
+        ? stripUnitFromPropertyValue(event.displayProps[property])
+        : 0;
     }
 
     // Increment value in corresponding bin if event has at least
--- a/static/js/components/tiles/histogram_tile.tsx
+++ b/static/js/components/tiles/histogram_tile.tsx
@@ -182,7 +182,7 @@ function binData(
       // Default to 0 if property can't be found in display props
       const propertyValueString = event.displayProps[property];
       eventValue = propertyValueString
-        ? stripUnitFromPropertyValue(event.displayProps[property])
+        ? stripUnitFromPropertyValue(propertyValueString)
         : 0;
     }
 
--- a/static/js/components/tiles/histogram_tile.tsx
+++ b/static/js/components/tiles/histogram_tile.tsx
@@ -19,7 +19,7 @@
  */
 
 import _ from "lodash";
-import React, { memo, useCallback, useEffect, useRef } from "react";
+import React, { memo, useCallback, useRef } from "react";
 
 import { DataPoint } from "../../chart/base";
 import { drawHistogram } from "../../chart/draw";
@@ -32,6 +32,7 @@ import {
 } from "../../types/disaster_event_map_types";
 import { EventTypeSpec } from "../../types/subject_page_proto_types";
 import { getDateRange } from "../../utils/disaster_event_map_utils";
+import { stripUnitFromPropertyValue } from "../../utils/property_value_utils";
 import { ReplacementStrings } from "../../utils/tile_utils";
 import { ChartTileContainer } from "./chart_tile";
 import { useDrawOnResize } from "./use_draw_on_resize";
@@ -179,7 +180,8 @@ function binData(
     let eventValue = 1;
     if (property) {
       // default to 0 if property can't be found in display props
-      eventValue = event.displayProps[property] || 0;
+      eventValue =
+        stripUnitFromPropertyValue(event.displayProps[property]) || 0;
     }
 
     // Increment value in corresponding bin if event has at least
--- a/static/js/components/tiles/histogram_tile.tsx
+++ b/static/js/components/tiles/histogram_tile.tsx
@@ -179,7 +179,7 @@ function binData(
     // Get value of property to aggregate
     let eventValue = 1;
     if (property && event.displayProps) {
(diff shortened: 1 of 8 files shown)


juliawu:
@shifucun @chejennifer (Reviewers): This is a large change, comments on the approach also welcome. If you think a different approach would work better, I can close this PR and implement the different approach instead.

static/js/utils/property_value_utils.tsx
```diff
@@ -0,0 +1,108 @@
...
+  if (!val) {
+    return "";
+  }
+  return formatPropertyValue(val);
+}
+
+/**
+ * Parses a lat,long pair from the first value of a property value node.
+ */
+export function parseLatLong(property: Node[]): [number, number] {
```
chejennifer, line 87: nit: don't need to export helper functions

static/js/utils/property_value_utils.tsx
```diff
@@ -0,0 +1,108 @@
...
+    unit = unit.trim();
+  }
+  const num = Number.parseFloat(val.substring(numIndex));
+  return formatNumber(num, unit);
+}
+
+/**
+ * Formats the first display value of a property value node as a number with unit.
+ */
+export function formatPropertyNodeValue(property: Node[]): string {
```
chejennifer, line 76: nit: for exported functions, in the comments, want to also include @param like: https://github.com/datacommonsorg/website/blob/master/static/js/utils/geojson_utils.ts#L25-L29

static/js/utils/disaster_event_map_utils.tsx
```diff
@@ -262,12 +262,8 @@ function fetchEventPoints(
         if (eventTypeSpec.displayProp) {
           for (const dp of eventTypeSpec.displayProp) {
             if (dp.prop in eventData.propVals) {
-              const val = parseEventPropVal(
-                eventData.propVals[dp.prop].vals,
-                dp.unit || "",
-                reqParams
-              );
-              if (!isNaN(val)) {
+              const val = eventData.propVals[dp.prop].vals[0];
```
chejennifer, line 265: check that vals is non empty before accessing the first element? Or just use the `getValue` util function to get the value with all of this handled

static/js/components/tiles/histogram_tile.tsx
```diff
@@ -179,7 +180,8 @@ function binData(
     let eventValue = 1;
     if (property) {
       // default to 0 if property can't be found in display props
-      eventValue = event.displayProps[property] || 0;
+      eventValue =
+        stripUnitFromPropertyValue(event.displayProps[property]) || 0;
```
chejennifer, line 184: should check that event.displayProps[property] is not undefined?

chejennifer (commented):
nice! this approach makes sense to me

static/js/utils/property_value_utils.tsx
juliawu: Added `@param`s to each exported function

static/js/utils/property_value_utils.tsx
juliawu: Keeping this function exported, it gets used here: https://github.com/datacommonsorg/website/blob/b8b597ff9a5d777f3d50b3d01cfd49509354cf1d/static/js/apps/event/app.tsx#L86

static/js/utils/disaster_event_map_utils.tsx
juliawu: Added check for non-empty, `getValue()` expects a list of nodes as input instead.

static/js/components/tiles/histogram_tile.tsx
juliawu: Added check!

static/js/components/tiles/histogram_tile.tsx
```diff
@@ -177,9 +178,12 @@ function binData(
...
     // Get value of property to aggregate
     let eventValue = 1;
-    if (property) {
-      // default to 0 if property can't be found in display props
-      eventValue = event.displayProps[property] || 0;
+    if (property && event.displayProps) {
+      // Default to 0 if property can't be found in display props
+      const propertyValueString = event.displayProps[property];
+      eventValue = propertyValueString
+        ? stripUnitFromPropertyValue(event.displayProps[property])
```
chejennifer, line 185: nit: pass in propertyValueString?

static/js/utils/disaster_event_map_utils.tsx
```diff
@@ -49,6 +49,7 @@ import {
   EventTypeSpec,
   SeverityFilter,
 } from "../types/subject_page_proto_types";
+import { getValue } from "./property_value_utils";
```
chejennifer, line 52: nit: seems unused?

chejennifer (approved):
thanks for the updates!

static/js/components/tiles/histogram_tile.tsx
juliawu: Thanks for the catch! Fixed.

static/js/utils/disaster_event_map_utils.tsx
juliawu: Thanks for the catch! Removed.

PR #2779, opened 2023-06-06
Deploy node app

chejennifer:
- Create config files needed for deploying the node server app.
- Tested on stanford staging: https://staging.stanford.datacommons.org/nodejs/query?q=population%20of%20counties%20in%20the%20us
- Removed the config changes for stanford staging & made changes for autopush and dev configs


build/nodejs_server/Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+FROM node:18.4.0
+
+ARG ENV
+ENV ENV=${ENV}
+
+WORKDIR /workspace
+
+# Install netstat which is useful for debugging.
+RUN apt-get -y update
+RUN apt-get -y install net-tools
```
shifucun, line 24: can remove these now?

build/nodejs_server/Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+
+# Install dependencies
+COPY static/package.json /workspace/static/package.json
+COPY static/package-lock.json /workspace/static/package-lock.json
+WORKDIR /workspace/static
+RUN npm install --omit=dev
+
+COPY static/. /workspace/static/
+# Expose the port that the Node.js application will listen on
+EXPOSE 8080
```
shifucun, line 34: Add a comment that 8080 is crucial for health check as we see in google.com projects

deploy/helm_charts/dc_website/templates/deployment.yaml
```diff
@@ -130,4 +130,73 @@ spec:
...
+          image:  "gcr.io/{{ .Values.website.image.project }}/datacommons-nodejs:{{ .Values.website.image.tag }}"
+          imagePullPolicy: {{ .Values.website.image.pullPolicy }}
+          args: []
+          ports:
+            - containerPort: 8080
+          resources:
+            limits:
+              memory: "3G"
+            requests:
+              memory: "3G"
```
shifucun, line 176: I think 1 or 2G memory should be more than enough, since all this does is to render an svg?

deploy/helm_charts/dc_website/values.yaml
```diff
@@ -122,3 +122,14 @@ nl:
...
+
+###############################################################################
+# Config for NodeJS Server
+###############################################################################
+nodejs:
+  # If nodejs is enabled, must also enable nl.
+  # When enabling the NodeJS server for the first time for a project, need to
+  # update the cluster config by following instructions here: https://github.com/datacommonsorg/website/blob/master/gke/README.md
+  enabled: false
+  env: "prod"
```
shifucun, line 134: s/prod/gke, i intially thought prod is the website env, but realized this is to contrast local

static/src/server.ts
```diff
@@ -49,7 +49,19 @@ import { NamedTypedPlace, StatVarSpec } from "../js/shared/types";
 import { TileConfig } from "../js/types/subject_page_proto_types";
 
 const app = express();
-const port = process.env.PORT || 3030;
+const DEFAULT_CONFIG = {
+  port: 8080,
+  apiRoot: "http://mci-website-mcs-svc-ta0n5jd94jdh7kdn:8080",
```
shifucun, line 54: can you check how to get a stable server name? This will likely change with new deployment, and will be different across projects.

static/src/server.ts
```diff
@@ -49,7 +49,19 @@ import { NamedTypedPlace, StatVarSpec } from "../js/shared/types";
...
+};
+const APP_CONFIGS = {
+  local: {
+    port: 3030,
+    apiRoot: "http://127.0.0.1:8080",
+  },
```
shifucun, line 60: APP_CONFIGS = {
  local: {}
  gke: {}
}

Then can remove default?

build/nodejs_server/Dockerfile
chejennifer: done

build/nodejs_server/Dockerfile
chejennifer: done

deploy/helm_charts/dc_website/templates/deployment.yaml
chejennifer: it will eventually be multiple svgs and multiple data csvs, but you're right, don't need 3G. Updated

deploy/helm_charts/dc_website/values.yaml
chejennifer: done

static/src/server.ts
chejennifer: the service name is derived from the MultiClusterService config & can't be controlled by the config .. but updated the deploy script to figure out the service name and then pass into the helm chart as an environment variable for the nodejs server container

static/src/server.ts
chejennifer: good idea, done

shifucun (approved):
thanks for the update! Look forward to see this running!

chejennifer:
thanks for the review!

PR #2784, opened 2023-06-06
[subject page] allow place groups for a topic page config

chejennifer:
- add list of allowed place groups in the subject page config
- allow equity page for all states in USA

Screenshot 2023-06-06 at 1 37 11 PM

bug fixes along the way (caused by updates in accepted parameter type and return value of parent_places):
- fix bug in getting parent place type in choropleth API
- fix bug in returned parent place list from subject page config place metadata

### DIFF
--- a/server/routes/topic_page/html.py
+++ b/server/routes/topic_page/html.py
@@ -24,6 +24,7 @@ from google.protobuf.json_format import MessageToJson
 import server.lib.subject_page_config as lib_subject_page_config
 import server.lib.util as libutil
 import server.routes.shared_api.place as place_api
+import server.services.datacommons as dc
 
 _NL_DISASTER_TOPIC = 'nl_disasters'
 _SDG_TOPIC = 'sdg'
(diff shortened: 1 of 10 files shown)


server/routes/topic_page/html.py
```diff
@@ -96,6 +96,11 @@ def topic_page(topic_id=None, place_dcid=None):
 
   more_places = request.args.getlist('places')
 
+  # TODO: should use place metadata API to fetch these data in one call.
+  place_type = place_api.get_place_type(place_dcid)
```
shifucun, line 100: Can just use dc.get_place_info() to fetch metadata?

shifucun (approved):
very nice!

server/routes/topic_page/html.py
chejennifer: done

chejennifer:
thanks for the review!

pradh:
This is cool!

PR #2795, opened 2023-06-08
[Maps Tool] Enable Census Tracts in States

juliawu:
Screenshot 2023-06-07 at 5 30 08 PM

This PR enables the map tool to plot "Places in <US State> of type <CensusTract>" for all US states. The states California, Florida, New York, and Texas have >5,000 census tracts, which requires special handling to avoid server limits. This PR makes multiple ("batched") api calls as as a temporary workaround until caches can be updated.

Note that render times for California, Florida, New York, and Texas is VERY slow (map takes ~40 seconds to load). 

More specifically, this PR:
* Edit webserver endpoints to make multiple calls for census tracts in CA, FL, NY, and TX
* Add helper functions for merging batched api call responses into a single response
* Add tests for the new helper functions

Testing:
Verified changes don't break Place pages, disaster dashboard, sustainability pages, scatter plot explorer, timelines explorer

### DIFF
--- a/server/lib/shared.py
+++ b/server/lib/shared.py
@@ -21,13 +21,14 @@ import server.lib.fetch as fetch
 # Determines when to make batched API calls to avoid server errors.
 # TODO(juliawu): This is a temporary workaround. Remove once cache is updated.
 NEEDS_SPECIAL_HANDLING = {
-  "CensusTract": [
-    "geoId/06", # California
-    "geoId/12", # Florida
-    "geoId/36", # New York (State)
-    "geoId/48", # Texas
-  ]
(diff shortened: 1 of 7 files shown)


juliawu:
@shifucun @chejennifer (reviewers): For CA, FL, TX, NY, the page load time is very long (~40 secs). If we think this is too long, I can also just disable CensusTracts as an option for those states instead.

juliawu:
Converting to draft -- A cache update will supersede these changes. Will close once cache changes land.

juliawu:
Closing because changes are superceded by a cache update.

PR #2797, opened 2023-06-09
Fix undefined issue in stat var summary page name display

shifucun:


chejennifer (approved):
thanks for the fix!

shifucun (commented):
thanks!

PR #2804, opened 2023-06-12
[Maps] Add Census Tracts in a State

juliawu:
Screenshot 2023-06-12 at 1 36 52 PM

Adds the ability to plot census tracts within a US state to the Map Tool. Load times are <10 secs when testing locally.

In particular, the following changes are made:
* Enable census tracts as an enclosed place type for states
* Make batched calls for observation series for the 4 states that have >5k tracts to avoid server hanging (series are too big to fit into cache)
* Add helper functions and tests for batched calls
* Add cached geojsons for the 4 states with >5k tracts to reduce load time.

Not in this PR, but will be added in followup PRs:
* Use optimized geojsons to further reduce load time.
* Let census tracts be selectable for states in the scatter plot explorer and data download tool.

Testing:
Verified changes don't break: map tool, scatter plot explorer, place pages, ranking pages, map tiles, nl interface, disaster dashboard

### DIFF
--- a/server/lib/shared.py
+++ b/server/lib/shared.py
@@ -21,13 +21,14 @@ import server.lib.fetch as fetch
 # Determines when to make batched API calls to avoid server errors.
 # TODO(juliawu): This is a temporary workaround. Remove once cache is updated.
 NEEDS_SPECIAL_HANDLING = {
-  "CensusTract": [
-    "geoId/06", # California
-    "geoId/12", # Florida
-    "geoId/36", # New York (State)
-    "geoId/48", # Texas
-  ]
+    "CensusTract": [
+        "geoId/06",  # California
+        "geoId/12",  # Florida
+        "geoId/36",  # New York (State)
+        "geoId/48",  # Texas
+    ]
 }
+
 
 def names(dcids):
   """Returns display names for set of dcids.
@@ -154,7 +155,7 @@ def divide_into_batches(all_items: list, batch_size: int):
   batch size. Used to make batched calls to mixer.
   """
   for i in range(0, len(all_items), batch_size):
-    yield all_items[i:i+batch_size]
+    yield all_items[i:i + batch_size]
 
 
 def merge_responses(resp_1: dict, resp_2: dict) -> dict:
@@ -168,11 +169,11 @@ def merge_responses(resp_1: dict, resp_2: dict) -> dict:
       if key in resp_2:
         if type(resp_2[key]) == dict:
           merged_resp[key] = merge_responses(resp_1[key], resp_2[key])
-      else: # key is not in second response
+      else:  # key is not in second response
         merged_resp[key] = val
     elif type(val) == list:
-        if key in resp_2 and type(resp_2[key]) == list:
-          merged_resp[key] = resp_1[key] + resp_2[key]
+      if key in resp_2 and type(resp_2[key]) == list:
+        merged_resp[key] = resp_1[key] + resp_2[key]
     else:
       merged_resp[key] = val
 
@@ -180,5 +181,5 @@ def merge_responses(resp_1: dict, resp_2: dict) -> dict:
   for key, val in resp_2.items():
     if not key in resp_1:
       merged_resp[key] = val
-  
+
   return merged_resp
--- a/server/lib/shared.py
+++ b/server/lib/shared.py
@@ -182,4 +182,3 @@ def merge_responses(resp_1: dict, resp_2: dict) -> dict:
       merged_resp[key] = val
   
   return merged_resp
-      
--- a/server/lib/shared.py
+++ b/server/lib/shared.py
@@ -17,6 +17,17 @@ import re
 
 import server.lib.fetch as fetch
 
+# Maps enclosed place type -> places with too many of the enclosed type
+# Determines when to make batched API calls to avoid server errors.
+# TODO(juliawu): This is a temporary workaround. Remove once cache is updated.
+NEEDS_SPECIAL_HANDLING = {
+  "CensusTract": [
+    "geoId/06", # California
+    "geoId/12", # Florida
+    "geoId/36", # New York (State)
+    "geoId/48", # Texas
+  ]
+}
 
 def names(dcids):
   """Returns display names for set of dcids.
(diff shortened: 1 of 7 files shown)


server/routes/shared_api/observation/series.py
```diff
@@ -64,6 +65,22 @@ def series_within():
   variables = list(filter(lambda x: x != "", request.args.getlist('variables')))
   if not variables:
     return 'error: must provide a `variables` field', 400
+  batch_size = request.args.get('batch_size') or 5000
```
chejennifer, line 68: nit: make 5000 a constant?

server/lib/shared.py
```diff
@@ -17,6 +17,17 @@
 
 import server.lib.fetch as fetch
 
+# Maps enclosed place type -> places with too many of the enclosed type
+# Determines when to make batched API calls to avoid server errors.
+NEEDS_SPECIAL_HANDLING = {
```
chejennifer, line 22: nit: BATCHED_CALL_PLACES or something that describes this special group of places. NEEDS_SPECIAL_HANDLING is very vague

also, curious why putting this constant in the shared file when it's specific to observation?

server/routes/shared_api/observation/series.py
```diff
@@ -64,6 +65,22 @@ def series_within():
...
+  # Make batched calls if it's a special case, as defined in shared.py
+  if (child_type in shared.NEEDS_SPECIAL_HANDLING and
```
chejennifer, line 70: nit: could just do `parent_entity in shared.NEEDS_SPECIAL_HANDLING.get(child_type, [])`

server/routes/shared_api/observation/series.py
```diff
@@ -64,6 +65,22 @@ def series_within():
...
+      parent_entity in shared.NEEDS_SPECIAL_HANDLING[child_type]):
+    try:
+      child_places = fetch.descendent_places([parent_entity],
```
chejennifer, line 73: nit: use `.get(parent_entity, [])` instead of just accessing it in case the key isn't in the response?

server/lib/shared.py
```diff
@@ -136,3 +147,38 @@ def date_lesser_equal_max(date, max_date):
...
+  for key, val in resp_1.items():
+    if type(val) == dict:
+      if key in resp_2:
+        if type(resp_2[key]) == dict:
+          merged_resp[key] = merge_responses(resp_1[key], resp_2[key])
+      else:  # key is not in second response
+        merged_resp[key] = val
+    elif type(val) == list:
+      if key in resp_2 and type(resp_2[key]) == list:
+        merged_resp[key] = resp_1[key] + resp_2[key]
```
chejennifer, line 175: if type(resp_2[key]) isn't a list, we probably want to set merged_resp[key] to be val right?

server/lib/shared.py
```diff
@@ -136,3 +147,38 @@ def date_lesser_equal_max(date, max_date):
...
+def merge_responses(resp_1: dict, resp_2: dict) -> dict:
+  """Merge the response of two calls to the same API into a single response.
+  Requires the two responses to have identical keys. In the case of conflicts,
+  will default to using values from the first given response.
+  """
+  merged_resp = {}
+  for key, val in resp_1.items():
+    if type(val) == dict:
+      if key in resp_2:
+        if type(resp_2[key]) == dict:
```
chejennifer, line 169: can probably combine this if statement with the one in 168?

chejennifer (commented):
nice!

server/lib/shared.py
juliawu: Great catch! Added.

server/lib/shared.py
juliawu: Good catch! Combined the two statements.

server/routes/shared_api/observation/series.py
juliawu: Good point, switched to using .get()

server/routes/shared_api/observation/series.py
juliawu: Great suggestion, changed!

server/lib/shared.py
juliawu: leaving it in shared file was a holdover from when `point.py` also needed to use it. Now that it's specific to series, moved this to `series.py` and renamed to `BATCHED_CALL_PLACES`

server/routes/shared_api/observation/series.py
juliawu: Made it a constant

server/lib/shared.py
```diff
@@ -136,3 +136,42 @@ def date_lesser_equal_max(date, max_date):
...
+  for key, val in resp_1.items():
+    if type(val) == dict:
+      if key in resp_2 and type(resp_2[key]) == dict:
+        merged_resp[key] = merge_responses(resp_1[key], resp_2[key])
+      else:
+        # key is not in resp_2, or values conflict, take value from resp_1
+        merged_resp[key] = val
+    elif type(val) == list:
+      if key in resp_2:
+        if type(resp_2[key]) == list:
```
chejennifer, line 164: I think you also want to keep this if statement combined with 163? because if key is not in resp_2, you also want to set merged_resp[key] to val?

server/routes/shared_api/observation/series.py
```diff
@@ -64,6 +79,21 @@ def series_within():
   variables = list(filter(lambda x: x != "", request.args.getlist('variables')))
   if not variables:
     return 'error: must provide a `variables` field', 400
+  batch_size = request.args.get('batch_size') or _MAX_BATCH_SIZE
```
chejennifer, line 82: nit: usually we want request arguments to be camelCase

chejennifer (commented):
thanks for the updates!

server/routes/shared_api/observation/series.py
juliawu: Switched to camelCase

server/lib/shared.py
juliawu: I think you're right! It should mirror the structure of the top if. Changed.

chejennifer (approved):
thanks for the updates!

PR #2805, opened 2023-06-13
[node server] update response format

chejennifer:
- update response format to match the selenium version
- return all tiles from nl response

Screenshot 2023-06-13 at 8 29 40 AM

### DIFF
--- a/static/js/components/tiles/chart_tile.tsx
+++ b/static/js/components/tiles/chart_tile.tsx
@@ -25,6 +25,7 @@ import { INITAL_LOADING_CLASS } from "../../constants/tile_constants";
 import { ChartEmbed } from "../../place/chart_embed";
 import {
   formatString,
+  getChartTitle,
   getMergedSvg,
   ReplacementStrings,
 } from "../../utils/tile_utils";
@@ -50,10 +51,7 @@ export function ChartTileContainer(props: ChartTileContainerProp): JSX.Element {
   const containerRef = useRef(null);
   const embedModalElement = useRef<ChartEmbed>(null);
   // on initial loading, hide the title text
-  const title =
-    props.title && !props.isInitialLoading
-      ? formatString(props.title, props.replacementStrings)
-      : "";
+  const title = !props.isInitialLoading ? getChartTitle : "";
   const showEmbed = props.allowEmbed && !props.isInitialLoading;
   return (
     <div
--- a/static/js/components/tiles/line_tile.tsx
+++ b/static/js/components/tiles/line_tile.tsx
@@ -81,15 +81,11 @@ export function LineTile(props: LineTilePropType): JSX.Element {
   }, [props, chartData]);
 
   useDrawOnResize(drawFn, svgContainer.current);
-
-  const rs: ReplacementStrings = {
-    placeName: props.place.name,
-  };
   return (
     <ChartTileContainer
       title={props.title}
       sources={chartData && chartData.sources}
-      replacementStrings={rs}
+      replacementStrings={getReplacementStrings(props)}
       className={`${props.className} line-chart`}
       allowEmbed={true}
       getDataCsv={chartData ? () => dataGroupsToCsv(chartData.dataGroup) : null}
@@ -109,6 +105,14 @@ export function LineTile(props: LineTilePropType): JSX.Element {
       ></div>
     </ChartTileContainer>
   );
+}
+
+export function getReplacementStrings(
+  props: LineTilePropType
+): ReplacementStrings {
+  return {
+    placeName: props.place.name,
+  };
 }
 
 export const fetchData = async (props: LineTilePropType) => {
--- a/static/js/components/tiles/map_tile.tsx
+++ b/static/js/components/tiles/map_tile.tsx
@@ -139,15 +139,11 @@ export function MapTile(props: MapTilePropType): JSX.Element {
     setSvgHeight(svgHeight);
   }, [props]);
 
-  const rs: ReplacementStrings = {
-    placeName: props.place.name,
-    date: mapChartData && mapChartData.dateRange,
-  };
   return (
     <ChartTileContainer
       title={props.title}
       sources={mapChartData && mapChartData.sources}
-      replacementStrings={rs}
+      replacementStrings={getReplacementStrings(props, mapChartData)}
       className={`${props.className} map-chart`}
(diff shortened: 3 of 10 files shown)


static/js/components/tiles/chart_tile.tsx
```diff
@@ -50,10 +51,7 @@ export function ChartTileContainer(props: ChartTileContainerProp): JSX.Element {
   const containerRef = useRef(null);
   const embedModalElement = useRef<ChartEmbed>(null);
   // on initial loading, hide the title text
-  const title =
-    props.title && !props.isInitialLoading
-      ? formatString(props.title, props.replacementStrings)
-      : "";
+  const title = !props.isInitialLoading ? getChartTitle : "";
```
shifucun, line 54: getChartTitle()

static/js/components/tiles/line_tile.tsx
```diff
@@ -111,6 +107,14 @@ export function LineTile(props: LineTilePropType): JSX.Element {
   );
 }
 
+export function getReplacementStrings(
```
shifucun, line 110: Add comment on what the replacement string is used for?

static/js/components/tiles/map_tile.tsx
```diff
@@ -243,6 +249,8 @@ export const fetchData = async (
       parentPlaces,
       borderGeoJson,
     };
+    //console.log("got raw")
+    //console.log(props)
```
shifucun, line 253: remove?

static/src/server.ts
```diff
@@ -172,15 +200,41 @@ function getTextLength(text: string): number {
...
+// Processes and serializes a svg for a chart.
+function getProcessedSvg(chartSvg: SVGSVGElement): string {
+  if (!chartSvg) {
+    return "";
+  }
+  // Set the font for all the text in the svg
+  chartSvg.querySelectorAll("text").forEach((node) => {
+    node.setAttribute("font-family", FONT_FAMILY);
+    node.setAttribute("font-size", FONT_SIZE);
+  });
```
shifucun, line 222: comment on why the font size was not size in the original svg?

static/src/server.ts
```diff
@@ -190,56 +244,92 @@ function getScatterTileSvg(
...
+    .then((chartData) => {
+      const tileContainer = document.createElement("div");
+      tileContainer.setAttribute("id", id);
+      document.getElementById(DOM_ID).appendChild(tileContainer);
+      drawLine(tileProp, chartData, null);
+      return {
+        svg: getProcessedSvg(tileContainer.querySelector("svg")),
+        data_csv: dataGroupsToCsv(chartData.dataGroup),
+        srcs: getSources(chartData.sources),
+        legend: chartData.dataGroup.map((dg) => dg.label || "A"),
```
shifucun, line 307: what is "A"?

static/src/server.ts
```diff
@@ -249,68 +339,140 @@ function getBarTileSvg(
...
+      // Get the legend svg and add it to the merged svg
+      const legendSvg = legendContainer.querySelector("svg");
+      legendSvg.setAttribute("width", String(legendWidth));
+      const legendG = document.createElementNS(
+        "http://www.w3.org/2000/svg",
+        "g"
+      );
+      legendG.setAttribute("transform", `translate(${mapWidth})`);
+      legendG.appendChild(legendSvg);
+      mergedSvg.appendChild(legendG);
```
shifucun, line 428: I wonder can the merge happen in the original map component? I wonder if this merge has to happen in multiple places now?

static/src/server.ts
```diff
@@ -342,78 +502,49 @@ app.get("/nodejs/query", (req: Request, res: Response) => {
...
       }
 
-      // Get the list of tiles and the stat var spec for the first category in
-      // the config.
-      const categories = config["categories"];
-      if (!_.isEmpty(categories)) {
-        const category = categories[0];
+      // If no place, return here
+      if (!place.dcid) {
+        res.status(200).send("No data found");
```
shifucun, line 507: Since response is a json, should this be a structured object. Like have an "err" field in the response and set it here.

static/src/server.ts
chejennifer: I'm not sure, but that's what's used for the legend text in a line chart: https://github.com/datacommonsorg/website/blob/master/static/js/chart/draw.ts#L1024-L1026

static/src/server.ts
chejennifer: - doing it as a "merge" (draw legend and map and then merge into one svg) doesn't make the most sense because it is forcing a redraw that will sometimes be unnecessary & will need some thought to account for resizing.
- doing it as just drawing legend and map into the same container will need some thought to ensure the drawing functions don't break other map components (e.g., map explorer we would want to keep the legend and actual map separate because we want to zoom the map without affecting the legend)

So I think this could be a future thing & I could open an issue for it?

static/src/server.ts
chejennifer: Updated so that the 200 responses with no data just return an empty array for charts & the 500 responses return a json object with an "err" field

static/src/server.ts
chejennifer: done

static/js/components/tiles/map_tile.tsx
chejennifer: done

static/js/components/tiles/line_tile.tsx
chejennifer: done

static/js/components/tiles/chart_tile.tsx
chejennifer: good catch! done

static/src/server.ts
shifucun: I am not fully sure why option 1 is a redraw since it just operates on html element. But ok to solve it later. Worth to add where this merging happens (I guess here and in chart embed?)

shifucun:
thanks for update!

static/src/server.ts
chejennifer: ah you're right, no need to redraw, just need to manipulate the html elements. But will take more thought especially around calculating BBox (which the chart embed merge function uses to determine where to move each element).

created issue: https://github.com/datacommonsorg/website/issues/2808

shifucun (approved):

PR #2806, opened 2023-06-13
Introduce basic PaLM API based detection

pradh:
**Usage:**

To use this, add `&llm=True` to the NL URL.

**API Access:**

This PR wraps the prompt and API call found in [this notebook](https://colab.sandbox.google.com/drive/1HjoxjCPRHUoVcMatIf1Q6LHWThxhHKvS?resourcekey=0-xsDaW2emy3gLWxfMZ-hLzQ#scrollTo=DfJ1DLe8x8U5).  There is a new PY API that Google recommends, which I tried out in [this notebook](https://colab.sandbox.google.com/drive/1yvNztmy82BC18f9ZwAM4qO1XGFjNDDV1?resourcekey=0-QwuOwSOfsyyGyVJ_NvUGjQ#scrollTo=w3L03bUQsAHT), but it seems to be more flaky given the same inputs.  So I'm going with the REST API.

**Detector Integration:**

There's now a fork/peer of detector, "llm_detector", which first calls LLM then does place-to-dcid and sv-to-dcid calls.

Mapping the other heuristics to Classifications data structure will come in follow on PR.

**Environments:**

The PaLM API requires a key which is now supported only in local and autopush.

**TODOs:**

* Integration test
* Inferring other heuristics


server/lib/nl/detection/palm_api.py
```diff
@@ -0,0 +1,336 @@
...
+}
+
+_SKIP_BEGIN_CHARS = ['`', '*']
+
+
+def call(query: str, history: str, ctr: counters.Counters) -> str:
+  logging.info(f'Doing PaLM API for {query}')
+  req_data = _REQ_DATA.copy()
+
+  if not history:
```
jehangiramjad, line 284: nit: a comment on how these are different and perhaps why we need both separately?

server/lib/nl/detection/palm_api.py
```diff
@@ -0,0 +1,336 @@
...
+                    headers=_API_HEADER)
+  resp = r.json()
+  ctr.timeit('palm_api_call', start_time)
+
+  return _parse_response(query, resp, ctr)
+
+
+def _parse_response(query: str, resp: Dict, ctr: counters.Counters) -> Dict:
+  if 'candidates' in resp and resp['candidates']:
+    ans = _extract_ans(resp['candidates'][0]['content'])
```
jehangiramjad, line 314: nit: supposedly if ans is empty it most likely means a response format which was not as expected, right? in that case, perhaps return {} and log and error/warning/info?

server/lib/nl/detection/llm_detector.py
```diff
@@ -0,0 +1,152 @@
...
+                   llm_resp=llm_resp)
+
+
+def _merge_sv_dicts(sv_list: List[str], svs_score_dicts: List[Dict]) -> Dict:
+  if not svs_score_dicts:
+    return _empty_svs_score_dict()
+  if len(svs_score_dicts) == 1:
+    return svs_score_dicts[0]
+
+  # This is the case of multiple since SV, which we should merge into
```
jehangiramjad, line 131: nit: not sure what this comment implies? Perhaps a typo somewhere?

jehangiramjad (approved):
Thanks. A few questions/comments

server/lib/nl/detection/palm_api.py
```diff
@@ -0,0 +1,336 @@
...
+import requests
+
+from server.lib.nl.common import counters
+
+_API_URL_BASE = f"https://generativelanguage.googleapis.com/v1beta2/models/chat-bison-001:generateMessage"
+_API_HEADER = {'content-type': 'application/json'}
+
+
+_PROMPT_CONTEXT_ORIG = \
+("""You are going to be a super helpful assistant that will convert an english sentence to a JSON object.
```
shifucun, line 31: do you want to have a text file to store the prompt only?

server/lib/nl/detection/palm_api.py
```diff
@@ -0,0 +1,336 @@
...
+            'content': input
+        },
+        'output': {
+            'content': json.dumps(output)
+        }
+    })
+
+  start_time = time.time()
+  req = json.dumps(req_data)
+  api_key = current_app.config['PALM_API_KEY']
```
shifucun, line 302: check api_key exists (or guard the caller with checking USE_PALM)?

shifucun (approved):

server/lib/nl/detection/llm_detector.py
pradh: oh oops, fixed.

server/lib/nl/detection/palm_api.py
pradh: Done.

server/lib/nl/detection/palm_api.py
```diff
@@ -0,0 +1,336 @@
...
+}
+```
+
+Do not make up new JSON properties outside of the JSON Schema.
+""")
+
+_SUFFIX = '\n\nIn your response, include just the JSON adhering to the above schema. Do not add JSON keys outside the schema.  Also, explain why you set specific enum values.'
+
+# TODO: Consider tweaking this.
+_TEMPERATURE = 1.0
```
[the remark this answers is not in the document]
pradh: Anecdotal and 2 months old, but I remember seeing better results with 1.0 than with 0.5 or so.  I didn't really tune it, having a TODO.

server/lib/nl/detection/palm_api.py
pradh: Good idea.  I'll do in followup PR (added TODO).

I'd then need to add load it into config in __init__.py and use it here right?

server/lib/nl/detection/palm_api.py
pradh: Good call.  Added a few checks.

server/lib/nl/detection/palm_api.py
pradh: This is checked before calling llm_detector.  Added comment.

pradh (commented):
Thank you both for the quick reviews!

PR #2807, opened 2023-06-13
[topic page] fix bug

chejennifer:
- base.html uses "config" assuming it is the app config, but topic page was passing in the page config as "config" causing the page to fail to load: https://autopush.datacommons.org/topic/poverty/country/USA


shifucun (approved):
Thanks for the fix!

PR #2814, opened 2023-06-14
Add css class to make stat var hierarchy screenshot stable

shifucun:
Also make place page sources stable.


server/webdriver/tests/timeline_test.py
```diff
@@ -150,9 +150,9 @@ def test_check_statvar_and_uncheck(self):
         (By.CLASS_NAME, 'dc-chart-exist'))
     WebDriverWait(self.driver, self.TIMEOUT_SEC).until(element_present)
 
-    # Assert there is one chart.
+    # Assert there is one chart and on opened hierarchy.
     charts = self.driver.find_elements(By.CLASS_NAME, 'dc-chart-exist')
-    self.assertEqual(len(charts), 1)
+    self.assertEqual(len(charts), 2)
```
chejennifer, line 155: nit: I think it's a little confusing that we are waiting on "dc-chart-exist" when waiting on stat var hierarchy to open .. Maybe we can either:

a. change the classnames to something more generic like dc-element-loaded & dc-container
b. have a separate classname for sv hierarchy exists

chejennifer (approved):
nice! Just one suggestion

server/webdriver/tests/timeline_test.py
shifucun: Good idea. Renamed with generic names.

chejennifer (approved):
thanks for updating!

PR #2817, opened 2023-06-14
Update screenshot pages with some new features

shifucun:
* Add "async" field to indicate if the page fetches (and should wait) asynchronously.
* Have per domain page.json as top level folder, hence moving PR merge screenshot to /local folder
* Separate out different Jinja html template for each page.
* Add a script to upload custom DC screenshot.


server/webdriver/screenshot/runner.py
```diff
@@ -25,23 +27,27 @@
...
     for page in config:
       # Set the window size. Testing different sizes.
       driver.set_window_size(width=WIDTH,
                              height=page['height'],
                              windowHandle='current')
-      driver.get(url + page['url'])
-      name = urllib.parse.quote_plus(page['url'].removeprefix('/'))
-      WebDriverWait(driver, shared.TIMEOUT).until(shared.charts_rendered)
+      driver.get(base_url + page['url'])
+      if page['async']:
```
chejennifer, line 44: Include a comment in the code about what 'async' means? Because looking at the code, it's not very clear & I had to refer to the PR description

chejennifer (approved):
this is cool!

server/webdriver/screenshot/runner.py
shifucun: Done

shifucun:
Thanks for review!

PR #2818, opened 2023-06-15
[nl] fix url prompts bug

chejennifer:
- fix bug with running the url prompts when the data fetch takes too long where the input will get cleared because the completed data fetch re-renders the search box. Fixed by inputting the next prompt once the previous data fetch is completed. 

screen-recording-_96_

### DIFF
--- a/static/js/apps/nl_interface/app.tsx
+++ b/static/js/apps/nl_interface/app.tsx
@@ -103,15 +103,11 @@ export function App(): JSX.Element {
 
   useEffect(() => {
     // If there are url prompts that have not been searched, input the next
-    // url prompt into the search box.
+    // url prompt into the search box when the last query's data fetch has
+    // completed.
     // TODO: Do this by going through state/props instead of directly
     // manipulating the DOM.
-    if (!urlPrompts.current.length) {
-      return;
-    }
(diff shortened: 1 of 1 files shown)


static/js/apps/nl_interface/app.tsx
```diff
@@ -101,21 +102,26 @@ export function App(): JSX.Element {
...
-    // If autoRun is enabled, runs every prompt (';' separated) from the url.
+    // If there are url prompts that have not been searched, input the next
+    // url prompt into the search box.
     // TODO: Do this by going through state/props instead of directly
     // manipulating the DOM.
-    if (urlPrompts.current.length) {
-      inputNextPrompt(false);
+    if (!urlPrompts.current.length) {
+      return;
+    }
```
pradh, line 111: Before this change, when `urlPrompts.current.length` was 0, we were clearing on unmount but not anymore.  That is intentional?

pradh (approved):
Thanks for the fix!

static/js/apps/nl_interface/app.tsx
chejennifer: good catch, updated

PR #2819, opened 2023-06-15
screenshot page support for remote domain

shifucun:
Add "domain=..." url param to show commit/compare pages for remote domain.

Also stabilize  place page by using seed=1 param.

### DIFF
--- a/server/routes/place/html.py
+++ b/server/routes/place/html.py
@@ -49,6 +49,7 @@ def place(place_dcid=None):
     return flask.redirect(flask.url_for('place.place', **redirect_args))
 
   dcid = flask.request.args.get('dcid', None)
+  seed = flask.request.args.get('seed', 0)
   if dcid:
     # Traffic from "explore more" in Search. Forward along all parameters,
     # except for dcid, to the new URL format.
@@ -76,7 +77,8 @@ def place(place_dcid=None):
                                place_name=place_name,
                                place_dcid=place_dcid,
                                category=category if category else '',
-                               maps_api_key=current_app.config['MAPS_API_KEY'])
+                               maps_api_key=current_app.config['MAPS_API_KEY'],
+                               seed=seed)
 
 
 def place_landing():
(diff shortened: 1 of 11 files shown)


server/routes/place/html.py
```diff
@@ -76,7 +77,8 @@ def place(place_dcid=None):
                                place_name=place_name,
                                place_dcid=place_dcid,
                                category=category if category else '',
-                               maps_api_key=current_app.config['MAPS_API_KEY'])
+                               maps_api_key=current_app.config['MAPS_API_KEY'],
+                               seed=seed)
```
chejennifer, line 81: curious why reading the url param value here instead of in place.ts

server/routes/screenshot/html.py
```diff
@@ -33,13 +35,44 @@
...
   """List recent commits and diff url
 
   Optional url argument "base" to set a base commit sha for comparison.
   """
+  if not env_valid():
+    flask.abort(404)
   base_sha = request.args.get('base', '')
+  domain = request.args.get('domain', '')
+
+  if domain:
```
chejennifer, line 55: So are we ignoring the base_sha if domain is set? Maybe add a comment about this?

server/templates/screenshot/commit.html
```diff
@@ -26,12 +26,16 @@
...
-      </div>
+      {% if sha|length == 7 %}
+        <div>Code Link:</div>
+        <div>
+            <a href="https://github.com/datacommonsorg/website/commit/{{ sha }}">
+              https://github.com/datacommonsorg/website/commit/{{ sha }}
+            </a>
+        </div>
+      {% else %}
+        <div><b>Date: {{ sha }}</b></div>
```
chejennifer, line 37: I think instead of re-using sha to hold the date, add a date field?

chejennifer (commented):
cool! Just some parts I think should be clarified with comments

shifucun (commented):
Thanks for review. Separate out the template for w/o domain to make it clear.

server/templates/screenshot/commit.html
shifucun: Use two templates for date and commit pages now

PR #2820, opened 2023-06-15
PaLM: Support filter queries

pradh:
With this PR, LLM detection has parity with heuristics approach, and slightly exceeds it.  Also, cope with some LLM misbehavior, where it returns two json blocks.

As with heuristics, it supports:
* california cities where the hispanic population is over 10000
* prevalence of Asthma in California cities with hispanic population over 10000

But also supports a "ranking" filter:  get X from places in Y where Z is the highest/lowest.

image

image


jehangiramjad (approved):

pradh:
Thanks Jehangir!

PR #2821, opened 2023-06-15
[node server] Add ranking tile results

chejennifer:
- add node server results for ranking tiles

Screenshot 2023-06-15 at 9 51 36 AM


static/src/server.ts
```diff
@@ -428,29 +444,108 @@ function getMapTileResult(
...
     })
     .catch(() => {
       console.log("Failed to get map tile result for: " + id);
       return null;
     });
 }
 
-// Gets the TileResult for a given tile config and supporting information.
+// Get the result for a single ranking unit
+function getRankingUnitResult(
```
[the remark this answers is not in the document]
chejennifer: yeup, that was on my own todo list, but added to the code now!

static/js/components/tiles/sv_ranking_units.tsx
```diff
@@ -163,71 +139,133 @@ export function SvRankingUnits(props: SvRankingUnitsProps): JSX.Element {
...
-                statVar: svName,
-              }
-            )}
-            numDataPoints={numDataPoints}
-            points={points.slice(0, rankingCount)}
-            isHighest={false}
-            svNames={isMultiColumn ? svNames : undefined}
-            formatNumberFn={formatNumber}
-          />
+          {getRankingUnit(
```
dwnoble, line 178: What do you think about making `getRankingUnit` a proper component? Maybe `<StatVarRankingUnit {...props} />`

static/src/server.ts
```diff
@@ -538,9 +644,9 @@ app.get("/nodejs/query", (req: Request, res: Response) => {
 
       Promise.all(tilePromises)
         .then((tileResults) => {
-          const filteredResults = tileResults.filter(
-            (result) => result !== null
-          );
+          const filteredResults = tileResults
+            .filter((result) => result !== null)
+            .flat(1);
```
dwnoble, line 649: Could any of the nested tile results be null? e.g., `[tileResult1, tileResult2, [tileResult3, null]]`- the filter might not catch these cases

static/src/server.ts
```diff
@@ -428,29 +446,108 @@ function getMapTileResult(
...
+    id,
+    title: tileConfig.title,
+    place,
+    enclosedPlaceType,
+    statVarSpec,
+    rankingMetadata: tileConfig.rankingTileSpec,
+    apiRoot: CONFIG.apiRoot,
+  };
+  return fetchRankingData(tileProp)
+    .then((rankingData) => {
```
dwnoble, line 520: consider using `await` here and making this function `async`

dwnoble (approved):

static/js/utils/svg_utils.ts
```diff
@@ -0,0 +1,70 @@
...
+  const svg = d3
+    .create("svg")
+    .attr("xmlns", SVGNS)
+    .attr("xmlns:xlink", XLINKNS)
+    .attr("width", svgWidth)
+    .attr("height", svgHeight);
+
+  svg
+    .append("g")
+    .attr("transform", `translate(${CHART_PADDING})`)
```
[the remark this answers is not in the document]
chejennifer: for now there are no callers that need to override this value .. so we can change this when needed?

static/js/components/tiles/sv_ranking_units.tsx
chejennifer: `getRankingUnit` does very few calculations and just renders a `RankingUnit` component so not sure it warrants its own component. I extracted it out as a function where most arguments come from props or datafetch so that it can be easily used by server to ensure any changes to the ranking tile will also result in a change in server. Let me know if you feel strongly about changing it into a component.

static/src/server.ts
chejennifer: the nested tile results couldn't be null, but flipped the order of flattening and filtering in case future changes could cause nested results to be null!

static/src/server.ts

static/js/utils/svg_utils.ts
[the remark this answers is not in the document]
dwnoble: sg!

static/js/components/tiles/sv_ranking_units.tsx
dwnoble: The function works for me

dwnoble (approved):

chejennifer:
thanks for the reviews!

PR #2825, opened 2023-06-16
Add screenshot capture for NL page

shifucun:
* Abstract out the test server setup as a base class
* Install default font for ubuntu


pradh (approved):
Nice change!

PR #2827, opened 2023-06-16
Add screenshot examples for feeding america; Support long url

shifucun:


static/js/shared/child_places.tsx
```diff
@@ -40,7 +40,7 @@ export function ChildPlaces(props: ChildPlacesPropType): JSX.Element {
   return (
     <div id="child-places">
       <span id="child-place-head">Places in {props.parentPlace.name}</span>
-      {Object.keys(props.childPlaces).map((placeType) => (
+      {[...Object.keys(this.props.childPlaces)].sort().map((placeType) => (
```
chejennifer, line 43: curious why we need to make a new array before sorting? I think you can just do `Object.keys(this.props.childPlaces).sort()`?

chejennifer (approved):
nice!

static/js/shared/child_places.tsx
shifucun: copied from another place, simplified now.

shifucun (commented):
thanks for review!

PR #2829, opened 2023-06-17
Update React from v16 to v17

dwnoble:
Updating react to v17 to fix an issue with exporting react chart components as web components.
React 17 fixes an issues where synthetic events were not properly observed when attached to a web component via the shadow DOM.

References:
* https://www.wpeform.io/blog/render-react-app-shadow-dom-styled-components/
* https://stackoverflow.com/questions/37866237/click-event-not-firing-when-react-component-in-a-shadow-dom

Dependency notes:
* React 17: https://legacy.reactjs.org/blog/2020/10/20/react-v17.html 
* Reactstrap 8.7.1: (smallest version bump to allow for react 17) https://github.com/reactstrap/reactstrap/blob/master/CHANGELOG.md
* Removed `enzyme-adapter-react-16`; added `@wojtekmaj/enzyme-adapter-react-17` - reference: https://stackoverflow.com/questions/64658031/which-enzyme-adapter-works-with-react-17
* Updated `react-test-renderer`

Tested changes locally (Map plot, scatter plot, topics page, sustainability page, NL)


shifucun (commented):
Nice! Can you check the npm test error, seems to be a lint error.

chejennifer (approved):
nice

PR #2837, opened 2023-06-22
[node server] clean up files

chejennifer:
- move functions into smaller files


static/nodejs_server/map_tile.ts
```diff
@@ -0,0 +1,114 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * Functions for getting tile result for a map tile
+ */
+
+// This import is unused in this file, but needed for draw functions
+import * as Canvas from "canvas";
```
shifucun, line 22: why only scatter and map need Canvas?

static/src/server.ts
```diff
@@ -124,20 +74,6 @@ const CHAR_WIDTHS = [
 const CHAR_AVG_WIDTH = 5.0341796875;
 // Height of a 10px Roboto character.
 const CHAR_HEIGHT = 13;
-// Height of the svg to render.
```
shifucun, line 127: move line 57-76 to constants.ts as well?

static/nodejs_server/map_tile.ts
chejennifer: map and scatter use the canvas for creating the color scale legend. The other tiles don't need this

static/src/server.ts
chejennifer: I kept those constants in server because they're only used in server.ts for mocking the getBBox functions ..

static/nodejs_server/bar_tile.ts
```diff
@@ -0,0 +1,101 @@
...
+ * @param apiRoot API root to use to fetch data
+ */
+export async function getBarTileResult(
+  id: string,
+  tileConfig: TileConfig,
+  place: NamedTypedPlace,
+  enclosedPlaceType: string,
+  statVarSpec: StatVarSpec[],
+  apiRoot: string
+): Promise<TileResult[]> {
```
dwnoble, line 53: Why return an array in these functions? Looks like they return a single `TileResult`

chejennifer: originally made it a list because one of the tiles (ranking) returns multiple TileResults so I just made all tiles return a list to keep it consistent. But that's not necessary, so updated

dwnoble (approved):

shifucun (approved):

chejennifer:
thanks for the reviews!

PR #2838, opened 2023-06-22
Fix screenshot test with better logging and handle NL page open

shifucun:


pradh (approved):
Thanks for the fix!

shifucun:
thanks!

PR #2841, opened 2023-06-22
[Library] Add bar chart web component

juliawu:
Add a web component to the new JS chart library to draw bar charts

Packaging the bar tile into a web component required a tweak in the way we draw/redraw tiles. Instead of identifying chart-related DOM elements by ID, we need to pass a reference to the container element to the draw function. This change is reflected in the diffs for `chart.tsx`, `server.ts`, `draw.ts`, `bar_tile.tsx` and `dev_page.tsx`.

Testing: Verified changes don't break place pages nor the NL interface

Limitations: This PR implements a 'simple' bar chart. Grouped bar charts will be added in a followup PR. Furthermore, the number of bars plotted is limited (see screenshot below, not all US States are plotted). Functionality to have the user set this limit can be implemented in a future PR.

Screenshot:
Screenshot 2023-06-22 at 4 36 54 PM


juliawu:
@pradh @shifucun fyi

experimental/datacommons-js/example.html
```diff
@@ -22,7 +22,7 @@
     <script src="https://cdnjs.cloudflare.com/ajax/libs/lodash.js/4.17.21/lodash.min.js"></script>
     <script src="https://cdnjs.cloudflare.com/ajax/libs/jquery/3.7.0/jquery.min.js"></script>
     <script src="https://cdnjs.cloudflare.com/ajax/libs/bootstrap/4.6.2/js/bootstrap.min.js"></script>
-    <script src="https://autopush.datacommons.org/datacommons.js"></script>
+    <script src="https://datacommons.org/datacommons.js"></script>
```
juliawu, line 25: Reviewers: Do we want to make this change here? Or should we stick to autopush?

dwnoble: Let's use `datacommons.org` since it will be deployed early next week

static/js/chart/draw.ts
```diff
@@ -829,6 +829,7 @@ function drawStackBarChart(
  * @param unit
  */
 function drawGroupBarChart(
+  containerElement: HTMLDivElement,
```
dwnoble, line 832: What's `containerElement` for?

juliawu: Before, updating the charts was done by finding the div within the tile with the correct id (old line 857 in this file). This no longer works because tile divs are now hidden under a shadow root. Instead, we pass in the div to update directly as `containerElement`. If you know of a better way to do this, please do share and I can update!

static/js/chart/draw.ts
```diff
@@ -853,8 +854,12 @@ function drawGroupBarChart(
     return;
   }
 
+  // clear old chart to redraw over
+  const container = d3.select(containerElement);
+  container.selectAll("*").remove();
+
   const svg = d3
-    .select("#" + id)
+    .select(containerElement)
```
chejennifer, line 862: nit: can replace d3.select(containerElement) with `container` that you defined in line 858

chejennifer (approved):
nice!

static/js/chart/draw.ts
juliawu: nice catch! Updated.

static/js/chart/draw.ts
dwnoble: That sounds good- maybe we can drop the id parameter all together in a future PR

PR #2842, opened 2023-06-23
[Library] Update map web component properties

juliawu:
Update the properties of the map web component to bring in line with other components

Specifically:
(1) Remove "statVarName", it is unused
(2) Rename "statVarDcid" to "variableDcid" to match convention of other components

As a result, a minimal example of embedding a datacommons map now looks like:
```
<script src="https://autopush.datacommons.org/datacommons.js"></script>
<datacommons-map
  title="Population Below Poverty Level Status in Past Year in States of United States (2020)"
  placeDcid="country/USA"
  enclosedPlaceType="State"
  variableDcid="Count_Person_BelowPovertyLevelInThePast12Months"
></datacommons-map>
```


pradh (approved):
Thanks.

We could let the folks tomm know to switch to variableDcid if they plan to use it for real.

PR #2844, opened 2023-06-23
LLM: Add some missing unit tests

pradh:
This covers much of core logic in the LLM flow, outside of API calls and such.


server/lib/nl/detection/llm_detector.py
```diff
@@ -200,37 +212,33 @@ def detect(query: str, context_history: Dict, index_type: str,
         continue
       cls = _handle_compare(llm_vals[0])
     elif t == filter_type:
-      # Earlier since we had set filter_type, should be safe to do llm_resp[f][0].
-      cls = _handle_quantity(llm_resp[f][0], t)
+      # Earlier caller had set filter_type by checking llm_resp,
+      # should be safe to do llm_resp[t][0].
+      cls = _handle_quantity(llm_resp[t][0], t)
```
shifucun, line 217: "cls" is a python reserved word, maybe use some other name? (i was a little confused when first see this)

server/tests/lib/nl/detection/llm_detector_test.py
```diff
@@ -0,0 +1,215 @@
...
+      }),
+      # When there are multiple, we get a MultiSV.
+      (['asthma', 'poverty'], [{
+          'SV': ['Count_Person_Asthma', 'Percent_Person_Asthma'],
+          'CosineScore': [0.9, 0.8]
+      }, {
+          'SV': ['Count_Person_BelowPoverty', 'Count_Person_FoodStamps'],
+          'CosineScore': [0.7, 0.4]
+      }], {
+          'CosineScore': [0.9, 0.8],
```
shifucun, line 61: why the CosineScore is only for the first sv here?

jehangiramjad (approved):

server/tests/lib/nl/detection/llm_detector_test.py
pradh: Good Q.  Because in this case of LLMs detecting multi-SV we *always* set AggCosineScore to be 1.0 so the MultiSV candidate will be picked over SV (single-sv candidate).  So the CosineScore does not matter as long as its <1.0
 https://github.com/datacommonsorg/website/blob/master/server/lib/nl/detection/llm_detector.py#L227-L228.

pradh (commented):
Thanks!

server/lib/nl/detection/llm_detector.py
pradh: Done.

PR #2846, opened 2023-06-23
[node server] add timing info

chejennifer:
- add timing info for the node server


pradh (approved):
Thanks.  It will be cool to see the huge timing improvements.

shifucun (approved):
nice! Really glad to see this launched!

PR #2847, opened 2023-06-23
[node server] add more precision to node server timing

chejennifer:
- just dividing big ints causes decimals to be lost. So instead, convert ns to ms, convert to number type & then divide by 1000 to get s with decimals

### DIFF
--- a/static/src/server.ts
+++ b/static/src/server.ts
@@ -271,6 +271,16 @@ function getDisasterBlockTileResults(
   return tilePromises;
 }
 
+// Get the elapsed time in seconds given the start and end times in nanoseconds.
+function getElapsedTime(startTime: bigint, endTime: bigint): number {
+  // Dividing bigints will cause decimals to be lost. Therefore, convert ns to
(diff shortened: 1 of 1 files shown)


static/src/server.ts
```diff
@@ -354,13 +355,15 @@ app.get("/nodejs/query", (req: Request, res: Response) => {
...
             timing: {
-              getNlResult: Number(
-                (nlResultTime - startTime) / TIMING_SCALE_FACTOR
-              ),
-              getTileResults: Number(
-                (endTime - nlResultTime) / TIMING_SCALE_FACTOR
-              ),
-              total: Number((endTime - startTime) / TIMING_SCALE_FACTOR),
+              getNlResult:
+                Number((nlResultTime - startTime) / NS_TO_MS_SCALE_FACTOR) /
```
pradh, line 359: nit: have a quick comment RE: precision?

pradh (approved):

static/src/server.ts

chejennifer:
thanks for the review!

PR #2848, opened 2023-06-23
[Library] Add support for comparing specific places to Bar Chart

juliawu:
Add support for the user to select specific places to plot in the bar chart

With this change, users can also now use
```
<script src="https://datacommons.org/datacommons.js"></script>
<datacommons-bar
  title="Population in States of United States"
  variableDcid="Count_Person"
></datacommons-bar>
```

to create the bar chart below:
image

### DIFF
--- a/experimental/datacommons-js/example.html
+++ b/experimental/datacommons-js/example.html
@@ -5,7 +5,7 @@
       rel="stylesheet"
       href="https://datacommons.org/css/nl_interface.min.css"
     />
-    <script src="http://localhost:8080/datacommons.js"></script>
+    <script src="https://datacommons.org/datacommons.js"></script>
   </head>
   <body>
     <h1>Datacommons Web Components</h1>
--- a/experimental/datacommons-js/example.html
+++ b/experimental/datacommons-js/example.html
@@ -5,7 +5,7 @@
       rel="stylesheet"
       href="https://datacommons.org/css/nl_interface.min.css"
     />
-    <script src="https://datacommons.org/datacommons.js"></script>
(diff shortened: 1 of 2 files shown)


shifucun:
For naming: the new param is "comparisonPlaces" without dcid, and we have other params using dcid. I wonder should we consolidate them, and might remove "dcid" like using "place=", "variable=" etc.. @btelles @pradh wdyt?

juliawu:
> For naming: the new param is "comparisonPlaces" without dcid, and we have other params using dcid. I wonder should we consolidate them, and might remove "dcid" like using "place=", "variable=" etc..

+1 to removing "dcid", this looks a lot cleaner

dwnoble:
This looks good to me, and +1 for dropping `dcid` from parameter names

dwnoble (approved):

juliawu:
Merging this so we have this feature in autopush, but will followup with a param renaming PR once we reach quorum/consensus on naming convention.

pradh:
> For naming: the new param is "comparisonPlaces" without dcid, and we have other params using dcid. I wonder should we consolidate them, and might remove "dcid" like using "place=", "variable=" etc.. @btelles @pradh wdyt?

Removing dcid from everywhere sg

pradh:
> > For naming: the new param is "comparisonPlaces" without dcid, and we have other params using dcid. I wonder should we consolidate them, and might remove "dcid" like using "place=", "variable=" etc.. @btelles @pradh wdyt?
> 
> Removing dcid from everywhere sg

Better to think through make these changes sooner.

PR #2851, opened 2023-06-23
Fix json dump bug in NL logging

shifucun:
The new MultiVarCandidates can not be dumped by json, using pickle instead.

Also fix some other small issues.

### DIFF
--- a/server/services/bigtable.py
+++ b/server/services/bigtable.py
@@ -16,6 +16,7 @@ from datetime import datetime
 from datetime import timedelta
 import json
 import os
+import pickle
 
 from flask import current_app
 import google.auth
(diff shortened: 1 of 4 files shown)


chejennifer (approved):
thanks for the fixes!

server/services/bigtable.py
```diff
@@ -87,7 +88,7 @@ async def write_row(session_info, data):
   row.set_cell(_COLUMN_FAMILY, _COL_PROJECT.encode(), project_id)
   row.set_cell(_COLUMN_FAMILY, _COL_SESSION.encode(), json.dumps(session_info))
   row.set_cell(_COLUMN_FAMILY, _COL_VERSION.encode(), json.dumps(version))
-  row.set_cell(_COLUMN_FAMILY, _COL_DATA.encode(), json.dumps(data))
+  row.set_cell(_COLUMN_FAMILY, _COL_DATA.encode(), pickle.dumps(data))
```
pradh, line 91: Maybe revert just this change, and I'll make a PR as followup?

pradh (approved):

server/services/bigtable.py
shifucun: I will hold off this PR for other purpose then, since this was aimed to fix this bug.

pradh: No, I don't think its from the recent change that added MultiSV.  Generally data_dict is not safe to convert to json.

So submit this change unbreak autopush for now, and I can look into why this is and switch to json?

pradh (approved):

server/services/bigtable.py
shifucun: sg!

PR #2855, opened 2023-06-23
Bio browser landing

rvguha:
Add landing page for bio browser


server/routes/browser/html.py
```diff
@@ -32,6 +32,9 @@
 def browser_main():
   return lib_render.render_page("browser/landing.html", "browser_landing.html")
 
+@bp.route('/bio')
+def browser_main():
```
shifucun, line 36: Please change `browser_main` to `bio_browser_main` (since this func is defined already)

And in render_page, the second argument should be "" (this is used for custom DC)

rvguha: Could you take another look?

shifucun (approved):

PR #2856, opened 2023-06-23
Added js templates to sdg static example page

dwnoble:


juliawu (approved):
Thanks for templatizing!

PR #2860, opened 2023-06-24
[Experimental] Add stand-alone example story page

juliawu:
Add a standalone html page demonstrating how our webcomponents can be integrated into a simple webpage.

Replit: https://replit.com/@JuliaWu7/SDG-Stories-Page


dwnoble (approved):
Thank you!

PR #2864, opened 2023-06-25
Added datacommons-map deprecated attributes for backwards compatibility

dwnoble:
Added back `datacommons-map` attributes `placeDcid`, `statVarDcid`, and `enclosedPlaceType` for backwards compatibility. 

@pradh FYI


shifucun (approved):
nice!

PR #2865, opened 2023-06-26
Temporary/quick e2e test against dev for nodejs

pradh:
The test_data files were copy/pasted from https://github.com/datacommonsorg/website/tree/master/server/integration_tests/test_data/screenshot. The only changes were that there are now multiple src links (improvement).


chejennifer (approved):
thanks for adding!

PR #2870, opened 2023-06-26
Fix SDG story page issue

shifucun:
* Fix ranking component import
* update hidden class to use height
* Show various column layout
* Update example country, some minor text change


static/js/ranking/component.ts
```diff
@@ -0,0 +1,27 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import React from "react";
+import ReactDOM from "react-dom";
+
+import { Page, RankingPagePropType } from "./ranking_page";
+
+export const renderRankingComponent = (
```
chejennifer, line 22: maybe add to here instead for consistency: https://github.com/datacommonsorg/website/blob/master/static/library/components.ts

chejennifer (commented):
thanks!

chejennifer (approved):

static/js/ranking/component.ts
shifucun: Added comment to clarify this is also used to render the internal ranking page.

PR #2877, opened 2023-06-27
[website revamp] add NL features needed for the homepage

chejennifer:
- add &d=1 flag which will disable the delay when autorunning a query on NL Interface. This way, when user enters a query in the homepage search and clicks submit, the result will be loaded without delay
- add &topic= flag which will show a list of example queries for a topic. Note: these examples I've added are just to show what this behavior looks like & needs to be updated to actual queries
- also fix click on datacommons in the nav bar to go back to old hompage instead of the new one

screen-recording-_98_


static/js/apps/nl_interface/query_examples.tsx
```diff
@@ -0,0 +1,105 @@
...
+    "How do housing prices in California compare with New York",
+    "What percentage of homes are rentals in USA",
+    "How old are the homes in counties in USA",
+  ],
+  sustainability: [
+    "Where were there fires in the US in the last year",
+    "What is the projected max temperature across the world",
+    "How does air pollution correlate with asthma prevalence around the world",
+  ],
+  biomedical: [
```
pradh, line 68: This is health?

static/js/apps/nl_interface/query_examples.tsx
```diff
@@ -0,0 +1,105 @@
...
+/**
+ * Component for a list of example queries
+ */
+
+import _ from "lodash";
+import React from "react";
+
+import { getUrlToken } from "../../utils/url_utils";
+
+// Map of topic to list of example queries
```
pradh, line 26: How did you come up with these queries?  something like "what is the main farming method used" we don't have data for

pradh (commented):
nice

static/js/apps/nl_interface/query_examples.tsx
chejennifer: this is following the topics used in the mock: https://screenshot.googleplex.com/7yRNh3s3Xi5tBAW

All the text is subject to being changed & I would think of all text as just placeholders to see the flow/behavior. Can change if you think that's needed now though

static/js/apps/nl_interface/query_examples.tsx
chejennifer: These are just random queries I thought of off the top of my head to show what the UI would look like/what the flow of actions would be. If there are specific queries you want me to take out or add, I can make those changes.

static/js/apps/nl_interface/query_examples.tsx
pradh: got it, can change later then.

shifucun (approved):

PR #2882, opened 2023-06-28
Extend flask cache expiration from 1 day to 7 days

shifucun:
After a deployment, only branch cache is updated which can cause data update. Since only covid is in branch cache and it's of low interest, we would increase the cache expiration from 1 day to 7 days. This would improve the latency a lot.


pradh (approved):
Nice refactor!

PR #2884, opened 2023-06-29
[Library] Add date attribute to map component

juliawu:
Allow users to specify a specific date to display on a map component.

Example usage:
```
<datacommons-map
    title="Population"
    place="country/USA"
    childPlaceType="State"
    variable="Count_Person"
    date="2010"
></datacommons-map>
```


shifucun (approved):
thanks for quick change!

PR #2885, opened 2023-06-29
LLM: Add integration test covering LLM flow

pradh:
Also, return debug info from nodejs server.


server/app_env/integration_test.py
```diff
@@ -19,4 +19,5 @@ class Config(_base.Config):
   INTEGRATION = True
   API_ROOT = 'https://autopush.api.datacommons.org'
   SECRET_PROJECT = 'datcom-website-dev'
-  SCHEME = 'http'
\ No newline at end of file
+  SCHEME = 'http'
```
jehangiramjad, line 22: Leaving a random comment for testing

jehangiramjad (commented):
Leaving a random comment

pradh (commented):
I have reverted the LLM fallback changes - based on Guha discussion, we want to do it slightly differently (follow on PR).

Now this PR is just a small integration test addition.

jehangiramjad (approved):

pradh:
Thanks!

PR #2888, opened 2023-06-29
NL: Query generation scripts

pradh:


tools/nl/querygen/README.md
```diff
@@ -0,0 +1,17 @@
+# Query Generation
```
jehangiramjad, line 1: nit: It would be great to know why this is needed? :) As in, why are we using seeds for rewriting by PaLM?

jehangiramjad (approved):

tools/nl/querygen/README.md
pradh: Good point, added!

pradh (commented):
Thanks for the review!

PR #2889, opened 2023-06-30
LLM: Use JSON5 to parse LLM response

pradh:
JSON5 is a lot more lenient in parsing JSON strings that LLM returns.

The docs say JSON5 can be 1000-3000x slower.  I did some benchmarking ([code](https://gist.github.com/pradh/63e5b9085505a7906062575314d1365a)), it is ~600x slower.

### DIFF
--- a/server/lib/nl/detection/palm_api.py
+++ b/server/lib/nl/detection/palm_api.py
@@ -19,6 +19,7 @@ import time
 from typing import Dict
 
 from flask import current_app
+import json5
 import requests
 
 from server.lib.nl.common import counters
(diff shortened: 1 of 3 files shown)


server/lib/nl/detection/palm_api.py
```diff
@@ -91,8 +92,11 @@ def parse_response(query: str, resp: Dict, ctr: counters.Counters) -> Dict:
       return {}
 
     try:
-      ans_json = json.loads(ans)
-    except json.decoder.JSONDecodeError as e:
+      # Use json5 to load since it is a lot more lenient (e.g.,
```
shifucun, line 95: can put some absolute latency number here? how many seconds for 600x?

shifucun (approved):

server/lib/nl/detection/palm_api.py
pradh: Done.

Here's my benchmark result: https://gist.github.com/pradh/63e5b9085505a7906062575314d1365a#file-benchmarkjson5-py-L1-L5

PR #2891, opened 2023-06-30
[Library] Add Lollipop component

juliawu:
This PR:
* Adds the ability to render a bar tile in lollipop style
* Wraps the lollipop style bar tile into a lollipop component
* Adds some examples
* Updates node server with lollipop changes

Because so much of the logic needed to render a lollipop chart is shared with a bar chart, I opted to add lollipop functionality to our existing bar tiles, instead of creating a separate lollipop tile. If requested, I can also amend this PR to add a separate lollipop tile instead.

Examples:
```
<datacommons-lollipop
  title="Population in States of United States"
  place="country/USA"
  childPlaceType="State"
  variable="Count_Person"
  useLollipop=true
></datacommons-lollipop>
<datacommons-lollipop
  title="Population in Alaska, Alabama, and Arizona"
  variable="Count_Person"
></datacommons-lollipop>
<datacommons-lollipop
  title="Median income by gender for Alaska, Alabama, and Arizona"
  comparisonVariables='["Median_Income_Person_15OrMoreYears_Male_WithIncome", "Median_Income_Person_15OrMoreYears_Female_WithIncome"]'
></datacommons-lollipop>
```

Screenshot 2023-06-30 at 8 31 48 AM


static/library/lollipop_component.ts
```diff
@@ -0,0 +1,117 @@
...
+ * ></datacommons-lollipop>
+ *
+ * <!-- Show a lollipop chart of population for specific US states -->
+ * <datacommons-lollipop
+ *      title="Population of US States"
+ *      variableDcid="Count_Person"
+ *      comparisonPlaces='["geoId/01", "geoId/02"]'
+ * ></datacommons-lollipop>
+ */
+@customElement("datacommons-lollipop")
```
dwnoble, line 49: Should we combine this component with `datacommons-bar` ? 

maybe:

```html
<datacommons-bar
  title="Population of US States"
  place="country/USA"
  childPlaceType="State"
  variable="Count_Person"
  lollipop
></datacommons-bar>
```

static/js/chart/draw.ts
```diff
@@ -937,6 +937,155 @@ function drawGroupBarChart(
...
+    .rangeRound([0, x0.bandwidth()])
+    .paddingOuter(0.05);
+
+  // Update and redraw the y-axis based on the new x-axis height.
+  y.rangeRound([chartHeight - bottomHeight, MARGIN.top]);
+  tempYAxis.remove();
+  addYAxis(yAxis, chartWidth, y, formatNumberFn, TEXT_FONT_FAMILY, unit);
+  updateXAxis(xAxis, bottomHeight, chartHeight, y);
+
+  const colorFn = getColorFn(statVars);
```
dwnoble, line 1028: Looks like some duplicate logic from `drawGroupBarChart` - is there a good way to abstract this?

juliawu: As per offline discussion, refactored the bar vs lollipop sections into separate helper functions.

static/library/lollipop_component.ts
juliawu: Good idea, done as you proposed above.

experimental/datacommons-js/example.html
```diff
@@ -66,5 +66,32 @@ <h3>Line</h3>
       place="country/USA"
       variables='["Count_Person"]'
     ></datacommons-line>
+    <h2>Lollipop</h2>
+    <datacommons-bar
+      title="Population in States of United States"
+      place="country/USA"
+      childPlaceType="State"
+      variable="Count_Person"
+      lollipop="true"
```
dwnoble, line 75: you can drop the `="true"` part here- and in fact `="false"` i dont think will work as expected. Just the presence of the `lollipop` attribute with no value should set the flag

static/library/bar_component.ts
```diff
@@ -45,6 +45,14 @@ import { convertArrayAttribute } from "./utils";
  *      variableDcid="Count_Person"
  *      comparisonPlaces='["geoId/01", "geoId/02"]'
  * ></datacommons-bar>
+ *
+ * <!-- Show a bar chart of population for specific US states with lollipops -->
+ * <datacommons-bar
+ *      title="Population of US States"
+ *      variableDcid="Count_Person"
+ *      comparisonPlaces='["geoId/01", "geoId/02"]'
+ *      lollipop="true"
```
dwnoble, line 54: ditto

static/library/bar_component.ts
```diff
@@ -80,6 +88,11 @@ export class DatacommonsBarComponent extends LitElement {
   @property({ type: Array<string>, converter: convertArrayAttribute })
   comparisonVariables;
 
+  // Optional: Whether to render as a lollipop
+  // Set to true to render using lollipops instead of bars
+  @property()
```
dwnoble, line 93: change to `@property({type: Boolean})` - otherwise i think Lit will actually make this a string (even though it's typed as a boolean)

static/library/lollipop_component.ts
```diff
@@ -0,0 +1,117 @@
+/**
```
dwnoble, line 1: delete file?

dwnoble (changes requested):

static/library/lollipop_component.ts
juliawu: Done!

static/library/bar_component.ts
juliawu: Thanks for the pointer! I think this is why I had to set lollipop="true" with the string above originally. Fixed.

static/library/bar_component.ts
juliawu: Done!

experimental/datacommons-js/example.html
juliawu: Done!

static/js/chart/draw.ts
```diff
@@ -904,7 +1011,8 @@ function drawGroupBarChart(
   chartHeight: number,
   dataGroups: DataGroup[],
   formatNumberFn: (value: number, unit?: string) => string,
-  options?: ChartOptions
+  options?: ChartOptions,
+  useLollipop?: boolean
```
dwnoble, line 1015: can you embed this to the above options object?

dwnoble (approved):

static/js/chart/draw.ts
juliawu: Good idea. Done!

PR #2892, opened 2023-06-30
NL: Add more per-chart emojis

pradh:
Intended meaning:

1. "This chart is not relevant for the query" -- 👎
2. "Somewhat relevant but not quite right (eg wrong place type)" -- ⚠️
3. "chart should be promoted" -- ⬆️
4. "chart should be demoted" -- ⬇️
5. "god, this is so embarrassing" -- 🤦

Example:

image


static/js/apps/nl_interface/query_result.tsx
```diff
@@ -149,7 +149,9 @@ export const QueryResult = memo(function QueryResult(
           </a>
           {chartsData && chartsData.sessionId && (
             <span
-              className={`thumb-down ${isThumbClicked ? "thumb-dim" : ""}`}
+              className={`feedback-emoji ${
+                isThumbClicked ? "feedback-emoji-dim" : ""
```
juliawu, line 153: Is the intended behavior for query_result to be specific to thumbs down?

If not, let's use `isEmojiClicked` instead and update the function on line 200 to onChartClick to match nl_feedback.tsx

static/js/utils/nl_interface_utils.ts
```diff
@@ -30,6 +30,10 @@ const DEBUG_INFO_PARAM_PREFIX = "&entry.1280679042=";
 
 export const CHART_FEEDBACK_SENTIMENT = {
   THUMBS_DOWN: "THUMBS_DOWN",
+  WARNING: "WARNING",
```
juliawu, line 33: nit: let's add quick comments here explaining what feedback each emoji means

juliawu (commented):
Thanks!

static/js/apps/nl_interface/query_result.tsx
pradh: Yes, thumbs-down for now.

static/js/utils/nl_interface_utils.ts
pradh: Done.

pradh (commented):
Thanks for the review!

juliawu (approved):
Thanks for the changes!

PR #2893, opened 2023-06-30
NL: Change query-level emoji to a warning

pradh:
So that will be used for when the result is empty, and when there are chart results, feedback will be on the chart emojis.

image


juliawu (approved):
Nice!

pradh:
Thanks!

PR #2897, opened 2023-07-03
[library] Added stacked bar chart support to datacommons-bar web component

dwnoble:
Screenshot:

Screenshot 2023-07-03 at 1 58 37 PM

### DIFF
--- a/experimental/datacommons-js/example.html
+++ b/experimental/datacommons-js/example.html
@@ -30,7 +30,7 @@
       title="Median income by gender"
       comparisonVariables='["Median_Income_Person_15OrMoreYears_Male_WithIncome", "Median_Income_Person_15OrMoreYears_Female_WithIncome"]'
       comparisonPlaces='["geoId/01", "geoId/02", "geoId/04"]'
-      stacked="false"
+      stacked
     ></datacommons-bar>
     <datacommons-bar
       title="Median income by gender for counties in Alaska"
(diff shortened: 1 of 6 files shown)


static/library/bar_component.ts
```diff
@@ -53,6 +53,12 @@ export class DatacommonsBarComponent extends LitElement {
     ${unsafeCSS(tilesCssString)}
   `;
 
+  /**
+   * Draw as a stacked chart instead of grouped chart
+   */
+  @property({ type: Boolean })
+  stacked?: boolean;
```
juliawu, line 60: Nit: Let's add an example of a call using stacked=true to the example usage comment doc

juliawu (approved):
Awesome!

experimental/datacommons-js/example.html
```diff
@@ -26,6 +26,12 @@ <h2>Bar Chart</h2>
       comparisonVariables='["Median_Income_Person_15OrMoreYears_Male_WithIncome", "Median_Income_Person_15OrMoreYears_Female_WithIncome"]'
       comparisonPlaces='["geoId/01", "geoId/02", "geoId/04"]'
     ></datacommons-bar>
+    <datacommons-bar
+      title="Median income by gender"
```
shifucun, line 30: nit: use double quote as the outer quote to make it consistent across args

shifucun (approved):
nice!

experimental/datacommons-js/example.html
dwnoble: Leaving this as is since we are planning on switching to space-separated values

static/library/bar_component.ts
dwnoble: Done!

PR #2904, opened 2023-07-07
NL: Support more map types, improve stop-words, etc.

pradh:
A few misc changes:

* Add support for census tracts as child type
* Allow map display on city, zip-code and tract child types, only for a US parent.  Add e2e tests for these
* Remove events from STOP_WORDS.  This is useful in also matching SVs corresponding to events (in case the event data is missing).
*  Add "districts" to the plural map.  You can see the improvement to the [districts of turkey with the highest fertility rate] query] e2e test query.
*  Minor: Fallback from "places-in" query-types to SIMPLE.

Screenshots:

image

image

image


server/lib/nl/common/utils.py
```diff
@@ -293,11 +293,21 @@ def has_map(place_type: any, places: List[types.Place]) -> bool:
...
+  if not places:
     return False
-  return places[0].country in constants.ADMIN_AREA_MAP_COUNTRIES
+
+  aatype = constants.ADMIN_DIVISION_EQUIVALENTS.get(place_type, None)
+  if aatype and places[0].country in constants.ADMIN_AREA_MAP_COUNTRIES:
+    return True
+
+  # If the parent place is in USA, check that the child type +
+  # parent type combination is supports map.
```
jehangiramjad, line 305: nit: typo?

jehangiramjad (approved):
Nice!

server/lib/nl/common/utils.py
pradh: Thanks, done!

PR #2906, opened 2023-07-07
[Library] Added basic slider web component and publish/subscribe pattern

dwnoble:
Screenshots (note texas population change on different dates)

texas1985
texas2020


static/library/slider_component.ts
```diff
@@ -0,0 +1,133 @@
...
+ *
+ * Example usage:
+ *
+ * <!-- Date slider example  -->
+ * <datacommons-slider
+ *      max="2023"
+ *      min="1950"
+ *      publish="dc-year"
+ *      value="2023"
+ * ></datacommons-bar>
```
juliawu, line 35: s/bar/slider

experimental/datacommons-js/example.html
```diff
@@ -52,6 +52,20 @@ <h2>Map</h2>
...
+      childPlaceType="State"
+      subscribe="dc-map"
+      variable="Count_Person"
+    ></datacommons-map>
+    <datacommons-slider
+      max="2023"
+      min="1950"
+      publish="dc-map"
+      value="2023"
+    ></datacommons-slider>
```
juliawu, line 68: I think "publish" and "subscribe" are not intuitive for people with limited coding backgrounds. Perhaps we can rename them to something like:
```
<datacommons-slider
...
name="time-slider-1"
></datacommons-slider>

<datacommons-map>
...
controlledBy="time-slider-1"
></datacommons-map>
```
@dwnoble @shifucun @pradh thoughts?

static/library/slider_component.ts
```diff
@@ -0,0 +1,133 @@
...
+  value: number;
+
+  private onSliderChange(e: Event) {
+    const target = e.currentTarget as HTMLInputElement;
+    const value = target.value;
+    this.value = Number(value);
+    this.dispatchEvent(
+      new CustomEvent<ChartEventDetail>(this.publish, {
+        detail: {
+          property: "date",
```
juliawu, line 105: For my own understanding, do we think we'll extend this slider to other properties beyond date?

juliawu (commented):
This is very cool!

static/library/slider_component.ts
dwnoble: That was my thinking - down the road we can use this to set any arbitrary property

static/library/slider_component.ts
dwnoble: Good catch

experimental/datacommons-js/example.html
dwnoble: That is fair about using the names `publish` / `subscribe`. I don't think `name`/`controlledBy` is optimal either though- there could be a many-to-many relationship between visualization components and controls. for example, a `datacommons-map` could be controlled by both a time slider and a to-be-created country picker. In this case the time slider and country picker would both broadcast on the same channel.

Alternative possible names
`publish`: `publishEvent`,   `triggerEvent`, `emit`, `fire`, `broadcast`, `channel`,
`subscribe`: `listen`, `on`, `handle`, `channel`

Another thought: Users can trigger these events directly from their own JS code and have a datacommons web component listen for them. We should balance making it clear what the functionality is for JS-savvy users, and that these attributes are JS event names for more advanced users.

juliawu: Sounds good. Happy with going with publish/subscribe for now, we can update if we think there's a better naming convention.

juliawu (approved):
Thanks for changes!

PR #2912, opened 2023-07-10
[Library] Accept comma and/or space separated lists for attribute values

juliawu:
This PR:
*  Allows comma and/or space separated lists as valid syntax for Array<String> type attribute values in web components. For backwards compatibility, providing values in JSON format is still supported.
* Updates the examples in documentation to use a comma/space separated list instead of JSON.

With this change, the following syntaxes are all valid (see `comparisonPlaces`):

Space separated
```
<datacommons-bar
  title="Population of select US states"
  variable="Count_Person"
  comparisonPlaces="geoId/01 geoId/02 geoId/04 geoId/05"
></datacommons-bar>
```

JSON list with double quotes
(Single quotes is not valid JSON)
```
<datacommons-bar
  title="Population of select US states"
  variable="Count_Person"
  comparisonPlaces='["geoId/01", "geoId/02", "geoId/04", "geoId/05"]'
></datacommons-bar>
```


static/library/utils.ts
```diff
@@ -17,14 +17,24 @@
...
+ *
  * @param attributeValue the attribute value provided to the web component
- * @returns JSON parsed list of strings
+ * @returns A string array of attribute values
  */
 export function convertArrayAttribute(attributeValue: string): string[] {
-  // JSON.parse requires double quotes for strings within objects.
-  const cleanedAttributeValue = attributeValue.replaceAll("'", '"');
-  return JSON.parse(cleanedAttributeValue);
+  const openingBrackets = ["(", "[", "{"];
```
dwnoble, line 30: Just use `[` here - valid json can't start with `(`, and an object starting with `{` won't have the return type of `string[]`

dwnoble (changes requested):
Thanks Julia! I vote to drop support for the single quoted JSON entirely.

dwnoble:
I'd also be up for dropping support for comma-separated values and just allow for space-separated and JSON-array syntax (for backwards compatibility). @shifucun wdyt?

shifucun:
> I'd also be up for dropping support for comma-separated values and just allow for space-separated and JSON-array syntax (for backwards compatibility). @shifucun wdyt?

Only supporting space separated values sg!

juliawu:
@dwnoble @shifucun Just to make sure: do we anticipate we'll expand instances where valid spaces will be tricky? For example, space separation wouldn't work for something like `locationNames="Martha's Vineyard Half Moon Bay"` for locations "Martha's Vineyard" and "Half Moon Bay".

This might come into play in getting human-readable stat var names (`statVarDisplayName="Gini Index of Economic Activity"`). Leaning towards comma separated values for this reason.

static/library/utils.ts
juliawu: Good call out, switched to just `[`.

static/library/utils.ts
```diff
@@ -17,14 +17,24 @@
...
+  if (openingBrackets.includes(attributeValue.charAt(0))) {
+    // Parse as JSON if attribute value begins with a bracket
+    // JSON.parse requires double quotes for strings within objects
+    const cleanedAttributeValue = attributeValue.replaceAll("'", '"');
```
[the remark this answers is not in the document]
juliawu: Done. Now only supports double quotes.

juliawu:
Per offline discussion, moved to supporting only space separated values and double-quoted JSON. PR description updated to reflect this.

dwnoble (approved):
Nice! thanks julia!

PR #2914, opened 2023-07-10
[map] update getting child types for map explorer

chejennifer:
- Have a list of places that have AA1/AA2 child place type enabled so that it will be easy to enable AA1/AA2 for places in the future
- Make &ept= param take priority as the place type used in the map explorer


dwnoble (approved):
Nice!

chejennifer:
thanks for the review!

PR #2924, opened 2023-07-12
NL: fix minor bugs in prior PR

pradh:
https://github.com/datacommonsorg/website/pull/2923

### DIFF
--- a/static/js/apps/nl_interface/query_result.tsx
+++ b/static/js/apps/nl_interface/query_result.tsx
@@ -174,39 +174,65 @@ export const QueryResult = memo(function QueryResult(
           {debugData && (
             <DebugInfo
               debugData={debugData}
-              pageConfig={chartsData ? chartsData.config : null}
+              chartsData={chartsData}
             ></DebugInfo>
           )}
+          {chartsData &&
+            !chartsData.placeFallback &&
+            (chartsData.placeSource === "PAST_QUERY" ||
+              chartsData.placeSource === "PAST_QUERY") && (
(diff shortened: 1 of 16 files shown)


shifucun (approved):

static/js/apps/nl_interface/query_result.tsx
```diff
@@ -128,7 +128,18 @@ export const QueryResult = memo(function QueryResult(
...
         } else {
-          setErrorMsg("Sorry, we couldn't answer your question.");
+          // If there was no place recognized, we might end up with 0
+          // categories.  Check if that's the case, and provide a different
+          // error message.
+          if ("placeSource" in resp.data && resp.data["placeSource"]) {
+            setErrorMsg(
+              "Could not recognize any place in the query! " +
+                "If you included a place, it is possible " +
+                "we don't recognize it yet!"
```
jehangiramjad, line 138: nit: this sounds like we have a a set of those places we recognize vs those that we don't etc. That could be a bit problematic to communicate to users. Could we simply state that "we failed to recognize it" (which is also an accurate reflection)?

jehangiramjad (approved):
Only a nitty comment

static/js/apps/nl_interface/query_result.tsx
pradh: Ok, done.  I wasn't sure if we wanted to suggest adding a place, but we can revisit.

pradh (commented):
Thanks!

PR #2927, opened 2023-07-12
[node server] switch node server to use chart url

chejennifer:
- add new /nodejs/chart endpoint on node server that takes information about a chart and returns that chart
- update /nodejs/query to return a url to /nodejs/chart for each tile instead of returning the actual svg. i.e., in the response object, replace the `svg` field with `chartUrl`. This is behind a flag &chartUrl=1

screen-recording-2023-07-12T102022 847
screen-recording-2023-07-12T102126 924
screen-recording-2023-07-12T102110 422

Additional change:
- updated parent place api call to take dcid as a req parameter instead of part of the request path to fix server side forgery vulnerability


pradh:
Very cool!  Can we do the switch based on a flag?  So if we roll this out to dev it won't break and we can test out and update?

static/src/server.ts
```diff
@@ -385,6 +480,32 @@ app.get("/nodejs/query", (req: Request, res: Response) => {
...
+  );
+  // Need to convert encoded # back to #.
+  const eventTypeSpecVal = (
+    req.query[CHART_URL_PARAMS.EVENT_TYPE_SPEC] as string
+  ).replaceAll("%23", "#");
+  const eventTypeSpec = JSON.parse(eventTypeSpecVal);
+  const tileConfig = JSON.parse(
+    req.query[CHART_URL_PARAMS.TILE_CONFIG] as string
+  );
+  res.setHeader("Content-Type", "text/html");
```
dwnoble, line 499: Continued: this could return `image/svg+xml`

chejennifer:
> Very cool! Can we do the switch based on a flag? So if we roll this out to dev it won't break and we can test out and update?

Yes, good idea, I will make that change now!

static/nodejs_server/utils.ts
```diff
@@ -52,7 +58,49 @@ export function getProcessedSvg(chartSvg: SVGSVGElement): string {
     node.setAttribute("font-family", FONT_FAMILY);
     node.setAttribute("font-size", FONT_SIZE);
   });
-  // Get and return the svg as an xml string
+  // Create and return an image element out of the chart svg.
   const svgXml = xmlserializer.serializeToString(chartSvg);
-  return "data:image/svg+xml," + encodeURIComponent(svgXml);
+  const img = document.createElement("img");
+  img.src = "data:image/svg+xml," + encodeURIComponent(svgXml);
+  return img.outerHTML;
```
dwnoble, line 65: Instead of returning an SVG wrapped in an image, why not return a proper SVG directly? Then the consumer could either use the SVG directly or wrap it in an image themselves if they wanted to

static/src/server.ts
```diff
@@ -271,6 +298,71 @@ function getDisasterBlockTileResults(
...
 }
 
+// Get the chart html for a tile
+function getTileChart(
+  tileConfig: TileConfig,
+  placeDcid: string,
+  childPlaceType: string,
+  svSpec: StatVarSpec[],
+  eventTypeSpec: Record<string, EventTypeSpec>
+): Promise<string> {
```
dwnoble, line 308: Instead of returning `strings`, I think it would be more clear to return a`HTMLElement` (or `SVGElement`), and only convert to string right before returning to the user (in the `app.get` handler)

static/nodejs_server/utils.ts
chejennifer: done

static/src/server.ts
chejennifer: returning image/svg+xml seems to cause a weird error .. and i think returning html here is ok?

dwnoble (approved):
Thanks Jennifer!

server/routes/shared_api/place.py
```diff
@@ -339,9 +339,10 @@ def child_fetch(parent_dcid):
   return result
 
 
-@bp.route('/parent/<path:dcid>')
+@bp.route('/parent')
 @cache.cache.memoize(timeout=cache.TIMEOUT)
```
shifucun, line 343: need "query_string=True" for caching?

chejennifer: good catch, done

static/nodejs_server/bar_tile.ts
```diff
@@ -95,8 +121,54 @@ export async function getBarTileResult(
       ),
       type: "BAR",
     };
+    if (useChartUrl) {
```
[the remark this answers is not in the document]
chejennifer: done

chejennifer:
thanks for the reviews!

PR #2928, opened 2023-07-12
Added useDrawOnResize to map tile. 

dwnoble:
This fixes a bug where the map tile wouldn't render properly if it was constructed while hidden from view


juliawu (approved):
Thanks for the fix!

PR #2933, opened 2023-07-12
Bumped nl default chart height to 200

dwnoble:
... to account for map tile no longer defaulting to a height of 200


juliawu (approved):
LGTM

pradh (approved):
Thanks Dan!

PR #2934, opened 2023-07-12
[library] Web component parameter refactoring

dwnoble:
* Renamed `comparisonVariables` to `variables`. 
* Renamed `comparisonPlaces` to `places`. 
  * In `datacommons-map`, marked `place` as `@deprecated` rather than removing it to avoid breaking existing partners.
* Renamed `place` to `parentPlace` where appropriate.

### DIFF
--- a/web-components/docs/README.md
+++ b/web-components/docs/README.md
@@ -8,7 +8,10 @@ must be included for the web components to work.
 
 ```html
 <link rel="stylesheet" href="https://datacommons.org/css/ranking.min.css" />
-<link rel="stylesheet" href="https://datacommons.org/css/nl_interface.min.css" />
+<link
+  rel="stylesheet"
+  href="https://datacommons.org/css/nl_interface.min.css"
+/>
 <script src="https://datacommons.org/datacommons.js"></script>
 ```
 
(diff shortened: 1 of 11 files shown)


dwnoble:
Note: Let's remove `experimental/datacommons-js` in a future PR

web-components/docs/README.md
```diff
@@ -443,21 +454,21 @@ Show a ranking of US States by population, highest to lowest
...
-    title="US States with the Lowest Population"
-    place="country/USA"
-    childPlaceType="State"
-    variable="Count_Person"
-    showLowest=true
+  title="US States with the Lowest Population"
+  parentPlace="country/USA"
+  childPlaceType="State"
+  variable="Count_Person"
+  showLowest="true"
```
juliawu, line 472: Note for later: We should get the ranking component to also allow just `showLowest`, without having to set `="true"`

juliawu (approved):
Thanks for the changes! This naming is much cleaner.

web-components/docs/README.md
dwnoble: Good catch! I went ahead and added a fix for this in this PR

PR #2935, opened 2023-07-13
NL: Improve fallbacks

pradh:
1. Switch to having a default place, but report that we are using a default place
2. Propagate specific message for comparison fallback, which has multiple places
3. Generally simplify comparison flow to *only* use the most recent place(s)
4. Disable fallback for default place

Wrote up the [fallback matrix here](https://docs.google.com/document/d/13-1zsYZm_RFkzhKb4GxVy7DQNU8ZiEYZSY8001R-GsU/edit).

Some screenshots:

[agriculture workers]
image

[GDP of China and USA] -> [compare with India]
image

[GDP of China and India] -> [compare fertility rate instead]
image

### DIFF
--- a/server/lib/nl/common/utterance.py
+++ b/server/lib/nl/common/utterance.py
@@ -95,20 +95,20 @@ class FulfillmentResult(str, Enum):
   """Where is a variable or place from?"""
   # From the current user provided query.
   CURRENT_QUERY = "CURRENT_QUERY"
-  # From a prior query in the context.
-  # Importantly, no variable was recognized in the current query.
+  # From a past user query in the context.
+  # Importantly, no place/variable was recognized in the current query.
   PAST_QUERY = "PAST_QUERY"
-  # Subset of the places/variables used are from a past query.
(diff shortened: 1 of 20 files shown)


server/lib/nl/common/utterance.py
```diff
@@ -93,10 +93,22 @@ class ChartType(IntEnum):
 
 class FulfillmentResult(str, Enum):
   """Where is a variable or place from?"""
+  # From the current user provided query.
   CURRENT_QUERY = "CURRENT_QUERY"
+  # From a past user query in the context.
+  # Importantly, no place/variable was recognized in the current query.
```
jehangiramjad, line 99: What does this comment mean? Is ``PAST_QUERY`` only set if no place/variable was recognized int he current query?

jehangiramjad (approved):
Thanks!

server/lib/nl/common/utterance.py
pradh: Correct.

pradh (commented):
Thanks Jehangir!

PR #2937, opened 2023-07-13
NL: For place-fallback and default-place hide charts by default

pradh:
Also, refactor commentary logic into a separate component.

hide1

hide2


static/js/apps/nl_interface/nl_commentary.tsx
```diff
@@ -0,0 +1,156 @@
...
+
+import { SearchResult } from "../../types/app/nl_interface_types";
+
+interface NLCommentaryPropType {
+  chartsData: SearchResult;
+  hideCharts: boolean;
+  setHideCharts: (v: boolean) => void;
+}
+
+export function shouldHideCharts(respData: any): boolean {
```
shifucun, line 31: should this function reside in a util module?

static/js/apps/nl_interface/nl_commentary.tsx
```diff
@@ -0,0 +1,156 @@
...
+}
+
+//
+// For more context, refer:
+//   https://docs.google.com/document/d/13-1zsYZm_RFkzhKb4GxVy7DQNU8ZiEYZSY8001R-GsU/edit
+//
+export function NLCommentary(props: NLCommentaryPropType): JSX.Element {
+  return (
+    <>
+      {!props.chartsData.placeFallback &&
```
shifucun, line 49: strongly recommend to have a few child component like:

```
export function A1(props: chartsData): JSX.Element {
  showA := <complicated && ||>
  return <>(showA? <div/>)</>
}
A2...
A3...
```

Then can have meaningful name for A1, A2..

static/js/apps/nl_interface/nl_commentary.tsx
pradh: The logic is very specific to what's going on in this file, so keeping it here instead of a general util file?

static/js/apps/nl_interface/nl_commentary.tsx
pradh: Thanks for the offline suggestion to factor the "control" flow separate from the display strings. PTAL now.

shifucun (approved):
very nice!

PR #2942, opened 2023-07-13
Removed search bar from new homepage. Updated topic categories

dwnoble:
* Made the new homepage the default in autopush and local.
* Old homepage accessible via `/old` in autopush/local
* Note: the vscode prettier plugin made a bunch of format changes to hex colors

Screenshot:

Screenshot 2023-07-13 at 4 28 50 PM


server/routes/static.py
```diff
@@ -30,9 +30,15 @@
...
+    return lib_render.render_page(
+      "static/homepage.html",
+      "homepage.html",
+      topics=current_app.config.get('HOMEPAGE_TOPICS', []),
+      partners=json.dumps(current_app.config.get('HOMEPAGE_PARTNERS', [])))
   return lib_render.render_page("static/homepage_old.html", "homepage.html")
 
-
 @bp.route('/new')
 def homepage_new():
```
chejennifer, line 43: we can remove this endpoint now?

chejennifer (commented):
looks nice!

server/templates/static/homepage.html
```diff
@@ -31,12 +31,21 @@
...
  {% block content %}
  <section id="homepage-top">
-  <div id="search-container"></div>
+  <div class="container">
+    <div class="header">Public data tells incredible stories</div>
+    <div class="sub-header">
+      Working with public data doesn't have to be painful.<br/>
+      Data Commons is a free, unified database of over
+      <a href="https://docs.datacommons.org/datasets/" target="_blank" rel="noopener noreferrer">
+        175K public datasets
```
pradh, line 40: Not 175K public datasets :)

We can use a qualitative sentence perhaps since there are different ways to quantify the size (# variables, # sources, # datasets, # series, # data points)...

pradh (approved):
nice!

server/templates/static/homepage.html
dwnoble: Removed

server/routes/static.py
dwnoble: Good call. Removed.

pradh:
> 

nit:  As a non-design-savvy-layperson person, the "Public data tells incredible stories" punchline occupies tad too much screen space, and the topic cards could perhaps use more focus?

chejennifer (approved):
thanks!

dwnoble:
> > 
> 
> nit: As a non-design-savvy-layperson person, the "Public data tells incredible stories" punchline occupies tad too much screen space, and the topic cards could perhaps use more focus?

Agreed I might have gotten a little carried a way. I dropped the font size a bit, but left it relatively large. ill get Ido to weigh in - i think more important than the topic cards is giving some clear messaging about what datacommons is (not saying i accomplish that with the current language).

PR #2943, opened 2023-07-13
[Library] Add legend to gauge tile / component

juliawu:
Adds a legend to the gauge tile, so that it matches the style of the other charts

Example:
Screenshot 2023-07-13 at 4 44 59 PM


dwnoble (approved):
Nice!

PR #2944, opened 2023-07-14
[nl interface] update the feedback form being used

chejennifer:
update nl interface to use this shorter feedback form instead that just asks user for their feedback: https://forms.gle/ADxVJzmjid3ynBo26

### DIFF
--- a/static/js/utils/nl_interface_utils.ts
+++ b/static/js/utils/nl_interface_utils.ts
@@ -19,14 +19,14 @@
  */
 
 const FEEDBACK_LINK =
-  "https://docs.google.com/forms/d/e/1FAIpQLSfqndIayVhN1bN5oeZT0Te-MhhBMBR1hn97Lgr77QTOpga8Iw/viewform?usp=pp_url";
+  "https://docs.google.com/forms/d/e/1FAIpQLSe9SG0hOfrK7UBiOkQbK0ieC0yP5v-8gnQKU3mSIyzqdv6WaQ/viewform?usp=pp_url";
 // Param prefixes found when following the instructions here to get a prefilled
 // link: https://support.google.com/docs/answer/2839588?hl=en&ref_topic=6063592#zippy=%2Csend-a-form-with-pre-filled-answers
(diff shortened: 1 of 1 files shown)


static/js/utils/nl_interface_utils.ts
```diff
@@ -19,14 +19,14 @@
  */
 
 const FEEDBACK_LINK =
-  "https://docs.google.com/forms/d/e/1FAIpQLSfqndIayVhN1bN5oeZT0Te-MhhBMBR1hn97Lgr77QTOpga8Iw/viewform?usp=pp_url";
+  "https://docs.google.com/forms/d/e/1FAIpQLSe9SG0hOfrK7UBiOkQbK0ieC0yP5v-8gnQKU3mSIyzqdv6WaQ/viewform?usp=pp_url";
```
pradh, line 22: 1) Will "What was the query" get auto-filled?  If so, move it after the "Thank you!".

2) Email - should we make it optional?

3) In the Feedback textbox, should we add a few words of text?  Like the Bard form.... concretely:

```
How would you rate the overall relevance of the charts to your question? 
- Not at all relevant
- Somewhat relevant
- Very relevant

If you have anything else to share on what we could fix or improve, please let us know.
[........]
```

pradh (commented):
Thanks!

static/js/utils/nl_interface_utils.ts
chejennifer: updated the form, let me know what you think: https://forms.gle/1ih4oPmKg2Q8Mg6A8

pradh: Thanks!

Make "If you have anything else to share on what we could fix or improve, please let us know." optional and should we still keep the email, but optional as 3rd question?

pradh (approved):

dwnoble (approved):
Thank you!

static/js/utils/nl_interface_utils.ts
chejennifer: done

Issue #2932, opened 2023-07-12
NL: bar chart with lots of variables is hard to read


chejennifer:
in this example: https://screenshot.googleplex.com/9yrFJwF2nwWSEKs

1. When the chart says "A compared with other variables", A doesn't appear in the legend, so it is hard to see how it compares to others.

2. The colours shades for bars are too close to find the matching legend. Maybe add a hover tooltip?

pradh:
The attached PR addresses (1).  Defer to you and @dwnoble for (2).

Issue #2925, opened 2023-07-12
NL: "electricity" recognized as CITY


pradh:
![image](https://github.com/datacommonsorg/website/assets/4375037/79cab5b8-4f2f-40aa-a340-5bfc090ac7fb)

Issue #2938, opened 2023-07-13
NL: should not show bar chart comparing very different SVs


pradh:
[Most profitable sources of energy production in Texas last year]

![image](https://github.com/datacommonsorg/website/assets/4375037/eb14018b-7e15-4d18-8112-ab9f74248a06)

PR #2945, opened 2023-07-14
NL: fix a few bugs

pradh:
1. Bug in extend_svs where we don't always check SV compatibility caused https://github.com/datacommonsorg/website/issues/2938

Screenshot 2023-07-13 at 7 54 21 PM

3. Fix https://github.com/datacommonsorg/website/issues/2932 by limiting the number of extended SVs we plot in a bar chart to 16, and also keep the "main SV" pinned as the first entry so its clear.

Screenshot 2023-07-13 at 7 52 53 PM

5. Fix https://github.com/datacommonsorg/website/issues/2925 by ensuring that contained-in doesn't match substrings


server/integration_tests/test_data/usa_map_types/query_2/chart_config.json
```diff
@@ -39,9 +39,24 @@
                 "tiles": [
                   {
                     "statVarKey": [
-                      "Median_Income_Household_HouseholderRaceWhiteAlone"
+                      "Median_Income_Household_HouseholderRaceAsianAlone"
                     ],
-                    "title": "Median Income of White Households in Census Tracts of Placer County (${date})",
+                    "title": "Median Income of Asian Households in Census Tracts of Placer County (${date})",
```
pradh, line 44: This and the next diff comes about because of the tweaks to the tracts maps in this PR.  We consume away "tract" from the query and match more SVs I think...

chejennifer (approved):
thanks for the fixes!

shifucun (approved):

PR #2946, opened 2023-07-14
NL: Start an e2e validator tool

pradh:
Please see included README.


tools/nl/validator/systest.py
```diff
@@ -0,0 +1,249 @@
...
+
+  p = cfg['metadata']['placeDcid'][0]
+  if p != pl:
+    ret['WrongPlace'] = '*'
+    counters['WrongPlace'] += 1
+
+  found_sv = False
+  for i, s in enumerate(
+      resp.get('debug', {}).get('sv_matching', {}).get('CosineScore', [])):
+    if s < 0.5:
```
jehangiramjad, line 169: why is this needed?

jehangiramjad (approved):
Thanks. Looks good

tools/nl/validator/systest.py
pradh: Since the backend completely ignores such SVs (i.e., they'll never show up on charts), excluding from the rank...those will get INF rank.

pradh (commented):
Thanks for the quick review Jehangir!

PR #2947, opened 2023-07-14
NL: Add proto support for query suggestions

pradh:
Just something quick for @dwnoble to experiment with.  Not hung up on proto names, and we can change in future PR.


shifucun (approved):

pradh:
Thanks!

PR #2950, opened 2023-07-17
[node server] update chart response format

chejennifer:
- for node server chart response, return the svg with the svg tag
Screenshot 2023-07-17 at 3 45 50 PM


pradh (approved):
Thanks!   Is there enclosing `<html><body>` tags in the response?

Other SVG files just have an <svg> tag.

```
<?xml version="1.0" standalone="no"?>
<svg xmlns="http://www.w3.org/2000/svg" version="1.1" width="147" height="147">
...
...
</svg>
```

chejennifer:
> Thanks! Is there enclosing `<html><body>` tags in the response?
> 
> Other SVG files just have an `<svg>` tag.
> 
> ```
> <?xml version="1.0" standalone="no"?>
> <svg xmlns="http://www.w3.org/2000/svg" version="1.1" width="147" height="147">
> ...
> ...
> </svg>
> ```

updated to: 

Screenshot 2023-07-17 at 4 44 29 PM

pradh:
Thanks!

pradh (approved):

PR #2952, opened 2023-07-18
NL: reject inappropriate words and improve schools

pradh:
* The initial bad-word list is located [here](https://storage.mtls.cloud.google.com/datcom-website-config/nl_bad_words.txt).

* Also, avoid school types from being considered as stop-words.  There were two other ripple effects to this change:
   1. It also requires special-handling for fallback logic (to not say "schools in sunnyvale" => "sunnyvale")
   2. It also requires not regressing the demo query [how big are public schools in sunnyvale]

Note: making the schools change also uncovered https://github.com/datacommonsorg/website/issues/2953.   This whole stop-word removal business needs streamlining!  Post fishfood maybe, and as part of fixing 2853.
 
Screenshot

image

### DIFF
--- a/nl_server/gcs.py
+++ b/nl_server/gcs.py
@@ -13,7 +13,6 @@
 # limitations under the License.
 
 BUCKET = 'datcom-nl-models'
-TEMP_DIR = '/tmp/'
 
 import os
 from pathlib import Path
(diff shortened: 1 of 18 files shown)


jehangiramjad (approved):
Thanks!

PR #2957, opened 2023-07-18
[node server] serve png instead of svg

chejennifer:
- have the endpoint /nodejs/chart serve a png instead of a svg

Screenshot 2023-07-18 at 10 04 03 AM


pradh (approved):
Nice!

Issue #2962, opened 2023-07-19
NL: ContainedInPlace should detect multiple types


pradh:
With DC place-recognition, [how many people are unemployed in zip codes of washington state] only consumes "washington" leaving out "state".  When we recognize place-types, we seem to recognize "state" and leave out "zip codes".

We should loop over multiple place types, and in this case a ContainedInPlace matching the place's type should be ignored.

PR #2963, opened 2023-07-19
NL: Switch NL place-detector default to `dc`

pradh:
Also, remove the check-box from frontend.  The diffs are pretty clean (except for an underlying bug - https://github.com/datacommonsorg/website/issues/2962).

Finally, check for bad-words also from punctuation-removed query.


jehangiramjad (approved):
Cool, thanks!

Spaceenter (approved):

PR #2978, opened 2023-07-20
update sdg topic page with new dcids

n-h-diaz:


pradh (approved):
Still looking good when testing locally?

PR #2980, opened 2023-07-20
NL: Fix a few bugs from query testing

pradh:
1. Allow DEFAULT_TYPE for country to start from county instead of state [prevalence of diabetes across US]
2. Fix a few queries and add a couple.
3. Add Guha's curated "no per-capita" sheet.
4. Load the SV titles as a config (long pending TODO)

Generally the tests pass, and a few pc SVs are dropped.


jehangiramjad (approved):
Thanks!

server/lib/util.py
```diff
@@ -179,6 +180,30 @@ def get_nl_disaster_config():
   return get_subject_page_config(filepath)
 
 
+# Returns disaster dashboard config for NL
```
jehangiramjad, line 183: why the disaster dashboard config?

pradh: sorry, copy/paste typo, fixed.

PR #2986, opened 2023-07-20
[tools revamp] add new visualization tool with initial set of selectors

chejennifer:
- create new app reachable at /tools/visualization
- implement the selection flow when a user arrives at the tool for the first time (does NOT include reading/writing to the url, editing options once the selection has been completed, updating selections when switching to a different visualization tool, these will be in next PRs)
- selections will be saved in a context & currently have a dummy component that reads this context and writes "plotting <selected options>" but this component will eventually take the context and draw the correct visualization (visualization components will receive information via props)

screen-recording-2023-07-20T153000 593


rvguha (approved):
lgtm

PR #2992, opened 2023-07-21
[library] Add support for multiple places in the line tile & web component

juliawu:
Adds the ability for the line chart to plot multiple places, and enables this in the web components.

More specifically, this PR:
* Adds `parentPlace`, `places`, `childPlaceType` attributes to the line chart web component
* Depreciates `place`, so that now the line chart's attributes match all the other web components (but old behavior is still preserved)
* If multiple places are provided, update legend labels so that the place name is shown

Things that should be discussed and come in a future PR:
* Use the stat var name in the legend instead of the stat var's dcid
* Fixing the way legends are displayed, especially when there are many lines
* A clearer way to display multiple variables, multiple places, both with legend and the lines themselves
* Allowing for different types of line (dotted, dashed), instead of having color be the only differentiator

Examples:
Specify multiple places
```
<datacommons-line
  title="Population Of California, Texas, and New York State"
  places="geoId/06 geoId/48 geoId/36"
  variables="Count_Person"
  colors="green red blue"
></datacommons-line>
```
Screenshot 2023-07-21 at 10 29 12 AM

Use all child places in a parent place
```
<datacommons-line
  title="Population Of Counties of Rhode Island"
  parentPlace="geoId/44"
  childPlaceType="County"
  variables="Count_Person"
></datacommons-line>
```
Screenshot 2023-07-21 at 10 29 19 AM

Multiple places, and multiple stat vars
```
<datacommons-line
  title="Population Of California, Texas, and New York State, by gender"
  places="geoId/06 geoId/48 geoId/36"
  variables="Count_Person_Male Count_Person_Female"
></datacommons-line>
```
Screenshot 2023-07-21 at 10 29 28 AM


dwnoble (approved):
Nice!

static/js/components/tiles/line_tile.tsx
```diff
@@ -45,6 +46,10 @@ export interface LineTilePropType {
   className?: string;
   // colors to use
   colors?: string[];
+  // List of place DCIDs to plot, instead of enclosedPlaceType in place
+  comparisonPlaces?: string[];
```
chejennifer, line 50: wonder if it would make more sense for comparison places to be places that we plot in addition to the main place? To be consistent with the behavior in other tiles (e.g., bar)

static/js/components/tiles/line_tile.tsx
```diff
@@ -125,19 +130,47 @@ export const fetchData = async (props: LineTilePropType) => {
...
     paramsSerializer: stringifyFn,
   });
-  return rawToChart(resp.data, props);
+
+  // get place names from dcids
+  const placeDcids = Object.keys(resp.data.data[statVars[0]]);
+  const placeNames = await getPlaceNames(placeDcids, props.apiRoot);
+  // How legend labels should be set
+  const options = {
+    // If many places and one stat var, legend should show only place labels
```
chejennifer, line 168: add a comment that if neither of these are set, we will show the stat vars in the legend labels?

chejennifer (commented):
cool!

static/js/components/tiles/line_tile.tsx
juliawu: Added!

static/js/components/tiles/line_tile.tsx
juliawu: I like the suggestion of maintaining consistency, but I think the current behavior in other tiles is to have comparisonPlaces to be _instead of_ the main place.

It looks like comparisonPlaces is only also used in the bar tile. The current data fetch doesn't get data for the main place if comparisonPlaces is set. See: https://github.com/datacommonsorg/website/blob/537b332b85959480abe982a10e7e564464140924/static/js/components/tiles/bar_tile.tsx#L172-L185

Based on how our TileConfig is set though, it looks like the intention is to have comparisonPlaces be in _addition to _the main place.

Should I update both the line tile and bar tile to plot main place in addition to comparisonPlaces? Or should match the line tile with the current behavior of the bar tile (either main place, or comparisonPlaces, but not both)?

juliawu: @dwnoble @chejennifer

juliawu: Going with matching comparisonPlaces with how it works in the bar chart. If requested, I'll update the behavior in a followup PR.

PR #3000, opened 2023-07-22
Get working chart on insights app

juliawu:
Get a proof of concept chart up on the insights page:

2 main changes from @pradh's branch:
* Added links to datacommons.js and datacommons.min.css to insights.html head tags
* Updated the &lt;datacommons-bar&gt; attributes to the new syntax
* Add custom elements to JSX.IntrinsicElements so webpack knows how to compile our web components

Screenshot:
Screenshot 2023-07-22 at 12 40 57 PM


juliawu:
Note: JSX will still complain that:       ```TS2339: Property 'datacommons-bar' does not exist on type 'JSX.IntrinsicElements'.```

Because JSX is expecting this non-standard html tag to actually be an incorrectly cased JSX element

@dwnoble do you know of how fix this? We can wrap our components into JSX elements, but I think there might be a cleaner solution.

pradh (approved):

static/index.d.ts
```diff
@@ -21,3 +21,47 @@ declare module "datacommons" {
...
+  lollipop?: boolean;
+  max?: number;
+  maxPlaces?: number;
+  min?: number;
+  parentPlace?: string;
+  place?: string;
+  places?: string[];
+  publish?: string;
+  showLowest?: boolean;
+  sort?: import("./js/chart/types").SortType;
```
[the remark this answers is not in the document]
juliawu: It looks like module declarations can't have imports at the top, otherwise it gets treated as a normal module file (See top ansew of this stack overflow page: https://stackoverflow.com/questions/39040108/import-class-in-definition-file-d-ts).

static/js/apps/insights/app.tsx
```diff
@@ -54,7 +54,7 @@ export function App(): JSX.Element {
         setChartData(chartData);
       }
     })();
-  }, [window.location.hash]);
```
pradh, line 57: Why is this removed?

juliawu: `window.location.hash` is outside the scope of the app. I don't think a change to `window.location.hash` will actually result in a re-`fetchData` if it gets changed.

static/index.d.ts
```diff
@@ -21,3 +21,47 @@ declare module "datacommons" {
...
+  title?: string;
+  value?: number;
+  variable?: string;
+  variables?: string[];
+  yAxisMargin?: number;
+}
+
+declare namespace JSX {
+  interface IntrinsicElements {
+    "datacommons-bar": WebComponentProps;
```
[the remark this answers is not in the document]
juliawu: Done!

static/index.d.ts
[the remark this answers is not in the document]
juliawu: Update, was able to extract these and use the `import("path/to/file")` syntax instead.

server/routes/nl/api.py
```diff
@@ -219,7 +219,7 @@ def fulfill_and_build_charts(query_detection: Detection,
 #
 # TODO: Add support for context
 #
-@bp.route('/data_via_dcid', methods=['POST'])
+@bp.route('/fulfill', methods=['POST'])
```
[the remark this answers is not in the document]
juliawu: Leaving it since it looks like this got readded in #3004

static/js/apps/insights/app.tsx
pradh: I see, ok to remove then.  Thanks for explaining...

juliawu:
Thanks for the reviews! Made the changes suggested, and readded `datacommons.js` to insights.html. Looks like some recent PRs fixed the double-defining issue.

server/templates/insights.html
```diff
@@ -26,6 +26,7 @@
  {% block head %}
  <link rel="stylesheet" href={{url_for('static', filename='css/insights.min.css', t=config['VERSION'])}} >
  <link rel="stylesheet" href="https://fonts.googleapis.com/icon?family=Material+Icons">
+ <script src="https://datacommons.org/datacommons.js"></script>
```
dwnoble, line 29: Remove?

dwnoble (approved):
Thanks Julia!

server/templates/insights.html
juliawu: This actually needs to stay in now for the charts to show up. It looks like the double-defining issue got fixed in one of the PRs, or was just a different mistake on my end.

PR #3020, opened 2023-07-25
Insight: perform fine-grained existence checks to avoid empty charts

pradh:
Previously, we did a fairly coarse grained existence check to decide whether an SV approximately has data or not (which mostly works for NL case because its either a specific place or child-place).

But for Insights page we have both the main place and child-places as well as other parent places.  So in this change, we maintain per-place per-sv results, and use that when adding specific charts.

This is maintained in PopulateState, and used only in Insights charts.


shifucun (approved):
How much time does the existence check take?

pradh:
> 

0.74s.  Its happening with a single call, largely.

PR #3027, opened 2023-07-25
[tools revamp] render simple charts based off selections

chejennifer:
- render a map, scatter, or timeline tile depending on selections

screen-recording-2023-07-25T131256 085

### DIFF
--- a/static/js/apps/visualization/chart.tsx
+++ b/static/js/apps/visualization/chart.tsx
@@ -20,13 +20,100 @@
 
 import React, { useContext } from "react";
 
+import { LineTile } from "../../components/tiles/line_tile";
+import { MapTile } from "../../components/tiles/map_tile";
+import { ScatterTile } from "../../components/tiles/scatter_tile";
+import { StatVarInfo } from "../../shared/stat_var";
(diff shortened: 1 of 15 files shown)


static/js/apps/visualization/chart.tsx
```diff
@@ -20,14 +20,101 @@
...
+          />
+        );
+      case VisType.SCATTER:
+        return (
+          <ScatterTile
+            id="vis-tool-scatter"
+            title={
+              statVarLabels[0] +
+              " (${yDate}) vs " +
+              statVarLabels[1] +
```
juliawu, line 75: What happens if there is only one StatVar and "SCATTER" is selected as VisType?

juliawu (approved):
Very cool!

static/js/apps/visualization/chart.tsx
chejennifer: there is a util function `isSelectionComplete` where if the selection is not complete, the initial selection UI will be rendered instead of the charts UI

chejennifer:
thanks!

PR #3031, opened 2023-07-26
[tools revamp] add per capita and log options

chejennifer:
- add per capita and log options

note: 
- the render chart function is getting a little out of hand. To make it easier to see the actual logic/code changes, I will do the clean up in a separate PR

screen-recording-2023-07-26T090816 941


chejennifer:
thanks for the review!

PR #3034, opened 2023-07-26
add option for custom embeddings size

n-h-diaz:


tools/nl/embeddings/README.md
```diff
@@ -52,6 +52,16 @@ This is a common sheet across the different index sizes.
    ```bash
    /run.sh -l medium <path_to_local_model>
    ```
+
+   To create custom embeddings (using the finetuned model in PROD):
+   ```bash
+   ./run.sh -c <embeddings_size> <local_sheets_csv_filepath> <worksheet_name>
```
jehangiramjad, line 58: What sheets url does this worksheet belong to? If you're recommending providing the worksheet name like this, then you must also ask for the sheets url, no (instead of using the local downloaded path?)

tools/nl/embeddings/run.sh
```diff
@@ -14,18 +14,19 @@
 # limitations under the License.
 
 function help {
-  echo "Usage: -bfl <embeddings-size> # 'small' or 'medium'"
+  echo "Usage: -bflc <embeddings-size>"
   echo "$0 -b <embeddings-size> # 'small' or 'medium'. This option uses the base default sentence_transformer model."
   echo "$0 -f <embeddings-size> # 'small' or 'medium'. This option uses the finetuned model on PROD."
   echo "$0 -l <embeddings-size> <local_model_path> # 'small' or 'medium'. This option uses the locally stored model to build the embeddings."
+  echo "$0 -c <embeddings-size> <local_sheets_csv_filepath> <sheets_url> <worksheet_name> # This option creates custom embeddings (using the finetuned model in PROD)."
```
jehangiramjad, line 21: Ok now i am a bit more confused :) Why provide the local path AND the sheets url+worksheet? They are one and the same thing, no? The local path is where the file from sheets_url+worksheet is downloaded to, no?

n-h-diaz: sorry for my confusion! it looks like the sheets url+worksheet is just getting written to the local path, so I added an option to avoid overwriting the existing file that's checked in. but I guess it doesn't matter if it's overwritten (can choose not to push, or can rerun with small/medium), so I removed this flag

tools/nl/embeddings/README.md
jehangiramjad: Ok, I am still a bit unsure what this means. I'll approve to unblock you though ;) Could you please document in the README exactly what supplying the local csv path means and what it means if you don't provide it etc? Thanks!

jehangiramjad (approved):
Approving to unblock but please add more documentation to the README :)

tools/nl/embeddings/README.md
n-h-diaz: sorry actually thinking about it more, I think it doesn't matter to supply the local path, so I'll remove it!

PR #3038, opened 2023-07-26
[tools revamp] fix some styling bugs

chejennifer:
- fix overflow/sizing on the selector pane
- make stat var selector a fixed size with scroll
- when user edits a place, close the window for editing once selection is made

screen-recording-2023-07-26T145258 419


juliawu (approved):
Nice!

chejennifer:
thanks for the review!

PR #3042, opened 2023-07-27
[tools revamp] add chips for timeline chart

chejennifer:
- add chips to the timeline chart
- cleaned up the code so that each vis type has their own file to generate the config for that vis type
- for map and scatter, in the selector pane, when selecting a place, show the selected place in the input so that users know they can only pick one place

Screenshot 2023-07-26 at 5 24 08 PM

Screenshot 2023-07-26 at 5 54 02 PM


juliawu (approved):
Nice refactor!

chejennifer:
thanks for the review!

PR #3044, opened 2023-07-27
[tools revamp] handle when there are no child places

chejennifer:
- if there are no allowed child place types, render a message telling user to select a different place so that user doesn't get into a bad state
- update the place label for timeline vis type to say "plot places" instead of "plot places in"

Screenshot 2023-07-26 at 6 32 45 PM


chejennifer:
thanks for the review!

PR #3046, opened 2023-07-27
[subject pages] make proto change to allow for per capita toggle on block level

chejennifer:


server/config/subject_page.proto
```diff
@@ -251,6 +251,11 @@ message Block {
     DISASTER_EVENT = 1;
   }
   BlockType type = 4;
+  // If denom is set, adds a per capita toggle on the block. If per capita is
+  // toggled on, all tiles in the block will use this denom to calculate per
+  // capita.
+  string denom = 6;
+  bool startPerCapita = 7;
```
pradh, line 258: nit: start_per_capita ... can you default to false?

pradh: or better call it `start_with_denom` ... since only Count_Person is per-capita.

pradh (approved):
Thanks!

server/config/subject_page.proto
chejennifer: done & yes, it will default to false

PR #3048, opened 2023-07-27
fix hardcoded place search

chejennifer:
- when new place search components are rendered, it adds another pac-container to the html. To get the current pac-container, we want the last one in the list, not the first


chejennifer:
thanks for the review!

PR #3049, opened 2023-07-27
update favicon

chejennifer:
left is the new icon, right is the old
Screenshot 2023-07-27 at 11 10 13 AM


juliawu:
Hm, in the screenshot, the left icon looks fuzzier to me than the right one. If we're making the left the new one, is that intentional?

chejennifer:
> Hm, in the screenshot, the left icon looks fuzzier to me than the right one. If we're making the left the new one, is that intentional?

So that's the icon I got from Ido, but let me check with him ..

chejennifer:
Updated - the one on the left is the new one 
Screenshot 2023-07-27 at 3 25 59 PM

juliawu (approved):
Thanks for checking!

PR #3051, opened 2023-07-27
[subject page] add block level per capita option

chejennifer:
- UI changes to allow for block level per capita option

screen-recording-2023-07-27T124816 969

### DIFF
--- a/static/js/components/subject_page/stat_var_provider.ts
+++ b/static/js/components/subject_page/stat_var_provider.ts
@@ -41,7 +41,7 @@ export class StatVarProvider {
   }
 
   // Gets the stat var spec for a list of keys. If blockDenom is non empty, set
-  // the denom each stat var spec in the list.
+  // the denom of each stat var spec in the list.
   getSpecList(keys: string[], blockDenom?: string): StatVarSpec[] {
     return keys
       .map((k) => this.getSpec(k, blockDenom))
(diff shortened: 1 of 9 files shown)


static/js/components/subject_page/stat_var_provider.ts
```diff
@@ -27,11 +28,23 @@ export class StatVarProvider {
...
+    return {
+      ...this._statVarSpecMap[key],
+      denom: blockDenom || this._statVarSpecMap[key].denom,
+    };
   }
 
-  getSpecList(keys: string[]): StatVarSpec[] {
-    return keys.map((k) => this._statVarSpecMap[k]);
+  // Gets the stat var spec for a list of keys. If blockDenom is non empty, set
+  // the denom each stat var spec in the list.
```
juliawu, line 44: nit: "the denom **of** each stat var spec"

juliawu (approved):
Awesome! Thanks for the additional tile cleanup.

static/js/components/tiles/bar_tile.tsx
```diff
@@ -257,6 +258,7 @@ function rawToChart(
         const svUnit = getUnit(raw.facets[stat.facet]);
         unit = unit || svUnit;
       }
+      console.log(props.statVarSpec);
```
pradh, line 261: drop?

pradh (approved):
Very nice!

static/js/components/tiles/bar_tile.tsx
chejennifer: good catch!

PR #3052, opened 2023-07-27
[tools revamp] stat var selector updates

chejennifer:
- change "continue" to "display"
- if going from map to scatter, make the selector header say "Select 1 more variable"
- on scatter or map, only show stat vars that are available for at least 10 places

screen-recording-2023-07-27T135434 502


juliawu (approved):

chejennifer:
thank you for the quick review!!

PR #3053, opened 2023-07-27
[tools revamp] update when to show the per capita option

chejennifer:
- extract out the nl check for whether per capita is relevant for a stat var
- use that check to decide whether or not to show the per capita option on a vis tool chart

screen-recording-2023-07-27T145726 368


juliawu (approved):
extracting out per capita check sounds good to me, but please wait for @pradh ack

pradh (approved):
nice.

PR #3065, opened 2023-07-28
[explore] add example query links

chejennifer:
- update the query link objects to take title and url. If no url is set, form link out of the tile, otherwise use url as the link


jehangiramjad (approved):
Many many thanks!

PR #3070, opened 2023-07-29
Insight: Use an in-memory Topic cache

pradh:
Trims ~5s or so on top-level topics.


server/lib/topic_cache.py
```diff
@@ -0,0 +1,157 @@
...
+from typing import Dict, List, Set
+
+from server.lib import fetch
+from server.lib.nl.common import utils
+from server.services import datacommons as dc
+
+_BATCH_SIZE = 100
+
+_CACHE_PATH = '~/.datacommons/'
+_CACHE_EXPIRY = 3600 * 24  # Cache for 1 day
```
shifucun, line 30: Do you need the expiry to be shorter?

shifucun (approved):
this is very nice!

server/lib/topic_cache.py
pradh: sg, made 3 hrs

PR #3075, opened 2023-07-29
Revert previous mixer-api-key change and don't require it in prod

pradh:
Essentially undo https://github.com/datacommonsorg/website/pull/3072, and instead of API key plumb through the context.  That way, it will work on prod and local.


shifucun (approved):
thanks!

PR #3081, opened 2023-07-30
Added 'An initiative from Google' byline

dwnoble:
Updated homepage & NL bylines per branding guidelines

Homepage:
home-byline

NL: 
nl-byline


pradh (approved):
Yay!

PR #3084, opened 2023-07-30
Handle empty space query for demo control

shifucun:


pradh:
@chejennifer @dwnoble -- maybe we should make a proper impl of this capability?

PR #3088, opened 2023-07-31
Build Topic Cache off MCF file and reenable

pradh:


pradh:
> To confirm, what this the code that generates the json?

The build_topic_cache_test.py is the script that both validates if its out of date as well as generates it.

PR #3090, opened 2023-07-31
Change insights to explore; Add redirect url for no place/topic case

shifucun:
The renaming is based on Guha's request.


pradh (approved):
This is great!

PR #3092, opened 2023-07-31
[tools revamp] add back and forward button functionality

chejennifer:
- add back and forward button functionality
- fix bugs in the landing page links
- add spinner when loading context from the url
- fix bug with place type changing when going from map -> scatter -> timeline -> scatter
- fix bug with variable staying selected even though it's greyed out in the stat var selector (because checking existence is checking for existence of at least one place, but hierarchy checks for existence of at least n places)

screen-recording-2023-07-31T134714 453


static/js/apps/visualization/app_context.tsx
```diff
@@ -330,13 +298,89 @@ export function AppContextProvider(
...
+  }, [places, enclosedPlaceType, statVars, visType]);
 
   return (
     <AppContext.Provider value={contextValue}>
       {props.children}
     </AppContext.Provider>
   );
+
+  // Updates context values based off URL params
+  function updateContextValue(): void {
```
juliawu, line 312: Nice refactor!

static/js/apps/visualization/app_context.tsx
```diff
@@ -330,13 +298,89 @@ export function AppContextProvider(
...
+        return statVars;
+      });
+    }
+    Promise.all([
+      placesPromise,
+      statVarsPromise,
+      childTypesPromise,
+      samplePlacesPromise,
+    ])
+      .then(([places, statVars, childTypes, samplePlaces]) => {
```
juliawu, line 371: Just to double check, does `shouldUpdateHash` also need to be set here?

juliawu (approved):
Thanks!

static/js/apps/visualization/app_context.tsx
```diff
@@ -76,18 +77,21 @@ export function AppContextProvider(
...
   const [samplePlaces, setSamplePlaces] = useState([]);
-  const [isInitializing, setIsInitializing] = useState(true);
+  const [isContextLoading, setIsContextLoading] = useState(true);
   const prevSamplePlaces = useRef(samplePlaces);
   const prevStatVars = useRef(statVars);
+  // List of booleans to track whether or not hash should be updated when
+  // places, enclosedPlaceType, statVars, or visType changes. Hash should be
+  // updated for user selections and NOT for selections that are auto-magically
+  // made
+  const shouldUpdateHash = useRef([]);
```
shifucun, line 94: does the order of this list corresponding to place, statVars etc..? If so, is a map better?

shifucun (approved):
this makes the logic simpler!

PR #3100, opened 2023-08-01
NL/Explore: fix some bugs, improve layout

pradh:
1. Add a highlight chart alongside single-var timeline
2. Fix for the [highest gini index in the world] query in NL
3. Fix query wordings per Guha's suggestion


pradh:
Thanks Jehangir!

PR #3105, opened 2023-08-01
update the byline logo

chejennifer:
- footers for every other page besides the homepage is light grey so the reverse logo does not show up well. Add a variable to set which logo to use in the footer

homepage
Screenshot 2023-08-01 at 7 46 03 AM

tools
Screenshot 2023-08-01 at 7 46 43 AM

### DIFF
--- a/server/templates/static/homepage.html
+++ b/server/templates/static/homepage.html
@@ -18,6 +18,7 @@
  {% set main_id = 'homepage' %}
  {% set page_id = 'page-homepage' %}
  {% set title = 'Home' %}
+ {% set brand_logo_path = "/images/google-logo-reverse.svg" %}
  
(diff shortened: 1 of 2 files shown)


server/templates/static/homepage.html
```diff
@@ -18,6 +18,7 @@
  {% set main_id = 'homepage' %}
  {% set page_id = 'page-homepage' %}
  {% set title = 'Home' %}
+ {% set brand_logo_path = "/images/google-logo-reverse.svg" %}
```
dwnoble, line 21: What do you think about making this a boolean like `brand_logo_dark = true`? might leave less room for error

dwnoble (approved):
thank you!

PR #3110, opened 2023-08-01
add flag to hide website revamp changes

chejennifer:
- flag hides changes specific for the new data commons website. flag is set to true for prod and staging.


shifucun (approved):
thanks!!

PR #3112, opened 2023-08-01
Use default 3 digits precision for ranking tile; Do not auto focus explore query box

shifucun:
2 digits gives inaccurate result as we use M, K (1000 step) as unit. Also it makes per capita value very weak (all 0.1 or all 0)

Fix some lint issues as well.


chejennifer (approved):
nice

PR #3120, opened 2023-08-02
[tools revamp] update selection flow

chejennifer:
- remove selector box when user has clicked continue/made a selection

screen-recording-2023-08-02T082208 144


chejennifer:
thanks for the review!

PR #3126, opened 2023-08-02
Show date in hight chart; Make font smaller

shifucun:
image

### DIFF
--- a/static/js/components/tiles/highlight_tile.tsx
+++ b/static/js/components/tiles/highlight_tile.tsx
@@ -70,6 +70,7 @@ export function HighlightTile(props: HighlightTilePropType): JSX.Element {
             false,
             NUM_FRACTION_DIGITS
           )}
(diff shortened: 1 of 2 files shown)


static/js/components/tiles/highlight_tile.tsx
```diff
@@ -70,6 +70,7 @@ export function HighlightTile(props: HighlightTilePropType): JSX.Element {
             false,
```
pradh, line 70: While here should we make a bit unit smaller?  USD all caps looks a bit odd.

pradh (approved):

static/js/components/tiles/highlight_tile.tsx
shifucun: This is a bit involved. Need to make formatNumber() function to return 2 parts; maybe a later change.

PR #3141, opened 2023-08-04
[tools revamp] Add some P0 features

chejennifer:
- allow specifying denom in the url for map and scatter variables
- allow specifiying date in the url for map
- add the get BQ SQL Query button for map
- add zoom functionality for map

screen-recording-2023-08-04T084540 036


static/js/tools/map/bq_query_utils.ts
```diff
@@ -180,8 +183,10 @@ function getBaseWhereClause(
 }
 
 export function getPcQuery(
```
shifucun, line 185: looks like the BQ modal is in many tools not only map. Should these live outside the map folder?

shifucun (approved):
map zoom is a very nice feature! A highly asked one from friends testing :)

static/js/tools/map/bq_query_utils.ts
chejennifer: The BQ modal lives in the shared folder. These are the query utils specific to getting the query for the map tool!

chejennifer:
thanks for the review!

PR #3145, opened 2023-08-04
[library] Add highlight component

juliawu:
Adds the ability to render a highlight tile as a web component.

This PR:
* Adds a highlight component that wraps the highlight tile 
* Adds the ability to specify a specific date in the highlight tile
* Adds a "part" to the highlight tile containing div to allow for web component styling via CSS
* Adds documentation for the new highlight component
* Also fixes an error in the map web component documentation (date attribute should be under optional, not required)

Now the syntax
```
<datacommons-highlight
  description="Population of the United States"
  place="country/USA"
  variable="Count_Person"
  date="2020"
></datacommons-highlight>
```
Produces:

Screenshot 2023-08-04 at 4 45 53 PM


packages/web-components/docs/components/highlight.md
```diff
@@ -0,0 +1,44 @@
+# Data Commons Gauge Chart Web Component
```
dwnoble, line 1: s/Gauge/Highlight/

packages/web-components/docs/components/highlight.md
```diff
@@ -0,0 +1,44 @@
...
+<datacommons-highlight
+  title="Percentage of US Population that are Internet Users"
+  place="country/USA"
+  variable="Count_Person_IsInternetUser_PerCapita"
+  min="0"
+  max="100"
+></datacommons-highlight>
+```
+
+<img src="../assets/gauge.png" width="620"/>
```
dwnoble, line 18: s/gauge/highlight/

dwnoble (approved):
Thanks Julia! I think @pradh added something to the highlight tile the other day to show the unit-can you verify this works with the web component?

packages/web-components/docs/components/highlight.md
juliawu: Thanks for the catch!

packages/web-components/docs/components/highlight.md
juliawu: Thanks for the catch! Added a new image and linked to that instead.

juliawu:
> Thanks Julia! I think @pradh added something to the highlight tile the other day to show the unit-can you verify this works with the web component?

The unit will show in the tile when a unit is passed into the statvar spec. Added an optional attribute to set a "unit" string to display to the web component.

PR #3146, opened 2023-08-05
[library] Add subtitle slot to web components

juliawu:
Adds a slot to the web components to add an optional subtitle.

Note that based on the inline comment in `chart_tile.tsx`, it looks like intended behavior was for space for the header to remain even if a title is not provided. The CSS changes are to maintain this behavior, while avoiding awkward blank space between the title and subtitle.

Also confirmed these changes don't change the way our tiles are rendered in our Search/Chat interfaces.

Example:
```
<datacommons-bar
  title="Most populous states in the US"
  parentPlace="country/USA"
  childPlaceType="State"
  variables="Count_Person"
  maxPlaces="15"
>
<p slot="subtitle" style="text-align:center; color: grey; font-size: 0.85rem">(top 15 places)</p>  
</datacommons-bar>
```
Screenshot 2023-08-04 at 5 30 42 PM

### DIFF
--- a/packages/web-components/examples/all-charts.html
+++ b/packages/web-components/examples/all-charts.html
@@ -76,7 +76,9 @@
       childPlaceType="State"
       variables="Count_Person"
       maxPlaces="15"
-    ></datacommons-bar>
+    >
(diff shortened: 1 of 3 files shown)


chejennifer (approved):
nice!

PR #3160, opened 2023-08-08
Create separate formatUnit and formatNumberAndUnit function

shifucun:
This is to make it possible format value and unit differently.

Also cleaned up the formatNumberFn args as it is not used anymore.


chejennifer (approved):
nice & thanks for the clean up!

shifucun:
I ended up reverting creating two separate functions but use translateUnit for i18n for the unit formatting only.

PR #3168, opened 2023-08-09
Explore: Fork off page config building for SDG

pradh:
And implement the 3-level structure on the RHS.  Right now default_sdg looks like main, but the plan is to drive it off a config in upcoming PR.

image

### DIFF
--- a/server/lib/explore/page_sdg.py
+++ b/server/lib/explore/page_sdg.py
@@ -17,12 +17,15 @@ import time
 from typing import Dict, List
 
 from server.config.subject_page_pb2 import SubjectPageConfig
-from server.lib.explore.detector import Params
-from server.lib.explore.page_type import default
+from server.lib.explore import params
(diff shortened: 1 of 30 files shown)


server/lib/explore/page_sdg.py
```diff
@@ -0,0 +1,167 @@
...
+      footnote=variable.get_sv_footnote(all_svs),
+  )
+  state.uttr.counters.timeit('get_sv_details', start)
+
+  # Decide on the levels of the hierarchy!
+  topic_tree = _get_topic_tree(chart_vars_list, sv2thing, dc)
+
+  builder = Builder(state, env_config, sv2thing, len(chart_vars_list))
+
+  if topic_tree.page:
```
shifucun, line 61: comment on what are page vs category vs block

pradh: Done, thanks!

PR #3172, opened 2023-08-09
[tools revamp] add date options for scatter

chejennifer:
- allow setting date options in the url for scatter stat vars
- update subject page proto so that stat var spec takes a date


juliawu (approved):
Nice!

chejennifer:
thanks for the review!

PR #3177, opened 2023-08-09
Various bugs

chejennifer:
- update explore more link on tiles to go to the new vis tools
- in explore page, when new place is clicked, clear the query input
- fix all cap section headers
- when choosing per capita at block level, use percent instead of ratio

screen-recording-2023-08-09T163509 554


dwnoble (approved):
Nice!

chejennifer:
thanks for the reviews!

PR #3181, opened 2023-08-10
[library] Swap title for header attribute and enable tooltip on hover for the line web component

juliawu:
This PR gives a tooltip when hovering over a line chart web component. The logic to render a tooltip was already present, this PR just enables the tooltips in the web components and adjusts the CSS.

**Important!:** In order to disable the browser default tooltip that also comes up on a hover, the `title` attribute has been renamed to `header`.

Screenshots:

Single line:
Screenshot 2023-08-10 at 4 50 10 PM

Long variable names will wrap to limit how much space the tooltip takes up
Screenshot 2023-08-10 at 4 48 16 PM

A scrollbar appears when there are too many places to fit:
Screenshot 2023-08-10 at 4 48 43 PM


chejennifer (approved):
nice!

dwnoble:
Thanks Julia! 

(1) Can you make all the existing title attributes still work but deprecated? We can remove all of the deprecated properties once we eventually move this to a proper npm package with versioning
(2) What do you think about changing the highlight chart's "description" attribute to "header"?

juliawu:
> Thanks Julia!
> 
> (1) Can you make all the existing title attributes still work but deprecated? We can remove all of the deprecated properties once we eventually move this to a proper npm package with versioning (2) What do you think about changing the highlight chart's "description" attribute to "header"?

Love both ideas. Titles are now deprecated but still work. The highlight chart's `description` is also deprecated in favor of `header`. PTAL!

dwnoble (approved):
Nice!

PR #3184, opened 2023-08-11
Explore: Fulfillment support for Topic Expansion

pradh:
That is, use any StatVarGroups defined in a Topic if there are too few charts, then classify the SVs based on the definition (i.e., differ only by contraint-value) into SVPGs, perform existence checks and add to the chart-config.

These are extra now on the [economy of Brazil] query.

image


jehangiramjad (approved):

pradh:
Thanks!

PR #3186, opened 2023-08-11
Update subject page files with chart tile changes

juliawu:
Updates the interfaces in the subject page protos, JS, and tests to reflect recent additions/changes to the chart tile props. This brings feature parity between the web components and the subject pages, which are based on tiles.

**Important**: The following settings are now enabled on the subject pages by default:

1. Allowing zooming on the map tiles
Screenshot 2023-08-11 at 4 10 29 PM

3. Tooltips on line tiles
Screenshot 2023-08-11 at 4 11 32 PM


server/config/subject_page.proto
```diff
@@ -164,9 +164,36 @@ message ScatterTileSpec {
...
+    SORT_DESCENDING = 2;
+    SORT_ASCENDING_POPULATION = 3;
+    SORT_DESCENDING_POPULATION = 4;
+  }
   // [optional] The root url to use for labels in the x-axis. If not specified,
   // labels in the x-axis will link to Place Explorer.
   string x_label_link_root = 1;
+  // Bar height for horizontal bar charts
+  double bar_height = 2;
+  // list of specific colors to use
```
pradh, line 178: nit: What are the color values, just english words, small-case?

server/config/subject_page.proto
```diff
@@ -164,9 +164,36 @@ message ScatterTileSpec {
...
+  repeated string colors = 3;
+  // whether to plot bars horizontally.
+  bool horizontal = 4;
+  // maximum number of places to display
+  int32 max_places = 5;
+  // bar sort order
+  // options:
+  // - "ascending"
+  // - "descending"
```
pradh, line 187: The enums are self-explanatory?

SortType sort = 6 [default = SORT_DESCENDING_POPULATION]

pradh (approved):
Thanks Julia!

server/config/subject_page.proto
juliawu: Removed the options comments. It looks like setting explicit defaults are disallowed in proto3, so I left a comment that default = descending population instead.

server/config/subject_page.proto
```diff
@@ -164,9 +164,36 @@ message ScatterTileSpec {
 }
 
 message BarTileSpec {
+  enum SortType {
+    SORT_ASCENDING = 1;
+    SORT_DESCENDING = 2;
+    SORT_ASCENDING_POPULATION = 3;
+    SORT_DESCENDING_POPULATION = 4;
+  }
```
juliawu, line 172: @pradh @dwnoble I'm less familiar with how protobufs and MessageToJson() works. This enum is intended to match these lines: https://github.com/datacommonsorg/website/blob/59601b22c428665893ff8b78e88da0551e597f24/static/js/chart/types.ts#L170-L178. Is this the right approach?

server/config/subject_page.proto
juliawu: Any CSS-compatible color string (hex, rgb, rgba, etc) is allowed. Updated the comment line here and everywhere else `colors` shows up in the file.

server/config/subject_page.proto
pradh: I think the JS code will see the sort field value as a string "SORT_DESCENDING", that you might then need to map to "descending".

I would recommend just using identically named enum names, so that's just a lower() op?

juliawu: Thanks! Swapped to identically named enum names, and added a typecasting function

pradh (approved):
Thanks!

PR #3190, opened 2023-08-12
World Bank DC setup; Custom DC config update

shifucun:
Added global client side variables to indicate the data source dcid and name used for the special category.


server/app_env/_base.py
```diff
@@ -66,4 +66,8 @@ class Config:
   # full serving path from /static folder.
   OVERRIDE_CSS_PATH = ''
   # Hide changes made for the website revamp.
-  HIDE_REVAMP_CHANGES = False
\ No newline at end of file
+  HIDE_REVAMP_CHANGES = False
+  # The dcid of the special data source to show as top level category in the hierarchy
+  DATA_SOURCE_DCID = ''
+  # The name of the speical data source to show as top level category in the hierarchy
```
chejennifer, line 72: special

static/js/stat_var_hierarchy/stat_var_group_node.tsx
```diff
@@ -246,11 +249,20 @@ export class StatVarGroupNode extends React.Component<
...
     const entityList = this.props.entities;
     this.dataFetchingEntities = this.props.entities;
+    let numEntitiesExistence = this.props.numEntitiesExistence;
+    // When dataSource is specified, the stat var group node info fetch should
+    // also be constrained by the data source. So adding the data source dcid
+    // to the `entities` list for this purpose.
+    const entityDcids = entityList.map((entity) => entity.dcid);
+    if (this.props.dataSource) {
+      entityDcids.push(this.props.dataSource);
+      numEntitiesExistence = entityDcids.length;
```
chejennifer, line 259: in the case where data source is specified, do all entity dcids have to have data for a stat var in order for that stat var to be displayed in the hierarchy? because in the contained in place examples, we sometimes pass in 50 entities, do all of them need to have data?

static/js/stat_var_hierarchy/stat_var_hierarchy.tsx
```diff
@@ -211,6 +212,12 @@ export class StatVarHierarchy extends React.Component<
                           showAllSV={this.state.showAllSV}
                           expandedPath={this.state.expandedPath.slice(1)}
                           numEntitiesExistence={this.props.numEntitiesExistence}
+                          dataSource={
+                            // This is a virtual node for holding stat vars of
+                            // a data source.
+                            // Add data source to constrain the this node.
```
chejennifer, line 218: constrain this

chejennifer (commented):
cool!

server/app_env/_base.py
shifucun: Done

static/js/stat_var_hierarchy/stat_var_group_node.tsx
shifucun: Yea, this is not optimal right now. Ideally, I think we should separate out the data source and the entities from the constraints. This need the API change though. I would add a todo for this.

Doing this to let the custom DC working for now.

static/js/stat_var_hierarchy/stat_var_hierarchy.tsx
shifucun: Done

shifucun (commented):
thanks!

chejennifer (approved):

PR #3193, opened 2023-08-13
Do not deploy to us-west1 for dev and autopush

shifucun:
us-west1 is not co-located with the BT cache, and could have longer mixer->web server latency.
it also makes deployment longer as the deployment need to happen sequentially for two clusters.


pradh (approved):
Thanks Bo!

PR #3197, opened 2023-08-14
[tools revamp] Bug fixes

chejennifer:
- resize chart when window resizes
- enable line chart tooltip

screen-recording-2023-08-14T160103 733


shifucun (approved):

chejennifer:
thanks for the review!

PR #3198, opened 2023-08-14
Explore: fixes for SDG and temperature

pradh:
1.  Support plumbing of descriptions and footnotes into Blocks
2.  Add Gauge tile for SDG SVs that support percentages
3.  Expand the number of SVs per topic to 20 so all goals show up on the RHS for dc/topic/sdg


server/lib/nl/config_builder/gauge.py
```diff
@@ -0,0 +1,34 @@
...
+from server.config.subject_page_pb2 import StatVarSpec
+from server.config.subject_page_pb2 import Tile
+from server.lib.nl.config_builder import base
+
+
+def gauge_block_for_percent(column, place, sv, sv2thing: base.SV2Thing):
+  chart_title = base.decorate_chart_title(title=sv2thing.name[sv], place=place)
+
+  tile = Tile(type=Tile.TileType.GAUGE, title=chart_title, stat_var_key=[sv])
+  tile.gauge_tile_spec.range.min = 0.0001
```
shifucun, line 24: why not 0?

shifucun (approved):

server/lib/nl/config_builder/gauge.py
pradh: There's a gauge bug Julia is fixing.

dwnoble (approved):
Nice

PR #3199, opened 2023-08-14
[explore] fix back button

chejennifer:
- fix back button problems caused by updating place & topic when given a query in the url

note: this is a hacky approach & should come up with a better way to do this

[screen-recording - 2023-08-14T163735.738.webm](https://github.com/datacommonsorg/website/assets/69875368/6124abba-9b8a-437c-8878-ed196c62d009)


dwnoble:
This logic is a little rough. I think the better solution might be to just not update the hash at all. we could have the query parameters be the search query OR the place and topic, but not both. this would also prevent an old search query from being associated with a "stale" pair of topic/places as we iterate on the backend logic.

chejennifer:
Chatted offline and will switch the approach to not updating the hash but instead adding an app state through context. Closing this PR

PR #3200, opened 2023-08-14
[Bug fixes] Fix single colors in gauge tile, proto enum

juliawu:
Fix 2 bugs:

### 1. Allow for specifying a single color to the `colors` prop of the gauge tile
Before the gauge tile would incorrectly render a single color passed to the `colors` prop as black, regardless of the color passed in. This is because behind the scenes D3 is expecting a color range instead of a single color value. This change passes a single color as a range by repeating the value.

### 2. Proto3 enums should start with 0
Change in https://github.com/datacommonsorg/website/pull/3186 prevents `./run_server.sh` from spinning up a server locally because proto3 requires enums to start from 0. Offending enum is updated.


server/config/subject_page.proto
```diff
@@ -165,10 +165,10 @@ message ScatterTileSpec {
...
 message BarTileSpec {
   enum SortType {
-    ASCENDING = 1;
-    DESCENDING = 2;
-    ASCENDING_POPULATION = 3;
-    DESCENDING_POPULATION = 4;
+    ASCENDING = 0;
+    DESCENDING = 1;
+    ASCENDING_POPULATION = 2;
+    DESCENDING_POPULATION = 3;
```
pradh, line 171: nit: Might make the default one as 0

pradh (approved):

pradh:
Also, in the PR, could you maybe drop this -- https://screenshot.googleplex.com/BjEJjn9LWPHaHGr.  Thanks!

juliawu:
> Also, in the PR, could you maybe drop this -- https://screenshot.googleplex.com/BjEJjn9LWPHaHGr. Thanks!

Done!

server/config/subject_page.proto
juliawu: Following offline conversation of @keyurva, using `TYPE_UNSPECIFIED` as 0 to match proto conventions.

PR #3210, opened 2023-08-15
[library] Add scatter chart web component

juliawu:
Add a web component wrapper for our scatter tile. This will let us fully implement our explore pages externally with web components instead of tiles.

Things that need to be implemented in followup PRs:
* The web component does not currently fetch stat var names, so the axes labels default to the variable's DCID. We need to add logic in the data fetch.
* Need to add support for custom color scales
* Need to add parts for custom styling

Example:
```
<datacommons-scatter
  header="Population vs Median Household Income for US States"
  variables="Count_Person Median_Income_Household"
  parentPlace="country/USA"
  childPlaceType="State"
></datacommons-scatter>
```
Screenshot 2023-08-15 at 12 52 14 PM


dwnoble (approved):
Nice!

Can you add a link to the scatter tile (and highlight tile while you're at it) here? https://screenshot.googleplex.com/3sTynZtE6vq83VS  in this file https://github.com/datacommonsorg/website/blob/master/packages/web-components/README.md

PR #3221, opened 2023-08-16
Update Stat Var explore section

shifucun:
* For the same stat var, only show the explore section once
* Use new list item style
* Show maximum 4 stat vars per section
image


static/js/components/subject_page/block_container.tsx
```diff
@@ -21,20 +21,31 @@
...
 const DELIM = "___";
 
+function camelCaseToWords(camelCaseString: string): string {
+  const words = camelCaseString
+    .replace(/([a-z])([A-Z])/g, "$1 $2") // Insert space before uppercase letters
+    .split(/[\s_-]+/) // Split by spaces, underscores, and hyphens
+    .filter((word) => word.length > 0) // Remove empty words
+    .map((word) => word.toLowerCase()); // Convert to lowercase
+
+  return words.join(" ");
```
pradh, line 38: Should we do init-caps on these?

shifucun: Done

shifucun (commented):
thanks!

PR #3231, opened 2023-08-16
Fix query bug, enable exploreMore by default

pradh:
Also, pick the exploreMore thing with most breakdown and pick just the top one.


shifucun (approved):

keyurva (approved):
Thanks.

PR #3235, opened 2023-08-16
Fix gauge tile sizing issues

juliawu:
The current gauge tile grows too large when the tile gets wide, which made it hard to add gauge tiles to the explore pages. This PR converts the previous minimum height prop to a fixed height, which matches the behavior of all other tiles.

Before:
Screenshot 2023-08-16 at 11 03 21 AM

After: 

Screenshot 2023-08-16 at 11 12 35 AM


dwnoble (approved):
Nice!

juliawu:
@pradh FYI, this should fix the sizing issues you were seeing. We can try bringing gauge tiles back to the explore pages again.

pradh:
> @pradh FYI, this should fix the sizing issues you were seeing. We can try bringing gauge tiles back to the explore pages again.

Go ahead, please send out PR and have Dan test it?

PR #3240, opened 2023-08-16
[stylesheets] add some global variables

chejennifer:
- add some global color and border variables

note: to make the PR easy to review, will move files around in next PR

also fix empty footer being shown


static/js/components/tiles/chart_footer.tsx
```diff
@@ -29,6 +29,9 @@ interface ChartFooterPropType {
 }
 
 export function ChartFooter(props: ChartFooterPropType): JSX.Element {
+  if (!props.handleEmbed && !props.exploreLink) {
```
dwnoble, line 32: did you mean for this to make it in?

dwnoble (approved):
Nice!

static/js/components/tiles/chart_footer.tsx
chejennifer: oh, yes, that was a small bug I found, let me add a note in the PR comment

PR #3242, opened 2023-08-17
NL: Simplify Context Handling

pradh:
Key context related changes:

1. NL fulfillment logic no longer refers to the context.
2. Instead, before fulfillment there is now a function `merge_with_context`, which is used by both Explore and NL, to merge past context into current utterance.
3. As part of this change, there is now only 1 unified past utterance.

Generally, the diffs are clean, except for one regression (CURRENT_QUERY -> UNFULFILLED status in response when there is place fallback).  This should be fixed in upcoming PRs.

The main diffs are from removing the `suggestions` response in chat, which is not currently used.


server/integration_tests/test_data/detection_api_context/correlatewithasthma/chart_config.json
```diff
@@ -16,7 +16,8 @@
   "context": {},
   "debug": {},
   "entities": [
-    "geoId/32"
+    "geoId/32",
```
jehangiramjad, line 19: this expanding to include another place is expected?

server/integration_tests/test_data/detection_api_context/medianincomeinsantaclaracountyandalamedacounty/chart_config.json
```diff
@@ -1,13 +1,13 @@
 {
   "childEntityType": "",
   "comparisonEntities": [
-    "geoId/06085"
+    "geoId/06001"
```
jehangiramjad, line 4: Seems like this order flipped. That's ok?

server/lib/nl/detection/context.py
```diff
@@ -0,0 +1,272 @@
...
+  data_dict.update({
+      Params.ENTITIES.value: places,
+      Params.VARS.value: vars[:max_returned_vars],
+      Params.SESSION_ID.value: uttr.session_id,
+      Params.CMP_ENTITIES.value: cmp_places,
+      Params.CMP_VARS.value: cmp_vars[:max_returned_vars],
+      Params.CHILD_TYPE.value: '' if not place_type else place_type.value,
+  })
+
+  # 6. Set the detected params in uttr ctx and clear past contexts.
```
jehangiramjad, line 118: nit: this should be # 7

server/lib/nl/detection/context.py
```diff
@@ -0,0 +1,272 @@
...
+  if hoist_topics:
+    _hoist_topic(parts[0].svs)
+    _hoist_topic(parts[1].svs)
+  return parts[0].svs, parts[1].svs
+
+
+#
+# A topic may not often be the top-most result. In that case,
+# we look for a topic for up to TOPIC_RANK_LIMIT, and hoist to top
+# (This is the same limit NL interface uses for opening up topic).
```
jehangiramjad, line 186: nit: if these are now unified, this comment becomes redundant?

jehangiramjad (approved):
Thanks. Some nits.

shifucun (approved):
this does simplify the fulfillment a lot!

server/lib/nl/detection/context.py
pradh: This comment is still relevant for now.

server/lib/nl/detection/context.py
pradh: ah, yes

server/integration_tests/test_data/detection_api_context/correlatewithasthma/chart_config.json
pradh: Good find.  I've fixed this!  (The Explore frontend though will only use the first one)

pradh (commented):
Thanks for the review both!

PR #3245, opened 2023-08-17
Move UN tracker css to a separate file.

keyurva:
* Moving css to a separate file makes it easier to use it in the UN custom DC.

[header-css.webm](https://github.com/datacommonsorg/website/assets/1221814/ddf8a8dd-bd16-4e2f-9fab-8349febf13b6)

### DIFF
--- a/experimental/sdg-tracker/src/components/layout/AppHeader.tsx
+++ b/experimental/sdg-tracker/src/components/layout/AppHeader.tsx
@@ -15,116 +15,39 @@
  */
 
 import { Link } from "react-router-dom";
-import styled from "styled-components";
 import { BrandingLink } from "./components";
+import "./AppHeader.css";
+import React from "react";
 
 type HeaderOptions = "home" | "countries" | "goals" | "topics" | "search";
 
-const HeaderContainer = styled.div`
-  width: 100%;
-  background: white;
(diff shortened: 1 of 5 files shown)


experimental/sdg-tracker/src/components/layout/AppHeader.tsx
```diff
@@ -15,116 +15,39 @@
...
-  align-items: center;
-  border-bottom: 1px solid #f1f1f1;
-  display: flex;
-  flex-direction: row;
-  height: 75px;
-  justify-content: space-between;
-  padding: 0 4rem;
-  width: 100%;
-`;
+const newNode = (className: string) => {
```
dwnoble, line 30: Instead of creating the new node here, i think it would be more clear to just replace these custom classes `<TopContainer />`, etc with divs directly in the AppHeader definition

E.g.,

```
const AppHeader = (props: { selected: HeaderOptions }) => {
  const { selected } = props;
  return (
    <div className="header-container">
      <div className="top-container">
      ...
```

keyurva: Done.

dwnoble (approved):
Nice!

PR #3246, opened 2023-08-17
Create a base image for web compose docker container

shifucun:
The website compose container need Python, Node, Golang, Protoc and Envoy. Created a base image for that to make the docker build much simpler. 

Before the change, different base image are combined which have incompatible debian system library.


keyurva (approved):
Very cool!

PR #3248, opened 2023-08-17
NL: Rework the flow

pradh:
This PR sets up the NL flow to be more amenable to Explore like structure, where we want to generate charts of different types for a single query type.

Currently, given some input SVs and places, the NL flow is like this:

1. For each inferred "query type" in fallback list
2. For each place in fallback list
3. Perform place + SV existence checks
4. Maybe try to add charts for each SV

If we fail to add charts in (4), we'd go back to (1).  This is rigid (and inefficient) when we want to do multiple things (containedin + ranking + timeline).

This change folds the "query type" down, so repeated attempts don't involve new existence checks.

1. For each place in fallback list
2. Perform place + SV existence checks
3. For each inferred "query type" in fallback list
4. Maybe try to add charts for each SV

Future PR will fold in the existence check to happen only once. The diffs seem clean (I also run through the demo scripts), with one regression in reporting of placeFallback from "counties of California" to "California".  Also to be fixed in a future PR.


jehangiramjad (approved):
I mostly looked at the diffs which look good (based on your notes).

pradh:
Thanks for the review!

PR #3249, opened 2023-08-18
[sdg-tracker] Added lazy web component loading

dwnoble:
Screencast:

https://github.com/datacommonsorg/website/assets/13766/432ff955-582a-4748-8463-60f85247d5da

### DIFF
--- a/experimental/sdg-tracker/src/components/countries/CountriesContent.tsx
+++ b/experimental/sdg-tracker/src/components/countries/CountriesContent.tsx
@@ -16,7 +16,7 @@
 
 import { CaretDownOutlined, LoadingOutlined } from "@ant-design/icons";
 import { AutoComplete, Breadcrumb, Layout, Spin } from "antd";
-import React, { Suspense, useEffect, useMemo, useRef, useState } from "react";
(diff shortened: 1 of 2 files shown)


experimental/sdg-tracker/src/components/countries/CountriesContent.tsx
```diff
@@ -31,6 +32,11 @@ import {
   FulfillResponse,
 } from "../../utils/types";
 import { SearchBar } from "../layout/components";
+
+// Approximate chart heights for lazy-loading
+const CHART_HEIGHT = 389;
```
juliawu, line 37: Something to think about for later: If we ever add horizontal bar charts, that chart height is dynamic based on number of vars/places to plot and if there's a set bar height. We should remember to revisit this if horizontal bar charts get added to the UN pages.

juliawu (approved):
Nice!

experimental/sdg-tracker/src/components/countries/CountriesContent.tsx
dwnoble: That sounds good- we can come up with a more dynamic chart height estimate depending on the parameters

PR #3256, opened 2023-08-18
NL: Clean up Config Builder API

pradh:
This is a zero (intentional) logic change refactor to avoid using a dict into the Config Builder API.

The chart diffs actually fix a bug with title in Explore currently, because it passes [`child_type` in the map](https://github.com/datacommonsorg/website/blob/b2ead6e13472bcd568cf1374c3ee66c774f60048/server/lib/explore/page_type/default_main.py#L115), and the map builder [checks for `place_type`](https://github.com/datacommonsorg/website/blob/b2ead6e13472bcd568cf1374c3ee66c774f60048/server/lib/nl/config_builder/map.py#L43) (a reason to avoid this dict in between).


jehangiramjad (approved):
Thanks!

PR #3258, opened 2023-08-19
NL: Switch to Explore style topic expansion and Per Capita

pradh:
This change includes:
* Using Topic expansion in Explore style using the same code (and deleting NL style topic expansion)
* Drop per-capita charts on NL in favor of showing a check box
* Switch the custom NL Correlation handling to be more like other query-type handlers (cleans it up a lot)
* Ensure the order of the Embeddings returned SVs is maintained by using an OrderedMap from `compute_chart_vars_map`.  This regressed in [this PR](https://github.com/datacommonsorg/website/pull/3248).
* Delete redundant charts in NL, like Explore.  But be sure to disambiguate TIME_DELTA_ACROSS* charts, which may plot the same variable multiple times (it may show up in different positions in ranking across abs vs. relative vs. per-capita).
* Try as much as possible to match NL style titles.

There is one known regression with layout of SIZE_ACROSS_ENTITIES results -- [how big are schools in Sunnyvale] -- instead of all results on the same row, we now have a table per-row ([demo2_cities_feb2023/query_1/chart_config.json](https://github.com/datacommonsorg/website/pull/3258/files#diff-162678b066943222a2e92c47da16c25085f744fccb14205363184dd2d3a05281)).

Besides that the diffs are largely from deleting Per Capita charts.

NOTE: I've disabled the fulfiller and chart_config unit-tests since updating them manually was becoming too time-consuming. I'll setup auto-gen for it, and re-enable, after the weekend.


shifucun (approved):

pradh:
Thanks Bo!

PR #3259, opened 2023-08-19
Add Explore Chart Layout, based on a flag

pradh:
This change adds a new fulfillment handler called "basic" which encompasses:
* simple
* containedin
* ranking_across_places
* ranking_across_vars
(the names need some tweaking, like what's basic vs. simple, but that's in followup)

The "basic" handler, depending on whether it is explore vs. chat, will route to one or more of these components appropriately.

Also, for every single-line timeline (in both modes), add a highlight, and demote a single-bar barchart to highlight.

IMPORTANT NOTE:  This change makes all of the existence checks to require "is it single point", so that makes things a LOT more expensive, esp for [What are the projected temperature extremes across california].

REGRESSION NOTES:

1.  After this change, chat will show a bunch more charts.  As an example, for query [asthma in california counties], suppose it returns asthma and childhood asthma, and suppose asthma doesn't exist at county-level (it does exist, but just suppose), then today we'll not show asthma at all.  After this change we'll show asthma line chart and childhood asthma map.

2. There are cases where the "extension SVs" maybe removed (in diffs), that will be fixed in upcoming PRs.

### DIFF
--- a/server/integration_tests/test_data/demo2_cities_feb2023/query_2/chart_config.json
+++ b/server/integration_tests/test_data/demo2_cities_feb2023/query_2/chart_config.json
@@ -15,6 +15,18 @@
                     "type": "LINE"
                   }
                 ]
+              },
+              {
+                "tiles": [
+                  {
+                    "description": "Asthma in Sunnyvale",
+                    "statVarKey": [
+                      "Percent_Person_WithAsthma"
+                    ],
+                    "title": "Asthma in Sunnyvale",
+                    "type": "HIGHLIGHT"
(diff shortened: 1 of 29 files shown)


shifucun (commented):
Should we add the number of observation in existence cache, or single point bit, to make the check much faster?

pradh:
> Should we add the number of observation in existence cache, or single point bit, to make the check much faster?

That would really help!

shifucun (approved):

pradh:
Thanks Bo!

PR #3265, opened 2023-08-21
Add a Date classifier

shifucun:
This only covers very basic cases for now. Will need to expand based on the usage.

### DIFF
--- a/server/lib/nl/detection/date.py
+++ b/server/lib/nl/detection/date.py
@@ -12,17 +12,17 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import re
 from datetime import datetime
+import re
 
 from server.lib.nl.common.counters import Counters
 from server.lib.nl.detection.types import Date
 from server.lib.nl.detection.types import DateClassificationAttributes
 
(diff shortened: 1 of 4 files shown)


server/lib/nl/detection/date.py
```diff
@@ -0,0 +1,47 @@
...
+import re
+from datetime import datetime
+
+from server.lib.nl.common.counters import Counters
+from server.lib.nl.detection.types import Date
+from server.lib.nl.detection.types import DateClassificationAttributes
+
+YEAR_RE = [r'(in|after|on|before|year) (\d{4})']
+YEAR_MONTH_RE = [
+    r'(in|after|on|before) (Jan|Feb|Mar|Apr|May|Jun|Jul|Aug|Sep|Oct|Nov|Dec)(?:,)? (\d{4})',
```
pradh, line 24: should we also infer whether it is "in/on" vs. "before" vs. "after" ?

pradh (approved):
Nice!

server/lib/nl/detection/date.py
shifucun: Added a preposition field for it

PR #3266, opened 2023-08-21
Fully switch to GA4

### LINKED ISSUE #1715: [BUG] Config parameters are ignored by the connected Google Analytics 4 tag for the Universal Analytics property
### Describe the bug
The connected Google Analytics 4 tag ignores any config parameters that customize data collection for the Universal Analytics property.  For the parameters we added to configs in DC prod account (which is a Universal Analytics property), these data won’t be sent to Google Analytics. [About connected site tags.](https://support.google.com/analytics/answer/9973999#considerations)
image

### To Reproduce
1. Open the DC prod website and view a random page
2. Open the Google Analytics DC prod account and go to the real-time report. https://analytics.google.com/analytics/web/#/p249581446/realtime/overview?params=_u..pageSize%3D25%26_u..nav%3Dmaui&collectionId=life-cycle
3. Click page_view event in the real-time report
The parameters of page_event do not include the "client_storage", "anonymize_ip" or "place_cateogry".

### Expected behavior
"client_storage", "anonymize_ip"  and "place_cateogry" parameters should be included in all the events. It works in the autopush account but doesn't work in the prod account.

### Screenshots
Prod account page_view parameters:
image
Autopush account page_view parameters:
image

### Solution
We can mitigate this issue by moving the parameters assigned in the gtag(‘config’) command for Universal Analytics to a gtag(‘set’) command before the gtag(‘config’) command in the file [base.html](https://github.com/datacommonsorg/website/blob/1e6e78400d3b137c1a5f3ba1fd37025052ba1a02/server/templates/base.html).
gtag(‘set’) allows you to set values that persist across all the subsequent gtag() calls on the page. [Google tag API reference](https://developers.google.com/tag-platform/gtagjs/reference).
Example:
`gtag('set', {<parameter-value-pair>, <parameter-value-pair>})`.

beets:
We had been collecting prod data to UA and GA4 for a while, aren't looking at UA reports, and UA data collection will end soon, switch our analytics tags to GA4 for prod and staging.

Also added a GA4 tag for Dev to collect metrics from fishfood.

For a primer on the differences in data collection: https://support.google.com/analytics/answer/11986666?hl=en&ref_topic=10737980&sjid=3091673578951483115-NA#zippy=%2Cin-this-article

Also fixes #1715


shifucun (approved):
nice!

beets:
thanks for the review! we'll need to keep an eye on the metrics after the next push.

shifucun:
There is a test error blocking merge

PR #3268, opened 2023-08-21
Parity: Enable event charts and ranking-across-vars

pradh:
This does two main things:

1.  Enables event charts in Explore, and blends in SV stuff along with event charts.
2.  Drops custom ranking logic of vars across bar chart and relies on the bar-chart-spec option Julia added.

image


juliawu (approved):
LGTM

PR #3272, opened 2023-08-22
[explore] make new detect + fulfill endpoint

chejennifer:
add new endpoint that does detection and fulfillment


pradh (approved):

chejennifer:
thanks for the review!

PR #3276, opened 2023-08-22
[explore] update what's saved in the URL param

chejennifer:
- update URL params to only take place (p), topic (t), query (q), disableExploreMore (em), dc (dc), and nlFulfillment (nl)
- when q is set, call the new detect-and-fulfill endpoint
- clean up flow of the explore app, tldr:
    - all links update the hash
    - when hash is updated, read the hash, fetch data using hash + saved context, save the fulfillment result to the state
    - use the context saved from fulfillment to generate the links on the page


server/lib/nl/detection/detector.py
```diff
@@ -142,7 +142,10 @@ def construct(entities: List[str], vars: List[str], child_type: str,
       if not any([child_type == x.value for x in types.ContainedInPlaceType]):
         counters.err('failed_detection_badChildEntityType', child_type)
         return None, f'Bad childEntityType value {child_type}!'
-      child_type = types.ContainedInPlaceType(child_type)
+      if child_type == places[0].place_type:
+        child_type = None
+      else:
```
pradh, line 147: Why is this needed?

server/routes/explore/api.py
```diff
@@ -115,7 +115,9 @@ def detect_and_fulfill():
   data_dict = copy.deepcopy(utterance.insight_ctx)
   utterance.prev_utterance = None
   data_dict[Params.CTX.value] = serialize.save_utterance(utterance)
-
+  data_dict[Params.ENABLE_NL_FULFILLMENT] = request.get_json().get(Params.ENABLE_NL_FULFILLMENT, "")
+  data_dict[Params.EXP_MORE_DISABLED] = request.get_json().get(Params.EXP_MORE_DISABLED, "")
```
pradh, line 119: Not sure if you want to use `.value` to be sure?  I'm never sure the enums just work

pradh (commented):
Very nice!  Bo should take a close look, and defer to Bo for approval.

server/lib/nl/detection/detector.py
chejennifer: when calling fulfill on the client code, we pass in the place type from the saved context (unless we don't want to do this?) & the passed in place type could end up being the same type as the place, which will then cause state.place_type to get set to None here: https://github.com/datacommonsorg/website/blob/f8b5fec8ecd06a758c3cfc8cfa8b2ada21ec034c/server/lib/nl/fulfillment/utils.py#L141 which then eventually, an error gets thrown when trying to access state.place_type

server/routes/explore/api.py
chejennifer: done

static/js/apps/explore_landing/topics.json
```diff
@@ -743,61 +743,61 @@
...
           },
           {
             "title": "Compare SDG goals on climate change between Brazil, China and India",
-            "url": "oq=Compare+SDG+goals+on+climate+change+between+Brazil,+China+and+India&dc=sdg"
+            "url": "q=Compare+SDG+goals+on+climate+change+between+Brazil,+China+and+India&dc=sdg"
           },
           {
             "title": "Violence vs poverty across the world",
-            "url": "oq=Violence+vs+poverty+across+the+world&dc=sdg"
+            "url": "q=Violence+vs+poverty+across+the+world&dc=sdg"
```
shifucun, line 800: @pradh any links with oq we have distriubuted?

pradh: it might be, not sure.  if its a one-liner, map oq to q?

chejennifer: sure

static/js/apps/explore/app.tsx
```diff
@@ -128,339 +94,224 @@ export function App(): JSX.Element {
...
+          })}`;
+        }
       }
-      if (!topic) {
-        updateHash({
-          t: DEFAULT_TOPIC,
-          dc,
-        });
-        return;
+      if (shouldSetQuery && chartData.mainTopic?.name && chartData.place.name) {
```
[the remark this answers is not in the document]
chejennifer: yes

static/js/apps/explore/app.tsx
```diff
@@ -128,339 +94,224 @@ export function App(): JSX.Element {
...
+    return (
+      <Container className="explore-container">
+        <div>
+          <Spinner isOpen={true} />
+        </div>
+      </Container>
+    );
+  }
+  if (loadingStatus === LoadingStatus.SUCCESS && chartData) {
+    const childPlaceType = Object.keys(chartData.childPlaces)[0];
```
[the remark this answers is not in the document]
chejennifer: good catch, added a check

shifucun (approved):
thanks!

chejennifer:
thanks for the reviews!

server/integration_tests/test_data/e2e_answer_places/californiacountieswiththehighestasthmalevels/chart_config.json
```diff
@@ -8,26 +8,26 @@
...
-                    "rankingTileSpec": {
-                      "rankingCount": 10,
-                      "showHighest": true
-                    },
                     "statVarKey": [
                       "Percent_Person_Children_WithAsthma"
                     ],
                     "title": "Percent of Children That Have Asthma in Counties of California (${date})",
-                    "type": "RANKING"
+                    "type": "MAP"
```
pradh, line 15: Why is this changing?

server/lib/nl/detection/detector.py
pradh: What is the eventual error that gets thrown?  Because the setting of state.place_type to None is intentional, and that downstream error should be fixed.

If user has not asked for "place_type" then it should be empty, and we'll default to DEFAULT_TYPE.

If the user has said "countries in USA", then we set place_type to empty, and only show US-level data.

chejennifer: reverted & updated the frontend code as we talked about offline

server/integration_tests/test_data/e2e_answer_places/californiacountieswiththehighestasthmalevels/chart_config.json
chejennifer: updated

static/js/apps/explore/app.tsx
```diff
@@ -128,339 +94,228 @@ export function App(): JSX.Element {
...
+    const childPlaceType = !_.isEmpty(chartData.childPlaces)
+      ? Object.keys(chartData.childPlaces)[0]
+      : "";
+    const placeUrlVal = (
+      exploreContext?.entities || [chartData.place.dcid]
+    ).join(DELIM);
+    const topicUrlVal = (exploreContext?.variables || []).join(DELIM);
+    const relatedPlaceTopic = _.isEmpty(chartData.mainTopic)
+      ? {
+          dcid: exploreContext?.variables,
```
pradh, line 135: This can be a list?  do you mean to set it to just dcid?

static/js/components/subject_page/block_container.tsx
```diff
@@ -99,7 +100,12 @@ export function BlockContainer(props: BlockContainerPropType): JSX.Element {
     const result: Item[] = [];
     for (const prop in svExtendedMap) {
       const urlSv = svExtendedMap[prop].join(DELIM);
-      const url = `/explore/#t=${urlSv}&p=${exploreData.place}&pcmp=${exploreData.cmpPlace}&pt=${exploreData.placeType}&dc=${exploreData.dc}&em=1`;
+      const url = `/explore/#${getUpdatedHash({
+        t: urlSv,
+        p: exploreData.place,
+        q: "",
+        em: "1",
+      })}`;
```
pradh, line 108: is DC no longer passed here?

static/js/apps/explore/result_header_section.tsx
```diff
@@ -0,0 +1,80 @@
...
+      if (topic.dcid == DEFAULT_TOPIC) {
+        // Do not show the root topic.
+        continue;
+      }
+      topicList.push({
+        text: topic.name,
+        url: `/explore/#${getUpdatedHash({
+          [URL_HASH_PARAMS.TOPIC]: topic.dcid,
+          [URL_HASH_PARAMS.PLACE]: props.placeUrlVal,
+          [URL_HASH_PARAMS.QUERY]: "",
```
pradh, line 56: is DC not passed here?

static/js/components/subject_page/block_container.tsx
chejennifer: getUpdatedHash will keep all params the same and update the ones you pass in

static/js/apps/explore/app.tsx
chejennifer: this was following what the app used to do: https://github.com/datacommonsorg/website/blob/65909ffa310fbb6ec191693f7dd2f6832c969e5f/static/js/apps/explore/app.tsx#L430-L438

is taht something we wanted to change?

chejennifer: oh, actually, need to use topicUrlVal, good catch

PR #3278, opened 2023-08-22
Use places in prior answer to answer follow on queries

pradh:
Add ability ask follow up queries based on answers of prior queries.  Also, support per-capita queries.

This works for:
- Ranking (highest X in)
- Time delta (growth the most in)
- Filter (X is over 1000 in)

Example: https://screencast.googleplex.com/cast/NjU3NzUxNDExMjYxNDQwMHwwNDE4ZWUzYS1kMA


jehangiramjad (approved):
Very cool!

PR #3285, opened 2023-08-23
Enable expando at block level

shifucun:
With the recent layout change, there is only one tile per column. So the expando should be at block level instead of column level now.

### DIFF
--- a/static/js/components/subject_page/block.tsx
+++ b/static/js/components/subject_page/block.tsx
@@ -120,13 +120,10 @@ export function Block(props: BlockPropType): JSX.Element {
           props.columns.map((column, idx) => {
             const id = getId(props.id, COLUMN_ID_PREFIX, idx);
             const columnTileClassName = getColumnTileClassName(column);
-            let className = "";
-            if (idx >= minIdxToHide) {
-              className = HIDE_TILE_CLASS;
-            }
+            const shouldHideColumn = idx >= minIdxToHide;
             return (
               <Column
-                className={className}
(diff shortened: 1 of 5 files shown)


static/js/components/subject_page/block.tsx
```diff
@@ -119,8 +120,13 @@ export function Block(props: BlockPropType): JSX.Element {
           props.columns.map((column, idx) => {
             const id = getId(props.id, COLUMN_ID_PREFIX, idx);
             const columnTileClassName = getColumnTileClassName(column);
+            let className = "";
+            if (idx >= minIdxToHide) {
+              className = HIDE_TILE_CLASS;
+            }
             return (
               <Column
+                className={className}
```
chejennifer, line 129: nit: make a new `shouldHideTiles: boolean` prop instead so that it's clear what's happening?

static/js/components/subject_page/block.tsx
```diff
@@ -138,10 +144,30 @@ export function Block(props: BlockPropType): JSX.Element {
             );
           })}
       </div>
+      {isNlInterface() && props.columns.length > NL_NUM_BLOCKS_SHOWN && (
```
chejennifer, line 147: I'm slightly confused, you mention in the comments of this PR that there is only one column per block .. so why are we checking length of columns?

Can you confirm when the expando should show up and what the layout in terms of tiles/columns/blocks is?

shifucun (commented):
thanks!

chejennifer (approved):
nice

PR #3287, opened 2023-08-23
[disaster block] update date for disaster tiles

chejennifer:
- allow configuring date for a disaster block
- keep state of date & filters in the block instead of in the url
- update default date range to last 3 years for all places except for earth which will keep default of last year


static/js/components/subject_page/disaster_event_block.tsx
```diff
@@ -108,55 +134,79 @@ export const DisasterEventBlock = memo(function DisasterEventBlock(
...
+    }
+    if (!_.isEqual(prevDate.current, date)) {
+      updatedParams[URL_HASH_PARAM_KEYS.DATE] = date;
+      prevDate.current = date;
+    }
+    const updatedHash = getUpdatedHash(updatedParams, props.id);
+    const currentHash = location.hash.replace("#", "");
+    const urlRoot = window.location.href.split("#")[0];
+    if (updatedHash && updatedHash !== currentHash) {
+      history.pushState({}, "", `${urlRoot}#${updatedHash}`);
```
shifucun, line 178: if there are multiple blocks in a page, would they result in multiple history state?

server/config/topic_page/dev/CA.textproto
```diff
@@ -22,6 +22,44 @@ metadata {
     key: "State"
     value: "County"
   }
+  event_type_spec {
```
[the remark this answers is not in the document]
chejennifer: yes

static/js/components/subject_page/disaster_event_block.tsx
chejennifer: multiple blocks will all push to the same history state

server/config/subject_page.proto
```diff
@@ -288,6 +288,22 @@ message Tile {
...
+message DisasterBlockSpec {
+  enum DateRange {
+    TYPE_NONE = 0;
+    THIRTY_DAYS = 1;
+    SIX_MONTHS = 2;
+    ONE_YEAR = 3;
+    THREE_YEARS = 4;
+  }
+  // [optional] If set, use as default date. Otherwise, default date is
+  // THREE_YEARS for all places except for Earth, which defaults to ONE_YEAR.
```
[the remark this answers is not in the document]
chejennifer: defaults are set in disaster_event_map_constants.ts

pradh:
Do we maybe need to change this too accordingly: https://github.com/datacommonsorg/website/blob/87114281c4d12c265213cc3c50cd661950dcc0e6/server/lib/nl/common/utils.py#L52-L68 ?

static/js/components/subject_page/disaster_event_block.tsx
shifucun: Why is this not at the app level? I would imagine component should not make page level state change?

chejennifer: this only updates the hash with the current block's changes .. and the page doesn't know the information about the block level controls so it'll also be tricky to do the update at the page level. Also wondering if we even need to update the hash because that's just so we can copy + paste the link, the app doesn't actually read state from the hash

shifucun: This is not blocking this PR but maybe we can talk more about this offline.

chejennifer:
> Do we maybe need to change this too accordingly:
> 
> https://github.com/datacommonsorg/website/blob/87114281c4d12c265213cc3c50cd661950dcc0e6/server/lib/nl/common/utils.py#L52-L68
> 
> ?

good call, updated to check last 3 years

chejennifer:
thanks for the reviews!

server/lib/nl/common/utils.py
```diff
@@ -60,11 +60,14 @@ def event_existence_for_place(place: str, event: types.EventType,
     counters.timeit('event_existence_for_place', start)
     cur_year = datetime.datetime.now().year
     prev_year = str(cur_year - 1)
+    two_years = str(cur_year - 2)
```
pradh, line 63: Thanks for this.

Should we match the Earth behavior?  Earth also might have a ton of data.

pradh (approved):

server/lib/nl/common/utils.py
chejennifer: Sorry, what do you mean? This change just adds two more years to the list of allowed years in the check & doesn't add more data processing?

PR #3289, opened 2023-08-24
Add a text-box tile & web component for the UN SDG tracker

juliawu:
Adds stories to the UN SDG Tracker pages via a new Text Tile

More specifically:
* Adds a new "Text" tile that displays some text with a title
* Adds a web component wrapper around the new tile
* Adds UN SDG stories to the JSON config files of the sdg-tracker
* Adds logic to show the stories using a text tile in the sdg tracker
* Updates types in sdg-tracker to match new fulfillment return structure
* Brings back the headers after fulfillment return structure changes.

Note that:
* Right now the tile takes raw strings to display. Logic to fetch strings from the KG will be implemented in a future PR after the schema for text/stories is set
* The UN stories are only shown if "Earth" is selected.
* The UN stories show under the goal headings, but the goal headings are missing in the "All SDGs" page. We should add them back, in a followup PR.

Screenshots:
Text starts out collapsed:
Screenshot 2023-08-23 at 4 59 31 PM

Text expands after hitting "Show more"
Screenshot 2023-08-23 at 4 59 45 PM


dwnoble (approved):
Nice!

dwnoble (approved):

PR #3290, opened 2023-08-24
NL: bunch of fixes, add India Demo script

pradh:
This PR includes:

1. With TIME_DELTA, support for not just increase/decrease but also change over time (test coverage)
2. Fix answer-places bug where it is place-type dependent, which trips non-US places that have a query type (district) different from KG type (AA2).  Net: be less strict about matching types for answer-places, there's just one list only (test coverage)
3. Fix bug with contained-in for non-US places (again due to the above reason).... [literacy across districts of India] now shows map.
4. Show SDG map charts (test coverage)
5. If user asked for highest/lowest, show ranking chart first on Explore (test diffs)
6. Fallback from time-delta to place-comparison... so if we're comparing SVs that have single data points we'll show bars.
7. Propagate counters and debug-logs from detection to fulfillment in explore.  After the detect-and-fullfill change, we were only return fulfill logs.
8. Fix a minor bug with incorrect correlation interpretation.
9. Fix a minor bug where if place_type is empty in Explore (under edge cases) we don't crash.
10. Add India Explore demo script


shifucun (approved):

pradh:
Thanks Bo!

PR #3291, opened 2023-08-24
Fix a performance issue for fetching map point names; Run screenshot test in parallel

shifucun:
* Right now facilities fetch parent place via place info cache. In USA map, this involves around 10K facilities and result in server error. Use regular name instead.

* Refactor the screenshot a bit to allow parallel run.


pradh (approved):
Nice!

PR #3294, opened 2023-08-24
NL: a few fixes incl per capita for answer-places

pradh:
This PR includes:

1. Per capita computation in the backend for ranking for answer-places ([states with the highest poverty levels per capita])
2. Track whether child-type came from a default-type or was user-specified. If latter, show map before line chart, and prefer time-delta-across-vars over time-delta-across-places.  Example query: [which industries have grown the most in california] in Explore.
3. Perform existence checks always with single-point.  When time-delta-across-places falls back to basic, we need that to not show single-dot timeline.  Example query: [how has literacy grown in states of india]
4. Allow for answer-place propagation to work even when we come with multiple places (it was incorrectly suppressed).


jehangiramjad (approved):

pradh:
Thanks Jehangir!

PR #3296, opened 2023-08-24
[sdg-tracker] downgraded dependencies

dwnoble:
Downgraded `react`, `react-router-dom` and `antd` to match UN versions from https://code.officialstatistics.org/undata2/data-commons/frontend-tester/-/blob/main/UNSD-Website-skeleton/UNSD.Website/ClientApp/package.json


juliawu (approved):
Thanks for handling this!

PR #3297, opened 2023-08-24
[explore] add autoplay

chejennifer:
- adds autoplay
    - url param to specify to queries: aq, with ___ as the delimiter between queries
    - to disable typing, use &at=1
    - to manually click enter after each query is inputted, use &ae=1
- extracted out some components from the main App component. Created new ErrorResult and SuccessResult components

[screen-recording - 2023-08-24T130539.840.webm](https://github.com/datacommonsorg/website/assets/69875368/464c2107-86e8-4f4a-bc22-38c04e808c64)


shifucun (approved):
very nice!!

shifucun:
not for this PR, but it would be nice that the spinner is transparent instead of showing a gray page , wdyt?

chejennifer:
thanks for the review! Transparent spinner could be nice, maybe add that to the NL Launch task list, but I would say that's at most a P1?

pradh (commented):
Thank you!

PR #3302, opened 2023-08-25
feat: add changes to display chart header only if data exists

suhana13:
For the chemical-gene associations, some of the genes have zero values for their associations, which leads to displaying a chart like this. 
Screenshot 2023-08-25 at 11 34 42 AM

We want to make sure that if we have an array of zeroes, we return a null array so it doesn't display the chart header or the chart.
Screenshot 2023-08-25 at 11 35 00 AM


static/js/biomedical/protein/data_processing_utils.ts
```diff
@@ -528,7 +544,9 @@ export function getChemicalGeneAssoc(data: GraphNodes): ProteinNumData[] {
     for (const m in chemAssoc) {
       result.push({ name: m, value: chemAssoc[m] });
     }
-
+    if (check_for_zeroes(result)) {
```
chejennifer, line 547: nit: don't really need the helper function because could just check that there is at least one value that is not zero like:
```
const hasNonZero = result.findIndex((obj) => obj.value !== 0) >= 0;
if (!hasNonZero) {
...
}
```

static/js/biomedical/protein/page.tsx
```diff
@@ -83,39 +94,50 @@ export interface InteractingProteinType {
 export class Page extends React.Component<PagePropType, PageStateType> {
   constructor(props: PagePropType) {
     super(props);
-    this.state = { data: null };
+    this.state = {
+      data: null,
```
chejennifer, line 98: nit: you might not need to keep data in the state anymore since you're keeping all the processed data in the state?

chejennifer (approved):
nice, just some nits

static/js/biomedical/protein/data_processing_utils.ts
suhana13: great suggestion! done!

static/js/biomedical/protein/page.tsx
suhana13: we're using `this.state.data` for protein description and protein link in lines 148-149!

PR #3303, opened 2023-08-25
Update the NL Search Bar

chejennifer:
- TextSearchBar was only used by NL, so extract that to an NlSearchBar & add a stylesheet specific for this component
- update styling according to: https://www.figma.com/file/6PEivw1xQeu4JznkEqj50Y/AO---Data-Commons?type=design&node-id=36-328&mode=design&t=xgIUzdeMV3LTgAG7-0
- update link colors

[screen-recording - 2023-08-25T115841.044.webm](https://github.com/datacommonsorg/website/assets/69875368/d992fa6f-69e2-4244-86aa-5f59e57b7cba)


static/css/base.scss
```diff
@@ -21,6 +21,9 @@ $full-footer-bg: #fafcff;
 
 :root {
   --font-family: "Roboto", "Helvetica Neue", Arial, sans-serif !important;
+  --font-family-sans-serif: "Google Sans Text", -apple-system, BlinkMacSystemFont,
```
dwnoble, line 24: Maybe use the variable `$font-family-sans-serif` here?

dwnoble (commented):
Nice!

dwnoble (approved):

static/css/base.scss
chejennifer: actually, don't need to redefine it here because setting $font-family-sans-serif already defines it as a variable!

chejennifer:
thanks for the review!

PR #3304, opened 2023-08-25
Add a "max variables" attribute to bar tile and web component

juliawu:
This PR allows for setting a maximum number of variables to show for each grouping in bar charts via a new `maxVariables` prop in the bar tiles & web component. 

Note that:
* If no value is set, all provided variables are shown.
* If multiple places are provided, the variables will be sorted by the value for the first place

Ex:
```
<datacommons-bar
  header="Lots of places, lots of variables"
  parentPlace="country/USA"
  childPlaceType="State"
  variables="Count_Person_1To4Years Count_Person_5To17Years Count_Person_18To24Years Count_Person_25To34Years Count_Person_35To44Years Count_Person_45To54Years Count_Person_55To64Years Count_Person_65To74Years Count_Person_75OrMoreYears"
  maxPlaces="5"
  maxVariables="5"
  sort="ascending"
></datacommons-bar>
```
Screenshot 2023-08-25 at 3 24 18 PM


server/config/subject_page.proto
```diff
@@ -182,15 +182,17 @@ message BarTileSpec {
   bool horizontal = 4;
   // maximum number of places to display
   int32 max_places = 5;
+  // maximum number of variables to display
+  int32 max_variables = 6;
   // bar sort order
```
pradh, line 187: nit: not that it matters here, but its generally good practice to not change existing tag numbers :)

pradh (approved):
Very nice!

Could you add me to the task/bug as a reminder for making the backend change?

PR #3305, opened 2023-08-25
[Bug fix] fix incorrect horizontal bar chart bar lengths

juliawu:
Fixes the bar lengths on horizontal bar charts, which were too long, resulting in incorrect values plotted.

Before (Note value in tooltip vs where top bar is on X axis):
Screenshot 2023-08-25 at 4 19 49 PM

After:
Screenshot 2023-08-25 at 4 19 24 PM


dwnoble (approved):
nice!

PR #3308, opened 2023-08-25
Use observation facet API to check single point observation

shifucun:
For query "What are the projected temperature extremes across california", the check goes from 22 seconds to 0.2 seconds.


pradh (approved):
Woohoo!

PR #3309, opened 2023-08-26
NL: a bunch of fixes

pradh:
This PR fixes a bunch of things:

1. Sets "answer places" from place-comparison handler.  This is so that we can string together not just 'time-delta' queries, but also comparison queries (this is relevant for the electrification demo).  One implication of this change is that in "Top Foo places" => "emissions there" => "GDP there", if the 2nd query only had data for a subset of places, then GDP will be for that subset only (not the first result).  I think this is reasonable/OK.  This explains the deletions in the golden diffs.

2. Improve existence checks for bar-chart place comparison

3. Fix place fallback for Explore flow.  Example query: [emissions in houston].  The issue was two-fold:  (a) because explore has a default child-type (and we got a default child-type of COUNTY, which we shouldn't), we were doing child-type fallback, and for that, the fix is to not do child-type fallback unless user has specified the child-type;  (b) the main place being returned in the response was the queried place, not the fallback place, so fix that.

4. As part of fixing (3b), unify response generation further for NL and Explore.  Also, make the default child-types same between NL and Explore (previously for USA they were different because of place-type fallback).

5. Opt: Reduce the number levels of SVG lookup, without much loss of Explore More function (as seen by lack of diffs).

6. Propagate the `userMessage` to the user.  For now early failure from backend is propagated.  Upcoming PRs will set explanation texts (when charts show up).

7.  Minor:  suppress block title for place-overview, since the place is also there in page title.

8.  Infer "place comparison" more proactively when there are multiple places (so we don't fallback from var comparison).

9.  Edge case: avoid duplicate SVs and places in the recognized set.  Add tests for these.

10. Fix a crash bug (Top sources of CO2 emissions in the US) by checking for place_type first.

11.  Fix a bug where NL was incorrectly going down the path of a correlation-across-vars query when in fact there were multiple places (see fix in `fulfiller.py`).  This was on [bachelor majors in austin vs. berkeley vs. cambridge]

12.  Fallback from var-correlation or place-comparison to basic.  Since var-correlation uses a different topic-expansion approach it doesn't make much sense now (i.e., post explore migration).

13. Minor tweaks to some error messages.


nl_server/embeddings.py
```diff
@@ -242,6 +242,9 @@ def _detect_multiple_svs(self, query: str) -> vars.MultiVarCandidates:
           # so drop this candidate.
           continue
 
+        if not vars.deduplicate_svs(candidate):
```
jehangiramjad, line 245: Why is this needed?

jehangiramjad (approved):
This is really nice!

nl_server/embeddings.py
pradh: To dedup SVs that show up in multiple parts of the candidate.  Updated test and added comment.

pradh (commented):
Thanks Jehangir!

PR #3311, opened 2023-08-26
NL: Improve/fix a few things

pradh:
1.  For SDG the default place is Earth
2.  For correlation query results, also include ranking + map for every chart-var below scatter charts.
3.  For correlation query, show "X vs Y" on the Explore page on the title and search box.  Also return peer and parent and child topics from both sides.  This change requires making "mainTopic" a repeated thing, and propagating all thru the stack.  Thus the many files touched!
4. Avoid using place-type from context for correlation query.  This leads to non-intuitive behavior, like [poverty across the world] > [diabetes vs. asthma in california], makes the 2nd query look for place-type country.

localhost_8080_explore_ (4)


jehangiramjad (approved):

pradh:
Thanks Jehangir!

static/js/apps/explore/app.tsx
```diff
@@ -197,11 +197,16 @@ export function App(props: { isDemo: boolean }): JSX.Element {
...
       if (
         shouldSetQuery &&
-        pageMetadata.mainTopic?.name &&
+        !_.isEmpty(pageMetadata.mainTopics) &&
         pageMetadata.place.name
       ) {
-        const q = `${pageMetadata.mainTopic.name} in ${pageMetadata.place.name}`;
-        setQuery(q);
+        if (pageMetadata.mainTopics.length == 2) {
+          const q = `${pageMetadata.mainTopics[0].name} vs. ${pageMetadata.mainTopics[1].name} in ${pageMetadata.place.name}`;
```
chejennifer, line 204: will topics always have names? In the old code, we checked for a name before actually setting the query so not sure if that was because sometimes topics might not have a name?

chejennifer (commented):
nice!

static/js/apps/explore/app.tsx
pradh: sg, I can make that change in follow up...

PR #3315, opened 2023-08-28
NL: Provide user feedback

pradh:
Return user feedback strings per [this doc](https://docs.google.com/document/d/13-1zsYZm_RFkzhKb4GxVy7DQNU8ZiEYZSY8001R-GsU/edit).

Move the string gen impl from frontend (only supported in NL) to the backend (now supported in both NL and Explore), so it gets validated in the e2e tests too.

As part of this, simplify the detect+fulfill path, which previously did independent detect and fulfill tasks.  In the new approach, we don't turn things into `insight_ctx` between those stages (i.e., we don't call `construct_for_explore`), instead we have a simpler function (`prepare_for_explore`) that prepares the utterance for Explore (by attaching child/peer places, and setting default child-type).  This was done to avoid passing more state (eg, sv_source, place_source, etc. which are needed for proper user feedback).

Finally, fix some duplicate place additions.

DIFFS:  Generally, they seem reasonable.  One meaningful diff is that the "poverty vs. poverty in CA" query now compares female vs. male poverty.

SCREENSHOTS:

image

image


jehangiramjad (approved):
Looks great (leaving the tsx files review to @shifucun / @chejennifer )

pradh:
> Looks great (leaving the tsx files review to @shifucun / @chejennifer )

Sorry had automerge -- @shifucun @chejennifer -- could you let me know if there's anything in the tsx changes (just a few lines)?

PR #3317, opened 2023-08-28
Add setting to prevent per-capita toggle from affecting some variables

juliawu:
Adds a `noPerCapita` attribute to statVarSpecs, which when set, will prevent the per capita toggle on subject pages from dividing their value by population. This is helpful for (1) allowing the per capita toggle to only change one of the axes on a scatter plot (by preventing the change on an axis that doesn't need it) and (2) preventing per capita on variables that are already per capita.

Note, this still needs the following backend changes to show up in explore results:
(1) We should enable the per capita toggle on comparison pages with scatter plots
(2) Fulfillment needs to emit "noPerCapita = true" in the statVarSpecs we wish to prevent from toggling per capita for.


chejennifer (approved):
nice!

PR #3318, opened 2023-08-28
[explore] various UI fixes

chejennifer:
- update show more to be more prominent according to figma mocks: https://screenshot.googleplex.com/BXs4PiHDr9bSYZJ
- update highlight tiles according to: https://www.figma.com/file/6PEivw1xQeu4JznkEqj50Y/AO---Data-Commons?type=design&node-id=722-2766&mode=design
- fix finnicky scatter plot tooltips


chejennifer:
thanks for the review!

PR #3319, opened 2023-08-28
[library] scatter plot improvements

juliawu:
This PR updates the scatter plot web components to match behavior and styling of scatter plots on our website. This will improve behavior of scatter plots for files in our experimental folder (like the UN sdg tracking page).

Changes:
1. Fetches the stat var names if one is not provided in StatVarSpec. This fixes the axes labels in the scatter plot web component, and brings behavior in line with the other tiles/web components.
Before:
Screenshot 2023-08-28 at 1 13 53 PM

After:
Screenshot 2023-08-28 at 12 18 26 PM

2. Fix tooltip styling in web components
Before:
Screenshot 2023-08-28 at 1 13 35 PM

After:
Screenshot 2023-08-28 at 12 33 16 PM

3. Allows the user to optionally set some variables as "per capita" by providing the list of variables to make per capita. Note that this does not yet add an indication this is per capita. We should discuss how we want to show this on the chart.
Ex:
```
<datacommons-scatter
  header="Population vs Median Household Income for US States"
  variables="Annual_Generation_Electricity Median_Income_Household"
  parentPlace="country/USA"
  childPlaceType="State"
  usePerCapita="Annual_Generation_Electricity"
></datacommons-scatter>
```
Screenshot 2023-08-28 at 12 18 32 PM

### DIFF
--- a/static/css/shared/_tiles.scss
+++ b/static/css/shared/_tiles.scss
@@ -175,6 +175,29 @@ $box-shadow-8dp: 0 8px 10px 1px rgba(0, 0, 0, 0.14),
   right: 0;
 }
 
+#scatter-tooltip {
+  background-color: var(--dc-gray-lite);
+  box-shadow: 0 3px 6px 0 rgba(0, 0, 0, 0.2);
+  font-size: 10pt;
+  padding: 10px;
+  border-radius: 8px;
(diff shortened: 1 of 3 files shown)


chejennifer (approved):
nice!

PR #3320, opened 2023-08-28
Fix explore more iteration bug

shifucun:
Apparently the length of the HTMLCollection object changes as class list is updated. That causes the issue that only a subset of the divs have class updated.


pradh (approved):
Thanks!

PR #3324, opened 2023-08-28
[explore] add new user message component

chejennifer:
- added a basic user message component that currnetly takes in the message and whether or not to show form. Will update the component as flask response is updated
- remove the "see more ..." in the place overview tile

e.g.,
Screenshot 2023-08-28 at 3 56 00 PM


pradh (approved):
Neat!

The showForm bit is being set in https://github.com/datacommonsorg/website/pull/3325.

PR #3325, opened 2023-08-28
NL: Tweak scatter/bar chart bits and user-messages

pradh:
1. Set titles and no-per-capita bit for scatter plots.  Example: https://screencast.googleplex.com/api/gif/NTExMjUzNTY2NDU1ODA4MHxjMWZhODU3Yy05Nw/image**
2. Limit number of bars, and also sort by default so the prominent vars show up.
3. Fix up messages based on feedback from Jehangir and Prem:
    * Don't show "inappropriate words" msg as a special thing
    * Trim some wording
4. Delete the deprecated explore flow (and the nlFulfillment param)
5. Return a `showForm` bit for frontend user-message module.
6. Minor followups from https://github.com/datacommonsorg/website/pull/3311.


static/js/constants/app/explore_constants.ts
```diff
@@ -25,7 +25,6 @@ export const URL_HASH_PARAMS = {
   QUERY: "q",
   DC: "dc",
   DISABLE_EXPLORE_MORE: "em",
```
shifucun, line 27: shall we remove "em" as well?

shifucun (approved):
nice deletion!!

static/js/constants/app/explore_constants.ts
pradh: ok. I'll do in follow up.

PR #3327, opened 2023-08-28
Homepage mobile-view tweaks: topic tiles and footer

juliawu:
### 1. Updates topic tile sizes in mobile view to be smaller and without descriptions so that two tiles can be shown on the same row.
Before:
Screenshot 2023-08-28 at 4 46 05 PM

After:
Screenshot 2023-08-28 at 4 45 47 PM

### 2. Converts the footer in mobile to two rows so it doesn't get cut off at the bottom of the page. 

Before:
Screenshot 2023-08-28 at 4 44 43 PM

After:
Screenshot 2023-08-28 at 4 44 31 PM


chejennifer (approved):
thanks!

PR #3328, opened 2023-08-29
Remove footer from unsdg base template.

keyurva:
* On the main site, the footer is conditionally displayed (e.g. it is not displayed on the nl page).
* With custom DC, the base template (base.html) is applied for all pages.
* Removing footer from unsdg until we discuss and figure out conditional templates for custom DCs.
* Screencast of unsdg DC pages without footer below.

[no-footer.webm](https://github.com/datacommonsorg/website/assets/1221814/a4990de5-5c58-4e3f-b461-21ddb3ff2512)


shifucun (approved):
thanks!

PR #3332, opened 2023-08-29
[explore] use the show form bit

chejennifer:
- use the `showForm` bit from flask response to determine whether or not to show a form in the user message


pradh (approved):
Thanks!

PR #3333, opened 2023-08-29
Split integration tests into two tests and run them in parallel in cloudbuild

shifucun:
Also made a few changes:

1. Do not run pip install in every test commands, and make them optional running
2. Enable using a random port for nl server, this will make it possible to run multiple tests together and without starting the local nl server.
3. Run integration test/golden update tests with pytest parallelization. 

NL integration tests now takes 4 min:

image


pradh (approved):
Very nice!

nl_server/__init__.py
```diff
@@ -28,6 +30,10 @@ def create_app():
 
   flask_env = os.environ.get('FLASK_ENV')
 
+  # https://github.com/UKPLab/sentence-transformers/issues/1318
+  if sys.version_info >= (3, 8) and sys.platform == "darwin":
+    torch.set_num_threads(1)
```
shifucun, line 35: @jehangiramjad This solves the issue!

PR #3334, opened 2023-08-29
[Disease Browser]: The Disease Prevalence Rate Map

suhana13:
This PR fixes the disease prevalence rate map on the disease browser page. 
Screenshot 2023-08-30 at 12 32 38 PM


chejennifer:
maybe update the styling to make the map section look better? Looks kind of odd to have everything so clustered on the left side of the page .. maybe move the legend to the right

static/css/biomedical/disease.scss
```diff
@@ -60,3 +61,26 @@ tbody tr:nth-child(even) {
   stroke-width: 1.5;
   border: 3px solid #000;
 }
+
+.map {
```
chejennifer, line 65: do you still need all these style changes after importing the tiles sheet?

chejennifer (approved):
nice! Before you merge, can you check if you even need the style additions you had made & if not, remove them

static/css/biomedical/disease.scss
suhana13: removed the unnecessary css elements!

PR #3342, opened 2023-08-30
Report a low-confidence msg < certain threshold

pradh:
image


server/lib/nl/detection/utils.py
```diff
@@ -96,6 +96,14 @@ def get_multi_sv_pair(
   return parts
 
 
+def get_top_sv_score(detection: Detection) -> float:
+  if is_multi_sv(detection):
```
jehangiramjad, line 100: the return type is a float so shouldn't this return something? else, the return type hint could be a Union[float, None] or something like that?

jehangiramjad (approved):
Thanks!

server/lib/nl/detection/utils.py
pradh: Sorry, not following the Q, it is returning the score, or 0?

pradh: I'll merge now and handle something I missed based on your response

jehangiramjad: Ooops sorry. Too late in the night. My screen was showing me a line break so it seemed like it was just a dangling return statement... my bad!

PR #3345, opened 2023-08-30
[explore] Update error messages

chejennifer:
- show topics in the error message (only for cases where the only tile returned is a place overview tile)
- put in the url for the response form
- also fixed css problem with sticky search bar

Screenshot 2023-08-30 at 10 39 50 AM


static/js/apps/explore/success_result.tsx
```diff
@@ -56,11 +56,28 @@ interface SuccessResultPropType {
...
+      for (const column of block.columns) {
+        for (const tile of column.tiles) {
+          if (tile.type !== "PLACE_OVERVIEW") {
+            return false;
+          }
+        }
+      }
+    }
+  }
+  return true;
```
pradh, line 72: nit: should we do:
if (#categories == 1 && #block-in-category == 1 && #column-in-block == 1 && #tile-in-column == 1 && tile.type == "PLACE_OVERVIEW") return true;

?

pradh (approved):
Nice!

static/js/apps/explore/success_result.tsx
chejennifer: oh, that's nicer, yes, updated!

chejennifer:
thanks for the review!

PR #3346, opened 2023-08-30
Add source url column to sanity test results.

keyurva:


server/webdriver/tests/standalone/sanity.py
```diff
@@ -69,6 +74,7 @@ def __init__(self, page: WebPage, status: str, comments: str = "") -> None:
     self.title = page.title
     self.status = status
     self.url = page.url
+    self.source_url = page.source_url
```
shifucun, line 77: Can we make the chrome driver headless at line 94 like, so this can ran in the background

    chrome_options = Options()
    chrome_options.add_argument("--headless=new")
    self.driver = webdriver.Chrome(service=Service(
        ChromeDriverManager().install()), options=chrome_options)

shifucun (approved):
Nice! thanks for adding this!

server/webdriver/tests/standalone/sanity.py
keyurva: Done.

PR #3363, opened 2023-08-31
[vis tool] fix scatter vis tool tooltip

chejennifer:
- the spinner was covering the chart and not allowing for hover


shifucun (approved):
thanks!

PR #3364, opened 2023-08-31
Update highlight tile unit display for large per-capita percentages.

juliawu:
Updates the styling of the units on the highlight tile, so that when per-capita results in a large percentage, we show the raw value instead of a percentage. This prevents us from showing something like "1.3K%" to the user instead of the actual value "13".

Before:
Screenshot 2023-08-31 at 1 40 05 PM

After:
Screenshot 2023-08-31 at 1 39 08 PM

### DIFF
--- a/static/js/components/tiles/highlight_tile.tsx
+++ b/static/js/components/tiles/highlight_tile.tsx
@@ -121,6 +121,7 @@ export function HighlightTile(props: HighlightTilePropType): JSX.Element {
   const [highlightData, setHighlightData] = useState<HighlightData | undefined>(
     null
   );
+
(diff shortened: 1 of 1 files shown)


chejennifer (approved):
thanks for the update! My next task was actually going to be to handle per-capita percentages for all the tile types .. but feel free to merge, I just will probably be overriding these changes

juliawu:
> thanks for the update! My next task was actually going to be to handle per-capita percentages for all the tile types .. but feel free to merge, I just will probably be overriding these changes

In that case, going to close, treatment for all the tile types will be better, and this saves you from having to undo the change.

PR #3365, opened 2023-08-31
Increase place threshold and change default child-type

pradh:
This fixes the Justice system equity query.

Note that I'd switched to County from State as a default assuming fallback, but then when user doesn't specify place-type we don't do fallback.

image


jehangiramjad (approved):
Thanks!

PR #3366, opened 2023-08-31
Bar tile: allow ascending and descending sort in single variable case

juliawu:
Updates bar tile so that in the single variable case, setting sort to "ascending" or "descending" will sort by places.

Ex:
```
<datacommons-bar
  header="Number of Farms"
  parentPlace="country/USA"
  childPlaceType="State"
  variables="Count_Farm"
  maxPlaces="15"
  sort="descending"
>
```
Screenshot 2023-08-31 at 3 55 33 PM


static/js/components/tiles/bar_tile.tsx
```diff
@@ -283,21 +283,30 @@ function rawToChart(
...
   if (props.sort === "ascending" || props.sort === "descending") {
-    // sort variables in first group by value
-    dataGroups[0].value.sort(
-      (a, b) => (a.value - b.value) * (props.sort === "ascending" ? 1 : -1)
-    );
+    // if only one variable, sort by places
+    if (props.statVarSpec.length == 1) {
+      dataGroups.sort(
+        (a, b) =>
+          (a.value[0].value - b.value[0].value) *
```
pradh, line 290: We're sure there will be one value in both a and b?  if not, add defensive check?

static/js/components/tiles/bar_tile.tsx
```diff
@@ -283,21 +283,30 @@ function rawToChart(
...
+          (props.sort === "ascending" ? 1 : -1)
+      );
+    } else {
+      // sort variables in first group by value
+      dataGroups[0].value.sort(
```
pradh, line 295: is there a dataGroups not empty check earlier?

pradh (approved):
Thanks!

static/js/components/tiles/bar_tile.tsx
juliawu: Good call out! Added defensive check.

static/js/components/tiles/bar_tile.tsx
juliawu: Added defensive check!

PR #3367, opened 2023-09-01
Fix screenshot test failure and support setting a maximum block number in results

shifucun:
* Fixed the issue that highlight tile does not have proper setup for async-element markup
* Use 3000px for nl/explore page height universally.


pradh (approved):
Nice!

PR #3368, opened 2023-09-01
NL: Fix two edge-case bugs

pradh:
1. In comparison query, do not co-locate SVs that aren't part of the same peer-group
2. Avoid deduping SVs incorrectly for correlation vars [immunization vs. debt in the world]


jehangiramjad (approved):

pradh:
Thanks!

PR #3369, opened 2023-09-01
more homepage updates

beets:
- Went through the hp with Ariana today and made requested style updates
- Also updated search query to match results


static/css/homepage.scss
```diff
@@ -15,7 +15,8 @@
  */
 
 /* Styles for the homepage */
-
+@use "sass:math";
```
shifucun, line 18: this is cool!

shifucun (approved):

PR #3372, opened 2023-09-01
Add prod/autopush/dev as remote screenshot target; Try to fix image corruption issue

shifucun:
All main DC site uses the same page.json as /local, Here created symbolic links.

Now there is two saving issue:

1. Some images are saved in the wrong name
2. Some images are corrupted and not fully saved.

These are very likely from the pytest parallel run. Disabling that.

### DIFF
--- a/server/integration_tests/test_data/e2e_answer_places/whatistheemissionsinthesecounties/chart_config.json
+++ b/server/integration_tests/test_data/e2e_answer_places/whatistheemissionsinthesecounties/chart_config.json
@@ -730,6 +730,13 @@
     "childPlaces": {},
     "childTopics": [
       {
+        "dcid": "dc/topic/CO2Emissions",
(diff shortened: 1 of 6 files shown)


pradh (approved):

server/integration_tests/test_data/e2e_answer_places/whatistheemissionsinthesecounties/chart_config.json
```diff
@@ -729,6 +729,13 @@
   "relatedThings": {
     "childPlaces": {},
     "childTopics": [
+      {
+        "dcid": "dc/topic/CO2Emissions",
+        "name": "Carbondioxide ( CO2)",
```
shifucun, line 734: @pradh i have to update this, but it seems not good, should be (CO2)?

PR #3375, opened 2023-09-01
Add a generic way for creating redirect for dc.org/demo/<name>

shifucun:


server/routes/redirects.py
```diff
@@ -101,4 +106,17 @@ def insights():
...
       url_for('explore.page',
               _scheme=current_app.config.get('SCHEME', 'https'),
-              code=302))
\ No newline at end of file
+              code=302))
+
+
+def load_redirects(name):
+  local_file = gcs.download_file(bucket=GLOBAL_CONFIG_BUCKET,
+                                 filename='redirect.json')
```
pradh, line 114: Should we also check-in the file so we have a record of the URLs we have and can restore to GCS, etc.

pradh (approved):
Very nice!

server/routes/redirects.py
```diff
@@ -101,4 +106,17 @@ def insights():
...
+  with open(local_file) as fp:
+    mapping = json.load(fp)
+    return mapping.get(name, '/')
+
+
+@bp.route('/link/<path:name>')
+def demo(name):
```
pradh, line 121: nit: link(name)

pradh (approved):

server/routes/redirects.py
shifucun: Done

server/routes/redirects.py
shifucun: Done

server/routes/redirects.py
```diff
@@ -101,4 +106,17 @@ def insights():
...
+  local_file = gcs.download_file(bucket=GLOBAL_CONFIG_BUCKET,
+                                 filename='redirects.json')
+  with open(local_file) as fp:
+    mapping = json.load(fp)
+    return mapping.get(name, '/')
+
+
+@bp.route('/link/<path:name>')
+def link(name):
+  return redirect(load_redirects(name), code=302)
```
keyurva, line 122: Will this download and load the file from GCS for every link call?

shifucun: Yes, it's doing that to make sure we can dynamically change the content. The performance seems ok.

PR #3377, opened 2023-09-01
Fix up suggested queries

beets:
- dropped green jobs for now (it returns mostly green businesses, farms, etc).
- fixed typos
- updated queries to be more "natural"
- reordered some suggestions so our better pages "rank" higher


static/js/apps/explore_landing/topics.json
```diff
@@ -296,20 +296,20 @@
             "url": "q=How+does+child+immunizations+correlate+with+life+expectancy+in+Asia"
           },
           {
-            "title": "Health conditions vs median age in Alameda County",
-            "url": "q=Health+conditions+vs+median+age+in+Alameda+County"
+            "title": "Compare health conditions vs median age in Alameda County",
```
pradh, line 299: nit: vs.

pradh (approved):
Thanks!

static/js/apps/explore_landing/topics.json
beets: thanks - also updated throughout

PR #3379, opened 2023-09-01
[tiles] Per capita calculation

chejennifer:
- update per capita calculation for tiles so that tiles fetch a series for the denom data and chooses the data point with the closest date to the main stat
    - create new data fetch methods for series and series/within
    - add helper function that gets the denom value to use for a given data point & use this in all the tiles
- when choosing unit and scaling factor, use % and scaling factor of 100 when per capita is selected and there isn't any other unit being used. Make this a helper function that is used in all the tiles

[screen-recording - 2023-09-01T155755.603.webm](https://github.com/datacommonsorg/website/assets/69875368/bf8b3eef-03fa-4752-93e2-5380db84084c)
[screen-recording - 2023-09-01T155636.661.webm](https://github.com/datacommonsorg/website/assets/69875368/b8916664-d15d-4f0f-8c4a-0615b3585963)


pradh (commented):
Thats a lot of code, without tests and at this late stage.  Please review carefully :)

static/js/utils/tile_utils.tsx
```diff
@@ -270,3 +276,88 @@ export function getSourcesJsx(sources: Set<string>): JSX.Element {
...
+  // Otherwise, try to get the unit from the stat data
+  if (!result.unit) {
+    let statMetadata = null;
+    if (statPointData) {
+      const obsWithFacet = Object.values(
+        statPointData.data[svSpec.statVar]
+      ).find((obs) => !!obs.facet);
+      statMetadata = statPointData.facets[obsWithFacet.facet];
+    }
+    if (statSeriesData) {
```
beets, line 305: super nit: else if?

static/js/utils/tile_utils.tsx
```diff
@@ -270,3 +276,88 @@ export function getSourcesJsx(sources: Set<string>): JSX.Element {
...
+}
+
+interface DenomInfo {
+  value: number;
+  date: string;
+  source: string;
+}
+
+/**
+ * Gets information needed to calculate per capita for a single stat data point.
```
beets, line 329: nit: add a note here about date alignment and that null is returned when the matching value is 0

beets (approved):
the code consolidation is nice

static/js/utils/tile_utils.tsx
chejennifer: done

static/js/utils/tile_utils.tsx
chejennifer: done

chejennifer:
thanks for the review!

PR #3380, opened 2023-09-01
Some more updates

chejennifer:
- make sv widget collapsible
- show more digits when highlight value is 0.0

[screen-recording - 2023-09-01T163635.868.webm](https://github.com/datacommonsorg/website/assets/69875368/5a5e7a8c-6515-4ba2-8060-4dd79f46b9bf)

Screenshot 2023-09-01 at 4 37 38 PM


static/js/components/tiles/highlight_tile.tsx
```diff
@@ -201,10 +201,16 @@ const fetchData = (props: HighlightTilePropType): Promise<HighlightData> => {
       if (props.statVarSpec.scaling) {
         value *= props.statVarSpec.scaling;
       }
+      // If value is a decimal, calculate the numFractionDigits is the number of
+      // digits so that at least 2 non-zero digits will show.
+      const numFractionDigits =
+        Math.abs(value) >= 1
+          ? NUM_FRACTION_DIGITS
+          : 1 - Math.floor(Math.log(Math.abs(value)) / Math.log(10));
```
beets, line 209: We might want to put a cap on this so it doesn't get arbitrarily long. Could be a TODO for now.

beets (approved):
nice math!

static/js/components/tiles/highlight_tile.tsx
juliawu: I think it won't get arbitrarily long. I think the logic actually finds the first non-zero digit and the digit after (eg. a value of 0.003004 gives a numFractionDigits of 4, not 6)

If we're just displaying the digits up to first sig fig I think the logic is OK, but we should update the comment to avoid confusion.

chejennifer: done

chejennifer:
thanks for the reviews!

PR #3384, opened 2023-09-03
NL: Fix a few bugs from DDD results

pradh:
This PR fixes a few bugs:

1. When doing place fallback do not report incorrect peer/child places (seen in "what kind of pollution affects seattle in summer")
2. Report "low confidence" by looking at the SV score for the SV in the top chart (rather than the detected SVs), since that's what the user sees (seen in "How does school size of urban schools compare to rural schools")
3. Handle child-place lookup blowup (eg schools in US) more gracefully (seen in "How does school size of urban schools compare to rural schools")
4. When getting SVs for correlation, consider only 2 part candidates (seen in "What is the relationship between housing size and home prices")
5. One recent silly regression crash in `_get_multi_sv_pair` (trigged by "What crimes are considered felonies vs misdemeanors")
6. Minor check-before-access fix with debug-info


shifucun (approved):

pradh:
Thanks Bo!

PR #3389, opened 2023-09-04
Switch to manual page view tracking for explore pages

beets:
Disable automatic page view tracking for explore pages, and instead track pageviews on hash changes (including initial page load tracking).

More info at:
https://developers.google.com/analytics/devguides/collection/ga4/views?client_type=gtag


shifucun (approved):
This is very cool!

beets:
thanks!

PR #3393, opened 2023-09-05
Use a lot more example query for dev screenshot scraping. Fix concurrent test issue due to same tmp file name.

shifucun:


pradh (approved):
Very nice!

PR #3397, opened 2023-09-05
Override descriptions for commute time variable

juliawu:
Add an override for the title and description for the variable `dc/e9gftzl2hm8h9` to make its meaning more clear in explore results.

Before:
Screenshot 2023-09-05 at 10 41 28 AM

After:
Screenshot 2023-09-05 at 10 41 00 AM

### DIFF
--- a/server/lib/nl/common/constants.py
+++ b/server/lib/nl/common/constants.py
@@ -327,6 +327,8 @@ SV_DISPLAY_NAME_OVERRIDE = {
         "Household Median Income",
     "Median_Earnings_Person":
         "Individual Median Earnings",
+    "dc/e9gftzl2hm8h9":
+        "Total Commute Time Across All Workers",
     "dc/6rltk4kf75612":
         "Work at home",
     "dc/vp8cbt6k79t94":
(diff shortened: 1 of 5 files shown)


jehangiramjad (approved):
I think this LGTM to me. Do wait for @pradh though :)

jehangiramjad:
To fix the nl_test failures, please run ./run_test -g from the top-level folder

server/lib/nl/common/constants.py
```diff
@@ -761,6 +763,8 @@
         "Relative to the Average Yearly Min Temperature between 1980-2010, the Predicted Min Temperature difference with 95% chance, at least once in the year, according to a CMIP6 Ensemble model as per SSP 245 (intermediate) scenario.",
     "DiffRelativeToAvg_1980_2010_MinTemp_Daily_Hist_95PctProb_LessThan_Atleast1DayAYear_CMIP6_Ensemble_SSP585":
         "Relative to the Average Yearly Min Temperature between 1980-2010, the Predicted Min Temperature difference with 95% chance, at least once in the year, according to a CMIP6 Ensemble model as per SSP 585 (pessimistic) scenario.",
+    "dc/e9gftzl2hm8h9":
+        "Total time spent on a single commute across all workers",
```
pradh, line 767: in a week?

server/lib/nl/common/constants.py
```diff
@@ -327,6 +327,8 @@
         "Household Median Income",
     "Median_Earnings_Person":
         "Individual Median Earnings",
+    "dc/e9gftzl2hm8h9":
+        "Total Commute Time Across All Workers",
```
pradh, line 331: Given the description that immediately follows it, and the other shorter titles, tempted to say, keep it short?  esp the "Across All Workers" part.

server/lib/nl/common/constants.py
juliawu: It looks like this is for a single commute actually. The original question here:https://www.census.gov/acs/www/about/why-we-ask-each-question/commuting/ suggests a single commute, and dividing the variable value by "Population: Outside of Home" also gives the average commute time cited by Census ACS on their website, which appears to be for a single commute.

server/lib/nl/common/constants.py
juliawu: Sounds good. Shortened to "Total Commute Time" while keeping the description the same.

server/lib/nl/common/constants.py
pradh: Ah, got it.  From [1] "Total time spent on a one-way commute from home to work across all workers."

[1] - https://www2.census.gov/programs-surveys/acs/tech_docs/subject_definitions/2021_ACSSubjectDefinitions.pdf

pradh (approved):

PR #3402, opened 2023-09-05
[explore] drop header section for place overview

chejennifer:
Screenshot 2023-09-05 at 3 54 45 PM
Screenshot 2023-09-05 at 3 55 40 PM


pradh (approved):
Nice!

chejennifer:
thanks for the review!

PR #3411, opened 2023-09-06
Add toggle for homepage animation, with cookies

juliawu:
Adds a light grey toggle to the bottom-right corner of the animation on the homepage, which when clicked, toggles hiding and showing the animation.

A cookie is set on the browser to save whether the animation was hidden, and if cookie is present, the homepage will start with the animation hidden. The cookie lasts for 24hrs.

https://github.com/datacommonsorg/website/assets/4034366/f3462169-859e-4e88-8a7c-05be398aa593


static/css/homepage.scss
```diff
@@ -333,6 +333,15 @@ $search-answer-bottom: 48px;
   }
 }
 
+#search-animation-toggle {
+  color:#bbb;
+  font-size: 18px;
+  margin-bottom: $topics-section-vmargin !important;
+  margin-left: auto;
+  margin-top: -24px;
+  width: fit-content;
```
beets, line 342: Also add cursor: pointer

chejennifer:
very cool, but can we make: `cursor: pointer` or `cursor: default` for the icon so that it looks clickable

chejennifer (approved):

juliawu:
> very cool, but can we make: cursor: pointer or cursor: default for the icon so that it looks clickable

Good idea! Done!

beets (approved):
Thanks for the change!

static/css/homepage.scss
juliawu: Done!

PR #3426, opened 2023-09-06
[SDG] first pass search landing updates

beets:
Most requested changes are in, save for the colored bullets based on goal #
Also added Roboto for headers which seems to be the direction the design is headed. Let me know if you'd prefer it scoped only to the search page.

desktop:
image

at narrow window widths:
image

### DIFF
--- a/experimental/sdg-tracker/src/components/layout/components.css
+++ b/experimental/sdg-tracker/src/components/layout/components.css
@@ -14,12 +14,6 @@
  * limitations under the License.
  */
 
-@import url('https://fonts.googleapis.com/css2?family=Roboto&display=swap');
-
-h1, h2, h3, h4 {
-    font-family: 'Roboto', sans-serif;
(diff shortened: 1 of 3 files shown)


experimental/sdg-tracker/src/components/layout/components.css
```diff
@@ -14,6 +14,12 @@
  * limitations under the License.
  */
 
+@import url('https://fonts.googleapis.com/css2?family=Roboto&display=swap');
```
dwnoble, line 17: Can you actually add this to the `src/index.css` file? UN wants roboto everywhere

dwnoble (approved):
Nice!

experimental/sdg-tracker/src/components/layout/components.css
beets: sure - done.

PR #3428, opened 2023-09-06
Update country/region top header in sdg-tracker

juliawu:
Updates the top header in the sdg-tracker to match mocks.

In particular:
* Adds a "PlaceHeaderCard" component for the new country/region page headers
* Removes the goal icon
* Extracts out the relevant components from CountriesContent.tsx into components.tsx
* Removes references to `PlaceChips`which is no longer used
* Adds back "World" heading in the goal pages
* Don't show breadcrumbs if "All goals" is the only entry

Screenshots:

Screenshot 2023-09-06 at 4 49 00 PM

Screenshot 2023-09-06 at 4 48 45 PM
Screenshot 2023-09-06 at 4 48 05 PM

Screenshot 2023-09-06 at 4 47 48 PM


experimental/sdg-tracker/src/components/layout/components.tsx
```diff
@@ -127,3 +130,148 @@ export const SearchBar: React.FC<{
     </SearchInputContainer>
   );
 };
+
+// Country selection dropdown
+const CountrySelectContainer = styled.div`
```
keyurva, line 135: For custom DC integration, it's easier if the styles are in their own CSS. Any chance we can do that?

juliawu: I've been using the styled divs to follow the pattern set by Dan -- @dwnoble thoughts?

dwnoble: for the header and footer and other reusable components i think that makes sense, but this and many other components in this project are tied to the [ant design](https://ant.design/) library, so they may not make sense to integrate with custom DC anyway

keyurva: SG. Thanks.

dwnoble:
Thanks @juliawu ! two things:

(1) looks like the "World" title text got removed from the Goals page-can you add it back?
(2) the breadcrumbs look a little off to me when only "All goals" is present- if it's the only one can you hide them?

pr-3428-fb

juliawu:
> Thanks @juliawu ! two things:
> 
> (1) looks like the "World" title text got removed from the Goals page-can you add it back? (2) the breadcrumbs look a little off to me when only "All goals" is present- if it's the only one can you hide them?
> 
> pr-3428-fb

Good catch! done and done. Screenshots and PR description have been updated accordingly. I also removed the goal icon per offline discussion.

dwnoble:
Thank you! last thing can you remove the empty header in the search results page
Screenshot 2023-09-06 at 5 16 23 PM

dwnoble (approved):

juliawu:
> Thank you! last thing can you remove the empty header in the search results page Screenshot 2023-09-06 at 5 16 23 PM

Done!

PR #3433, opened 2023-09-07
Update country/region pages in UN sdg tracker site

juliawu:
This PR:
* Updates the country/region pages to use one card per target instead of one card total
* Adds a header for each target
* Adds store logic for the text that accompanies each target's header
* Adds a pink headline blurb tile for indicators we have text for, at the start of each indicator
* Includes some minor CSS updates

More features and some code cleanup will come in followup PRs.

Screenshot:
Screenshot 2023-09-06 at 9 19 34 PM


dwnoble (approved):
Nice!

PR #3434, opened 2023-09-07
SDG: Implement Variable groupings

pradh:
Modify the Topic generation script to incorporate variable-groups from Luis (checked in as variable_groups.csv).

Minor tweaks to backend code to always prefer timelines, and in place of bar-chart to show highlights (since they don't like bars...).

Also, kill minisdg generation.

image


jehangiramjad (approved):

dwnoble (approved):
Nice!

juliawu (approved):
Thanks for this!

PR #3441, opened 2023-09-07
Add lint tests for sdg-tracker code

juliawu:
Makes running `./run_test.sh -l` and `./run_test.sh -f` also lint and fix the code in `experimental/sdg-tracker`

We can remove these changes once the sdg-tracker code is ported to the UN.

### DIFF
--- a/experimental/sdg-tracker/package.json
+++ b/experimental/sdg-tracker/package.json
@@ -27,19 +27,14 @@
     "styled-components": "^6.0.6"
   },
   "devDependencies": {
-    "@types/react-dom": "^17.0.2",
     "@types/react": "^17.0.2",
+    "@types/react-dom": "^17.0.2",
     "@typescript-eslint/eslint-plugin": "^6.0.0",
     "@typescript-eslint/parser": "^6.0.0",
     "@vitejs/plugin-react-swc": "^3.3.2",
-    "eslint-config-prettier": "^6.11.0",
-    "eslint-plugin-prettier": "^3.1.4",
+    "eslint": "^8.45.0",
     "eslint-plugin-react-hooks": "^4.6.0",
     "eslint-plugin-react-refresh": "^0.4.3",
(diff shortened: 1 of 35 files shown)


experimental/sdg-tracker/src/App.tsx
```diff
@@ -45,7 +45,7 @@ const InitializeStore: React.FC = () => {
   const initializeAppState = useStoreActions((a) => a.initializeAppState);
   useEffect(() => {
     initializeAppState();
-  }, []);
+  }, [initializeAppState]);
```
shifucun, line 48: should really test out these dependency... I have noticed that these are indeed issues when we initially develop, but they may introduce breakage, and need property bigger fix.

juliawu: Ok, reverted updating the dependency arrays for now.

experimental/sdg-tracker/package.json
```diff
@@ -26,14 +27,19 @@
...
   },
   "devDependencies": {
-    "@types/react": "^17.0.2",
     "@types/react-dom": "^17.0.2",
+    "@types/react": "^17.0.2",
     "@typescript-eslint/eslint-plugin": "^6.0.0",
     "@typescript-eslint/parser": "^6.0.0",
     "@vitejs/plugin-react-swc": "^3.3.2",
-    "eslint": "^8.45.0",
+    "eslint-config-prettier": "^6.11.0",
```
dwnoble, line 35: Can you remove the newly added dependencies here that arent actually used by the linter for now? It might be all of the new dependencies, but at least prettier &  import-sort

juliawu: Removed all newly added dev dependencies.

juliawu:
Closing this since we've switched folders and UN is moving out of our repo. We can reopen if that changes.

PR #3444, opened 2023-09-07
[sdg] Use text in [ ] for line chart legend labels 

chejennifer:
- do this by adding a new regex property to the line web component that allows processing of stat var names using regex.

Screenshot 2023-09-07 at 4 14 55 PM

### DIFF
--- a/static/library/line_chart_component.ts
+++ b/static/library/line_chart_component.ts
@@ -86,6 +86,12 @@ export class DatacommonsLineComponent extends LitElement {
   // Statistical variable DCIDs
   @property({ type: Array<string>, converter: convertArrayAttribute })
   variables!: Array<string>;
+  
+  // Optional: Regex used to process variable names
+  // If provided, will only use the first case of the variable name that matches
+  // this regex.
+  @property()
+  variableNameRegex!: string;
 
   render(): HTMLElement {
     const lineTileProps: LineTilePropType = {
@@ -110,6 +116,7 @@ export class DatacommonsLineComponent extends LitElement {
       })),
       svgChartHeight: 200,
       title: this.header || this.title,
+      statVarNameRegex: this.variableNameRegex
     };
     return createWebComponentElement(LineTile, lineTileProps);
   }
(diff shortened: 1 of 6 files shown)


static/library/line_chart_component.ts
```diff
@@ -110,6 +116,7 @@ export class DatacommonsLineComponent extends LitElement {
       })),
       svgChartHeight: 200,
       title: this.header || this.title,
+      statVarNameRegex: this.variableNameRegex
```
pradh, line 119: super nit: use consistent name for it

pradh (commented):
Thanks for the change!

static/library/line_chart_component.ts
chejennifer: right now users specify "variables" in the web component which in turn maps to "statVarSpecs" used by the line tile. I tried to keep the name consistent with how each component references variable/statVar .. so not sure if we should consistently use statVarNameRegex or variableNameRegex here

pradh: i see, ok, that makes sense!

dwnoble: I like `variableNameRegex` for the web component attribute to be consistent with the `variable` web component attribute

static/library/line_chart_component.ts
```diff
@@ -86,6 +86,12 @@ export class DatacommonsLineComponent extends LitElement {
   // Statistical variable DCIDs
   @property({ type: Array<string>, converter: convertArrayAttribute })
   variables!: Array<string>;
+  
+  // Optional: Regex used to process variable names
+  // If provided, will only use the first case of the variable name that matches
+  // this regex.
```
dwnoble, line 92: can you add an example regex and the before/after of what happens to the variable to this comment

dwnoble (commented):
Nice!

Can you add an example here: https://github.com/datacommonsorg/website/blob/d30b7e80db881708b0bfd45e7ac9362d5ffeeb4a/packages/web-components/examples/all-charts.html

and add a description & example usage of this attribute here: https://github.com/datacommonsorg/website/blob/d30b7e80db881708b0bfd45e7ac9362d5ffeeb4a/packages/web-components/docs/components/line.md

dwnoble:
also can you update the title in `line.md` above to `Data Commons Line Chart Web Component` while you are at it :)

static/library/line_chart_component.ts
chejennifer: done

juliawu (approved):
Very nice!

PR #3446, opened 2023-09-08
Explore: Fix place title issue

pradh:
When there are answer places flowing, like [which of these countries have seen the greatest reduction in poverty?], the places going into the query may not be the same that are shown in the result.  Currently, we list the places going in, which is important to do for certain types of queries (e.g., [most poor states in the US] has states in result but the main place is US), but when there are answer-places show those instead.

Also, add the demo queries in NL integration tests.

**Before:**
image

**After:**
image


jehangiramjad (approved):
Nice!

PR #3447, opened 2023-09-08
NL: Improve bad-words matching

pradh:
1.  Support phrases
2. Support multi-word matching, where every word has to appear in the query (in any order), with `:` separation.
3. Add tests and prod file validation

Note: Using regex as input might be very flexible but since it is easy to get wrong or eye-ball, I chose this simpler config of just `:` separated values.

### DIFF
--- a/server/integration_tests/test_data/e2e_electrification_demo/whichcountriesinafricahavehadthegreatestincreaseinelectricityaccess/chart_config.json
+++ b/server/integration_tests/test_data/e2e_electrification_demo/whichcountriesinafricahavehadthegreatestincreaseinelectricityaccess/chart_config.json
@@ -503,13 +503,6 @@
           ]
         },
         {
-          "dcid": "country/IOT",
-          "name": "British Indian Ocean Territory",
-          "types": [
-            "Country"
-          ]
-        },
-        {
           "dcid": "country/KEN",
           "name": "Kenya",
           "types": [
@@ -673,6 +666,13 @@
         {
           "dcid": "country/STP",
           "name": "S\u00e3o Tom\u00e9 and Pr\u00edncipe",
+          "types": [
+            "Country"
+          ]
+        },
+        {
+          "dcid": "country/SWZ",
+          "name": "Eswatini",
           "types": [
             "Country"
           ]
--- a/server/lib/nl/common/bad_words.py
+++ b/server/lib/nl/common/bad_words.py
@@ -40,32 +40,36 @@ class Entry:
   # This is a multi-word entry, where we must match every
   # word (in any order) in the query.  When sorted, the first
   # word is the key of Entry, and the remaining words are here.
+  # NOTE: The first word only appears in the key, not here.
   #
-  # In case of "idiot:cat:bat", only if idiot, bat and cat
-  # appear in the query (regardless of ordering),
-  # will we match.  We will have: ["cat", "idiot"]
-  other_words: List[str] = field(default_factory=list)
+  # In case of lines "idiot:cat:bat", "mat:bat" only if
+  # all idiot, bat and cat appear in the query, or bat and mat
+  # both appear in the query will we match (regardless of ordering),
+  # In that case, the key of the Entry will be "bat", and
+  # in other_words, we'll have:
+  #   [ ["cat", "idiot"], ["mat"] ]
+  other_words: List[List[str]] = field(default_factory=list)
 
 
 @dataclass
-class BadWords:
+class BannedWords:
   # The key is a word.  It is either a singleton word,
   # or the first word in a phrase, or the first sorted
   # word in a multi-word (aka colon-delimited) entry.
-  words: Dict[str, Entry]
+  entries: Dict[str, Entry]
 
 
 #
 # Loads a list of bad words from a text file.
 #
-def load_bad_words() -> BadWords:
+def load_bad_words() -> BannedWords:
   local_file = gcs.download_file(bucket=GLOBAL_CONFIG_BUCKET,
                                  filename=BAD_WORDS_FILE)
   return load_bad_words_file(local_file)
 
 
-def load_bad_words_file(local_file: str, validate: bool = False) -> BadWords:
-  bad_words = BadWords(words={})
+def load_bad_words_file(local_file: str, validate: bool = False) -> BannedWords:
+  bad_words = BannedWords(entries={})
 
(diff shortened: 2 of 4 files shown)


server/tests/lib/nl/bad_words_test.py
```diff
@@ -0,0 +1,53 @@
...
+    # Non-existent words
+    self.assertTrue(bad_words.is_safe('number of people', bw))
+
+    # Single-word: "stupid"
+    self.assertFalse(bad_words.is_safe('is it that stupid', bw))
+    self.assertFalse(bad_words.is_safe('who said stupid is as stupid does', bw))
+
+    # Phrase: "very bad word"
+    self.assertFalse(bad_words.is_safe('is that a very bad word really', bw))
+    self.assertFalse(
```
jehangiramjad, line 35: nit: this should have a comment about what it's testing (currently seems like a continuation of the above which is confusing)

server/lib/nl/common/bad_words.py
```diff
@@ -12,40 +12,138 @@
...
+  #
+  phrases: List[str] = field(default_factory=list)
+
+  # This is a multi-word entry, where we must match every
+  # word (in any order) in the query.  When sorted, the first
+  # word is the key of Entry, and the remaining words are here.
+  #
+  # In case of "idiot:cat:bat", only if idiot, bat and cat
+  # appear in the query (regardless of ordering),
+  # will we match.  We will have: ["cat", "idiot"]
```
jehangiramjad, line 46: also "bat" in the list?

server/lib/nl/common/bad_words.py
```diff
@@ -12,40 +12,138 @@
...
+  other_words: List[str] = field(default_factory=list)
+
+
+@dataclass
+class BadWords:
```
jehangiramjad, line 51: as a nit, since this is a new class, perhaps a good time to call this "BannedWords"? That's essentially what they are and feels more "standard"?

server/lib/nl/common/bad_words.py
```diff
@@ -12,40 +12,138 @@
...
+      line = ' '.join(line.split())
       # Ignore comments
       if line.startswith('#') or line.startswith('//'):
         continue
 
-      bad_words.add(line.lower())
+      # Generally, be resilient to duplicates.  Only assert when `validate` is set.
+      if _DELIM in line:
+        words = sorted([w.strip() for w in line.split(_DELIM) if w.strip()])
+        if words[0] not in bad_words.words:
```
jehangiramjad, line 82: what is you have a situation where the same first word is part of multiple DELIM sequences?

server/lib/nl/common/bad_words.py
```diff
@@ -12,40 +12,138 @@
...
+          bad_words.words[words[0]] = Entry(is_singleton=False)
+        bad_words.words[words[0]].other_words = words[1:]
+
+        if validate:
```
jehangiramjad, line 86: suggestion: factor out the validation routines in a separate function (or three different functions)

server/lib/nl/common/bad_words.py
```diff
@@ -12,40 +12,138 @@
...
+          assert ' ' not in line, f'Line {line} has a ":" and a phrase!'
+          for w in words:
+            if w in bad_words.words:
+              assert not bad_words.words[
+                  w].is_singleton, f'{line} is redundant because {w} already exists!'
+
+      elif ' ' in line:
+        # This is the case of a phrase.
+        parts = line.split(' ', 1)
+        if parts[0] not in bad_words.words:
```
jehangiramjad, line 96: same question as above: what if parts[0] appears in multiple phrases?

jehangiramjad (commented):
Thanks! Some clarifying questions

server/tests/lib/nl/bad_words_test.py
pradh: Added!

server/lib/nl/common/bad_words.py
pradh: Clarified that bat is only in the key

server/lib/nl/common/bad_words.py
pradh: Good idea.  Changed class name, but keeping function names the same to match file name.

server/lib/nl/common/bad_words.py
pradh: Great catch!  Added support and a test.

server/lib/nl/common/bad_words.py
pradh: Good call... see now?

server/lib/nl/common/bad_words.py
pradh: This is currently handled because the `phrases` is a repeated list of all phrases with parts[0] as the first word.

pradh (commented):
Very helpful review, thanks Jehangir!

jehangiramjad (approved):
Thanks!

server/integration_tests/test_data/e2e_electrification_demo/whichcountriesinafricahavehadthegreatestincreaseinelectricityaccess/chart_config.json
```diff
@@ -502,13 +502,6 @@
             "Country"
           ]
         },
-        {
```
jehangiramjad, line 505: any idea why this got removed? mixer?

server/tests/lib/nl/bad_words_test.py
jehangiramjad: Thanks!

server/lib/nl/common/bad_words.py
jehangiramjad: Thanks, got it!

server/lib/nl/common/bad_words.py
jehangiramjad: Thanks!

server/lib/nl/common/bad_words.py
jehangiramjad: Great, thanks!

server/lib/nl/common/bad_words.py
jehangiramjad: Got it, yes!

jehangiramjad (commented):
Thanks! LGTM :)

PR #3448, opened 2023-09-08
Link to a different form for "low confidence" message

pradh:
This is as a disclaimer to flag inappropriate content.  We might need a different form than the one used, will talk to Prem tomm.

image


chejennifer (approved):
nice

PR #3449, opened 2023-09-08
[sdg] Migrated search to use detect and fulfill endpoint

dwnoble:
Will address in a follow-up

* Supporting multiple country responses
* Adding search context


experimental/sdg-tracker/src/components/search/Search.tsx
```diff
@@ -245,22 +245,24 @@ const Search = () => {
...
       setErrorMessage("");
-      const responseObj = await dataCommonsClient.detect(searchQuery || query);
+      const fulfillResponse = await dataCommonsClient.detectAndFulfill(
+        searchQuery || query
+      );
       setIsSearching(false);
 
-      if (responseObj.failure) {
-        setErrorMessage(responseObj.failure);
+      if (fulfillResponse.failure || fulfillResponse.userMessage) {
```
pradh, line 253: Jennifer should confirm, but I think this is how error is detected: https://github.com/datacommonsorg/website/blob/989710ebc78b643c0abee61333eba875702ee11a/static/js/apps/explore/app.tsx#L159-L167

userMessage can be valid even for a non-error case.

experimental/sdg-tracker/src/components/countries/CountriesContent.tsx
```diff
@@ -150,6 +152,12 @@ const CountriesContent: React.FC<{
    * Fetch page content
    */
   useEffect(() => {
+    // If a fulfill response was passed in, use that
+    if (fulfillResponse) {
+      setLocalFulfillResponse(fulfillResponse);
+      return;
+    }
+    // Otherwise fetch a fulfill response based on the
```
chejennifer, line 160: nit: finish the sentence

experimental/sdg-tracker/src/components/search/Search.tsx
chejennifer, line 253: userMessage and failure should be handled differently? Because user message can be for fallback in which case I think we would still want to show charts? e.g., user message can be something like "we couldn't detect a place, so showing charts for world"

chejennifer (commented):
nice!

experimental/sdg-tracker/src/components/search/Search.tsx
chejennifer: yes, I also left a comment about this that failure and userMessage should be handled separately

dwnoble: Separated the handling logic for userMessage and failure

experimental/sdg-tracker/src/components/countries/CountriesContent.tsx
dwnoble: done

experimental/sdg-tracker/src/components/search/Search.tsx
dwnoble: done

dwnoble:
Example of displaying a userMessage when there is no failure:
Screenshot 2023-09-08 at 2 00 39 PM

Comparing two countries
Screenshot 2023-09-08 at 2 01 29 PM

experimental/sdg-tracker/src/components/search/Search.tsx
```diff
@@ -223,6 +224,8 @@ const Search = () => {
   const [variableDcids, setVariableDcids] = useState<string[]>([]);
   const [placeDcids, setPlaceDcids] = useState<string[]>([]);
   const [errorMessage, setErrorMessage] = useState("");
+  const [userMessage, setUserMessage] = useState("");
```
chejennifer, line 227: nit: do we need the userMessage as a state? because we can just read the userMessage from the fulfillResponse?

experimental/sdg-tracker/src/components/countries/CountriesContent.tsx
```diff
@@ -109,38 +107,49 @@ const Spinner: React.FC<{ fontSize?: string }> = ({ fontSize }) => {
...
   fulfillResponse?: FulfillResponse;
   hidePlaceSearch?: boolean;
+  isFetchingFulfillment?: boolean;
   onSearch?: (query: string) => void;
-  placeDcid?: string;
+  placeDcids: string[];
   query?: string;
   setPlaceDcid: (placeDcid: string) => void;
   showNLSearch?: boolean;
+  userMessage?: string;
```
chejennifer, line 119: curious why not just read userMessage from fulfillResponse?

juliawu (approved):
Thanks for responding to everyone's comments, this now lgtm.

experimental/sdg-tracker/src/components/countries/CountriesContent.tsx
```diff
@@ -573,7 +591,7 @@ const ChartTile: React.FC<{ placeDcid: string; tile: ChartConfigTile }> = ({
           apiRoot={WEB_API_ENDPOINT}
           header={tile.title}
           variable={tile.statVarKey.join(" ")}
-          place={placeDcid}
+          place={placeDcids.join(" ")}
```
chejennifer, line 594: I believe gauge web component takes a single place dcid?

chejennifer (commented):
thanks for the updates!

experimental/sdg-tracker/src/components/countries/CountriesContent.tsx
juliawu: Good catch! @dwnoble please fix before merge.

dwnoble: Fixed

chejennifer (approved):
thanks!

PR #3454, opened 2023-09-08
[sdg] add un regions

chejennifer:
- updates homepage to use same place dropdown as countries page
- update the dropdown to include regions
- get regions on app initialization

Screenshot 2023-09-08 at 1 46 59 PM

Screenshot 2023-09-08 at 1 45 53 PM

### DIFF
--- a/experimental/sdg-tracker/src/state/index.ts
+++ b/experimental/sdg-tracker/src/state/index.ts
@@ -48,6 +48,7 @@ const MenuImageIcon = styled.img`
   border-radius: 0.25rem;
 `;
 
+const REGION_PLACE_TYPES = ["UNGeoRegion", "GeoRegion", "ContinentalUnion"];
 export interface Place {
   name: string;
   dcid: string;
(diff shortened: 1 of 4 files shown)


experimental/sdg-tracker/src/state/index.ts
```diff
@@ -48,6 +48,7 @@ const MenuImageIcon = styled.img`
   border-radius: 0.25rem;
 `;
 
+const REGION_PLACE_TYPES = ["UNGeoRegion", "GeoRegion", "ContinentalUnion"];
```
juliawu, line 51: To confirm, do we want `Continent` as part of this list? Not sure if the UNGeoRegions/GeoRegions include whole continents or if we have data for them.

juliawu (approved):
Nice!

experimental/sdg-tracker/src/state/index.ts
chejennifer: @n-h-diaz can you confirm these are the 3 regions we need?

n-h-diaz: yes we do have data for 'Continent' too, but I was thinking of this as separate from regions (so yes those are the only three region types). not sure if we also want Continent, but maybe that's a question for UN

chejennifer:
thanks for the reviews!

PR #3456, opened 2023-09-08
[sdg] update un regions 

chejennifer:
- update to only show UNGeoRegion, Continent, and ContinentalUnion place types
- update to only show regions that have the name property


pradh (approved):
Thanks!

PR #3462, opened 2023-09-09
[tiles] fix unit overrides

chejennifer:
- use unit overrides for all tile types

(unit here used to be "thousands")
Screenshot 2023-09-08 at 11 03 00 PM


juliawu (approved):
Nice!

PR #3464, opened 2023-09-09
[UN SDG tracker] Add asterisk to maps and show disclaimer at bottom

juliawu:
Adds a disclaimer to the page about map borders, and an asterisk at the end of map titles

Example of a page scrolled all the way down:
See asterisk in map chart title and disclaimer at the bottom.

Screenshot 2023-09-09 at 10 04 10 AM


pradh (approved):
Nice!

PR #3466, opened 2023-09-09
Enable explore link in UN SDG tracker app

shifucun:


experimental/sdg-tracker/src/components/countries/CountriesContent.tsx
```diff
@@ -512,10 +512,7 @@ const ChartGoalBlock: React.FC<{
   return (
     <>
       {placeDcids[0] === EARTH_PLACE_DCID && (
-        <GoalOverview
-          goalNumber={Number(goal)}
-          showExploreLink={false}
-        />
+        <GoalOverview goalNumber={Number(goal)} showExploreLink={false} />
```
pradh, line 515: meant to be false?

pradh (approved):

jehangiramjad (approved):

experimental/sdg-tracker/src/components/countries/CountriesContent.tsx
shifucun: I think this is a different explore link, not related to the chart tile. I will leave it untouched.

pradh: sg

juliawu: False is correct. it should only be true if we're on the "all goals for the world" page.

PR #3468, opened 2023-09-09
Redirect unsdg home page; Use a faster url for map webdriver test

shifucun:


pradh (approved):

server/routes/static.py
```diff
@@ -31,6 +32,8 @@
 @bp.route('/')
 def homepage():
   # Return old homepage if hiding revamp changes
+  if request.url == 'https://unsdg.datacommons.org/':
```
pradh, line 35: Do we want to include "http://" as well as without trailing "/" ?

shifucun: from local try, looks like they are in it..

pradh: sg

PR #3477, opened 2023-09-10
[sdg] add variable name regex for bar tile

chejennifer:
Screenshot 2023-09-09 at 6 38 50 PM

### DIFF
--- a/experimental/sdg-tracker/src/components/countries/CountriesContent.tsx
+++ b/experimental/sdg-tracker/src/components/countries/CountriesContent.tsx
@@ -58,6 +58,7 @@ import { theme } from "../../utils/theme";
 // Approximate chart heights for lazy-loading
 const CHART_HEIGHT = 389;
 const HIGHLIGHT_CHART_HEIGHT = 155;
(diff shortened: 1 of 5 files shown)


experimental/sdg-tracker/src/components/countries/CountriesContent.tsx
```diff
@@ -658,6 +658,7 @@ const ChartTile: React.FC<{
           places={placeDcids.join(" ")}
           sort="descending"
           showExploreMore={true}
+          variableNameRegex={"(?<=\\[)(.*?)(?=\\])"}
```
juliawu, line 661: Let's extract this out to a constant since it's shared with the line chart?

juliawu (approved):
Thanks!

experimental/sdg-tracker/src/components/countries/CountriesContent.tsx
chejennifer: done

chejennifer:
thanks for the review

PR #3479, opened 2023-09-10
[UN SDG] Bug fix: show target cards in order

juliawu:
Sorts the goals/targets/indicators so that the goals, cards, and tiles show up in lexical order by indicator number.

Before:
Screenshot 2023-09-09 at 7 42 13 PM

After:
Screenshot 2023-09-09 at 7 41 52 PM


chejennifer (commented):
nice!

chejennifer (approved):

PR #3480, opened 2023-09-10
[UN SDG] Fix: Remove "[c]" from text for target 14.6

juliawu:
Removes extra "[c]" from the label for target 14.6, which gets used in the sidebar and card headers.


chejennifer (approved):
thanks

PR #3481, opened 2023-09-10
[UN SDG] Ignore ranking charts in fulfillment response

juliawu:
This PR Ignore ranking charts in fulfilment responses, to avoid showing "unknown chart type RANKING ..." errors to the user. Decision is not to have ranking tiles for now. We can add them in properly in a followup PR if wanted.

Before:
Screenshot 2023-09-09 at 8 39 50 PM

After
Screenshot 2023-09-09 at 8 40 05 PM


chejennifer (approved):
thanks

PR #3482, opened 2023-09-10
[UN SDG] Enable multi-place line charts

juliawu:
This PR:
* Updates the interface of ChartConfigTile to match new return type of detect-and-fulfill endpoint
* Updates logic in line tile rendering to support multiple places
* Allows for placeDcidOverride to override the places passed into the line chart. 

Screenshot 2023-09-09 at 9 01 05 PM


chejennifer (approved):
nice

PR #3483, opened 2023-09-10
Fix "explore more" hash for bar tiles

juliawu:
Updates the link for "Explore in Timeline Tool" so the resulting page doesn't have extra place chips.

### On UN SDG Pages, removes extra blank chip
Before:
Screenshot 2023-09-09 at 9 21 48 PM

After:
Screenshot 2023-09-09 at 9 21 35 PM

### On explore pages, removes duplicate chip
Before:
Screenshot 2023-09-09 at 9 28 33 PM

After:
Screenshot 2023-09-09 at 9 29 17 PM


chejennifer (approved):
nice

PR #3485, opened 2023-09-10
Fixed country selector dropdown arrow so clicking it shows the dropdown

dwnoble:
Before: this arrow didnt trigger the dropdown

country_select


chejennifer (approved):
nice

PR #3491, opened 2023-09-10
Some minor fix for unsdg tracker

shifucun:


static/js/utils/tile_utils.tsx
```diff
@@ -407,7 +407,9 @@ export function getStatFormat(
     const seriesWithFacet = Object.values(
       statSeriesData.data[svSpec.statVar]
     ).find((series) => !!series.facet);
-    statMetadata = statSeriesData.facets[seriesWithFacet.facet];
+    if (seriesWithFacet) {
```
chejennifer, line 410: can you also add add this same check for obsWithFacet in the statPointData handling

chejennifer (approved):
thanks!

static/js/utils/tile_utils.tsx
shifucun: Done

PR #3492, opened 2023-09-10
Fix maps of Europe being cut-off

juliawu:
Updates the projection scaling for maps of Europe so that the borders are no longer cut off when the chart is much wider than its height.

 Before:
Screenshot 2023-09-10 at 2 15 48 PM

After:
Screenshot 2023-09-10 at 2 15 21 PM

What Map tool now looks like:
Screenshot 2023-09-10 at 2 18 01 PM

What explore tiles now look like:
Screenshot 2023-09-10 at 2 18 33 PM


chejennifer (approved):
nice

PR #3493, opened 2023-09-10
Add date to tooltips in map tile

juliawu:
Adds the date to tooltips in the map tile

Screenshot:
Screenshot 2023-09-10 at 3 16 54 PM

### DIFF
--- a/static/js/components/tiles/map_tile.tsx
+++ b/static/js/components/tiles/map_tile.tsx
@@ -113,7 +113,6 @@ interface RawData {
 }
 
 export interface MapChartData {
-  dataDates: { [dcid: string]: string };
   dataValues: { [dcid: string]: number };
   metadata: { [dcid: string]: DataPointMetadata };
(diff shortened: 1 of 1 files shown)


static/js/components/tiles/map_tile.tsx
```diff
@@ -339,6 +340,7 @@ function rawToChart(
   const metadataMap = rawData.placeStat.facets || {};
   const placeStat = rawData.placeStat.data[statVarSpec.statVar] || {};
 
+  const dataDates = {};
```
chejennifer, line 343: the date for each value is already in the metadata (line 345), no need to add this

juliawu: Ooh thanks for the heads up! Swapped to using `metadata` instead.

chejennifer (approved):
thanks for the update

dwnoble (approved):

PR #3497, opened 2023-09-11
Add flask endpoint to fetch observation facets

shifucun:


server/lib/fetch.py
```diff
@@ -73,18 +73,24 @@ def _get_processed_facets(facets):
 
 
 def _convert_v2_obs_point(facet):
-  return {
+  result = {
       'facet': facet['facetId'],
-      'date': facet['observations'][0]['date'],
-      'value': facet['observations'][0]['value']
   }
+  if 'observations' in facet:
```
chejennifer, line 79: not sure if facet['observations'] can be empty .. if it can be, maybe also check that facet['observations'] is not empty before accessing the first element?

chejennifer (approved):
very nice, thank you!

server/lib/fetch.py
shifucun: Done

PR #3500, opened 2023-09-11
update points and series flask endpoints to filter by facet id

chejennifer:


server/routes/shared_api/observation/point.py
```diff
@@ -69,8 +69,10 @@ def point_within():
   if not variables:
     return 'error: must provide a `variables` field', 400
   date = request.args.get('date') or 'LATEST'
+  facet_id = request.args.get('facetId') or None
```
shifucun, line 72: shall we use "facetIds" here as well for consistency?

server/routes/shared_api/observation/series.py
```diff
@@ -60,7 +60,8 @@ def series_all():
     return 'error: must provide a `entities` field', 400
   if not variables:
     return 'error: must provide a `variables` field', 400
-  return fetch.series_core(entities, variables, True)
+  facet_ids = list(filter(lambda x: x != "", request.args.getlist('facetIds')))
+  return fetch.series_core(entities, variables, True, facet_ids)
```
shifucun, line 64: also add after line 49?

server/routes/shared_api/observation/point.py
chejennifer: since this is only a single facetId, I think facetId makes more sense? wdyt?

shifucun: Can you remind me what does a list of facet id filter implies? Say i want to fetch [facet1, facet2], but some data has either facet1 or facet2, would that work? 

The result here does return 1 facet, but can the filter be  multiple so I can ask for either 
Census or WorldBank data, whichever has data.

shifucun (approved):
thanks for update!

chejennifer:
thanks for the review!

PR #3503, opened 2023-09-11
[GA] Add GA4 timing event around detec/fulfill endpoints

beets:
This is still somewhat experimental, since GA4 removed native support for page timing events:
https://support.google.com/analytics/answer/9964640#settings-with-no-GA4-equivalent&zippy=%2Cin-this-article

@pradh Adding minimal information first that we can continue to augment if this works, though please let me know if any of the other parameters would be useful to track.

### DIFF
--- a/static/js/apps/explore/app.tsx
+++ b/static/js/apps/explore/app.tsx
@@ -30,7 +30,7 @@ import {
   URL_DELIM,
   URL_HASH_PARAMS,
 } from "../../constants/app/explore_constants";
-import { GA_EVENT_PAGE_VIEW, triggerGAEvent } from "../../shared/ga_events";
+import { GA_EVENT_NL_DETECT_FULFILL, GA_EVENT_NL_FULFILL, GA_EVENT_PAGE_VIEW, triggerGAEvent } from "../../shared/ga_events";
 import {
   QueryResult,
   UserMessageInfo,
@@ -319,6 +319,7 @@ const fetchFulfillData = async (
   disableExploreMore: string
 ) => {
   try {
+    let startTime = window.performance ? window.performance.now() : undefined;
     const resp = await axios.post(`/api/explore/fulfill`, {
       dc,
       entities: places,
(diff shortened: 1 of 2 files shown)


static/js/apps/explore/app.tsx
```diff
@@ -330,6 +336,17 @@ const fetchFulfillData = async (
       classifications: classificationsJson,
       disableExploreMore,
     });
+    if (startTime) {
+      const endTime = window.performance
```
chejennifer, line 340: nit: this wouldn't be end time right? this is the time taken?

static/js/shared/ga_events.ts
```diff
@@ -26,6 +26,7 @@ export function triggerGAEvent(
   if (window.gtag) {
     window.gtag("event", eventName, parameter);
   }
+  console.log(eventName, parameter);
```
chejennifer, line 29: nit: I don't think we need this?

static/js/shared/ga_events.ts
```diff
@@ -97,6 +98,22 @@ export const GA_EVENT_TOOL_CHART_OPTION_CLICK = "tool_chart_option_click";
...
+ */
+export const GA_EVENT_NL_DETECT_FULFILL = "explore_detect_fulfill";
+
+/**
+ * Triggered when detection results are returned in NL search.
+ * Parameters:
+ *   "topic": array of topics
+ *   "time_ms": e2e request/response timing in ms.
+ */
+export const GA_EVENT_NL_FULFILL = "explore_fulfill";
```
chejennifer, line 115: fulfill is based on topic + place, should we also include places as a parameter?

chejennifer (approved):
nice! Just some nits/suggestions

static/js/shared/ga_events.ts
beets: thanks for catching that!

static/js/apps/explore/app.tsx
beets: yes, elapsed time - updated.

pradh (approved):
Query is good.

Past queries in the context would also be useful since that can impact the latency of a query.  But that can come after.

beets:
thanks for the reviews! i've added a todo to add past queries.

PR #3504, opened 2023-09-11
Add a spinner loading prompt after 3 seconds

shifucun:
image

### DIFF
--- a/static/css/base.scss
+++ b/static/css/base.scss
@@ -115,7 +115,7 @@ $headings-font-family: "Google Sans", "Public Sans", $font-family-sans-serif;
 #main-nav .navbar-brand .sep {
   color: var(--gm-3-ref-neutral-neutral-40);
   display: inline-block;
-  margin: 0 .5rem;
+  margin: 0 0.5rem;
 }
 
 .main-content {
@@ -261,7 +261,7 @@ $headings-font-family: "Google Sans", "Public Sans", $font-family-sans-serif;
     justify-content: space-between;
   }
   height: fit-content;
-  
+
   @media (max-width: $sub-footer-breakpoint) {
     .container {
       align-items: center;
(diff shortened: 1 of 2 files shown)


static/css/base.scss
```diff
@@ -356,6 +356,28 @@ $headings-font-family: "Google Sans", "Public Sans", $font-family-sans-serif;
...
+  align-items: center;
+}
+
+#spinner-text::after {
+  content: "Still loading ...";
+  z-index: 10;
+  margin-top: 150px;
+  margin-left: 20px;
+  font-size: 20px;
+  color: white;
```
beets, line 376: is white text visible enough against the loading background? could you add a screenshot

static/css/base.scss
```diff
@@ -356,6 +356,28 @@ $headings-font-family: "Google Sans", "Public Sans", $font-family-sans-serif;
...
+  opacity: 0; /* Initially hidden */
+  animation: fadeIn 0.5s forwards 3s; /* Animation after 3s delay */
```
beets, line 378: fancy :)

static/css/base.scss
```diff
@@ -356,6 +356,28 @@ $headings-font-family: "Google Sans", "Public Sans", $font-family-sans-serif;
...
+  display: flex;
+  justify-content: center;
+  align-items: center;
+}
+
+#spinner-text::after {
+  content: "Still loading ...";
+  z-index: 10;
+  margin-top: 150px;
+  margin-left: 20px;
```
beets, line 374: is the margin necessary if it's already centered?

beets (approved):
this is a very cool, css only solution :)

static/css/base.scss
shifucun: updated screenshot, can you check?

static/css/base.scss
shifucun: done

static/css/base.scss
shifucun: from genAI :)

chejennifer (approved):
nice!

static/css/base.scss
beets: i think it's better to remove since it does look off center now..

beets:
thanks for the screenshot - looks good!

static/css/base.scss
shifucun: I tried that but since there is "...", the text part seems off center. The 20px is used to offset that.

PR #3509, opened 2023-09-11
Homepage speed updates

beets:
Address low hanging fruit from https://pagespeed.web.dev/analysis/https-dev-datacommons-org/gklwbb79ky?form_factor=mobile

- Clean up unused fonts
- Combine fonts to a single call (material icons stays separate so we can block display until font is loaded)
- Add aria labels
- Add image sizes


shifucun (approved):
nice find and this speed tool is really cool! How much improvement do we get from this?

beets:
thanks!

we'll need to measure from dev - it's mostly reducing the "red marks" we get on that report and issues that might cause bad press..

PR #3510, opened 2023-09-11
[UN SDG] Fix missing charts in search results

juliawu:
Some charts were not showing up in search results because of filtering by `mainTopic`.

This PR changes the search results pages:
* results are all in one card with all tiles displayed
* tiles are no longer ordered by goal/target/indicator
* target headings have been taken out
* Adds topic name to the header behind the place name using mainTopics (this approach matches what's currently used on our explore pages)

Screenshot 2023-09-11 at 1 36 16 PM

Screenshot 2023-09-11 at 1 36 32 PM

### DIFF
--- a/experimental/sdg-tracker/src/components/countries/CountriesContent.tsx
+++ b/experimental/sdg-tracker/src/components/countries/CountriesContent.tsx
@@ -275,6 +275,8 @@ function buildTileHierarchy(
   topicDcids.sort();
   if (mainTopics.length == 2) {
     topicNameStr = `${mainTopics[0].name} vs. ${mainTopics[1].name}`;
+  } else if (mainTopics.length == 1) {
+    topicNameStr = mainTopics[0].name;
   }
   return {hierarchy, orderedTiles, topicNameStr};
 }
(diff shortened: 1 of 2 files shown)


experimental/sdg-tracker/src/components/countries/CountriesContent.tsx
```diff
@@ -204,6 +228,59 @@ function addTileToHierarchy(
...
+        varToTopics[statVar].forEach((topic) => topicDcids.push(topic.dcid));
+      });
+    });
+  });
+
+  let topicNameStr = "";
+  topicDcids.sort();
+  if (mainTopics.length == 2) {
+    topicNameStr = `${mainTopics[0].name} vs. ${mainTopics[1].name}`;
+  } else if (mainTopics.length == 1) {
```
dwnoble, line 278: Can you make this an `} else {` clause with no `mainTopics.length === 1` conditional ? That way it will trigger even if mainTopics.length > 2 (in case we support that in the future)

dwnoble (approved):
Nice refactor!

experimental/sdg-tracker/src/components/countries/CountriesContent.tsx
juliawu: Done!

PR #3520, opened 2023-09-11
[bar tile] fix bug when first data group is missing datapoints

chejennifer:
- fix bug that occurs when drawing a bar chart for a list of DataGroups that don't all have the same number of DataPoints.

before: 
Screenshot 2023-09-11 at 3 41 27 PM

with the fix:
Screenshot 2023-09-11 at 3 41 40 PM


dwnoble (approved):
Thank you!

chejennifer:
thanks for the review!

PR #3522, opened 2023-09-12
Sdg country select fix

dwnoble:
* fixed country switching bug.
* Added null check to formatDate in highlight tile


pradh (approved):

juliawu (approved):
Thank you!

PR #3524, opened 2023-09-12
[vis tools] source selector

chejennifer:
- add source selector for vis tools
    - note: facet id is not remembered in the url hash because facet Ids can become stale

[screen-recording - 2023-09-11T173921.005.webm](https://github.com/datacommonsorg/website/assets/69875368/b43ca6ef-cd46-44e0-b8c0-d8f37b7e4fe6)
[screen-recording - 2023-09-11T174046.492.webm](https://github.com/datacommonsorg/website/assets/69875368/ade0a4d5-c045-4710-82f9-7783ed572fc4)
[screen-recording - 2023-09-11T174208.924.webm](https://github.com/datacommonsorg/website/assets/69875368/db4904cd-923d-4656-b365-fb866f89c2ae)


static/js/utils/data_fetch_utils.ts
```diff
@@ -196,12 +209,101 @@ export function getSeriesWithin(
...
+            result[svDcid][obs.facet] = respData.facets[obs.facet];
+          }
+        });
+      }
+      return result;
+    });
+}
+
+/**
+ * Gets the available favets for each variable given a list of entities.
```
shifucun, line 273: s/favets/facets

shifucun (commented):
This is very cool! I would suggest adding one webdriver test for each tool, and make sure facets can be selected and charts can show up!

static/js/utils/data_fetch_utils.ts
chejennifer: done

chejennifer:
> This is very cool! I would suggest adding one webdriver test for each tool, and make sure facets can be selected and charts can show up!

thanks for the review! I agree with adding webdriver tests, but maybe in a separate PR?

shifucun (approved):

PR #3534, opened 2023-09-12
[map] update more projections

chejennifer:
EU & Melanesia 

Screenshot 2023-09-12 at 10 35 52 AM
Screenshot 2023-09-12 at 10 36 04 AM


beets (approved):
Melanasia looks better now, but it looks very skewed / stretched. It might need a different projection..

chejennifer:
thanks for the review! I'll add a TODO to look  into a new melanesia projection

PR #3535, opened 2023-09-12
[UN SDG] Fix infographic for 12.2.1

juliawu:
Updates the JSON to point to the right infographic.

Also made a pass through of all other indicators in the JSON to make sure indicator number in the image path matched the indicator key, but the rest look ok.


dwnoble (approved):
Thank you!! good catch.

PR #3545, opened 2023-09-12
Consolidate sanity/screenshot/adversarial cloudbuild with domain support

shifucun:


build/ci/cloudbuild.website_testing.yaml
```diff
@@ -13,16 +13,38 @@
 # limitations under the License.
 
 steps:
+  # Run Website sanity
+  - id: website_adversarial
```
keyurva, line 17: I'm wondering if it's worthwhile keeping these cloudbuild files around so an individual test can be triggered manually?

keyurva (approved):
Thanks Bo!

build/ci/cloudbuild.website_testing.yaml
shifucun: Now we can still trigger those, but they would all run, would that work? I am bit worried maintaining several copies of very similar files..

run_website_sanity.sh
```diff
@@ -15,21 +15,18 @@
...
+# DC Instance domain like: "dev.datacommons.org", "datacommons.org"
+domain=$1
+echo "Domain: $domain"
+
 export FLASK_ENV=webdriver
 export GOOGLE_CLOUD_PROJECT=datcom-website-dev
 
 python3 -m venv .env
 source .env/bin/activate
 pip3 install -r server/requirements.txt
```
pradh, line 27: nit: maybe add a check that domain doesn't start with `http`?

pradh (commented):
Thanks!

PR #3539, opened 2023-09-12
[UN SDG] Add footnotes to tiles

juliawu:
Add footnotes to the chart tiles:
* Plumb through footnotes to from fulfill endpoints
* Add a Chart footnote react element
* Adds "show more/show less" toggling for long text

Screenshots:
Screenshot 2023-09-12 at 12 39 21 PM
Screenshot 2023-09-12 at 12 34 18 PM
Screenshot 2023-09-12 at 12 30 19 PM
Screenshot 2023-09-12 at 12 30 29 PM

PR #3547, opened 2023-09-12
Move tile footnote changes to correct folder, fix search page bug

juliawu:
Move changes from #3539 into the new sdg-tracker-un folder. Also fixes the search bug in both sdg-tracker and sdg-tracker-un.


dwnoble (approved):
Nice

PR #3550, opened 2023-09-12
[UN SDG] Update place and topic name headers

juliawu:
Updates logic of place and topic name header so that:
* Don't show the topic name of dc/topic/Root ("Statistics")
* Don't show topic name on search page if place wasn't found
* Don't show place name on search page if no topic was found

Place, but no topic:
Screenshot 2023-09-12 at 3 42 09 PM

No place or topic:
Screenshot 2023-09-12 at 3 14 51 PM

Place and topic (no change to behavior):
Screenshot 2023-09-12 at 3 42 09 PM


dwnoble (approved):
Nice!

PR #3554, opened 2023-09-13
Add custom time scale to x-axis of line charts

juliawu:
Allows for selection of a custom time scale in line web component, and sets the line charts in the UN sdg pages to use only years in the x-axis.

Example:
```
<datacommons-line
  header="Sample line"
  places="country/COL"
  variables="sdg/SI_POV_NAHC"
/>
```
Gives
Screenshot 2023-09-12 at 6 52 14 PM

Instead of
Screenshot 2023-09-12 at 6 52 48 PM


static/js/chart/types.ts
```diff
@@ -170,6 +170,7 @@ export interface HorizontalBarChartOptions extends ChartOptions {
 
 export interface LineChartOptions extends ChartOptions {
   handleDotClick?: (dotData: DotDataPoint) => void;
+  timeScale?: string;
```
dwnoble, line 173: Use typing `"year" | "month" | "day"` instead of `string`

static/js/components/tiles/line_tile.tsx
```diff
@@ -80,6 +80,8 @@ export interface LineTilePropType {
   showTooltipOnHover?: boolean;
   // Function used to get processed stat var names.
   getProcessedSVNameFn?: (name: string) => string;
+  // Time scale to use on x-axis. "year", "month", or "day"
+  timeScale?: string;
```
dwnoble, line 84: same as above

static/library/line_chart_component.ts
```diff
@@ -111,6 +111,10 @@ export class DatacommonsLineComponent extends LitElement {
   @property({ type: Boolean })
   showExploreMore: boolean;
 
+  // Optional: Timescale to use for x-axis labels
+  @property()
+  timeScale?: string;
```
dwnoble, line 116: ditto

static/js/chart/types.ts
juliawu: Good call! Done!

static/js/components/tiles/line_tile.tsx
juliawu: Done.

static/library/line_chart_component.ts
juliawu: Done.

dwnoble (approved):

PR #3568, opened 2023-09-13
Remove ga event console logging

shifucun:


beets (approved):
ouch - thanks for the fix

PR #3583, opened 2023-09-14
Remove browser landing page bogus statement on search box

shifucun:


beets (approved):
thanks for the catch and fix!

PR #3585, opened 2023-09-14
[sdg] remove explore more link for region maps

chejennifer:
Screenshot 2023-09-14 at 11 00 33 AM
Screenshot 2023-09-14 at 11 00 56 AM

also remove for tiles rendered by dc:
Screenshot 2023-09-14 at 1 52 09 PM


static/library/utils.ts
```diff
@@ -41,6 +41,17 @@ export function convertArrayAttribute(attributeValue: string): string[] {
...
 
+/**
+ * Custom attribute converter for boolean type attributes.
+ * Checks that the attribute value equals "true"
+ *
+ * @param attributeValue the attribute value provided to the web component
+ * @returns boolean
+ */
+export function convertBooleanAttribute(attributeValue: string): boolean {
+  return attributeValue.toLowerCase() !== "false";
```
beets, line 52: might be better to explicitly test for "true"?

beets (approved):

static/library/utils.ts
chejennifer: that's what I originally wanted to do, but therea re example cases like these https://github.com/datacommonsorg/website/blob/45a226a04d47d0296849afe61c3ce3fe87376fa3/packages/web-components/examples/all-charts.html#L158-L159 that need to evaluate to true ..

chejennifer:
thanks for the review!

static/library/utils.ts
beets: Ah - got it. A comment to that effect would be good to have.

chejennifer: udpated in https://github.com/datacommonsorg/website/pull/3595

PR #3595, opened 2023-09-14
[sdg] updates

chejennifer:
- update "m^s" and "km^2" to use superscripts
- fix color coding on map legend
- update a util function comment from https://github.com/datacommonsorg/website/pull/3585
Screenshot 2023-09-14 at 3 35 44 PM
Screenshot 2023-09-14 at 3 36 42 PM
Screenshot 2023-09-14 at 4 17 48 PM


static/js/utils/tile_utils.tsx
```diff
@@ -376,6 +377,18 @@ export function getSourcesJsx(sources: Set<string>): JSX.Element {
   );
 }
 
+// Processes a unit string
```
juliawu, line 380: Let's add a comment that this converts a units with "^" into superscript.

juliawu (approved):
Very nice!

static/js/utils/tile_utils.tsx
chejennifer: done

chejennifer:
thanks for the review!

PR #3598, opened 2023-09-15
[UN SDG] Only show goal overviews on goal page

juliawu:
Adds logic to prevent search results showing the sdg goal overviews instead.

Before, when searching "compare sdg goals in BRICS countries":
Screenshot 2023-09-14 at 5 32 45 PM

After:
Screenshot 2023-09-14 at 5 32 55 PM


chejennifer (approved):
nice

PR #3599, opened 2023-09-15
SDG: Add maps for all the regions

pradh:
Might need to drop "Oceania excluding Aus and NZ", and Northern Europe could use some tweaks.

UNMaps2


packages/web-components/examples/all-charts.html
```diff
@@ -332,6 +332,175 @@ <h2>UN Map Comparisons</h2>
       variable="Count_Person"
       geoJsonProp="geoJsonCoordinatesUN"
     ></datacommons-map>
+    <h3>UN Regions</h3>
```
chejennifer, line 335: nit: not sure we want to add all these maps to this example html since it's supposed to be used to show how different features of the web components are used .. what do you guys think, @dwnoble @juliawu?

chejennifer (approved):
nice!

packages/web-components/examples/all-charts.html
pradh: Let me move it to a different html?  I think its a very useful testing tool :)

pradh: Any ideas where to put the new html?

chejennifer: yeah moving to a different html sounds good to me & we could just keep it in the examples folder!

pradh: Moving it to server/config/geojson/examples.html...

pradh: Moved PTAL.

chejennifer: lgtm!

PR #3600, opened 2023-09-15
Hack around lack of containedInPlace

pradh:
Also drop "Oceania without Aus and NZ" from the maps.

Also drop the block list in UNSDG app.

image

image


static/library/constants.ts
```diff
@@ -15,7 +15,7 @@
  */
 
 /** Default website API endpoint for datacommons library */
-export const DEFAULT_API_ENDPOINT = "https://datacommons.org";
+export const DEFAULT_API_ENDPOINT = "http://localhost:8080";
```
dwnoble, line 18: Revert to datacommons.org

dwnoble (approved):

static/library/constants.ts
pradh: Thanks for the catch!!

PR #3601, opened 2023-09-15
[sdg] Updated navigation links to properly update header when clicking a new goal or on a search query

dwnoble:
The issue was due to a `<BrowserRouter />` wrapped in another `<BrowserRouter />` . The inner (data commons) browser router was preventing the outer (UN) browser router from properly updating its state when navigating around the page


experimental/sdg-tracker-un/frontend-tester/UNSD-Website-skeleton/UNSD.Website/ClientApp/src/templates/DataCommons/components/search/Search.tsx
```diff
@@ -336,7 +342,7 @@ const Search = () => {
                 {QUERIES.general.map((q, i) => (
                   <StyledLinkContainer key={i}>
                     <IconSquare color={theme.sdgColors[q.goal - 1]} />
-                    <StyledLink to={`/search?q=${q.query}`}>
+                    <StyledLink to={`search?q=${q.query}`}>
```
juliawu, line 345: Just to be sure, should this, and line 362 match line 378?

This code works fine when I click the links locally, but maybe we want to standardized the "to=" field?

juliawu (approved):
Amazing!

experimental/sdg-tracker-un/frontend-tester/UNSD-Website-skeleton/UNSD.Website/ClientApp/src/templates/DataCommons/components/search/Search.tsx
dwnoble: Good catch! made these consistent

PR #3602, opened 2023-09-15
[sdg] Updated web components CSS file to use the minimum required styles

dwnoble:
Previously the datacommons.css file (used only by web components) was exporting too many styles, causing the UN's site to default to Google Sans. 

This updates datacommons.css to export only the minimum css settings, namely:
* CSS variables used by web web components
* material design icons
* google sans fonts (only applied to the web components themselves, no longer to the parent page)
* bootstrap - still required for the "Download" popup, though we should remove this requirement in the future


pradh (approved):
Nice fix!

PR #3616, opened 2023-09-15
[sdg] fix "invalid size" error

chejennifer:
- fix a problem where we are trying to draw a scatter chart in a container that has 0 width. This is a quick fix & I've added a todo to figure out/fix the root cause


juliawu (approved):
Thanks for the quick fix!

chejennifer:
thanks for the review

PR #3618, opened 2023-09-16
NL: Improve logging, etc

pradh:
This PR makes 3 fixes:

1. Logs to BT when we abort a query (for bad-words or otherwise).
2. Supports `explore` to specify a url param `test=<type>`, which gets plumbed into the response which is logged.
3. Drops a few signal words in LLM response to avoid false positive blocks.

To validate (1), I ran tests locally and was able to find it in BT (see screenshot)

image

image

image

### DIFF
--- a/static/js/apps/explore/app.tsx
+++ b/static/js/apps/explore/app.tsx
@@ -259,6 +259,7 @@ export function App(props: { isDemo: boolean }): JSX.Element {
     );
     const detector = getSingleParam(hashParams[URL_HASH_PARAMS.DETECTOR]);
     const llmApi = getSingleParam(hashParams[URL_HASH_PARAMS.LLM_API]);
+    const testMode = getSingleParam(hashParams[URL_HASH_PARAMS.TEST_MODE]);
 
     let fulfillmentPromise: Promise<any>;
     const gaTitle = query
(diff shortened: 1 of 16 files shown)


keyurva (approved):

shifucun (approved):
very nice!

static/js/apps/explore/app.tsx
```diff
@@ -365,14 +370,16 @@ const fetchDetectAndFufillData = async (
...
+  testMode: string
 ) => {
   const detectorTypeArg = detector ? `&detector=${detector}` : "";
   const llmApiArg = llmApi ? `&llm_api=${llmApi}` : "";
+  const testModeArg = testMode ? `&test=${testMode}` : "";
   try {
     const startTime = window.performance ? window.performance.now() : undefined;
     const resp = await axios.post(
-      `/api/explore/detect-and-fulfill?q=${query}${detectorTypeArg}${llmApiArg}`,
+      `/api/explore/detect-and-fulfill?q=${query}${detectorTypeArg}${llmApiArg}${testModeArg}`,
```
shifucun, line 382: for later, shall we move these from url param to POST data? There is a combination of both now.

pradh: That makes sense!

PR #3621, opened 2023-09-18
fix some key errors

chejennifer:
- fix some key errors in flask


beets (approved):
thanks for these fixes!

chejennifer:
thanks for the review!

PR #3622, opened 2023-09-19
Add a chart picker util

pradh:


shifucun (commented):
At a high-level, do we produce chart for each config with the same data? Is there a plan to cache the data fetching?

tools/charts/picker.py
```diff
@@ -0,0 +1,368 @@
...
+  ctx.writer.add('map', parts)
+
+
+def _ranking(v: str, title: str, ctx: Context):
+  parts = [
+      f'header="{title}"',
+      f'parentPlace="{ctx.place}"',
+      f'childPlaceType="{ctx.child_type}"',
+      f'variable="{v}"',
+      'rankingCount=10',
```
juliawu, line 161: `rankingCount` isn't supported by the web components yet so this line and line 171 will get ignored.

juliawu (approved):
Very nice!

chejennifer (approved):
cool

tools/charts/picker.py
juliawu: Update: working on adding this now, we can leave this in for now!

PR #3623, opened 2023-09-20
[web-components] Added css variable overrides for web component font family

dwnoble:
Added css variables

```
--dc-font-family
--dc-headings-font-family
```

for web component users to set the font family

Example (Top: google sans (default), bottom: monospace (overridden))
Screenshot 2023-09-19 at 4 53 35 PM


juliawu (approved):
Nice!

PR #3626, opened 2023-09-21
Removed datacommon.css requirement from web components

dwnoble:
Previously datacommons.css was used to 

1. include google sans and material icons web fonts
2. include bootstrap styles used by the "download" modal
3. define css variables used by web component

This change:
1. Moves google sans and material icons definition to datacommons.js
2. Moves bootstrap modal to inside the web component itself, meaning we no longer need to include these styles at the parent component
3. Adds css variable definitions locally to the web components using `:host` definition 
4. Updates README and examples to remove datacommons.css


chejennifer (approved):
nice, thanks for the clean up!

PR #3629, opened 2023-09-22
use un geojsons for all maps on unsdg dc

chejennifer:
- add new GEO_JSON_PROP to the app config & read that when retrieving geojsons

### DIFF
--- a/server/routes/shared_api/choropleth.py
+++ b/server/routes/shared_api/choropleth.py
@@ -237,6 +237,8 @@ def geojson():
   if not place_type:
     place_dcid, place_type = get_choropleth_display_level(place_dcid)
   place_name_prop = request.args.get("placeNameProp")
+  # If the request has a geoJsonProp, use that. Otherwise, use the default
+  # property specified in the app config.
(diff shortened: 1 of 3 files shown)


pradh (approved):
Very nice!

@shifucun is there some way to test this change before rolling out to unsdg?

server/routes/shared_api/choropleth.py
```diff
@@ -238,7 +238,7 @@ def geojson():
     place_dcid, place_type = get_choropleth_display_level(place_dcid)
   place_name_prop = request.args.get("placeNameProp")
   geojson_prop = request.args.get("geoJsonProp",
-                                  CHOROPLETH_DEFAULT_GEOJSON_PROP)
+                                  current_app.config["GEO_JSON_PROP"])
```
pradh, line 241: To confirm, when web-components sets geoJsonProp that will override the backend config?  Add a comment to that effect maybe?

chejennifer: correct, if the client calls this endpoint with a geoJsonProp, that value will be used. Added a comment.

PR #3632, opened 2023-09-25
Deepcopy topic page config before it is mutated

shifucun:
Thanks @caravansteve for finding this issue in the techsoup DC instance!


chejennifer (approved):
nice

PR #3633, opened 2023-09-26
Unify Ranking tile, web component, and subject page proto

juliawu:
Updates the web component and subject page proto so that the fields for ranking charts match the ranking tile props.

In particular:
* RankingTileSpec entries & behavior is added to the web component
* Hiding footer and showing loading spinner are added as options to the proto and web component
* Web component documentation is updated with new fields
* Adds the ability to show rankings for multiple variables.
* Adds the ability to show variables per capita for the web component
* Adds the ability to adjust how many places are shown in the web component

Note that:
* `showExploreMore` is already implemented at block level in subject page protos
*  `onHoverToggled` expects a function defining mouseover behavior, so this is skipped in web component and proto.
* `diffbaseDate` doesn't appear to do anything. This is removed.


server/config/subject_page.proto
```diff
@@ -279,6 +279,14 @@ message Tile {
   // [optional] If set, used instead of main place for the page.
   string place_dcid_override = 11;
 
+  // [optional] used only in ranking tiles
+  // set to true to hide the footer with download / explore more link
+  bool hide_footer = 17;
+
+  // [optional] used only in ranking tiles
+  // whether to show loading spinner during data fetch
+  bool show_loading_spinner = 18;
```
chejennifer, line 288: I don't actually think show_loading_spinner is something we should add as a config/param ..all tiles should just work the same way in terms of loading treatment and I don't think there's been a case so far where we've wanted a specific loading treatment for a tile .. wdyt?

packages/web-components/docs/components/ranking.md
```diff
@@ -37,12 +37,54 @@ Required:
...
+
+  Default: 5.
+
+- `showExploreMore` _boolean_
+
+  Whether to show "Explore more" link, which takes the user to Datacommons.org's visualization tools.
+
+- `showHighestLowest` _boolean_
+
+  Whether to show a high-to-low ranking stacked on top of a low-to-high ranking, in one chart.
```
chejennifer, line 68: this actually just shows low to high with "..." between top N and bottom N
e.g., 1, 2, 3 ... 9, 10, 11

static/library/ranking_component.ts
```diff
@@ -61,26 +65,64 @@ export class DatacommonsRankingComponent extends LitElement {
   @property()
   apiRoot: string;
 
+  // Type of child place to rank (ex: State, County)
+  @property()
+  childPlaceType!: string;
+
+  @property()
+  diffBaseDate?: string;
```
chejennifer, line 73: add a comment for what this is? But also, can you check if this is even used anywhere & maybe we can get ride of it

static/library/ranking_component.ts
```diff
@@ -61,26 +65,64 @@ export class DatacommonsRankingComponent extends LitElement {
...
 
-  // DCID of the statistical variable to compare values for
-  @property()
-  variable!: string;
+  // Optional: Set to true to show an "explore more" link to visualization tools
+  // Default: false
+  @property({ type: Boolean, converter: convertBooleanAttribute })
+  showExploreMore?: boolean;
+
+  // Optional: Set to true to show both high-to-low and low-to-high rankings
```
chejennifer, line 108: update this comment

static/library/ranking_component.ts
```diff
@@ -89,32 +131,50 @@ export class DatacommonsRankingComponent extends LitElement {
   @property()
   title!: string;
 
+  // DCID of the statistical variable to compare values for
```
chejennifer, line 134: nit: rank values for?

static/library/ranking_component.ts
```diff
@@ -89,32 +131,50 @@ export class DatacommonsRankingComponent extends LitElement {
...
+
+  // List of DCIDs of the statistical variable to compare values for
+  @property({ type: Array<string>, converter: convertArrayAttribute })
+  variables?: string[];
+
   render(): HTMLElement {
+    const variables = this.variables || [this.variable];
+    const statVarSpec = variables.map((statVar) => {
+      return {
+        denom: "",
```
chejennifer, line 146: hm, one more unifying thing is we should allow per capita for web components? and other relevant stat var spec things

chejennifer (commented):
nice!

server/config/subject_page.proto
juliawu: This sounds good to me! Removed as a config/param.

packages/web-components/docs/components/ranking.md
juliawu: Good catch! Updated comments and docs.

static/library/ranking_component.ts
juliawu: Updated!

static/library/ranking_component.ts
juliawu: Checked, this doesn't appear to be used anywhere. Removed all references.

static/library/ranking_component.ts
juliawu: Changed!

static/library/ranking_component.ts
juliawu: Good idea. Added a perCapita field for the web components. I'll add these in as we update each of the tiles.

server/config/subject_page.proto
```diff
@@ -279,6 +279,10 @@ message Tile {
   // [optional] If set, used instead of main place for the page.
   string place_dcid_override = 11;
 
+  // [optional] used only in ranking tiles
+  // set to true to hide the footer with download / explore more link
+  bool hide_footer = 17;
```
chejennifer, line 284: nit: maybe should put this in RankingTileSpec since it's specific to ranking tiles?

chejennifer (approved):
thanks for the updates!

server/config/subject_page.proto
juliawu: Looked into this, if we put `hide_footer` into `RankingTileSpec` in the proto, cleanest implementation is to also move `hideFooter` into `rankingMetadata` on the tile side, to avoid type errors in `block.tsx` but that feels a little odd to me. Thoughts?

juliawu: Going to merge this for now, and we can followup with a props cleanup PR after some more discussion about how to refactor/abstract.

PR #3634, opened 2023-09-26
Add support for fetching descendant names for non-place entities.

keyurva:
API results:

[non-place-descendants.webm](https://github.com/datacommonsorg/website/assets/1221814/03d54559-00aa-45b8-b97e-9ba040a11446)

With this change, we can now do scatter plots for non-place entities:

image

### DIFF
--- a/server/lib/fetch.py
+++ b/server/lib/fetch.py
@@ -436,10 +436,15 @@ def triples(nodes, out=True):
 
 
 def descendent_places(nodes, descendent_type):
-  return property_values(nodes,
-                         'containedInPlace+',
-                         out=False,
-                         constraints='{{typeOf:{}}}'.format(descendent_type))
+  # When the only node being requested is also the descendent_type, fetch all nodes of that type.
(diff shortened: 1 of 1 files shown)


server/lib/fetch.py
```diff
@@ -436,10 +436,15 @@ def triples(nodes, out=True):
 
 
 def descendent_places(nodes, descendent_type):
-  return property_values(nodes,
-                         'containedInPlace+',
-                         out=False,
-                         constraints='{{typeOf:{}}}'.format(descendent_type))
+  # When the only node being requested is also the descendent_type, fetch all nodes of that type.
```
shifucun, line 439: Can you add a TODO to re-visit this. Maybe have a bug to track too.

shifucun (approved):
thanks!

server/lib/fetch.py
keyurva: Done. b/302167251

keyurva:
Thanks Bo.

PR #3635, opened 2023-09-26
[tiles] fix per capita scaling bug

chejennifer:
- fix bug when svSpec.scaling was undefined, we were trying to multiply undefined by 100 for per capita

Screenshot 2023-09-26 at 1 23 24 PM


pradh (approved):
Thanks!

PR #3636, opened 2023-09-26
NL: Link to Place Pages

pradh:
https://github.com/datacommonsorg/website/assets/4375037/c11b98bb-a380-4bdb-af2f-f1213fb89a64

### DIFF
--- a/static/src/server.ts
+++ b/static/src/server.ts
@@ -162,7 +162,6 @@ function getTextLength(text: string): number {
   };
 };
 
-
 // Gets a promise for a single tile result
 function getTileResult(
   id: string,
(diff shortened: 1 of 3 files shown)


static/src/server.ts
```diff
@@ -174,47 +173,47 @@ function getTileResult(
...
-          false
-        )
+        id,
+        tileConfig,
+        place,
+        svSpec,
+        CONFIG.apiRoot,
+        "",
+        false
+      );
```
pradh, line 183: These changes are all from running lint.

chejennifer (approved):
nice!

pradh:
Thanks for the review!

beets:
> nice!
> 
> One thing about the UI/UX, I'm not sure how I feel about the blue for the page header & not sure how obvious it is to a user that the header is clickable .. maybe we can think about adding an icon instead or some link that a user can click to go to the place page. (Fine to defer to future PR)

+1 : maybe we can add another line under "related topics" with "explore more: <place name>" which would be more obvious. that could work for multi-place queries too

pradh:
On Wed, Sep 27, 2023 at 10:03 AM Carolyn Au ***@***.***>
wrote:

> nice!
>
> One thing about the UI/UX, I'm not sure how I feel about the blue for the
> page header & not sure how obvious it is to a user that the header is
> clickable .. maybe we can think about adding an icon instead or some link
> that a user can click to go to the place page. (Fine to defer to future PR)
>
> +1 : maybe we can add another line under "related topics" with "explore
> more: " which would be more obvious. that could work for multi-place
> queries too
>
> —
>

Sorry, I missed Jennifer’s original comment.  Explore More: idea is
interesting, but most of the times there will only be one place?  Would an
icon be better?

static/src/server.ts
shifucun: @juliawu can you help fix the lint issue in cloudbuild?

beets:
I'm not sure where we'd place an icon and it would be confusing for the
multi-place query. I could see the "explore more" line growing to include
more out-links..

pradh:
> I'm not sure where we'd place an icon and it would be confusing for the multi-place query. I could see the "explore more" line growing to include more out-links..

fair enough.  RE: out-links, what others did you have in mind?

another thing is we also have an "explore" thing for SV slicing, not sure if we want to call it different? 
image

do you want me to roll this back, so one of the FE folks can do the "explore more:" thing?

beets:
I don't think we should roll it back, but we could rope in others for their
thoughts.

Maybe "more about: <place 1>, <place 2>"
Other outlinks could be topic pages etc..

pradh:
> Maybe "more about: <place 1>, <place 2>" Other outlinks could be topic pages etc..

For a user, topic pages vs. explore isn't distinguishable, so do you think those links could be confusing with Related topics"?

Multi place comparison doesn't seem common, so most of the time it will jut be "more about: <place_1>" which repeats the name of the place.  Would it be odd to do "more about:" only for multi places?

PR #3637, opened 2023-09-26
Fixes for ranking and bar chart legend links

dwnoble:
* Fixes ranking and bar chart legend links so they no longer get redirected to `file:///`
* Updated ranking and bar chart legend links to redirect to `/browser` page instead of `/places` if the entities are not places


keyurva (approved):
Very cool! Thanks Dan.

PR #3640, opened 2023-09-27
Increase element wait timeout for screenshot tests

shifucun:
Removed duplicated ASYNC_ELEMENT_CLASS tagging, which exist twice in ranking tile.

### DIFF
--- a/static/js/components/tiles/sv_ranking_units.tsx
+++ b/static/js/components/tiles/sv_ranking_units.tsx
@@ -21,7 +21,6 @@ import React, { RefObject, useRef } from "react";
 
 import { VisType } from "../../apps/visualization/vis_type_configs";
 import { URL_PATH } from "../../constants/app/visualization_constants";
-import { ASYNC_ELEMENT_CLASS } from "../../constants/css_constants";
 import {
   RankingData,
   RankingGroup,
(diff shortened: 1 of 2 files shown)


static/js/components/tiles/sv_ranking_units.tsx
```diff
@@ -100,9 +99,7 @@ export function SvRankingUnits(props: SvRankingUnitsProps): JSX.Element {
   return (
     <React.Fragment>
       {rankingMetadata.showHighestLowest || props.errorMsg ? (
-        <div
-          className={`ranking-unit-container ${ASYNC_ELEMENT_CLASS} highest-ranking-container`}
```
keyurva, line 104: Do the webdriver tests run without issues with this change?

static/js/components/tiles/sv_ranking_units.tsx
```diff
@@ -21,7 +21,6 @@ import React, { RefObject, useRef } from "react";
 
 import { VisType } from "../../apps/visualization/vis_type_configs";
 import { URL_PATH } from "../../constants/app/visualization_constants";
-import { ASYNC_ELEMENT_CLASS } from "../../constants/css_constants";
```
pradh, line 24: For my understanding, why remove this?  There are other places that use it too.

pradh (approved):
Thanks!

static/js/components/tiles/sv_ranking_units.tsx
shifucun: This is a dup. The child html also have this class added and should be preferred.

static/js/components/tiles/sv_ranking_units.tsx
shifucun: ASYNC_ELEMENT_CLASS in this file is a dup. Wherever it's used, the child html also have this class added and should be preferred.

PR #3642, opened 2023-09-27
[node server] enable multiple places for line chart

chejennifer:
Screenshot 2023-09-27 at 4 06 16 PM

url used to get this: http://localhost:3030/nodejs/chart-info?config={%22comparisonPlaces%22:[%22self%22,%22geoId/06%22],%22title%22:%22Prevalence%20of%20Health%20Conditions%20in%20Placer%20County%20(${date})%22,%22type%22:%22LINE%22}&place=geoId/06061&svSpec=[{%22name%22:%22Chronic%20Kidney%20Disease%22,%22statVar%22:%22Percent_Person_WithChronicKidneyDisease%22,%22unit%22:%22%25%22},{%22name%22:%22Chronic%20Obstructive%20Pulmonary%20Disease%22,%22statVar%22:%22Percent_Person_WithChronicObstructivePulmonaryDisease%22,%22unit%22:%22%25%22},{%22name%22:%22Coronary%20Heart%20Disease%22,%22statVar%22:%22Percent_Person_WithCoronaryHeartDisease%22,%22unit%22:%22%25%22},{%22name%22:%22Diabetes%22,%22statVar%22:%22Percent_Person_WithDiabetes%22,%22unit%22:%22%25%22}]

### DIFF
--- a/static/js/components/subject_page/block.tsx
+++ b/static/js/components/subject_page/block.tsx
@@ -27,7 +27,6 @@ import {
   COLUMN_ID_PREFIX,
   HIDE_COLUMN_CLASS,
   HIDE_TILE_CLASS,
-  SELF_PLACE_DCID_PLACEHOLDER,
   TILE_ID_PREFIX,
 } from "../../constants/subject_page_constants";
 import { NamedPlace, NamedTypedPlace } from "../../shared/types";
@@ -41,6 +40,7 @@ import {
   getId,
   getMinTileIdxToHide,
 } from "../../utils/subject_page_utils";
+import { getComparisonPlaces } from "../../utils/tile_utils";
 import { BarTile } from "../tiles/bar_tile";
 import { BivariateTile } from "../tiles/bivariate_tile";
 import { DonutTile } from "../tiles/donut_tile";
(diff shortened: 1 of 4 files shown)


static/js/components/subject_page/block.tsx
```diff
@@ -200,11 +200,7 @@ function renderTiles(
     const place = tile.placeDcidOverride
```
pradh, line 200: To confirm, does line tile in nodejs support child-type?  I know web-component does (https://github.com/datacommonsorg/website/blob/9aad7497955e20b180c1c661b2ef57986a096d0e/packages/web-components/examples/all-charts.html#L328-L333)

pradh (approved):
Nice, thanks!

static/js/components/subject_page/block.tsx
chejennifer: it does not .. do we want to enable it?

I chose not to enable because:
1. line tile doesn't set max number of places so the number of lines can get crazy
2. the legend can't be hovered so if there are lots of lines, don't have a way to tell which line is which

pradh: Got it.  I think this is fine for chart generation.

I think we should do (1), given web-components support setting childPlaceType?

You mean (2) doesn't work only for the SVGs right, but the tile supports it:
Screenshot 2023-09-27 at 4 27 06 PM

chejennifer: yeup, agreed that we should do (1) & yes, (2) only applies to node server svgs

PR #3639, opened 2023-09-27
Create initial script to call into Vertex AI and ingest query CSVs.

keyurva:
* PR created to get Julia unblocked for her tasks.
* More functionality / docs to follow.

PR #3644, opened 2023-09-28
Add script for using translate API on csvs

juliawu:
Followup to #3639.

Adds a script to handle en -> language, and language -> en translations of a CSV of queries.


tools/nl/translate_compare/translate_csvs.py
```diff
@@ -0,0 +1,220 @@
...
+import glob
+import os
+import pandas as pd
+from google.cloud import translate
+
+CSV_FILE_EXTENSION = ".csv"
+DEFAULT_NEW_COLUMN_NAME = "translation"
+DEFAULT_OUTPUT_LANUGAGE_CODE = "en"
+
+project_id = os.environ.get("PROJECT_ID", "")
```
keyurva, line 43: Can we accept this as a flag instead of an env variable and default it to `datcom-204919`?

tools/nl/translate_compare/translate_csvs.py
```diff
@@ -0,0 +1,220 @@
...
+        Returns:
+            translated text
+        """
+        response = self.client.translate_text(
+            parent=PARENT,
+            contents=texts,
+            source_language_code=input_lang,
+            target_language_code=output_lang,
+        )
+        translations = [resp.translated_text for resp in response.translations]
```
keyurva, line 140: Couple questions regarding the translation API: Are the translations guaranteed to be the same size and in the same order as the input texts

keyurva: Looks like it does: https://screenshot.googleplex.com/3ehF92MS6Rf3Bjk

keyurva (approved):
Very cool!

And thanks for the excellent documentation - makes it super easy to follow (and maintain long term).

tools/nl/translate_compare/translate_csvs.py
juliawu: Switched to a glag with default variable!

pradh (approved):

PR #3646, opened 2023-09-28
Fixed bug where bar charts with non-place entities were not rendering

dwnoble:
Example chart:

```
    <datacommons-bar
      apiRoot="https://floret.datacommons.org/"
      header="Sectors with lowest risk"
      parentPlace="NAICSEnum"
      childPlaceType="NAICSEnum"
      variables="sector_risk"
    ></datacommons-bar>
```

## Before
Screenshot 2023-09-28 at 1 51 01 PM

## After
Screenshot 2023-09-28 at 1 51 10 PM


static/js/components/tiles/bar_tile.tsx
```diff
@@ -227,9 +226,24 @@ export const fetchData = async (props: BarTilePropType) => {
...
+    if (_.isEmpty(statResp.data[FILTER_STAT_VAR])) {
+      const entityDcidsSet = new Set<string>();
+      Object.keys(statResp.data).forEach((statVarKey) => {
+        Object.keys(statResp.data[statVarKey]).forEach((entityDcid) => {
+          entityDcidsSet.add(entityDcid);
+        });
+      });
+      entityDcidsSet.forEach((entityDcid) => {
+        popPoints.push({
+          placeDcid: entityDcid,
```
keyurva, line 242: At some point (not advocating in this PR) it'll be good to have a cleaner distinction between place and non-place entities.

static/js/stat_var_hierarchy/stat_var_section.tsx
```diff
@@ -25,8 +25,12 @@ import React from "react";
 
 import { StatVarCharts } from "../browser/stat_var_charts";
 import { Context } from "../shared/context";
-import { NamedNode, StatVarHierarchyType } from "../shared/types";
-import { StatVarInfo, StatVarSummary } from "../shared/types";
+import {
```
keyurva, line 28: optional: Can't tell if changes in this package are related to the bar chart changes? If they are not, consider doing it in a separate PR.

keyurva (approved):
Thanks Dan! And very cool to see tests for web components!

static/js/components/tiles/bar_tile.tsx
dwnoble: Agreed - we should use the more generic "entity" terminology vs place

static/js/stat_var_hierarchy/stat_var_section.tsx
dwnoble: While not directly related, after adding `bar_tile.test.tsx`, the client side tests started failing on these test due to typing errors. Not totally sure why this started happening now, but it was blocking the PR submission

PR #3649, opened 2023-09-29
Update dev and magiceye replicas

shifucun:
DEV: Reduce web server replicas as launch as complete; Increase NodeJs replicas for large scale offline charts generation.

MagicEye: Reduce SVG replicas as it's the least needed. Update Bigtable version. Note this cache substantially increased the SVG in-memory cache built time due to the large amount of stat vars added.


pradh (approved):
Thanks!

PR #3650, opened 2023-09-29
Add support for non-English queries.

keyurva:
* Currently behind an `i18n=true` feature flag.
* Will add tests next once I figure out how to pass api keys in tests.

[non-english-query.webm](https://github.com/datacommonsorg/website/assets/1221814/9659f6e4-dd73-4e9c-a1d3-266669117d2d)


shared/lib/translator.py
```diff
@@ -0,0 +1,51 @@
...
+from langdetect import detect as lang_detect
+import requests
+
+_API_URL = "https://translation.googleapis.com/language/translate/v2"
+_API_HEADER = {'content-type': 'application/json'}
+_EN_LANG = "en"
+
+
+# Detects the query language and translates non-English queries to English.
+# TODO: Add tests once I understand out how to access the api key in tests.
```
pradh, line 27: Follow what happens with PALM_API_KEY: https://github.com/datacommonsorg/website/blob/e57ac74c8f2b6fcb0e28198bafce11f64b10d47a/server/__init__.py#L383-L392 ?

shared/lib/translator.py
```diff
@@ -0,0 +1,51 @@
...
+    request = {"q": query, "format": "text", "source": lang, "target": _EN_LANG}
+    response = requests.post(f'{_API_URL}?key={api_key}',
+                             json=request,
+                             headers=_API_HEADER).json()
+
+    translation = response.get("data",
+                               {}).get("translations",
+                                       [{}])[0].get("translatedText", "")
+    if translation:
+      logging.info("Translated '%s' to: %s", query, translation)
```
pradh, line 46: Don't log stuff?

server/routes/nl/helpers.py
```diff
@@ -104,6 +106,9 @@ def parse_query_and_detect(request: Dict, app: str, debug_logs: Dict):
   else:
     llm_api_type = LlmApiType(llm_api_type)
 
+  if i18n and i18n.lower() == 'true':
+    original_query = detect_lang_and_translate(original_query)
```
pradh, line 110: Let's move `counters = ctr.Counters()` up, pass it into detect_lang_and_translate()

And in there, update `counters.info()` when translation occurs.  And also time it with `counters.timeit()`.

shared/lib/translator.py
```diff
@@ -0,0 +1,51 @@
...
+_API_HEADER = {'content-type': 'application/json'}
+_EN_LANG = "en"
+
+
+# Detects the query language and translates non-English queries to English.
+# TODO: Add tests once I understand out how to access the api key in tests.
+def detect_lang_and_translate(query, api_key=None):
+  try:
+    lang = lang_detect(query)
+    logging.info("Language detected for '%s': %s", query, lang)
```
pradh, line 31: Drop

shared/lib/translator.py
```diff
@@ -0,0 +1,51 @@
...
+                             json=request,
+                             headers=_API_HEADER).json()
+
+    translation = response.get("data",
+                               {}).get("translations",
+                                       [{}])[0].get("translatedText", "")
+    if translation:
+      logging.info("Translated '%s' to: %s", query, translation)
+      return translation
+    return query
```
pradh, line 48: Update `counters.err` here

shared/lib/translator.py
```diff
@@ -0,0 +1,51 @@
...
+    response = requests.post(f'{_API_URL}?key={api_key}',
+                             json=request,
+                             headers=_API_HEADER).json()
+
+    translation = response.get("data",
+                               {}).get("translations",
+                                       [{}])[0].get("translatedText", "")
+    if translation:
+      logging.info("Translated '%s' to: %s", query, translation)
+      return translation
```
pradh, line 47: update `counters.info` here

shared/lib/translator.py
```diff
@@ -0,0 +1,51 @@
...
+# TODO: Add tests once I understand out how to access the api key in tests.
+def detect_lang_and_translate(query, api_key=None):
+  try:
+    lang = lang_detect(query)
+    logging.info("Language detected for '%s': %s", query, lang)
+    if lang.startswith(_EN_LANG):
+      return query
+
+    if not api_key:
+      api_key = current_app.config['PALM_API_KEY']
```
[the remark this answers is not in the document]
shifucun: I have suggested to reuse the existing key but add another API support there so we don't have to re-create api key, create secret manager for all projects.

But we should rename the key as "NL_API_KEY" here?

shared/lib/translator.py
shifucun, line 27: If we want to test the actual behavior of the translate API, then this should be covered by a golden test?  If you add one golden integration test now with other language, i would expect to work since integration test can find the api key from gcp now: https://github.com/datacommonsorg/website/blob/master/server/app_env/integration_test.py#L21

shifucun (commented):
With i18n, can we put other languages in the url with "q=". I tried it and it seems to be changed to ascii code. Does this have effect in this flow?

nl_server/requirements.txt
```diff
@@ -4,6 +4,7 @@ Flask==2.3.2
 google-cloud-pubsub==2.14.0
 google-cloud-storage==2.8.0
 gunicorn==20.1.0
+langdetect=1.0.9
```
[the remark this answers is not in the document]
keyurva: Done.

shared/lib/translator.py
```diff
@@ -0,0 +1,51 @@
+# Copyright 2023 Google LLC
```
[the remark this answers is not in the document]
keyurva: Done.

shared/lib/translator.py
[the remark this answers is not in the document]
keyurva: Should it be "CLOUD_API_KEY"? Let's discuss and I'll rename in a separate PR.

shared/lib/translator.py
keyurva: Done.

server/routes/nl/helpers.py
keyurva: Done. Didn't know about this - counters are great!

https://screenshot.googleplex.com/A4B4bCWjVfazjor

shared/lib/translator.py
keyurva: Done.

shared/lib/translator.py
keyurva: Done.

keyurva:
> With i18n, can we put other languages in the url with "q=". I tried it and it seems to be changed to ascii code. Does this have effect in this flow?

Not sure what you mean by "changed to ascii code". The characters will be URL encoded like any other special characters.

server/integration_tests/nl_test.py
```diff
@@ -247,6 +248,14 @@ def test_demo_usa_map_types(self):
             'how many people are unemployed in zip codes of washington?'
         ])
 
+  def test_translate(self):
```
pradh, line 251: Can you add this to explore_test.py?  We should have 2 tests: one for detect/ and another for detect-and-fulfill/

server/lib/translator.py
```diff
@@ -0,0 +1,57 @@
...
+
+# Detects the query language and translates non-English queries to English.
+def detect_lang_and_translate(query, counters: Counters, api_key=None):
+  try:
+    lang = lang_detect(query)
+    if lang.startswith(_EN_LANG):
+      return query
+
+    if not api_key:
+      api_key = current_app.config['PALM_API_KEY']
```
pradh, line 36: Drop the api_key param, no one is using it?

Also, check 'PALM_API_KEY' in current_app.config before access.  And if not, return query.

server/lib/translator.py
```diff
@@ -0,0 +1,57 @@
...
+          "query_translation", {
+              "query": query,
+              "translation": translation,
+              "source_language": lang,
+              "target_language": _EN_LANG
+          })
+      return translation
+    return query
+  except Exception as e:
+    counters.err(f"Error detecting / translating query: {query}\n{str(e)}", "")
```
pradh, line 56: counters.err("error_query_translation", {"query": query, "error": str(e)})

server/lib/translator.py
```diff
@@ -0,0 +1,57 @@
...
+      api_key = current_app.config['PALM_API_KEY']
+    request = {"q": query, "format": "text", "source": lang, "target": _EN_LANG}
+    response = requests.post(f'{_API_URL}?key={api_key}',
+                             json=request,
+                             headers=_API_HEADER).json()
+
+    translation = response.get("data",
+                               {}).get("translations",
+                                       [{}])[0].get("translatedText", "")
+    if translation:
```
pradh, line 45: Under what situation will translation be empty?  Is that worth marking it as an error?

pradh (commented):
Thanks for the edits.

shifucun:
> > With i18n, can we put other languages in the url with "q=". I tried it and it seems to be changed to ascii code. Does this have effect in this flow?
> 
> Not sure what you mean by "changed to ascii code". The characters will be URL encoded like any other special characters.

The video is great, to show it works! How about query with context in i18n?

shared/lib/translator.py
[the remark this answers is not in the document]
shifucun: Since we also have other API keys saved. It's a point to organize them. Good for a later discussion.

shifucun (approved):

shared/lib/translator.py
keyurva: Done.

server/integration_tests/nl_test.py
keyurva: Done.

server/lib/translator.py
keyurva: Done.

server/lib/translator.py
keyurva: Done.

server/lib/translator.py
keyurva: It should never be empty. Just being defensive. Marked it as an error if it at all happens.

server/integration_tests/explore_test.py
```diff
@@ -323,6 +336,11 @@ def test_e2e_superlatives(self):
     ],
                                 test='unittest')
 
+  def test_e2e_translate(self):
+    # Chinese query for "which cities in the Santa Clara County have the highest larceny?"
+    self.run_detect_and_fulfill('e2e_translate_chinese', ['圣克拉拉县哪些城市的盗窃率最高？'],
```
pradh, line 341: To Bo's point, maybe we can also add another query "what about car theft?", exercise context.

pradh (approved):
Nice job!

server/integration_tests/explore_test.py
keyurva: Good idea both. Done.

server/integration_tests/explore_test.py
```diff
@@ -55,20 +61,26 @@ def run_detect_and_fulfill(self,
...
+              'dc': dc,
+          }).json()
       ctx = resp['context']
       if len(queries) == 1:
         d = ''
       else:
         d = q.replace(' ', '').replace('?', '').lower()
+        # For some queries like Chinese, no characters are replaced and leads to unwieldy folder names.
+        # Use the query index for such cases.
+        if d == q:
```
pradh, line 82: maybe do if iso18n is true?

keyurva: Yeah, that's safer. Done.

keyurva:
Thanks guys for the detailed reviews!

PR #3651, opened 2023-09-29
[Bug Fix] Fix stacked bar charts rendering NaN values + correct tooltip values.

juliawu:
Fixes stacked bar charts such that:

1. When no data is available for a place+statVar, it doesn't get plotted.

Before:
Screenshot 2023-09-29 10 01 03 AM

After:
Screenshot 2023-09-29 9 59 56 AM

3. Tooltips now show the correct values.

Before:
Screenshot 2023-09-29 10 01 39 AM

After:
Screenshot 2023-09-29 9 58 53 AM


static/js/chart/draw_bar.ts
```diff
@@ -1058,7 +1062,6 @@ export function drawHorizontalBarChart(
       curr[dataPoint.label] = dataPoint.value;
       curr.dcid = dataPoint.dcid;
       curr.date = dataPoint.date;
-      curr.value = dataPoint.value;
```
juliawu, line 1061: This used to be used to render values on the tooltips, but is no longer used. `d3.stack()` mangles the value downstream, resulting in incorrect tooltips. Bar lengths and values are already calculated using the value in line 1058.

static/js/chart/draw_bar.ts
```diff
@@ -326,14 +326,16 @@ export function drawStackBarChart(
...
-    return d.map((item) => ({
-      date: item.data.date,
-      place: item.data.label,
-      statVar: d.key,
-      unit: options?.unit,
-      value: item.data.value,
-      ...item,
-    }));
+    return d
+      .filter((item) => !isNaN(item[0]) && !isNaN(item[1]))
```
chejennifer, line 330: curious, will item always be an array with at least 2 values? maybe leave a comment about this?

server/config/subject_page_pb2.py
```diff
@@ -1,10 +1,10 @@
 # -*- coding: utf-8 -*-
 # Generated by the protocol buffer compiler.  DO NOT EDIT!
 # source: subject_page.proto
-"""Generated protocol buffer code."""
-from google.protobuf.internal import builder as _builder
+
```
chejennifer, line 4: should there be changes in this file ..?

chejennifer (commented):
nice!

static/js/chart/draw_bar.ts
dwnoble: +1
Would it make sense to add a bounds check here and filter out the items if the size != 2?

juliawu: Yes, item will always have at least 2 items. It's a `d3.SeriesPoint` with the weird syntax [elem1, elem2, data:{...}]. I think adding a bounds check is a great idea though. Added to both instances of .filter().

server/config/subject_page_pb2.py
juliawu: These changes are generated for me whenever I run `./run_server.sh`. Since it's autogenerated it shouldn't matter, but removed these changes anyway.

chejennifer (approved):
thanks for the updates!

PR #3661, opened 2023-10-03
Refactor Tile Interfaces 

juliawu:
Following offline conversation, refactors the hierarchy of tile interfaces. Also changes the tile props of the ranking tile as an example.


dwnoble (approved):
Nice refactor! thank you

PR #3667, opened 2023-10-05
Bundle import tool into RSI docker image; Add website endpoint for RSI data load

shifucun:


server/routes/import_wizard/html.py
```diff
@@ -30,3 +31,58 @@ def main():
 @bp.route('/')
 def main_new():
   return render_template('/import_wizard2.html')
+
+
+# This is only used in CSV + SQL backend, converting raw CSV data into resolved
+# CSV.
+@bp.route('/simple/convert', methods=['POST'])
+def convert():
+  RAW_DATA_PATH = os.environ.get("RAW_DATA_PATH")
```
keyurva, line 40: I'm wondering if dath path can be an optional parameter? It can help with testing.

server/routes/import_wizard/html.py
```diff
@@ -30,3 +31,58 @@ def main():
...
+  SQL_DATA_PATH = os.environ.get("SQL_DATA_PATH")
+  command1 = [
+      "python",
+      "import/simple/stats/main.py",
+      "--input_path",
+      f"{RAW_DATA_PATH}/input.csv",  # TODO: use gcs and local folder path.
```
keyurva, line 46: Once the script supports folders with a config.json, I'm assuming we'll change this as well?

server/routes/import_wizard/html.py
```diff
@@ -30,3 +31,58 @@ def main():
...
+      "--output_dir",
+      f"{SQL_DATA_PATH}",
+  ]
+  # TODO: save debug information in another folder
+  command2 = ["rm", f"{SQL_DATA_PATH}/debug_resolve.csv"]
+  output = []
+  for command in [command1, command2]:
+    try:
+      result = subprocess.run(command, capture_output=True, text=True)
```
keyurva, line 55: Is this blocking? Any thoughts on how this would work for longer running jobs?

keyurva (approved):
This is a great start - thanks!

server/routes/import_wizard/html.py
shifucun: Yes, that can use "subprocess.Popen". We should definitely update to use that once there is way to check job status.

server/routes/import_wizard/html.py
shifucun: That's a good idea.

shifucun (commented):
thanks!

chejennifer (approved):

server/integration_tests/test_data/translate_hindi/query_1/chart_config.json
```diff
@@ -3,6 +3,230 @@
     "categories": [
       {
         "blocks": [
+          {
```
shifucun, line 6: @pradh when you have time, can you check why this config changes?

PR #3668, opened 2023-10-05
Updated tile spec: added subtitle, geo_json_prop (map), place_name_prop (map)

dwnoble:
* Added new "asia" dev topic to test `geoJsonProp` and `placeNameProp` at country level
* Added dev topic pages to screenshot test

Screenshots - first two use UN geojson (includes taiwan) and UN place names (Uses "Russian Federation" instead of "Russia")
## geo_json_prop
geoJsonProp

## place_name_prop
pr-place-name-prop

## subtitle
subtitles


chejennifer (approved):
nice!

PR #3671, opened 2023-10-06
[subject-page-spec] Updated tile props

dwnoble:
* Moved place_name_prop to a tile level field
* Added time_scale to line chart
* Added variable_name_regex to line and bar charts

## Screenshots

tiles2a
tiles2b


chejennifer (approved):
nice!

PR #3674, opened 2023-10-09
Add a flag for use default place in NL fulfillment

shifucun:


server/routes/nl/helpers.py
```diff
@@ -66,6 +65,8 @@ def parse_query_and_detect(request: Dict, app: str, debug_logs: Dict):
 
   test = request.args.get(params.Params.TEST.value, '')
   i18n = request.args.get(params.Params.I18N.value, '')
+  use_default_place = request.args.get(params.Params.USE_DEFAULT_PLACE.value,
```
beets, line 68: should we convert to a boolean here? then it'll be clearer below on 172 what the acceptable values are.

we should also probably call lower() before comparing.

beets (approved):
thanks bo!

shifucun (commented):
thanks for review!

PR #3675, opened 2023-10-09
[homepage] Add data size blurb

beets:
Demo here: https://serve-dot-zipline.googleplex.com/asset/d90717e2-8b4f-5ee4-9598-3f2b9483b423/


shifucun (approved):
nice! maybe in a separate PR, worth documenting how to update the sprite and use it.

beets:
> nice! maybe in a separate PR, worth documenting how to update the sprite and use it.

good point - will do in a followup

PR #3677, opened 2023-10-10
Bar Tile Prop Refactoring

juliawu:
Updates the Bar tile to use the new refactored structure in `tile_types.ts`.

Changes:
* Switched places from "NamedTypePlace" to just the dcid string as input to decouple tiles from website repo specific constructs.
* Fixed typo in `INITAL_LOADING_CLASS`
* Removed reference to `BarTileSpec` in the bar tile props which is specific to the website repo
* Updated some existence checks and field accesses in bar tile functions as required by typescript when working with union/intersect types.
* Move subtitle and svgChartHeight fields into the shared `ChartOptions` interface.

Testing: Verified changes don't break bar and ranking tiles in homepage queries, visually verified /topic/dev/geoId/06, and checked bar tiles in node server locally.


dwnoble (approved):
Nice refactor!

PR #3678, opened 2023-10-10
Added "date" field support in highlight tile for subject page config

dwnoble:


juliawu (approved):
Nice! Thanks for also adding to the dev topic page!

PR #3682, opened 2023-10-12
Add a convenience script for updating submodules

juliawu:
I always need to lookup which commands to run to update the submodules of this repo. This PR adds the commands as a script for convenience.

### DIFF
--- a/update_git_submodules.sh
+++ b/update_git_submodules.sh
@@ -1,0 +1,28 @@
+#!/bin/bash
+# Copyright 2023 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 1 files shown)


update_git_submodules.sh
```diff
@@ -0,0 +1,24 @@
...
+# limitations under the License.
+
+
+# Convenience script for updating git submodules
+# 
+# Requires: git to be installed
+#
+# Usage: ./update_git_submodules.sh from root directory
+
+git submodule foreach git pull origin master
```
juliawu, line 23: Reviewers: To confirm, should we pull from origin master or dc master? (after adding dc as a remote)

keyurva (approved):
Thanks - this will save time every time!

update_git_submodules.sh
shifucun: it should be origin. We don't set up "dc" as remote for the submodules.

shifucun (approved):
thanks!

PR #3685, opened 2023-10-12
Added web components npm package

dwnoble:
Added a basic npm package containing for the data commons web components that holds the type definitions. Web component definitions themselves are still in the website/static directory.
 
* Added npm package definition to web components
* Added `wireit` to website build to orchestrate building web components dependency
* migrated type definitions to packages/web-components
* migrated website to use donut web component instead of react tile


static/js/components/subject_page/block.tsx
```diff
@@ -404,22 +404,21 @@ function renderTiles(
           ></GaugeTile>
         );
       case "DONUT":
+        // TODO(dwnoble): Add footnote, svg chart height, and stat var spec support to this tile
         return (
-          <DonutTile
+          <datacommons-pie
```
juliawu, line 409: To confirm, is the plan to eventually migrate everything in the website repo (minus visualization tools) to web components?

We might want to keep using `DonutTile` instead of `datacommons-pie` until all the parity/refactoring work is done. This way we don't lose current footnote or chart height functionality.

packages/web-components/src/main.d.ts
```diff
@@ -15,9 +15,20 @@
...
+export type SORT_ASCENDING_POPULATION = "ascendingPopulation";
+export type SORT_DESCENDING_POPULATION = "descendingPopulation";
+export type ChartSortOption =
+  | SORT_ASCENDING
+  | SORT_DESCENDING
+  | SORT_ASCENDING_POPULATION
+  | SORT_DESCENDING_POPULATION;
 
-import { SortType } from "../js/chart/types";
+export interface ChartEventDetail {
```
juliawu, line 28: nit: can we add a quick comment about what this is for?

juliawu (commented):
Awesome work!

static/js/components/subject_page/block.tsx
dwnoble: That makes sense - reverted this to DonutTile for now

packages/web-components/src/main.d.ts
dwnoble: done!

shifucun (commented):
Can you help try to run custom DC locally following https://github.com/datacommonsorg/website/blob/master/docs/custom_dc.md#ui-updates. Want to make sure the changes are compatible.

juliawu (approved):
Thanks for the updates! LGTM, please address shifucun's comments before merging.

dwnoble:
> Can you help try to run custom DC locally following https://github.com/datacommonsorg/website/blob/master/docs/custom_dc.md#ui-updates. Want to make sure the changes are compatible.

Good catch-  i saw some build errors here and resolved them. I was able to get the local docker image running successfully after.

PR #3686, opened 2023-10-12
removed legacy sdg-static and sdg-tracker directories

dwnoble:


juliawu (approved):
Woohoo! Thanks for the cleanup.

PR #3691, opened 2023-10-13
Add screenshot tests for homepage and explore landing pages

beets:


server/webdriver/screenshot/local/page.json
```diff
@@ -134,6 +139,41 @@
     "height": 1080,
     "async": true
   },
+  {
+    "url": "/explore/economics",
+    "height": 1080,
+    "async": true
```
shifucun, line 145: async should be "false" I think for these pages.

When it's set to "true", the test would check for parent dom with class "dc-async-element-holder" to have children "dc-async-element". These classes needed to added for the component with async data fetch

beets: thanks - i've updated to false, added the viz landing pages and tested locally.
the viz tools will need some updates for the screenshot tests to work on the charting calls

shifucun (approved):

server/webdriver/screenshot/local/page.json
shifucun: Right, need to decorate those classes for viz tool.

PR #3693, opened 2023-10-13
Translate strings in page config.

keyurva:
* Previously we added support for non-English queries [(PR)](https://github.com/datacommonsorg/website/pull/3650).
* However, the response was still in English.
* This PR translates the response strings to the queried language.

[translate-page-config.webm](https://github.com/datacommonsorg/website/assets/1221814/72125a9d-aae6-4405-9b9b-686950759254)

### DIFF
--- a/server/lib/nl/common/utterance.py
+++ b/server/lib/nl/common/utterance.py
@@ -175,3 +175,7 @@ class Utterance:
   insight_ctx: Dict = field(default_factory=dict)
   # Test param.
   test: str = ''
+  # i18n param.
+  i18n: str = ''
+  # For i18n enabled requests, this field encodes the language of the query.
+  query_lang: str = ''
--- a/server/lib/nl/common/utterance.py
+++ b/server/lib/nl/common/utterance.py
@@ -175,7 +175,5 @@ class Utterance:
   insight_ctx: Dict = field(default_factory=dict)
   # Test param.
   test: str = ''
-  # i18n param.
-  i18n: str = ''
   # For i18n enabled requests, this field encodes the language of the query.
-  query_lang: str = ''
+  i18n_lang: str = ''
--- a/server/lib/translator.py
+++ b/server/lib/translator.py
@@ -20,10 +20,10 @@ from langdetect import detect as lang_detect
 import requests
 
 from server.lib.nl.common.counters import Counters
+from shared.lib.constants import EN_LANG_CODE
 
 _API_URL = "https://translation.googleapis.com/language/translate/v2"
 _API_HEADER = {"content-type": "application/json"}
-_EN_LANG = "en"
 
 
 # Detects the query language and translates non-English queries to English.
@@ -31,10 +31,10 @@ _EN_LANG = "en"
 def detect_lang_and_translate(query, counters: Counters) -> (str, str):
   try:
     lang = lang_detect(query)
-    if lang.startswith(_EN_LANG):
+    if lang.startswith(EN_LANG_CODE):
       return lang, query
 
-    translations = _translate([query], lang, _EN_LANG)
+    translations = _translate([query], lang, EN_LANG_CODE)
 
     if translations:
       counters.info(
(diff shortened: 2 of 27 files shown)


server/lib/translator.py
```diff
@@ -21,42 +22,168 @@
...
+
+
+class TranslationStringTokens:
+  """Class that replaces tokens (like "${date}") in strings 
+with replacements that won't be translated (like "___") to send for translation.
+
+It also supports a method to reinsert tokens back into the translated string.
+"""
+
+  _TOKEN_PATTERN = r"\${[^}]+}"
```
pradh, line 65: comment what the pattern is?

server/lib/translator.py
```diff
@@ -21,42 +22,168 @@
...
+  def reinsert_tokens(self, translation):
+    for token in self.tokens:
+      translation = re.sub(TranslationStringTokens._TOKEN_REPLACEMENT,
+                           token,
+                           translation,
+                           count=1)
+    return translation
+
+
+# Translates display strings in the provided page config and
```
pradh, line 87: There's a lot of logic here that could use some unit tests

server/lib/translator.py
```diff
@@ -21,42 +22,168 @@
...
+    for block in category.get("blocks", []):
+      _maybe_add(block.get("title"))
+      for column in block.get("columns", []):
+        for tile in column.get("tiles", []):
+          _maybe_add(tile.get("description"))
+          _maybe_add(tile.get("title"))
+    for sv in category.get("statVarSpec", {}).values():
+      _maybe_add(sv.get("name"))
+
+  return list(strings)
```
pradh, line 166: I wonder if the logic would be simpler (without the need for `TranslationStringTokens`), if we just stripped the `${date}`-like stuff in the returned strings (in _maybe_add), and again when checking the map (in _maybe_update)?

server/lib/nl/common/utterance.py
```diff
@@ -174,4 +174,8 @@ class Utterance:
   # Past complete context for insight flow.
   insight_ctx: Dict = field(default_factory=dict)
   # Test param.
-  test: str = ''
\ No newline at end of file
+  test: str = ''
+  # i18n param.
```
pradh, line 178: Wonder if we should have just 1 field `i18n_lang`?

server/routes/explore/api.py
```diff
@@ -160,6 +160,7 @@ def _fulfill_with_insight_ctx(request: Dict, debug_logs: Dict,
                               counters: ctr.Counters) -> Dict:
   insight_ctx = request.get_json()
   test = request.args.get(Params.TEST.value, '')
+  i18n = request.args.get(Params.I18N.value, '')
```
pradh, line 163: This flow is for `/fulfill`, and it never sets the lang.  So I think if we want to support i18n for this flow, we should pass in `i18n_lang` as param.  In this PR, maybe drop changes to this file, and have a TODO to support i18n for `/fulfill` API?

server/routes/nl/helpers.py
```diff
@@ -47,11 +47,15 @@
...
 from server.lib.translator import detect_lang_and_translate
+from server.lib.translator import translate_page_config
 from server.lib.util import get_nl_disaster_config
 from server.routes.nl import helpers
 import server.services.bigtable as bt
 import shared.lib.utils as shared_utils
 
+# English language code.
+_EN_LANG = "en"
+
```
pradh, line 58: move it to constants.py?

pradh (commented):
Nicely done!

server/lib/translator.py
keyurva: Done.

server/lib/translator.py
keyurva: Done.

server/lib/nl/common/utterance.py
keyurva: Done.

server/routes/explore/api.py
keyurva: Done.

server/routes/nl/helpers.py
keyurva: Done.

pradh (approved):
Thanks for the tests!

PR #3702, opened 2023-10-17
NL: Fix Earth title bug and don't log test queries

pradh:
This fixes the issue where `Earth` would show up as the title instead of World.

Also, avoid logging test queries.


shifucun (approved):
nice clean up!

chejennifer (approved):

pradh:
Thanks!

PR #3705, opened 2023-10-17
fixed nodejs docker image build error

dwnoble:
* Added docker image dependencies to allow building on M1 mac
* Updated build logic to copy over and build entire web components folder before installing dependencies in `website/static`


shifucun (approved):
thanks!

PR #3709, opened 2023-10-18
Update map tool geo json ready hook

shifucun:
Current logic treat geoJson is "ready" even though the data features is empty. This poses an issue when the geoJson is computed for 1deg grid. The render function is entered once, but as geoJson is not ready, it's unmounted, making the later render fail. Not clear why this works for the ipcc_grid, but local testing shows the fix works for existing charts.


chejennifer (approved):
nice

PR #3711, opened 2023-10-20
[nl] pass detected date range to PopulateState and ChartSpec

chejennifer:
- detect date from query and pass the detected date range to PopulateState and ChartSpec
- add date to debug info

note: Changes for actually using the date in existence checks and chart configs will come in future PRs


server/integration_tests/test_data/detection_api_sdg/chart_config.json
```diff
@@ -13,17 +14,5 @@
...
-    "dc/topic/sdg_3.c.1",
-    "dc/topic/sdg_3.b.3",
-    "dc/topic/sdg_3.d",
-    "dc/topic/sdg_3.4.1",
-    "dc/topic/sdg_3.c",
-    "dc/topic/sdg_3.4",
-    "dc/topic/sdg_3.b.2"
-  ]
+  "variables": []
 }
```
pradh, line 18: why this diff?

server/integration_tests/test_data/e2e_sdg/compareprogressonpovertyinmexico,nigeriaandpakistan/chart_config.json
```diff
@@ -8,534 +8,13 @@
               {
                 "tiles": [
                   {
-                    "barTileSpec": {
-                      "maxPlaces": 15,
-                      "maxVariables": 15,
-                      "sort": "DESCENDING"
-                    },
-                    "comparisonPlaces": [
-                      "country/MEX",
```
pradh, line 17: expected?

server/lib/nl/common/serialize.py
```diff
@@ -106,7 +108,8 @@ def classification_to_dict(classifications: List[NLClassifier]) -> List[Dict]:
       cdict['correlation'] = True
     elif isinstance(c.attributes, QuantityClassificationAttributes):
       cdict['quantity'] = _quantity_to_dict(c.attributes)
-
+    elif isinstance(c.attributes, DateClassificationAttributes):
+      cdict['dates'] = c.attributes.dates
```
pradh, line 112: Do we need to turn dataclass into a dict?

pradh (commented):
Nice!  Mainly wondering about the diffs...

server/integration_tests/test_data/detection_api_sdg/chart_config.json
chejennifer: caused by updating goldens without clearing embeddings cache first

server/integration_tests/test_data/e2e_sdg/compareprogressonpovertyinmexico,nigeriaandpakistan/chart_config.json
chejennifer: caused by updating goldens without clearing embeddings cache first

server/lib/nl/common/serialize.py
chejennifer: done

server/integration_tests/test_data/translate_hindi/query_1/chart_config.json
```diff
@@ -3,230 +3,6 @@
     "categories": [
       {
         "blocks": [
-          {
```
chejennifer, line 6: @keyurva these seem related to translation, do you have any ideas what could be causing these?

keyurva: Thanks for updating the goldens. Unfortunately the translations returned by the api can vary and the goldens will need to be updated when they do.

pradh: Is the failure non-determinism vs. some sporadic model change? (if the former, we should consider checking for select keywords instead of full strings)

keyurva: I don't think it'll be sporadic. I've filed b/307967463 to address it.

pradh (approved):
Thanks!

chejennifer:
thanks for the review!

PR #3715, opened 2023-10-24
Fix Test Failures

chejennifer:
- the nl server uses the package `datasets` which uses the package `fsspec`. The most recent version of `fsspec` causes problems with `datasets` as mentioned here: https://github.com/huggingface/datasets/issues/6333


shifucun (approved):
nice find!

PR #3716, opened 2023-10-24
Change old admin page (based on oauth and multi-tenent model) to RSI model with data load support

shifucun:
On server start, can set an environment variable "ADMIN_SECRET". When user perform actions in this admin page, they need to input the secret. This is a naive way for authentication before we have a full solution.

image


server/routes/admin/html.py
```diff
@@ -0,0 +1,83 @@
...
+          "status": "success",
+          "stdout": result.stdout.strip().splitlines()
+      })
+    except subprocess.CalledProcessError as cpe:
+      return jsonify({
+          "status": "failure",
+          "error": cpe.stderr.strip().splitlines()
+      }), 500
+    except Exception as e:
+      return jsonify({"status": "error", "message": str(e)}), 500
```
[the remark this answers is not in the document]
juliawu: Perhaps we can log the error so we can debug internally and instead return a more generic message to the user?

juliawu (approved):
Nice! Just some quick thoughts on the alerts.

server/routes/admin/html.py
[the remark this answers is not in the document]
shifucun: The log here is for the custom DC dev uses, so the errors are desired. 

Moving forward, we should put more stuff in a status.json file for the page rendering purpose.

server/routes/admin/html.py
```diff
@@ -0,0 +1,83 @@
...
+      "-X",
+      "POST",
+      "localhost:8081/import",
+      "-d",
+      f'{{"data_path": "{output_dir}"}}',
+  ]
+  output = []
+  for command, cwd in [(command1, "import/simple"), (command2, ".")]:
+    try:
+      result = subprocess.run(command,
```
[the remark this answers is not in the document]
shifucun: Added some check for command, but this warning would still persist.

shifucun (commented):
thanks for review!

import
chejennifer, line 1: did you mean to update this and the import submodules?

static/js/admin/app.ts
chejennifer, line 1: nit: maybe make this `main.ts` to be consistent with the other pages where `main.ts` is the file with the "window.onload"

static/js/admin/page.tsx
```diff
@@ -0,0 +1,78 @@
...
+ * Admin Page.
+ */
+
+import axios from "axios";
+import _ from "lodash";
+import React, { useState } from "react";
+import { Button } from "reactstrap";
+
+export function Page(): JSX.Element {
+  const [loading, setLoading] = useState(Boolean);
```
chejennifer, line 27: is this a new syntax for `useState`? Because shouldn't the state type be declared like `useState<boolean>()`

chejennifer (commented):
cool!

nl_server/requirements.txt
```diff
@@ -1,20 +1,20 @@
 datasets==2.12.0
```
shifucun, line 1: sorted the packages

server/requirements.txt
```diff
@@ -1,25 +1,26 @@
 absl-py==1.4.0
```
shifucun, line 1: sorted the pip packages here

static/js/admin/page.tsx
shifucun: Good catch, there were totally off!

PR #3725, opened 2023-10-26
Refactor NL Server bootstrap code for readability

pradh:
This is to make follow on custom DC specific changes cleaner.

There is no intentional logic change other than using a single cache entry (instead of one per index) in test environment.

A key simplification is to more directly rely on the embeddings name to extract the base / tuned-model (see `config._parse()`) instead of the current code which uses `models.yaml` and also relies on the naming partially.

### DIFF
--- a/nl_server/__init__.py
+++ b/nl_server/__init__.py
@@ -49,6 +49,11 @@ def create_app():
   return app
 
 
+#
+# On prod the yaml files are in /datacommons/nl/, whereas
+# in test-like environments it is the checked in path
+# (deploy/nl/).
+#
 def get_env_path(file_name: str) -> str:
   flask_env = os.environ.get('FLASK_ENV')
   if flask_env in ['local', 'test', 'integration_test', 'webdriver']:
--- a/nl_server/__init__.py
+++ b/nl_server/__init__.py
@@ -12,7 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import logging
 import os
 import sys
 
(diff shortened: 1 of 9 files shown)


nl_server/config.py
```diff
@@ -0,0 +1,102 @@
...
+
+
+def _parse(embeddings_map: Dict[str, str]) -> List[Index]:
+  indexes: List[Index] = []
+
+  for key, value in embeddings_map.items():
+    parts = value.split('.')
+    idx = Index(name=key, embeddings_file_name=value)
+    if len(parts) == 4:
+      # Expect: <embeddings_version>.<fine-tuned-model-version>.<base-model>.csv
```
keyurva, line 91: I see a blob like `embeddings_medium_2023_10_24_10_44_32.ft_final_v20230717230459.all-MiniLM-L6-v2.csv` in GCS. Does this parse such values or a different one?

It will help if you can add a comment with a concrete example for those not familiar with the format.

keyurva:
It'd be helpful if you can add a short README for the role the 2 yaml files play in this folder: https://github.com/datacommonsorg/website/tree/master/deploy/nl

(Ignore if it's already documented somewhere)

keyurva (approved):
Thanks for the cleanup.

nl_server/__init__.py
```diff
@@ -12,62 +12,48 @@
...
-    embeddings_map = yaml.full_load(f)
-    if not embeddings_map:
-      logging.error("No configuration found for embeddings")
-      return
 
-    app.config['EMBEDDINGS_VERSION_MAP'] = embeddings_map
-    loader.load_embeddings(app, embeddings_map, models_downloaded_paths)
+def get_env_path(file_name: str) -> str:
+  flask_env = os.environ.get('FLASK_ENV')
+  if flask_env in ['local', 'test', 'integration_test', 'webdriver']:
```
shifucun, line 54: Add a comment on why different env needs to read different files?

nl_server/config.py
```diff
@@ -0,0 +1,102 @@
...
+      assert parts[
+          2] == EMBEDDINGS_BASE_MODEL_NAME, f'Unexpected base model {parts[3]}'
+      idx.tuned_model = f'{parts[1]}.{parts[2]}'
+    else:
+      # Expect: <embeddings_version>.csv
+      assert len(parts) == 2, f'Unexpected file name format {value}'
+    assert parts[
```
shifucun, line 100: nit: do this assert at the very beginning.

nl_server/config.py
```diff
@@ -0,0 +1,102 @@
...
+
+# App Config constants.
+NL_MODEL_KEY = 'NL_MODEL'
+NL_EMBEDDINGS_KEY = 'NL_EMBEDDINGS'
+NL_EMBEDDINGS_VERSION_KEY = 'NL_EMBEDDINGS_VERSION_MAP'
+
+
+# Defines one embeddings index config.
+@dataclass
+class Index:
```
shifucun, line 36: Maybe "EmbeddingIndex"?

nl_server/config.py
pradh: Yes, added example.

nl_server/__init__.py
pradh: Done

nl_server/config.py
pradh: Done

nl_server/config.py
pradh: Done

pradh (commented):
Thanks for the review both!

pradh:
> It'd be helpful if you can add a short README for the role the 2 yaml files play in this folder: https://github.com/datacommonsorg/website/tree/master/deploy/nl
> 
> (Ignore if it's already documented somewhere)

I'll add it to https://github.com/datacommonsorg/website/blob/master/tools/nl/embeddings/README.md in a follow on PR

shifucun (approved):

PR #3729, opened 2023-10-26
Merging for Custom DC Embeddings

pradh:
This PR allows for queries spanning Custom DC and Main DC embeddings.

Specifically, when `embeddings.yaml` contains a `custom` key, we merge those embeddings into `medium_ft` embeddings on NL server boot.

Bonus:
* Docs on Embeddings Config files
* Modify the NL Server param from `sz` to `idx` (only website and a test uses it now).


nl_server/loader.py
```diff
@@ -46,18 +52,66 @@ def load_server_state(app: Any, embeddings_map: Dict[str, str],
...
+
+  return embeddings_map, models_map
+
+
+def _maybe_load_custom_dc_yaml():
+  base = os.environ.get('SQL_DATA_PATH')
+  if not base:
+    return None
+
+  file_path = os.path.join(base, f'data/nl/{_CUSTOM_EMBEDDINGS_YAML}')
```
keyurva, line 94: nit: Add a comment referencing the custom dc loader: https://github.com/datacommonsorg/website/blob/5ff56b7cdabb0c6c79e0c6c90bee8280c398ac2d/server/routes/admin/html.py#L40

nl_server/routes.py
```diff
@@ -92,3 +93,9 @@ def search_verbs():
 @bp.route('/api/embeddings_version_map/', methods=['GET'])
 def embeddings_version_map():
   return json.dumps(current_app.config[config.NL_EMBEDDINGS_VERSION_KEY])
+
+
+@bp.route('/api/load/', methods=['GET'])
+def load():
+  loader.load_server_state(current_app)
```
keyurva, line 100: This is beautiful.

nl_server/tests/custom_embeddings_test.py
```diff
@@ -0,0 +1,105 @@
...
+    cls.custom = store.Store(
+        config.load({
+            'medium_ft': default_file,
+            'custom_ft': custom_file,
+        }, {'tuned_model': _TUNED_MODEL}))
+
+  def test_entries(self):
+    self.assertEqual(1, len(self.main.get('medium_ft').dcids))
+    self.assertTrue(not self.main.get('custom_ft'))
+    self.assertEqual(2, len(self.custom.get('medium_ft').dcids))
```
keyurva, line 67: Is this 2 because it is default + custom concatenated?

keyurva (approved):
This is amazing!

I actually can't decide which one is more beautiful - the `/api/load` implementation or the tests :)

nl_server/loader.py
shifucun, line 94: Maybe add an optional arg for load_server_state(), which is the folder path, since "data" can become dynamic later.

nl_server/routes.py
```diff
@@ -92,3 +93,9 @@ def search_verbs():
 @bp.route('/api/embeddings_version_map/', methods=['GET'])
 def embeddings_version_map():
   return json.dumps(current_app.config[config.NL_EMBEDDINGS_VERSION_KEY])
+
+
+@bp.route('/api/load/', methods=['GET'])
+def load():
```
shifucun, line 99: also nice to have another param for the custom embedding data path.

shifucun (approved):

nl_server/loader.py
pradh: hmm, to confirm, when "data" becomes dynamic, on subsequent boot-up (say server restarts), how will that dynamic path be available?

pradh: So it almost feels like we should only rely on an env path `CUSTOM_DC_NL_BASE` (which is made of `SQL_DATA_PATH`/data)?

That way, on restart, we can set that env to the new dynamic path.

nl_server/tests/custom_embeddings_test.py
pradh: Yes, left a comment.

nl_server/loader.py
shifucun: I was thinking the "version" is stored in "SQL_DATA_PATH/version.txt" which can be retrieved, but RSI need to produce that first.

pradh: That makes sense.   Should we just change loader.py to always read "SQL_DATA_PATH/version.txt" when we have that ready?

So the load flow is same as boot up flow (no override arg for `load/)

pradh: In any case, added a TODO in this PR for follow up.

pradh (commented):
Thanks for the review guys! :)

nl_server/routes.py
pradh: I added a TODO in the other file based on what we decide about always readying "version.txt".  Based on that, will add in follow on PR.

nl_server/loader.py
pradh: Done.

keyurva: > I was thinking the "version" is stored in "SQL_DATA_PATH/version.txt" which can be retrieved, but RSI need to produce that first.

That's a good idea.

@shifucun - Can you create a bug for this to record what exactly we want in this file? Also, think about whether a yaml or json might work better for future extensibility.

PR #3732, opened 2023-10-27
Optimize  website compose pip package install; Sync submodules

shifucun:
Transformer package depends on torch, which is a few G due to the pulling of nvida library. For us, we only need the cpu version of that. So doing the following:

```
RUN pip install --upgrade pip setuptools light-the-torch
RUN ltt install torch --cpuonly
```

This PR also sorted requirements.txt files and upgrade a few packages to be aligned with those are used in "import" repo


shifucun:
Thanks!

pradh:
To confirm, how do we validate this?  Is a autopush successful load sufficient?

PR #3736, opened 2023-10-27
Added support for web components in screenshot test

dwnoble:
Updated `charts_rendered` helper to wait for the web components as well as charts. The original check wasn't working for web components because selenium's element selectors can't reach into the web component shadow root

### DIFF
--- a/server/webdriver/shared.py
+++ b/server/webdriver/shared.py
@@ -23,7 +23,8 @@ MAX_NUM_SPINNERS = 3
 ASYNC_ELEMENT_HOLDER_CLASS = 'dc-async-element-holder'
 ASYNC_ELEMENT_CLASS = 'dc-async-element'
 TIMEOUT = 60
-# Keep in sync with packages/web-components/src/main.ts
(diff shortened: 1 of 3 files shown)


server/webdriver/shared.py
```diff
@@ -23,6 +23,11 @@
 ASYNC_ELEMENT_HOLDER_CLASS = 'dc-async-element-holder'
 ASYNC_ELEMENT_CLASS = 'dc-async-element'
 TIMEOUT = 60
+WEB_COMPONENT_TAG_NAMES = [
```
shifucun, line 26: Add a comment in here and the components definition page to sync both places?

shifucun (approved):
Thanks! Can you run ./run_test.sh -s locally to verify?

server/webdriver/shared.py
dwnoble: done!

dwnoble:
> Thanks! Can you run ./run_test.sh -s locally to verify?

Verified the tests pass locally

PR #3737, opened 2023-10-27
A set of custom DC improvements

shifucun:
* Enable Admin page via environment variables. This way, /admin shows up only on demand (and in RSI custom DC by default).
* Set default svg_blocklist which custom DC can use
* Handle when sql_data_path is not set.


server/__init__.py
```diff
@@ -379,6 +380,8 @@ def create_app(nl_root=DEFAULT_NL_ROOT):
   if os.path.isfile(BLOCKLIST_SVG_FILE):
     with open(BLOCKLIST_SVG_FILE) as f:
       blocklist_svg = json.load(f) or []
+  else:
+    blocklist_svg = ["dc/g/Uncategorized", "oecd/g/OECD"]
```
juliawu, line 384: Just want to make sure, do we want Uncategorized and OECD blocked by default?

juliawu (approved):
Nice!

server/__init__.py
shifucun: Since we don't show them in prod, we should not show them in custom DC, which is this line covers.

PR #3743, opened 2023-10-30
Fix Lint Tests

juliawu:
The `website-pull-request-npm` can pass even when lint tests fail, causing some badly formatted code to be successfully merged. This PR fixes the test, and corrects the files with lint issues.

Root cause was that `website-pull-request-npm` did not exit when the lint test fails and instead silently proceeded, passing if client tests passed. Added `set -e` so that bash exits on a test failure.


shifucun (approved):
thanks for fixing!

dwnoble (approved):

pradh (approved):
Thanks for the fix!

PR #3745, opened 2023-10-30
add sdg chart titles

chejennifer:
- add sdg chart titles to the cached chart titles
- updates chart titles to include a list of blocklisted dcs
    - this is needed because we want curated titles for sdg svs on the main dc because otherwise titles on the explore page will have things like '2. No Hunger' which is confusing when there's no context into what the number is for. On the other hand, for the sdg static site, the UN wants all the sdg goal/target/indicator numbers and the exact original names.


tools/nl/titles/gen_chart_titles_by_sv.py
```diff
@@ -5,27 +5,39 @@
 import pandas as pd
 
 SHEET_URL = 'https://docs.google.com/spreadsheets/d/1lmNAnqECpcvkuOlIkdo50Ve1KAalOoyP_lUlOuLmIAU'
-SHEET_NAME = 'Main'
-CSVFILE = 'demo_svs_1.3k_names.csv'
-JSONFILE = '../../../server/config/nl_page/chart_titles_by_sv.json'
+SHEET_LIST = [{
```
pradh, line 8: Update README.md in this directory to mention there is an additional SDG-specific sheet?

server/lib/topic_cache.py
```diff
@@ -121,8 +122,9 @@ def load_files(fpath_list: List[str], name_overrides: Dict) -> TopicCache:
     dcid = node['dcid'][0]
     typ = node['typeOf'][0]
     name = node.get('name', [''])[0]
-    if name_overrides.get(dcid):
-      name = name_overrides.get(dcid)
+    name_override_info = name_overrides.get(dcid)
+    if name_override_info and not dc in name_override_info['blocklist']:
```
pradh, line 126: nit:  be defensive name_override_info.get('blocklist', []), and below

server/integration_tests/test_data/sdg/query_1/chart_config.json
```diff
@@ -11,14 +11,14 @@
                     "statVarKey": [
                       "sdg/SI_POV_DAY1"
                     ],
-                    "title": "Proportion of Population Below International Poverty Line in Countries of Africa (${date})",
+                    "title": "Population Below International Poverty Line in Countries of Africa (${date})",
```
pradh, line 14: For one of these with SDG response, can we add another mode that sets `dc=sdg` as validation that the prior titles are unaffected?

pradh (commented):
Nice!

Can you expand the description with the motivation of this change?  Specifically, call out how this behavior is different on UN SDG DC vs. main DC.

tools/nl/titles/gen_chart_titles_by_sv.py
chejennifer: done

server/lib/topic_cache.py
chejennifer: done

server/integration_tests/test_data/sdg/query_1/chart_config.json
chejennifer: the sdg tests in explore_test.py actually already set dc=sdg, but I've added an sdg test that doesn't set dc there!

pradh:
> Nice!
> 
> Can you expand the description with the motivation of this change? Specifically, call out how this behavior is different on UN SDG DC vs. main DC.

in case you missed this :)

server/integration_tests/explore_test.py
```diff
@@ -370,6 +370,11 @@ def test_e2e_sdg(self):
     ],
                                 dc='sdg')
 
+    self.run_detect_and_fulfill('e2e_sdg_mainDc', [
```
pradh, line 373: super nit: ok for this, but not a fan of mixing snake and camel :)

tools/nl/titles/README.md
```diff
@@ -2,8 +2,11 @@ This directory is for storing and processing curated stat-var title spreadsheets
 
 State as of March 2023:
 
-* The spreadsheet with 1.2K SVs is [HERE](https://docs.google.com/spreadsheets/d/1lmNAnqECpcvkuOlIkdo50Ve1KAalOoyP_lUlOuLmIAU/edit#gid=599439456),
-  and checked in as [`demo_svs_1.3k_names.csv`](demo_svs_1.3k_names.csv).
+* The spreadsheet with the information to generate chart titles is [HERE](https://docs.google.com/spreadsheets/d/1lmNAnqECpcvkuOlIkdo50Ve1KAalOoyP_lUlOuLmIAU/edit#gid=599439456) and it has 2 sheets that are used:
+
+   * __Main__: This is the sheet for stat var titles that are used in all DCs & is checked in as [`demo_svs_1.3k_names.csv`](demo_svs_1.3k_names.csv).
+
+   * __Main_SDG__: This is the sheet for stat var titles for SDG topics and variables that are used in the main DC, but not in sdg or sdgmini DCs. This is checked in as [`sdg_svs_names.csv`](sdg_svs_names.csv).
```
pradh, line 9: "because for the [UN SDG Frontend] (https://unstats.un.org/UNSDWebsite/undatacommons/sdgs) requires original SDG strings"

pradh (approved):
Thanks for the fixes!

chejennifer:
> > Nice!
> > Can you expand the description with the motivation of this change? Specifically, call out how this behavior is different on UN SDG DC vs. main DC.
> 
> in case you missed this :)

oh yes, totally missed that, updated now!

server/integration_tests/explore_test.py
chejennifer: done

tools/nl/titles/README.md
chejennifer: done

chejennifer:
thanks for the review!

PR #3749, opened 2023-10-31
trim chart config footnotes

chejennifer:
- trim chart config footnotes so that it's easier to look at golden diffs


pradh (approved):
Thanks!

PR #3752, opened 2023-10-31
Bug quick fix: Fix per-capita toggle for US counties scatter plots

juliawu:
**Problem:** The "See per-capita" toggle fails on some scatter plots because the scatter tile sends out a duplicate API call for "Count_Person", one for each axis, which can overwhelm mixer when making concurrent fetches because there are too many places to fetch data for.

**Fix:** Remove duplicate denom dcids in the data fetch.

**Testing:** Verified changes locally on explore pages and scatter plot explorer.

| Per Capita Off | Per Capita On |
--- | --- |
| Screenshot 2023-10-31 at 11 20 35 AM |  Screenshot 2023-10-31 at 11 29 00 AM |

### DIFF
--- a/static/js/components/tiles/scatter_tile.tsx
+++ b/static/js/components/tiles/scatter_tile.tsx
@@ -230,16 +230,21 @@ function getPopulationPromise(
   statVarSpec: StatVarSpec[],
   apiRoot?: string
 ): Promise<SeriesApiResponse> {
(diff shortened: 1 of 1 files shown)


chejennifer (approved):
nice, thanks for the quick fix!

pradh (approved):
yay, thanks for the quick fix!

PR #3757, opened 2023-11-02
Add NL server to docker image

keyurva:
* Adds NL server to docker image.
* Currently always starts NL server for custom DC. Will help with adoption and awareness. Can make it optional if needed in a separate PR.
* Uses an anonymous client to download models so customers can use it without a GCP account.
* Banned words are in a private bucket, so currently not loading them for custom DCs.
* Added `MIXER_API_KEY` back since Mixer still uses it (vs `DC_API_KEY`). Can remove in a separate PR.
* Added `devcontainer.json` to enable dev containers in VS code when needed.
* NEXT: Running custom DC NL loader in `/load-data` and serving NL queries for custom SVs.

Quick demo of starting a docker container with NL server:

[nl-in-docker.webm](https://github.com/datacommonsorg/website/assets/1221814/17f9f509-0208-4b63-887c-da173d965963)


shared/lib/gcs.py
```diff
@@ -23,11 +23,14 @@
 # Downloads the `filename` from GCS to TEMP_DIR
 # and returns its path.
 #
-def download_file(bucket: str, filename: str) -> str:
-  storage_client = storage.Client()
+def download_file(bucket: str,
+                  filename: str,
+                  use_anonymous_client: bool = False) -> str:
+  storage_client = storage.Client.create_anonymous_client(
+  ) if use_anonymous_client else storage.Client()
```
pradh, line 30: For readability:

```
if use_anonymous_client:
  storage_client = X
else:
  storage_client = Y
```

nl_server/gcs.py
```diff
@@ -69,7 +71,7 @@ def download_model_from_gcs(gcs_bucket: Any, local_dir: str,
 # Downloads the `model_folder` from GCS to /tmp/
 # and return its path.
 def download_model_folder(model_folder: str) -> str:
-  sc = storage.Client()
+  sc = storage.Client.create_anonymous_client()
```
pradh, line 74: Comment why so?  Also this gets called from multiple places

nl_server/gcs.py
```diff
@@ -26,7 +26,9 @@
 
 
 def download_embeddings(embeddings_filename: str) -> str:
-  return gcs_lib.download_file(bucket=BUCKET, filename=embeddings_filename)
+  return gcs_lib.download_file(bucket=BUCKET,
+                               filename=embeddings_filename,
+                               use_anonymous_client=True)
```
pradh, line 31: Comment why use_anonymous_client=True?

nl_server/loader.py
```diff
@@ -110,7 +110,7 @@ def _maybe_load_custom_dc_yaml():
 # (deploy/nl/).
```
pradh, line 110: nit: ... in "custom DC" and test-like ...

server/lib/nl/common/bad_words.py
```diff
@@ -69,8 +69,14 @@ class BannedWords:
...
 # Loads a list of bad words from a text file.
+# If the load argument is False (True by default),
+# then an empty BannedWords object will be returned.
+# It will only be True for Custom DCs at this time.
 #
-def load_bad_words() -> BannedWords:
+def load_bad_words(load: bool = True) -> BannedWords:
+  if not load:
+    return BannedWords(entries={})
+
```
pradh, line 79: Instead of this additional bool arg, maybe we can create an empty entry explicitly if CUSTOM DC?

Also, in `is_safe`, return True if `not bad_words.entries`?

server/__init__.py
```diff
@@ -371,7 +371,7 @@ def create_app(nl_root=DEFAULT_NL_ROOT):
         secret_response = secret_client.access_secret_version(name=secret_name)
         app.config['PALM_API_KEY'] = secret_response.payload.data.decode(
             'UTF-8')
-    app.config['NL_BAD_WORDS'] = bad_words.load_bad_words()
+    app.config['NL_BAD_WORDS'] = bad_words.load_bad_words(load=not cfg.CUSTOM)
```
pradh, line 374: Should we be consistent with use of `flask_env == 'custom'` vs. `cf.CUSTOM`?

pradh (commented):
Very nice!!   Looked at just the `.py` files.

build/web_compose/Dockerfile
```diff
@@ -56,14 +58,20 @@ COPY web_app.py /workspace/web_app.py
 COPY server/requirements.txt /workspace/server/requirements.txt
```
shifucun, line 58: Nit: change line 56 to "Web Server"

build/web_compose/Dockerfile
```diff
@@ -56,14 +58,20 @@ COPY web_app.py /workspace/web_app.py
...
+COPY nl_app.py /workspace/nl_app.py
+COPY nl_server/requirements.txt /workspace/nl_server/requirements.txt
+RUN pip install --upgrade pip setuptools light-the-torch
+RUN ltt install torch --cpuonly
+RUN pip3 install -r /workspace/nl_server/requirements.txt
 
 # Import Tool
 COPY import/simple/requirements.txt /workspace/import/simple/requirements.txt
 RUN pip install --upgrade pip setuptools light-the-torch
 RUN ltt install torch --cpuonly
```
shifucun, line 73: can remove line 72,73 now

build/web_compose/Dockerfile
```diff
@@ -105,5 +118,5 @@ ENV DC_API_ROOT=https://api.datacommons.org
 ENV ENABLE_ADMIN=true
 
 COPY build/web_compose/compose.sh .
-EXPOSE 8080 8081
+EXPOSE 6060 8080 8081
```
shifucun, line 121: actually should only expose 8080 now?

mixer and nl server should be internal servers now

custom_dc/cloudsql_env.list
```diff
@@ -1,7 +1,10 @@
 FLASK_ENV=custom
 USE_CLOUDSQL=true
+ENABLE_MODEL=true
 
+DC_API_ROOT=https://api.datacommons.org
```
shifucun, line 5: no need to set DC_API_ROOT as it has a default value in build/web_compose/compose.sh

custom_dc/sqlite_env.list
```diff
@@ -1,8 +1,11 @@
 FLASK_ENV=custom
 USE_SQLITE=true
+ENABLE_MODEL=true
 SQL_DATA_PATH=/sqlite
 
+DC_API_ROOT=https://api.datacommons.org
 DC_API_KEY=
+MIXER_API_KEY=
```
shifucun, line 8: ditto

nl_server/loader.py
```diff
@@ -110,7 +110,7 @@ def _maybe_load_custom_dc_yaml():
...
 #
 def get_env_path(flask_env: str, file_name: str) -> str:
-  if flask_env in ['local', 'test', 'integration_test', 'webdriver']:
+  if flask_env in ['local', 'test', 'integration_test', 'webdriver', 'custom']:
```
shifucun, line 113: have a helper function for this check?

server/__init__.py
shifucun: We should have some one line boolean helper function for such checks?

server/app_env/custom.py
```diff
@@ -23,6 +23,8 @@ class Config(_base.Config):
   LOGO_PATH = "/custom_dc/custom/logo.png"
   SHOW_DISASTER = False
   LITE = True
+  USE_PALM = False
+  USE_MEMCACHE = False
```
shifucun, line 27: No need to set since these are default?

build/web_compose/Dockerfile
keyurva: Expose only documents that the container uses these ports. They are published via the `-p` command line argument.

custom_dc/sqlite_env.list
keyurva: Ditto.

custom_dc/cloudsql_env.list
keyurva: I don't see a default value in [compose.sh](https://github.com/datacommonsorg/website/blob/master/build/web_compose/compose.sh).

build/web_compose/Dockerfile
shifucun: Oh, you are right.

custom_dc/cloudsql_env.list
shifucun: sorry, the DOCKERFILE https://github.com/datacommonsorg/website/blob/master/build/web_compose/Dockerfile#L101-L105

server/__init__.py
keyurva: I went by how it was used elsewhere in the file.

Would like to understand the difference though. As I understand it, `flask_env` is read from the env whereas cfg.CUSTOM is a `bool` in the config. What happens if both are different?

server/lib/nl/common/bad_words.py
keyurva: > Also, in is_safe, return True if not bad_words.entries?

Done.

> create an empty entry explicitly

Want to keep it encapsulated in this method, in case we want to load them differently?

server/app_env/custom.py
keyurva: Perhaps good to be explicit about it in case the defaults change?

build/web_compose/Dockerfile
keyurva: Done.

nl_server/gcs.py
keyurva: Done.

nl_server/gcs.py
keyurva: Done.

nl_server/loader.py
keyurva: Done.

shared/lib/gcs.py
keyurva: Done.

build/web_compose/Dockerfile
keyurva: Done.

custom_dc/cloudsql_env.list
keyurva: Done.

nl_server/loader.py
keyurva: Done.

server/__init__.py
shifucun: flask_env could be "floret", "stanford" or "custom" but they all have cfg.CUSTOM set. I guess here we are really looking at a new cfg field like "BLOCK_BAD_WORDS"..?

server/app_env/custom.py
shifucun: sg

server/__init__.py
keyurva: Ah ok. In that case, we are currently not loading banned words for all custom DCs so cfg.CUSTOM is appropriate.

However, we'll need to change the logic in the NL loader.

The `get_env_path` method there uses the env variable. However, we need it to pick the `deploy/nl` path for all custom DCs.

@pradh - do we have access to the config in the loader?

shifucun (approved):

server/__init__.py
pradh: Oh I think the `cfg` thing is a website server thing, not in NL server I think.

server/lib/nl/common/bad_words.py
pradh: > > create an empty entry explicitly
> 
> Want to keep it encapsulated in this method, in case we want to load them differently?

nit: It just seems odd to call load_bad_words(load=False)?  Feels slightly better to leave it empty, or call `empty_bad_words()` in the caller.

keyurva: Ok. Done.

server/__init__.py
keyurva: Hmm. In nl server, will `flask_env` be `custom` for all custom DCs or will it be `floret`, `stanford`, etc?

shifucun: flask_env could be floret, stanford, this env is global for all the servers

keyurva: In that case, should we flip the logic like below?

```
def get_env_path(flask_env: str, file_name: str) -> str:
  if flask_env in ['autopush', 'staging', 'production']:
    return return f'/datacommons/nl/{file_name}'

  return os.path.join(
        os.path.dirname(os.path.dirname(os.path.abspath(__file__))),
        f'deploy/nl/{file_name}')
```

keyurva: I'm closing this conversation and opening another one in the affected file.

nl_server/loader.py
```diff
@@ -105,18 +107,17 @@ def _maybe_load_custom_dc_yaml():
 
 
 #
-# On prod the yaml files are in /datacommons/nl/, whereas
-# in test-like environments it is the checked in path
+# On GCP, the yaml files are in /datacommons/nl/, whereas
+# in custom DC and test-like environments it is the checked in path
 # (deploy/nl/).
 #
 def get_env_path(flask_env: str, file_name: str) -> str:
```
keyurva, line 114: @pradh @shifucun - The change we were discussing in the other thread. Please review the updated logic in here.

shifucun: I think we should keep the original logic and, for RSI, mount the folder to /datacommons/nl/.

Today, we have "climate_trace" and other instances that have NL and also need embedding file in /datacommons/nl/. We can not track every such instance in _GCP_ENVIRONMENTS

keyurva: That's a great idea. Done.

shifucun: nice!

PR #3759, opened 2023-11-03
Fix webdriver tests

juliawu:
The webdriver tests have started failing after the last data push, because it is expecting older data values. This PR updates the failing tests to check for new values.

Note: Also updates integration test goldens.

Screenshots of current autopush behavior of pages being tested:
Screenshot 2023-11-03 at 9 36 29 AM
Screenshot 2023-11-03 at 9 34 19 AM


shifucun (commented):
Thank you so much! To make this more robust, shall we do inequality check like number of points > 5? @chejennifer  @dwnoble

chejennifer:
> Thank you so much! To make this more robust, shall we do inequality check like number of points > 5? @chejennifer @dwnoble

I think that's a good idea!

shifucun (approved):

juliawu:
> > Thank you so much! To make this more robust, shall we do inequality check like number of points > 5? @chejennifer @dwnoble
> 
> I think that's a good idea!

Updated to an inequality check!

PR #3760, opened 2023-11-03
Support google login for screenshot test

shifucun:


server/webdriver/screenshot/runner.py
```diff
@@ -31,6 +32,10 @@
 
 WAIT_TIMEOUT = 40
 
+_INSTANCE_WITH_IAP = ['https://autopush.datacommons.org']
```
pradh, line 35: could staging also go here?

pradh (approved):
Very cool!

server/webdriver/screenshot/runner.py
shifucun: Done

PR #3761, opened 2023-11-03
Start a react-tiles package folder

juliawu:
Create a containing folder for a new react-tiles NPM package. Initializes with a minimal `package.json` and `LICENSE` files. Future PRs will move our react chart tile code into this new package.


packages/react-tiles/package.json
```diff
@@ -0,0 +1,15 @@
+{
+  "name": "@datacommonsorg/react-tiles",
+  "version": "1.0.0",
```
dwnoble, line 3: nit: change to 0.0.1 (let's bump to 1.x when it's production-ready)

juliawu: Done!

PR #3764, opened 2023-11-07
[nodejs] update query response

chejennifer:
- add `unit` field for all charts
- add `highlight` field for single line line charts
- only return first 3 charts (use `&allCharts=1` to show all charts)

updated response:
https://paste.googleplex.com/5876459951357952

updated response w/ `&all=1`:
https://paste.googleplex.com/5318110008573952

### DIFF
--- a/static/src/server.ts
+++ b/static/src/server.ts
@@ -93,6 +93,10 @@ const NS_TO_MS_SCALE_FACTOR = BigInt(1000000);
 const MS_TO_S_SCALE_FACTOR = 1000;
 // The param value for the chartUrl param which indicates using svg
 const CHART_URL_PARAM_SVG = "0";
+// The param value if we should return all charts in /nodejs/query. Otherwise,
+// return QUERY_MAX_RESULTS.
+const SHOW_ALL_URL_PARAM = "1";
+const QUERY_MAX_RESULTS = 3;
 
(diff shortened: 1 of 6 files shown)


static/src/server.ts
```diff
@@ -457,6 +461,7 @@ app.get("/nodejs/query", (req: Request, res: Response) => {
   const startTime = process.hrtime.bigint();
   const query = req.query.q;
   const useChartUrl = req.query.chartUrl !== CHART_URL_PARAM_SVG;
+  const allResults = req.query.all === SHOW_ALL_URL_PARAM;
```
pradh, line 464: nit:  if we have a param, should we do maxResults?  Default = 3, to disable set 0.

if you prefer a bool, have a more descriptive param name?

pradh (approved):
Thanks!

Once the PR lands, could you please push to dev and update https://github.com/datacommonsorg/website/blob/3a1ca19f6dff04d010da7fe28c122d8e485ede28/server/tests/nodejs_e2e_test.py#L4 goldens?

static/src/server.ts
chejennifer: I prefered a bool so updated the param name to be `allCharts`

chejennifer:
thanks for the review! will do the push and golden updates!

PR #3769, opened 2023-11-07
Add latency column to sanity results.

keyurva:
image


server/webdriver/tests/standalone/sanity.py
```diff
@@ -78,10 +78,18 @@ def __init__(self,
...
+               page: WebPage,
+               status: str,
+               latency_sec: float,
+               comments: str = "") -> None:
     self.page_type = page.page_type
     self.title = page.title
     self.status = status
+    # Note that this latency is NOT the whole page latency.
+    # It is only the latency incurred until the page sanity is deemed to be a pass or a failure.
+    # It tends to be a small fraction of the full page latency.
```
pradh, line 91: any idea how small in practice?

pradh (approved):
Thanks!

server/webdriver/tests/standalone/sanity.py
keyurva: Anecdotally, the entire page load could be 3-4x this time. The perceived latency (display of charts above the fold) would be 1.5-2x.

pradh: What does "above the fold" mean?

The top few charts are often only in the user's "view", so I suppose latency for top-N those loading is more useful than the whole page.

Does the success criteria for this test cover some of that?  If not, we might consider a version of it that does.

keyurva: Above the fold is what the user sees on screen without scrolling. In our case it is typically the first chart.

The sanity check does not explicitly wait for the first chart to be completely rendered but for certain elements to become visible. 

@dwnoble @chejennifer @juliawu - is there an indicator (an element, css class, etc.) we can use as a signal that a chart was completely rendered?

chejennifer: "dc-async-element" class gets added to a chart when it's drawn: 

https://github.com/datacommonsorg/website/blob/0134730a843379b66e87bf9ca25550c68af410b1/static/js/constants/css_constants.ts#L22

https://github.com/datacommonsorg/website/blob/0134730a843379b66e87bf9ca25550c68af410b1/server/webdriver/shared.py#L24

keyurva: Thanks!

PR #3772, opened 2023-11-08
Do not use memcache for dev, staging and autopush

shifucun:
This is to make the screenshot test and latency profiling consistent with version changes.


pradh (approved):

server/integration_tests/test_data/e2e_edge_cases2/povertyvs.unemploymentrateindistrictsoftamilnadu/chart_config.json
```diff
@@ -31,12 +31,44 @@
             ],
             "denom": "Count_Person",
             "title": "Population Below Poverty Line"
+          },
```
shifucun, line 34: @pradh is this change valid?

pradh: This part is valid and comes from new data addition.

PR #3774, opened 2023-11-08
[seo] choose interesting rankings for place summaries

beets:
Some big updates:
- Added a script that filters all place explorer configs to ones that will show on the overview page. The intermediate file can then be updated to work with a shorter list of SV's, if needed.
- Updated all the calls in dc.py to use the Mixer directly - heavily borrowed from code in the flask server. Exposes more v1 APIs that can be used to extract more rankings per dcid, for example.
- Updated palm.py to naively pick a few top rankings to generate sentences from. I did minimal testing here, and there are some spurious results for states when there are blank prompts. Currently only processes Demographics SVs.


juliawu (approved):
Thanks for this!

PR #3775, opened 2023-11-08
Added perCapita support to bar, gauge, highlight, line, map, and ranking web components

dwnoble:
Screenshots

Screenshot 2023-11-08 at 2 40 08 PM
Screenshot 2023-11-08 at 2 40 02 PM
Screenshot 2023-11-08 at 2 12 38 PM
Screenshot 2023-11-08 at 10 29 04 AM
Screenshot 2023-11-08 at 8 19 07 AM


static/js/components/tiles/bar_tile.tsx
```diff
@@ -200,7 +200,7 @@ export const fetchData = async (props: BarTilePropType) => {
       ? Promise.resolve(null)
       : getSeriesWithin(
           apiRoot,
-          props["parentPlaces"],
+          props["parentPlace"],
```
chejennifer, line 203: nit: maybe should do `props.parentPlace` and `props.enclosedPlaceType` to avoid errors like this?

beets:
Fly by review - should the gauge chart value include "%" like the others?

packages/web-components/examples/all-charts.html
```diff
@@ -178,6 +191,18 @@ <h2>Pie/Donut</h2>
         </ul>
       </div>
     </datacommons-pie>
+    <datacommons-pie
+      header="Environmental Disasters in California"
+      place="geoId/06"
+      variables="Count_WildfireEvent Count_EarthquakeEvent"
+      donut
+    ></datacommons-pie>
+    <datacommons-pie
```
chejennifer, line 200: I think you wanted to add perCapita to this one?

static/library/highlight_component.ts
```diff
@@ -61,6 +66,10 @@ export class DatacommonsHighlightComponent extends LitElement {
   @property()
   place!: string;
 
+  // Optional: List of variable DCIDs to plot per capita
+  @property({ type: Array<string>, converter: convertArrayAttribute })
+  perCapita?: string[];
+
```
juliawu, line 72: Since highlight/gauge only take one variable, would we want to make this a boolean for highlight/gauge? Or keep the perCapita spec uniform across all tiles?

juliawu (approved):
Thanks for adding this!

static/library/map_component.ts
```diff
@@ -199,7 +204,10 @@ export class DatacommonsMapComponent extends LitElement {
           parentPlace: placeDcid,
           variable: {
             date,
-            denom: "",
+            denom:
+              this.perCapita && this.perCapita.includes(this.variable)
```
chejennifer, line 208: I think you want `this.perCapita.includes(variable)` here

static/library/map_component.ts
```diff
@@ -217,7 +225,10 @@ export class DatacommonsMapComponent extends LitElement {
           enclosedPlaceType: childPlaceType,
           parentPlace: place,
           variable: {
-            denom: "",
+            denom:
+              this.perCapita && this.perCapita.includes(this.variable)
```
chejennifer, line 229: also here

static/library/map_component.ts
```diff
@@ -245,7 +256,10 @@ export class DatacommonsMapComponent extends LitElement {
       },
       showExploreMore: this.showExploreMore,
       statVarSpec: {
-        denom: "",
+        denom:
```
chejennifer, line 259: looks like this is the empty variable case, so maybe keep denom as `""`

chejennifer (commented):
nice

packages/web-components/examples/all-charts.html
dwnoble: Removed the datacommons-pie example- I ended up not adding perCapita to the pie chart, because there's currently no tooltips and since pie charts are about a single place, the per capita and non per capita visualizations look the same

static/js/components/tiles/bar_tile.tsx
dwnoble: Good call- updated

static/library/highlight_component.ts
dwnoble: I went back and forth on this and and ended up going with uniform across the board. It's a little awkward for the single variable case, but i think i like the uniform appraoch. wdyt?

static/library/map_component.ts
dwnoble: Good catch

static/library/map_component.ts
dwnoble: Fixed

static/library/map_component.ts
dwnoble: Good call

dwnoble:
> Fly by review - should the gauge chart value include "%" like the others?

Agreed. ill address this in a follow-up: im seeing a case where the unit name is long (e.g., "Millions of tonnes") and overlaps with the gauge itself, and I want to make sure we handle it properly.

PR #3776, opened 2023-11-09
Improve custom embeddings load performance.

keyurva:
* Avoids downloading main dc embeddings and spacy when loading custom dc embeddings.
* Avoids loading main dc embeddings into a dataframe.
* Brings down NL load time (in docker) from ~30 seconds to ~5 seconds.

image


nl_server/embeddings_store.py
```diff
@@ -88,14 +113,17 @@ def _merge_custom_index(default: EmbeddingsIndex,
...
-  # Set ignore_index to reset index in output
-  concat_df = pd.concat([main_df, custom_df], ignore_index=True)
-  concat_df.to_csv(output_embeddings_path, index=False)
+  # First copy default embeddings to output file.
+  shutil.copy(default.embeddings_local_path, output_embeddings_path)
+  # Load custom embeddings into a dataframe
+  custom_df = pd.read_csv(custom.embeddings_local_path)
+  # Append custom embeddings to output file.
+  with open(output_embeddings_path, "a") as out:
+    out.write(custom_df.to_csv(header=None, index=False))
```
pradh, line 126: Call out the skipping of header explicitly?

nl_server/embeddings_store.py
```diff
@@ -49,6 +56,24 @@ def __init__(self, indexes: List[EmbeddingsIndex]):
   def get(self, index_type: str = DEFAULT_INDEX_TYPE) -> Embeddings:
     return self.embeddings_map.get(index_type)
 
+  def merge_custom_index(self, custom_idx: EmbeddingsIndex):
+    """Merges the specified custom index with the default index.
+
+    This method will be called if a new custom index is loaded at runtime
+    via the /api/load/ call.
+    """
+    default_idx = dataclasses.replace(self.original_default_idx)
```
pradh, line 65: Is the intent of `dataclasses.replace` the same as doing `copy.copy` / `copy.deepcopy`?  Maybe prefer the latter?

nl_server/loader.py
```diff
@@ -35,39 +38,61 @@
...
+  embeddings_map, _ = _load_yamls(flask_env)
+  custom_embeddings_local_path = embeddings_map.get(config.CUSTOM_DC_INDEX)
+  if not custom_embeddings_local_path:
+    logging.warning("No custom DC embeddings found, so none will be loaded.")
+    return
+
+  custom_idx = config.EmbeddingsIndex(
+      name=config.CUSTOM_DC_INDEX,
+      embeddings_file_name=os.path.basename(custom_embeddings_local_path),
+      embeddings_local_path=custom_embeddings_local_path)
```
pradh, line 83: Is this now making an assumption that `custom_embeddings_local_path` always gets updated?  Also that it is always local?

Can we check we're in "local GKE" or "RSI" mode?  Or use `config._parse()`, and it will then decide whether the path is local or not.

nl_server/loader.py
```diff
@@ -86,6 +111,29 @@ def _load_yamls(flask_env: str) -> tuple[Dict[str, str], Dict[str, str]]:
...
 
+def _update_app_config(app: Flask, nl_model: NLAttributeModel,
+                       nl_embeddings: embeddings_store.Store,
+                       embeddings_map: Dict[str, str]):
+  app.config[config.NL_MODEL_KEY] = nl_model
+  app.config[config.NL_EMBEDDINGS_KEY] = nl_embeddings
+  app.config[config.NL_EMBEDDINGS_VERSION_KEY] = embeddings_map
+
+
+def _maybe_cache(flask_env, nl_embeddings: embeddings_store.Store,
```
pradh, line 122: nit: _maybe_update_cache

also, `flask_env: str`

pradh (commented):
Very nice!

nl_server/embeddings_store.py
keyurva: Done.

nl_server/embeddings_store.py
keyurva: Done.

nl_server/loader.py
keyurva: Chatted offline. Called out in comment and added a TODO to call `config._parse()` instead.

nl_server/loader.py
keyurva: Done.

nl_server/embeddings_store.py
```diff
@@ -39,7 +39,7 @@ def __init__(self, indexes: List[EmbeddingsIndex]):
 
     # If a new custom index is loaded after startup, it will be merged with the
     # original default index maintained under this variable.
-    self.original_default_idx = dataclasses.replace(default_idx)
+    self.original_default_idx = copy.copy(default_idx)
```
pradh, line 42: To confirm, you don't want deepcopy?

nl_server/loader.py
```diff
@@ -119,7 +123,7 @@ def _update_app_config(app: Flask, nl_model: NLAttributeModel,
   app.config[config.NL_EMBEDDINGS_VERSION_KEY] = embeddings_map
 
 
-def _maybe_cache(flask_env, nl_embeddings: embeddings_store.Store,
+def _maybe_cache(flask_env: str, nl_embeddings: embeddings_store.Store,
```
pradh, line 126: _maybe_update_cache?

pradh (approved):

nl_server/embeddings_store.py
keyurva: It's not a compound object so should not be needed. But changed to `deepcopy` in case it is in the future.

nl_server/loader.py
keyurva: Done.

PR #3777, opened 2023-11-09
Use existence check to filter sources in stat var explorer page

shifucun:
Currently source filtering result in a large number api call that degrade the performance, especially in custom DC setup. Switch the check from stat var info to existence check, which is much more efficient.

Also no need to fetch the source names as they are already from the property value response.


keyurva (approved):
Nice.

Consider adding tests.

shifucun:
> Nice.
> 
> Consider adding tests.

This would involve a webdriver tests. Let me try to add later on.

PR #3780, opened 2023-11-10
[nodejs] Updates to query results

chejennifer:
- add a `dcLink` field that links to the datacommons explore page
- enable ranking results and return place names in the data csv
- hide map results

example for the query "which counties in california have the largest population"
https://paste.googleplex.com/5076381028319232

### DIFF
--- a/static/nodejs_server/types.ts
+++ b/static/nodejs_server/types.ts
@@ -41,4 +41,6 @@ export interface TileResult {
     value: number;
     date: string;
   };
+  // The link to the data commons explore page for this result.
+  dcLink?: string;
 }
--- a/static/nodejs_server/types.ts
+++ b/static/nodejs_server/types.ts
@@ -42,5 +42,5 @@ export interface TileResult {
     date: string;
   };
   // The link to the data commons explore page for this result.
-  dcLink?: string;
+  dcUrl?: string;
 }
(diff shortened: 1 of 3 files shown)


static/nodejs_server/types.ts
```diff
@@ -41,4 +41,6 @@ export interface TileResult {
     value: number;
     date: string;
   };
+  // The link to the data commons explore page for this result.
+  dcLink?: string;
```
pradh, line 45: nit: dcUrl ?  (to consistently refer to all URLs as such in this json)

static/src/server.ts
```diff
@@ -297,25 +301,22 @@ function getBlockTileResults(
               useChartUrl
             )
           );
-          break;
-        /* TODO: foreignobject doesn't work with the svg to png converter.
-                  Need to find a different way to render an svg of the ranking
-                  table & then re-enable this.
+          break;*/
```
pradh, line 304: maybe in future we might want a "client=" and decide based on that.  have a TODO?

pradh (approved):
Nice!

static/nodejs_server/types.ts
chejennifer: done

static/src/server.ts
chejennifer: added the client param

pradh (approved):
Thanks!

PR #3785, opened 2023-11-14
Plumb through client tags from JS to NL log

pradh:


static/js/constants/app/explore_constants.ts
```diff
@@ -37,6 +37,21 @@ export const URL_HASH_PARAMS = {
   MAXIMUM_BLOCK: "mb",
   TEST_MODE: "test",
   I18N: "i18n",
+  CLIENT: "client",
+};
+export const CLIENT_TYPES = {
+  // User typed in a p=X&t=Y URL
+  ENTITY: "ui_entity",
```
chejennifer, line 44: nit: should this be "ui_topic"? since this is a topic page?

.vscode/launch.json
```diff
@@ -20,6 +20,7 @@
       "env": {
         "FLASK_APP": "${workspaceRoot}/web_app.py",
         "FLASK_ENV": "local",
+        "ENV_PREFIX": "DC",
```
chejennifer, line 23: curious why this needs to be set?

chejennifer (approved):
nice!

static/js/constants/app/explore_constants.ts
pradh: More or less all of the below is from topic page.  The CLIENT_TYPES (maybe I need a better name) indicate the specific context for it.  This is where user provided a url with entity, maybe I'll make it `ui_url_entity` if its clearer/

.vscode/launch.json
pradh: This was a recent change made to the script: https://github.com/datacommonsorg/website/blob/9b3df3ee1b94d9c10d036da75672d342bf9896f1/run_server.sh#L62

But it breaks ability to run server in vscode for debugging, and this change fixes it.

pradh (commented):
Thanks for the review!

PR #3787, opened 2023-11-14
Address flakiness with i18n integration tests.

keyurva:


server/integration_tests/nl_test.py
```diff
@@ -131,6 +138,33 @@ def handle_response(self,
           self.assertEqual(dbg["main_place_dcid"], expected["main_place_dcid"])
           self.assertEqual(dbg["main_place_name"], expected["main_place_name"])
 
+  def handle_i18n_response(self, resp, i18n_lang):
```
pradh, line 141: Should we put this in a common place?

pradh (approved):
Thanks for the change!

server/integration_tests/nl_test.py
keyurva: For tests, prefer to keep assertion logic in the same file. I think `handle_response` in both files is also similar but kept separate for similar reasons?

pradh: Yeah, that's also worth deduping.  We can have a TODO

pradh: Sure

PR #3788, opened 2023-11-14
[testing cronjob] add nodejs & sanity tests

chejennifer:
- update the testing cronjob to run new nodejs query test & sanity tests
    - new nodejs query test runs queries against nodejs query endpoint & outputs results to gcs 

To come in future PRS:
- differ tool for nodejs query results between different runs
- try adding screenshot and adversarial tests


build/website_cron_testing/push_image.sh
```diff
@@ -19,6 +19,11 @@ set -e
 
 gcloud config set project datcom-ci
 
+cp ../../server/integration_tests/standalone/nodejs_query.py .
```
shifucun, line 22: this is a bit hacky, worth to find a better solution later.

run_nodejs_query.sh
```diff
@@ -0,0 +1,55 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# bash script to run nodejs query tests and save output to gs://datcom-website-periodic-testing
+
+set -e
```
shifucun, line 18: why is this file needed?

shifucun (commented):
Do you want to add a step in https://github.com/datacommonsorg/website/blob/master/build/ci/cloudbuild.push_image.yaml for this docker image build?

run_nodejs_query.sh
chejennifer: Not needed, but followed how the rest of adversarial, screenshot and sanity does it (having a script to manually run the test). This can be used if we want to manually do a run of the nodejs query test? But I'm also ok removing it, wdyt

server/integration_tests/standalone/nodejs_query.py
```diff
@@ -0,0 +1,76 @@
...
+    all_charts = test_case.get('all_charts', '')
+    client = test_case.get('client', '')
+    resp = requests.get(
+        f'{FLAGS.base_url}/nodejs/query?q={query}&allCharts={all_charts}&client={client}').json(
+        )
+    for chart in resp.get('charts', []):
+      chart['chartUrl'] = ''
+    file_name = test_case.get('name', '') + '.json'
+    json_file = os.path.join(FLAGS.output_dir, file_name)
+    with open(json_file, 'w') as out:
```
pradh, line 56: To confirm, does this not run diffs against test_data?

chejennifer: yes, this just saves the results in gcs & a follow up is to write a differ script that diffs between 2 sets of the outputs in gcs

run_nodejs_query.sh
shifucun: This script would not work locally for autopush now? As we want to add IAP for dev, it won't work there either. 

We should have a script to trigger the cron job to run in the GKE instead...

chejennifer: ah, ok, I didn't realize we were adding IAP for dev as well. I will remove this script & add a TODO for triggering the cron job to run in the GKE

shifucun: Sg

server/integration_tests/standalone/README.md
```diff
@@ -76,3 +76,17 @@ Note that leaving the `dc` flag to be empty will also default to the `--dc=main`
 ```shell
 python3 adversarial.py --mode=run_all --dc=sdg
 ```
+
```
shifucun, line 79: I guess no need to add the documentation below now?

shifucun (approved):

run_nodejs_query.sh
chejennifer: done

build/website_cron_testing/push_image.sh
chejennifer: this script is now removed

server/integration_tests/standalone/README.md
chejennifer: might be good to still describe what the test does? wdyt

chejennifer:
thanks for the review!

server/integration_tests/standalone/README.md
shifucun: Sure, but can mention it running in gke and remove from line 87?

shifucun (approved):
This is a really good addition to our testing infra! Great job!

server/integration_tests/standalone/README.md
chejennifer: done

PR #3790, opened 2023-11-14
Temporarily disable creating custom embeddings if data is in GCS.

keyurva:
* Disabling it temporarily until we have full support for GCS data based embeddings.

### DIFF
--- a/server/routes/admin/html.py
+++ b/server/routes/admin/html.py
@@ -47,6 +47,8 @@ def load_data():
   output_dir = os.path.join(sql_data_path, 'data')
   nl_dir = os.path.join(output_dir, "nl")
   sentences_path = os.path.join(nl_dir, "sentences.csv")
+  # TODO: Enable NL for GCS paths once we add support for it.
+  load_nl = not sql_data_path.lower().startswith("gs://")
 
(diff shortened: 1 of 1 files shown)


shifucun (approved):
Add a todo to have an option in docker run to enable/disable NL functions?

server/routes/admin/html.py
```diff
@@ -47,6 +47,9 @@ def load_data():
   output_dir = os.path.join(sql_data_path, 'data')
   nl_dir = os.path.join(output_dir, "nl")
   sentences_path = os.path.join(nl_dir, "sentences.csv")
+  # TODO: Support enabling / disabling NL via env / Dockerfile.
+  # TODO: Enable NL for GCS paths once we add support for it.
+  load_nl = not sql_data_path.lower().startswith("gs://")
```
pradh, line 52: Should we have a flag to explicitly enable/disable NL, even if sentences.csv is local?

keyurva: How about using `ENABLE_MODEL` for this?

pradh: Good idea

PR #3794, opened 2023-11-15
[nl] update existence

chejennifer:
- update fetching existence information to match the new facet response from mixer and add a date parameter (that will be used in a future PR)

### DIFF
--- a/server/lib/fetch.py
+++ b/server/lib/fetch.py
@@ -84,7 +84,10 @@ def _convert_v2_obs_point(facet):
 
 def _convert_v2_obs_series(facet):
   result = {
-      'facet': facet['facetId'],
+      'facet': facet.get('facetId', ''),
+      'obsCount': facet.get('obsCount', 0),
+      'earliestDate': facet.get('earliestDate', ''),
+      'latestDate': facet.get('latestDate', '')
   }
   if 'observations' in facet:
     result['series'] = facet['observations']
@@ -211,7 +214,7 @@ def point_within_core(ancestor_entity,
 
 
 def series_core(entities, variables, all_facets, facet_ids=None):
-  """Fetchs observation series for given entities and variables.
+  """Fetches observation series for given entities and variables.
 
   The response is in the following format:
   {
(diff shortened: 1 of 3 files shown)


server/lib/nl/common/utils.py
```diff
@@ -108,28 +108,40 @@ def sv_existence_for_places(places: List[str], svs: List[str],
...
-      if not place_data.get('series'):
-        continue
-      num_series = place_data['series'][0]["value"]
-      existing_svs[sv] = existing_svs.get(sv, False) | (num_series == 1)
-      if sv not in existsv2places:
-        existsv2places[sv] = {}
-      existsv2places[sv][pl] = (num_series == 1)
+      for facet_data in place_data:
+        # If there is a date specified, but facet dates are different
+        # granularity, move on to next facet
```
pradh, line 128: If the query asks for 2020, and we have 2020-06 I think we can answer?

server/lib/nl/fulfillment/existence.py
```diff
@@ -65,7 +65,7 @@ def _run(self):
     # TODO: Optimize this!
     self.existing_svs, existsv2places = \
       utils.sv_existence_for_places_check_single_point(
-        self.places, list(self.all_svs), self.state.uttr.counters)
+        self.places, list(self.all_svs), None, self.state.uttr.counters)
```
pradh, line 68: nit: do date=None ?

server/lib/nl/common/utils.py
pradh: If you agree, have a TODO for revisiting (so we fix autopush first)?

server/lib/fetch.py
```diff
@@ -211,7 +215,7 @@ def point_within_core(ancestor_entity,
 
 
 def series_core(entities, variables, all_facets, facet_ids=None):
```
pradh, line 217: It seems this changes the returned structure, so to confirm, are all callers of series_core good?

pradh (approved):

server/lib/fetch.py
```diff
@@ -242,21 +246,29 @@ def series_facet(entities, variables, all_facets):
...
       }
     }
   }
 
   """
   resp = dc.series_facet(entities, variables)
-  return _compact_series(resp, all_facets)
+  compacted_series = _compact_series(resp, all_facets)
+  processed_series = {'facets': compacted_series.get('facets', {}), 'data': {}}
+  # Update compacted series so that the entity data is always a list.
```
shifucun, line 263: Instead of updating here, do you want to update compact_series() function directly?

I feel it's nice to make _compact_series and _compact_point to return uniform response (the origin decision to return either {} or [] is not optimal)

chejennifer: This will be a bigger change (need to also update the frontend logic that calls series/point APIs), so added a TODO for this

server/lib/fetch.py
chejennifer: yeup, all callers of series_core are good

server/lib/nl/common/utils.py
chejennifer: that makes sense to me, but will need more thought about how to pass this date to the chart config. Added a TODO & I will also create a buganizer issue for this.

chejennifer:
thanks for the reviews!

PR #3796, opened 2023-11-16
Switch nodejs to use explore backend

pradh:
Switch Nodejs server to use the Explore backend.  Because Explore returns more charts (e.g., ranking for just contained-in query), and used in production.

Skip additional processing to build `relatedThings` in Nodejs flow (aka `mode=strict`), since the caller (bard) doesn't use it.

### Results for [obesity in counties of california]
image


chejennifer (approved):
thanks for the change!

PR #3798, opened 2023-11-16
[nodejs] fix ranking csv bug

chejennifer:
fix bug where ranking table csv is in the wrong order

updated response for family earnings in california counties
https://paste.googleplex.com/6309252029743104

### DIFF
--- a/static/nodejs_server/ranking_tile.ts
+++ b/static/nodejs_server/ranking_tile.ts
@@ -107,8 +107,11 @@ function getRankingUnitResult(
   useChartUrl: boolean,
   apiRoot: string
 ): TileResult {
+  const points = isHighest
+    ? rankingGroup.points.slice().reverse()
+    : rankingGroup.points;
   const result: TileResult = {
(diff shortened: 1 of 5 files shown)


static/nodejs_server/ranking_tile.ts
```diff
@@ -107,8 +107,11 @@ function getRankingUnitResult(
   useChartUrl: boolean,
   apiRoot: string
 ): TileResult {
+  const points = isHighest
+    ? rankingGroup.points.slice().reverse()
```
pradh, line 111: Will this sort by RankingPoint.value?

pradh (approved):
Thanks!

static/nodejs_server/ranking_tile.ts
chejennifer: yes, rankingGroup.points is already sorted by value, but from lowest to highest

chejennifer:
thanks for the review!

PR #3799, opened 2023-11-17
In strict mode, bail on finding temporal words

pradh:
In strict mode, when we know we don't specially handle certain temporal words (like december), better to to not continue processing.

image


shifucun (approved):

pradh:
Thanks!

PR #3800, opened 2023-11-17
Update custom DC README.

keyurva:


custom_dc/README.md
```diff
@@ -6,6 +6,17 @@ custom to a third party's need.
 This document illustrates development and deployment details for a custom Data
 Commons instance.
 
+## Custom and Base Data Commons
+
+<img src="https://github.com/datacommonsorg/website/assets/1221814/63efdb88-580c-4bc2-967b-5dce61844eea"
```
pradh, line 11: nit: there appears to be a lot of whitespace around the image, any chance of trimming?

custom_dc/README.md
```diff
@@ -6,6 +6,17 @@ custom to a third party's need.
 This document illustrates development and deployment details for a custom Data
 Commons instance.
 
+## Custom and Base Data Commons
+
+<img src="https://github.com/datacommonsorg/website/assets/1221814/63efdb88-580c-4bc2-967b-5dce61844eea" 
+     alt="Custom Data Commons" 
+     style="width: 40vw;"/>
+
+The diagram above is a high level view of the relationship between a custom and base DC.
```
pradh, line 15: high-level

pradh, line 15: "custom and base DC" => Custom and Base DC instances"

pradh: Also, add "(DC)" the first time "Data Commons" is mentioned above

custom_dc/README.md
```diff
@@ -6,6 +6,17 @@ custom to a third party's need.
...
+As indicated, custom and base DCs largely share the same software stack.
+Additionally a custom DC interfaces with its own custom data and as well as with main DC.
+The latter enables combining custom DC datasets with those from base DC.
```
pradh, line 18: suggestion:  "Additionally, a custom DC combines its own local datasets with base DC datasets to provide a unified view."

custom_dc/README.md
```diff
@@ -193,6 +204,9 @@ before the operation.
 
 ### Prepare Custom Data
 
+> Note that the treatment of variables in this section is much simplified for the purposes of this discussion.
```
pradh, line 207: It seems we do not introduce the word "variables" or "statistical variables" at all in this doc?  In the context of the CSV below, we say that `stat_var_1` and `stat_var_2` column names correspond to variables?  And then we can add this note.

pradh (commented):
Thanks!

custom_dc/README.md
keyurva: It seems to add a paragraph tag around it. Tried a few things but could not get rid of it.

custom_dc/README.md
keyurva: Done.

keyurva: Done.

custom_dc/README.md
keyurva: Done.

custom_dc/README.md
keyurva: I think a separate high-level doc might be a better option to introduce concepts. There are a few that could be useful: variables, entities, dcid mapping / resolution, etc.

Make a suggestion if you want a short blurb in this doc.

pradh: I feel inlining a one line intro (to avoid needing to read a new doc) is helpful

For example:

```
The first column header is a property that identifies the observed entity,
supported properties are `dcid`, `name`, `geoId`. When `dcid` is used, the
entity should have been resolved (from a previous step).

The second column is a property that identifies the date.

The last two columns identify IDs of metrics or statistical variables, with the column containing the corresponding numeric values.

> Note that the treatment of statistical variables is much simplified for the purposes of this discussion. For a more detailed discussion on the DC data model, please refer to the document on [Representing statistics in Data Commons](https://github.com/datacommonsorg/data/blob/master/docs/representing_statistics.md).

```

pradh (approved):

custom_dc/README.md
keyurva: This is good. Applied it with a couple changes.

keyurva (commented):
Thanks for the review.

PR #3801, opened 2023-11-17
[nodejs] update highlighted point in line result

chejennifer:
- pass date in line config
- update the highlighted point in the nodejs line result to be the specified date
- skip returning a line chart result if there is a specified date, but no data point for that date

### DIFF
--- a/static/nodejs_server/line_tile.ts
+++ b/static/nodejs_server/line_tile.ts
@@ -97,15 +97,32 @@ export async function getLineTileResult(
     // If it is a single line chart, add highlight information.
     if (chartData.dataGroup && chartData.dataGroup.length === 1) {
       const dataPoints = _.cloneDeep(chartData.dataGroup[0].value);
-      if (!_.isEmpty(dataPoints)) {
+      const highlightDate = statVarSpec[0].date;
+      let highlightPoint = null;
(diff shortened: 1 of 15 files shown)


static/nodejs_server/line_tile.ts
```diff
@@ -97,15 +97,32 @@ export async function getLineTileResult(
     // If it is a single line chart, add highlight information.
     if (chartData.dataGroup && chartData.dataGroup.length === 1) {
       const dataPoints = _.cloneDeep(chartData.dataGroup[0].value);
-      if (!_.isEmpty(dataPoints)) {
+      const highlightDate = statVarSpec[0].date;
+      let highlightPoint = null;
+      if (!_.isEmpty(dataPoints) && highlightDate) {
+        highlightPoint = dataPoints.find(
+          (point) => point.label === highlightDate
+        );
```
pradh, line 105: TODO: consider if we want to allow for monthly stats if year is asked

pradh (approved):
Thanks!

static/nodejs_server/line_tile.ts
chejennifer: added

chejennifer:
thanks for the review!

PR #3805, opened 2023-11-17
[cron testing] Update dev and bard yamls to handle cron tests

chejennifer:
update bard and dev deployment yamls to be able to handle cron tests

### DIFF
--- a/deploy/helm_charts/envs/bard.yaml
+++ b/deploy/helm_charts/envs/bard.yaml
@@ -47,3 +47,6 @@ serviceGroups:
 
 svg:
   blocklistFile: ["dc/g/Uncategorized", "oecd/g/OECD"]
+
+cronTesting:
+  enabled: true
--- a/deploy/helm_charts/envs/dev.yaml
+++ b/deploy/helm_charts/envs/dev.yaml
@@ -45,3 +45,6 @@ nl:
 nodejs:
   enabled: true
   replicas: 50
+
+cronTesting:
+  enabled: true


deploy/helm_charts/envs/bard.yaml
```diff
@@ -47,3 +47,6 @@ serviceGroups:
 
 svg:
   blocklistFile: ["dc/g/Uncategorized", "oecd/g/OECD"]
+
+cronTesting:
```
shifucun, line 51: Can you add 

```
nodejs:
  enabled: true
  replicas: 50
```
?

Now nodejs is not enabled in bard

chejennifer: I'm not sure who's standing up the bard instance and whether or not they've completed the setup, because adding this would be part of that whole setup process right? so I was just going to add the testing changes so that once nodejs is up and running on bard, it can do the tests

shifucun: I see. With `cronTesting: enabled`, would the cron testing start to run every 4 hours?

shifucun (approved):

deploy/helm_charts/envs/bard.yaml
chejennifer: no, `cronTest: enabled` just writes values to the cronTesting configMap during deployment

also, when cron testing runs, it will check that there is a node server & if no node server, it will skip the nodejs query tests

PR #3811, opened 2023-11-28
Start custom dc NL server and load embeddings conditionally.

keyurva:


server/routes/admin/html.py
```diff
@@ -47,9 +47,10 @@ def load_data():
   output_dir = os.path.join(sql_data_path, 'data')
   nl_dir = os.path.join(output_dir, "nl")
   sentences_path = os.path.join(nl_dir, "sentences.csv")
-  # TODO: Support enabling / disabling NL via env / Dockerfile.
   # TODO: Enable NL for GCS paths once we add support for it.
-  load_nl = not sql_data_path.lower().startswith("gs://")
+  load_nl = os.environ.get(
```
shifucun, line 51: split and reformat this expression for better readability

shifucun (approved):

server/routes/admin/html.py
keyurva: Yeah, not a fan of this formatting either. But it was done by the linter so leaving it as-is.

keyurva (commented):
Thanks for the review.

PR #3814, opened 2023-11-29
Custom DC update to sync recent mixer change

shifucun:
* Use mixer /update-cache endpoint.
* Do not use "LITE" mode for "custom" flaskEnv. This would require MPAS_API_KEY to set.
* Update mixer cacheSVG flag arg
* Update mixer sqlite database path
* Disable NL by default


keyurva (approved):
Cool!

PR #3815, opened 2023-11-29
[CUSTOM DC] Clean up variable definition and management

shifucun:


custom_dc/README.md
```diff
@@ -68,6 +69,13 @@ This brings up a local instance with sample data that are stored under
 Now you can open `localhost:8080/tools/timeline` to browse these sample data.
 Also note the base Data Commons data are also avaiable in this instance.
 
+[Note]: The local sample data path should always bind to `/userdata`.
+
+[Note]: The SQLite database is in the Docker runtime and does not persist. To
+persist the SQLite database, can bind it to a local directory by adding `-v
+<local_path>:/sqlite/` in the `docker run` command. By doing this, when you run
+the command and starting the servers, it can use the existing database.
+
```
keyurva, line 78: Perhaps another note indicating that sqlite should be used for local development only:

> NOTE: SQLite should be used for local development only. For production cloud deployments, use Cloud SQL.

build/web_compose/compose.sh
```diff
@@ -15,36 +15,38 @@
...
+    export USER_DATA_PATH=$GCS_DATA_PATH
+fi
+
 /go/bin/mixer \
     --use_bigquery=false \
     --use_base_bigtable=false \
     --use_custom_bigtable=false \
     --use_branch_bigtable=false \
-    --sqlite_path=$SQL_DATA_PATH/data/datacommons.db \
+    --sqlite_path=$SQLITE_PATH \
```
keyurva, line 30: Assuming mixer is ok if `sqlite_path` is empty.

keyurva (approved):

build/web_compose/compose.sh
shifucun: Mixer would only use sqlite_path if use_sqlite == true, so it's fine to be empty.

custom_dc/README.md
shifucun: Yea, maybe do it in a follow up PR. As you suggested, should have a table diagram with a lot more discussions on storage and database.

PR #3817, opened 2023-11-29
Add plumbing for place summaries

juliawu:
This PR adds plumbing to show place summaries and have an allowlist of places to show summaries for on staging and prod.

In particular, we can now add the summaries to show to `place_summary_content.json` in GCS, and control which places to show summaries for by editing the allow list in `place_summary_config.json` in GCS.

Also fixes a typo in the README for summary generation.

Changes tested by verifying some sample place pages locally.


server/__init__.py
```diff
@@ -284,8 +285,10 @@ def create_app(nl_root=DEFAULT_NL_ROOT):
   if app.config['ENABLE_ADMIN']:
     register_routes_admin(app)
 
-  # Load place explorer summaries
-  app.config['PLACE_EXPLORER_SUMMARIES'] = libutil.get_place_summaries()
+  # Load place explorer summaries & allowlist of places to show summaries for
+  # Used when rendering place pages
+  app.config['PLACE_SUMMARY_ALLOW_LIST'] = place_summaries.get_place_allowlist()
+  app.config['PLACE_EXPLORER_SUMMARIES'] = place_summaries.get_place_summaries()
```
shifucun, line 291: This config is created on server start time. If we change the json content, needs to restart server. This does not require a release but should trigger a GKE pod restart by updating a dummy field in the yaml file.

server/routes/place/html.py
```diff
@@ -73,20 +73,27 @@ def place(place_dcid=None):
     place_name = place_dcid
 
   place_summary = None
-  if os.environ.get('FLASK_ENV') in ['autopush', 'local']:
+  show_summary = False
+  if os.environ.get('FLASK_ENV') in ['autopush', 'local', 'staging', 'prod']:
     if not category:
       # Only show summary for Overview
       place_summary = current_app.config['PLACE_EXPLORER_SUMMARIES'].get(
```
shifucun, line 80: looks like PLACE_EXPLORER_SUMMARIES already have the allowed list as the key, why still need json for PLACE_SUMMARY_ALLOW_LIST?

juliawu: Was originally thinking we would have separate entries: one EXPLORER_SUMMARIES with content that would show on _autopush_ while we develop, and ALLOW_LIST are the subset of summaries allowed on _prod_. Would we prefer it to be just one file?

shifucun: I see, the if condition here is a bit confusing. From what you explain, I feel the condition should be:


show_summary = (place_dcid in place_allow_list) or (place_dcid in current_app.config['PLACE_EXPLORER_SUMMARIES']) # check prod then check other environment
place_summary = current_app.config['PLACE_EXPLORER_SUMMARIES'][place_dcid]

shifucun (approved):

server/routes/place/html.py
juliawu: Good call out. Reworked the if conditions to make the logic more clear.

PR #3827, opened 2023-12-06
[legacy scatter] Fix duplicate API calls

chejennifer:
fix bug where scatter plot fetches the same data 3 times


shifucun (approved):
nice!

PR #3828, opened 2023-12-06
Added internal load balancer option for website gke deployments

dwnoble:
See `datcom-bard` project for example configuration


gke/README.md
```diff
@@ -25,7 +25,7 @@ You should have owner/editor role to perform the following tasks.
 
 ### One time setup
```
shifucun, line 26: Add the step to manually setup apigee? Feel like need to talk about that here but it also deserves a standalone setup guide..

shifucun (commented):
this is very cool!

gke/README.md
```diff
@@ -55,16 +55,19 @@ You should have owner/editor role to perform the following tasks.
 
     # Deploy esp service
     ./setup_esp.sh
+
+    # [Internal load balancer services only] Configure internal load balancer network and dns settings
```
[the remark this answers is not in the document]
dwnoble: Done

gke/configure_internal_load_balancer.sh
```diff
@@ -0,0 +1,60 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Configures internal load balancer requirements
+# (1) proxy subnet
+# (2) self -signed certificate for internal load balancer
```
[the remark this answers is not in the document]
dwnoble: fixed

deploy/apigee/deploy_proxy.sh
```diff
@@ -0,0 +1,61 @@
...
+set -e
+
+ENV=$1
+
+if [[ $ENV == "" ]]; then
+  echo "Missing arg 1 (env)"
+  exit 1
+fi
+
+PROJECT_ID=$(yq eval '.project' config.yaml)
```
chejennifer, line 28: I wonder if this script should live under gke where the rest of the one time setup scripts are because:

1. The instructions for running this script is in the README for that folder
1. I think this script needs to be run from that folder in order for this line to work right? because I don't think there is a config.yaml file in this folder?

dwnoble: That makes sense to me. Moved to gke directory.

chejennifer (approved):
thanks for updating!

PR #3840, opened 2023-12-13
Update custom dc sample data.

keyurva:
* The current sample CSVs are more about examples of what the input CSVs could be. So moved them to a new `examples` folder.
* Added more real world CSVs (OECD wage data) to the `sample` folder instead. New users will likely try out the quick start guide first and since it loads data from the sample folder, it will give them a more real world idea of the possibilities.


juliawu (approved):
Nice change! Real-world sample data is a nice touch.

keyurva:
Thanks for the review.

PR #3825, opened 2023-12-04
Update Place Overview UI

juliawu:
Lays the groundwork for the UI update for the place overview of the place pages based on figma mocks. This PR leaves the summary text as a paragraph in the overview tile. 

Left for follow-up PRs:
* Switching to sentence-by-sentence highlights will come in a followup PR.
* Updating the other components on the place pages to match Figma mocks.

Pages With Summary
Screenshot 2023-12-04 at 9 25 48 AM

Pages Without Summary
Screenshot 2023-12-04 at 9 25 57 AM

Testing:
Verified changes locally don't break the place pages, and that the place overview shown on the NL interface does not change.

PR #3841, opened 2023-12-13
[Place Page] Add nl search bar to place page

juliawu:
As followup to https://github.com/datacommonsorg/website/pull/3825, this PR replaces the tiny "Graph Browser" link on the place pages with the NL search bar on top of the page, following the design set in our figma mocks.

Updating the UI treatment for switching places, tabbing sections, and chart styling will come in future PRs.

Before:
Screenshot 2023-12-13 at 12 57 29 PM

After:
Screenshot 2023-12-13 at 12 57 49 PM

### DIFF
--- a/static/js/place/place.ts
+++ b/static/js/place/place.ts
@@ -20,7 +20,15 @@ import React from "react";
 import ReactDOM from "react-dom";
 
 import { PageData } from "../chart/types";
+import { NlSearchBar } from "../components/nl_search_bar";
 import { loadLocaleData } from "../i18n/i18n";
+import {
+  GA_EVENT_NL_SEARCH,
+  GA_PARAM_QUERY,
+  GA_PARAM_SOURCE,
+  GA_VALUE_SEARCH_SOURCE_EXPLORE_LANDING,
+  triggerGAEvent,
+} from "../shared/ga_events";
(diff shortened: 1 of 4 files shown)


static/js/place/place.ts
```diff
@@ -123,6 +131,18 @@ async function getLandingPageData(
...
 }
 
+/**
+ * Handler for NL search bar
+ * @param q search query entered by user
+ */
+function onSearch(q: string): void {
+  triggerGAEvent(GA_EVENT_NL_SEARCH, {
+    [GA_PARAM_QUERY]: q,
+    [GA_PARAM_SOURCE]: GA_VALUE_SEARCH_SOURCE_EXPLORE_LANDING,
```
dwnoble, line 141: Should this source be the place page instead of explore landing?

juliawu: Good catch! Updated the source to be the place page.

dwnoble (approved):

pradh:
@juliawu catching up on this...

1. do we need to do anything special to ensure a question that's asked uses the current place as the default one?

2. curious about why we remove the graph browser link?  (IIRC it was there for devs to get access to the place's props, incl dcid)

PR #3842, opened 2023-12-13
[nl] Remove date range from query before variable detection

chejennifer:
This fixes a problem of the date range part of a string causing the variable match score to be low.
For example, the query "Population of California in 2013" was showing as low confidence because the query used to detect the variable was "Population in 2013" even though the actual variable part of the query should be "Population".

Changes made:
- change the order of detection to: 
    - detect heuristics
    - use detected heuristics to update the query
    - detect variable using the updated query
- update debug info in the ui to have "query without places" AND "query used for variable detection"
- update DateClassificationAttributes to return the trigger string

Also updated date parsing to understand "past decade"


server/lib/nl/detection/utils.py
```diff
@@ -207,3 +209,17 @@ def get_multi_sv(main_vars: List[str], cmp_vars: List[str],
   if not dvars.deduplicate_svs(res.candidates[0]):
     return None
   return res
+
+
+# Gets the query string to use for sv detection.
+def get_sv_detection_query(query: str,
```
jehangiramjad, line 215: nit: since we're only doing date string trigger removals, would it be more appropriate for this function to named something like ``remove_date_detection_from_query`` ?

server/tests/lib/nl/date_parser_test.py
```diff
@@ -34,6 +34,8 @@ class TestDateParser(unittest.TestCase):
       ('Earthquakes in the last 5 years',
        [Date('before', 2022, 0, year_span=5)]),
       ('Flood in previous year', [Date('before', 2022, 0, year_span=1)]),
+      ('Population over the past decade',
```
jehangiramjad, line 37: nit: would be nice to have some strings which do not trigger the date detector.

jehangiramjad (approved):
Thanks!

server/lib/nl/detection/utils.py
chejennifer: The thought behind making this function more general was that in the future we might want to remove other trigger strings. But since we're only removing date for now, I've updated the function to be specific to removing date & we can always make the function more general again in the future when we want to remove more things.

server/tests/lib/nl/date_parser_test.py
chejennifer: done

chejennifer:
thanks for the review!

PR #3844, opened 2023-12-15
[nl] Use facetId in timeline config

chejennifer:
Fix a problem with date handling where only a specific series has data for specified date, but the LINE tile may use a different series. This fix makes NL pass the facet id of the series that we want used (if it matters) to the config.

before: https://screenshot.googleplex.com/4LGnpRH3x9FsULs (notice that the line chart starts after 2001)
after: https://screenshot.googleplex.com/9cgjkFrcUCEEtcN

### DIFF
--- a/server/integration_tests/test_data/e2e_single_date/populationintheusinthelastyear/chart_config.json
+++ b/server/integration_tests/test_data/e2e_single_date/populationintheusinthelastyear/chart_config.json
@@ -10,7 +10,7 @@
                 "tiles": [
                   {
                     "statVarKey": [
-                      "Count_Person_2022"
+                      "Count_Person_2022_2176550201"
                     ],
                     "title": "Population in United States",
                     "type": "LINE"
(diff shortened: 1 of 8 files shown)


jehangiramjad (approved):
This looks good to me! I have approved it to unblock you and as long as all the tests pass this should be Ok. But would be great to make sure you eventually get @pradh and @shifucun to review (they are both OOO today).

chejennifer:
thanks for the review! Will merge for now and can have a follow up PR for any new comments

server/lib/nl/config_builder/timeline.py
```diff
@@ -25,6 +25,13 @@
 import server.lib.nl.fulfillment.types as types
 
 
+# Get facet id to use
```
pradh, line 28: Maybe add some comments that we use a facet *only* if the date is set (or better yet, rename the function).  Currently that is hidden in the logic...

server/integration_tests/test_data/e2e_single_date/populationintheusinthelastyear/chart_config.json
```diff
@@ -10,7 +10,7 @@
                 "tiles": [
                   {
                     "statVarKey": [
-                      "Count_Person_2022"
+                      "Count_Person_2022_2176550201"
```
pradh, line 13: Maybe add a focussed test-case for this?

pradh (commented):
Nicely done!

PR #3848, opened 2023-12-19
[nl] handle date range for timeline charts

chejennifer:
Add handling in NL for specifying a date range for timeline charts. Date range would be stuff like "before 2020", "after 2013", "last 5 years" etc.

Changes:
- Update existence check to check for date range
- pass date range from heuristic classification to chart spec
- Use the date range when specifying a timeline chart

TODO in future PRs:
- Frontend changes to show only a specific date range in a timeline tile
- Handle ranked timeline blocks
- Figure out what to do about date ranges for other chart types


server/lib/nl/common/utils.py
```diff
@@ -133,19 +149,20 @@ def _facet_contains_date(facet_data, facet_metadata, date) -> bool:
...
     # of the facet
-    year_diff = date.year - int(facet_earliest_date[:4])
-    if date.month:
+    year_diff = single_date.year - int(facet_earliest_date[:4])
+    if single_date.month:
       # Difference in number of months between the date of interest
       # and the earliest date of the facet
-      month_diff = year_diff * 12 + date.month - int(facet_earliest_date[5:])
+      month_diff = year_diff * 12 + single_date.month - int(
+          facet_earliest_date[5:])
```
jehangiramjad, line 157: nit: I know this has been there before but would be great to add a comment about the magic number 5

server/lib/nl/common/utils.py
```diff
@@ -369,6 +389,17 @@ def get_single_date(uttr: nl_uttr.Utterance) -> types.Date:
...
 
 
+def get_date_range(uttr: nl_uttr.Utterance) -> types.Date:
+  classification = futils.classifications_of_type_from_utterance(
+      uttr, types.ClassificationType.DATE)
+  if (classification and isinstance(classification[0].attributes,
+                                    types.DateClassificationAttributes)):
+    # Return the first date in the classification if it is not a single date
+    if not classification[0].attributes.is_single_date:
+      return classification[0].attributes.dates[0]
```
jehangiramjad, line 399: curious: why not return the full range instead of just the first date?

server/lib/nl/config_builder/timeline.py
```diff
@@ -96,21 +98,27 @@ def ranked_timeline_collection_block(builder: base.Builder,
...
+  date_string = get_date_string(single_date)
   if date_string:
     sv_key += f'_{date_string}'
-  facet_id = _get_facet_id(sv_dcid, date, cv)
+  facet_id = _get_facet_id(sv_dcid, single_date or date_range, cv)
   if facet_id:
     sv_key += f'_{facet_id}'
   tile = Tile(type=Tile.TileType.LINE, title=title, stat_var_key=[sv_key])
+  if date_range:
+    start_date, end_date = get_date_range(date_range)
```
jehangiramjad, line 119: hmmm. related to the comment earlier on this function, here it seems you're returning a tuple but the function comment makes it seem to expect only one date?

server/lib/nl/config_builder/timeline.py
```diff
@@ -147,9 +156,13 @@ def single_place_multiple_var_timeline_block(column,
 
   # Line chart for the stat var
   tile = Tile(type=Tile.TileType.LINE, title=title, stat_var_key=[])
-  date_string = get_date_string(date)
+  if date_range:
+    start_date, end_date = get_date_range(date_range)
```
jehangiramjad, line 160: this block seems to be repeated a couple of times. would a helper function be warranted?

jehangiramjad (commented):
Thanks! A bit complex so might need a couple of rounds so I am clear what's going on :)

server/lib/nl/config_builder/timeline.py
chejennifer: hm, I feel like the function comment indicates 2 strings are returned? Or do you have any suggestions to improve this? 

"""
# Gets the date range as 2 ISO-8601 formatted strings (i.e., YYYY-MM or YYYY).
# First string is the start date and the second string is the end date.
"""

server/lib/nl/common/utils.py
chejennifer: Each Date object in DateClassificationAttributes is actually either a date range or single date. There really should only be one date in the list of dates, so for simplicity, just return the first object in the list for now.

server/lib/nl/common/utils.py
```diff
@@ -108,19 +108,35 @@ def sv_existence_for_places(places: List[str], svs: List[str],
   return existing_svs, existsv2places
 
 
-def _facet_contains_date(facet_data, facet_metadata, date) -> bool:
-  date_string = server.lib.nl.detection.date.get_date_string(date)
+def _facet_contains_date(facet_data, facet_metadata, single_date,
+                         date_range) -> bool:
+  start_date, end_date = server.lib.nl.detection.date.get_date_range(date_range)
```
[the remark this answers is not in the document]
chejennifer: Yes, this will always succeed & return two strings (although, if no date range, it will return 2 empty strings)

server/lib/nl/common/utils.py
chejennifer: done

server/lib/nl/config_builder/timeline.py
chejennifer: done

server/lib/nl/common/utils.py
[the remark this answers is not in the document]
jehangiramjad: Got it, thanks!

server/lib/nl/common/utils.py
jehangiramjad: Thanks!

server/lib/nl/common/utils.py
jehangiramjad: Interesting. I must have missed this but would be nice to have this document somewhere obvious.

jehangiramjad (approved):
Thanks for the updates and the replies.

I see the confusion about the documentation for the function ``get_date_range()``. Turns out, in this PR you have two of them? One is in ``lib/nl/detection/date.py`` (this is the one that you were referring to) and the other one seems to be in ``... lib/nl/common/utils.py`` (that's the one I was referring to). Perhaps a good idea to disambiguate them by changing the function name for one of them?

server/lib/nl/config_builder/timeline.py
jehangiramjad: Turns out, in this PR you have two of these functions (``get_date_range()``)? One is in ``lib/nl/detection/date.py`` (this is the one that you were referring to) and the other one seems to be in ``... lib/nl/common/utils.py`` (that's the one I was referring to). Perhaps a good idea to disambiguate them by changing the function name for one of them?

server/lib/nl/config_builder/timeline.py
jehangiramjad: Thanks!

server/lib/nl/config_builder/timeline.py
chejennifer: ohh you're right, that IS confusing. Updated the one in date.py to `get_date_range_strings`

chejennifer:
thanks for the review!

server/lib/nl/detection/date.py
```diff
@@ -31,12 +31,16 @@
...
+    r'(in|after|on|before|since|by|until|from|between) (Jan|Feb|Mar|Apr|May|Jun|Jul|Aug|Sep|Oct|Nov|Dec)(?:,)? (\d{4})',
+    r'(in|after|on|before|since|by|until|from|between) (\d{4})(?:,)? (Jan|Feb|Mar|Apr|May|Jun|Jul|Aug|Sep|Oct|Nov|Dec)',
 ]
 
 # List of date preps that indicate single date
 _SINGLE_DATE_PREPS = ['in', 'on']
+# List of date preps that indicate that the base date is a start date
+_START_DATE_PREPS = ['after', 'since', 'from']
+# List of date preps that indicate that the base date is an end date
+_END_DATE_PREPS = ['before', 'by', 'until']
```
pradh, line 43: To confirm, should we handle <= vs. < ?

pradh (commented):
Nicely done!  Just one comment...

server/lib/nl/detection/date.py
chejennifer: ah, yes, that would make sense, will do in a follow up PR

PR #3853, opened 2023-12-26
[embeddings] Remove dependency on google sheets

chejennifer:
Remove dependency on google sheets when building embeddings by using the curated input csv files on github instead.

Also in this PR:
- update sdg embeddings because sdg curated input (in the sheet) had been updated since the last time the sdg embeddings were built.
    - sv differ validation: https://pantheon.corp.google.com/storage/browser/_details/datcom-embedding-diffs/chejennifer_sdg_ft_2023_12_26_10_22_37.html;tab=live_object?mods=-monitoring_api_staging. Diffs mainly caused by:
        - targets/indicators that were removed. Checked that these are targets/indicators that do not show up on the [UN goal explorer](https://unstats.un.org/UNSDWebsite/undatacommons/sdgs/goals)
        - typos in descriptions that got fixed
- remove `data/` from gitignore file because it was causing all files in tools/nl/embeddings/data to be ignored ..
- separate main and sdg alternatives files because the main embeddings was created without the extra descriptions from the sdg palm alternatives


tools/nl/embeddings/README.md
```diff
@@ -9,16 +9,9 @@ There are multiple sizes of indexes: small, medium.
...
-
-Latest sheet as of May 2023 is
-[`DC_NL_SVs_Curated`](https://docs.google.com/spreadsheets/d/1-QPDWqD131LcDTZ4y_nnqllh66W010HDdows1phyneU).
-
-This is a common sheet across the different index sizes.
-
 ## Making a change to the embeddings.
 
-1. For any carefully curated SVs/topics, make edits to the [latest sheet](https://docs.google.com/spreadsheets/d/1-QPDWqD131LcDTZ4y_nnqllh66W010HDdows1phyneU). The columns in this sheet are:
+1. For any carefully curated SVs/topics, make edits to the [latest sheet](https://docs.google.com/spreadsheets/d/1-QPDWqD131LcDTZ4y_nnqllh66W010HDdows1phyneU). The columns in this sheet are: <a name="curated-input"></a>
```
jehangiramjad, line 14: A bit unclear here. Why do we still need this reference to the Google sheet?

tools/nl/embeddings/README.md
```diff
@@ -52,15 +41,17 @@ This is a common sheet across the different index sizes.
...
    To generate the `sdg_ft` embeddings:
    ```bash
-   ./run.sh -c sdg https://docs.google.com/spreadsheets/d/1-QPDWqD131LcDTZ4y_nnqllh66W010HDdows1phyneU SDG_GOALS_INDICATORS_ONLY data/curated_input/sdg_sheets_svs.csv
+   ./run.sh -c sdg data/curated_input/sdg_sheets_svs.csv data/alternatives/sdg/*.csv
    ```
 
    You can also create custom embeddings (using the finetuned model in PROD):
    ```bash
-   ./run.sh -c <embeddings_size> <sheets_url> <worksheet_name> data/curated_input/sheets_svs.csv
+   ./run.sh -c <embeddings_size> <curated_input_csv_path> <alternatives_filepattern>
```
jehangiramjad, line 49: nit: how about an example (hypothetical)? That will make clear how to provide relative paths or absolute paths to these files, right? Or do these need to be under secific subpaths?

tools/nl/svindex_differ/differ.py
```diff
@@ -224,6 +224,8 @@ def run_diff(base_file, test_file, base_model_path, test_model_path, query_file,
   for _, diff_table_rows in diffs:
     for row in diff_table_rows:
       for info in row:
+        if not info:
```
jehangiramjad, line 227: question: when would this happen?

jehangiramjad (commented):
Thanks! Mostly looks good but just some comments/questions. Also, I believe we might want to converge on the resolution for user-friendly editing UX on the email thread wih @pradh . But I suspect we can do that as a follow up to this PR?

tools/nl/embeddings/README.md
chejennifer: done, added an example in the Notes

tools/nl/embeddings/README.md
chejennifer: Ah, I had changed that but somehow undid it at some point. Updated

tools/nl/svindex_differ/differ.py
chejennifer: Turns out when getting the diff_table_rows, I wasn't handling a 4th case that the Differ library returns where it returns a "?" which means the line is not present in either input.. fixed

tools/nl/embeddings/build_embeddings.py
```diff
@@ -42,23 +42,16 @@
...
-                    'Local Sheets csv (relative) file path')
-flags.DEFINE_string(
-    'sheets_url',
-    'https://docs.google.com/spreadsheets/d/1-QPDWqD131LcDTZ4y_nnqllh66W010HDdows1phyneU',
-    'Google Sheets Url for the latest SVs')
-flags.DEFINE_list(
-    'worksheet_names', ['Demo_SVs', 'MAIN_DC_SDG_GOALS_INDICATORS_ONLY'],
-    'List of names of worksheets in the Google Sheets file to use')
+flags.DEFINE_string('curated_input_path',
+                    'data/curated_input/main_sheets_svs.csv',
```
pradh, line 46: super nit: feel free to ignore or defer.  like we have data/alternatives/main|sdg|xyz, we could have sub-dirs in data/curated_input/

pradh (approved):
Nice!  Any context on why/when for the SDG changes being updated here?

tools/nl/embeddings/build_embeddings.py
chejennifer: done

pradh (approved):
Thanks.

chejennifer:
> Nice! Any context on why/when for the SDG changes being updated here?

I think most of the changes happened when adding the sdg index to the main dc index, but I didn't rebuild the sdg index.

- For the removals, I remember deleting the ones that didn't exist in our KG (I'm not sure when they got removed from the KG though & I also double checked the UN site and they don't appear in the goals explorer)
- For the typo fixes, I remember some of the cases like: eduation -> education & "aged 10-14 years; aged 15 -19..." -> "aged 10-19 years"

jehangiramjad (approved):
Thank for the updates and clarifications!

chejennifer:
Thanks for the reviews!

Failing tests are not due to this PR:
- Python test goldens will be fixed here: https://github.com/datacommonsorg/website/pull/3854
- Webdriver tests are failing due to a missing drought event.

PR #3854, opened 2023-12-26
Fix repo tests

chejennifer:
update goldens for latest nodejs response update (from last week) where:
- scatter plots re-enabled
- ranking table can show top ... bottoms ranks

Also:
- fix webdriver tests by updating the events tested for on the events page are not autogenerated events (because autogenerated events may sometimes be modified). 
- update event page to handle when no location found


pradh (approved):
Thanks for the fixes!

chejennifer:
Thanks for the review!

PR #3859, opened 2024-01-02
Add support for downloading default FT model and embeddings.

keyurva:
* Added a mode to the `build_custom_dc_embeddings.py` script to download default FT model and embeddings.
* Updated `Dockerfile` to call this script when building the docker image.
* Server startup with NL enabled improved from O(mins) to O(seconds).
* Does not download place summaries for custom DCs (to circumvent a GCP auth failure) for now.


keyurva:
~ make sure we'll be using the appropriate/latest base model+embeddings version?

This does not change the version the custom DC image would be using. The embeddings.yaml is already burnt into the image. Even prior to this change, that was the version used by the custom DC NL server ([go/custom-dc-embeddings](http://goto.google.com/custom-dc-embeddings)). That continues to be the case after this change as well. The only difference is that the model and embeddings are burnt into the image so they don't need to be downloaded on every docker run.

shared/lib/gcs.py
```diff
@@ -26,13 +27,18 @@
 def download_file(bucket: str,
                   filename: str,
                   use_anonymous_client: bool = False) -> str:
+  local_embeddings_path = os.path.join(TEMP_DIR, filename)
+  if os.path.exists(local_embeddings_path):
+    logging.info("File already exists, skipping GCS download: %s",
```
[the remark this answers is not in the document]
keyurva: Done.

tools/nl/embeddings/build_custom_dc_embeddings.md
```diff
@@ -28,6 +28,17 @@ To use a different model version, specify the `--model-version` flag.
 --output_dir=/tmp
 ```
 
+## Download default FT model and embeddings
```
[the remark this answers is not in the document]
keyurva: Done.

shared/lib/gcs.py
```diff
@@ -26,13 +27,18 @@
 def download_file(bucket: str,
                   filename: str,
                   use_anonymous_client: bool = False) -> str:
+  local_embeddings_path = os.path.join(TEMP_DIR, filename)
+  if os.path.exists(local_embeddings_path):
+    logging.info("File already exists: %s, skipping GCS download.",
+                 local_embeddings_path)
+    return local_embeddings_path
+
```
pradh, line 35: Looks like this is a general lib used by many callers?  To be sure, do we want to do this check just in the callsite?

nit:  Maybe rename "local_embeddings_path" to "local_file_path" ?

pradh:
> ~ make sure we'll be using the appropriate/latest base model+embeddings version?
> 
> This does not change the version the custom DC image would be using. The embeddings.yaml is already burnt into the image. Even prior to this change, that was the version used by the custom DC NL server ([go/custom-dc-embeddings](http://goto.google.com/custom-dc-embeddings)). That continues to be the case after this change as well. The only difference is that the model and embeddings are burnt into the image so they don't need to be downloaded on every docker run.

Got it!

pradh (approved):

juliawu (approved):
Reviewed the place summary changes, which lgtm. Thanks for the fix!

server/__init__.py
```diff
@@ -287,8 +287,12 @@ def create_app(nl_root=DEFAULT_NL_ROOT):
...
   # Load place explorer summaries & allowlist of places to show summaries for
   # Used when rendering place pages
-  app.config['PLACE_SUMMARY_ALLOW_LIST'] = place_summaries.get_place_allowlist()
-  app.config['PLACE_EXPLORER_SUMMARIES'] = place_summaries.get_place_summaries()
+  app.config[
+      'PLACE_SUMMARY_ALLOW_LIST'] = [] if cfg.CUSTOM else place_summaries.get_place_allowlist(
+      )
+  app.config[
+      'PLACE_EXPLORER_SUMMARIES'] = {} if cfg.CUSTOM else place_summaries.get_place_summaries(
+      )
```
shifucun, line 295: nit: shall we have a regular if else block so the if block can hold or CUSTOM handlings so it's easier to update later.

shifucun (approved):
very nice change and refactor!

server/__init__.py
keyurva: Done.

keyurva:
Thanks all for the reviews!

PR #3861, opened 2024-01-05
[nl] Enable date range for ranking timelines

chejennifer:
Enabled date range for ranked timeline blocks
- added logic to rank a group of timelines for a specified date range
- include date range when building ranked timeline blocks
- update the sv_existing_facet_id field in ChartVars to be a dict of sv to dict of place to facet id because different places may have data for different facets. This change means when getting the facet id to build a line chart, if the line chart has more than one place, need to choose the facet with data for the most places.
    - this also fixes a bug in pages like this (https://screenshot.googleplex.com/5QmasfDuf3cCg32) because when facet was saved in sv => facet format, the facet that was saved could've been the facet that exists for a child place and not the parent place


server/lib/nl/fulfillment/types.py
```diff
@@ -71,9 +71,9 @@ class ChartVars:
   # Set if is_topic_peer_group is set.
   svpg_id: str = ''
 
-  # Map of sv to facet metadata that has data for this sv. Used by LINE tiles
+  # Map of sv to place to facet metadata that has data for this sv. Used by LINE tiles
   # when there is a date specified, and by the "is_coplottable" check.
-  sv_exist_facet = Dict[str, Dict[str, str]]
+  sv_exist_facet = Dict[str, Dict[str, Dict[str, str]]]
```
pradh, line 76: Should we use a dataclass now?  So its more clear everywhere we pass around?

server/lib/nl/fulfillment/utils.py
```diff
@@ -165,6 +166,26 @@ def get_default_contained_in_place(places: List[Place],
   return constants.DEFAULT_PARENT_PLACES.get(ptype, None)
 
 
+# Get facet id to use when there is a date specified (use a facet id that has
+# data). Gets the facet id that has data for the most places.
+def get_facet_id(sv: str, date: Date, sv_exist_facet: Dict[str, Dict[str, str]],
```
pradh, line 171: Does the new structure of sv_exist_facet impact is_coplottable?

server/lib/nl/common/rank_utils.py
```diff
@@ -163,15 +190,29 @@ def rank_svs_by_series_growth(place: str, svs: List[str],
...
   earliest = None
-  # TODO: Apparently series is ordered, so simplify.
-  for s in series:
-    if not latest or s['date'] > latest['date']:
-      latest = s
-    if not earliest or s['date'] < earliest['date']:
-      earliest = s
+  latest = None
+  if date_range:
+    # Get the earliest and latest dates in the series that are within the date
```
pradh, line 199: could this use a helper?

server/lib/nl/fulfillment/types.py
chejennifer: discussed offline to have a followup PR to try to enhance https://github.com/datacommonsorg/website/blob/368d8d7cdfc1cd5086809d15b39c820d1fe763e7/server/lib/nl/fulfillment/types.py#L122-L123 to save facet metadata, so won't make the structure change now

server/lib/nl/fulfillment/utils.py
chejennifer: good call, updated!

server/lib/nl/common/rank_utils.py
chejennifer: done

server/integration_tests/test_data/e2e_date_range/whichcountriesinafricahavehadthegreatestincreaseinelectricityaccessoverthelast10years/chart_config.json
```diff
@@ -19,7 +19,7 @@
                       "country/DZA"
                     ],
                     "lineTileSpec": {
-                      "endDate": "2023",
+                      "endDate": "2024",
```
pradh, line 22: Is this new data?

pradh (approved):
Thanks and sorry for the delay!

server/integration_tests/test_data/e2e_date_range/whichcountriesinafricahavehadthegreatestincreaseinelectricityaccessoverthelast10years/chart_config.json
chejennifer: this was from the inclusive/exclusive date range change earlier this week (last 10 years should include the current year)

chejennifer:
thanks for the review!

PR #3862, opened 2024-01-05
Update Nodejs loadtest results

chejennifer:
It's been awhile since we've updated the loadtest results and there have been updates to the endpoint response.


pradh (approved):
Hmmm, are things seeming slower than before in general?

chejennifer:
> Hmmm, are things seeming slower than before in general?

I haven't noticed things being slower .. looking at the results, some queries slowed down but others have gotten faster, but these times will always fluctuate

pradh:
> > Hmmm, are things seeming slower than before in general?
> 
> I haven't noticed things being slower .. looking at the results, some queries slowed down but others have gotten faster, but these times will always fluctuate

From my skim things are a bit slower more often than not.  Might be worth keeping an eye on...

PR #3864, opened 2024-01-06
Support Gemini Pro as an option

pradh:
Drop the currently unused PaLM text API support, and replace it with Gemini Pro.

Rename "PALM" things with "LLM", since we now have both PaLM and Gemini models.

**Testing**

* Manually tested with `llm_api=geminipro&detector=llm` (example, http://localhost:8080/explore/#q=income+in+california+counties+with+median+age+over+40&detector=llm&llm_api=geminipro)

* Need to run adversarial with `llm_api=geminipro`

**Notes**

Generally, Gemini seems to adhere to the prompt better.  For instance, [employment where obama was born], PaLM will expand the place to Hawaii, but Gemini won't.  Gemini also does better with 2 variable constraint queries, like [income in california counties with median age over 40], while PaLM often gets it wrong.

The Gemini API is also faster than PaLM Chat (the one we use).


server/integration_tests/standalone/adversarial.py
```diff
@@ -52,8 +52,8 @@ class Mode:
 
 flags.DEFINE_enum(
     "llm_api",
-    "chat",
-    ["chat", "text"],
+    "palm",
+    ["palm", "geminipro"],
     f"Value for the llm_api parameter. Valid values: chat, text.",
```
chejennifer, line 57: update valid values here?

server/lib/nl/detection/detector.py
```diff
@@ -65,13 +65,13 @@ def detect(
   #
   # In the absence of the PALM API key, fallback to heuristic.
   #
-  if (detector_type in _PALM_API_DETECTORS and
-      'PALM_API_KEY' not in current_app.config):
+  if (detector_type in _LLM_API_DETECTORS and
+      'LLM_API_KEY' not in current_app.config):
     counters.err('failed_palm_keynotfound', '')
```
chejennifer, line 70: should we update this to 'failed_llm_keynotfound'?

server/lib/nl/detection/detector.py
```diff
@@ -65,13 +65,13 @@ def detect(
...
     detector_type = RequestedDetectorType.Heuristic.value
 
-  if (detector_type in _PALM_API_DETECTORS and
-      'PALM_PROMPT_TEXT' not in current_app.config):
+  if (detector_type in _LLM_API_DETECTORS and
+      'LLM_PROMPT_TEXT' not in current_app.config):
     counters.err('failed_palm_promptnotfound', '')
```
chejennifer, line 75: should we update this to 'failed_llm_promptnotfound'?

server/lib/translator.py
```diff
@@ -168,9 +168,9 @@ def _maybe_add(string: str):
 
 def _translate(queries: list[str], source_lang: str,
                target_lang: str) -> list[str]:
-  # The name "PALM_API_KEY" is a misnomer.
+  # The name "LLM_API_KEY" is a misnomer.
   # Both palm and translation apis are enabled under this key.
```
chejennifer, line 172: palm should be llm?

server/tests/lib/nl/detection/llm_api_test.py
```diff
@@ -106,16 +106,19 @@ class TestParseResponse(unittest.TestCase):
   def test_chat(self, input, want):
```
chejennifer, line 106: should test_chat and test_text_block be updated to test_gemini and test_palm? and then use the respective extraction functions in each function?

chejennifer (commented):
nice! Just a few places that I think need a naming update

server/tests/lib/nl/detection/llm_api_test.py
pradh: test_chat has both palm and gemini, but renamed them slightly.

server/lib/translator.py
pradh: yes!

server/lib/nl/detection/detector.py
pradh: yes!

server/integration_tests/standalone/adversarial.py
pradh: yes!

server/lib/nl/detection/detector.py
pradh: yes!

pradh (commented):
Thanks for the careful review!

chejennifer (approved):
thanks for the updates!

PR #3866, opened 2024-01-09
Added redis config to bard env, updated node distribution to be more balanced

dwnoble:


deploy/helm_charts/envs/bard.yaml
```diff
@@ -21,9 +21,16 @@ namespace:
 
 website:
   flaskEnv: custom
-  replicas: 4
+  replicas: 12
```
pradh, line 24: 20?

pradh (approved):
Thanks!

deploy/helm_charts/envs/bard.yaml
dwnoble: done

PR #3868, opened 2024-01-09
[nl] Allow fulfillment when there are only multi svs detected

chejennifer:
When there are no svs detected because the scores were too low, we should still try to populate charts for multi sv candidates (if there are any detected)

This fixes the empty result in strict mode for the query 'Does obesity correlate with lack of sleep in US counties' (it returns results in regular mode).

Example: nodejs/query?q=Does%20obesity%20correlate%20with%20lack%20of%20sleep%20in%20US%20counties
currently: empty result
with this change: https://paste.googleplex.com/5581611641339904

### DIFF
--- a/server/lib/nl/fulfillment/base.py
+++ b/server/lib/nl/fulfillment/base.py
@@ -188,9 +188,6 @@ def _maybe_switch_parent_type(
 # Add charts given a place and a list of stat-vars.
 def _add_charts_with_existence_check(state: PopulateState,
                                      places: List[Place]) -> bool:
-  svs = state.uttr.svs
-  logging.info("Add chart %s" % (', '.join(_get_place_names(places))))
-
   # This may set state.uttr.place_fallback
   _maybe_set_fallback(state, places)
 
@@ -251,7 +248,7 @@ def _add_charts_with_existence_check(state: PopulateState,
         not state.ranking_types and num_charts < _MAX_NUM_CHARTS):
       # Note that we want to expand on existing_svs only, and in the
       # order of `svs`
-      ordered_existing_svs = [v for v in svs if v in existing_svs]
+      ordered_existing_svs = [v for v in state.uttr.svs if v in existing_svs]
       found |= _add_charts_for_extended_svs(state=state,
                                             places=places,
                                             svs=ordered_existing_svs,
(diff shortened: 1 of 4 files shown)


server/lib/nl/fulfillment/base.py
```diff
@@ -55,7 +55,7 @@ def populate_charts(state: PopulateState) -> bool:
   places = state.uttr.places
   handle_contained_in_type(state, places)
 
-  if not state.uttr.svs:
+  if not state.uttr.svs and not state.uttr.multi_svs:
```
pradh, line 58: Should we check `state.chart_vars_map` instead?

Also, do the same in line 68?

server/lib/nl/fulfillment/base.py
```diff
@@ -55,7 +55,7 @@ def populate_charts(state: PopulateState) -> bool:
...
     state.uttr.counters.err('num_populate_fallbacks', 1)
     state.uttr.sv_source = FulfillmentResult.UNRECOGNIZED
     return False
```
pradh, line 61: Also, don't add `state.uttr.svs` in RHS line 67.

Also, drop RHS line 190+191 (printing state.uttr.svs to log)?

pradh (commented):
Nice find!

server/lib/nl/fulfillment/base.py
chejennifer: ah yes, that makes sense, updated

server/lib/nl/fulfillment/base.py
chejennifer: done

server/lib/nl/fulfillment/base.py
```diff
@@ -188,7 +189,7 @@ def _maybe_switch_parent_type(
 def _add_charts_with_existence_check(state: PopulateState,
                                      places: List[Place]) -> bool:
   svs = state.uttr.svs
```
pradh, line 191: Drop the logging (we don't really use it)?   Also is `svs` variable used?

pradh (approved):
Thanks!

server/lib/nl/fulfillment/base.py
chejennifer: dropped! the `svs` variable was used later in line 251, but removed `svs` now and updated that line to just use `state.uttr.svs`

chejennifer:
thanks for the review!

PR #3869, opened 2024-01-10
Use docker run --pull=always in custom DC documentation

shifucun:
Docker cache images and prevents users to use the latest docker image. This PR fixed that.

### DIFF
--- a/custom_dc/README.md
+++ b/custom_dc/README.md
@@ -34,7 +34,7 @@ Additionally, a custom DC combines its own local datasets with base DC datasets 
 
 ## Quick Start
 
-Once the prerequisites and api keys from above are in place, 
+Once the prerequisites and api keys from above are in place,
 here's how you can start a local custom DC instance quickly.
 
 > Note that this is only a quick start section. See the rest of the sections for more details.
@@ -61,6 +61,7 @@ To start the custom DC services, in the root of this repository, run Docker as f
 
 ```bash
 docker run -it \
+--pull=always \
 -p 8080:8080 \
 -e DEBUG=true \
 --env-file $PWD/custom_dc/sqlite_env.list \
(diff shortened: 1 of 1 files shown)


custom_dc/README.md
```diff
@@ -125,6 +128,7 @@ In the root of this repository, run:
 
 ```bash
 docker run -it \
+--pull=always \
```
keyurva, line 131: I think pulling when running might sometimes lead to unintended surprises.

Perhaps add a section that guides them to pull the latest image explicitly from the repo: `docker pull gcr.io/datcom-ci/datacommons-website-compose:latest`?

shifucun: Why would that give unintended surprises? If we always want users to get refreshed image, then might just have one command to do it?

keyurva: When they run it the first time and they get the refreshed image, that's perfectly ok.

I'm concerned about when they are just bouncing their servers and it pulls in a new image because of `--pull-always`. The new image might have changes / issues they were not expecting.

shifucun: Currently, after rebuilding local docker image, when re-run it, the new docker image is not picked up due to caching. I think that's what we see from Alan's issue.

shifucun: If we really want users to stick with a particular docker image, then shall use <TAG> instead of "latest" or "stable" throughout the documentation. Now "stable" does not match the "stable" version is surprising.

keyurva: Ah, this is for the locally built image and not the one from the repo.

OK - giving lgtm.

We should do a separate discussion on whether moving forward we want customers to build their own image at all or always use ones from our repo. Basically, the service (image) comes from us and the env, data and any webpage customizations from them.

keyurva (approved):

PR #3871, opened 2024-01-10
Set unit to % for all variations where scaling factor equals 100

chejennifer:
scaling factor can be 100, 100.0, 100.00, etc. We should set the unit to % for all variations.


jehangiramjad (approved):
Thanks!

PR #3872, opened 2024-01-10
[nl] allow passing a default place to detect and fulfill

chejennifer:
This allows the Explore page to take a default place that will be used when trying to detect and fulfill a query. If no default place is passed to the endpoint, the regular default place logic will be used (if sdg, default place is Earth, otherwise, default place is USA)

e.g., when default_place is set to India: https://screenshot.googleplex.com/8fcrdGyvgrzS23j


server/routes/nl/helpers.py
```diff
@@ -47,17 +47,34 @@
...
 
 _SANITY_TEST = 'sanity'
 
 
+# Get the default place to be used for fulfillment. If there is a place in the
+# request, use that. Otherwise, use pre-chosen places.
+def _get_default_place(request: Dict, is_sdg: bool):
+  default_place_dcid = request.args.get('default_place', default='', type=str)
+  if default_place_dcid:
+    place_type = get_place_type(default_place_dcid)
```
pradh, line 68: nit:  Does this add an API call?  Should we short-circuit types for `country/` and Earth DCIDs?

pradh (commented):
Nice!

static/js/apps/explore/app.tsx
```diff
@@ -413,6 +418,9 @@ const fetchDetectAndFufillData = async (
   if (client) {
     argsMap.set(URL_HASH_PARAMS.CLIENT, client);
   }
+  if (defaultPlace) {
```
dwnoble, line 421: Is this necessary? won't this be handled by the search context?

chejennifer: sorry, what do you mean by the search context? Do you mean to just always set default place in the args map instead of having it in a conditional?

server/routes/nl/helpers.py
chejennifer: done

static/js/apps/explore/app.tsx
dwnoble: i was thinking-do we have to set this url parameter at all for subsequent searches? if the user doesn't specify a place, will the user still get the same default place (only out of the context history instead of the default_place arg)

chejennifer: if there is a place in the context history, subsequent searches will just use that. But this call is used in both the first search and later searches, so we would still need to set default_place?

server/routes/nl/helpers.py
```diff
@@ -47,17 +47,34 @@
...
+    place_name = names([default_place_dcid])[default_place_dcid]
```
pradh, line 76: Sorry for missing this before, instead of 2 API calls, I think we can just call place.get_place_from_dcids() which will use `get_place_info` to return `Place` objects.

pradh (approved):

server/routes/nl/helpers.py
chejennifer: oh, that's a useful method, updated!

pradh (approved):

server/routes/nl/helpers.py
```diff
@@ -62,19 +61,17 @@
...
   elif default_place_dcid:
-    if default_place_dcid.startswith(constants.COUNTRY_DCID_PREFIX):
-      place_type = 'Country'
-    else:
-      place_type = get_place_type(default_place_dcid)
-    place_name = names([default_place_dcid])[default_place_dcid]
-    return Place(default_place_dcid, place_name, place_type)
+    places, _ = get_place_from_dcids([default_place_dcid], debug_logs)
+    if len(places) > 0:
+      return places[0]
```
pradh, line 74: If for whatever reason places is empty, then this function won't return anything?  Perhaps we want to return None and check in the caller?

chejennifer: Actually, what should the caller do if there is no default place returned? Should the caller just continue without a default place or use the original logic for getting default place?

pradh: Hmmm, three options:
1. Use the original logic
2. Make up a place as:  `Place(type='Place', name=dcid, dcid=dcid)`
3. Continue without default place (aka return None here)

I prefer (1) less.  (2) assumes the DCID exists, which may fail in other ways.

So I was suggesting (3), in which case, I think we would return an error message like no place found in the query.  Perhaps test it out...

static/js/apps/explore/app.tsx
dwnoble: That makes sense!

dwnoble (approved):

server/routes/nl/helpers.py
chejennifer: updated to return an error message when there is no place: https://screenshot.googleplex.com/587DDdCAGTF2XH2

server/integration_tests/test_data/explore_strict_default_place/howtoearnmoneyonlinewithoutinvestment/chart_config.json
```diff
@@ -1,24 +1,16 @@
 {
+  "blocked": true,
```
pradh, line 2: why is blocked true in this case?

server/integration_tests/test_data/inappropriate_query/query_1/chart_config.json
```diff
@@ -10,5 +10,7 @@
     "name": "",
     "place_type": ""
   },
-  "userMessage": "Sorry, could not complete your request."
+  "userMessage": {
```
pradh, line 13: What does the client read?  Wondering how this was working before

pradh (commented):
Thanks.

server/integration_tests/test_data/inappropriate_query/query_1/chart_config.json
chejennifer: the client reads userMessage.msg. The client actually wasn't reading the userMessage from "abort" cases & just using a DEFAULT_MESSAGE set on the client side

pradh: Good find

server/integration_tests/test_data/explore_strict_default_place/howtoearnmoneyonlinewithoutinvestment/chart_config.json
chejennifer: ah, accidentally passed that to the abort message, fixed

pradh (approved):

chejennifer:
thanks for the reviews!

PR #3876, opened 2024-01-11
Use the right native and foreign SVs in the index

pradh:
In the index today, we do not have SV for "US native population" (`nativity=Native`).  What exists is "US natives born outside US".  So fix that.

Also, for "foreign born population", we have `placeOfBirth=ForeignBorn`, rather than `nativity=ForeignBorn`.  This makes comparing against native population not entirely accurate (esp considering nativity has a dpv of >5 years).

So, this PR makes the following replacements in the SV index:

* Count_Person_ForeignBorn => Count_Person_5OrMoreYears_ForeignBorn
* Count_Person_NativeBornOutsideTheUnitedStates => Count_Person_Native

Diffs are clean - https://storage.mtls.cloud.google.com/datcom-embedding-diffs/shanth_medium_ft_2024_01_10_17_44_39.html - except for a "use electricity" change that I'm not fully sure (will poke some more).


jehangiramjad (approved):
Thanks! 

Regarding the Electricity topic diff, it's not a real diff. See: https://screenshot.googleplex.com/BQcBuyxgW37F4u6 and notice the top SV in all these three topics is the same (hence, the same top matching score). So I bet it's randomly changing the order of the topics here. We should probably figure out if these three topics are different enough or could be collapsed somehow?

pradh:
> Thanks!
> 
> Regarding the Electricity topic diff, it's not a real diff. See: https://screenshot.googleplex.com/BQcBuyxgW37F4u6 and notice the top SV in all these three topics is the same (hence, the same top matching score). So I bet it's randomly changing the order of the topics here. We should probably figure out if these three topics are different enough or could be collapsed somehow?

Nice, thanks for finding that!

pradh:
> Thanks!
> 
> Regarding the Electricity topic diff, it's not a real diff. See: https://screenshot.googleplex.com/BQcBuyxgW37F4u6 and notice the top SV in all these three topics is the same (hence, the same top matching score). So I bet it's randomly changing the order of the topics here. We should probably figure out if these three topics are different enough or could be collapsed somehow?

Sorry missed addressing the question.  I think we should sort deterministically.  While not ideal, we do have the same concept represented by different SVs/topics, where different SVs have data for different places (US vs. global vs. SDG).  We could merge, but it would also be subject to #svs we open up for performance.

I've filed b/319546168 for follow up.

PR #3878, opened 2024-01-11
[nodejs] Fix the root of returned chartUrl for API proxy instances

chejennifer:
The url root used when generating the chartUrl used an internal protocol/host when the nodejs server was behind an API proxy, but we want to use the actual protocol/host that the user inputted
- update apigee config to forward the original host in the request header
- on the node server, read the protocol and host from forwarded values in the request header if available
- on the node server, also read and add apikey to the chartUrl


dwnoble (approved):
Nice!

chejennifer:
thanks for the review!

PR #3884, opened 2024-01-16
Load place page summaries immediately on load

juliawu:
Move the place summaries out of the overview tile and into the DOM directly. This means the summary text will load before our heavy typescript/react logic. This will help our SEO by preventing search crawlers from seeing a completely empty page when charts render too slowly.

This is intended as a quick change to allow us to start experimenting with ways to fix "Soft 404" errors. Larger code refactor / optimization to come in future PRs.

Screenshot:

Screenshot 2024-01-15 at 7 16 21 PM


static/js/place/place.ts
```diff
@@ -279,7 +278,6 @@ function renderPage(): void {
           categoryStrings: data.categories,
           locale,
           highlight: data.highlight,
-          summaryText,
```
dwnoble, line 282: Can we remove `summaryText` as a prop from `MainPanePropType` altogether now?

dwnoble (approved):

static/js/place/place.ts
juliawu: Done!

PR #3888, opened 2024-01-17
[nl] update date range to consider > vs >=

chejennifer:
Handle the difference between exclusive prepositions (e.g., "before 2020", "after 2020") vs inclusive (e.g., "until 2020", "since 2020").

Also fix:
- when determining if a date is a single date or a date range, should just look at the preposition
- for cases like "over the last x year(s)", we should also include current date


server/lib/nl/detection/date.py
```diff
@@ -35,13 +35,18 @@
     r'(in|after|on|before|since|by|until|from|between) (\d{4})(?:,)? (Jan|Feb|Mar|Apr|May|Jun|Jul|Aug|Sep|Oct|Nov|Dec)',
 ]
 
+# prep to use for LAST_YEAR dates which should include the current year.
+_LAST_YEARS_PREP = 'last_years'
```
pradh, line 39: nit: mention why the `_`?

server/lib/nl/detection/date.py
```diff
@@ -35,13 +35,18 @@
...
 # List of date preps that indicate single date
-_SINGLE_DATE_PREPS = ['in', 'on']
+_SINGLE_DATE_PREPS = ['in', 'on', 'year']
```
pradh, line 41: nit: `year` isn't a preposition, but do you mean for queries like: "population in year 2020" ?

server/lib/nl/detection/date.py
```diff
@@ -86,27 +87,27 @@ def parse_date(query: str, ctr: Counters) -> DateClassificationAttributes:
       count, = match.groups()
       count_num = _YEARS_STRING_TO_NUM.get(count, count)
       year = datetime.date.today().year
-      dates.append(Date('before', year - 1, year_span=int(count_num)))
+      dates.append(Date(_LAST_YEARS_PREP, year, year_span=int(count_num)))
       trigger_strings.append(query[match.start():match.end()])
 
   for pattern in LAST_YEAR:
```
pradh, line 93: Maybe add some comments about what is going on here?

server/lib/nl/detection/date.py
```diff
@@ -115,30 +116,60 @@ def get_date_string(date: Date) -> str:
...
+  # if date range excludes the specified date, need to do some calculations to
+  # get the base date.
+  if date.prep in _EXCLUSIVE_DATE_PREPS:
+    # if specified date is an end date, base date should be earlier than
+    # specified date
+    if date.prep in _END_DATE_PREPS:
+      # if date is monthly, use date that is one month before the specified date
+      if date.month >= _MIN_MONTH:
+        base_year = base_year - 1 if date.month == _MIN_MONTH else base_year
+        base_month = _MAX_MONTH if date.month == _MIN_MONTH else base_month - 1
```
pradh, line 136: Would something like dateutil (which we already use) make it easier without scope for off-by-ones?  https://pypi.org/project/python-dateutil/

pradh (approved):
Thanks!

server/lib/nl/detection/date.py
pradh: I was puzzled by the `_` in `last_years`, but later realized these are sentinels you use to indicate things rather than anything else, right?  (Perhaps an enum for these might be more readable)

chejennifer: ah yes, this is just a placeholder to use as the "prep" for all the dates that are last x years. This is the only case of using a placeholder prep, so maybe I can make that more clear in the comments & can switch to an enum if there are more cases of this

server/lib/nl/detection/date.py
chejennifer: done

server/lib/nl/detection/date.py
chejennifer: good idea, updated!

pradh (approved):

chejennifer:
thanks for the reviews!

PR #3889, opened 2024-01-17
[vis tool] fix source selector on timeline

chejennifer:
This fixes the source selector for the timelines on vis tool (https://screenshot.googleplex.com/kX3735zMTDF7uFR). The problem was that the endpoint for fetching the facets enforced strict slashes, but the client was not passing in a slash and so the client was being redirected to a location it could not access.

### DIFF
--- a/server/routes/shared_api/facets.py
+++ b/server/routes/shared_api/facets.py
@@ -62,7 +62,7 @@ def get_facets_within():
                                   True)
 
 
-@bp.route('/')
+@bp.route('/', strict_slashes=False)
 def get_facets():
   """Gets the available facets for a list of stat vars for a list of places.
   """


shifucun (commented):
should we do the same thing for other endpoints?

chejennifer:
> should we do the same thing for other endpoints?

I did a quick search through the website repo and the other cases where the endpoint ends in a '/' are all routes for a page (e.g., /browser/) & those redirects are ok. The problem only occurs when the client tries to fetch data from an endpoint that redirects.

dwnoble (approved):

juliawu (approved):
Thanks for the fix!

chejennifer:
thanks for the reviews!

PR #3891, opened 2024-01-18
[Place Page SEO] Add canonical link to header

juliawu:
Adds a [canonical link](https://developers.google.com/search/docs/crawling-indexing/consolidate-duplicate-urls#rel-canonical-link-method) to the header of our place pages, following the [guidance to consolidate duplicate URLs](https://developers.google.com/search/docs/crawling-indexing/consolidate-duplicate-urls). This change will help us fix "Duplicate without user-selected canonical" indexing errors. The canonical urls were chosen to match the entries in our sitemaps.

### DIFF
--- a/server/templates/place.html
+++ b/server/templates/place.html
@@ -21,14 +21,16 @@ limitations under the License.
 {% set subpage_title = _('Place Explorer') %}
 {% set subpage_url = url_for('place.place') %}
 {% set place_category = category %}
+{% set canonical_base_url = url_for('place.place', place_dcid=place_dcid) %}
+{% set canonical_category_url = url_for('place.place', place_dcid=place_dcid, category=category) %}
 
 {% block head %}
 <link rel="stylesheet" href={{url_for('static', filename='css/place_page.min.css' , t=config['VERSION'])}}>
 <link rel="stylesheet" href="https://fonts.googleapis.com/css2?family=Material+Symbols+Outlined" />
 {% if category == '' %}
-<link rel="canonical" href="https://{{ request.host }}/place/{{ place_dcid }}" />
+<link rel="canonical" href="https://{{ request.host }}{{ canonical_base_url }}" />
 {% elif category != '' %}
-<link rel="canonical" href="https://{{ request.host }}/place/{{ place_dcid }}?category={{ category }}" />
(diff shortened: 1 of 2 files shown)


dwnoble:
Do the URLs that appear in the browser address bar ever differ from the canonical URLs?

server/templates/place.html
```diff
@@ -25,6 +25,11 @@
 {% block head %}
 <link rel="stylesheet" href={{url_for('static', filename='css/place_page.min.css' , t=config['VERSION'])}}>
 <link rel="stylesheet" href="https://fonts.googleapis.com/css2?family=Material+Symbols+Outlined" />
+{% if category == '' %}
+<link rel="canonical" href="https://datacommons.org/place/{{ place_dcid }}" />
```
dwnoble, line 29: Should we use the site's actual hostname here (in case this is a custom DC instance)

juliawu: Great callout. Switched to using actual hostname.

juliawu:
> Do the URLs that appear in the browser address bar ever differ from the canonical URLs?

Yes, for example, going to `/place/geoId/06/` with trailing slash stays in the browser address, but the canonical url according to our sitemaps is without trailing slash. Is this a problem?

dwnoble:
> > Do the URLs that appear in the browser address bar ever differ from the canonical URLs?
> 
> Yes, for example, going to `/place/geoId/06/` with trailing slash stays in the browser address, but the canonical url according to our sitemaps is without trailing slash. Is this a problem?

Got it-that makes sense that the trailing slash could trigger the crawler to complain about duplicate pages.

But, reading the [duplicate url guidance](https://developers.google.com/search/docs/crawling-indexing/consolidate-duplicate-urls), it seems like the intent of the `canonical` meta tag is if we had multiple URL schemes for reaching the same page. For example, if say  https://datacommons.org/place/geoId/06 and https://datacommons.org/place/california both sent you to the same page.

For the trailing slash case, it sounds like a [301 redirect](https://developers.google.com/search/docs/crawling-indexing/consolidate-duplicate-urls#redirects-method) might be the better solution. WDYT?

dwnoble (approved):

dwnoble:
Approving to test canonical links for this week's release. We can do a follow-up test next release with the 301 redirects.

PR #3892, opened 2024-01-19
[nodejs] Update chart image response

chejennifer:
- increase the size of the png image of the chart to improve image quality when consumed by Bard
- add an option to return the chart as an svg to be used by Bard to test consuming our charts as an svg instead of a png

### DIFF
--- a/static/src/server.ts
+++ b/static/src/server.ts
@@ -104,6 +104,8 @@ const BARD_ALLOWED_CHARTS = new Set(["LINE", "BAR", "RANKING", "SCATTER"]);
 // The root to use to form the dc link in the tile results
 // TODO: update this to use bard.datacommons.org
 const DC_URL_ROOT = "https://datacommons.org/explore#q=";
+// Size of the PNG to return for the chart query
+const PNG_WIDTH = 1600;
 
 const dom = new JSDOM(
   `<html><body><div id="dom-id" style="width:500px"></div></body></html>`,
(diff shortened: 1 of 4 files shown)


static/src/server.ts
```diff
@@ -105,6 +104,8 @@ const BARD_ALLOWED_CHARTS = new Set(["LINE", "BAR", "RANKING", "SCATTER"]);
 // The root to use to form the dc link in the tile results
 // TODO: update this to use bard.datacommons.org
 const DC_URL_ROOT = "https://datacommons.org/explore#q=";
+// Size of the PNG to return for the chart query
+const PNG_WIDTH = 1600;
```
dwnoble, line 108: Would it make sense to return SVGs at the same width (maybe set this in the SVG chart itself?) Now SVGs get returned with a 550px width and pngs with 1600

dwnoble (commented):
Nice!

The SVG charts look fine in chrome, but in safari & the mac OS preview (when you press "spacebar" when the file is selected in finder) the title gets cut off
Screenshot 2024-01-22 at 10 19 39 AM

static/src/server.ts
chejennifer: The reason for returning pngs at 1600 is to just use a really large size to preserve image quality no matter what size Bard actually chooses to render the image at. I tried drawing the actual svg at 1600 and the scaling makes everything super hard to see. Instead, I updated the returned svg to use viewbox with the original dimensions & caller can then update the width and height to whatever they need.

chejennifer:
> Nice!
> 
> The SVG charts look fine in chrome, but in safari & the mac OS preview (when you press "spacebar" when the file is selected in finder) the title gets cut off Screenshot 2024-01-22 at 10 19 39 AM

thanks for the review! updated the svg charts to work in safari as well 
Screenshot 2024-01-22 at 11 28 36 AM

static/src/server.ts
dwnoble: Sounds reasonable

dwnoble (approved):

PR #3894, opened 2024-01-19
[Place Page SEO] Add meta description and update titles

juliawu:
Adds a meta description and updates the titles of our place pages to be more descriptive.

Adding more `<h1>` headers is being left for a future PR. Section headers are currently missing in the 'view source' because they take too long to load, and won't be seen by the crawler. Addressing the latency issue will require a larger fix.

Screenshot of new header:

Overview page:
Screenshot 2024-01-19 at 10 41 16 AM

Category page:
Screenshot 2024-01-19 at 10 41 35 AM


server/templates/place.html
```diff
@@ -32,6 +40,8 @@
 {% elif category != '' %}
 <link rel="canonical" href="https://{{ request.host }}{{ canonical_category_url }}" />
 {% endif %}
+
+<meta name="description" content="{{ description }}" />
```
dwnoble, line 44: Can you use the `place_summary` here if it's available, otherwise default to the description

juliawu: Great idea, done!

server/templates/place.html
```diff
@@ -17,7 +17,15 @@
 
 {% set main_id = 'dc-places' %}
 {% set page_id = 'page-dc-places' %}
-{% set title = place_name + ' ' + category + ' - ' + _('Place Explorer') %}
+{% set category_title = place_name + ' ' + category %}
```
dwnoble, line 20: did you mean to drop `_('Place Explorer') ` ?

server/templates/place.html
```diff
@@ -17,7 +17,15 @@
...
+{% set overview_title = place_name + ' - Economics, Health, Equity, Crime, Education, Demographics, Housing, Environment' %}
```
dwnoble, line 21: Not every place will have all of these categories (example: no crime here http://localhost:8080/place/geoId/06013 )

dwnoble: can we pull these categories programmatically and translate them (both here and in the overview description below)

dwnoble: Nit: move category_title and overview_title inline below in the if / else statement

dwnoble (commented):
Some follow-up findings from https://github.com/datacommonsorg/website/pull/3891

server/templates/place.html
```diff
@@ -17,7 +17,15 @@
...
+{% if category != '' %}
+  {% set title = category_title %}
+  {% set description = 'Statistics about ' + place_name + ' ' + category %}
```
dwnoble, line 24: Add localization

server/templates/place.html
juliawu: Moved the title logic into the if/else statement.

As discussed offline, pulling the category names is a larger change. The current logic lies downstream and needs to wait for all the data to load before calculating which categories are available. Leaving the current title as-is and title updates can be a future change.

server/templates/place.html
juliawu: Added localization plumbing to the page. Translations won't show up until after the PR goes in and we extract and submit these strings to Google's translation pipeline.

server/templates/place.html
juliawu: Originally intentional to match Adriana's suggestions. Reverted to original title (see discussion below)

server/templates/place.html
juliawu: Update: added localization to the titles, while keeping the title format the same.

server/templates/place.html
```diff
@@ -17,21 +17,27 @@
...
 {% set subpage_url = url_for('place.place') %}
 {% set place_category = category %}
-{% set canonical_base_url = url_for('place.place', place_dcid=place_dcid) %}
-{% set canonical_category_url = url_for('place.place', place_dcid=place_dcid, category=category) %}
 
 {% block head %}
 <link rel="stylesheet" href={{url_for('static', filename='css/place_page.min.css' , t=config['VERSION'])}}>
 <link rel="stylesheet" href="https://fonts.googleapis.com/css2?family=Material+Symbols+Outlined" />
-{% if category == '' %}
-<link rel="canonical" href="https://{{ request.host }}{{ canonical_base_url }}" />
```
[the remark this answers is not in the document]
juliawu: Good point. I've removed the canonical link now, we can go back to add it in if we're still noticing the "duplicate without user selected-canonical" errors.

server/templates/place.html
```diff
@@ -49,7 +55,8 @@
       <div id="title" data-dcid="{{ place_dcid }}">
         <div id="place-heading-container">
           <div id="place-heading">
-            <h1 id="place-name" data-pn="{{ place_name }}">{{ place_name }}</h1>
+            {# TRANSLATORS: Name of the place the page is about. #}
+            <h1 id="place-name" data-pn="{{ place_name }}">{% trans %}{{ place_name }}{% endtrans %}</h1>
```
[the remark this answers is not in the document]
juliawu: Good catch on the place name already being localized. Removed the double localization logic for place name.

According to [this source](https://docs.ckan.org/en/2.9/contributing/string-i18n.html) linked from [our docs](https://github.com/datacommonsorg/website/blob/0bea5bef632f9a8bd8b525af2a887d9f10173480/docs/localization.md), we should be able to wrap variables like so:

Screenshot 2024-01-23 at 12 28 55 PM

server/templates/place.html
```diff
@@ -60,7 +67,8 @@ <h3 id="locale" data-lc="{{ locale }}"></h3>
         </div>
       </div>
       <div id="subtitle" class="row"></div>
-      <div id="place-summary" class="row-col place-summary-container">{{ place_summary or '' }}</div>
+      {# TRANSLATORS: A summary highlight of the information shown on the page. #}
+      <div id="place-summary" class="row-col place-summary-container">{% trans %}{{ place_summary }}{% endtrans %}</div>
```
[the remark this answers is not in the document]
juliawu: Agreed we definitely need more thought on how to handle the localization of the place summaries (and by the same logic, our descriptions). For now, we'll only show the place summaries and descriptions in the "en" locale.

server/templates/place.html
```diff
@@ -17,21 +17,26 @@
...
 {% block head %}
 <link rel="stylesheet" href={{url_for('static', filename='css/place_page.min.css' , t=config['VERSION'])}}>
 <link rel="stylesheet" href="https://fonts.googleapis.com/css2?family=Material+Symbols+Outlined" />
-{% if category == '' %}
-<link rel="canonical" href="https://{{ request.host }}{{ canonical_base_url }}" />
-{% elif category != '' %}
-<link rel="canonical" href="https://{{ request.host }}{{ canonical_category_url }}" />
+
+{% if locale == "en" %}
+<meta name="description" content="{{ place_summary or description }}" />
```
dwnoble, line 37: looks like some redundancy with the above `{% set description = (place_summary or description_text) %}`

server/templates/place.html
```diff
@@ -17,21 +17,26 @@
...
 {% set page_id = 'page-dc-places' %}
-{% set title = place_name + ' ' + category + ' - ' + _('Place Explorer') %}
+{% if category == '' %}
+  {% set title = place_name + ' - ' + _('Place Explorer') %}
+  {% set description = 'Statistics about economics, health, equity, crime, education, demographics, housing, and environment in {place_name}.'.format(place_name=place_name) %}
+{% else %}
+  {% set title = place_name + ' - '+ _('CHART_TITLE-CHART_CATEGORY-{category}'.format(category=category)) + ' - ' + _('Place Explorer') %}
+  {% set description = 'Statistics about {category} in {place_name}.'.format(category=category, place_name=place_name) %}
+{% endif %}
+{% set description = (place_summary or description_text) %}
```
dwnoble, line 27: `description_text` looks undefined

server/templates/place.html
```diff
@@ -17,21 +17,26 @@
 
 {% set main_id = 'dc-places' %}
 {% set page_id = 'page-dc-places' %}
-{% set title = place_name + ' ' + category + ' - ' + _('Place Explorer') %}
+{% if category == '' %}
+  {% set title = place_name + ' - ' + _('Place Explorer') %}
+  {% set description = 'Statistics about economics, health, equity, crime, education, demographics, housing, and environment in {place_name}.'.format(place_name=place_name) %}
```
dwnoble, line 22: Wrap with `_` for these like described here: https://docs.ckan.org/en/2.9/contributing/string-i18n.html#internationalizating-strings-in-jinja2-templates - then you can drop the `{% if locale == "en" %}` check below.

Also might be worth running through https://github.com/datacommonsorg/website/blob/master/docs/localization.md to ensure the new strings are getting extracted properly (you dont have to commit the new extracted strings to the repo, i think itd be good from a sanity check)

server/templates/place.html
```diff
@@ -17,21 +17,26 @@
...
+{% else %}
+  {% set title = place_name + ' - '+ _('CHART_TITLE-CHART_CATEGORY-{category}'.format(category=category)) + ' - ' + _('Place Explorer') %}
+  {% set description = 'Statistics about {category} in {place_name}.'.format(category=category, place_name=place_name) %}
```
dwnoble, line 25: Wrap with `_`

server/templates/place.html
```diff
@@ -60,7 +65,7 @@ <h3 id="locale" data-lc="{{ locale }}"></h3>
         </div>
       </div>
       <div id="subtitle" class="row"></div>
-      <div id="place-summary" class="row-col place-summary-container">{{ place_summary or '' }}</div>
+      <div id="place-summary" class="row-col place-summary-container">{{ place_summary if locale == "en" }}</div>
```
dwnoble, line 68: Nit: can you move the locale check to `place/html.py`? i think eventually that's where the place summary localization logic will live, so might make sense to move the "en" check there now. wdyt?

server/templates/place.html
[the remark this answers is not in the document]
dwnoble: Good callout here with being able to wrap variables in trans text. Looking more into it, it looks like the text would then be sent to a translator as "Hello {{ user }}", so the translators job would be to translate the word "Hello" only and move the variable around as needed so it makes sense in the context of that language.

server/templates/place.html
juliawu: Good catch! Deduped.

server/templates/place.html
juliawu: Good catch! Fixed.

server/templates/place.html
juliawu: Wrapped. I left the locale check because we don't have translations for the descriptions ready yet, otherwise the descriptions for non-english locales will be in english.

I tried the frist step in the localization dogs, but the scripts used now have some errors because they're so old and some of the methods they call are depreciated now. Will look into fixing that after we get this PR in.

server/templates/place.html
juliawu: Wrapped!

server/templates/place.html
juliawu: Good idea. Moved to html.py

server/templates/place.html
```diff
@@ -17,21 +17,25 @@
...
 
 {% block head %}
 <link rel="stylesheet" href={{url_for('static', filename='css/place_page.min.css' , t=config['VERSION'])}}>
 <link rel="stylesheet" href="https://fonts.googleapis.com/css2?family=Material+Symbols+Outlined" />
-{% if category == '' %}
-<link rel="canonical" href="https://{{ request.host }}{{ canonical_base_url }}" />
-{% elif category != '' %}
-<link rel="canonical" href="https://{{ request.host }}{{ canonical_category_url }}" />
+
+{% if locale == "en" %}
```
dwnoble, line 35: do we still need this check?

dwnoble (approved):

server/templates/place.html
juliawu: Yes, right now we don't have translations for the descriptions, and getting those translations will take some time. The description string defaults to the original text when a translation is not available, so we'd end up with english descriptions for non-english pages in the meantime.

dwnoble: That makes sense- thank you!

PR #3898, opened 2024-01-22
[Place Page SEO] Add new sitemap for testing

juliawu:
Add a new sitemap consisting of 50 US States, Washington DC, and the top 100 US cities by population. This sitemap can be used as a "test set" for our SEO changes, and allows for us to more easily request re-indexing and track indexing over time for just this priority set of places.

This PR contains both the new sitemap, an updated robots.txt, and the code used to generate the sitemap.


tools/sitemap/main.py
```diff
@@ -87,6 +88,51 @@ def write_place_url(place_type):
...
 
 
+def write_priority_places_sitemap():
+  """Write a custom sitemap for SEO testing.
+  
+  Writes a sitemap with 50 US states, Washington D.C., and the top 100
+  US cities by population.
+  """
+  # Get US states and Washington DC
+  sparql = '''
```
shifucun, line 98: curious why not fetch State the same way as City, then can merge logic for both of them

static/sitemap/PriorityPlaces.0.txt
```diff
@@ -0,0 +1,151 @@
+https://datacommons.org/place/geoId/01
```
shifucun, line 1: The states are already covered in State.0.txt, I wonder if we need a duplicated sitemap, or could that be an issue for search engine.

juliawu: From what I understand having the duplicates shouldn't break anything, but also won't increase crawl rate (having the link twice will not make googlebot crawl it twice).

Having a dedicated sitemap is still helpful for us because the recommended way to request re-crawling for a batch of URLs is to put them in a sitemap. Otherwise, we'd have to manually request re-crawling for each of the 100 top cities. (The cities sitemap has 45,000+ entries, more than we would want for a test set.)

tools/sitemap/main.py
juliawu: The ranking call for States will also return places like Puerto Rico, which has multiple dcids and place types. I felt it was easier to match the way the other sitemaps are built instead of filtering the ranking call.

Should I switch to using a call to /api/ranking for State to match City?

static/sitemap/PriorityPlaces.0.txt
shifucun: That makes sense.

tools/sitemap/main.py
shifucun: This works for now. If you add more changes, worth to consolidate.

shifucun (approved):

PR #3901, opened 2024-01-22
fix webdriver tests

chejennifer:
Fix webdriver tests and improve source selector checking by:
- picking the source by source name
- checking that the source has in fact changed in the chart

Also update NL goldens. The diffs all have to do with new commute data that was recently imported which caused svgs to get dropped from consideration when fetching related topics.

### DIFF
--- a/server/integration_tests/test_data/e2e_answer_places/howabouttheuninsuredpopulation/chart_config.json
+++ b/server/integration_tests/test_data/e2e_answer_places/howabouttheuninsuredpopulation/chart_config.json
@@ -721,7 +721,6 @@
       "Count_Person_Female_NoHealthInsurance": {
         "age": [
           "Count_Person_18To64Years_Female_NoHealthInsurance",
-          "Count_Person_21To64Years_Female_NoHealthInsurance",
           "dc/001qmh9sdqzeb",
           "dc/0flj62fpwwxd",
           "dc/bhlcb3wz4b3v2",
@@ -738,8 +737,6 @@
       },
       "Count_Person_Male_NoHealthInsurance": {
         "age": [
-          "Count_Person_18To64Years_Male_NoHealthInsurance",
-          "Count_Person_21To64Years_Male_NoHealthInsurance",
           "dc/1qdvp8x0t3l11",
           "dc/260gxgmeglpb",
           "dc/33z5t6xqjepp1",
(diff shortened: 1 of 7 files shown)


shifucun (approved):

dwnoble (approved):

server/integration_tests/test_data/fulfillment_api_comparison/chart_config.json
```diff
@@ -394,12 +394,9 @@
     "exploreMore": {
       "dc/0gettc3bc60cb": {
         "hasOccupation": [
-          "dc/b9plt3q5k6gyb",
```
chejennifer, line 397: the dropped svs are commute related (dropped because there is a limit to number of svgs that get opened and new commute data has caused these svs to not be seen)

server/integration_tests/test_data/fulfillment_api_correlation/chart_config.json
```diff
@@ -814,48 +814,7 @@
         ]
       }
     ],
-    "exploreMore": {
```
chejennifer, line 817: the dropped svs are commute related (dropped because there is a limit to number of svgs that get opened and new commute data has caused these svs to not be seen)

server/integration_tests/test_data/e2e_answer_places/howabouttheuninsuredpopulation/chart_config.json
```diff
@@ -721,7 +721,6 @@
       "Count_Person_Female_NoHealthInsurance": {
         "age": [
           "Count_Person_18To64Years_Female_NoHealthInsurance",
-          "Count_Person_21To64Years_Female_NoHealthInsurance",
```
chejennifer, line 724: Although these dropped svs are not themselves commute related, these dropped svs still have to do with the commute import because: 

- the svgs that these svs are part of were not expanded because new svgs part of the commute import were expanded instead (limit of 50 svgs being expanded at each level): 
Screenshot 2024-01-23 at 12 04 51 AM

- there is an increase in svs that have to do with income of person age 16+ with 25 in prod (https://screenshot.googleplex.com/Bg23G4mDVFJxeQz ) and 57 in autopush (https://screenshot.googleplex.com/zUCtttk5HtPVxyZ)
- svs like “Count_Person_16OrMoreYears_NoHealthInsurance_IncomeOf15000To24999USDollar_WithEarnings” have svgs in their memberOf like “dc/g/Person_Age-Years16Onwards_Income-USDollar15000To24999” which open up to svgs like “dc/g/Person_Age-Years16Onwards_CommuteMode_Income-USDollar15000To24999” (new commute svgs)

PR #3903, opened 2024-01-24
[Place Page SEO] read text summaries from GCS directly

juliawu:
This PR loads summary text on the place pages by making a read into the GCS bucket in the place page endpoint, instead of loading them all to the server app config on server load. This change means we won't have to restart the prod server each time we make a change to the content of the summaries.


server/routes/place/html.py
```diff
@@ -81,8 +82,7 @@ def place(place_dcid=None):
   else:
     place_name = place_dcid
 
-  place_summary = current_app.config['PLACE_EXPLORER_SUMMARIES'].get(
-      place_dcid, {'summary': ''})
+  place_summary = place_summaries.get_place_summaries().get(place_dcid, "")
```
shifucun, line 85: can you add a logging to time this function call?

shifucun (approved):

server/routes/place/html.py
juliawu: Added logging!

PR #3904, opened 2024-01-25
[nl] Clean up how existence results get passed around

chejennifer:
Move storing sv->place->facets (comes from existence check) from ChartVars to PopulateState and use PopulateState to set sv_place_facet_ids in ChartSpec to be used by chart builder. This is more inline with the purposes of each of the ChartVars, PopulateState, and ChartSpec structures


server/lib/nl/common/rank_utils.py
```diff
@@ -82,9 +83,7 @@ def rank_places_by_series_growth(
     counters: ctr.Counters,
     place_type: str = '',
     min_population: int = 0,
-    date_range: types.Date = None,
-    sv_exist_facet: Dict[str, Dict[str, Dict[str, str]]] = None
-) -> GrowthRankedLists:
+    date_range: types.Date = None) -> (GrowthRankedLists, Dict[str, str]):
```
pradh, line 86: Document ret value?

server/lib/nl/config_builder/timeline.py
```diff
@@ -108,7 +107,9 @@ def single_place_single_var_timeline_block(column, place: Place, sv_dcid: str,
                                            sv2thing: types.SV2Thing,
                                            single_date: types.Date,
                                            date_range: types.Date,
-                                           cv: ChartVars):
+                                           sv_place_facet_id: Dict[str,
+                                                                   Dict[str,
+                                                                        str]]):
```
pradh, line 112: Should we use a dataclass for this since it gets passed around a lot?

server/lib/nl/fulfillment/utils.py
```diff
@@ -213,11 +217,24 @@ def is_coplottable(chart_vars: ChartVars, places: List[Place]) -> bool:
...
+# the results of an existence check
+def get_sv_place_facet_ids(svs: List[str], places: List[Place],
+                           exist_checks: Dict[str, Dict[str, ExistInfo]]):
+  sv_place_facet_ids = {}
+  for sv in svs:
+    sv_place_facet_ids[sv] = {}
+    for pl in places:
+      sv_place_facet_ids[sv][pl.dcid] = exist_checks.get(sv, {}).get(
+          pl.dcid, ExistInfo()).facet.get('facetId', '')
+  return sv_place_facet_ids
```
pradh, line 240: Can this be in https://github.com/datacommonsorg/website/blob/e48cbfa692be3b6d458f2938dbab74490ef27f8a/server/lib/explore/existence.py#L4?

server/lib/nl/fulfillment/types.py
```diff
@@ -85,6 +80,12 @@ class SV2Thing:
   footnote: Dict
 
 
+@dataclass
+class ExistInfo:
+  is_single_point: bool = False
+  facet: Dict[str, str] = field(default_factory=dict)
```
pradh, line 86: nit: document key-values?

server/lib/nl/fulfillment/existence.py
```diff
@@ -62,21 +63,51 @@ def __init__(self, state: PopulateState, place2keys: Dict):
     # has data for that SV.
     self.existing_svs = {}
 
+  def _get_sv_place_facets(self):
```
pradh, line 66: Maybe add descriptive summary of what the function does, esp the computing the facets with max occurrences?

server/integration_tests/test_data/e2e_date_range/whichcountriesinafricahavehadthegreatestincreaseinelectricityaccessoverthelast10years/chart_config.json
```diff
@@ -10,20 +10,23 @@
...
+                      "country/KEN",
+                      "country/RWA",
+                      "country/ETH",
+                      "country/UGA",
+                      "country/TZA",
+                      "country/MLI",
+                      "country/ZMB",
+                      "country/SDN",
+                      "country/ZWE",
+                      "country/AGO"
```
pradh, line 22: Is this change from the fact that we use facet-ids from within the time-delta handlers?

pradh (commented):
Very nice!  Thanks for the change!

server/lib/nl/common/rank_utils.py
chejennifer: done

server/lib/nl/config_builder/timeline.py
chejennifer: I'm not sure how this dataclass will look because there aren't really fields we're accessing, we just want the map of sv -> place -> facetid .. 

One thought is that I could pass in ChartSpec instead and access chartspec.sv_place_facet_id to make things more clear? wdyt?

server/lib/nl/fulfillment/utils.py
chejennifer: good idea, done

server/lib/nl/fulfillment/types.py
chejennifer: done

server/lib/nl/fulfillment/existence.py
chejennifer: done

server/lib/nl/config_builder/timeline.py
pradh: I see.  In C++ to make this case more readable we'd just use typedef. So should we try:

`Place2Facet = Dict[str, str]`
`Sv2Place2Facet = Dict[str, Place2Facet]`

So just replace the type annotations with `Sv2Place2Facet`, impl doesn't change.

chejennifer: oh, super cool, didn't know that worked in python, done!

pradh (approved):
Thanks for the changes!

PR #3907, opened 2024-01-25
[Place page SEO] Only show summaries on main DC, Reduce latency in fetch logic

juliawu:
This PR makes the following changes:

* Only read the GCS bucket if we're in one of the main flask environments. This will prevent our custom dc's from trying to read a file they don't have access to.

* removes the allow list. We can directly remove summaries from the main json if we wish to hide them.

* Have server store the blob to the place summary file on GCS. This reduces the latency on the summary text fetch.

* Moves the logic in `place_summaries.py` to the place page's file. The logic there is now short enough not to need its own dedicated page.


server/routes/place/html.py
```diff
@@ -32,6 +33,19 @@
 }
 
 
+def get_place_summaries() -> dict:
+  """Load place summary content from GCS"""
+  try:
+    tmp_path = os.path.join('/tmp', PLACE_SUMMARY_CONTENT_FILENAME)
+    current_app.config['PLACE_SUMMARY_BLOB'].download_to_filename(tmp_path)
```
shifucun, line 40: This may have race condition, when two client requests in the same time and both write to the file. If the whole setup works, later on can change this to read bytes directly.

shifucun (approved):
very nice!

juliawu:
Superceded by https://github.com/datacommonsorg/website/pull/3908

PR #3909, opened 2024-01-26
[NL] Add Plumbing for per chart messaging - server

chejennifer:
These are the server side changes to add plumbing for per chart messaging:
- add new field in Block proto for taking an info message
- add logic to pass a user message from populate handler to the builder to add to the block config. If all blocks have the same user message, pass the message to page level instead
- update page level 'user message' to take a list of messages instead of just a single message
- implement user message for the comparison handler

client side changes are in a separate PR: https://github.com/datacommonsorg/website/pull/3910

block level messages:
Screenshot 2024-01-26 at 11 45 40 AM

page level message that comes from every block having the same message:
Screenshot 2024-01-26 at 11 51 26 AM

when there was already a page level message, but every block has the same message:
Screenshot 2024-01-26 at 11 51 43 AM


server/lib/nl/config_builder/builder.py
```diff
@@ -40,10 +40,11 @@
 
 #
 # Given an Utterance, build the final Chart config proto.
+# Returns the chart config proto and user message (empty if no message to show).
 #
-def build(state: PopulateState, config: Config) -> SubjectPageConfig:
+def build(state: PopulateState, config: Config) -> (SubjectPageConfig, str):
```
pradh, line 45: nit: maybe return a struct?

server/lib/nl/config_builder/builder.py
```diff
@@ -73,11 +74,24 @@ def build(state: PopulateState, config: Config) -> SubjectPageConfig:
...
 
   builder = base.Builder(uttr, sv2thing, config)
+  user_message = ''
+  for idx, cspec in enumerate(uttr.rankedCharts):
+    cspec = cast(ChartSpec, cspec)
+    if idx == 0:
+      user_message = cspec.info_message
+    elif cspec.info_message != user_message:
+      user_message = ''
+      break
```
pradh, line 84: nit: _get_page_level_msg helper?

server/lib/nl/config_builder/builder.py
```diff
@@ -73,11 +74,24 @@ def build(state: PopulateState, config: Config) -> SubjectPageConfig:
...
 
   # Build chart blocks
   for cspec in uttr.rankedCharts:
     cspec = cast(ChartSpec, cspec)
     cv = cspec.chart_vars
+    # if there is a user message, that means every chart spec had the same info
+    # message and we should show the message at the page level instead of block
+    # level.
+    if bool(user_message):
```
pradh, line 93: Curious what the `bool()` does over just `user_message`?

server/lib/nl/fulfillment/comparison.py
```diff
@@ -24,6 +24,36 @@
...
+_MISSING_PLACE_MSG = 'Data for "{missing_places}" is currently unavailable. Nonetheless, take a look at the following insights on "{exist_places}"'
+
+
+def _get_place_list_str(places: List[str]):
+  if len(places) < 1:
+    return ''
+  elif len(places) == 1:
+    return places[0]
+  else:
+    return ', '.join(places[0:len(places) - 1]) + f' and {places[-1]}'
```
pradh, line 36: could this maybe move to a util?

pradh (commented):
Nice!

server/integration_tests/test_data/demo2_cities_feb2023/query_7/chart_config.json
```diff
@@ -136,6 +136,7 @@
               }
             ],
             "denom": "Count_Person",
+            "infoMessage": "Data for \"South Sudan\" is currently unavailable. Nonetheless, take a look at the following insights on \"Algeria, Morocco, Libya, Tunisia, Mali, Eritrea, Malawi, Senegal and Gambia\"",
```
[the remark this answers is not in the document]
chejennifer: That's a good point, and there isn't really a way to know at the point of message generation what the order of the places will be.. This came from Adriana, so I will get her opinions on this!

server/lib/nl/config_builder/builder.py
chejennifer: done

server/lib/nl/config_builder/builder.py
chejennifer: done

server/lib/nl/config_builder/builder.py
chejennifer: I was using it to return false if user_message is empty, but an empty string itself will return false, so removed the `bool()`

server/lib/nl/fulfillment/comparison.py
chejennifer: done

static/js/apps/explore/app.tsx
```diff
@@ -229,7 +229,7 @@ export function App(props: { isDemo: boolean }): JSX.Element {
       }
     }
     const userMessage = {
-      msg: fulfillData["userMessage"] || "",
+      msg: !_.isEmpty(fulfillData["userMessages"]) ? fulfillData["userMessages"][0] : "",
```
chejennifer, line 232: this is a temporary change to make sure UI doesn't break if this PR gets merged before the UI changes (it will be updated by the UI changes in a separate PR).

server/integration_tests/test_data/demo2_cities_feb2023/query_7/chart_config.json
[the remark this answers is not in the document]
chejennifer: @miss-o-soup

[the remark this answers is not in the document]
miss-o-soup: Hi there,
I believe the comment relates to this example
8PMQ8Yupv2koDgQ
If so, the intention is to give context in one line of what is available or not from your original  query. Meaning you don’t have to compare and do the string match/math between that message and the search box.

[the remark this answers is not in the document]
pradh: Should we then do that only for the top-level msg, not per chart message?  In the per chart case, it seems crowded, especially for time-delta queries.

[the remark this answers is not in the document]
pradh: Hmm actually, even for this case, the bar charts prominently have those existing place names?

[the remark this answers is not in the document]
miss-o-soup: mmm.... I think I see that different. What I'm concerned about is that just saying something doesn't exist, doesn't fully paint a picture on what is presented next (the user then has to do the logic in his head once he readd the next/plot). It's an incomplete message. Having said that, to alleviate for the crowded part that you are concerned about - What about this?:

Data for “[places-with-no-data]” is currently unavailable. Nonetheless, take a look at the following insights on the other countries.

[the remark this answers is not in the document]
pradh: I see, ok, that sounds good!  

nit:  maybe we can use slightly fewer words + similar language [as we do now](https://github.com/datacommonsorg/website/blob/c2eb71f3ad6d30a059553f0a59cebef37c5c90c0/server/lib/nl/common/commentary.py#L35-L40)?

    "Data unavailable for '[places-with-no-data]', but see statistics for the other places."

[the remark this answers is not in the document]
miss-o-soup: Yeah, what about the following - we can take out the "Nonetheless" and some other words to keep it short, but still warm sounding:

Data for “[places-with-no-data]” is currently unavailable. See the following statistics on the other places.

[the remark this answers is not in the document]
pradh: sounds good!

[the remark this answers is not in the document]
chejennifer: updated!

server/lib/nl/fulfillment/comparison.py
```diff
@@ -23,17 +23,9 @@
...
-
-
-def _get_place_list_str(places: List[str]):
-  if len(places) < 1:
-    return ''
-  elif len(places) == 1:
-    return places[0]
-  else:
-    return ', '.join(places[0:len(places) - 1]) + f' and {places[-1]}'
+_MISSING_PLACE_MSG = 'Data for "{missing_places}" is currently unavailable. See the following statistics on the other places'
```
pradh, line 28: nit: Should we move this to `commentary.py`, so we kinda have all the msgs in one place in case we want to reuse?

static/js/apps/explore/app.tsx
```diff
@@ -229,7 +229,9 @@ export function App(props: { isDemo: boolean }): JSX.Element {
       }
     }
     const userMessage = {
-      msg: fulfillData["userMessage"] || "",
+      msg: !_.isEmpty(fulfillData["userMessages"])
+        ? fulfillData["userMessages"][0]
```
pradh, line 233: maybe comment why we use first one?

pradh (approved):
Thanks for the changes!

server/lib/nl/fulfillment/comparison.py
chejennifer: good idea, done

static/js/apps/explore/app.tsx
chejennifer: this change will be overwritten when https://github.com/datacommonsorg/website/pull/3910 goes in. This is just a temporary change so that there isn't any breakage in between the two PRs hitting autopush

pradh: ah, makes sense.

chejennifer:
thanks for the reviews!

PR #3912, opened 2024-01-27
[Place Page SEO] Check-in place page summary generation script

juliawu:
Adds the scripts used to generate the summary text shown on our place pages.

There are two scripts and two sets of variables and text provided:

(1) A "ranking based" summary text, where sentences are written for variables a place ranks highly in. This is the approach currently used for the summaries we have up. This includes the following files:
* `build_ranking_based_summaries.py`
* `stat_vars_detailed.json`
* `place_summary_content_us_states.json` 

(2) A templated summary, where we write sentences about the same set of stat vars for every place. These summaries are intended to be used as starting draft to feed into LLMs to rewrite into a final draft.
* `build_template_summaries.py`
* `stat_vars_to_highlight.json`
* `priority_places.json`
* `place_summaries_from_template.json`


tools/summaries/build_template_summaries.py
```diff
@@ -0,0 +1,190 @@
...
+  summaries = {
+      place_dcid: {
+          "summary": " ".join(sentence_list)
+      } for place_dcid, sentence_list in sentences.items()
+  }
+
+  return summaries
+
+
+def main():
```
dwnoble, line 168: ditto here

tools/summaries/build_ranking_based_summaries.py
```diff
@@ -0,0 +1,216 @@
...
+  for place, sentence_list in sentences.items():
+    summaries[place] = {"summary": " ".join(sentence_list)}
+
+  # Write summaries to file
+  with open(output_file, "w") as out_file:
+    json.dump(summaries, out_file, indent=4)
+
+
+def main():
+  build_ranking_based_summaries(place_type="State",
```
dwnoble, line 209: For a future PR: take these in as command line arguments (look into using [click](https://click.palletsprojects.com/en/8.1.x/))

dwnoble (approved):
Nice!

tools/summaries/build_ranking_based_summaries.py
juliawu: Thanks for the pointer! Will add in next PR.

PR #3913, opened 2024-01-29
Apply best practices to base custom dc docker image.

keyurva:
* Applied some of the [best practices][practices] for building docker images for the base custom dc docker image.
* Although the overall image size did not change by much, it reduced number of layers with non-zero size in the base image from 18 to 1. Since the base image won't change often, collapsing it into a single layer should help in the future.
* Ran cloud build and pushed the new image to the [cloud directory][directory].
  + Will use this image with the custom DC docker image locally and verify it works before pushing the latter to the cloud.

[practices]: https://www.ecloudcontrol.com/best-practices-to-reduce-docker-images-size
[directory]: gcr.io/datcom-ci/python-node-go-protoc-envoy

### DIFF
--- a/build/python_node_go_protoc_envoy/Dockerfile
+++ b/build/python_node_go_protoc_envoy/Dockerfile
@@ -29,7 +29,7 @@ ENV GO111MODULE=on
 ENV PATH="${PATH}:/root/go/bin"
 
 RUN apt update \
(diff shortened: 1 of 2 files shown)


shifucun (approved):
very nice!

keyurva:
Thanks for the review!

PR #3914, opened 2024-01-29
[nodejs] use compression to shorten chartUrl

chejennifer:
Sometimes chartUrl that is returned in the response is too long to send a GET request to. Use compression to shorten the URL

For example, "Which industries have grown the most in California" chartUrls:
current (5810 characters): https://screenshot.googleplex.com/4WUDYVS5N2MQsVj
with compression (1513 characters): https://screenshot.googleplex.com/8HAWVmq7Yef9BQE

### DIFF
--- a/static/nodejs_server/utils.ts
+++ b/static/nodejs_server/utils.ts
@@ -30,7 +30,7 @@ import {
   TileConfig,
 } from "../js/types/subject_page_proto_types";
 import {
-  CHART_URL_PARAMS,
+  CHART_PARAMS,
   COMPRESSED_VAL_ENCODING,
   FONT_FAMILY,
   FONT_SIZE,
@@ -38,6 +38,7 @@ import {
   SVG_PADDING,
   SVG_WIDTH,
 } from "./constants";
+import { ChartProps } from "./types";
 
 /**
  * Gets a list of source objects with name and url from a set of source urls.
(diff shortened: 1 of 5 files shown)


static/nodejs_server/utils.ts
```diff
@@ -139,9 +148,13 @@ export function getChartUrl(
       if (!paramVal) {
         return;
       }
```
pradh, line 150: So, we're compressing each value independently?  Is there value in compressing stuff like place, enclosed_place_type.

Or how about just compressing the whole thing in one shot and passing `arg=<base64-encoded-stuff>`?

static/nodejs_server/utils.ts
```diff
@@ -139,9 +148,13 @@ export function getChartUrl(
...
-      url += `${idx === 0 ? "" : "&"}${paramKey}=${paramVal}`;
+      let compressedVal = zlib.deflateSync(paramVal).toString("base64");
```
pradh, line 151: How about having a helper for `compress` and `decompress` that wraps this COMPRESSED_VAL_ENCODING replacement?

Also, maybe have a simple unit-test that compresses, decompresses and confirms they match?

static/src/server.ts
```diff
@@ -616,21 +621,35 @@ app.get("/nodejs/query", (req: Request, res: Response) => {
...
+    let val = req.query[key] as string;
+    // Convert all the manually escaped characters back to their original
+    // characters.
+    for (const c in COMPRESSED_VAL_ENCODING) {
+      val = val.replaceAll(COMPRESSED_VAL_ENCODING[c], c);
+    }
+    // decompress the value
+    val = zlib.inflateSync(Buffer.from(val, "base64")).toString();
+    decompressedReqVals[key] = val;
+  }
```
pradh, line 636: Put in decompress wrapper?

pradh (commented):
Thanks!   Are we not dropping the names?

static/nodejs_server/utils.ts
chejennifer: ah, good idea, updated

static/nodejs_server/utils.ts
chejennifer: done

static/src/server.ts

chejennifer:
> Thanks! Are we not dropping the names?

Thanks for the review! I saw that just by using compression, the URLs are shortened by a lot, so wanted to get this in first and then dropping names can be in a future PR?

pradh (approved):

pradh:
> > Thanks! Are we not dropping the names?
> 
> Thanks for the review! I saw that just by using compression, the URLs are shortened by a lot, so wanted to get this in first and then dropping names can be in a future PR?

Sounds good!  Thanks for the changes and test.

PR #3916, opened 2024-01-30
[Place Page SEO] Add meta tag to prevent indexing on non-prod place pages

juliawu:
This PR adds `<meta name="robots" content="noindex">` as to place pages when not on the main `https://datacommons.org` domain to [prevent search crawlers from indexing](https://developers.google.com/search/docs/crawling-indexing/block-indexing).

This means only the `https://datacommons.org/place/` pages will get indexed, and the crawler will ignore Custom DCs and alternate domains like `browser.datacommons.org` and `dev.datacommons.org`. This will reduce the spurious indexing error alerts for pages that shouldn't be indexed.

### DIFF
--- a/static/package-lock.json
+++ b/static/package-lock.json
@@ -115,7 +115,7 @@
         "sass-loader": "^13.3.2",
         "tslib": "^2.6.2",
         "typescript": "^5.0.2",
-        "vite": "^4.4.5"
+        "vite": "^4.4.12"
       },
       "devDependencies": {
         "@types/lodash": "^4.14.199",
@@ -24716,7 +24716,7 @@
         "sass-loader": "^13.3.2",
         "tslib": "^2.6.2",
         "typescript": "^5.0.2",
-        "vite": "^4.4.5"
+        "vite": "^4.4.12"
       }
     },
     "@discoveryjs/json-ext": {
--- a/static/package-lock.json
+++ b/static/package-lock.json
@@ -115,7 +115,7 @@
         "sass-loader": "^13.3.2",
         "tslib": "^2.6.2",
         "typescript": "^5.0.2",
-        "vite": "^4.4.12"
+        "vite": "^4.4.5"
       },
       "devDependencies": {
         "@types/lodash": "^4.14.199",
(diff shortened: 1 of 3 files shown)


dwnoble:
@juliawu can you verify this works as expected in a test GCP environment? I suspect the base_url might not get set properly when going from the gcp load balancer -> kubernetes cluster

juliawu:
> @juliawu can you verify this works as expected in a test GCP environment? I suspect the base_url might not get set properly when going from the gcp load balancer -> kubernetes cluster

Just tested by deploying to dev.datacommons.org. `base_url` does get set to `http://datacommons.org` (note HTTP instead of HTTPS). Switched the logic to allow both HTTP and HTTPS for now since I'm not sure which base_url will give when deployed to the prod server.

dwnoble:
> > @juliawu can you verify this works as expected in a test GCP environment? I suspect the base_url might not get set properly when going from the gcp load balancer -> kubernetes cluster
> 
> Just tested by deploying to dev.datacommons.org. `base_url` does get set to `http://dev.datacommons.org` (note HTTP instead of HTTPS). Switched the logic to allow both HTTP and HTTPS for now since I'm not sure which base_url will give when deployed to the prod server.

Great! So we can understand what the URL looks like in prod, can you add some sort of logging (maybe another meta tag directly under the noindex tag? `<meta name="dc-debug" content="{url}">`. Also-can you update update the logic to not show the noindex tag in staging (so it properly mimics the prod env)?

static/package-lock.json
dwnoble, line 1: Nit: remove package-lock.json changes

juliawu: Removed!

juliawu:
> > > @juliawu can you verify this works as expected in a test GCP environment? I suspect the base_url might not get set properly when going from the gcp load balancer -> kubernetes cluster
> > 
> > 
> > Just tested by deploying to dev.datacommons.org. `base_url` does get set to `http://dev.datacommons.org` (note HTTP instead of HTTPS). Switched the logic to allow both HTTP and HTTPS for now since I'm not sure which base_url will give when deployed to the prod server.
> 
> Great! So we can understand what the URL looks like in prod, can you add some sort of logging (maybe another meta tag directly under the noindex tag? `<meta name="dc-debug" content="{url}">`. Also-can you update update the logic to not show the noindex tag in staging (so it properly mimics the prod env)?

Good call outs. Added logging statements we can view in GKE logging (instead of adding meta tag end users also see).  Also updated the logic to not show noindex tag in staging.

dwnoble (approved):
Thank you!

PR #3919, opened 2024-01-31
Fix sv index differ info item bug

shifucun:


tools/nl/svindex_differ/differ.py
```diff
@@ -137,8 +137,9 @@ def _get_diff_table(diff_list, base_sv_info, test_sv_info):
       continue
     # If theres no + or -, that means this line is the same in both base and test.
     elif not diff.startswith('+') and not diff.startswith('-'):
-      info = base_sv_info.get(diff_sv, test_sv_info.get(diff_sv, {}))
-      diff_table_rows.append((info, info))
+      base_info = base_sv_info.get(diff_sv, {})
+      test_info = test_sv_info.get(diff_sv, {})
+      diff_table_rows.append((base_info, test_info))
```
chejennifer, line 142: I'm curious what the problem here was because when the diff doesn't start with a + or -, that row shouldn't have any diffs so both infos should be the saem?

chejennifer (commented):
thanks for the update but I'm curious what the issue was

tools/nl/svindex_differ/differ.py
shifucun: From what I see, the diff only look at the sv dcid. But even the same dcid has the same ranking, the score and the sentences matched could be different.

chejennifer: ohh i see, ok, makes sense to me

chejennifer (approved):
thanks for the fix!

PR #3920, opened 2024-02-01
NL Server: When at fault, admit it

pradh:
This fixes the NL server to not suppress faults.  Also, add a real health-check.

Context:
* This is in response to an outage (b/323064085) where a BAD machine caused embeddings to return empty on a small subset of servers, and redis then caches the empty results and always returns it
* With this fix, the website flask cache won't cache the faulty value (if any). To confirm, this [code here](https://github.com/pallets-eco/flask-caching/blob/a367b9f13109d35c1e3df069f5da4c29110a689c/src/flask_caching/__init__.py#L878-L882) in Flash-Caching only caches the return value, not exceptions.

### DIFF
--- a/nl_server/routes.py
+++ b/nl_server/routes.py
@@ -29,7 +29,14 @@ bp = Blueprint('main', __name__, url_prefix='/')
 
 @bp.route('/healthz')
 def healthz():
-  return ""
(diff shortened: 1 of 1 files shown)


shifucun (approved):

chejennifer (approved):
thanks for the fix!

PR #3922, opened 2024-02-01
Mention that you have to enable analytics hub

KilimAnnejaro:
I had to also enable Analytics Hub to make BQ queries work, so noting it.


chejennifer (approved):
thanks for adding this!

PR #3929, opened 2024-02-02
[nl/nodejs] date range updates for bard

chejennifer:
- updated date parser to handle "x years ago" or "decade ago" as a single date
- when there is a single date, update timeline to handle that by highlighting that data point on the chart by showing a single dot there

recording-_1


server/lib/nl/detection/date.py
```diff
@@ -25,6 +25,8 @@
     r'(in|after|on|before|year)(?: year)? (\d{4})',
 ]
 
+YEARS_AGO_RE = [r'(decade) ago', r'(\d+) years ago']
```
juliawu, line 28: Not for implementing in this PR: Do we plan to add support for queries using words instead of digits ("what was the population of California ten years ago")?

juliawu (commented):
Nice! Excited for this change!

static/js/chart/draw_line.ts
```diff
@@ -440,8 +445,11 @@ export function drawLineChart(
     });
     const hasGap = shouldFillInValues(dataset);
     hasFilledInValues = hasFilledInValues || hasGap;
+    // If there is a specific date to be highlighted on the chart, dont' show
```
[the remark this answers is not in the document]
chejennifer: done

static/js/chart/draw_line.ts
```diff
@@ -440,8 +445,11 @@ export function drawLineChart(
...
+    // any other dots.
     const shouldAddDots =
-      dataset.length < MIN_POINTS_FOR_DOTS_ON_LINE_CHART || showAllDots;
+      !options.highlightDate &&
+      (dataset.length < MIN_POINTS_FOR_DOTS_ON_LINE_CHART || showAllDots);
```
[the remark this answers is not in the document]
chejennifer: good point, 

- moved showAllDots to LineChartOptions
- if both showAllDots and highlightDate are set, then we still show the other dots and the highlighted one is just more prominent. If only highlight is set, we only show the highlighted dot. Went with this because".  
    - showAllDots only gets set as true for graph browser and graph browser wont be highlighting any points
    - highlighting dots when there are other dots shown actually still looks ok (https://screenshot.googleplex.com/36WVxaWevYNytww), just less obvious for certain colors (https://screenshot.googleplex.com/8u99pCkxkpcJmwV)

server/lib/nl/detection/date.py
chejennifer: I think we will want to keep improving these & hopefully move away from regex and towards using a model to tell us what dates are in the query

static/js/chart/draw_line.ts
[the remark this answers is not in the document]
juliawu: This looks good!

juliawu (approved):
Thanks for the updates!

chejennifer:
thanks for the review!

pradh:
Will take a look in a few mins.  Also please hold off one merge, since prod push is being triaged and might need cherrypick.

server/lib/nl/detection/date.py
```diff
@@ -103,6 +105,15 @@ def parse_date(query: str, ctr: Counters) -> DateClassificationAttributes:
       dates.append(Date(_LAST_YEARS_PREP, year, year_span=1))
       trigger_strings.append(query[match.start():match.end()])
 
+  for pattern in YEARS_AGO_RE:
```
pradh, line 108: To confirm, can each of these patterns be subset of each other? and thus need an ordered if else check?

pradh, line 108: This func is complicated enough that I might add sprinkle some comments on what each block is handling?

pradh (commented):
Nice!

server/lib/nl/detection/date.py
chejennifer: I don't think these patterns could be subsets of each other? also currently, if a query matches to more than one pattern, we add all the matched dates to the DateClassificationAttributes, but only use the first one in the list. Definitely something we can discuss changing, but probably a separate issue?

chejennifer: done

pradh: I think if we use only one, then we might order them from the longest match to shortest match or something like that.  But, agreed, separate concern...

pradh (approved):
Thanks

PR #3931, opened 2024-02-02
Rollforward healthz and increase timeouts

pradh:
Turns out the prod push failures were because the default health check timeouts (1s) was too small, so increasing it to 5s, and rolling forward the original healthz fix.


shifucun (approved):

pradh:
Thanks!

PR #3935, opened 2024-02-06
Don't lint nested python virtual envs

juliawu:
Updates the lint fix test `run_test.sh -f` to avoid linting files in python envs nested in subdirectories.

The scripts under `tools/summaries` and `tools/sitemap` use their own python `.env` virtual environment which sometimes contain packages that yapf can't parse.


shifucun (approved):
very helpful!

PR #3934, opened 2024-02-06
[Place Page SEO] Add global cities with population at least 500k to priority places sitemap

juliawu:
This PR adds cities around the world with a population of at least 500,000 to our `PriorityPlaces.0.txt` sitemap. This will allow us to more easily evaluate and track indexing status for the place pages corresponding to those cities.

This PR checks in the BQ query used, a csv export of BQ query results, and updates the sitemap generation script to add these populous cities to our sitemap.

PR #3936, opened 2024-02-06
[Place Page SEO] Add summaries for global cities with population >= 500k

juliawu:
This PR builds on top of changes made in #3934 and #3932. 

This PR:

* Adds summaries to the place summary config for all global cities with a population of at least 500,000
* Updates script used to generate those summaries
* Adds sorting by key when outputting by json so future diffs will be easier to parse


tools/summaries/fetch_place_summaries.py
```diff
@@ -237,12 +230,19 @@ def build_template_summaries_for_sitemap(
 @click.option('--batch_size',
               default=_BATCH_SIZE,
               help='how many places to process at once')
-def main(sitemap: str, stat_var_json: str, output_file: str, batch_size: int):
+@click.option('--num_skip_lines',
```
chejennifer, line 233: nit: num_skip_lines was slightly confusing to me at first (in terms of naming), maybe something like 'sitemap_start_idx'

chejennifer (approved):
nice, just a small naming nit

tools/summaries/fetch_place_summaries.py
juliawu: Good idea, renamed to "start index" and updated the help line and comments to clarify.

PR #3911, opened 2024-01-27
Updated stat var hierarchy to accept multiple root node filters.

dwnoble:
* Updated stat var hierarchy to accept multiple root node filters.
* Updated UN staging environment to filter hierarchy by two root nodes

Screenshots of various dc environments

## Base DC
base-dc

## Base Custom DC
base-custom-dc

## UNSDG Staging
unsdg-staging

## UNSDG Staging Map Tool
unsdg-staging-map

## UNSDG 
unsdg

## World Bank
worldbank

## World bank (known issue-counts are off when a place is selected)
worldbank-timeline
worldbank-timeline-ghana

Counts are off in the world bank hierarchy. This looks similar to an existing issue with https://worldbank.datacommons.org - (counts arent updating with place selection). Can address in a future PR.

PR #3937, opened 2024-02-07
Remove stat var hierarchy config from custom dc base.html

juliawu:
A quick fix for a bug in Custom DC. The script in custom/base.html leads to a server error broken page when following the custom DC quick start guide. 

This PR removes the script so that users can follow the custom DC quick start guide. A more complete fix to re-enable features from #3911 will come in a follow-up PR once we determine root cause of the issue.


keyurva (approved):
Thanks Julia!

@premr - once this is in and you sync your repo, your custom dc instance should work.

PR #3941, opened 2024-02-09
Fix website single cluster backendconfig bug; Increase website esp timeout to 60s

shifucun:
With this, timeout of website is 60s everywhere.


keyurva (approved):

pradh (approved):
Nice!

PR #3942, opened 2024-02-11
Add search box in world bank custom DC home page

shifucun:
This can be a general way to to add search box to custom DC home page


.gcloudignore
```diff
@@ -20,10 +20,10 @@
...
 /setup.cfg
 
-**/node_modules/**
+static/node_modules/
 
-.env/
+server/dist/
 
-screenshots/*.png
+.env/
```
juliawu, line 27: Should we use `**/.env/` to include the venvs from the tools directory?

server/templates/custom_dc/worldbank/homepage.html
```diff
@@ -24,17 +24,18 @@
 {% set is_hide_sub_footer = true %}
 
 {% block head %}
-<link rel="stylesheet" href={{url_for('static', filename='css/static.min.css', t=config['VERSION'])}}>
+<link rel="stylesheet" href={{url_for('static', filename='css/homepage.min.css', t=config['VERSION'])}}>
+ <script src={{url_for('static', filename='homepage_custom_dc.js', t=config['VERSION'])}} async></script>
```
juliawu, line 28: nit: make these two lines `<` aligned. I think there's an extra space in front of `<script`.

server/templates/custom_dc/worldbank/homepage.html
```diff
@@ -24,17 +24,18 @@
...
+
 {% endblock %}
 
 {% block content %}
+<div class="container">
+  <h4>Explore World Bank Data</h4>
```
juliawu, line 34: Just want to confirm move from `<h3>` to `<h4>` is intentional.

static/webpack.config.js
```diff
@@ -126,6 +126,10 @@ const config = {
       __dirname + "/js/apps/homepage/main.ts",
       __dirname + "/css/homepage.scss",
     ],
+    homepage_custom_dc: [
+      __dirname + "/js/apps/homepage/main_custom_dc.ts",
+      __dirname + "/css/homepage.scss",
+    ],
```
juliawu, line 132: I see both `homepage` and `homepage_custom_dc` entry points use `/css/homepage.scss`. Is the intention for both main and custom homepages to share `/css/homepage.scss`? Why not give each its own scss?

juliawu (commented):
Thanks! Just some quick questions

.gcloudignore
shifucun: Done

server/templates/custom_dc/worldbank/homepage.html
shifucun: Done

server/templates/custom_dc/worldbank/homepage.html
shifucun: Yea, it feels too big with the query box..

static/webpack.config.js
shifucun: custom dc has a way to override which is in "static/custom_dc/custom/overrides.css". Here needs to use either base.css or hompage.css to make the default UI works.

shifucun (commented):
thans for review!

PR #3943, opened 2024-02-12
update goldens and webdriver

chejennifer:
Update goldens and webdriver tests. Looks like there's been new census data

### DIFF
--- a/server/integration_tests/test_data/e2e_answer_places/whatistheemissionsinthesecounties/chart_config.json
+++ b/server/integration_tests/test_data/e2e_answer_places/whatistheemissionsinthesecounties/chart_config.json
@@ -15,16 +15,124 @@
                       "sort": "DESCENDING"
                     },
                     "comparisonPlaces": [
+                      "geoId/06049",
+                      "geoId/06041"
+                    ],
+                    "statVarKey": [
+                      "Annual_Emissions_GreenhouseGas_NonBiogenic_multiple_place_bar_block"
+                    ],
+                    "title": "Annual Emissions of Non Biogenic Greenhouse Gases (${date})",
+                    "type": "BAR"
+                  }
+                ]
+              }
+            ],
+            "denom": "Count_Person",
+            "infoMessage": "Data for \"Sierra County, Trinity County, Plumas County, Calaveras County, Mariposa County, Nevada County, Amador County and Tuolumne County\" is currently unavailable. See the following statistics on the other places.",
+            "title": "Annual Emissions of Non Biogenic Greenhouse Gases"
+          },
+          {
+            "columns": [
+              {
+                "tiles": [
+                  {
+                    "barTileSpec": {
+                      "maxPlaces": 15,
+                      "maxVariables": 15,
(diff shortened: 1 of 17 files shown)


server/integration_tests/test_data/e2e_edge_cases2/howdoesschoolsizeofurbanschoolscomparetoruralschoolsinus/chart_config.json
```diff
@@ -27,12 +27,20 @@
               {
                 "tiles": [
                   {
+                    "barTileSpec": {
```
chejennifer, line 30: the data change has caused the best facet for these stat vars to only have a single data point: https://screenshot.googleplex.com/58S3k5urvoosXqC

server/integration_tests/test_data/e2e_no_default_place_specified/chart_config.json
```diff
@@ -239,17 +239,6 @@
           },
           {
             "columns": [
-              {
```
chejennifer, line 242: the data change has caused the best facet for these stat vars to only have a single data point: https://screenshot.googleplex.com/58S3k5urvoosXqC 

so this was demoted from TIMELINE_WITH_HIGHLIGHT to BAR_CHART & because it is only single sv and single place, highlight chart was added instead of a bar

pradh: @ajaits This requires updating `CensusUSAMedianAgeIncome`

https://source.corp.google.com/piper///depot/google3/datacommons/import/mcf/manifest/prod_us_census_mcfs.textproto;l=95-107?q=CensusUSAMedianAgeIncome&ss=piper%2FGoogle%2FPiper:google3%2Fdatacommons%2Fimport%2Fmcf%2Fmanifest%2F

Or should we fold that into main Census (but that requires rebuilding census)?

pradh (approved):

server/integration_tests/test_data/e2e_no_default_place_specified/chart_config.json
shifucun: What exact place and variable is this? I can try it in mixer golden setup.

server/integration_tests/test_data/e2e_answer_places/whatistheemissionsinthesecounties/chart_config.json
```diff
@@ -15,39 +15,21 @@
                       "sort": "DESCENDING"
                     },
                     "comparisonPlaces": [
-                      "geoId/06105",
```
chejennifer, line 18: this got moved down because dc/topic/Emissions is a better match for the query & no data existed for that topic before

server/integration_tests/test_data/e2e_no_default_place_specified/chart_config.json
pradh: https://autopush.datacommons.org/tools/timeline#statsVar=Median_Age_Person_Female&place=country/USA

PR #3947, opened 2024-02-14
[node server] Handle block level per capita

chejennifer:
Handle cases in node server where the per capita specification comes from the block

example results:

query: What is the GDP per capita in Switzerland?
response: https://paste.googleplex.com/4980846865416192

query: What is the GDP in Switzerland?
response: https://paste.googleplex.com/6015469654376448


pradh (approved):

chejennifer:
thanks for the review!

PR #3948, opened 2024-02-15
Update goldens

chejennifer:
update goldens after updating dev with latest data


server/integration_tests/test_data/detection_api_context/stateswithhighestphds/chart_config.json
```diff
@@ -1,34 +1,16 @@
 {
-  "childEntityType": "State",
```
chejennifer, line 2: I'm not sure what change would've caused this, but tried running this on prod and got the same result: https://paste.googleplex.com/5743034006765568

server/integration_tests/test_data/detection_api_context/stateswithhighestphds/chart_config.json
```diff
@@ -1,34 +1,16 @@
...
-  ]
+  "failure": "Sorry, could not complete your request.",
+  "place": {
+    "dcid": "",
+    "name": "",
+    "place_type": ""
+  },
+  "userMessage": {
+    "msg": "Sorry, could not complete your request."
+  }
```
juliawu, line 15: Is having this entry be a failed response intentional?

server/integration_tests/test_data/detection_api_context/stateswithhighestphds/chart_config.json
pradh: Looks like this hits LLM and it is blocking.  Might be OK for now, and I will try to switch to using Gemini API, instead of PaLM.

pradh (approved):

server/integration_tests/test_data/detection_api_context/stateswithhighestphds/chart_config.json
chejennifer: see my comment earlier about this diff & prashanth's response

server/integration_tests/test_data/detection_api_context/stateswithhighestphds/chart_config.json
chejennifer: ok sounds good, I will merge this

chejennifer:
thanks for the reviews!

PR #3950, opened 2024-02-15
Switch to Gemini Pro instead of PaLM as default

pradh:
Plan to keep this PR outside of today's release.

TODO:  Validate the adversarial run after this is submitted.


chejennifer (approved):
thanks!

pradh:
Thanks!

PR #3952, opened 2024-02-15
[node server] add sv dcid place and place type to response

chejennifer:
Add following fields to chart response on node server:
- svs: list of sv dcids that are shown in the chart
- places: list of place dcids (or parent place dcid) show in the chart
- placeType: the place type shown in the chart if it's a contained in place chart (e.g., map, scatter, ranking)

example: https://paste.googleplex.com/5278230934716416


static/nodejs_server/types.ts
```diff
@@ -32,6 +32,13 @@ export interface TileResult {
   title: string;
   // The type of the tile
   type: string;
+  // Dcids of the stat vars shown in the chart
+  svs: string[];
```
pradh, line 36: from the standpoint of end-user readability of JSON, call it `vars` in place of `svs` maybe?

pradh (approved):
Thanks!

static/nodejs_server/types.ts
chejennifer: done

PR #3954, opened 2024-02-16
[node server] Add ability to run queries asynchronously in load test 

chejennifer:
Adds a version of the loadtest that runs the list of queries asynchronously.

Looking at the results, it seems that running queries asynchronously increases the timing by quite a bit, need to look into this.


tools/nl/loadtest/loadtest.py
```diff
@@ -72,9 +95,79 @@ def load_test(query_file, output_file):
...
+    }
+
+
+# Runs asynchronous load test where all the queries are requested in parallel
+async def run_async_load_test(query_file, output_file):
+  results = {}
+  result_keys = []
+  async with aiohttp.ClientSession() as session:
+    with open(query_file) as inf:
+      tasks = []
```
pradh, line 140: Should we control the parallelism?

tools/nl/loadtest/loadtest.py
```diff
@@ -72,9 +95,79 @@ def load_test(query_file, output_file):
...
+        _RESULT_KEY_STATUS: 'FAILURE',
+        _RESULT_KEY_QUERY: query,
+    }
+
+
+# Runs asynchronous load test where all the queries are requested in parallel
+async def run_async_load_test(query_file, output_file):
+  results = {}
+  result_keys = []
+  async with aiohttp.ClientSession() as session:
```
pradh, line 138: Also, I wonder about using concurrent.futures() instead?

So the same code can be used for single-threaded (if num-workers = 1) or parallel load?

chejennifer: done

tools/nl/loadtest/loadtest.py
chejennifer: Do we not want to issue all the queries in parallel? I've updated to using concurrent.futures where we could set how many threads to have, but for the parallel mode, I use as many threads as there are queries

pradh: So when you ran this what was the QPS?

The number of queries seems too few.  What I think might be helpful as a throughput test is, to run this + say https://github.com/datacommonsorg/website/blob/ab8cc12df92b283b523b63d3aedfc623517e2e2d/static/js/apps/explore_landing/topics.json queries in a loop as a sustained load for ~30 mins at a fixed parallelism of say `100`.  Then we look at the QPS.  Does that make sense?

shifucun:
Also need to clear redis cache before you run this?

chejennifer:
closing this because created a new tool for qps testing here: https://github.com/datacommonsorg/website/pull/3978

PR #3962, opened 2024-02-22
Update deployment YAMLs: dev instance IP and certificate name

juliawu:
This PR checks in the changes made during the redeployment of dev.datacommons.org on Feb 21, 2024:

Changes:

* The dev instance now lives on a new IP address. `dev.yaml` is updated with the new IP address.
* The ssl cert generated by `setup_ssl.sh` uses a different name for the created cert than is expected by `mci.yaml`. Updated `mci.yaml.tpl` to use the correct cert name.

### DIFF
--- a/deploy/gke/dev.yaml
+++ b/deploy/gke/dev.yaml
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 project: datcom-website-dev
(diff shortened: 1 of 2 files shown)


chejennifer (approved):
nice, thanks for updating!

PR #3966, opened 2024-02-23
[Place Page SEO] Suppress "sorry" message if summary is present

juliawu:
Updates the place pages to not show the "Sorry, there was an error loading charts for this place." message if summary text is being shown (i.e., there is content on the page). 

This addresses the hypothesis that our "Soft 404" errors come from the crawler picking up our "sorry" message when the crawler cancels our data fetch for being too slow.

Addressing the data fetch latency is a larger change that will come in future PRs.

Before (what crawler sees):
Screenshot 2024-02-22 at 5 41 08 PM

After (what crawler sees):
Screenshot 2024-02-22 at 5 40 32 PM

Behavior when no summary is present:
Screenshot 2024-02-22 at 5 52 08 PM

### DIFF
--- a/static/js/place/place.ts
+++ b/static/js/place/place.ts
@@ -150,7 +150,7 @@ function onSearch(q: string): void {
  * Set the text in the page-loading div to a sorry message.
  * Set the text to empty if summary text is present on the page.
  */
-function setSorryMessage(): void {
+function setErrorMessage(): void {
   const summaryText = document.getElementById("place-summary").innerText;
   const loadingElem = document.getElementById("page-loading");
   if (summaryText) {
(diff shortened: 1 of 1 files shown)


static/js/place/place.ts
```diff
@@ -146,6 +146,22 @@ function onSearch(q: string): void {
   window.location.href = `/explore#q=${encodeURIComponent(q)}`;
 }
 
+/**
+ * Set the text in the page-loading div to a sorry message.
+ * Set the text to empty if summary text is present on the page.
+ */
+function setSorryMessage(): void {
```
dwnoble, line 153: Nit: rename to `setErrorMessage`

dwnoble (approved):

static/js/place/place.ts
juliawu: Done!

PR #3955, opened 2024-02-17
Add new undata DC backend with SDG+WHO datasets

pradh:
This is to support UN's upcoming WHO launch.

PR #3968, opened 2024-02-23
Improve UN topic NL descriptions

pradh:
* Also, drop a couple of variables related to estimate availability from the topics.

* The "poverty" variables in UN topic seems not as good as poverty from SDG, so slightly demote it.

* Also, address comments in https://github.com/datacommonsorg/website/pull/3955

This fixes the query below:

image


dwnoble (approved):
Nice!

PR #3977, opened 2024-02-26
UNData: incorporate typo and fix bar order

pradh:
Before:

image

After:

image


server/lib/nl/config_builder/bar.py
```diff
@@ -85,9 +88,22 @@ def multiple_place_bar_block(column,
...
+  # Use no default sort_order for special DC, since UN
+  # want the order to be as provided in variable groupings.
+  sort_order = None if is_special_dc(state.uttr.insight_ctx) \
+    else BarTileSpec.DESCENDING
+
+  if (RankingType.LOW in cspec.ranking_types and
+      RankingType.HIGH not in cspec.ranking_types):
+    sort_order = BarTileSpec.ASCENDING
+  elif RankingType.HIGH in cspec.ranking_types:
+    sort_order = BarTileSpec.DESCENDING
```
chejennifer, line 107: this logic is slightly different from before for the case of both RankingType.HIGH and RankingType.LOW showing up in ranking_types or if neither show up, is that something you meant to change? Because before, those cases would still sort in descending order & now it won't do any sorting.

chejennifer (approved):
nice, just one question about whether logic change was intended

pradh:
Thanks for the review!

PR #3978, opened 2024-02-26
Load test changes

chejennifer:
Set of changes for getting bard instance to a qps of 50: 
- new load testing tool that runs x total queries and runs it over y parallel requests at a time
- rename old load testing tool as latencies test
- deployment updates
    - new nl server docker image specific for bard that uses more workers
    - updated bard instance gke settings
    - corresponding mixer changes: 
    - increase web server workers from 4 to 8


build/ci/cloudbuild.deploy.yaml
```diff
@@ -61,6 +61,21 @@ steps:
         docker push gcr.io/datcom-ci/datacommons-nl:latest
     waitFor: ["-"]
 
+  - id: push-nl-server
```
shifucun, line 64: push-nl-bard-server?

build/nl_server_bard/Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+RUN ltt install torch --cpuonly
+RUN pip3 install --no-cache-dir \
+    -r /workspace/nl_server/requirements.txt
+# Flask
+COPY nl_server/. /workspace/nl_server/
+COPY shared/. /workspace/shared/
+
+# Run server
+WORKDIR /workspace
+CMD exec gunicorn --timeout 1000 --bind :6060 -w 3 nl_app:app
```
shifucun, line 37: shall we also update this for our nl server? Then we just have one nl server to maintain.

If you really want to have separate one, can just use an ENV in dockerfile and set it in helm yaml, then you can just do 

-w <ENV_NUM_NL_WORKER>

shifucun, line 37: why not adding preload?

deploy/helm_charts/dc_website/templates/deployment.yaml
```diff
@@ -159,12 +163,12 @@ spec:
             httpGet:
               path: /healthz
               port: 6060
-            timeoutSeconds: 5
+            timeoutSeconds: 300
```
shifucun, line 166: add comments why long timeout?

build/ci/cloudbuild.deploy.yaml
chejennifer: removed this because of next comment

build/nl_server_bard/Dockerfile
chejennifer: This update also requires more memory for the pods, so I think it's better to only make this update for instances we need it for. 

Updated to using ENV

chejennifer: there was an issue with preload where the NL server wouldn't start

deploy/helm_charts/dc_website/templates/deployment.yaml
chejennifer: done

shifucun (approved):
looking good for the GKE changes. Will let leave the load script for shanth to review.

pradh:
> looking good for the GKE changes. Will let leave the load script for shanth to review.

Could you maybe leave out the script changes and make all the prod changes in this PR?

pradh:
> > looking good for the GKE changes. Will let leave the load script for shanth to review.
> 
> Could you maybe leave out the script changes and make all the prod changes in this PR?

(Generally good practice to keep prod changes isolated and small)

chejennifer:
> > looking good for the GKE changes. Will let leave the load script for shanth to review.
> 
> Could you maybe leave out the script changes and make all the prod changes in this PR?

done, moved script changes to a new PR: https://github.com/datacommonsorg/website/pull/3981

build/nl_server/Dockerfile
```diff
@@ -34,4 +34,4 @@ COPY shared/. /workspace/shared/
 
 # Run server
 WORKDIR /workspace
-CMD exec gunicorn --timeout 60 --bind :6060 nl_app:app
\ No newline at end of file
+CMD exec gunicorn --timeout 1000 --bind :6060 -w $((NUM_WORKERS + 0)) nl_app:app
```
pradh, line 37: Add a comment why 1000?

build/web_server/Dockerfile
```diff
@@ -47,4 +47,4 @@ RUN npm run-script build
 COPY server/. /workspace/server/
 COPY shared/. /workspace/shared/
 WORKDIR /workspace
-CMD exec gunicorn --preload --timeout 1000 --bind :8080 -w 4 web_app:app
+CMD exec gunicorn --preload --timeout 1000 --bind :8080 -w 8 web_app:app
```
pradh, line 50: Add a comment why 8?

Also was 8 necessary, or was it mostly the nl_server thing?

deploy/helm_charts/envs/bard.yaml
```diff
@@ -41,21 +42,28 @@ ingress:
...
   node:
-    replicas: 15
+    replicas: 25
+    nodePool: "pool-e2-cpu"
   observation:
     replicas: 15
+    nodePool: "default-pool"
   default:
     replicas: 9
+    nodePool: "default-pool"
```
pradh, line 67: comment on rationale for the choices?

pradh (approved):
Nice!

build/nl_server/Dockerfile
chejennifer: done

build/web_server/Dockerfile
chejennifer: Switching to 8 web app workers was a change I tried early on in the load testing process, so I'm honestly not sure if 4 would've also worked with the later changes. Since all the latest testing has been done with 8 web app workers and I don't see any big red flags for using that many, I think it's ok to just keep as 8?

deploy/helm_charts/envs/bard.yaml
chejennifer: done

build/web_server/Dockerfile
pradh: I see.  what is the mem-usage with 4 vs. 8?

pradh: But yeah, I generally agree with the change (even if we don't have evidence of it alone improving things).

chejennifer: I don't recall the exact mem-usage, but I remember the change did not require increasing the memory on the pods & the website-app container is limited to 2G of memory. Taking a look at the graphs, looks like current memory usage (20 pods, 8 workers each) is about 35G

pradh: SG!

chejennifer:
thanks for the reviews!

PR #3985, opened 2024-02-27
update goldens

chejennifer:
Looks like the ranking of the countries have changed with additional 2022 data

latest data on autopush is 2022: https://screenshot.googleplex.com/Rwi8a76s3L7xWVQ
latest data on prod is 2021: https://screenshot.googleplex.com/3sRUxGZMtPWv7hJ


pradh (approved):
Thanks!

PR #3996, opened 2024-03-05
UNData: increase topic limit and fix edge-cases

pradh:
1.  Improves further on https://github.com/datacommonsorg/website/commit/a820bde8a37f9eb4fc3eeec69b239e2680b55161 by opening more topics, but having an SV-based limit for Special DCs.

2.  Sync mixer to fix the "number of cholera cases" query.

image

3.  For the edge-case, the following query was failing.  Because "filter_with_single_var" fulfiller would not fallback to place-comparison, rather only to Basic (and basic won't fulfill multi-place query).
image


hareesh-ms (approved):

pradh:
Thanks Hareesh!

PR #4000, opened 2024-03-06
Disable 'snap to date with highest coverage' checkbox if the highest coverage data is also the latest data available

dwnoble:
* Disable 'snap to date with highest coverage' checkbox if the highest coverage data is also the latest data available
* Added @datacommonsorg/client dependency for coverage check

Screenshot:

BXLNGcHFAk9hhFx


miss-o-soup:
Looking good Dan, thank you so much for making this extra addition. For the tooltip text, I like how you frame it here in this item. Inspired by this can we have the following text for the tooltip "?"

"The highest coverage data is also the latest data available for this chart"

static/js/utils/app/explore_utils.ts
```diff
@@ -61,3 +63,64 @@ export function getTopics(
...
+      date: DATE_LATEST,
+    });
+
+  // Return false if we find any "latest observation dates" that differ from the
+  // "highest coverage date"
+  const highestCoverageVariableDcids = Object.keys(
+    highestCoverageObservations.data
+  );
+  for (let i = 0; i < highestCoverageVariableDcids.length; i++) {
+    const variableDcid = highestCoverageVariableDcids[i];
```
juliawu, line 103: nit: Can you make this a for of loop?

juliawu (approved):
Thanks!

dwnoble:
> The highest coverage data is also the latest data available for this chart

@miss-o-soup  Done

6Pjduos8ct4ZUpK

static/js/components/subject_page/block.tsx
```diff
@@ -115,33 +116,64 @@ const rankingTileLatestDataAvailableFooter = intl.formatMessage({
...
+async function shouldEnableSnapToHighestCoverage(
+  placeDcid: string,
+  enclosedPlaceType: string,
+  columns: ColumnConfig[],
+  statVarProvider: StatVarProvider
+): Promise<boolean> {
+  // Check if highest coverage & latest date observations are the same
+  const tiles = _.flatten(_.flatten(columns.map((c) => c.tiles)));
+  const statVarKeys = _.flatten(tiles.map((tile) => tile.statVarKey));
+  const statVarSpecs = statVarProvider.getSpecList(statVarKeys);
```
chejennifer, line 165: nit: could have a helper for getting all the stat var specs for a list of columns since this code is repeated in isEligibleForSnapToHighestCoverage

static/js/utils/app/explore_utils.ts
dwnoble: done

static/js/apps/explore/app.test.tsx
```diff
@@ -0,0 +1,109 @@
...
+import Adapter from "@wojtekmaj/enzyme-adapter-react-17";
+import Cheerio from "cheerio";
+import Enzyme, { mount } from "enzyme";
+import React from "react";
+
+import { App } from "./app";
+
+Enzyme.configure({ adapter: new Adapter() });
+/*
+const mockedAxios = axios as jest.Mocked<typeof axios>;
```
chejennifer, line 30: curious if you meant to keep this commented out section

static/js/utils/app/explore_utils.ts
```diff
@@ -61,3 +63,63 @@ export function getTopics(
...
+    const entityDcid = Object.keys(
+      highestCoverageObservations.data[variableDcid]
+    ).pop();
+    if (!entityDcid) {
+      continue;
+    }
+    const highestCoverageDate =
+      highestCoverageObservations.data[variableDcid][entityDcid].date;
+    const latestObservationEntityDcids = Object.keys(
+      latestObservations.data[variableDcid]
```
chejennifer, line 112: should check that there exists variableDcid in latestObservations.data first or `latestObservations.data[variableDcid] || {}`

static/js/utils/app/explore_utils.ts
```diff
@@ -61,3 +63,63 @@ export function getTopics(
...
+      date: DATE_LATEST,
+    });
+
+  // Return false if we find any "latest observation dates" that differ from the
+  // "highest coverage date"
+  const highestCoverageVariableDcids = Object.keys(
+    highestCoverageObservations.data
+  );
+  for (let variableDcid of highestCoverageVariableDcids) {
+    const entityDcid = Object.keys(
```
chejennifer, line 103: curious, are you just checking one entity per variable?

chejennifer (approved):
nice

static/js/apps/explore/app.test.tsx
dwnoble: Whoops didnt mean to include this file. removed.

static/js/components/subject_page/block.tsx
dwnoble: Good callout with the redundancy. In one of these cases, i'm also extracting the "tile types" as well as stat vars from the columns, and that intermediate step makes it less useful to generalize this to a function (i could do a helper that converts columns to a list of tiles, but that's already a one-liner).

im leaning keeping it as is for now- wdyt?

static/js/utils/app/explore_utils.ts
dwnoble: done

static/js/utils/app/explore_utils.ts
dwnoble: In this case yes- all entities in the "highest coverage observations" list are guaranteed to have the same date, so i'm just checking the date value of the first one

dwnoble: Added a comment

PR #4002, opened 2024-03-06
Normalize embedding vector for hosted NL models

shifucun:


model_server/serve.py
```diff
@@ -58,6 +59,10 @@ def create_model(model_name):
 app = Flask(__name__)
 
 
+def normalize(embeddings):
+  return [(e / np.linalg.norm(e)).tolist() for e in embeddings]
```
pradh, line 63: Add a comment why this is done?

pradh (approved):

model_server/serve.py
shifucun: Done

PR #4003, opened 2024-03-06
[Biomedical DC] Add the main landing page

juliawu:
This PR adds the main landing page for biomedical DC.

Changes Include:
* Adding a route /bio that takes the user to the new landing page (but hidden behind 404 on prod)
* CSS, TS, HTML for the new landing page
* React components for the search box and clickable cards

NOT included, but to be added in followup PRs:
* A custom header, menu, and footer for biomedical pages
* A "more examples" button for the sample questions section
* Moving the homepage to its final destination in custom DC
* Internationalization of the content

Sample screenshots:


miss-o-soup:
This is looking awesome Julia.
This might not be the right place to call it out but wanted to bring attention to the tool bar treatment. We will place the DC logo + "• Biomedical" and not "Data Commons Biomedical". This is so that there is room for the search queries to have more space to render. Also added more details on the break points and how it relates to the main DC so that we have the same functionality. See from slide 25 onwards in our specs deck: https://screenshot.googleplex.com/7khCEbLqC7YZdUW

server/routes/biomedical/html.py
```diff
@@ -0,0 +1,41 @@
...
+import os
+
+import flask
+from flask import Blueprint
+from flask import current_app
+from flask import render_template
+
+bp = Blueprint('biomedical', __name__, url_prefix='/bio')
+
+PAGE_CONFIG_FILE = "config/biomedical_landing_page/display_items.json"
```
chejennifer, line 26: nit: use _<variable name> for private constants (`_PAGE_CONFIG_FILE`)

server/routes/biomedical/html.py
```diff
@@ -0,0 +1,41 @@
...
+
+def get_page_config() -> dict:
+  path = os.path.join(current_app.root_path, PAGE_CONFIG_FILE)
+  with open(path) as f:
+    return json.load(f)
+
+
+@bp.route('/')
+def main():
+  if os.environ.get('FLASK_ENV') == 'production':
```
chejennifer, line 37: Since there are lots of different environments, should we limit this landing page to just autopush, dev, and local with: `if os.environ.get('FLASK_ENV') not in ['autopush', 'dev', 'local']`

static/css/biomedical/biomedical_landing.scss
```diff
@@ -0,0 +1,265 @@
...
+      line-height: var(--title-caption-line-height);
+    }
+  }
+
+  .thematic-image {
+    margin-top: auto;
+  }
+}
+
+.card {
```
chejennifer, line 219: curious what card this is referring to because I see that there is a file for Card with its own css

static/js/biomedical/landing/biomedical_landing.ts
```diff
@@ -0,0 +1,156 @@
...
+
+import { BiomedicalSearchBox } from "./biomedical_search_box";
+import { CardProps, CardTheme } from "./card";
+import { CardWall } from "./card_wall";
+
+// Search url for biomedical searches
+const BIOMEDICAL_SEARCH_URL = "/explore/#q=";
+
+// Mapping of categories to color theme to apply to cards in that category
+const CATEGORIES_TO_THEME = {
```
chejennifer, line 30: nit: I wonder if it would be better to include the color theme in the config so information about each category isn't spread out in mulitple places

static/js/biomedical/landing/biomedical_landing.ts
```diff
@@ -0,0 +1,156 @@
...
+  Diseases: "blue",
+  Ecology: "muted-blue",
+  Pharmacology: "red",
+};
+
+// Order to show card categories in, from left to right
+const CARD_CATEGORY_ORDER = ["Ecology", "Diseases", "Pharmacology"];
```
chejennifer, line 37: and this could also go into the config where the config just returns a list of category objects like:

```
[
{title: "Ecology",
themeColor: "muted-blue",
questions: ["question1",...]
},
...
]
```

static/js/biomedical/landing/biomedical_landing.ts
```diff
@@ -0,0 +1,156 @@
...
+  "E.g. What disease are the following genetic variances associated with? rs7903146, rs2237897, rs4712524, rs6769511, rs5219";
+
+/**
+ * Get colors to apply to a Card as a theme, based on name of theme
+ * @param themeName text name of the theme to apply
+ * @returns CardTheme matching the theme name
+ */
+function getColorTheme(themeName: string): CardTheme {
+  switch (themeName) {
+    case "red":
```
chejennifer, line 50: could have a map of theme name to CardTheme?

static/js/biomedical/landing/biomedical_search_box.tsx
```diff
@@ -0,0 +1,86 @@
...
+/**
+ * Search Box for the Biomedical DC landing page
+ */
+
+import React, { useEffect, useRef, useState } from "react";
+import { Input, InputGroup } from "reactstrap";
+
+interface BiomedicalSearchProps {
+  // Whether to allow searching for an empty string
+  allowEmptySearch?: boolean;
```
chejennifer, line 26: curious, is there a case where we allow empty search?

chejennifer (commented):
nice!

static/js/biomedical/landing/card.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+export function Card(props: CardProps): JSX.Element {
+  function handleClick(): void {
+    if (props.url) {
+      window.open(props.url, "_blank").focus();
+    }
+  }
+  return (
+    <Button className="card" onClick={handleClick}>
+      <Text className="card-text" theme={props.theme}>
```
dwnoble, line 87: To jennifer's comment above about the `.card` class in css, with styled components you can do something like:

```
const StyledText = styled(Text)`
   // your css properties here
`;
```
and use it like:

```
<StyledText theme={props.theme}>
  ...
</StyledText>
```

static/js/biomedical/landing/biomedical_landing.ts
```diff
@@ -0,0 +1,156 @@
...
+    document.getElementById("config-data").dataset.config
+  );
+
+  // Render cards for the sample questions
+  const questionCards = getCardSpecsFromConfig(
+    config,
+    "sample-questions",
+    true
+  );
+  ReactDOM.render(
```
dwnoble, line 135: For consistency sake I would say either make the entire page render from a flask template or make the entire page render from a single react component. I think this will make the page flow easier to follow, and will avoid splitting the styles between css and styled components

static/js/biomedical/landing/biomedical_search_box.tsx
```diff
@@ -0,0 +1,86 @@
...
+        autoFocus={false}
+        className="pac-target-input search-input"
+        id="search-input"
+        innerRef={inputRef}
+        invalid={invalid}
+        onChange={(e) => {
+          setValue(e.target.value);
+          setInvalid(false);
+        }}
+        onKeyDown={(e) => e.key === "Enter" && handleSearch()}
```
dwnoble, line 70: Hitting enter makes a newline before running the search. Maybe it needs a `e.preventDefault()`?

static/js/biomedical/landing/card.tsx
```diff
@@ -0,0 +1,95 @@
...
+}
+
+export function Card(props: CardProps): JSX.Element {
+  function handleClick(): void {
+    if (props.url) {
+      window.open(props.url, "_blank").focus();
+    }
+  }
+  return (
+    <Button className="card" onClick={handleClick}>
```
dwnoble, line 86: Nit: use an anchor and href instead of a button here for better accessibility [go/pan-a11y-reference#buttons](https://g3doc.corp.google.com/cloud/console/g3doc/producer/develop/accessibility/developer-reference.md?cl=head#buttons)

static/js/biomedical/landing/card_wall.tsx
```diff
@@ -0,0 +1,112 @@
...
+`;
+
+interface CardStackProps {
+  // Specs of cards in the stack
+  cardSpecs: CardProps[];
+  // Direction to stack. One of "row" or "column". Defaults to "row".
+  direction?: string;
+}
+
+interface CardWallProps {
```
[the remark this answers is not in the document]
juliawu: Good idea, switched it all to "stack" to be consistent with naming.

server/routes/biomedical/html.py
juliawu: Added _

server/routes/biomedical/html.py
juliawu: Good callout, limited to autopush, dev, and local like you suggested

static/js/biomedical/landing/biomedical_search_box.tsx
juliawu: I originally kept it to match the specs of the NL search bar on main DC, but removed it since I can't think of a case where we'd allow empty search.

static/js/biomedical/landing/biomedical_landing.ts
juliawu: Thanks! I like this. Swapped to putting these in the config as a list.

static/js/biomedical/landing/biomedical_landing.ts
juliawu: Good idea. Put color theme in the config

static/css/biomedical/biomedical_landing.scss
juliawu: It was originally there to allow for repsonsive text resizing, with the css card file as its own css. I've switched to doing all rendering with a single react component, so all styling is now in the Card file.

static/js/biomedical/landing/card.tsx
juliawu: Thanks! Used that to handle colors.

static/js/biomedical/landing/biomedical_landing.ts
juliawu: Made entire page render from a single app react component, so now all styling is in the styled components. The main .scss file now just has some overrides for our base mani DC styling

static/js/biomedical/landing/biomedical_search_box.tsx
juliawu: Good find! Added e.preventDefault()

static/js/biomedical/landing/card.tsx
juliawu: Good callout! Switched to using a stylized anchor

static/js/biomedical/landing/header_and_search_section.tsx
```diff
@@ -0,0 +1,118 @@
...
+        <StyledText>
+          <h1>Data Commons • Biomedical</h1>
+          <h2>A bridge to open biomedical data</h2>
+          <h3>
+            Find relationships between 25 different categories of biomedical
+            data like genes, proteins, diseases, and more
+          </h3>
+          <BiomedicalSearchBox
+            onSearch={(query) => {
+              window.location.href = `${BIOMEDICAL_SEARCH_URL}${query}`;
```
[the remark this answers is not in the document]
juliawu: Used urlSearchParams to process the query to address this.

static/js/biomedical/landing/highlights_section.tsx
```diff
@@ -0,0 +1,92 @@
...
+  locale: string;
+}
+
+export function HighlightsSection(props: HighlightsSectionProps): JSX.Element {
+  return (
+    <HighlightsContainer className="container">
+      {props.config.map((callout, index) => {
+        const label = Object.keys(callout)[0];
+        if (label === "Entities") {
+          const value = Intl.NumberFormat(props.locale, {
```
chejennifer, line 71: curious if internationalization of the numbers is something we are worried about? If not, we could just update the config to be formatted like: 

```
[
  {
    label: string,
    value: string
  },
  ...
]
```
Actually, even if you do want internationalization of the numbers, this format may be cleaner instead of getting the label by reading the first key of the object

static/js/biomedical/landing/app.tsx
```diff
@@ -0,0 +1,102 @@
...
+ *                     when clicked
+ * @returns a grid of CardProps to render on the page
+ */
+function getCardSpecsFromConfig(
+  config: Record<string, CategoryConfig[]>,
+  sectionName: string,
+  addSearchUrl: boolean
+): CardProps[][] {
+  // Build card specs for each entry in config
+  const sectionConfig = config[sectionName];
```
chejennifer, line 65: add `|| []` at the end in case the sectionName isn't found?

static/js/biomedical/landing/app.tsx
```diff
@@ -0,0 +1,102 @@
...
+        cardSpec.url = `${BIOMEDICAL_SEARCH_URL}${BIOMEDICAL_SEARCH_QUERY_PARAM}=${cardSpec.text}`;
+      }
+      cardSpec.tag = categoryConfig.category;
+      cardSpec.theme =
+        COLOR_TO_CARD_THEME[
+          categoryConfig.themeColor || DEFAULT_CARD_COLOR_THEME
+        ];
+    });
+  }
+  return sectionConfig.map((categoryConfig) => {
```
chejennifer, line 78: nit: a little hard to tell that you were updating the cards in place and then going back through the updated object to get the cards, might be a little clearer to do this or add some comments:

```
const cards = []
... 
for (const categoryConfig of sectionConfig) {
 categoryCards = categoryConfig.cards.map(cardSpec => ...)
 cards.push(categoryCards)
}
return cards
```

static/js/biomedical/landing/card_wall.tsx
```diff
@@ -0,0 +1,110 @@
...
+
+  @media ${BREAKPOINTS.md} {
+    flex-wrap: wrap;
+  }
+`;
+
+interface CardStackProps {
+  // Specs of cards in the stack
+  cardSpecs: CardProps[];
+  // Direction to stack. One of "row" or "column". Defaults to "row".
```
chejennifer, line 61: nit: could use an enum for direction?

chejennifer (commented):
Thanks for the updates, some nits

static/js/biomedical/landing/biomedical_search_box.tsx
dwnoble, line 1: I don't think there's anything biomedical-specific to this component. How about calling it something like "multi_line_search_box" or "text_area_search_box"?

static/js/biomedical/landing/shared.tsx
dwnoble, line 1: Nit: i think this can be a .ts file (doesnt have any react components)

static/js/biomedical/landing/card.tsx
```diff
@@ -0,0 +1,101 @@
...
+  // main text to display on card
+  text: string;
+  // url to link to when card is clicked
+  url?: string;
+}
+
+export function Card(props: CardProps): JSX.Element {
+  function handleClick(): void {
+    if (props.url) {
+      window.open(props.url, "_blank").focus();
```
dwnoble, line 92: For better accessibility, attach the url & target directly to the anchor tag below (CardContainer in this case) like `<CardContainer href={props.url} target="_blank">` - then you can remove the click handler altogether

static/js/biomedical/landing/card.tsx
```diff
@@ -0,0 +1,101 @@
...
+
+export interface CardProps {
+  // tag or category to label card with
+  tag: string;
+  // styling options
+  theme: CardTheme;
+  // main text to display on card
+  text: string;
+  // url to link to when card is clicked
+  url?: string;
```
dwnoble, line 86: Why make url optional? will it ever be empty?

static/js/biomedical/landing/biomedical_search_box.tsx
```diff
@@ -0,0 +1,135 @@
...
+
+  // Adjust search box height while query is being entered
+  useEffect(() => {
+    setSearchBoxHeight();
+  }, [props.placeholderText, value]);
+
+  return (
+    <StyledInputGroup>
+      <SearchIcon>
+        <span className="material-icons-outlined">search</span>
```
dwnoble, line 103: I think this should be clickable- wdyt?

static/js/biomedical/landing/biomedical_search_box.tsx
```diff
@@ -0,0 +1,135 @@
...
+      </SearchIcon>
+      <StyledInput
+        autoComplete="off"
+        autoFocus={false}
+        innerRef={inputRef}
+        invalid={invalid}
```
dwnoble, line 109: Hitting enter with no query gives me this: https://screenshot.googleplex.com/3F6GPMBD47jonjz

static/js/biomedical/landing/card_wall.tsx
dwnoble: or 
```
direction?: "row" | "column;
```

static/js/biomedical/landing/highlights_section.tsx
dwnoble: jennifer's suggestion sgtm. Seems too tightly coupled to have the entities label in the config file and an override for it here.

dwnoble: also depending on the type of `"value"` you could apply number internationalization or not.

```
const displayValue = typeof item.value === 'number' ? formatter.format(item.value) : item.value
```

dwnoble (changes requested):

static/js/biomedical/landing/shared.tsx
```diff
@@ -0,0 +1,120 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * Shared styling, components, and constants for the biomedical landing pages
```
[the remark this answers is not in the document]
juliawu: Good idea, split into constants and shared_styled_components files

static/js/biomedical/landing/header_and_search_section.tsx
```diff
@@ -0,0 +1,128 @@
+/**
+ * Copyright 2023 Google LLC
```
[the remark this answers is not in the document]
juliawu: Good catch! Updated.

static/js/biomedical/landing/highlights_section.tsx
juliawu: Good ideas. I updated the config to use "label" and "value" keys, and also added logic to only format if the value is a number.

static/js/biomedical/landing/app.tsx
juliawu: Done

static/js/biomedical/landing/app.tsx
juliawu: Good call out. I used the cards = [] approach, please take a look to see if the logic is more clear now

static/js/biomedical/landing/biomedical_search_box.tsx
juliawu: Good catch! Looks like that's from Reactstrap's default invalid behavior. We don't do anything with the "invalid" state, so I've set invalid to false instead.

static/js/biomedical/landing/biomedical_search_box.tsx
juliawu: Good idea! Made the search icon clickable, and the cursor is now a pointer when hovering over.

static/js/biomedical/landing/card.tsx
juliawu: Thanks! Attached url directly.

static/js/biomedical/landing/shared.tsx
juliawu: Made a .ts file

static/js/biomedical/landing/biomedical_search_box.tsx
juliawu: Renamed to multi_line_search_box.

static/js/biomedical/landing/card.tsx
juliawu: Good question. Originally wanted to have a default URL be the biomedical search url, but moved that logic to app.tsx instead. Made url required now.

static/js/biomedical/landing/card_wall.tsx
juliawu: Switched to "row" | "column" for succinctness

static/js/biomedical/landing/highlights_section.tsx
```diff
@@ -0,0 +1,118 @@
...
+      {props.config.map((callout, index) => {
+        const isEntities = callout.label === "Entities";
+        const value =
+          typeof callout.value === "number"
+            ? formatNumber(callout.value, props.locale, isEntities)
+            : callout.value;
+        if (isEntities) {
+          return (
+            <Label key={index}>
+              Over {value} entities
```
dwnoble, line 103: With restructuring the callout items to have `label` and `value` keys, as well as only formatting `value` if it's a number is you can change the "Entities" entry to look like:
```
{
  "label": "Over 195 million entities,
  "value":  "& counting"
}
```

& totally remove the "Entities" reference here

dwnoble (commented):
Thanks Julia! 1 minor change requested & lgtm!

dwnoble (approved):

static/js/biomedical/landing/highlights_section.tsx
juliawu: Reason I implemented entities was so that the "195 million" number could also be updated programmatically as we add new entities to the KG. Removed "entities" reference for now, and we can revisit if/when we implement a script to update the highlighted values.

static/js/biomedical/landing/highlights_section.tsx
```diff
@@ -0,0 +1,108 @@
...
+ * @param locale locale to localize number to
+ * @param useWords whether to format a number using words. For example, for the
+ *                 number 3000000 if useWords is true, will format to
+ *                 "3 million". If false, will format to 3,000,000 in en locale.
+ * @returns formatted number for display.
+ */
+function formatNumber(
+  number: number,
+  locale: string,
+  useWords?: boolean
```
chejennifer, line 83: nit: useWords doesn't seem to be used?

chejennifer (approved):
thanks for the changes! Just one small nit

static/js/biomedical/landing/highlights_section.tsx
juliawu: Thanks! Removed useWords

PR #4015, opened 2024-03-08
NL: Delete NER logic

pradh:
This has not been used since launching DC place recognition.


server/lib/nl/common/debug_utils.py
```diff
@@ -108,7 +108,7 @@ def result_with_debug_info(data_dict: Dict, status: str,
       'original_query': query_detection.original_query,
       'detection_type': query_detection.detector,
       'llm_api_type': query_detection.llm_api,
-      'place_detection_type': query_detection.place_detector,
+      'place_detection_type': 'DC',
```
shifucun, line 111: should this be 'dc'?

shifucun (approved):
very nice with this huge clean up!

server/lib/nl/common/debug_utils.py
pradh: Done.

PR #4016, opened 2024-03-09
Try to run NL integration tests against staging

pradh:


shifucun (approved):
thanks for the update!

server/app_env/integration_test.py
```diff
@@ -20,4 +20,10 @@ class Config(_base.Config):
   API_ROOT = 'https://autopush.api.datacommons.org'
```
shifucun, line 20: should make this AutopushConfig then?

shifucun (approved):

server/app_env/integration_test.py
pradh: Done.

shifucun: Also update below to be StagingConfig(base._Config): with 5 fields defined?

pradh: Thanks!  Indeed, I badly broke tests.

PR #4019, opened 2024-03-11
[biomed nl] add entities to response

chejennifer:
this PR adds a new entities field to the nl response


pradh (approved):
nice!  Thanks for separating out this PR, so much better!

chejennifer:
thanks for the review

PR #4020, opened 2024-03-11
fix autopush deployment

chejennifer:
Update autopush website-app to deploy in default-pool. This fixed deploying autopush when trying manually

Background:
- autopush deployments have been failing since website cron tests have been added
- there are 2 different node pools used in the autopush instance, one with 4 CPU and one with 8 CPU
- I noticed that a lot of website-app pods were deployed in the 8 CPU node pool and they seemed to have a lot of restarts. In a past experiment, we've learned that website-app pods have trouble starting in node pools that have more CPU than 4.
- So I tested moving all the cron jobs to the 8 CPU node pool and all the website-app pods to the 4 CPU node pool and it seemed to work. (Will see how well it does once this PR is in though)

### DIFF
--- a/deploy/helm_charts/dc_website/values.yaml
+++ b/deploy/helm_charts/dc_website/values.yaml
@@ -167,3 +167,5 @@ cronTesting:
   screenshotDomain:
   # whether or not sanity tests are enabled
   enableSanity: true
+  # node pool to use
+  nodePool:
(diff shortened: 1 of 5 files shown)


shifucun (approved):
thanks for the fix!

Maybe add some background and investigation results in the PR description?

chejennifer:
thanks for the review!

PR #4021, opened 2024-03-11
Update some curated SV titles to remove %

n-h-diaz:
This is to address one of the World Bank Issues "Fix Inaccurate labeling of per 100K StatVars as percentages" 

I reverted the curated name to the original SV name (e.g. https://datacommons.org/browser/Count_Death_IntentionalSelfHarm_AsFractionOf_Count_Person)


pradh:
Thanks!   Can we update the names in the schema too?

pradh (approved):

PR #4022, opened 2024-03-12
[biomed nl] Fulfill logic for single entity and out prop

chejennifer:
Simple fulfill logic to add a ANSWER_MESSAGE tile and ENTITY_OVERVIEW tile for single entity and single prop

Screenshot 2024-03-11 at 10 23 59 PM


server/config/subject_page.proto
```diff
@@ -261,6 +261,40 @@ message MapTileSpec {
   string geo_json_prop = 3;
 }
 
+// Spec for getting value from property values api call
+message PropertySpec {
+  string property = 1;
+  string direction = 2;
```
pradh, line 267: describe the fields.  should direction be an enum?

server/config/subject_page.proto
```diff
@@ -261,6 +261,40 @@ message MapTileSpec {
...
+  repeated string value = 1;
+  repeated string source = 2;
+}
+
+message AnswerMessageTileSpec {
+  oneof answer {
+    PropertySpec property = 1;
+    DisplayValueSpec display_value = 2;
+  }
+  string entity = 3;
```
pradh, line 282: When is this set?

server/config/subject_page.proto
```diff
@@ -261,6 +261,40 @@ message MapTileSpec {
...
+
+message AnswerTableTileSpec {
+  message TableColumn {
+    string header = 1;
+    PropertySpec property = 2;
+  }
+  repeated TableColumn column = 1;
+  repeated string entity = 2;
+}
+
```
pradh, line 293: Please describe the spec, how this matches to a table.  And is "entity" required or optional?

server/config/subject_page.proto
```diff
@@ -261,6 +261,40 @@ message MapTileSpec {
...
+message EntityOverviewTileSpec {
+  string entity = 1;
```
pradh, line 295: Are we not having entity_ids in the page metadata?  I thought this would be an override/customization.  (But I also don't want to go off into long discussions, so if that was what decided, fine, document well)

server/lib/nl/common/utterance.py
```diff
@@ -78,6 +79,7 @@ class ChartType(IntEnum):
   SCATTER_CHART = 5
   EVENT_CHART = 6
   RANKED_TIMELINE_COLLECTION = 7
+  ANSWER_WITH_ENTITY_OVERVIEW = 8
```
pradh, line 82: nit: just ENTITY_OVERVIEW

pradh (commented):
Nice!  A first pass before end-of-day on protos.  I'll take a look again first thing tomm morning.

server/lib/nl/common/utterance.py
chejennifer: hm, would ANSWER_WITH_ENTITY_OVERVIEW be more clear though because it's going to be an answer tile with an optional entity overview tile ..

server/config/subject_page.proto
chejennifer: done

server/config/subject_page.proto
chejennifer: always

server/config/subject_page.proto
chejennifer: done

server/config/subject_page.proto
pradh: hmm, then do we want to add this to the Tile message, as a repeated `entity`.  We can then set this for places (they can use the same field) and switch over.

(if we agree, switchover in later PR is OK)

server/lib/nl/common/utterance.py
pradh: Oh I see, got it!  (add that comment:  "Represents an answer tile with an optional entity overview tile")

server/lib/nl/config_builder/builder.py
```diff
@@ -107,6 +108,16 @@ def build(state: PopulateState, config: Config) -> BuilderResult:
     # level.
     if user_message:
       cspec.info_message = ''
+
+    if cspec.chart_type == ChartType.ANSWER_WITH_ENTITY_OVERVIEW:
+      if len(cspec.props) > 1 or len(cspec.entities) > 1:
+        # TODO: handle this case
+        continue
```
pradh, line 115: nit: update the counter with unsupported (just in case we see it while debugging)

server/lib/nl/fulfillment/fulfiller.py
```diff
@@ -161,7 +169,13 @@ def _perform_strict_mode_checks(uttr: Utterance) -> bool:
 
 
 def _produce_query_types(uttr: Utterance) -> List[QueryType]:
-  query_types = [handlers.first_query_type(uttr)]
+  query_types = []
+  if params.is_bio(uttr.insight_ctx) and uttr.entities and uttr.properties:
+    query_types.append(QueryType.TRIPLE)
+  # The remaining query types require places to be set
+  if not uttr.places:
+    return query_types
```
pradh, line 177: To confir, even if set, this is not handled in the callsite now, right?

server/lib/nl/fulfillment/triple.py
```diff
@@ -0,0 +1,49 @@
...
+#
+# Handler for TRIPLE queries.
+#
+
+
+def populate(uttr: nl_uttr.Utterance) -> bool:
+  if not uttr.entities:
+    uttr.counters.err('triple_failed_no_entities', 1)
+  if not uttr.properties:
+    uttr.counters.err('triple_failed_no_properties', 1)
```
pradh, line 31: Don't we want to return?

server/lib/nl/fulfillment/triple.py
```diff
@@ -0,0 +1,49 @@
...
+    if not any(prop_values.values()):
+      continue
+    state = PopulateState(uttr=uttr)
+    chart_vars = ChartVars(props=[prop])
+    return add_chart_to_utterance(ChartType.ANSWER_WITH_ENTITY_OVERVIEW,
+                                  state,
+                                  chart_vars, [],
+                                  entities=uttr.entities)
+
+  uttr.counters.err('triple_failed_no_properties_exist', 1)
```
pradh, line 48: nit: suggest moving this to 36 / 40 with the property that doesn't exist instead of `1`

pradh (approved):
Thanks for the updates!  Exciting!

server/lib/nl/fulfillment/triple.py
chejennifer: done

server/lib/nl/fulfillment/triple.py
chejennifer: yes, you're right

server/lib/nl/fulfillment/fulfiller.py
chejennifer: these other query types will only be handled if the triple handler fails AND there are both places and svs

server/lib/nl/config_builder/builder.py
chejennifer: done

server/lib/nl/common/utterance.py
chejennifer: done

server/config/subject_page.proto
chejennifer: done, switched it in this PR

pradh (approved):
Thank you!

chejennifer:
thanks for the review!

PR #4026, opened 2024-03-12
Filter 'per #' SVs from per capita

n-h-diaz:
Adds some rows to filter based on querying BQ (these substrings come from https://github.com/datacommonsorg/website/blob/master/server/lib/shared.py#L197) 

```
select id from `datcom-store.dc_kg_latest.StatisticalVariable`
where measurement_denominator is not null
and id not like "%Temperature%"
and id not like "%Precipitation%"
and id not like "%BarometricPressure%"
and id not like "%CloudCover%"
and id not like "%PrecipitableWater%"
and id not like "%Rainfall%"
and id not like "%Snowfall%"
and id not like "%Visibility%"
and id not like "%WindSpeed%"
and id not like "%ConsecutiveDryDays%"
and id not like "%Percent%"
and id not like "%Area_%"
and id not like "%Median_%"
and id not like "%LifeExpectancy_%"
and id not like "%AsFractionOf%"
and id not like "%AsAFractionOf%"
and id not like "%UnemploymentRate_%"
and id not like "%Mean_Income_%"
and id not like "%GenderIncomeInequality_%"
and id not like "%FertilityRate_%"
and id not like "%GrowthRate_%"
and id not like "%sdg/%"
and id not like "%FemaNaturalHazardRiskIndex_%"
and id not like "%FemaCommunityResilience_%"
and id not like "%FemaSocialVulnerability_%"
and id not like "%MothersAge_%"
and id not like "%IntervalSinceLastBirth_%"
and id not like "%BirthWeight_%"
and id not like "%Covid19MobilityTrend_%"
and id not like "%Average_%"
and id not like "%Cancer_Risk%"
and id not like "%IncrementalCount_%"
and id not like "%HouseholdSize_%"
and id not like "%LmpGestationalAge_%"
and id not like "%who/%"
and id not like "%WHO/%"
order by id
```

This outputs 27 SVs

Then I merged this with the existing list (if part of 27, modify existing row to 'n', else add row) 

This also filters all WHO SVs from per capita

(Now "See per capita" does not show for the example in the World Bank bug) 
Screenshot 2024-03-12 at 11 17 39 AM


server/config/nl_page/nl_vars_percapita_ranking.csv
```diff
@@ -1868,6 +1883,7 @@ Count_Household_HasComputer,,,
 Count_Household_HasComputer_NoInternetAccess,,,
 Count_Household_HasComputer_WithInternetSubscription_BroadbandInternetSuchAsCableFiberOpticOrDsl,,,
 Count_Household_HasComputer_WithInternetSubscription_DialUpInternetOnly,,,
+Count_Household_Homeless,n,,
```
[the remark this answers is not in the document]
n-h-diaz: The name is "Number of homeless households per 10,000 households" - should I also use the populationType to filter?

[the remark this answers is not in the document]
pradh: Ok, that's poorly modeled SV :(.  Left a comment in b/328603978#comment10

Just using populationType won't catch this.

I think for schemaful we could maybe double-check with schema (has measurementDenominator), and for schema-less go by description?

server/config/nl_page/nl_vars_percapita_ranking.csv
```diff
@@ -4193,74 +4218,26 @@ CumulativeCount_MedicalConditionIncident_COVID_19_ConfirmedCase,,,
 CumulativeCount_MedicalConditionIncident_COVID_19_ConfirmedOrProbableCase,,,
 CumulativeCount_MedicalConditionIncident_COVID_19_PatientDeceased,,,
 CumulativeCount_MedicalTest_ConditionCOVID_19,,,
-CumulativeCount_Vaccine_COVID_19_Administered,,,
+CumulativeCount_Vaccine_COVID_19_Administered,n,,n
```
pradh, line 4221: Why `n` here?

pradh (commented):
Thanks!

Hard to tell what's changed, so sent out a PR for sorting things first: https://github.com/datacommonsorg/website/pull/4032

Could you PTAL at that, then we can sync and see what's being really added/deleted?

server/config/nl_page/nl_vars_percapita_ranking.csv
n-h-diaz: This SV is listed twice in the file (also at the very end) - which line is correct? @pradh

server/lib/util.py
```diff
@@ -348,7 +348,6 @@ def get_nl_chart_titles():
 # Returns a set of SVs that should not have Per-capita.
 # TODO: Eventually read this from KG.
 def get_nl_no_percapita_vars():
-  # NOTE: This is a checked-in version of https://shorturl.at/afpMY
```
pradh, line 351: In its place, could you please link to instructions on regenerating this that you used?

pradh (approved):
Thanks!  The diffs look much better now!

PR #4035, opened 2024-03-14
[biomed nl] basic frontend to handle entity/prop answer

chejennifer:
Add some basic frontend to be able to display the answer from a biomed nl query. This adds:
- answer message tile
- entity overview tile (without the actual data, just a placeholder that says this is the entity overview tile)
- remove original entity pv config stuff (started by using entity pv config to handle biomed nl, but decided we want to include biomed nl response in the same config as regular nl charts)

Screenshot 2024-03-14 at 9 08 49 AM


static/js/components/tiles/answer_message_tile.tsx
```diff
@@ -0,0 +1,122 @@
...
+
+const fetchData = (
+  props: AnswerMessageTilePropType
+): Promise<DisplayValueSpec> => {
+  if (props.displayValue || _.isEmpty(props.property)) {
+    return Promise.resolve(props.displayValue);
+  }
+
+  return axios
+    .get(`/api/node/propvals/${getPropertyDirection(props.property)}`, {
```
dwnoble, line 85: Would you be up for giving `@datacommonsorg/client` a try here?

Import from here: 
https://github.com/datacommonsorg/website/blob/341c5bc00707eb2f7cc5a6e4ef3c0c481db2a34c/static/js/utils/datacommons_client.ts#L17

Call this: 
https://github.com/datacommonsorg/website/blob/master/packages/client/src/data_commons_web_client.ts#L49

static/js/components/tiles/answer_message_tile.tsx
```diff
@@ -0,0 +1,122 @@
...
+      params: { dcids: [props.entity], prop: props.property.property },
+      paramsSerializer: stringifyFn,
+    })
+    .then((resp) => {
```
dwnoble, line 89: Nit: use async/await to flatten nested calls

static/js/components/tiles/answer_message_tile.tsx
chejennifer: it looks like datacommons client only handles `out` properties right now and this tile should be able to handle both `in` and `out` .. so if we want to use datacommons client, maybe we can do it in a separate pr where we also update the client to handle in properties

static/js/components/tiles/answer_message_tile.tsx
```diff
@@ -0,0 +1,122 @@
...
+      const respValues = resp.data[props.entity] || [];
+      const values: Set<string> = new Set();
+      const provIds: Set<string> = new Set();
+      respValues.forEach((respVal) => {
+        values.add(respVal.name || respVal.dcid || respVal.value);
```
juliawu, line 94: Is there ever a case where `dcid` and `value` are both present in respVal? If so, should `value` come before `dcid` instead?

static/js/components/tiles/answer_message_tile.tsx
```diff
@@ -0,0 +1,122 @@
...
+    return null;
+  }
+
+  return (
+    <div
+      className={`chart-container answer-message-tile ${ASYNC_ELEMENT_HOLDER_CLASS}`}
+    >
+      <div className={`answer-message ${ASYNC_ELEMENT_CLASS}`}>
+        <span>{props.title}</span>
+        <span>{answerData.value.join(", ")}</span>
```
juliawu, line 70: should we make `DisplayValueSpec` have plural `values` and `sources` properties instead? Was a little confusing at first to see value (singular) with .join().

juliawu (commented):
Thanks! Just a few minor comments.

static/js/components/tiles/answer_message_tile.tsx
dwnoble: sounds good!

dwnoble (approved):

static/js/components/tiles/answer_message_tile.tsx
chejennifer: done

static/css/shared/_tiles.scss
```diff
@@ -598,3 +598,35 @@ $box-shadow-8dp: 0 8px 10px 1px rgba(0, 0, 0, 0.14),
...
+
+/**
+ * Entity overview tile
+ */
+
+ .entity-overview-tile {
+  border: var(--border-primary);
+  border-radius: var(--border-radius-primary);
+  padding: 32px;
+ }
```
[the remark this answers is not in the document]
chejennifer: done

static/js/components/tiles/answer_message_tile.tsx
chejennifer: there should not be a case like this, but switched them just in case

static/js/components/tiles/answer_message_tile.tsx
chejennifer: good call, updated the repeated proto fields to be plural

juliawu (approved):

PR #4036, opened 2024-03-14
[biomed nl] some followups from basic detect/fulfill

chejennifer:
- Add ability to get entity/prop from past query
- Update llm fallback to not fallback if there is entity/prop detected
- Update detection to not remove entities from query when detecting svs


server/lib/nl/detection/context.py
```diff
@@ -100,12 +100,17 @@ def merge_with_context(uttr: nl_uttr.Utterance, default_place: Place = None):
...
+  # 7. Detect properties leveraging context
+  properties = _detect_props(uttr)
+
   # 6. Populate the returned dict
   data_dict.update({
       Params.ENTITIES.value:
-          places,
-      Params.VARS.value:
-          main_vars[:_MAX_RETURNED_VARS],
+          places + entities,
```
pradh, line 112: hmm, why are we merging them?  Should we keep them separate so "fulfill" can do the right thing?

server/lib/nl/detection/context.py
```diff
@@ -217,8 +222,9 @@ def _detect_places(uttr: nl_uttr.Utterance,
       if _handle_answer_places(uttr, child_type, places, cmp_places):
         return places, cmp_places
     else:
-      # There are NO places, so if there are answer places, bail.
-      if _handle_answer_places(uttr, child_type, places, cmp_places):
+      # There are NO places, so if there are answer places or entities, bail.
+      if _handle_answer_places(uttr, child_type, places,
```
pradh, line 226: update comment on why the change?  I don't follow

server/lib/nl/detection/context.py
```diff
@@ -327,3 +334,35 @@ def _append(src: List[Place], dst: List[str]):
...
+
+
+def _detect_props(uttr: nl_uttr.Utterance) -> List[str]:
+  props = []
+  if uttr.properties:
+    props = uttr.properties
+    uttr.properties_source = nl_uttr.FulfillmentResult.CURRENT_QUERY
+  # If svs were detected in the current query, don't try to use any past entities
+  elif uttr.svs and uttr.sv_source != nl_uttr.FulfillmentResult.PAST_QUERY:
+    uttr.entities_source = nl_uttr.FulfillmentResult.UNFULFILLED
```
pradh, line 362: or UNRECOGNIZED?

server/lib/nl/detection/context.py
```diff
@@ -327,3 +334,35 @@ def _append(src: List[Place], dst: List[str]):
...
       dst.append(p.dcid)
+
+
+def _detect_entities(uttr: nl_uttr.Utterance) -> List[str]:
+  entities = [e.dcid for e in uttr.entities]
+  if uttr.entities:
+    uttr.entities_source = nl_uttr.FulfillmentResult.CURRENT_QUERY
+  # If places were detected in the current query, don't try to use any past entities
+  elif uttr.places and uttr.place_source != nl_uttr.FulfillmentResult.PAST_QUERY:
+    uttr.entities_source = nl_uttr.FulfillmentResult.UNFULFILLED
```
pradh, line 345: or UNRECOGNIZED?

server/lib/nl/detection/context.py
pradh, line 362: should this be properties_source?

server/lib/nl/detection/place.py
```diff
@@ -66,7 +70,7 @@ def detect_from_query_dc(orig_query: str,
           d['dcid'] for d in item['places'] if 'dcid' in d
       ]
     else:
-      nonplace_query_parts.append(item['span'].lower())
+      query_parts_with_dcids.append(item['span'].lower())
```
pradh, line 73: Add comment on what this else means?  Is the span missing DCID vs. non-place thing?

server/lib/nl/detection/place.py
```diff
@@ -85,6 +90,16 @@ def detect_from_query_dc(orig_query: str,
     parent_places = parent_map.get(main_place.dcid, [])
 
   # Set PlaceDetection.
+  places_str = [
```
pradh, line 93: nit: for readability, do procedural?

```
places_str = []
for ...
```

pradh (commented):
Nice!

1.  Sorry for the delay -- I'll be quicker with future PR rounds!
2.  We should test all the demo scripts, the context logic is subtle and I tripped a lot before :)
3.  If you prefer getting the PR in quicker, feel free to decouple the context change out of this, and keep the LLM fallback and detection change.

server/lib/nl/detection/context.py
chejennifer: I had originally thought to merge them because I know we eventually want to move towards only having entities and only having variables & right now, it looks like the insight context is just used for generating place + topic urls on the front end. But maybe makes more sense to separate them for now and then merge later when we do want to merge entities + places, and variable + properties. Updated

server/lib/nl/detection/context.py
chejennifer: The change is that if entities were detected, we skip getting place from context, which the comment was updated with "... or entities, bail"

server/lib/nl/detection/context.py
chejennifer: done

server/lib/nl/detection/context.py
chejennifer: done

server/lib/nl/detection/context.py
chejennifer: good catch, done

server/lib/nl/detection/place.py
chejennifer: done

server/lib/nl/detection/place.py
chejennifer: done

chejennifer:
Address PR comments, but still need to test the demo scripts

chejennifer:
Tested the demo scripts and they all still work!

server/lib/nl/detection/context.py
```diff
@@ -327,3 +340,35 @@ def _append(src: List[Place], dst: List[str]):
...
+
+def _detect_entities(uttr: nl_uttr.Utterance) -> List[str]:
+  entities = [e.dcid for e in uttr.entities]
+  if uttr.entities:
+    uttr.entities_source = nl_uttr.FulfillmentResult.CURRENT_QUERY
+  # If places were detected in the current query, don't try to use any past entities
+  elif uttr.places and uttr.place_source != nl_uttr.FulfillmentResult.PAST_QUERY:
+    uttr.entities_source = nl_uttr.FulfillmentResult.UNRECOGNIZED
+  else:
+    if uttr.prev_utterance and uttr.prev_utterance.entities:
```
pradh, line 353: nit: add comment for else flow

server/lib/nl/detection/place.py
```diff
@@ -37,10 +37,13 @@ def detect_from_query_dc(orig_query: str,
 
   query_items = dc.recognize_places(query)
 
-  nonplace_query_parts = []
-  places_str = []
   mains = []
   main2corrections = {}
+  # Dictionary of dcid to the query substring that matched to the dcid
```
pradh, line 42: I was confused before, and now I see what you're trying to do.  Would it be more direct to do:

```
class QueryPart:
  substr: str
  dcid: str = ''

query_parts: List[QueryPart] = []
```

instead of a list and a map?

pradh (approved):
Thanks for the changes!   LG pending demo query checks.

server/lib/nl/detection/place.py
chejennifer: i like that, updated

pradh (approved):

chejennifer:
thanks for the review!

PR #4038, opened 2024-03-14
Use regular pip install for torch+cpu package; Deprecate light-the-torch

shifucun:
light-the-torch can not specify the torch version, which is a tricky when matching sentence-transformer and other ml library packages


chejennifer (approved):
cool! Curious what the difference between light-the-torch and regular pip install is?

shifucun:
> cool! Curious what the difference between light-the-torch and regular pip install is?

light-the-torch can not specify the torch version, which is a tricky when matching sentence-transformer and other ml library packages.

build/web_compose/Dockerfile
```diff
@@ -64,8 +64,8 @@ RUN pip3 install -r /workspace/server/requirements.txt
 COPY nl_app.py /workspace/nl_app.py
 COPY deploy/nl/. /datacommons/nl/
 COPY nl_server/requirements.txt /workspace/nl_server/requirements.txt
-RUN pip install --upgrade pip setuptools light-the-torch
-RUN ltt install torch --cpuonly
+RUN pip install --upgrade pip
+RUN pip3 install torch==2.2.1 --extra-index-url https://download.pytorch.org/whl/cpu
```
pradh, line 68: To check, this will increase the docker image (that was the original motivation for light-the-torch) how much?

pradh (approved):
Thanks!

build/web_compose/Dockerfile
shifucun: Nope, this also installs cpu only torch and the size is same.

shifucun: I should have found this way instead of light-the-torch :)

PR #4043, opened 2024-03-15
Use GCP logging client so python logs can show correct severity level

shifucun:
With this, we should use logging.error()/warning()/info() distinctively to better use the logging filtering.


chejennifer (approved):
nice, this will be helpful!

run_test.sh
```diff
@@ -186,7 +186,7 @@ function update_integration_test_golden {
 
   export ENV_PREFIX=Autopush
   python3 -m pytest -vv server/integration_tests/topic_cache
-  # python3 -m pytest -vv server/tests/nodejs_e2e_test.py
+  python3 -m pytest -vv server/tests/nodejs_e2e_test.py
```
pradh, line 189: Thank you for restoring!

server/__init__.py
```diff
@@ -226,6 +225,19 @@ def register_routes_common(app):
 def create_app(nl_root=DEFAULT_NL_ROOT):
   app = Flask(__name__, static_folder='dist', static_url_path='')
 
+  cfg = lib_config.get_config()
+
+  if not cfg.LOCAL:
+    client = google.cloud.logging.Client()
+    client.setup_logging()
+  logging.basicConfig(
```
pradh, line 233: future PR is fine, should we do this logging config in a util and reuse from everywhere

pradh (approved):
Very nice!

server/__init__.py
shifucun: The cfg is for the web server only. Now nl does not have it. Should look into a general way to check if running in a GCP environment..

PR #4048, opened 2024-03-18
Use simple accuracy score for model and stat var description evaluation; Update some golden statvars for queries.

shifucun:
NDCG score is used for ranking items evaluation. In our case, we don't have a strict ranking criteria for matching stat vars. So treat every matched stat vars the same and only look at match accuracy is more robust at this stage.


static/js/apps/nl_eval/search_result.tsx
```diff
@@ -120,18 +120,16 @@ export function SearchResult(props: SearchResultProps): JSX.Element {
     props.overrideStatVars,
   ]);
 
-  // Crop the rankedStatVars to the length of goldenStatVars. Need to evaluate
-  // this further.
-  const evalScore = ndcg(
-    rankedStatVars.slice(0, props.goldenStatVars.length).map((x) => x.dcid),
+  const evalScore = accuracy(
+    rankedStatVars.map((x) => x.dcid),
```
pradh, line 124: Any comments on why this score is better than ncdg?

pradh (approved):

static/js/apps/nl_eval/search_result.tsx
shifucun: Updated in the PR comment.

PR #4051, opened 2024-03-19
[biomed nl] allow in arcs

chejennifer:
Add fulfillment handling for properties that exist in the 'in' direction

Note: Updating title formatting is on the to-do list. Currently just displaying something to show that the flow is working.

Screenshot 2024-03-19 at 4 34 08 PM


server/lib/nl/config_builder/answer.py
```diff
@@ -17,19 +17,32 @@
...
+  return prop, _ARROW_TO_DIRECTION.get(arrow)
 
 
 def answer_message_block(builder: base.Builder, cspec: ChartSpec):
-  prop = cspec.props[0]
   entity = cspec.entities[0]
-  title = _SINGULAR_TITLE.format(property=prop,
-                                 entity=entity.name or entity.dcid)
+  prop, direction = _get_prop_and_direction(cspec.props[0])
+  title_format_str = _OUT_TITLE if direction == PropertySpec.PropertyDirection.OUT else _IN_TITLE
```
pradh, line 39: nit: maybe a helper to take entity, prop, direction and return a title?

server/lib/nl/fulfillment/triple.py
```diff
@@ -23,6 +23,42 @@
...
+    return None
+  prop_with_direction = f'{_OUT_ARROW}{prop}' if out else f'{_IN_ARROW}{prop}'
+  chart_vars = ChartVars(props=[prop_with_direction])
+  return chart_vars
+
+
+# Handles properties with no direction specified
+def _populate_any_direction_prop(uttr: nl_uttr.Utterance, prop: str):
+  out_chart_vars = _get_chart_vars(uttr, prop, True)
+  in_chart_vars = _get_chart_vars(uttr, prop, False)
```
pradh, line 45: TODO:  Try to do concurrent calls / prefer index to use properties with arrows?

server/lib/nl/fulfillment/triple.py
```diff
@@ -37,15 +73,11 @@ def populate(uttr: nl_uttr.Utterance) -> bool:
...
-    chart_vars = ChartVars(props=[prop])
-    return add_chart_to_utterance(ChartType.ANSWER_WITH_ENTITY_OVERVIEW,
-                                  state,
-                                  chart_vars, [],
-                                  entities=uttr.entities)
+    else:
+      if _populate_any_direction_prop(uttr, prop):
+        # Currently only handling one prop, so if populate succeeds for any one
+        # prop, return.
+        return True
```
pradh, line 80: Any specific reason why that is?

server/lib/nl/fulfillment/triple.py
```diff
@@ -23,6 +23,42 @@
...
+
+# Handles properties with no direction specified
+def _populate_any_direction_prop(uttr: nl_uttr.Utterance, prop: str):
+  out_chart_vars = _get_chart_vars(uttr, prop, True)
+  in_chart_vars = _get_chart_vars(uttr, prop, False)
+  chart_added = False
+  state = PopulateState(uttr=uttr)
+  # Skip the entity overview in the first chartspec if we got chart vars
+  # for both the in and out direction
+  skip_first_overview = out_chart_vars and in_chart_vars
```
pradh, line 50: Would it be cleaner to break it up the cspec into ANSWER and ENTITY_OVERVIEW, so we add, ENTITY_OVERVIEW for the entity once always?

server/lib/nl/fulfillment/triple.py
```diff
@@ -23,6 +23,42 @@
...
+_OUT_ARROW = '->'
+_IN_ARROW = '<-'
+
+
+# Gets the chart vars for a property and direction if values exist for that
+# property and direction,
+def _get_chart_vars(uttr: nl_uttr.Utterance, prop: str, out: bool) -> ChartVars:
+  entity_dcids = [e.dcid for e in uttr.entities]
+  prop_values = raw_property_values(entity_dcids, prop, out)
+  if not any(prop_values.values()):
```
pradh, line 35: Update counter?

pradh (commented):
Nice!

server/lib/nl/config_builder/answer.py
chejennifer: done

server/lib/nl/fulfillment/triple.py
chejennifer: done

server/lib/nl/fulfillment/triple.py
chejennifer: no specific reason, but easiest to handle for now & from the mocks, it looks like we're trying to answer the user's question with a single answer OR if user asks for multiple props in the query, then we show the table, but I think that would be similar to a multi-var case

server/lib/nl/fulfillment/triple.py
chejennifer: done

server/lib/nl/fulfillment/triple.py
chejennifer: done

server/lib/nl/fulfillment/triple.py
```diff
@@ -68,16 +65,23 @@ def populate(uttr: nl_uttr.Utterance) -> bool:
...
         # Currently only handling one prop, so if populate succeeds for any one
         # prop, return.
-        return True
-      else:
-        uttr.counters.err('triple_property_failed_existence', prop)
-  return False
+        break
+  # If we are populating for a single entity and we've added an answer for this
+  # entity, also add an entity overview chart
+  if chart_added and len(uttr.entities) == 1:
```
pradh, line 82: Even if `chart_added` is false, does it make sense to show ENTITY_OVERVIEW (with a message about property not existing)?

Feel free to add TODO for follow up, if you agree.

server/lib/nl/fulfillment/triple.py
pradh: I see.  Have a TODO to revisit this?  Because I think for the general case matching multiple props could be useful.

server/lib/nl/fulfillment/triple.py
```diff
@@ -68,16 +65,23 @@ def populate(uttr: nl_uttr.Utterance) -> bool:
...
     else:
       if _populate_any_direction_prop(uttr, prop):
+        chart_added = True
         # Currently only handling one prop, so if populate succeeds for any one
         # prop, return.
-        return True
-      else:
-        uttr.counters.err('triple_property_failed_existence', prop)
-  return False
+        break
```
pradh, line 79: Why drop the `else` triple_property_failed_existence counter update?

pradh: In fact, it might also be worth considering messaging if the top match(es) fail existence? (TODO?)

pradh (approved):

server/lib/nl/fulfillment/triple.py
chejennifer: dropped the counter here because I moved to updating the counter now for each direction of the property, so I don't think we need another counter update here? Given a prop A, we will try to get values for both its in and out values where we will also update the triple_property_failed_existence counter if there's no values for each of '->A' and '<-A' .. 

& added todo

server/lib/nl/fulfillment/triple.py
chejennifer: I think this is something we will need to revisit because if we eventually interleave entity/prop charts with the regular sv charts, I don't think it would make sense to add the entity overview here. Rather, we might want to add the entity overview at the very end of all the charts? Added a TODO to revisit

server/lib/nl/fulfillment/triple.py
chejennifer: done

chejennifer:
thanks for the review!

PR #4058, opened 2024-03-21
[biomed nl] Handle property as a relation expression

chejennifer:
Instead of using property, direction, and type filtering to define a property, use a relation expression for biomed nl (https://docs.datacommons.org/api/rest/v2#relation-expressions).

Changes include:
- update the protobuf to pass a property expression for answer tiles & update fulfillment and frontend accordingly
- new flask endpoint to get values for a property expression


server/routes/shared_api/node.py
```diff
@@ -45,3 +51,59 @@ def get_property_value(direction):
...
+
+@bp.route('/propvals', methods=['GET', 'POST'])
+def expression_property_value():
+  """
+  Returns the property values for given node dcids and a relation expression for
+  a property (https://docs.datacommons.org/api/rest/v2#relation-expressions)
+  """
+  dcids = request.args.getlist('dcids')
+  if not dcids:
+    dcids = request.json['dcids']
```
shifucun, line 64: nit if this is a get request but dcids are not specified, request.json is None?

server/routes/shared_api/node.py
```diff
@@ -45,3 +51,59 @@ def get_property_value(direction):
...
+    prop_expression = request.json['propExpr']
+  if not dcids:
+    return 'error: must provide a `dcids` field', 400
+  if not prop_expression:
+    return 'error: must provide a `propExpr` field', 400
+  response = get_expression_property_value(dcids, prop_expression)
+  return Response(json.dumps(response), 200, mimetype='application/json')
+
+
+def get_expression_property_value(dcids, expression):
```
shifucun, line 76: property_value_from_expression?

server/routes/shared_api/node.py
```diff
@@ -45,3 +51,59 @@ def get_property_value(direction):
...
+  """
+  Returns the property values for given node dcids and a relation expression for
+  a property (https://docs.datacommons.org/api/rest/v2#relation-expressions)
+  """
+  # Get a list of single property relation expressions (the expression passed to
+  # this function can chain multiple relation expressions)
+  expression_matches = re.finditer(_PROPERTY_EXPRESSION_RE, expression)
```
shifucun, line 83: Maybe define a wrapper function to do re.finditer and iterate through matches? And can have a unit test for it.

The data fetch will be based on the match data.

shifucun, line 83: Can you give an example of the chaining here? Given the complexity of the logic, I feel this function worth to have a test.

server/routes/shared_api/node.py
```diff
@@ -45,3 +51,59 @@ def get_property_value(direction):
...
+      # Get the original node dcid these values should apply to
+      orig_dcid = curr2orig_dcids[dcid]
+      results[orig_dcid] = results.get(orig_dcid, []) + resp.get(dcid, [])
+      # for each value that has a dcid, add it to curr_dcids for next round of
+      # property value fetching
+      for val in resp.get(dcid, []):
+        val_dcid = val.get('dcid', '')
+        if not val_dcid:
+          continue
+        curr_dcids.append(val_dcid)
```
shifucun, line 107: why do you expand curr_dcids here for the next iteration of match?

server/routes/shared_api/node.py
```diff
@@ -45,3 +51,59 @@ def get_property_value(direction):
...
+  """
+  # Get a list of single property relation expressions (the expression passed to
+  # this function can chain multiple relation expressions)
+  expression_matches = re.finditer(_PROPERTY_EXPRESSION_RE, expression)
+  results = {}
+  # the list of dcids to use to fetch the next round of property values
+  curr_dcids = dcids
+  # map of dcids from curr_dcids to the original dcid passed to this function
+  curr2orig_dcids = {dcid: dcid for dcid in curr_dcids}
+  for match in expression_matches:
```
shifucun, line 89: I am a bit confused of the logic here. Can you comment with a concrete example and the steps to do the data fetch?

chejennifer: not sure if you wanted an example here or in the code, but I'll give the example here and I can add it to the code as well:

node dcids: [rs13317, rs7903146]
prop expression: <-referenceSNPClusterID{typeOf:GeneticVariantGeneAssociation}->geneSymbol

1. we have 2 expression matches:
     - <-referenceSNPClusterID{typeOf:GeneticVariantGeneAssociation}
     - ->geneSymbol
2. For the first expression match, we fetch data using:
`fetch.raw_property_values(nodes=[rs13317, rs7903146], prop=referenceSNPClusterID, out=False, constraints='{typeOf:GeneticVariantGeneAssociation}')`
3. Say from that data fetch, we get back:
`{rs13317: [{dcid: a}, {dcid: b}], rs7903146: [{dcid: e}, {dcid: f}]}`
6. For the next expression match, we fetch data using 
`fetch.raw_property_values(nodes=[a, b, e, f], prop=geneSymbol, out=True, constraints='')`
7. Say from that data fetch, we get back 
`{a: [{dcid: aa}, {dcid: bb}], b: [{dcid: cc}, {dcid: dd}], e: [{dcid: ee}, {dcid: ff}], f: [{dcid: gg}, {dcid: hh}]}`
8. In the final result, we want
`{rs13317:  [{dcid: aa}, {dcid: bb}, {dcid: cc}, {dcid: dd}], rs7903146: [{dcid: ee}, {dcid: ff}, {dcid: gg}, {dcid: hh}]}`

shifucun: I see. thanks for the detailed example! I wonder if you want to have 

- one function that: parse expression, fetch data and return dcid2dcid map.
- one function that takes a list of such mappings and join them together to get the final result?

server/routes/shared_api/node.py
chejennifer: done

server/routes/shared_api/node.py
chejennifer: added unit test

server/routes/shared_api/node.py
chejennifer: added check

server/routes/shared_api/node.py
chejennifer: added a test that has a chaining example

server/routes/shared_api/node.py
chejennifer: explained the logic in later comment

server/routes/shared_api/node.py
chejennifer: I tried splitting out the functions but not sure if it's how you envisioned

server/routes/shared_api/node.py
shifucun: I was thinking one function that does the data fetch for a single expression only, that will be easy for testing.

And have a wrapper to chain the result (recursively) so something like below (naming could be totally bad though, just to give idea):

```
def get_pv_from_simple_expr(dcids, expr):
    return sub2obj_map
```

```
def parse_expr(expr_str):
   ... 
   return List[expr_obj]
```

```
get_pv_from_expr(dcids, List[expr_obj]):
  expr_list = parse(expr_str)
  if len(expr_list) == 0:
     return {dcid: dcid for dcid in dcids}
  # get direct property
  sub2obj_direct = get_pv_from_simple_expr(dcids, expr_list[0])
  # get all the rest
  sub2obj_remote = get_pv_from_expr(sub2obj_direct.values(), expr_list[1:])
  return {sub: sub2obj_remote[obj]  for sub, obj in sub2obj_direct.items()}
```

chejennifer: updated

server/routes/shared_api/node.py
```diff
@@ -45,3 +62,95 @@ def get_property_value(direction):
...
+  # Generate results by mapping the results for the remaining props to their
+  # original dcids
+  result = {}
+  for dcid, values in first_prop_result.items():
+    # For each value in this result, get the values for the remaining props
+    result_values = []
+    for val in values:
+      val_dcid = val.get('dcid', '')
+      if val_dcid not in remaining_prop_results:
+        continue
```
shifucun, line 124: no need to check since line 125 would return [] anyway?

shifucun (approved):
thanks for the update, esp the unit test!

server/routes/shared_api/node.py
chejennifer: you're right, updated

chejennifer:
thanks for the review!

PR #4061, opened 2024-03-22
Only perform GCP related operations when it's running on GCP

shifucun:


server/lib/gcp.py
```diff
@@ -0,0 +1,27 @@
...
+
+url = "http://metadata.google.internal/computeMetadata/v1/project/project-id"
+
+
+def in_google_network():
+  try:
+    req = urllib.request.Request(url, headers={"Metadata-Flavor": "Google"})
+    resp = urllib.request.urlopen(req)
+    resp.read().decode()
+    return True
```
keyurva, line 25: Can we cache this in-memory so the request is made only once per session?

server/lib/gcp.py
```diff
@@ -0,0 +1,27 @@
...
+  except Exception:
+    return False
```
keyurva, line 27: Perhaps log an info level exception (actually should log the true case as well)? If can help in troubleshooting in case there is a real issue. Assuming the value is cached, it will be logged only once.

server/lib/gcp.py
shifucun: I think this function as util should not cache but the caller should. That said, it is only invoked once in create_app() function during server start, so should be okay.

server/lib/gcp.py
shifucun: Done

keyurva: nit: Wondering if this should be a warning or info? I can argue either way - go with what feels right.

server/lib/gcp.py
keyurva: Currently, yes. It could be called from multiple places in the future. Perhaps add a comment to cache if that happens.

keyurva (approved):
Thanks for the fix!

server/lib/gcp.py
shifucun: Done

server/lib/gcp.py
shifucun: yea, changed to info!

PR #4067, opened 2024-03-26
Disabled nodejs test to avoid dependency on dev

dwnoble:


server/tests/nodejs_e2e_test.py
```diff
@@ -11,6 +11,9 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+
+# Disabled nodejs test to avoid dependency on dev 
+"""
```
pradh, line 16: Should we instead comment out here: https://github.com/datacommonsorg/website/blob/dd82de42a2906867b7de1d8a2469c42804824025/run_test.sh#L189?

dwnoble: I think it will still get triggered from this line: https://github.com/datacommonsorg/website/blob/dd82de42a2906867b7de1d8a2469c42804824025/run_test.sh#L101

juliawu: Drive by comment: +1 on shanth's comment. Commenting out in `run_test.sh` and using the `--ignore=server/tests/nodejs_e2e_test.py` pytest flag is cleaner to me.

dwnoble: sgtm. updated!

pradh (approved):
Thanks!

PR #4074, opened 2024-03-27
Added script to clear undata redis cache.

dwnoble:
- Added script to clear undata redis cache
- Updated other clearcache scripts to work when called from the parent website directory
- Updated `tools/clearcache/run.sh` script to accept an optional REDIS_REGION parameter (in case the redis region is different from the cluster region)


juliawu (approved):
Awesome! Thanks!

PR #4079, opened 2024-03-29
NL Server: use small spacy model instead of large

pradh:
Per SO (https://stackoverflow.com/a/57337084) and spacy website (https://spacy.io/models/en), the accuracy difference for Parts-Of-Speech task (for verb detection) between large vs. small is very small.  The verb test sanity passes.

This should cut down the memory usage per NL server a bunch, since the small model is apparently just 10MB (vs. 780MB).

I see that heap-usage drops from 970MB to 230MB.

Also, drop a super edge case handling for test environment which loads SentenceTransformer always.

### DIFF
--- a/nl_app.py
+++ b/nl_app.py
@@ -27,8 +27,6 @@ if __name__ == '__main__':
   logging.info("Run nl server in local mode")
 
   if len(sys.argv) == 3 and sys.argv[2] == 'opt':
-    log = logging.getLogger('werkzeug')
-    log.setLevel(logging.ERROR)
     app.run(host='127.0.0.1', port=int(sys.argv[1]))
   else:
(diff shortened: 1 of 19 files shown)


shifucun (commented):
Maybe can test a few cases from the initial bard review cases for the verb detection. Also I think there are unit test to cover some cases.

pradh:
There is one loss here.  Based on discussion with @shifucun, this was anyway a quick bandaid fix, and so we will go ahead and downgrade it for now to use lesser memory.  Soon, we should get rid of it altogether and migrate to a trained classifier (that's tracked in b/331939961).

shifucun (approved):

pradh:
> Maybe can test a few cases from the initial bard review cases for the verb detection. Also I think there are unit test to cover some cases.

Sanity checked a few and it does kick in.

PR #4081, opened 2024-03-29
Minor fix to avoid faulting on mode= param

pradh:
i.e., fix https://autopush.datacommons.org/explore/#q=Population%20by%20age%20in%20the%20united%20states&mode=strict

### DIFF
--- a/static/js/utils/app/explore_utils.ts
+++ b/static/js/utils/app/explore_utils.ts
@@ -40,13 +40,11 @@ export function getTopics(
   pageMetadata: SubjectPageMetadata,
   placeUrlVal: string
 ): Item[] {
-  if (_.isEmpty(pageMetadata?.childTopics)) {
-    return [];
-  }
   const topicList = [];
-  const topics = pageMetadata?.childTopics
-    .concat(pageMetadata?.peerTopics)
-    .concat(pageMetadata?.parentTopics);
+  const topics = []
+    .concat(pageMetadata?.childTopics || [])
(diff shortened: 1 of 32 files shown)


static/js/utils/app/explore_utils.ts
```diff
@@ -40,6 +40,9 @@ export function getTopics(
   pageMetadata: SubjectPageMetadata,
   placeUrlVal: string
 ): Item[] {
+  if (_.isEmpty(pageMetadata?.childTopics)) {
```
chejennifer, line 43: could there be a case where there are no childTopics but there are peerTopics or parentTopics? Which, in that case, should we just do `topics = [].concat(pageMetadata?.childTopics || []).concat(... || [])...`

chejennifer (commented):
nice, one question though

static/js/utils/app/explore_utils.ts
pradh: good catch.  ok, can I do a good'ol for loop though?  This whole `.concat(xyz || [])` feels so ugly, or is it the JS way? :)

chejennifer: haha yes, you can do a for loop

pradh: I went with your js expert suggestion after all :), the need to distinguish between null vs. empty makes it more verbose I realize now.

chejennifer (approved):
thanks for the fix!

pradh:
Thanks for the review!

PR #4082, opened 2024-03-29
Updated CSV export

dwnoble:
# Updates
- Updated CSV export for react tiles & web components
- Updated CSV export for place page tiles
- Added facetOverride to DataCommonsClient to account for SDG variable scaling factors not yet in the knowledge graph
- Added startDate/endDate filters to getCsvSeries and getDataRowSeries to accommodate line chart csv export
- Updates behavior when clicking on CSV content in ChartEmbed modal to allow for user to select a subset of the text (previously all text selections would result in the entire content being selected)
- Updated line tile's startDate & endDate filters to perform date comparison after truncating the observation date to match the startDate & endDate lengths respectively. Aligned DataCommonsClient startDate & endDate filters to match this behavior.

## Old CSV fields:
- observation date
- observation value
- variable name

## New CSV fields:
- `Entity DCID`
- `Entity isoCode`
- `Entity name`
- `Variable DCID`
- `Variable observation date`
- `Variable observation value`
- `Variable observation metadata importName`
- `Variable observation metadata provenanceUrl`
- `Variable observation metadata scalingFactor`
- `Variable observation metadata unit`
- `Variable observation metadata unitDisplayName`
- `Variable observation value`
- `Variable properties name`

## New CSV fields for per-capita exports
- `Variable perCapita dcid`
- `Variable perCapita observation date`
- `Variable perCapita observation metadata importName`
- `Variable perCapita observation metadata provenanceUrl`
- `Variable perCapita observation metadata scalingFactor`
- `Variable perCapita observation metadata unit`
- `Variable perCapita observation metadata unitDisplayName`
- `Variable perCapita observation value`
- `Variable perCapita properties name`
- `Variable perCapita quotientValue`

## Example previous csv format:
Screenshot 2024-03-29 at 3 44 22 PM

```csv
label,Coronary Heart Disease
2021,5.2
2020,5.5
2018,5.9
```

## Example new csv format:
Screenshot 2024-03-29 at 3 44 48 PM

```csv
"entity.dcid","entity.properties.isoCode","entity.properties.name","variable.dcid","variable.observation.date","variable.observation.metadata.importName","variable.observation.metadata.provenanceUrl","variable.observation.metadata.scalingFactor","variable.observation.metadata.unit","variable.observation.metadata.unitDisplayName","variable.observation.value","variable.properties.name"
"country/USA","US","United States of America","Percent_Person_WithCoronaryHeartDisease","2018","CDC500","https://www.cdc.gov/places/index.html","","","",5.9,"Percentage of Population With Coronary Heart Disease"
"country/USA","US","United States of America","Percent_Person_WithCoronaryHeartDisease","2020","CDC500","https://www.cdc.gov/places/index.html","","","",5.5,"Percentage of Population With Coronary Heart Disease"
"country/USA","US","United States of America","Percent_Person_WithCoronaryHeartDisease","2021","CDC500","https://www.cdc.gov/places/index.html","","","",5.2,"Percentage of Population With Coronary Heart Disease"
```

## Explore page examples
Screenshot 2024-03-29 at 3 57 43 PM
Screenshot 2024-03-29 at 3 57 35 PM
Screenshot 2024-03-29 at 3 57 29 PM

## Place page examples
Screenshot 2024-03-29 at 3 58 22 PM
Screenshot 2024-03-29 at 3 58 35 PM


static/js/components/tiles/sv_ranking_units.tsx
```diff
@@ -93,11 +92,11 @@ export function SvRankingUnits(props: SvRankingUnitsProps): JSX.Element {
...
       chartHeight,
       chartHtml,
-      points,
-      props.sources || Array.from(rankingGroup.sources),
-      rankingGroup.svName
+      chartTitle,
+      props.sources || Array.from(rankingGroup.sources)
     );
   }
+  const chartTitle = getDownloadFilename(title, rankingGroup);
```
juliawu, line 99: Reads a little weird to have the chartTitle as a filename. It looks like tihs is the only place getDownloadFilename gets called, and it doesn't return a file extension. Perhaps rename to getChartTitle?

static/js/components/tiles/scatter_tile.tsx
```diff
@@ -155,18 +156,29 @@ export function ScatterTile(props: ScatterTilePropType): JSX.Element {
...
-                scatterChartData.yStatVar.statVar,
-                scatterChartData.yStatVar.denom,
-                scatterChartData.points
-              )
-          : null
-      }
+      getDataCsv={() => {
+        // Assume both variables will have the same date
+        const date =
+          props.statVarSpec.length > 0 ? props.statVarSpec[0].date : undefined;
```
juliawu, line 162: What would it take to build support for fetching variables with two different dates? This assumption breaks for charts like the top of https://datacommons.org/explore/#q=housing%20vs%20obesity%20in%20US%20states

static/css/shared/_chart_embed.scss
```diff
@@ -52,4 +52,12 @@
...
   align-items: center;
-}
\ No newline at end of file
+}
+
+.modal {
+  .modal-textarea {
+    white-space: pre;
+    overflow-wrap: normal;
+    overflow-x: scroll;
```
juliawu, line 61: supernit: alphabetize

juliawu (commented):
Very nice! Cool to see the data commons client in action in the tiles.

packages/client/src/data_commons_client.test.ts
```diff
@@ -532,6 +528,69 @@ describe("DataCommonsWebClient", () => {
...
+      parentEntity: "country/MOCK",
+      perCapitaVariables: ["Has_Data"],
+      variables: ["Has_Data", "No_Data"],
+      startDate: "2020",
+    });
+    expect(response3.length).toBe(12);
+    response3.forEach((row) => {
+      const observationDate = row.variable.observation.date || "";
+      expect(observationDate >= "2020").toBeTruthy();
+      expect(observationDate <= "2023").toBeTruthy();
```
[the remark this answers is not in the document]
dwnoble: That came from just looking at the raw data, but to your point i removed `<= "2023"` to make the test more clear

packages/client/src/data_commons_client.test.ts
```diff
@@ -532,6 +528,69 @@ describe("DataCommonsWebClient", () => {
...
+      fieldDelimiter: ".",
+      parentEntity: "country/MOCK",
+      perCapitaVariables: ["Has_Data"],
+      variables: ["Has_Data", "No_Data"],
+      endDate: "2020",
+    });
+    expect(response4.length).toBe(6);
+    response4.forEach((row) => {
+      const observationDate = row.variable.observation.date || "";
+      expect(observationDate >= "2019").toBeTruthy();
```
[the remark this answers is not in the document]
dwnoble: Updated

packages/client/src/data_commons_client.test.ts
```diff
@@ -543,4 +602,60 @@ describe("DataCommonsWebClient", () => {
...
+    expect(facetOverrideResponse.length).toBe(15);
+    facetOverrideResponse.forEach((row) => {
+      expect(row.variable.dcid).toBe("Has_Data");
+      expect(row.variable.observation.metadata.unitDisplayName).toBe(
+        "US Dollars"
+      );
+      // Mock Has_Data values are set to 1/10th of the population values, so
+      // quotientValue (per-capita) should equal:
+      // quotientValue =  N / (N * 10 ) / scalingFactor
+      // Since scaling factor is 0.005, quotientValue = 0.1 / 0.05 = 2
```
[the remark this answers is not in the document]
dwnoble: Fixed

packages/client/src/data_commons_client_types.ts
```diff
@@ -59,22 +57,45 @@ export interface BaseGetDataRowsParams {
...
+  | BaseGetDataRowsParamsWithin
+  | BaseGetDataRowsParamsEntities;
+
+/**
+ * Parameters for data commons client getDataRow and getCsv methods
+ */
+export type GetDataRowsParams = BaseGetDataRowsParams & DataRowsDateFilter;
+
+/**
+ * Parameters for etDataRowSeries and getCsvSeries methods
```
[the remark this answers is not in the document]
dwnoble: Fixed

packages/client/README.md
```diff
@@ -49,7 +49,9 @@ To find the DCID of a entity or variable:
 | entities            | string[] | Entity DCIDs. Required if `parentEntity` and `childType` are empty. Example: `["country/USA", "country/IND"]`                                |
 | parentEntity        | string   | Parent entity DCID. Required if `entities` is empty.                                                                                         |
 | childType           | string   | Child entity type. Required if `entities` is empty. Example: `"State"`                                                                       |
-| date?               | string   | \[optional\] Only return observations from this date. Example: `"2023"`                                                                      |
+| date?               | string   | \[optional\] [getDataRow, getCsv, and getGeoJSON only] Only return observations from this date. Example: `"2023"`                            |
```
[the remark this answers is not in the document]
dwnoble: Updated

static/js/components/tiles/bar_tile.tsx
```diff
@@ -141,9 +142,33 @@ export function BarTile(props: BarTilePropType): JSX.Element {
       replacementStrings={getReplacementStrings(barChartData)}
       className={`${props.className} bar-chart`}
       allowEmbed={true}
-      getDataCsv={
-        barChartData ? () => dataGroupsToCsv(barChartData.dataGroup) : null
-      }
+      getDataCsv={() => {
```
chejennifer, line 145: nit: maybe make this a helper function?

static/js/components/tiles/bar_tile.tsx
```diff
@@ -141,9 +142,33 @@ export function BarTile(props: BarTilePropType): JSX.Element {
...
+        // Assume all variables will have the same date
+        const date =
+          props.variables.length > 0 ? props.variables[0].date : undefined;
+        const denoms = props.variables.map((v) => (v.denom ? v.statVar : ""));
```
chejennifer, line 149: this is slightly confusing because the variable is named `denom` but it actually holds the stat vars that have denoms? not sure if it should be updated to actually be mapping the denoms or the variable should be renamed

(here and all the other tiles)

chejennifer, line 149: maybe also want to remove empty strings?

static/js/components/tiles/donut_tile.tsx
```diff
@@ -114,9 +114,18 @@ export function DonutTile(props: DonutTilePropType): JSX.Element {
       replacementStrings={getReplacementStrings(props, donutChartData)}
       className={`${props.className} bar-chart`}
       allowEmbed={true}
-      getDataCsv={
-        donutChartData ? () => dataGroupsToCsv(donutChartData.dataGroup) : null
-      }
+      getDataCsv={() => {
+        // Assume all variables will have the same date
+        const date =
```
chejennifer, line 119: may be a good idea to extract out a `getDate` function that gets used by both the regular data fetch and the get csv call to make sure that the logic for getting date is the same in both cases. Because if you look at the fetchData call in this file, it looks like it never actually fetches data for a specific date.

(Here and in all the other tiles)

static/js/components/tiles/line_tile.tsx
```diff
@@ -144,7 +145,36 @@ export function LineTile(props: LineTilePropType): JSX.Element {
       replacementStrings={getReplacementStrings(props)}
       className={`${props.className} line-chart`}
       allowEmbed={true}
-      getDataCsv={chartData ? () => dataGroupsToCsv(chartData.dataGroup) : null}
+      getDataCsv={() => {
+        const denoms = props.statVarSpec.map((v) => (v.denom ? v.statVar : ""));
+        const entityProps = props.placeNameProp
+          ? [props.placeNameProp, ISO_CODE_ATTRIBUTE]
+          : undefined;
+        if ("enclosedPlaceType" in props && props.enclosedPlaceType) {
```
chejennifer, line 153: nit: I think you only need `props.enclosedPlaceType` here because if the field doesn't exist, `props.enclosedPlaceType` would evaluate to `undefined` which would be falsy?

static/js/components/tiles/line_tile.tsx
```diff
@@ -144,7 +145,36 @@ export function LineTile(props: LineTilePropType): JSX.Element {
...
+            childType: props.enclosedPlaceType,
+            entityProps,
+            parentEntity: props.place.dcid,
+            perCapitaVariables: _.uniq(denoms),
+            startDate: props.startDate,
+            endDate: props.endDate,
+            variables: props.statVarSpec.map((v) => v.statVar),
+          });
+        } else {
+          const entities =
```
chejennifer, line 164: hm, like with getting date from props, might also be a good idea to have a function for getting entities and use it in both the regular data fetch and the get csv since there's some logic involved.

packages/client/src/data_commons_client.ts
```diff
@@ -372,9 +391,52 @@ class DataCommonsClient {
       populationObservations
     );
 
+    // Filter by start/end date if specified
```
chejennifer, line 394: What should happen if the observation date is different granularity than the params?

case 1:
observationDate: 2023-01-06
startDate: 2023

case 2:
observationDate: 2023
startDate: 2023-01-06

packages/client/src/data_commons_client.ts
```diff
@@ -372,9 +391,52 @@ class DataCommonsClient {
...
+          return true;
+        })
+      );
+    }
+
     return Promise.resolve(dataRows);
   }
 
+  /**
+   * Overrides any of the passed in facets with values from statMetadataOverride
```
chejennifer, line 417: what is the statMetadataOverride you are referring to here? I tried searching the term in this file but could not find it, maybe update the comment

packages/client/src/data_commons_client.ts
```diff
@@ -372,9 +391,52 @@ class DataCommonsClient {
...
+   * @param facets
+   * @returns facets with overriden values
```
chejennifer, line 419: nit: overridden

packages/client/src/data_commons_client.ts
```diff
@@ -372,9 +391,52 @@ class DataCommonsClient {
...
+  overrideFacetValues(facets: FacetStore): FacetStore {
+    if (!this.facetOverride || Object.keys(this.facetOverride).length === 0) {
+      return facets;
+    }
+    const newFacets = {
+      ...facets,
+    };
+    Object.keys(newFacets).forEach((facetId) => {
+      const facet = newFacets[facetId];
+      if (facet.unit && facet.unit in this.facetOverride) {
```
chejennifer, line 430: also, are we only overriding units? if that's the case, then maybe be more specific in the naming/description of this function

packages/client/src/data_commons_client.ts
```diff
@@ -507,23 +574,33 @@ class DataCommonsClient {
...
+          );
 
           const closestPopulationObservation = this.getClosestObservationToDate(
             series.series,
             observation.date
           );
+          const scalingFactor = facet.scalingFactor || 1;
+          const quotientValue =
+            closestPopulationObservation &&
+            !_.isEmpty(closestPopulationObservation) &&
```
chejennifer, line 590: curious why you need both `closestPopulationObservation` and `!_.isEmpty(closestPopulationObservation)` checks?

chejennifer (commented):
cool!

static/js/components/tiles/sv_ranking_units.tsx
dwnoble: Done

static/js/components/tiles/scatter_tile.tsx
dwnoble: I think this assumption technically [still holds](https://github.com/datacommonsorg/website/blob/ead9e365979eb0e4f66f4981b8103e55b7a2359a/server/lib/nl/config_builder/scatter.py#L51) with the scatter plot -in this case the date is just set to `"LATEST"` or `""`

Granted I'd still like to make getCsv a little more flexible and support more of a "stat var spec" style interface - i  can do that in a follow-up.

static/css/shared/_chart_embed.scss
dwnoble: done

packages/client/src/constants.ts
```diff
@@ -0,0 +1,32 @@
...
+// Name attribute for entities and variables
+export const NAME_ATTRIBUTE = "name";
+// ISO 3166-2 code property name for place entities
+export const ISO_CODE_ATTRIBUTE = "isoCode";
+// Fetch these entity and variable properties by default
+export const DEFAULT_ENTITY_PROPS = [NAME_ATTRIBUTE, ISO_CODE_ATTRIBUTE];
+export const DEFAULT_VARIABLE_PROPS = [NAME_ATTRIBUTE];
+// GeoJSON is stored in this property name by default
+export const DEFAULT_GEOJSON_PROPERTY_NAME = "geoJsonCoordinatesDP1";
+// Delimit fields
```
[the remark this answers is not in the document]
dwnoble: fixed

static/js/components/tiles/bar_tile.tsx
dwnoble: Good callout-renamed `denoms` to `perCapitaVariables` to match the getCsv parameter name

dwnoble: Updated logic

static/js/components/tiles/donut_tile.tsx
dwnoble: Updated and added a todo to specify dates independently for each stat var

static/js/components/tiles/line_tile.tsx
dwnoble: Done, and simplified the regular fetch data logic

packages/client/src/data_commons_client.ts
dwnoble: fixed

packages/client/src/data_commons_client.ts
dwnoble: Hmm looks like the logic here deviates from the logic in line_tile.tsx
`data_commons_client.ts` -> lexicographically comparison
`line_tile.tsx` -> truncates startDate/endDate to the length of the observation date, then does a lexicographical comparison

I think the current line_tile.tsx behavior isn't quite right-- consider this example:

```
  <datacommons-line
  header="US Population in May 2010"
  places="country/USA"
  variables="Count_Person"
  colors="blue"
  startDate="2010-05-05"
  endDate="2010-05-06"
></datacommons-line>
```

Shows a chart with an observation from "2010":

Screenshot 2024-04-03 at 3 15 25 PM

But, i would expect it to only show observations strictly between 2010-05-05 and 2010-05-06, and thus no data to be shown

Open question:
- if startDate = 2014, and endDate=2015, does this mean only include observations for 2014, or include observations for 2014 and 2015?

I vote we do either 
(1) (preference) strict lexicographical comparison where startDate <= obs <= endDate , as i think this is easy enough to explain to users.
(2) instead of truncating the start/end dates to match the observation's length, truncate the observation's length to match the start/end dates length, and run the final calculation as startDate <= truncated(obs) <= endDate. this would mean startDate=2014 and endDate=2015 would show observations from any time in 2014 or 2015, which seems slightly off to me (i prefer it would just show 2014 or observations from exactly 2015, but open to other ideas)

packages/client/src/data_commons_client.ts
dwnoble: Updated

packages/client/src/data_commons_client.ts
dwnoble: It's overriding the entire facet / StatMetadata, but we know which values to override based on the facet's unit (since the FacetStore is keyed by facetId, which isn't a stable key for overriding)

dwnoble: Updated the comment to hopefully make this more clear

packages/client/src/data_commons_client.ts
dwnoble: Good catch- removed `closestPopulationObservation &&`

packages/client/src/data_commons_client.ts
chejennifer: hm, good points. I think:
- if startDate = 2014 and endDate=2015, we should include observations for both 2014 and 2015 because as per the comment, endDate is "latest date to show on the chart" https://github.com/datacommonsorg/website/blob/ead9e365979eb0e4f66f4981b8103e55b7a2359a/static/js/components/tiles/line_tile.tsx#L94-L97
- & since I think we should include observations from 2015, I prefer (2)

dwnoble: Done

static/js/components/tiles/scatter_tile.tsx
```diff
@@ -214,6 +206,42 @@ export function ScatterTile(props: ScatterTilePropType): JSX.Element {
...
+ * Returns callback for fetching chart CSV data
+ * @param props Chart properties
+ * @returns Async function for fetching chart CSV
+ */
+function getDataCsvCallback(
+  props: ScatterTilePropType,
+  scatterChartData: ScatterChartData
+): () => Promise<string> {
+  return () => {
+    // Assume both variables will have the same date
```
chejennifer, line 219: the fetchData function does not assume both variables will have the same date and we wouldn't want csv data to be different from the chart data? I did see your thread with Julia about this, but technically a user can put anything in the date field when generating the tile, so lets at least add a TODO to flag the difference. Maybe make one function for getting date and use it both here and the chart data fetch and add a TODO in the function to handle multiple dates. 

Another problem is that like map_tile, there will be some stat vars that get a special date set, so that also needs to be handled: https://github.com/datacommonsorg/website/blob/master/static/js/tools/scatter/util.ts#L72-L76

static/js/components/tiles/ranking_tile.tsx
```diff
@@ -99,22 +103,32 @@ export function RankingTile(props: RankingTilePropType): JSX.Element {
...
+    sources: string[]
   ): void {
     embedModalElement.current.show(
       "",
-      rankingPointsToCsv(rankingPoints, svNames),
+      () => {
+        const perCapitaVariables = props.variables
+          .filter((v) => v.denom)
+          .map((v) => v.statVar);
+        return datacommonsClient.getCsv({
```
chejennifer, line 115: ranking tile should also have a date? & this tile also uses the same special date logic as map and scatter

packages/client/src/utils.ts
```diff
@@ -222,7 +263,53 @@ export function dataRowsToCsv(
...
+ * @param startDate Start date string
+ * @param endDate End date string
+ * @returns boolean
+ */
+export function isDateInRange(
+  observationDate: string,
+  startDate: string | undefined,
+  endDate: string | undefined
+): boolean {
+  if (!startDate && !endDate) {
```
chejennifer, line 305: nit: i don't think this check is necessary because both of the following checks checks for the presence of a startDate/endDate

packages/client/README.md
```diff
@@ -208,7 +210,7 @@ const response = await client.getCsvSeries({
 ```
 
 ```text
-"entity.dcid","entity.properties.name","variable.dcid","variable.denominator.dcid","variable.denominator.observation.date","variable.denominator.observation.value","variable.denominator.properties.name","variable.denominator.quotientValue","variable.observation.date","variable.observation.value","variable.properties.name"
+"entity.dcid","entity.properties.name","variable.dcid","variable.perCapita.dcid","variable.perCapita.observation.date","variable.perCapita.observation.value","variable.perCapita.properties.name","variable.perCapita.perCapitaValue","variable.observation.date","variable.observation.value","variable.properties.name"
```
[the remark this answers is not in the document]
dwnoble: I went back and forth on this, and while i like the conciseness of dropping "properties", i also think it adds some clarity and avoids an (albeit unlikely) collision between property names and attributes defined in the entity & variable types

packages/client/src/data_commons_client_types.ts
```diff
@@ -21,26 +21,24 @@ import { Observation, StatMetadata } from "./data_commons_web_client_types";
  */
 
 /**
- * Denominator observation with calculated quotient value. Used for storing
- * per-capita derived values along side the original population observation.
- * "date' and "value" fields from the parent Observation interface will be set
- * to the original observation dates and values, and quotientValue is the
- * derived (per-capita) value.
+ * Observation with calculated per capita quotient value. Stores derived per
```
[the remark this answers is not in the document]
dwnoble: done

static/js/place/chart.tsx
```diff
@@ -451,7 +424,40 @@ class Chart extends React.Component<ChartPropType, ChartStateType> {
     }
     this.embedModalElement.current.show(
       svgXml,
-      this.dataCsv(),
+      () => {
```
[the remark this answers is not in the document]
dwnoble: That is a good callout. I'm inclined to just leave it with the new CSV export behavior for now, and add a TODO to address changing the denominator in a follow-up PR. I'd argue that even without the per-capita calculation, the new CSV export is still a step in the right direction since it gives more transparency as to where the data came from.

packages/client/src/utils.ts
dwnoble: Removed

static/js/components/tiles/map_tile.tsx
```diff
@@ -253,9 +256,23 @@ export function MapTile(props: MapTilePropType): JSX.Element {
...
-      getDataCsv={
-        mapChartData ? () => mapDataToCsv(mapChartData.layerData) : null
-      }
+      getDataCsv={() => {
+        const entityProps = props.placeNameProp
+          ? [props.placeNameProp, ISO_CODE_ATTRIBUTE]
+          : undefined;
+        return datacommonsClient.getCsv({
+          childType: props.enclosedPlaceType,
+          date: props.statVarSpec.date,
```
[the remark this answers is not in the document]
dwnoble: Did some refactoring and added call to getCappedStatVarDate to the getDataCsv callback

static/js/components/tiles/scatter_tile.tsx
dwnoble: Updated scatter_tile with a TODO to handle dates on a per-variable basis, and added capped stat var date handling

static/js/components/tiles/ranking_tile.tsx
dwnoble: Good catch- added

static/js/components/tiles/bivariate_tile.tsx
```diff
@@ -154,6 +145,32 @@ export function BivariateTile(props: BivariateTilePropType): JSX.Element {
...
+ * Returns callback for fetching chart CSV data
+ * @param props Chart properties
+ * @returns Async function for fetching chart CSV
+ */
+function getDataCsvCallback(
+  props: BivariateTilePropType
+): () => Promise<string> {
+  return () => {
+    // Assume all variables will have the same date
+    const date = getFirstStatVarSpecDate(props.statVarSpec);
```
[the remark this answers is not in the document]
dwnoble: Kept data fetch as is since it already handles case where different variables have different dates. Added TODO here to handle multiple dates in the CSV export

static/js/shared/util.ts
```diff
@@ -114,30 +115,61 @@ export function isDateTooFar(date: string): boolean {
...
+/**
+ * Gets the first date from a list of stat var spec objects
+ *
+ * Tiles in the subject config page currently operate with the assumption that
+ * all dates set for a subject page config will have the same date
+ *
+ * @param variables stat var spec variables
+ * @returns first date found or undefined if stat var spec list is empty
+ */
+export function getFirstCappedStatVarSpecDate(
```
chejennifer, line 164: nit: maybe move this to tile_utils.ts because StatVarSpec is a concept specific to tiles

chejennifer (approved):
thanks for all the changes!

static/js/shared/util.ts
dwnoble: done!

PR #4083, opened 2024-03-29
Save embeddings.yaml with only default embeddings for custom DCs.

keyurva:
Currently, all embeddings specified in embeddings.yaml (default, biomed, UN) are loaded in Custom DCs. We only need the default in them.

When building a custom DC docker image, we [run a script](https://github.com/datacommonsorg/website/blob/3112669701480bc5fe9c1e42051445d631cecace/build/web_compose/Dockerfile#L120) to download and save the default model in the docker image. With this change, we additionally save embeddings.yaml with just the default embeddings in it.

Before:

image

After:

image


pradh (approved):
Nice

PR #4092, opened 2024-04-05
NL: Fix a few bugs with superlative, etc

pradh:
1.  Fix the "diabetes vs. poor latinos in california" query, which currently shows hispanics without health insurance.  See screenshot below.  This is done by ordering SUPERLATIVE to not be higher than correlation.

2. Superlatives is broken with context now (asthma in california => richest counties in nevada), fix that by skipping SUPERLATIVE when SVs are found in CURRENT_QUERY (not PAST_QUERY, etc).

3.  In context detection, ensure that even if `is_multi_sv` is false, we set the 'CURRENT_QUERY' correctly in merge_context.  This is important since downstream (route_correlation_or_comparison) may still try to fulfill it as multi-sv query, and at that point CURRENT_QUERY won't be set.  This also improves the separate detection + fulfill flow to be more in line with detection-and-fulfill.

4.  Update the number of multi-SV parts to be 2 instead of 4, since that reduces the number of embeddings matches we do, etc.   This has some effect on behavior, since we detect multi-sv case based on the top-most score (a bug).

5.  Fix low-confidence commentary when correlation query has a query in the context.  This shows up in test edits.

6.  Also fixes broken topic-cache test

7.  And adds a recent demo link

image


pradh:
I'm running through all the demos now...

pradh:
> I'm running through all the demos now...

Looking good

chejennifer (approved):
nice, thanks for the fixes

pradh:
Thanks for the review!!

PR #4095, opened 2024-04-08
[nl] fix counter error

chejennifer:
values passed to the counter needs to be JSON serializable and entities was not an object type that was serializable so was causing an error when trying to get debug logs


server/lib/nl/detection/context.py
```diff
@@ -356,7 +356,7 @@ def _detect_entities(uttr: nl_uttr.Utterance) -> List[str]:
     if uttr.prev_utterance and uttr.prev_utterance.entities:
       entities = uttr.prev_utterance.entities
       uttr.entities = entities
-      uttr.counters.info('insight_entity_ctx', entities)
+      uttr.counters.info('insight_entity_ctx', [e.dcid for e in entities])
```
pradh, line 359: For simplifying future debugging, maybe we could add `traceback.format_stack(limit=10)` to https://github.com/datacommonsorg/website/blob/9e23f8cd7216470b45fa345457a296c218203f4a/server/lib/nl/common/counters.py#L70 ?

pradh (approved):
Thanks for the fix!

server/lib/nl/detection/context.py
chejennifer: tried out including stack info and it's actually not very useful because the exception gets thrown at the end when the counters are being formatted, but the bug is caused when adding an entry to the counters which isn't part of the stack info .. but I can still add stack info if you want it

pradh: Ohh good point, then never-mind.  This needs more tracking in the counter class to do it...

pradh (approved):

chejennifer:
thanks for the review!

PR #4103, opened 2024-04-11
fixed bar chart csv export error

dwnoble:
Fixed bar chart CSV export error for multi-place charts

Example:
http://localhost:8080/explore/#client=ui_query&q=How+does+the+GDP+of+USA+compare+with+that+of+China

## Before
before1
before2

## After
after


n-h-diaz (approved):

chejennifer (approved):
thanks for the fix

PR #4106, opened 2024-04-15
[cron testing] Tool for updating nodejs query diff goldens

chejennifer:
Various updates for nodejs query diffing

- updates differ process so that nodejs query differ runs diffs against local goldens
- updating goldens means pulling the results from the latest run and saving them as the new goldens
- adds instructions for debugging diffs to the email


tools/nl/nodejs_query_differ/README.md
```diff
@@ -11,4 +11,28 @@ To run this:
...
+
+### Push goldens
+
+To push the new goldens that have been saved locally to gcs (The differ runs diffs against the gcs goldens), you can run:
+
+```bash
+./update_goldens.sh -m push -e <ENV>
+```
+
+- ENV is the environment to update the goldens for. Default is 'autopush'
```
pradh, line 38: Just so I understand, is there a reason the test doesn't read the jsons from the checked-in version?

chejennifer: I chose to put it in gcs so that if we want to update the goldens, we don't have to go through an extra step of rebuilding the cron testing job docker image to use these new goldens.. but I guess we could also update cloudbuild to re-build cron testing job docker image on PR submissions. Do you have a preference

pradh: Having this checked-in and auto-building docker image would be very nice.  But I'm not sure how much more work the latter is?

tools/nl/nodejs_query_differ/update_goldens.py
```diff
@@ -0,0 +1,111 @@
...
+    del golden['debug']
+    with open(f'goldens/{FLAGS.env}/{filename}', 'w') as out:
+      out.write(json.dumps(golden))
+
+
+def update_goldens(gcs_bucket):
+  """Upload the local golden files to gcs"""
+  gcs_golden_folder = f'{FLAGS.env}/{_GOLDEN_FOLDER}/nodejs_query'
+  for blob in gcs_bucket.list_blobs(prefix=gcs_golden_folder):
+    blob.delete()
```
pradh, line 91: Do we need to delete and write?

tools/nl/nodejs_query_differ/update_goldens.py
```diff
@@ -0,0 +1,111 @@
...
+        "Could not get new_golden_folder name, please enter one manually.")
+    return
+  new_golden_folder = f'{FLAGS.env}/{new_golden_folder}/nodejs_query/'
+  new_golden_blobs = gcs_bucket.list_blobs(prefix=new_golden_folder)
+  for b in new_golden_blobs:
+    filename = b.name.split('/')[-1]
+    golden = json.loads(b.download_as_string())
+    del golden['debug']
+    with open(f'goldens/{FLAGS.env}/{filename}', 'w') as out:
+      out.write(json.dumps(golden))
```
pradh, line 84: Can we do an `indent=2` to make it readable?

tools/nl/nodejs_query_differ/README.md
chejennifer: auto-building docker image should be simple, but will need to test out packaging the golden files with the docker image and then using that to run the differ. Let me try it out!

tools/nl/nodejs_query_differ/update_goldens.py
chejennifer: done

chejennifer:
Updated process to:

- read local golden files for diff
- when updating golden files, update local goldens & on PR submission, this will also update the golden files used in the cron job

pradh (approved):
Thanks for the edits!

chejennifer:
thanks for the reviews

PR #4108, opened 2024-04-16
Update NL yaml config in prep for LanceDB Integration

pradh:
`embeddings.yaml` format is changed to identify explicitly the model (without inferring from filename), use absolute embeddings path and the type of store to load it into.


deploy/nl/embeddings.yaml
```diff
@@ -1,4 +1,20 @@
-medium_ft: embeddings_medium_2024_03_14_16_38_53.ft_final_v20230717230459.all-MiniLM-L6-v2.csv
-sdg_ft: embeddings_sdg_2023_12_26_10_03_03.ft_final_v20230717230459.all-MiniLM-L6-v2.csv
-undata_ft: embeddings_undata_2024_03_20_11_01_12.ft_final_v20230717230459.all-MiniLM-L6-v2.csv
-bio_ft: embeddings_bio_2024_03_19_16_39_03.ft_final_v20230717230459.all-MiniLM-L6-v2.csv
+medium_ft:
+  store: MEMORY
+  embeddings: embeddings_medium_2024_03_14_16_38_53.ft_final_v20230717230459.all-MiniLM-L6-v2.csv
```
shifucun, line 3: can we use full gcs path for embeddings?

nl_server/embeddings_map.py
```diff
@@ -61,9 +63,14 @@ def __init__(self, indexes: List[EmbeddingsIndex]):
...
-      self.embeddings_map[idx.name] = Embeddings(
-          model=self.name2model[idx.tuned_model],
-          store=MemoryEmbeddingsStore(idx.embeddings_local_path))
+      if idx.store_type == StoreType.MEMORY:
+        self.embeddings_map[idx.name] = Embeddings(
+            model=self.name2model[idx.tuned_model],
+            store=MemoryEmbeddingsStore(idx.embeddings_local_path))
+      elif idx.store_type == StoreType.LANCEDB:
+        self.embeddings_map[idx.name] = Embeddings(
+            model=self.name2model[idx.tuned_model],
```
shifucun, line 72: can we get rid of "tuned"..?

nl_server/loader.py
```diff
@@ -27,7 +27,6 @@
 from shared.lib.gcs import is_gcs_path
 from shared.lib.gcs import join_gcs_path
 
-_MODEL_YAML = 'models.yaml'
 _EMBEDDINGS_YAML = 'embeddings.yaml'
```
shifucun, line 30: this is much better!

nl_server/embeddings_map.py
pradh: Done.

pradh:
Since there was too much going on in the PR, I removed the LanceDB integration out of it, and this is just a config change PR.

nl_server/config.py
```diff
@@ -34,97 +35,84 @@
...
-  tuned_model_local_path: str = ""
+  model_local_path: str = ""
 
 
 #
 # Validates the config input, downloads all the files and returns a list of Indexes to load.
 #
-def load(embeddings_map: Dict[str, str],
-         models_map: Dict[str, str]) -> List[EmbeddingsIndex]:
+def load(embeddings_map: Dict[str, Dict[str, str]]) -> List[EmbeddingsIndex]:
```
chejennifer, line 66: nit: maybe comment on what the keys/values for the Dicts are or have a dataclass?

nl_server/gcs.py
```diff
@@ -37,25 +35,25 @@ def local_path(embeddings_file: str) -> str:
...
-  """Downloads a Sentence Tranformer model (or finetuned version) from GCS.
+def download_folder_from_gcs(gcs_bucket: Any, local_dir: str,
+                             folder_name: str) -> str:
+  """Downloads a folder from GCS.
 
   Args:
     gcs_bucket: The GCS bucket.
     local_dir: the local folder to download to.
-    model_folder_name: the GCS bucket name for the model.
+    model_folder_name: the GCS folder name for the model.
```
chejennifer, line 45: nit: folder_name

chejennifer (approved):
nice!

nl_server/config.py
pradh: The inputs are raw yaml representation being turned into dataclass in this function, so just added comments.

nl_server/gcs.py
pradh: Done

pradh (commented):
Thanks!

nl_server/config.py
```diff
@@ -34,97 +35,90 @@
...
+  # Values are: MEMORY, LANCEDB
+  store_type: StoreType
+
   # File name provided in the yaml file.
-  embeddings_file_name: str
+  embeddings_path: str
   # Local path.
   embeddings_local_path: str
 
   # Fine-tuned model name ("" if embeddings uses base model).
```
shifucun, line 57: update comment here

nl_server/gcs.py
```diff
@@ -67,24 +65,31 @@ def download_model_from_gcs(gcs_bucket: Any, local_dir: str,
...
-  bucket = sc.bucket(bucket_name=BUCKET)
+
+  if gcs_lib.is_gcs_path(path):
+    bucket_name, base_name = gcs_lib.get_gcs_parts(path)
+    bucket = sc.bucket(bucket_name=bucket_name)
+  else:
+    bucket = sc.bucket(bucket_name=BUCKET)
+    base_name = path
   directory = gcs_lib.TEMP_DIR
+  local_dir = os.path.basename(base_name) if '/' in base_name else base_name
```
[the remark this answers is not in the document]
pradh: Good catch.

keyurva (approved):
Thanks - nicely setup for lancedb!

nl_server/embeddings_map.py
```diff
@@ -51,19 +52,20 @@ def __init__(self, indexes: List[EmbeddingsIndex]):
...
       self.name2model[model_name] = LocalSentenceTransformerModel(model_path)
 
     # NOTE: Not excluding CUSTOM_DC_INDEX from the map, so should the
     # custom DC customers want queries to work within their variables
     # they can set `idx=custom`.
     for idx in indexes:
-      self.embeddings_map[idx.name] = Embeddings(
-          model=self.name2model[idx.tuned_model],
-          store=MemoryEmbeddingsStore(idx.embeddings_local_path))
+      if idx.store_type == StoreType.MEMORY:
```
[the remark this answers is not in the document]
pradh: Actually, on second thought, I'm not sure there will be many more.  This code is kinda like the "factory" for creating the right entries, so maybe this is ok?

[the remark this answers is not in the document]
keyurva: A factory type file / class hierarchy would be useful in 2 cases - if there is a possibility of more store types or there are conditionals based on the store type in multiple places. If neither is true, then we're ok. If either is true, the separation can help.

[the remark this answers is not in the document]
pradh: Oh, to be clear, we already have a class hierarchy for EmbeddingsStore and EmbeddingsModel.  This is just the logic that constructs them, and while there can be more store types, I don't anticipate conditionals based on the store type in multiple places.

So I really don't see a need for another hierarchy, perhaps moving it to a separate helper at best.

PR #4109, opened 2024-04-16
Add CrossEncoder model to Vertex AI Model Server

pradh:
Preliminary PR for reranking support.  I will build docker image and create endpoint after merging.


shifucun (approved):
nice!

PR #4110, opened 2024-04-16
LanceDB Integration

pradh:
This PR adds support for LanceDB, a new on-disk embeddings DB, so we can support multiple, much larger embeddings without having them all loaded into memory.  LanceDB also supports quicker load times.

Changes:

* Implementation of the LanceDBStore.  A new index `lance_medium_ft` using that store is added for testing.

* Support to create LanceDB index in `build_embeddings` tool.

* A simple integration test for lancedb index that tests parity with memory index.

Size impact:  Adds about ~90MB of packages and one more dependency on torch


static/js/apps/explore/app.tsx
```diff
@@ -269,6 +269,7 @@ export function App(props: { isDemo: boolean }): JSX.Element {
     const topic = getSingleParam(hashParams[URL_HASH_PARAMS.TOPIC]);
     const place = getSingleParam(hashParams[URL_HASH_PARAMS.PLACE]);
     const dc = getSingleParam(hashParams[URL_HASH_PARAMS.DC]);
+    const idx = getSingleParam(hashParams[URL_HASH_PARAMS.IDX]);
```
shifucun, line 272: what is this idx?

shifucun (approved):
very nice!

static/js/apps/explore/app.tsx
pradh: This is the embeddings index listed in embeddings.yaml.  I added a comment.

pradh: Mainly for testing.

pradh:
Thanks for the review!

PR #4111, opened 2024-04-16
[NL] Update timeline date range to work with FE change

chejennifer:
There was a FE change for how timeline tiles handle date ranges. Timelines used to cut dates to the lowest granularity between specified date range and observation date before comparing them, but now timeline will only cut observation dates to the granularity of the specified date range before comparing dates. On the NL side, we want the original behavior so that if a start date like 2020-04 is specified, observations from 2020 still show. This is a change to the date range we pass in the tile config from the NL so that we get the original behavior.

Additional change:
- update the regex for date recognition because query will actually be in lower case by the time we get to the date parsing code
- make sv_place_facet and sv_place_facets consistent to all be singular


server/tests/lib/nl/date_parser_test.py
```diff
@@ -25,8 +25,8 @@
 class TestDateParser(unittest.TestCase):
 
   @parameterized.expand([
-      ('USA population in May, 2021', [Date('in', 2021, 5)]),
-      ('USA population in Sep 2021', [Date('in', 2021, 9)]),
+      ('USA population in may, 2021', [Date('in', 2021, 5)]),
```
pradh, line 28: To confirm, this is fine because somewhere upstream the small casing happens?

server/lib/nl/config_builder/timeline.py
```diff
@@ -31,11 +31,25 @@
 
 # set the line tile spec field for a date range or a single date.
 def _set_line_tile_spec(date_range: types.Date, single_date: types.Date,
-                        line_tile_spec: LineTileSpec):
+                        line_tile_spec: LineTileSpec,
+                        sv_place_facet: types.Sv2Place2Facet):
   if date_range:
     start_date, end_date = get_date_range_strings(date_range)
```
pradh, line 37: nit: Can we put date-range handling in a helper?

pradh (approved):
Thanks!

server/tests/lib/nl/date_parser_test.py
chejennifer: yes, I believe so, but thinking more, I think it's better to just have both cases

server/lib/nl/config_builder/timeline.py
chejennifer: done

PR #4112, opened 2024-04-16
[biomed nl] update place detection to use new endpoint for entities

chejennifer:
Use new recognize entities endpoint for detecting entities. Also, always strip entities from the query if no places found and entities are allowed.

Note: the common word entities that are returned in the goldens are fixed in autopush mixer but the goldens are generated from staging mixer


server/lib/nl/detection/place.py
```diff
@@ -48,93 +48,56 @@ def detect_from_query_dc(orig_query: str,
...
-  resolved_entities = []
-  parent_map = {}
-  if mains:
-    resolved_places, parent_map = get_place_from_dcids(
-        mains, debug_logs["place_resolution"])
-    if allow_triples:
-      resolved_entities = _get_non_place_entities(mains, resolved_places)
-  resolved_place_dcids = set([p.dcid for p in resolved_places])
-  resolved_entity_dcids = set([e.dcid for e in resolved_entities])
+  place_query_parts = _get_query_parts(query, debug_logs["dc_recognize_places"],
```
pradh, line 55: nit: having the API call in `_get_query_parts` feels a bit odd based on a bool is a bit odd, could you instead inline it here, maybe?

server/lib/nl/detection/place.py
```diff
@@ -48,93 +48,56 @@ def detect_from_query_dc(orig_query: str,
...
+                                       True)
+  # Use the first DCID for now.
+  place_main_dcids = [p.dcids[0] for p in place_query_parts if p.dcids]
+  resolved_places, parent_map = get_place_from_dcids(
```
pradh, line 59: For consistency, update other callsites of `get_place_from_dcids` to not check empty before calling?

server/lib/nl/detection/place.py
```diff
@@ -279,6 +244,66 @@ def _get_non_place_entities(all_entities: List[str],
...
+# recognized.
+#
+def _get_resolved_entities(query_parts: List[QueryPart]) -> List[Entity]:
+  entity_dcids = set()
+  for e in query_parts:
+    if not e.dcids:
+      continue
+    entity_dcids.update(e.dcids)
+
+  return _get_non_place_entities(list(entity_dcids), [])
```
pradh, line 304: Why is 2nd arg empty? Comment...

server/lib/nl/detection/place.py
```diff
@@ -48,93 +48,56 @@ def detect_from_query_dc(orig_query: str,
...
+  if allow_triples:
+    debug_logs["dc_recognize_entities"] = {}
+    entity_query_parts = _get_query_parts(stripped_query_str,
+                                          debug_logs["dc_recognize_entities"],
+                                          False)
+    resolved_entities = _get_resolved_entities(entity_query_parts)
+    resolved_entity_dcids = set([e.dcid for e in resolved_entities])
+    entities_str, stripped_entities_query_str = _get_stripped_string(
+        entity_query_parts, resolved_entity_dcids)
+    if not places_str or len(entities_str) > _MAX_ENTITIES_QUERY:
```
pradh, line 83: Why do we still need _MAX_ENTITIES_QUERY?

server/lib/nl/detection/place.py
```diff
@@ -48,93 +48,56 @@ def detect_from_query_dc(orig_query: str,
...
-      stripped_query_parts.append(p.substr)
-  query_without_place_substr = ' '.join(stripped_query_parts)
+  if allow_triples:
+    debug_logs["dc_recognize_entities"] = {}
+    entity_query_parts = _get_query_parts(stripped_query_str,
+                                          debug_logs["dc_recognize_entities"],
+                                          False)
+    resolved_entities = _get_resolved_entities(entity_query_parts)
+    resolved_entity_dcids = set([e.dcid for e in resolved_entities])
+    entities_str, stripped_entities_query_str = _get_stripped_string(
```
pradh, line 81: Add vertical spacing, and call out why we pass `stripped_query_str`?

pradh (commented):
Nice!

server/lib/nl/detection/place.py
chejennifer: updated to do the API call in this function and then pass the response to _get_query_parts

server/lib/nl/detection/place.py
chejennifer: done

server/lib/nl/detection/place.py
chejennifer: made the second arg optional in the helper function and updated the helper function definition

server/lib/nl/detection/place.py
chejennifer: I kept it there in case there's a case where a place is detected?

server/lib/nl/detection/place.py
chejennifer: done

server/lib/nl/detection/place.py
```diff
@@ -295,13 +313,16 @@ def _get_stripped_string(query_parts: List[QueryPart],
...
   for e in query_parts:
     if not e.dcids:
       continue
-    entity_dcids.update(e.dcids)
-
-  return _get_non_place_entities(list(entity_dcids), [])
+    new_dcids = [d for d in e.dcids if not d in seen_dcids]
+    seen_dcids.update(new_dcids)
+    # sort the dcids so that the list is deterministic
+    entity_dcids.extend(sorted(new_dcids))
```
pradh, line 324: Should we keep the order in which user specified the list?  I "think" place entities maintains order

pradh: So for example, if user asks for X, Y and Z, we know which one is first.

server/lib/nl/detection/place.py
pradh: ok

pradh (approved):
Thanks for the edits!

server/lib/nl/detection/place.py
chejennifer: yes, this still maintains order because it is sorting the dcids within each query part. In biomed, there will be lots of entities with the same name, so each query part can return multiple dcids & there's preference for showing the answer for all the dcids

pradh: Hmm, then should we sort better maybe (a more common type ahead of less common one?).  Have a TODO if you think so too.

chejennifer: I do agree with sorting better, but types probably won't do much because a lot of times, the entities have the same type (e.g., FGFR1 is the name for hg19_FGFR1 and hg38_FGFR1 which are both genes)>

But added a TODO to sort better

chejennifer:
thanks for the reviews!

PR #4121, opened 2024-04-18
NL: Implement reranking support

pradh:
After this change if `reranker=cross-encoder-mxbai-rerank-base-v1` is added to the URL param on autopush, we'll perform reranking.

The reranking results will show up in debug-info UI in two ways.

**1. Counters of SV lists `pre_reranking` and `post_reranking`:**
image

**2. Reranking scores in Sentence table:**
image


.vscode/launch.json
```diff
@@ -23,7 +23,8 @@
         "ENV_PREFIX": "DC",
         "FLASK_DEBUG": "1",
         "ENABLE_MODEL": "true",
-        "GOOGLE_CLOUD_PROJECT": "datcom-website-dev"
+        "GOOGLE_CLOUD_PROJECT": "datcom-website-dev",
+        "ENABLE_EVAL_TOOL": "true",
```
chejennifer, line 27: might want to rename this flag to something that's not eval tool specific since it looks like this is the flag used to enable vertex ai models?

server/lib/nl/detection/rerank.py
```diff
@@ -22,7 +22,59 @@
...
-           candidates: vars.VarCandidates,
+           var_candidates: vars.VarCandidates,
            debug_logs: Dict) -> vars.VarCandidates:
-  # TODO: Implement me
-  return candidates
+  # 1. Prepare indexes and inputs
+
+  # List of query-sentence pairs.
+  qs_pairs: List[List[str, str]] = []
+  # Sentence to index into var_candidates parallel arrays.
```
chejennifer, line 31: nit: mention that this is the index of the sentence's matching sv's index in var_candidates.svs

chejennifer (approved):
this is very cool!

.vscode/launch.json
shifucun: maybe have another config like "Flask with NL and VertexAI"?

static/js/types/app/nl_interface_types.ts
```diff
@@ -60,8 +60,9 @@ export interface SVScores {
 }
 
 export interface SentenceScore {
-  Sentence: string;
-  Score: string;
+  sentence: string;
+  score: number;
+  rerank_score: number;
```
shifucun, line 65: nit: rerankScore (camel case in js)

server/integration_tests/test_data/detection_api_reranking/debug_info.json
```diff
@@ -0,0 +1,80 @@
...
+              "SV": [
+                "dc/topic/Income"
+              ]
+            }
+          ]
+        }
+      ]
+    },
+    "Query": "population that is rich in",
+    "SV": [
```
shifucun, line 58: i don't see post_rerank field in this golden json?

shifucun (commented):
What is a general range for rerank score and how to interpret them?

.vscode/launch.json
pradh: @shifucun ok if I change it to ENABLE_VERTEX_AI ?

pradh:
> What is a general range for rerank score and how to interpret them?

I see that each model has its own range, but generally, the higher the relative value the closer the match.  Maybe if we find a model works best, then we can look at its range for further heuristics?

.vscode/launch.json
shifucun: sg

server/integration_tests/test_data/detection_api_reranking/debug_info.json
pradh: Good point, I added it now

static/js/types/app/nl_interface_types.ts
pradh: Done.

server/lib/nl/detection/rerank.py
pradh: Done

.vscode/launch.json
pradh: On second thought and post discussion with Bo, I reverted this change.  I plan to move this to NL server, so let Web Server's use of Vertex AI continue to be gated under EVAL tool.

pradh (commented):
Thanks for the review!

shifucun (approved):

pradh:
Thanks for the reviews!

PR #4123, opened 2024-04-18
[biomed nl] handle duplicate answers and no props

chejennifer:
1. fix duplicate values in answer table tile
    - autopush: https://screenshot.googleplex.com/3EktF5rAWdFTpHC
    - local: https://screenshot.googleplex.com/4zFdCfpXEsUm2n5
3. return entity overview when there is entity detected, but no props
   - autopush: https://screenshot.googleplex.com/9dM6RvpBWkuynid
   - local: https://screenshot.googleplex.com/4k5aibpTAK6cgCB

Note: the integration test chart_configs are wrong because it uses staging mixer and staging mixer does not handle entity names that are also common words (autopush mixer does handle this)


static/js/components/tiles/answer_table_tile.tsx
```diff
@@ -149,14 +149,15 @@ const fetchData = async (
...
         val.forEach((singleVal) => {
-          entityResults.push(
+          entityResults.add(
             singleVal.name || singleVal.value || singleVal.dcid
           );
           provIds.add(singleVal.provenanceId);
         });
-        values[entity][props.columns[i].propertyExpr] = entityResults;
+        values[entity][props.columns[i].propertyExpr] =
+          Array.from(entityResults);
```
pradh, line 160: Does order matter?

pradh (approved):
Should we perhaps push to staging so we get these tests working right?

static/js/components/tiles/answer_table_tile.tsx
chejennifer: no, I don't think the order matters when displaying the values

server/integration_tests/test_data/e2e_triple/whatstrandorientationdoesfgfr1have/chart_config.json
```diff
@@ -27,7 +26,7 @@
                       "bio/mm10_Fgfr1",
                       "bio/mm9_Fgfr1"
                     ],
-                    "title": "The strandOrientation for Orientation, FGFR1 and 6 more other are as follows:",
+                    "title": "The strandOrientation for FGFR1, FGFR1 and 5 more other are as follows:",
```
juliawu, line 29: I think this should say other**s** (plural)

chejennifer: doing title/messaging changes in a separate pr!

chejennifer:
thanks for the reviews!

PR #4124, opened 2024-04-18
Adapt building custom dc embeddings to the recent yaml changes.

keyurva:
* This should fix the recent docker build failures: https://screenshot.googleplex.com/SEVhMwYnxZeyTRR
* There is a bug in this PR where the default model and embeddings paths don't match what we do in main dc.  I'll address this in a follow-up PR.

### DIFF
--- a/tools/nl/embeddings/utils.py
+++ b/tools/nl/embeddings/utils.py
@@ -43,7 +43,7 @@ _DEFAULT_EMBEDDINGS_INDEX_TYPE = "medium_ft"
 
 _CHUNK_SIZE = 100
 
-_MODEL_ENDPOINT_RETRYS = 3
+_MODEL_ENDPOINT_RETRIES = 3
 
 _GCS_PATH_PREFIX = "gs://"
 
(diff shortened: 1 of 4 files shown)


tools/nl/embeddings/utils.py
```diff
@@ -40,21 +40,20 @@
...
 _DEFAULT_EMBEDDINGS_INDEX_TYPE = "medium_ft"
-"""The embeddings filename pattern in embeddings.yaml.
 
-Expect: <embeddings_version>.<fine-tuned-model-version>.<base-model>.csv
-Example: embeddings_sdg_2023_09_12_16_38_04.ft_final_v20230717230459.all-MiniLM-L6-v2.csv
+_CHUNK_SIZE = 100
 
-Model version: <fine-tuned-model-version>.<base-model>
-Example: ft_final_v20230717230459.all-MiniLM-L6-v2
+_MODEL_ENDPOINT_RETRYS = 3
```
pradh, line 46: nit: RETRIES

pradh (approved):

tools/nl/embeddings/utils.py
keyurva: Done.

PR #4125, opened 2024-04-18
Fixed issue where highest coverage date would override date provided by stat var spec

dwnoble:
Example query: "Female population in California counties before 2020"

Before:
https://screenshot.googleplex.com/9rSK5p6z6sDia5k

After: 
https://screenshot.googleplex.com/3tjcbu8Rayzz5HF


chejennifer (approved):
thanks for the fix

PR #4126, opened 2024-04-19
Skip LanceDB for docker

pradh:


build/web_compose/Dockerfile
```diff
@@ -66,6 +66,8 @@ RUN pip3 install -r /workspace/server/requirements.txt
 COPY nl_app.py /workspace/nl_app.py
 COPY deploy/nl/. /datacommons/nl/
 COPY nl_server/requirements.txt /workspace/nl_server/requirements.txt
+# TODO: Undo this once we figure out why lancedb breaks Custom DC Docker
+RUN sed -i '' '/lancedb/d' /workspace/nl_server/requirements.txt
```
keyurva, line 70: Can we skip removing this since it works with requirements.txt as-is?

build/web_compose/compose.sh
```diff
@@ -27,6 +27,7 @@ if [[ $GCS_DATA_PATH != "" ]]; then
 else
     export USER_DATA_PATH=/userdata/
 fi
+export IS_CUSTOM_DC=true
```
keyurva, line 30: nit: Add a comment?

nl_server/embeddings_map.py
```diff
@@ -66,9 +65,13 @@ def __init__(self, indexes: List[EmbeddingsIndex]):
             model=self.name2model[idx.model_name],
             store=MemoryEmbeddingsStore(idx.embeddings_local_path))
       elif idx.store_type == StoreType.LANCEDB:
-        self.embeddings_map[idx.name] = Embeddings(
-            model=self.name2model[idx.model_name],
-            store=LanceDBStore(idx.embeddings_local_path))
+        if not _is_custom_dc():
+          from nl_server.store.lancedb import LanceDBStore
```
keyurva, line 69: nit: Add comment why we are doing this?

nl_server/embeddings_map.py
```diff
@@ -157,3 +160,7 @@ def _merge_custom_index(default: EmbeddingsIndex,
       f'custom {custom.embeddings_local_path} into {output_embeddings_path}')
 
   return output_embeddings_path
+
+
+def _is_custom_dc() -> bool:
```
keyurva, line 165: Can we add this method to the shared lib so it can be used in other places?

keyurva (approved):
Thanks for the fix.

nl_server/embeddings_map.py
pradh: There's not a good file, and need to create one for this, lets move it when its used another time?

build/web_compose/Dockerfile
pradh: I don't understand why that is?

build/web_compose/compose.sh
pradh: Self explanatory?

nl_server/embeddings_map.py
pradh: Done.

build/web_compose/Dockerfile
pradh: Also, did you delete any lancedb.py file and refs, vs. keep them but not have the code run?

pradh:
Still testing this ...

build/web_compose/Dockerfile
keyurva: I ran with lancedb in requirements.txt but without refs and without lancedb.py

build/web_compose/compose.sh
keyurva: ok

nl_server/embeddings_map.py
keyurva: SG

PR #4127, opened 2024-04-19
Fix downloaded model and embeddings paths for custom DC.

keyurva:
* This fixes the bug mentioned in #4124 with the downloaded model and embeddings paths.
* It also makes local paths of downloaded GCS files consistent with downloaded GCS folders - in that the former is also prefixed with the bucket name now like the latter.
* Finally, the embeddings are downloaded using an anonymous client like they were [previously](https://github.com/datacommonsorg/website/blob/master/nl_server/gcs.py#L31).

### DIFF
--- a/tools/nl/embeddings/build_custom_dc_embeddings.py
+++ b/tools/nl/embeddings/build_custom_dc_embeddings.py
@@ -53,7 +53,6 @@ flags.DEFINE_string(
     "Path where the default FT embeddings.yaml will be saved for Custom DC in download mode."
 )
 
-MODELS_BUCKET = 'datcom-nl-models'
 EMBEDDINGS_CSV_FILENAME_PREFIX = "custom_embeddings"
 EMBEDDINGS_YAML_FILE_NAME = "custom_embeddings.yaml"
 
@@ -143,7 +142,8 @@ def _download_model(model_version: str) -> utils.Context:
 
 
 def _ctx_no_model() -> utils.Context:
-  bucket = storage.Client.create_anonymous_client().bucket(MODELS_BUCKET)
+  bucket = storage.Client.create_anonymous_client().bucket(
+      utils.DEFAULT_MODELS_BUCKET)
   return utils.Context(model=None, model_endpoint=None, bucket=bucket)
 
 
--- a/tools/nl/embeddings/build_custom_dc_embeddings.py
+++ b/tools/nl/embeddings/build_custom_dc_embeddings.py
@@ -14,6 +14,7 @@
 """Build embeddings for custom DCs."""
 
 import os
+import sys
 
 from absl import app
 from absl import flags
@@ -23,6 +24,15 @@ from google.cloud import storage
 import pandas as pd
 import utils
 import yaml
+
+# Import gcs module from shared lib.
+# Since this tool is run standalone from this directory,
+# the shared lib directory needs to be appended to the sys path.
+_THIS_DIR = os.path.dirname(os.path.abspath(__file__))
+_SHARED_LIB_DIR = os.path.join(_THIS_DIR, "..", "..", "..", "shared", "lib")
+print("SHARED LIB DIR", _SHARED_LIB_DIR)
+sys.path.append(_SHARED_LIB_DIR)
+import gcs  # type: ignore
 
 FLAGS = flags.FLAGS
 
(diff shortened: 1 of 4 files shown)


tools/nl/embeddings/utils.py
```diff
@@ -267,11 +277,13 @@ def get_or_download_file_from_gcs(ctx: Context, file_name: str) -> str:
   If the file is already downloaded, it returns the file path.
   Otherwise, it downloads the file from GCS to the local file system and returns that path.
   """
+  bucket_name = DEFAULT_MODELS_BUCKET
```
pradh, line 280: Can this function use the one that nl server also uses?

pradh (commented):
Thanks!

tools/nl/embeddings/utils.py
keyurva: Updated to using the shared lib now. However it needs to be imported awkwardly. PTAL.

tools/nl/embeddings/build_custom_dc_embeddings.py
```diff
@@ -71,7 +80,8 @@ def download(embeddings_yaml_path: str):
 
   # Download embeddings.
   embeddings_file_name = default_ft_embeddings_info["embeddings"]
-  utils.get_or_download_file_from_gcs(ctx, embeddings_file_name)
+  gcs.download_gcs_file(embeddings_file_name)
+  # utils.get_or_download_file_from_gcs(ctx, embeddings_file_name)
```
pradh, line 84: delete

tools/nl/embeddings/utils.py
```diff
@@ -163,22 +166,6 @@ def dedup_texts(df: pd.DataFrame) -> Tuple[Dict[str, str], List[List[str]]]:
...
-  Returns the path to the local directory where the file was downloaded to.
-  ```
-  """
-  local_file_path = os.path.join(ctx.tmp, file_name)
-  blob = ctx.bucket.get_blob(file_name)
-  blob.download_to_filename(local_file_path)
-  return local_file_path
-
-
 def _download_model_from_gcs(ctx: Context, model_folder_name: str) -> str:
```
pradh, line 169: TODO:  Can this use download_folder?  (We can move it to shared)

tools/nl/embeddings/build_custom_dc_embeddings.py
```diff
@@ -24,6 +25,15 @@
 import utils
 import yaml
 
+# Import gcs module from shared lib.
+# Since this tool is run standalone from this directory,
+# the shared lib directory needs to be appended to the sys path.
+_THIS_DIR = os.path.dirname(os.path.abspath(__file__))
+_SHARED_LIB_DIR = os.path.join(_THIS_DIR, "..", "..", "..", "shared", "lib")
```
pradh, line 32: nit: why not "../../../shared/lib"

tools/nl/embeddings/build_custom_dc_embeddings.py
```diff
@@ -24,6 +25,15 @@
 import utils
 import yaml
 
+# Import gcs module from shared lib.
```
pradh, line 28: is this needed in docker?

pradh (commented):
Thanks!

tools/nl/embeddings/build_custom_dc_embeddings.py
keyurva: Yes.

tools/nl/embeddings/build_custom_dc_embeddings.py
keyurva: Just a good practice to let the os module insert the system path separator ("/" or "\") instead of hardcoding it.

tools/nl/embeddings/build_custom_dc_embeddings.py
keyurva: Done.

tools/nl/embeddings/utils.py
keyurva: Yes. Elaborated the TODO below.

tools/nl/embeddings/build_custom_dc_embeddings.py
pradh: I see.  Readability is poor, but sure :)

pradh (approved):

keyurva:
Thanks for the review!

PR #4129, opened 2024-04-19
Updated highest coverage check to only filter out future dates for specific variables

dwnoble:
* fixes bug where projected temperature readings were mistakenly being filtered  out

## before
uLcoMCqG4AUCiGi

## after
6WeUmhKL3darhKr (1)


server/lib/util.py
```diff
@@ -253,6 +253,11 @@
     'chart_titles_by_sv.json', 'chart_titles_by_sv_sdg.json'
 ]
 
+# Filter out observations with dates in the future for these variable DCIDs
+# when finding the date of highest coverage
+FILTER_FUTURE_OBSERVATIONS_FROM_HIGHEST_COVERAGE_FOR_VARIABLE_DCIDS = set(
```
pradh, line 258: frozenset

pradh, line 258: can we do a slightly shorter name?

server/lib/util.py
```diff
@@ -593,15 +598,21 @@ def _get_highest_coverage_date(observation_entity_counts_by_date,
     max_years_to_check: Only consider entity counts going back this number of
       years
   """
-  # Get observation dates in descending order, and filter out dates in the
-  # future (to exclude erroneous data)
+  # Get observation dates in descending order
   todays_date = str(date.today())
```
pradh, line 602: move todays_date to under the if condition?

pradh (approved):
Thanks Dan!

server/lib/util.py
dwnoble: done

dwnoble: changed to `FILTER_FUTURE_OBSERVATIONS_FROM_VARIABLES`

server/lib/util.py
dwnoble: done

PR #4131, opened 2024-04-19
[biomed nl] some naming updates

chejennifer:
- add chart title for entity overview charts

autopush: https://screenshot.googleplex.com/8URapF9YZabLWb5
local: https://screenshot.googleplex.com/8KC75DLA3PbmVUw

- update "and <n> more other" -> "and <n> more others"
- updates property names in table headers to be an actual name
- updates the entity string to only show unique entity names

autopush: https://screenshot.googleplex.com/3s9BQHG3xavYQXm
local: https://screenshot.googleplex.com/3FPMc4KEdb8UagP

naming/messaging that still need to be improved & will come in a separate PR:
- the template for the message (i.e., right now it's `the <property> for <entity> is:`)
- how the table heading should differentiate between in and out arcs

### DIFF
--- a/server/lib/nl/fulfillment/triple.py
+++ b/server/lib/nl/fulfillment/triple.py
@@ -53,7 +53,7 @@ def _get_entity_string(entities: List[Entity]) -> str:
     entity_name_lowercase = entity_name.lower()
     if not entity_name_lowercase in seen_entity_names:
       unique_entity_names.append(entity_name)
-    seen_entity_names.add(entity_name_lowercase)
+      seen_entity_names.add(entity_name_lowercase)
 
   # if there is only 1 unique entity name, return it
   if len(unique_entity_names) == 1:
--- a/server/lib/nl/fulfillment/triple.py
+++ b/server/lib/nl/fulfillment/triple.py
@@ -18,7 +18,7 @@ from flask import current_app
 
 from server.lib.fetch import properties
 from server.lib.nl.common.utterance import ChartType
-import server.lib.nl.common.utterance as nl_uttr
+from server.lib.nl.config_builder.base import get_property_display_name
 from server.lib.nl.fulfillment.types import ChartVars
 from server.lib.nl.fulfillment.types import Entity
 from server.lib.nl.fulfillment.types import PopulateState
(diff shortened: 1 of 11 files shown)


server/lib/nl/fulfillment/triple.py
```diff
@@ -109,7 +126,9 @@ def _add_multi_entity_overview(state: PopulateState) -> bool:
   for prop_list in properties(entity_dcids).values():
     prop_set.update(prop_list)
   prop_list = [f'{_OUT_ARROW}{prop}' for prop in sorted(list(prop_set))]
-  cv = ChartVars(props=prop_list)
+  chart_title = _OVERVIEW_TITLE.format(
+      entity=_get_entity_string(state.uttr.entities))
```
pradh, line 130: Should this happen in config builder?  (typically fulfillers don't concern with formatting names)

server/lib/nl/fulfillment/triple.py
```diff
@@ -44,20 +45,34 @@ def _entity_name(entity: Entity) -> str:
...
-    return _entity_name(entities[0])
+  # Get the list of unique entity names
+  seen_entity_names = set()
+  unique_entity_names = []
+  for e in entities:
+    entity_name = _entity_name(e)
+    entity_name_lowercase = entity_name.lower()
+    if not entity_name_lowercase in seen_entity_names:
+      unique_entity_names.append(entity_name)
+    seen_entity_names.add(entity_name_lowercase)
```
pradh, line 56: super nit: add to set when non-existent

pradh (approved):
Thanks!

server/lib/nl/fulfillment/triple.py
chejennifer: There's a TODO to move getting title to config builder. There will need to be at least one more PR for title change when we have the actual title templates, so was going to try to make the move when we do that!

server/lib/nl/fulfillment/triple.py
chejennifer: sorry, what do you mean add to set when non-existent?

pradh: Do we need to insert seen_entity_names only under the if condition?

chejennifer: ohh, yes, updated

chejennifer:
thanks for the review!

PR #4135, opened 2024-04-22
Disabled HIGHEST_COVERAGE option in SQL custom data commons

dwnoble:
Setting date=HIGHEST_COVERAGE falls back to date=LATEST in SQL custom data commons

### DIFF
--- a/server/routes/shared_api/observation/point.py
+++ b/server/routes/shared_api/observation/point.py
@@ -17,6 +17,7 @@ from flask import request
 
 from server.lib import fetch
 from server.lib.cache import cache
+from server.lib.config import is_highest_coverage_enabled
 from server.lib.util import fetch_highest_coverage
 from server.routes import TIMEOUT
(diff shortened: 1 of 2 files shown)


server/routes/shared_api/observation/point.py
```diff
@@ -73,6 +74,9 @@ def point_within():
   if not variables:
     return 'error: must provide a `variables` field', 400
   date = request.args.get('date') or DATE_LATEST
+  # If querying by highest coverage is disabled, default to DATE_LATEST
+  if date == DATE_HIGHEST_COVERAGE and not is_highest_coverage_enabled():
```
keyurva, line 78: Consider adding a TODO to remove this once b/336358566 is fixed.

keyurva (approved):
Thanks for the fix.

server/routes/shared_api/observation/point.py
dwnoble: good call. added

PR #4137, opened 2024-04-23
Use anonymous GCS client only for custom dc and local user data paths.

keyurva:


nl_server/util.py
```diff
@@ -14,6 +14,29 @@
 
 import os
 
+from shared.lib.gcs import is_gcs_path
+
+_IS_CUSTOM_DC_ENV_VAR = 'IS_CUSTOM_DC'
```
pradh, line 19: Why use constant if not referenced >1 time?  Adds more hops for the reader of code :)

pradh (approved):
Thanks

nl_server/util.py
keyurva: I prefer to declare instead of inlining constants but have removed it here.

keyurva:
Thanks for the review.

PR #4139, opened 2024-04-23
[biomed nl] add debug info and context messages

chejennifer:
- add entities and props to the debug info pane
    - entities: https://screenshot.googleplex.com/c7xQrYaW5sX3LPo
    - props: https://screenshot.googleplex.com/8smBYvhVPZHWYs8
- return user messages when using context for entities or props
    - using entities from context: https://screenshot.googleplex.com/AJ2hgmcdQqXVX3v
    - using props from context: https://screenshot.googleplex.com/9Pj3vCWWfSvBN5p


server/lib/nl/detection/context.py
```diff
@@ -100,13 +99,15 @@ def merge_with_context(uttr: nl_uttr.Utterance, default_place: Place = None):
...
-  entities = _detect_entities(uttr)
-
-  # 7. Detect properties leveraging context
+  # 6. Detect properties leveraging context
   properties = _detect_props(uttr)
 
-  # 6. Populate the returned dict
+  # 7. Detect entities leveraging context. Should detect entities after
+  #    properties because we only want to use context if properties didn't use
+  #    context.
```
pradh, line 107: sorry, don't follow.  What's an example for why entities should happen after props?

static/js/apps/nl_interface/debug_info.tsx
```diff
@@ -279,6 +283,17 @@ export function DebugInfo(props: DebugInfoProps): JSX.Element {
...
+            <b>Entity Detection:</b>
+          </Row>
+          <Row>
+            <Col>
+              Entities Detected: {debugInfo.entitiesDetected.join(", ")}
+            </Col>
+          </Row>
+          <Row>
+            <Col>Entities Resolved: {debugInfo.entitiesResolved}</Col>
+          </Row>
```
pradh, line 296: Could you attach this screenshot to desc?

server/integration_tests/test_data/e2e_triple/whattranscriptsdoesithave/chart_config.json
```diff
@@ -0,0 +1,98 @@
...
+    "exploreMore": {},
+    "mainTopics": [],
+    "parentPlaces": [],
+    "parentTopics": [],
+    "peerPlaces": [],
+    "peerTopics": []
+  },
+  "svSource": "UNKNOWN",
+  "userMessages": [
+    "See relevant information for FGFR1 and Fgfr1 based on the previous query."
```
pradh, line 96: hmm, same name (data issue?)

pradh (approved):
Nice!

server/lib/nl/detection/context.py
chejennifer: if properties use context, we dont' want entities to also use context because then we would jsut be returning the exact same result as the previous query even though the user asked for something else

static/js/apps/nl_interface/debug_info.tsx
chejennifer: done

server/integration_tests/test_data/e2e_triple/whattranscriptsdoesithave/chart_config.json
chejennifer: ah yes, needed to filter out the duplicates, updated

server/routes/nl/helpers.py
```diff
@@ -404,10 +404,9 @@ def abort(error_message: str,
           blocked: bool = False,
           test: str = '',
           client: str = '') -> Dict:
-  query = str(escape(shared_utils.remove_punctuations(original_query)))
-  escaped_context_history = []
-  for ch in context_history:
-    escaped_context_history.append(escape(ch))
+  query = shared_utils.escape_strings(
```
pradh, line 407: Why do we need escape_strings here?

server/routes/nl/helpers.py
```diff
@@ -404,10 +404,9 @@ def abort(error_message: str,
...
+      shared_utils.remove_punctuations(original_query))
+  escaped_context_history = shared_utils.escape_strings(context_history)
```
pradh, line 409: What is the motivation for this change?  Might be better to make changes like in a separate PR (OK for this one)?

Also, are there other places we escape context_history?

shared/lib/utils.py
```diff
@@ -125,3 +126,19 @@ def remove_punctuations(s, include_comma=False):
 
 def is_debug_mode() -> bool:
   return os.environ.get('DEBUG', '').lower() == 'true'
+
+
+# Converts a passed in object and escapes all the strings in it.
+def escape_strings(data):
+  if isinstance(data, dict):
+    for k, v in data.items():
+      data[k] = escape_strings(v)
```
pradh, line 135: Do we not want to escape the key?

shared/lib/utils.py
```diff
@@ -125,3 +126,19 @@ def remove_punctuations(s, include_comma=False):
...
+    return data
+  elif isinstance(data, list):
+    for i, item in enumerate(data):
+      data[i] = escape_strings(item)
+    return data
+  elif isinstance(data, str):
+    return str(escape(data))
+  else:
+    return data
```
pradh, line 144: Is the last `else` an error case?

server/lib/nl/detection/context.py
```diff
@@ -359,9 +359,9 @@ def _detect_entities(uttr: nl_uttr.Utterance) -> List[str]:
   else:
     # If there were entities detected in the previous query, use those entities
     if uttr.prev_utterance and uttr.prev_utterance.entities:
-      entities = uttr.prev_utterance.entities
-      uttr.entities = entities
-      uttr.counters.info('insight_entity_ctx', [e.dcid for e in entities])
+      uttr.entities = uttr.prev_utterance.entities
+      entities = [e.dcid for e in uttr.prev_utterance.entities]
```
pradh, line 363: super nit: uttr.entities, since its shorter?

server/routes/nl/helpers.py
chejennifer: This functionality didn't change, we were always escaping this string, I'm just using a shared util for doing this instead

server/routes/nl/helpers.py
chejennifer: this is because when we just used `escape(ch)`, it turned the context_history into a list of strings instead of a list of dicts, which then caused problems when the context was being loaded for the next query because load_utterance is expecting a dict, not a string. This came up because I added a test case where there is a query that goes down this abort path and then the query after that would error out.

shared/lib/utils.py
chejennifer: no, if it's a number or some other type, we don't have to do anything about it so just return the original item

shared/lib/utils.py
chejennifer: done

server/lib/nl/detection/context.py
chejennifer: done

server/routes/nl/helpers.py
pradh: Hmm, I see, the flow of context in the abort case might be untested/unknown.  What happens with the stats flow?  Might be worth following up...

shared/lib/utils.py
pradh: Add comment?

pradh (approved):
Thanks for the changes!

server/routes/nl/helpers.py
pradh: If you agree, have a TODO in abort maybe, for follow up.

chejennifer: added TODO

shared/lib/utils.py
chejennifer: done

chejennifer:
thanks for the reviews!

PR #4142, opened 2024-04-23
update nodejs differ golden  and doc

shifucun:


tools/nl/nodejs_query_differ/goldens/autopush/disaster.json
```diff
@@ -1,7 +1,7 @@
 {
   "charts": [
     {
-      "data_csv": "label,Count of Fire Event\r\n2024-04,124\r\n2024-03,350\r\n2024-02,234\r\n2024-01,296",
+      "data_csv": "label,Count of Fire Event\r\n2024-04,210\r\n2024-03,350\r\n2024-02,234\r\n2024-01,296",
```
shifucun, line 4: @chejennifer shall we pick a disaster that does not have changing data?

chejennifer: do we know of one that does not have changing data? Or we can also remove the disaster test case because we don't support disaster event for bard

chejennifer (approved):
thanks for improving the documentation!

tools/nl/nodejs_query_differ/goldens/autopush/disaster.json
shifucun: Ajai might know better. If it's hard to find a stable one, okay to remove it.

PR #4143, opened 2024-04-24
[cron testing] updates

chejennifer:
- remove cron testing from dev environment because not really needed & the testing overloads the servers
- update the cron testing job on bard environment to only run the nodejs tests because bard only serves nodejs endpoints
- remove disaster test case because disaster data is unstable (i.e., the count of disaster event will keep changing)
- update the tool for updating nodejs diff goldens to update the goldens for every environment


shifucun (approved):
thanks for the update!

Issue #4145, opened 2024-04-25
[BUG] Double quote characters in directory names cause clone error


mapsiva:
### Describe the bug
Error cloning repository at the same directory. Maybe the directory name containing quotes triggers this error. Please, rename this file and remove quotes from directory name.

Cloning into 'website'...
remote: Enumerating objects: 53683, done.
remote: Counting objects: 100% (1006/1006), done.
remote: Compressing objects: 100% (593/593), done.
remote: Total 53683 (delta 505), reused 733 (delta 341), pack-reused 52677
Receiving objects: 100% (53683/53683), 84.08 MiB | 12.54 MiB/s, done.
Resolving deltas: 100% (38884/38884), done.
fatal: cannot create directory at 'server/integration_tests/test_data/detection_api_multivar/compare"malepopulation"with"femalepopulation"': Invalid argument
warning: Clone succeeded, but checkout failed.
You can inspect what was checked out with 'git status'
and retry with 'git restore --source=HEAD :/'

### To Reproduce
Git client: 2.33.1
OS: DeepIn 23 (debian based distro)

1. git clone https://github.com/datacommonsorg/website.git

### Expected behavior
Successful checkout

mapsiva:
![error](https://github.com/datacommonsorg/website/assets/334335/70f62bae-b636-4bfe-93ff-1cd1ce9e3533)

shifucun:
This is fixed now. Can you try again?

PR #4147, opened 2024-04-25
Remove quote in integration test queries to avoid folder name with quote

shifucun:


shifucun:
To fix https://github.com/datacommonsorg/website/issues/4145

server/integration_tests/explore_test.py
```diff
@@ -261,7 +261,7 @@ def test_detection_multivar(self):
         'compare obesity vs. poverty',
         'show me the impact of climate change on drought',
         'how are factors like obesity, blood pressure and asthma impacted by climate change',
-        'Compare "Male population" with "Female Population"',
+        'Compare Male population with Female Population',
```
pradh, line 264: I think this is part of the test? (Multi-var handling tries to respect the double quotes)

Should we instead replace `"` in https://github.com/datacommonsorg/website/blob/cc3c3f29d3b8b13011295896c59c7bb381f7a290/server/integration_tests/explore_test.py#L91 ?

shifucun: Done

server/integration_tests/explore_test.py
```diff
@@ -61,7 +62,8 @@ def run_detection(self,
       if len(queries) == 1:
         d = ''
       else:
-        d = q.replace(' ', '').replace('?', '').lower()
+        d = re.sub(r'[ ?"]', '', q).lower()
+      print(d)
```
pradh, line 66: Drop print?

pradh (approved):
Thanks!

PR #4149, opened 2024-04-26
[Biomedical DC] Update Landing Page

juliawu:
Updates the landing page of biomedical DC to include recent updates to the mocks:
* New text including links at top of the page (above search bar)
* Update styling of the link to the knowledge graph browser in the KG numeric callouts section
* Example queries have been updated to currently working queries
* Example queries now open in the same tab instead of a new tab
* Notebook cards now include a "open in new tab" icon

Screenshots:
localhost_8080_


juliawu:
Update: switched the order of the label and new-tab icon on the cards. They now look like:

Screenshot 2024-04-26 at 12 38 34 PM

static/js/biomedical/landing/highlights_section.tsx
```diff
@@ -58,6 +76,40 @@ const Highlight = styled.div`
...
+  }
+
+  .icon {
+    font-size: 32px;
+  }
+`;
+
 interface HighlightConfigEntry {
   label: string;
   value: string | number;
```
dwnoble, line 115: update to just `number`

static/js/biomedical/landing/highlights_section.tsx
```diff
@@ -85,18 +137,24 @@ function formatNumber(number: number, locale: string): string {
...
-          <Label key={index}>
-            {callout.label}
-            <Highlight>{value}</Highlight>
-          </Label>
-        );
-      })}
+      <NumbersContainer>
+        {props.config.map((callout, index) => {
+          const value =
+            typeof callout.value === "number"
```
dwnoble, line 143: Remove this logic since the `string` path is no longer used

static/js/biomedical/landing/header_and_search_section.tsx
```diff
@@ -47,6 +47,9 @@ const HeadingContainer = styled(ContentContainer)`
 const StyledText = styled.div`
   flex-grow: 1;
 
+  a {
+    color: #146c2e;
```
dwnoble, line 51: Nit: This color looks like it's used a few places in biomedical. Maybe extract it to `CSS_THEME`?

dwnoble (commented):
Thanks Julia! I found this link a little odd - wdyt @miss-o-soup  ? 
4gjd25Ekg4h6CDt

static/js/biomedical/landing/highlights_section.tsx
juliawu: removed check for number

static/js/biomedical/landing/highlights_section.tsx
juliawu: Updated to just number

static/js/biomedical/landing/header_and_search_section.tsx
juliawu: good callout, moved to CSS_THEME

juliawu (commented):
Thanks for the review! I also updated the click area so only hovering over the text would give a pointer/link.

static/js/biomedical/landing/highlights_section.tsx
miss-o-soup: Hi there, 
Regarding this:
https://screenshot.googleplex.com/6Mc3ibgDgLoeqNU
Yes, I agree that is an odd behaviour, that tile should be in line with the rest of the numbers.
If the text wraps it should bit, but all in all a the moment we are missing to fix all the breaking points and margins, so that would help too.

beets (approved):

PR #4151, opened 2024-04-29
[nl] fix userMessage vs userMessages

chejennifer:
the abort path and the regular path in NL was returning user messages differently. This was causing some cases to actually not show the user message returned by NL. This PR makes the two paths return user messages in the same way - key `userMessages` and value as a list of strings. 

Example:
autopush: https://screenshot.googleplex.com/6T4AjXJDqNr2U8J
local: https://screenshot.googleplex.com/3pKVLCXxUvs9Vbe

### DIFF
--- a/server/routes/nl/helpers.py
+++ b/server/routes/nl/helpers.py
@@ -418,9 +418,7 @@ def abort(error_message: str,
       'config': {},
       'context': escaped_context_history,
       'failure': error_message,
-      'userMessage': {
-          'msg': error_message
-      },
+      'userMessages': [error_message]
   }
 
   if not counters:
--- a/server/integration_tests/test_data/demo_feb2023/query_10/chart_config.json
+++ b/server/integration_tests/test_data/demo_feb2023/query_10/chart_config.json
@@ -9,7 +9,7 @@
     "name": "",
     "place_type": ""
   },
-  "userMessage": {
-    "msg": "Sorry, could not complete your request. No place found in the query."
-  }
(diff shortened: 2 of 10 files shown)


server/routes/nl/helpers.py
```diff
@@ -418,9 +418,7 @@ def abort(error_message: str,
       'config': {},
       'context': escaped_context_history,
       'failure': error_message,
-      'userMessage': {
-          'msg': error_message
-      },
+      'userMessages': [error_message]
```
pradh, line 421: Can this impact non-explore clients talking to the API - un, nodejs?

https://github.com/datacommonsorg/website/blob/99b87eb1d7180a840f93af512f283db2a586864f/experimental/sdg-tracker-un/frontend-tester/UNSD-Website-skeleton/UNSD.Website/ClientApp/src/templates/DataCommons/components/search/Search.tsx#L281

chejennifer: ah, you're right, maybe we should return both userMessage and userMessages then & eventually deprecate userMessage? What is the best practice when it comes to updating a response like this?

pradh: Yes, have both, change client, then drop.

@dwnoble can point you to the repo where the client code is.

chejennifer: updated to have both & a TODO to remove userMessage once UN client is updated

pradh (approved):
Thanks for the edits

PR #4152, opened 2024-04-29
Fetch topic info for custom DCs from KG only.

keyurva:


keyurva:
Closing this PR - decided to go with a different approach.

PR #4161, opened 2024-04-30
Fix a few wrong subject page config params when passing to jinja

shifucun:
GCP error due to this: `UndefinedError: 'None' has no attribute 'CUSTOM'`


juliawu (approved):
Thanks!

shifucun:
thanks for review!

PR #4163, opened 2024-04-30
add script for getting the screenshot diffs between prod and autopush

chejennifer:
This tool gets the link to see the diffs between screenshots taken against autopush.datacommons.org between the last autopush that was pushed to prod & the current autopush

### DIFF
--- a/tools/screenshot_diff/diff.sh
+++ b/tools/screenshot_diff/diff.sh
@@ -1,0 +1,50 @@
+#!/bin/bash
+# Copyright 2024 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -e
+
+prod_commit_date=$(git log -1 --date=format:"%Y_%m_%d_%H_%M_%S" --format="%ad" prod)
+latest_commit_date=$(git log -1 --date=format:"%Y_%m_%d_%H_%M_%S" --format="%ad")
+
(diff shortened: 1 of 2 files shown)


tools/screenshot_diff/diff.sh
```diff
@@ -0,0 +1,50 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -e
+
+prod_commit_date=$(git log -1 --date=format:"%Y_%m_%d_%H_%M_%S" --format="%ad" prod)
```
shifucun, line 18: when I try line 18, got a fatal error:

fatal: ambiguous argument 'prod': unknown revision or path not in the working tree.

shifucun (commented):
this is very nice!!

tools/screenshot_diff/diff.sh
chejennifer: hm, I wonder if you need to pull tags first, can you try on a clean master:
- run `git pull origin --tags` (origin or dc or what you use when you do `git pull x master`)
- then run the script

shifucun: okay, this works! 
I wonder how is "prod" set in Github? And can we also add "git pull" command in this script?

shifucun (approved):

tools/screenshot_diff/diff.sh
chejennifer: updating the "prod" tag is actually part of the release process: https://screenshot.googleplex.com/BDEjF2MVMnzxYQi

I think the problem with adding git pull command in the script is that some people use `origin` and some people use `dc` .. but maybe I can add that as a step in the README?

shifucun: Adding to documentation sg!

For prod tag, I have been missing that (not strictly following the steps). We can maybe monitor how this goes and decide if need to use this https://gist.github.com/rponte/fdc0724dd984088606b0 later.

chejennifer: updated & ok, we can monitor!

chejennifer:
thanks for the review!

PR #4166, opened 2024-04-30
Restrict flask /api/variable-group/info to only use POST and ensure 'dcid' exists in request

shifucun:
Calling this endpoint without dcid will overload the mixer and internal server error. Only POST request is used in the client, hence removing GET to reduce its exposure even more.

### DIFF
--- a/server/routes/shared_api/variable_group.py
+++ b/server/routes/shared_api/variable_group.py
@@ -21,21 +21,19 @@ import server.services.datacommons as dc
 bp = Blueprint("variable-group", __name__, url_prefix='/api/variable-group')
 
 
-@bp.route('/info', methods=['GET', 'POST'])
-def get_variable_group_info():
+@bp.route('/info', methods=['POST'])
+def variable_group_info():
   """Gets the stat var group node information.
 
   This is to retrieve the adjacent nodes, including child stat vars, child stat
   var groups and parent stat var groups for the given stat var group node.
   """
-  if request.method == 'GET':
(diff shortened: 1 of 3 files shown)


server/routes/shared_api/variable_group.py
```diff
@@ -21,21 +21,19 @@
 bp = Blueprint("variable-group", __name__, url_prefix='/api/variable-group')
 
 
-@bp.route('/info', methods=['GET', 'POST'])
-def get_variable_group_info():
+@bp.route('/info', methods=['POST'])
```
pradh, line 24: To confirm, there are no clients that use GET today?  Can we tell from monitoring dashboard?

shifucun: All client uses POST now: 
https://github.com/search?q=repo%3Adatacommonsorg%2Fwebsite%20%2Fapi%2Fvariable-group%2Finfo&type=code

pradh: I thought in the past the Stanford team was using it?

pradh (approved):

server/routes/shared_api/variable_group.py
shifucun: don't quite remember, maybe it's mixer api? Also external client should not use/rely on the flask endpoints to fetch data..

pradh: uh oh, I should not review code, I was thinking this was mixer :)

PR #4168, opened 2024-04-30
Add script to GC old cron testing job outputs

chejennifer:
A script to delete old outputs from cron testing jobs.

I set this script to delete files older than 30 days, but have not run it to completion yet (tested by just deleting the earliest file and breaking out of the loop), so I can still increase or decrease this number as you guys see fit!


tools/delete_cron_outputs/README.md
```diff
@@ -0,0 +1,11 @@
+# Delete Cron Outputs
```
shifucun, line 1: very nice!

I assume this would be part of oncall task?

shifucun (approved):

tools/delete_cron_outputs/README.md
chejennifer: Yeup, plan to add it as one of the oncall GC steps

PR #4177, opened 2024-05-02
[fixit] use the right date for the csv in place charts

chejennifer:
the csv export for place page charts was always returning latest even if the chart showed a different date. This fixes it so that the csv will return the same date as the chart.

prod: https://screenshot.googleplex.com/86h6y9ERRXzowYJ
with the change: https://screenshot.googleplex.com/8fNVtSb9EJN8kW4


dwnoble (approved):
Thanks Jennifer!

PR #4179, opened 2024-05-02
A bunch of pip and npm packages update to fix security alerts

shifucun:
Tested with local instances and locally built custom DC docker image


dwnoble (approved):
Thanks bo!

PR #4180, opened 2024-05-02
Delete old NL interface [Part 1]

pradh:
This has not been used in a long time and deleting this will simplify things.  This PR is for frontend, backend cleanup in next part.

Testing:
image
And a test query on explore page still LG.

NOTE: I did not want to mess with css, in the fear of breaking something.  Left a TODO.


dwnoble (approved):
Thanks Prashanth!

pradh:
Thanks for the review!

PR #4182, opened 2024-05-02
Upgrade d3 to v7.4

shifucun:
Also

* Upgraded jest versions
* Removed babel libs since they are not used.


static/js/chart/draw_d3_map.ts
```diff
@@ -342,9 +342,7 @@ export function getProjection(
       projection = d3.geoEquirectangular().rotate([-100, 0]).precision(0.1);
       break;
     default:
-      projection = isUSAPlace
-        ? geo.geoAlbersUsaTerritories()
-        : d3.geoEquirectangular();
+      projection = isUSAPlace ? d3.geoAlbersUsa() : d3.geoEquirectangular();
```
shifucun, line 345: @chejennifer which page should I verify for this projection change?

chejennifer: you can check any place that has US map charts - probably want to check US and a US state and US territory so maybe: 

/place/country/USA
/place/geoId/06
/place/geoId/72

packages/client/package.json
```diff
@@ -47,14 +44,13 @@
     "@types/react-dom": "^18.2.17",
     "@types/react-syntax-highlighter": "^15.5.11",
     "@types/turf": "^3.5.32",
-    "babel-jest": "^29.7.0",
     "d3-scale": "^4.0.2",
     "jest": "^29.7.0",
```
shifucun, line 48: @dwnoble This file and static/package.json has some common deps. I wonder if some more can be removed from here?

static/js/chart/draw_d3_map.ts
shifucun: Reverted as the lib does not block the version update

static/js/biomedical/protein/chart.ts
```diff
@@ -968,15 +969,16 @@ export function drawVarGeneAssocChart(
     .attr("cy", (d) => y(d.id))
     .attr("r", "6")
     .attr("id", (d, i) => circleIdFunc(i))
-    .style("fill", (d) => ERROR_BAR_VAR_COLOR[d.name])
+    .style("fill", (d: VarGeneDataPoint) => ERROR_BAR_VAR_COLOR[d.name])
     // variant redirect
-    .on("click", (d) => {
+    .on("click", (d: VarGeneDataPoint) => {
```
chejennifer, line 983: I see you updated a lot of the other `.on(<event>, () => {})` calls by adding an extra `event` param in the callback function, do you need to do that here as well?

static/js/tools/timeline/__snapshots__/page.test.tsx.snap
```diff
@@ -1,169 +1,169 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
 exports[`Single place and single stat var 1`] = `
-"<div class=\\"chart-container dc-async-element-holder\\">
-  <div class=\\"card\\">
-    <div class=\\"statVarChipRegion\\">
-      <div class=\\"chip\\"><span class=\\"chip-text chip-clickable-text\\">Age</span><button class=\\"chip-action\\"><i class=\\"material-icons\\">cancel</i></button></div>
+"<div class="chart-container dc-async-element-holder">
```
chejennifer, line 4: is this updated because of jest updates?

chejennifer (approved):
nice!

static/js/tools/timeline/__snapshots__/page.test.tsx.snap
shifucun: Yes!

static/js/biomedical/protein/chart.ts
shifucun: Good catch. Fixed here and a few other places.

PR #4183, opened 2024-05-02
add support for vertex ai in nl server

chejennifer:
- add new model_type field in embeddings.yaml which can be LOCAL or VERTEXAI
- handle VERTEXAI model_type
- split out models and indexes in vertex_ai_endpoints.yaml & in eval tool (in both /server and /tools)


server/config/nl_page/nl_vertex_ai_models.yaml
```diff
@@ -0,0 +1,17 @@
...
+  model: "uae-large-v1-model"
+
+sfr-embedding-mistral:
+  index: "sfr-embedding-mistral-index"
+  model: "sfr-embedding-mistral-model"
+
+cross-encoder-ms-marco-miniilm-l6-v2:
+  model: "cross-encoder-ms-marco-miniilm-l6-v2"
+
+cross-encoder-mxbai-rerank-base-v1:
```
pradh, line 16: hmm, this file is actually the index?  Also, should this be in shared too?

What do you think of the following structure:

Split `vertex_ai_endpoints.yaml` into two:

1. vertex_ai_models.yaml:  Has only the models
2. vertex_ai_indexes.yaml:  Has the names in this file, with a reference to the model (in vertex_ai_models.yaml) and endpoint info.

NL Server only reads #1 fully.

This way, we don't need `sfr-embedding-mistral` separate from `sfr-embedding-mistral-index` and `sfr-embedding-mistral-model`.

nl_server/config.py
```diff
@@ -73,11 +82,15 @@ def load(embeddings_map: Dict[str, Dict[str, str]]) -> List[EmbeddingsIndex]:
   #
   # Download all the models.
   #
-  models_set = set([i.model_name for i in indexes if i.model_name])
+  models_set = set([
+      i.model_name
+      for i in indexes
+      if i.model_name and i.model_type == ModelType.LOCAL
+  ])
```
pradh, line 89: nit: Pull loop outside for readability?

nl_server/config.py
```diff
@@ -73,11 +82,15 @@ def load(embeddings_map: Dict[str, Dict[str, str]]) -> List[EmbeddingsIndex]:
...
   model2path = {d: gcs.download_folder(d) for d in models_set}
   for idx in indexes:
     if idx.model_name:
-      idx.model_local_path = model2path[idx.model_name]
+      idx.model_local_path = model2path.get(idx.model_name, "")
```
pradh, line 93: When won't it exist?  comment

pradh (commented):
Nice!  We should maybe have an integration test?

server/lib/util.py
```diff
@@ -639,3 +640,20 @@ def _get_highest_coverage_date(observation_entity_counts_by_date,
   } for obs in observation_dates]
   best_coverage = max(date_counts, key=lambda date_count: date_count['count'])
   return best_coverage['date']
+
+
+def get_vertex_ai_models():
+  import shared.model.loader as model_loader
```
pradh, line 646: Why import here?

server/lib/util.py
```diff
@@ -639,3 +640,20 @@ def _get_highest_coverage_date(observation_entity_counts_by_date,
...
+  vertex_ai_endpoints = model_loader.load()
+  filepath = os.path.join(
```
pradh, line 648: Should this be ENV FLAG Protected? Because auth will fail for custom DC, etc.

nl_server/loader.py
```diff
@@ -45,6 +46,7 @@ def load_server_state(app: Flask):
   flask_env = os.environ.get('FLASK_ENV')
 
   embeddings_map = _load_yaml(flask_env)
+  vertex_ai_endpoints = model_loader.load()
```
pradh, line 49: This should be ENV flag protected?

server/config/nl_page/nl_vertex_ai_models.yaml
chejennifer: oh I like that, done

nl_server/config.py
chejennifer: done

nl_server/config.py
chejennifer: done

server/lib/util.py
chejennifer: originally in __init__.py, this import was inside the `if os.environ.get('ENABLE_EVAL_TOOL') == 'true'`so was just keeping that consistent

nl_server/loader.py
chejennifer: ah, yes, done

deploy/nl/embeddings.yaml
```diff
@@ -2,19 +2,29 @@ medium_ft:
   store: MEMORY
   embeddings: gs://datcom-nl-models/embeddings_medium_2024_03_14_16_38_53.ft_final_v20230717230459.all-MiniLM-L6-v2.csv
   model: ft_final_v20230717230459.all-MiniLM-L6-v2
+  model_type: LOCAL
```
pradh, line 5: Custom DC also generates the YAML (@keyurva), so could you also update:

https://github.com/datacommonsorg/website/blob/d61af7aff064b0c92db702473c0f7113e5e17dda/tools/nl/embeddings/build_custom_dc_embeddings.py#L140-L150
https://github.com/datacommonsorg/website/blob/d61af7aff064b0c92db702473c0f7113e5e17dda/tools/nl/embeddings/testdata/custom_dc/expected/custom_embeddings.yaml#L4

nl_server/loader.py
```diff
@@ -45,6 +47,9 @@ def load_server_state(app: Flask):
   flask_env = os.environ.get('FLASK_ENV')
 
   embeddings_dict = _load_yaml(flask_env)
+  vertex_ai_models = {}
+  if not is_custom_dc():
```
pradh, line 51: This is only gated by !custom-dc?  We don't want it on prod etc until we setup the backend service, etc.  So for now, only include in autopush?

deploy/nl/embeddings.yaml
```diff
@@ -2,19 +2,29 @@ medium_ft:
...
 medium_lance_ft:
   store: LANCEDB
   embeddings: gs://datcom-nl-models-dev/lancedb/lancedb_embeddings_medium_2024_04_09_07_22_21.ft_final_v20230717230459.all-MiniLM-L6-v2
-  model: ft_final_v20230717230459.all-MiniLM-L6-v2
\ No newline at end of file
+  model: ft_final_v20230717230459.all-MiniLM-L6-v2
+  model_type: LOCAL
+medium_ft_vertex:
```
pradh, line 26: end with `_ft` for consistency?

server/lib/util.py
pradh: oh lol I forget why I did that (maybe to not import aiplatform unnecessarily?). but since this is going to be loaded commonly now, maybe ok putting it on top.

shared/model/loader.py
```diff
@@ -14,39 +14,54 @@
...
-                                 'vertex_ai_endpoints.yaml')
-  with open(model_yaml_file) as f:
+def _load_yaml(filename: str):
+  yaml_file = os.path.join(os.path.dirname(os.path.abspath(__file__)), filename)
+  with open(yaml_file) as f:
     return yaml.full_load(f)
 
 
-def load():
+def load_models(model_types: List[str] = None):
```
pradh, line 30: nit: if the callers pass a single value, then avoid a list?

shared/model/loader.py
```diff
@@ -14,39 +14,54 @@
...
+    aiplatform.init(project=index_info['project_id'],
+                    location=index_info['location'])
+    vector_search_client = aiplatform_v1.MatchServiceClient(
+        client_options={"api_endpoint": index_info['index_endpoint_root']})
+    indexes[index_name] = {
+        'vector_search_client': vector_search_client,
+        'index_endpoint': index_info['index_endpoint'],
+        'index_id': index_info['index_id'],
+    }
+    indexes[index_name].update(models.get(index_info['model'], {}))
```
pradh, line 65: Given this is checked-in yaml, we should probably assert and fail here if indexes refers to a non-existing model (esp since tests exercise this path)?

nl_server/config.py
```diff
@@ -73,11 +82,15 @@ def load(embeddings_map: Dict[str, Dict[str, str]]) -> List[EmbeddingsIndex]:
   #
   # Download all the models.
   #
-  models_set = set([i.model_name for i in indexes if i.model_name])
+  models_set = set()
+  for i in indexes:
+    if i.model_name and i.model_type == ModelType.LOCAL:
+      models_set.add(i.model_name)
```
pradh, line 88: Can we assert or fail if for VERTEXAI model_name isn't in the config[VERTEX_AI_MODELS] map?

pradh (commented):
Looking nice!

server/lib/util.py
chejennifer: done

deploy/nl/embeddings.yaml
chejennifer: done

nl_server/loader.py
chejennifer: done & also included in local and tests (since I added an integration test for using a vertex ai model)

deploy/nl/embeddings.yaml
chejennifer: done

shared/model/loader.py
chejennifer: I was thinking that eventually we may have more types and caller may want a list, but we can change that when the time comes, updated to single value

shared/model/loader.py
chejennifer: done

nl_server/config.py
chejennifer: hm, I tried accessing current_app.config from this point, but I don't think the app config is set yet at this point. I could try passing the config through, but that also feels kind of messy. Maybe it could be better to just have a test to check that embeddings.yaml refers to valid vertex ai models?

deploy/nl/embeddings.yaml
```diff
@@ -2,19 +2,29 @@ medium_ft:
...
   store: LANCEDB
   embeddings: gs://datcom-nl-models-dev/lancedb/lancedb_embeddings_medium_2024_04_09_07_22_21.ft_final_v20230717230459.all-MiniLM-L6-v2
-  model: ft_final_v20230717230459.all-MiniLM-L6-v2
\ No newline at end of file
+  model: ft_final_v20230717230459.all-MiniLM-L6-v2
+  model_type: LOCAL
+medium_vertex_ft:
+  store: MEMORY
+  embeddings: gs://datcom-nl-models/embeddings_medium_2024_03_14_16_38_53.ft_final_v20230717230459.all-MiniLM-L6-v2.csv
+  model: dc-all-minilm-l6-v2-model
```
shifucun, line 29: Given this cross reference of model name and 4 yaml files to manage now. I wonder if we should just have one model_embedding.yaml to hold all these model and embedding information? @pradh

shared/model/vertex_ai_indexes.yaml
```diff
@@ -1,38 +1,23 @@
 dc-all-minilm-l6-v2:
-  type: EMBEDDING
   project_id: "datcom-website-dev"
   location: "us-central1"
-  prediction_endpoint_id: "8518340991868993536"
   index_endpoint_root: "302175072.us-central1-496370955550.vdb.vertexai.goog"
   index_endpoint: "projects/496370955550/locations/us-central1/indexEndpoints/8500794985312944128"
   index_id: "dc_all_minilm_l6_v2_ft_1709655496660"
+  model: "dc-all-minilm-l6-v2-model"
```
shifucun, line 7: why not put this in deploy/nl/embeddings.yaml so we can see all embeddings in one place?

shared/model/vertex_ai_models.yaml
```diff
@@ -0,0 +1,29 @@
+dc-all-minilm-l6-v2-model:
```
shifucun, line 1: why not put this in deploy/nl/models.yaml?

nl_server/loader.py
```diff
@@ -121,12 +128,15 @@ def _load_yaml(flask_env: str) -> Dict[str, str]:
   return embeddings_map
 
 
-def _update_app_config(app: Flask, nl_model: NLAttributeModel,
+def _update_app_config(app: Flask,
+                       nl_model: NLAttributeModel,
```
shifucun, line 132: find nl_model and vertex_ai_models a bit confusing. I feel either have one model class to represent both or to make the nl_model more specific.

nl_server/loader.py
```diff
@@ -45,6 +47,9 @@ def load_server_state(app: Flask):
   flask_env = os.environ.get('FLASK_ENV')
 
   embeddings_dict = _load_yaml(flask_env)
+  vertex_ai_models = {}
```
shifucun, line 50: I find it a bit hard to following load_server_state process here. As I mentioned in the other comment, maybe we should have one yaml file to hold all the model/embedding information. So this function would read the yaml and switch between the "type" and process?

Imaging if we decide to use huggingface endpoint or other 3p model endpoint. The change would just be add a new switch path in here. But now there are "vertext_ai" specific naming, functions and logic route. It would not be easy to add a new type of endpoint.

shifucun (changes requested):
This is very cool!

Have some high level question about making the config and loading more generic.

shifucun:
As we can foresee more changes to the model/embedding yaml files, to avoid breaking old users, can we add a version tag in the the yaml file and have versioned loader function like loader_v1() loader_v2().

If we do this, then have everything model/embedding related in one yaml file would make more sense. @pradh

shared/model/vertex_ai_indexes.yaml
chejennifer: from my udnerstanding, right now vertex_ai indexes are only used by nl eval tools & this pr is to add support for vertex_ai models in nl server

nl_server/loader.py
chejennifer: it looks like nl_model is used for detecting verbs, maybe I can rename it to nl_attribute_model?

shifucun: maybe we can remove "nl" (which applies to everything) and use "attribute_model"?

nl_server/loader.py
chejennifer: chatted & will chat more offline about changes to make after this PR goes in

pradh (approved):
Thanks for all the changes!

deploy/nl/embeddings.yaml
keyurva: Will it cause issues if `MODEL_TYPE` is not present? If it does, can we prevent it? In general, for custom DCs, we need to preserve backward compatibility.

chejennifer:
thanks for the reviews

shifucun (approved):

PR #4186, opened 2024-05-02
Added website frontend caching to NL endpoints

dwnoble:
Before: "How has access to electricity improved in world" takes 3.44s
Screenshot 2024-05-02 at 4 51 06 PM

After: Same query takes 23ms
Screenshot 2024-05-02 at 4 50 32 PM


pradh (approved):
Nice find!

PR #4187, opened 2024-05-03
SV Differ: compare model and/or store

pradh:
More deets in README.

Same index diff: https://storage.mtls.cloud.google.com/datcom-embedding-diffs/shanth_medium_ft_2024_05_02_18_12_18.html

Different index diff: https://storage.mtls.cloud.google.com/datcom-embedding-diffs/shanth_medium_ft_2024_05_02_18_21_37.html


shifucun (approved):
This is so much cleaner now!

pradh:
> This is so much cleaner now!

Totally!  thanks for the review...

PR #4188, opened 2024-05-03
Move reranking to NL Server

pradh:
This is so that all Vertex AI calls go through NL Server.  Additionally, the multi-var case also gets reranking support (if enabled).


nl_server/embeddings_map.py
```diff
@@ -60,24 +67,27 @@ def reset_index(self, embeddings_info: EmbeddingsConfig):
...
-      model = None
       try:
-        if model_info.type == ModelType.VERTEXAI and allow_vertex_ai():
-          model = VertexAIModel(model_info)
+        if (allow_vertex_ai() and model_info.type == ModelType.VERTEXAI):
+          if model_info.usage == ModelUsage.EMBEDDINGS:
+            model = VertexAIEmbeddingsModel(model_info)
+            self.name_to_emb_model[model_name] = model
+          else:
+            # model_info.usage == ModelUsage.RERANKING
```
shifucun, line 81: uncomment?

nl_server/rerank.py
```diff
@@ -0,0 +1,92 @@
...
+
+from nl_server.ranking import RerankingModel
+import shared.lib.detected_variables as vars
+
+# Given a list of query <-> candidates pairs as input, returns
+# a corresponding list of scores.
+RerankCallable = Callable[[List[tuple[str, str]]], List[float]]
+
+
+def rerank(rerank_model: RerankingModel,
```
shifucun, line 25: would be nice to have 2 helper functions for prepare data and group result and have test for them, given the logic involved.

maybe a TODO for future.

nl_server/embeddings_map.py
pradh: Done

nl_server/rerank.py
pradh: Good point, added a TODO

pradh:
Thanks for the review!

shifucun (approved):

PR #4189, opened 2024-05-03
Disable filter queries unless LLM infers things

pradh:
This change is safe to disable because neither our landing page queries, nor our demo queries rely on filter.


chejennifer (approved):
nice!

pradh:
Thanks!

PR #4191, opened 2024-05-03
Fallback to LLM for real (instead for safety) and fully

pradh:
We used `hybridsafety` before launch worrying what LLM would hallucinate/do.  But since then Gemini happened, and our own heuristics have proven not very good, so lets try `hybrid` as a step towards `LLM` for real perhaps.

As well, do not distinguish whether we use just place or sv detected by LLM, when we fallback.  Its hard to reason, and simpler to just use what the LLM detected fully.

TODO:  Run adversarial tests for sanity!  (Although, for real, the QPS to LLM should be the same, the only difference is whether we only use LLM result for safety check or fulfill)


chejennifer (approved):
nice!

pradh:
Thanks for the review!

pradh:
This will be in adversarial to be run as part of index changes.

shifucun:
nice!

PR #4193, opened 2024-05-03
deprecate userMessage

chejennifer:
Now that UN has switched over to using userMessages, deprecating userMessage in NL response


dwnoble (approved):

pradh (commented):
To confirm, have they rolled out their client?

Otherwise, things would break if we only rollout to their backend with this change.

chejennifer:
> They've rolled it out to their staging instance, but not prod. We could wait to merge this until it reaches prod

Ok, I can wait until it reaches prod, can you let me know?

dwnoble (commented):
Will do

dwnoble:
This PR looks good to merge. UN updated the site on their end to migrate from `userMessage` to `userMessages`, and I ran some tests locally and verified it's working for me.

chejennifer:
> This PR looks good to merge. UN updated the site on their end to migrate from `userMessage` to `userMessages`, and I ran some tests locally and verified it's working for me.

Thank you for checking, can you approve this PR then if this looks safe?

dwnoble (approved):

PR #4203, opened 2024-05-07
Remove /nl urls from screenshot tests

chejennifer:
The nl app has been deleted, so remove the /nl urls which were causing exceptions in the screenshot tests.

This also lead to finding out that cloudbuild service account didn't have IAM permissions to write to logs, so granted log permissions for the cloudbuild service account


shifucun (approved):
thanks!

pradh (approved):
Thanks for this!  Sorry for the miss.

PR #4204, opened 2024-05-07
Fix nginx redirects.

keyurva:
* The fix is to use `$http_host` instead of `$host` in `nginx.conf` so that both host and port are used for redirects.

#### Before fix

```
$ curl -H "Host: foo.com" -is http://localhost:8080/admin | grep "Location" 
Location: http://foo.com/admin/
$ curl -H "Host: foo.com:9090" -is http://localhost:8080/admin | grep "Location"
Location: http://foo.com/admin/
```

#### After fix

```
$ curl -H "Host: foo.com" -is http://localhost:8080/admin | grep "Location"     
Location: http://foo.com/admin/
$ curl -H "Host: foo.com:9090" -is http://localhost:8080/admin | grep "Location"
Location: http://foo.com:9090/admin/
```

### DIFF
--- a/build/web_compose/nginx.conf
+++ b/build/web_compose/nginx.conf
@@ -6,7 +6,7 @@ http {
 
     location / {
         proxy_pass http://0.0.0.0:7070;
-        proxy_set_header Host $host;
+        proxy_set_header Host $http_host;
         proxy_read_timeout 3600;
     }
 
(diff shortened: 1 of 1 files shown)


shifucun:
also test in cloud run?

shifucun (approved):

PR #4206, opened 2024-05-07
Add support for loading custom dc topic cache.

keyurva:
* Custom DC topic cache will be loaded from `$USER_DATA_PATH/datacommons/nl/custom_dc_topic_cache.json`.
* When custom dc topic cache is loaded, it is available under `dc=custom` and more importantly, it is merged into a main dc cache (by default into `main` but overridable via a `"dc"` field in the cache) at startup.
  + Merging them at runtime (like embedding index) will take some effort so going with merged at load for now.
  + We can implement runtime merging later if it is useful.
* There is no hot reload of topic cache.


shared/lib/custom_dc_util.py
```diff
@@ -27,6 +30,15 @@ def get_user_data_path() -> str:
...
 
 
+def get_topic_cache_path() -> str:
+  base_path = get_user_data_path()
+  if not base_path:
+    return base_path
+  if is_gcs_path(base_path):
+    return join_gcs_path(base_path, _TOPIC_CACHE_PATH)
+  return os.path.join(base_path, _TOPIC_CACHE_PATH)
+
```
pradh, line 40: Why is this in shared lib when used only by server?

shared/lib/custom_dc_util.py
```diff
@@ -27,6 +30,15 @@ def get_user_data_path() -> str:
   return os.environ.get('USER_DATA_PATH', '')
 
 
+def get_topic_cache_path() -> str:
```
pradh, line 33: This and others should probably have `custom` in their name?

server/lib/topic_cache.py
```diff
@@ -154,7 +191,32 @@ def load_files(fpath_list: List[str], name_overrides: Dict,
...
-  topic_cache_map = {}
+  topic_cache_map: dict[str, TopicCache] = {}
   for dc, fpath_list in TOPIC_CACHE_FILES.items():
     topic_cache_map[dc] = load_files(fpath_list, name_overrides, dc)
+
+  if is_custom_dc():
+    custom_dc_topic_cache = _load_custom_dc_topic_cache(name_overrides)
+    if custom_dc_topic_cache:
+      topic_cache_map[DCNames.CUSTOM_DC.value] = custom_dc_topic_cache
+      topic_cache_map[DCNames.MAIN_DC.value]._merge(custom_dc_topic_cache)
```
pradh, line 202: This may not be the right thing to do, e.g. for undata.  Should we maybe get what DC the topic-cache is for from the JSON itself?

server/lib/topic_cache.py
```diff
@@ -154,7 +191,32 @@ def load_files(fpath_list: List[str], name_overrides: Dict,
...
+    return None
+  return load_files([local_path], name_overrides, DCNames.CUSTOM_DC.value)
+
+
+def _get_local_custom_dc_topic_cache_path() -> str:
+  path = get_topic_cache_path()
+  if not path:
+    logging.info("No Custom DC topic cache path specified.")
+    return path
+  logging.info("Custom DC topic cache will be downloaded from: %s", path)
```
pradh, line 219: nit: Should the "download" msg be under if is_gcs_path?

server/lib/topic_cache.py
```diff
@@ -154,7 +191,32 @@ def load_files(fpath_list: List[str], name_overrides: Dict,
 
 
 def load(name_overrides: Dict) -> Dict[str, TopicCache]:
-  topic_cache_map = {}
+  topic_cache_map: dict[str, TopicCache] = {}
```
pradh, line 194: nit: localfile consistency `Dict`?

server/lib/nl/common/utils.py
```diff
@@ -42,15 +42,15 @@
...
 
 
 def is_svg(sv):
-  return sv.startswith("dc/g/")
+  return sv.startswith("dc/g/") or sv.startswith("c/g/")
 
 
 def is_svpg(sv):
-  return sv.startswith("dc/svpg/")
+  return sv.startswith("dc/svpg/") or sv.startswith("c/svpg/")
```
pradh, line 54: comment what `c/` prefixes are?

pradh (commented):
Thanks!

server/lib/nl/common/utils.py
keyurva: Done.

server/lib/topic_cache.py
keyurva: Using the built-in types in this file instead of the [deprecated](https://docs.python.org/3/library/typing.html#typing.Dict) `typing.*` types.

server/lib/topic_cache.py
keyurva: Are you implying we should not merge custom with main or that we should support `dc=<any>` vs a constant `dc=custom`?

server/lib/topic_cache.py
keyurva: I wanted to log in both cases. Changed "downloaded" to just "loaded".

shared/lib/custom_dc_util.py
keyurva: Done.

shared/lib/custom_dc_util.py
keyurva: To localize knowledge about relative custom dc paths (`datacommons/nl/custom_dc_topic_cache.json`) in this file. Want to move others such as (`datacommons/nl/custom_embeddings.yaml`) and any future ones in this file as well.

If you feel otherwise, I can move it.

server/lib/topic_cache.py
pradh: For UN, we should merge the topic-cache into `undata` not `main`.  So let the topic-cache builder make that choice?

shared/lib/custom_dc_util.py
pradh: To me, `shared` implies it can be safely called by multiple servers (nl-server, web-server, model-server), but that's probably not true here?

server/lib/topic_cache.py
pradh: We need to do a bulk s/Dict/dict/ then for consistency

shared/lib/custom_dc_util.py
pradh: If we should be mindful of something, add a comment on top?  I would additionally point out this is really only for sharing between Web and NL server.

server/tests/lib/topic_cache_test.py
```diff
@@ -0,0 +1,102 @@
...
+
+
+@patch.dict(topic_cache.TOPIC_CACHE_FILES, {
+    DCNames.MAIN_DC.value: [_MAIN_DC_TOPIC_CACHE_FILE],
+    DCNames.UNDATA_DC.value: [_UNDATA_TOPIC_CACHE_FILE]
+},
+            clear=True)
+class TestTopicCacheLoader(unittest.TestCase):
+
+  @patch.dict(os.environ, {"IS_CUSTOM_DC": "false"})
```
pradh, line 80: Appreciate the detailed tests here, but could you please add some comments on what we're doing?

server/lib/topic_cache.py
```diff
@@ -68,10 +82,35 @@ class Node:
...
 class TopicCache:
 
-  def __init__(self, out_map: Dict[str, Node], in_map: Dict[str, Set[str]]):
+  def __init__(self, dc: str, out_map: Dict[str, Node],
+               in_map: Dict[str, Dict[str, Set[str]]]):
+    self.dc = dc
     self.out_map = out_map
     self.in_map = in_map
 
+  def _merge(self, other: Self):
```
pradh, line 91: nit: `_merge` -> `merge` since its called from outside?

pradh (commented):
Thanks for the edits!

shared/lib/custom_dc_util.py
keyurva: Done.

server/lib/topic_cache.py
keyurva: Done.

server/lib/topic_cache.py
```diff
@@ -150,11 +195,43 @@ def load_files(fpath_list: List[str], name_overrides: Dict,
...
+      topic_cache_map[merge_dc]._merge(custom_dc_topic_cache)
+
   return topic_cache_map
+
+
+def _load_custom_dc_topic_cache(name_overrides: Dict) -> TopicCache:
+  local_path = _get_local_custom_dc_topic_cache_path()
+  if not local_path:
+    return None
+  return load_files([local_path], name_overrides, DCNames.CUSTOM_DC.value)
```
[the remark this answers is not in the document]
keyurva: Good idea. Done.

server/tests/lib/topic_cache_test.py
keyurva: Done.

pradh (approved):

keyurva:
Thanks for the review.

PR #4207, opened 2024-05-07
Fix typos in SV title overrides

juliawu:
Ran spellcheck over `chart_titles_by_sv.json` and `chart_titles_by_sv_sdg.json`, checked each hit manually and fixed the true positive hits.

Things NOT changed:
* Typos in DCIDs
* Sentence vs Title case


server/config/nl_page/chart_titles_by_sv.json
```diff
@@ -1325,47 +1325,47 @@
...
   },
   "Annual_Amount_Emissions_Oxo_Oxochromiooxy_Chromium_SCC_2_InternalCombustionEngines": {
     "blocklist": [],
-    "title": "Annual Emissions: Oxo Chromium, Internal Combustion Engine"
+    "title": "Annual Emissions: Chromium Oxide, Internal Combustion Engine"
   },
   "Annual_Amount_Emissions_Oxo_Oxochromiooxy_Chromium_SCC_3_IndustrialProcesses": {
     "blocklist": [],
-    "title": "Annual Emissions: Oxo Chromium, Industrial Processes"
+    "title": "Annual Emissions: Chromium Oxide, Industrial Processes"
```
chejennifer, line 1360: Is Oxo Chromium and Chromium oxide the same thing? (I don't know the chemistry)

chejennifer (approved):
thank you!! please also update the integration test goldens with ./run_test.sh -g

server/config/nl_page/chart_titles_by_sv.json
juliawu: I think the original Oxo(Oxochromioxy) Chromium refers to Cr2O3 (two chromiums, three oxygen atoms) which is called Chromium (III) oxide, but has many possible names depending on convention (see [its wikipedia entry](https://en.wikipedia.org/wiki/Chromium(III)_oxide)).

Chromium Oxide would be the class of CrXOY compounds which Chromium (III) oxide falls under. I couldn't find the original name used by the EPA (their data portal requires permissions to access) but they do have fact sheets about air emissions that refer to chromium compounds (https://www.epa.gov/sites/default/files/2015-07/documents/acrylic_etc_fr_fs061507.pdf and https://www.epa.gov/sites/default/files/2016-09/documents/chromium-compounds.pdf) so I've omitted the (III) from the SV name just in case.

As far as I'm aware, oxo chromium is not a valid name for any chemical.

pradh (commented):
I think these titles are generated from https://github.com/datacommonsorg/website/blob/26ccbd2071650fc495ef775f77fe3176dce1b50f/tools/nl/titles/README.md#L4 so we should ideally fix the CSVs and regenerate?

PR #4209, opened 2024-05-08
Fix Google Sans font breaking in Custom DC

juliawu:
Adds alternate body fonts to the tools, because Google Sans is not an open source font. This will allow other fonts to render in custom DC instead of the default serif font of the browser.

### DIFF
--- a/static/css/base.scss
+++ b/static/css/base.scss
@@ -58,10 +58,10 @@ $theme-colors: (
   "primary": $dc-primary-color,
 );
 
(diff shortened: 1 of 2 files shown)


dwnoble (approved):
Thanks Julia! As a follow up we should remove google sans entirely from custom dc and replace it with Poppins https://fonts.google.com/?query=poppins

juliawu:
Thanks for the review! Will add Poppins as followup PR.

PR #4211, opened 2024-05-08
[nl server] remove local cache

chejennifer:
this removes the local cache from the nl server & is part of the changes for updating the embeddings.yaml structure

The difference in time taken for the server to restart after a change
with cache: 1-2s
without cache: 6-7s


shifucun (approved):
Making things simpler!

Is there documentation or script that using ~/.datacommons should be cleaned?

chejennifer:
> Making things simpler!
> 
> Is there documentation or script that using ~/.datacommons should be cleaned?

good call, updated

PR #4220, opened 2024-05-10
Added caching to /node-geojson endpoint (speeds up explore page loading)

dwnoble:
- Added flask caching to /node-geojson endpoint
- Added helper for caching post requests, setting the key to the full path (url + query params) and json-encoded post body
- Added helper for caching detect and detect-and-fulfill POST requests (does not cache if a contextHistory is set)
- Bug fix for cached /detect and /detect-and-fulfill endpoints that was ignoring contextHistory for cached queries

Before: load time is between 1.1s and 2.5s
Screenshot 2024-05-10 at 2 16 30 PM

After: load time is between ~00ms  and 900ms 
Screenshot 2024-05-10 at 2 17 09 PM


juliawu (approved):
Thanks for adding caching!

shifucun (commented):
just to confirm the cache works for the POST request with json data.

dwnoble:
> just to confirm the cache works for the POST request with json data.

Good catch- looks like i need to make a custom cache key for the post request. 
(Note to self: also run some tests on https://github.com/datacommonsorg/website/pull/4186)

dwnoble:
@shifucun @juliawu  added a helper to use the post body in cache keys- can you take another look?

server/routes/explore/api.py
```diff
@@ -84,7 +87,9 @@ def detect():
 #  - childEntityType: A type of child entity (optional)
 #
 @bp.route('/fulfill', methods=['POST'])
-@cache.cached(timeout=TIMEOUT, query_string=True)
+@cache.cached(timeout=TIMEOUT,
```
pradh, line 90: Will doing cache for POST without the custom key cause trouble?  That is, should this PR go in before prod push?

server/routes/explore/helpers.py
```diff
@@ -466,3 +467,20 @@ def _set_blocked(err_json: Dict):
...
+
+def explore_post_body_cache_key():
+  """
+  Builds flask cache key for /detect and /detect-and-fulfill POST
+  requests. Does not cache requests with a contextHistory.
+  """
+  body_object = request.get_json()
+  if "contextHistory" in body_object and body_object["contextHistory"]:
+    # Don't cache requests with a contextHistory by returning None.
+    # All context histories are be unique due to a unique session ID, so caching
```
pradh, line 480: nit: will

pradh, line 480: If that's the only reason, it might be worth constructing the key without the session ID in that case?  Have a TODO?

server/routes/explore/api.py
dwnoble: We should get this PR in the release if possible (cc @keyurva ). Without the custom key, contextHistory is ignored for cached /fulfill and /detect-and-fulfill requests. the query string itself is still used, just not the context history.

server/routes/explore/helpers.py
dwnoble: fixed

dwnoble: Added a todo

dwnoble:
> Wonder is there a size limit for the caching key?

Looks like the key size for redis is 512mb: https://redis.io/docs/latest/develop/use/keyspace/#:~:text=The%20maximum%20allowed%20key%20size%20is%20512%20MB.

server/routes/explore/helpers.py
```diff
@@ -450,4 +451,23 @@ def abort(error_message: str,
...
+  requests. Does not cache requests with a contextHistory.
+  """
+  body_object = request.get_json()
+  if "contextHistory" in body_object and body_object["contextHistory"]:
+    # Don't cache requests with a contextHistory by returning None.
+    # All context histories are unique due to a unique session ID, so caching
+    # isn't helpful.
+    # TODO: Strip out session id, debug info, and potentially other fields from
+    # contextHistory so it can be cached
+    return None
```
shifucun, line 469: I see, so if the cache key is None, flask-caching will do nothing even using the default url path?

shifucun (approved):
this is great!

server/routes/explore/helpers.py
dwnoble: yeah that's right

server/routes/explore/helpers.py
```diff
@@ -19,6 +19,7 @@
 
 import flask
 from flask import current_app
+from flask import request
```
pradh, line 22: nit: is this needed now?

pradh (approved):
Thanks!

server/routes/explore/helpers.py
dwnoble: Removed

PR #4221, opened 2024-05-10
Fix flaky golden test

chejennifer:
SV scores from sv detector can change a little between each run (after 6 decimal digits). So sort variables by truncating their score to 6 decimal digits and if variables have the same truncated score, sort alphabetically


server/integration_tests/explore_test.py
```diff
@@ -235,463 +248,8 @@ def handle_i18n_response(self, resp, i18n_lang):
 
     self.assertTrue(success, f"wanted: {i18n_lang}, got {detected}")
 
-  def test_detection_basic(self):
```
shifucun, line 238: revert these deletions?

server/integration_tests/explore_test.py
```diff
@@ -235,463 +248,8 @@ def handle_i18n_response(self, resp, i18n_lang):
...
-                       check_detection=True)
-
-  def test_detection_context(self):
-    self.run_detection('detection_api_context', [
-        'States with highest PHDs', 'Commute in tracts of California',
-        'Compare with Nevada', 'Correlate with asthma',
-        'countries with greenhouse gas emissions',
-        'median income in Santa Clara county and Alameda county'
-    ])
-
```
pradh, line 307: Don't delete all the tests? :)

server/integration_tests/explore_test.py
```diff
@@ -111,6 +111,19 @@ def handle_response(self,
...
+    # runs. Sort by scores cut off after 6 digits after the decimal and for
+    # variables with the same truncated score, sort alphabetically
+    resp_var_to_score = {}
+    for i, sv in enumerate(dbg['sv_matching']['SV']):
+      score = dbg['sv_matching']['CosineScore'][i]
+      resp_var_to_score[sv] = float("{:.6f}".format(score))
+    if 'variables' in resp:
+      sorted_variables = sorted(resp['variables'],
+                                key=lambda x: (-resp_var_to_score[x], x))
+      resp['variables'] = sorted_variables
```
shifucun, line 125: should all the golden scores change to 6 digit now?

server/integration_tests/explore_test.py
chejennifer: good catch, updated

server/integration_tests/explore_test.py
chejennifer: good catch, updated

server/integration_tests/explore_test.py
chejennifer: no, this just updates the ordering of variables

shifucun (approved):

pradh (approved):
Thanks!

Would the "proper fix" be to change NL Server to be more deterministic?  Should we have TODO?

chejennifer:
> Thanks!
> 
> Would the "proper fix" be to change NL Server to be more deterministic? Should we have TODO?

added TODO

PR #4222, opened 2024-05-11
move vertex ai model info to deployment values

chejennifer:
We want each environment to be able to use different project/location/endpoints for each vertex ai model so move the specification of those values to the deployment yamls. When running locally, read from autopush values

### DIFF
--- a/nl_server/config.py
+++ b/nl_server/config.py
@@ -18,6 +18,7 @@ from enum import Enum
 import json
 import logging
 import os
+from pathlib import Path
 from typing import Dict
 
 import yaml
@@ -107,14 +108,16 @@ class EmbeddingsConfig:
 #
 # Get Dict of vertex ai model to its info
 #
-def _get_vertex_ai_model_map() -> Dict[str, any]:
+def _get_vertex_ai_model_info() -> Dict[str, any]:
+  # This is the path to model info when deployed in gke.
   if os.path.exists(_VERTEX_AI_MODEL_CONFIG_PATH):
     with open(_VERTEX_AI_MODEL_CONFIG_PATH) as f:
       return json.load(f) or {}
+  # If that path doesn't exist, assume we are running locally and use the values
+  # from autopush.
   else:
-    autopush_env_values = os.path.join(
-        os.path.dirname(os.path.dirname(os.path.abspath(__file__))),
-        f'deploy/helm_charts/envs/autopush.yaml')
+    current_file_path = Path(__file__)
+    autopush_env_values = f'{current_file_path.parent.parent}/deploy/helm_charts/envs/autopush.yaml'
     with open(autopush_env_values) as f:
       autopush_env = yaml.full_load(f)
       return autopush_env['nl']['vertex_ai_models']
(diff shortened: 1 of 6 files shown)


nl_server/config.py
```diff
@@ -99,13 +104,30 @@ class EmbeddingsConfig:
...
 
+#
+# Get Dict of vertex ai model to its info
+#
+def _get_vertex_ai_model_map() -> Dict[str, any]:
+  if os.path.exists(_VERTEX_AI_MODEL_CONFIG_PATH):
+    with open(_VERTEX_AI_MODEL_CONFIG_PATH) as f:
+      return json.load(f) or {}
+  else:
+    autopush_env_values = os.path.join(
```
pradh, line 115: comment what these modes are?

nl_server/config.py
```diff
@@ -124,12 +147,17 @@ def parse_v1(embeddings_map: Dict[str, any]) -> EmbeddingsConfig:
                                             usage=model_info['usage'],
                                             gcs_folder=model_info['gcs_folder'])
     elif model_type == ModelType.VERTEXAI:
+      if model_name not in vertex_ai_model_map:
```
pradh, line 150: throw exception!?

pradh (approved):
Nice!

pradh (commented):
Maybe wait for Bo's review, I don't have much of a clue about helm?

nl_server/config.py
chejennifer: hm, throwing an exception worries me because I've only added the vertex_ai_model map to autopush and dev, so if we deploy this code anywhere else that spins up an nl server, then this exception will get thrown .. 

one thing I could do is check if vertex ai is allowed here before doing this?

nl_server/config.py
```diff
@@ -99,13 +104,30 @@ class EmbeddingsConfig:
...
+#
+# Get Dict of vertex ai model to its info
+#
+def _get_vertex_ai_model_map() -> Dict[str, any]:
+  if os.path.exists(_VERTEX_AI_MODEL_CONFIG_PATH):
+    with open(_VERTEX_AI_MODEL_CONFIG_PATH) as f:
+      return json.load(f) or {}
+  else:
+    autopush_env_values = os.path.join(
+        os.path.dirname(os.path.dirname(os.path.abspath(__file__))),
```
shifucun, line 116: nit: Can do the following:

from pathlib import Path
current_file_path = Path(__file__)
autopush_env_values = current_file_path.parent.parent / 'deploy/helm_charts/envs/autopush.yaml'

nl_server/config.py
```diff
@@ -99,13 +104,30 @@ class EmbeddingsConfig:
   models: Dict[str, ModelConfig]
 
 
+#
+# Get Dict of vertex ai model to its info
+#
+def _get_vertex_ai_model_map() -> Dict[str, any]:
```
shifucun, line 110: nit: _get_vertex_ai_model_info() and use info instead of map below?

nl_server/config.py
shifucun: looks like we need a bit in embeddings.yaml to indicate whether a model should be loaded or not?

chejennifer: hm, should this be a bit in embeddings.yaml or instance specific

shifucun: This should be instance specific i think. Different instance (bio, sdg, un, custom dc, dev) will use very different indexes. Maybe add them to the instance yaml?

nl_server/config.py
chejennifer: done

nl_server/config.py
chejennifer: done

PR #4230, opened 2024-05-14
NL: Make score threshold a configurable parameter per model

pradh:
Every model config now specifies a `score_threshold` parameter which gets used in place of the hard-coded `0.5`.  If a model doesn't set it, the value defaults to `0.5` (so backward compatible for custom DC).

This PR intends to make NO real functional change, except when using the `uae-large-v1-model` model.  And this is seen by the only "real" diff being in `detection_api_uae_idx/chart_config.json` file.

Also, when looking at empty variables return empty.  As validation, the [which are the richest counties in California] query has [empty SV match table](https://screenshot.googleplex.com/DtEGSwnhKsngUYy).


server/lib/nl/detection/detector.py
```diff
@@ -232,21 +232,29 @@ def construct_for_explore(entities: List[str], vars: List[str], child_type: str,
...
+    # it `scores` > `sv_threshold`.
     sv_detection = types.SVDetection(query='',
                                      single_sv=dutils.VarCandidates(
                                          svs=vars,
                                          scores=[0.51] * len(vars),
                                          sv2sentences={}),
                                      prop=empty_var_candidates(),
-                                     multi_sv=get_multi_sv(vars, cmp_vars, 1.0))
+                                     multi_sv=get_multi_sv(vars, cmp_vars, 1.0),
+                                     sv_threshold=0.5,
```
shifucun, line 245: should update 0.5 here?

shared/lib/constants.py
```diff
@@ -375,18 +375,27 @@
...
 
-# The default Cosine score threshold beyond which Stat Vars
-# are considered a high confidence match.
-SV_SCORE_HIGH_CONFIDENCE_THRESHOLD = 0.7
+# The Cosine score bump (aka increase over default model-threshold)
+# to be applied for a high confidence match.
+# See compute_final_threshold()
+#
+# bump of 0.4 => a threshold of 0.7 if default is 0.5
+#             => a threshold of 0.82 if default is 0.7
```
shifucun, line 383: for the two example here, how is 0.4 used? Have more explanation?

shifucun (approved):
thanks for the quick update!

server/lib/nl/detection/detector.py
pradh: It actually doesn't matter as long as score is > threshold here, and made that clear.  This is the case of fulfillment-only flow after scores are employed and no model is involved.

shared/lib/constants.py
pradh: Added more comments, ptal?

pradh (commented):
Thanks for the review!

shared/lib/constants.py
shifucun: ah, i see.

shifucun (approved):
thanks for update!

shared/lib/detected_variables.py
```diff
@@ -109,10 +113,11 @@ def var_detection_result_to_dict(res: VarDetectionResult) -> Dict:
   return result
 
 
-def dict_to_var_detection_result(input: Dict) -> VarDetectionResult:
-  return VarDetectionResult(single_var=dict_to_var_candidates(input),
-                            multi_var=dict_to_multivar_candidates(
-                                input.get('MultiSV', {})))
+def test_dict_to_var_detection_result(input: Dict) -> VarDetectionResult:
```
chejennifer, line 116: is this only used in testing (because of the 'test' prefix)? if so, add a comment

chejennifer (approved):
nice!

shared/lib/detected_variables.py
pradh: Done.

pradh:
Thanks for the review!

PR #4233, opened 2024-05-14
[nl server] don't try to read vertex ai model info if custom dc

chejennifer:


nl_server/config.py
```diff
@@ -112,6 +113,10 @@ class EmbeddingsConfig:
 # Get Dict of vertex ai model to its info
 #
 def _get_vertex_ai_model_info() -> Dict[str, any]:
+  # Custom DC doesn't use vertex ai so just return an empty dict
+  if is_custom_dc():
```
keyurva, line 117: Do we want custom dcs to use them?

If we do, the model info can be added in the custom dc docker image [here](https://github.com/datacommonsorg/website/blob/63b5002f1c1a0b9dbbdca0ad1a0bb3c66801136e/build/web_compose/Dockerfile#L67).

chejennifer: hm, I don't think we want custom dcs to use them right now

keyurva: SG. Consider adding a TODO if we plan to in the future.

keyurva (approved):
Thanks for the fix!

nl_server/config.py
chejennifer: done

PR #4234, opened 2024-05-14
Major update of base embedding descriptions

shifucun:
This is a major effort to improve base DC variable and topic descriptions. 

* Keep only 1 well curated description for each stat var.
* Use consistent descriptions across similar stat vars.
* Removed a bunch of multi PV variables which mostly adds noise.
* Remove variables for "other" that needs context to represent.

As a result. There is no need for various alternative csv files for the base DC embeddings.


nl_server/model/sentence_transformer.py
```diff
@@ -35,14 +35,8 @@ def __init__(self, model_info: LocalModelConfig):
     if model_info.gcs_folder:
       logging.info(f'Downloading tuned model from: {model_info.gcs_folder}')
       model_path = gcs.download_folder(model_info.gcs_folder)
```
pradh, line 37: Maybe skip the if condition, since model_info.gcs_folder is a required parameter now?

pradh (approved):
WOOHOO!!

nl_server/model/sentence_transformer.py
shifucun: Done

PR #4235, opened 2024-05-14
 [NL Server] Introduce ServerConfig and config reader

shifucun:
Use ServerConfig as the NL server runtime config source of truth. It holds the usable model index information.

Make a standard process to obtain this server config by:
1. read catalog config from ALL available paths and merge them. This makes custom DC user path one of the path so no special treatment is needed. Moving forwards, read_config() could take an argument if necessary.
2. read environment config from env yaml file. This still has a check of is_custom_dc() call which could be removed if making the xx_ENV_PATH ranked. We will see if that's better.
3. build the server config from catalog and env.

process

The Registry object can be solely constructed from a server_config now. 

This makes it easier for build embeddings tool. The tool can just set a default_index in its yaml file (even in str format) and obtain the registry. At this point, all the resource are downloaded and ready to use. No special check is needed anywhere.

---
This PR also constructs Flask app after the registry is built. This makes it much simple to do `/healthz` check.


nl_server/config_reader.py
```diff
@@ -0,0 +1,214 @@
...
+  for p in all_paths:
+    logging.info('Loading index and model catalog from: %s', p)
+    with open(p) as f:
+      partial_catalog = yaml.safe_load(f.read())
+      # read version
+      ver = catalog.version
+      this_ver = partial_catalog['version']
+      if (ver and this_ver and ver != this_ver):
+        raise ValueError(
+            f'Inconsistent version in config files: {ver} and {this_ver}')
```
keyurva, line 117: Perhaps include the path in the error message - helpful for debugging.

nl_server/config_reader.py
```diff
@@ -0,0 +1,214 @@
...
+
+      if this_ver not in _SUPPORTED_VERSIONS:
+        raise ValueError(f'Unknown version: {this_ver}')
```
keyurva, line 120: s/Unknown/Unsupported

nl_server/config_reader.py
```diff
@@ -0,0 +1,214 @@
...
+_CATALOG_USER_PATH_SUFFIX = 'datacommons/nl/custom_catalog.yaml'
+_CATALOG_TMP_PATH = '/tmp/catalog.yaml'
+
+# env paths
+_ENV_CODE_PATH = (Path(__file__).parent /
+                  '../deploy/helm_charts/envs/autopush.yaml')
+_ENV_MOUNT_PATH = '/datacommons/nl/env.yaml'
+_ENV_USER_PATH = Path(__file__).parent / 'custom_dc_env.yaml'
+
+_SUPPORTED_VERSIONS = [1]
```
keyurva, line 50: nit: can be a set for efficient lookups.

nl_server/config_reader.py
```diff
@@ -0,0 +1,214 @@
...
+
+      if this_ver not in _SUPPORTED_VERSIONS:
+        raise ValueError(f'Unknown version: {this_ver}')
+      catalog.version = this_ver
+
+      indexes = catalog.indexes
+      models = catalog.models
+
+      # TODO: when another version is added, extract separate version read as
+      # a function.
```
keyurva, line 127: Consider already extracting as a function?

nl_server/config_reader.py
```diff
@@ -0,0 +1,214 @@
...
+  logging.debug(json.dumps(asdict(catalog), indent=2))
+  return catalog
+
+
+def read_env() -> Env:
+  """Reads the envs
+  """
+  # TODO: Make this generic by checking a user provided config file path
+  if custom_dc_util.is_custom_dc():
+    c = from_dict(data_class=Env, data=yaml.safe_load(open(_ENV_USER_PATH)))
```
keyurva, line 161: Hmm - this "leaks" the fact that `Env` is a `dataclass`. How about adding a class level `from_dict` method to `Env` itself and using `dacite` in there? It will keep the dataclass aspect encapsulated in the class and will be more flexible.

nl_server/config_reader.py
```diff
@@ -0,0 +1,214 @@
...
+  elif os.path.exists(_ENV_MOUNT_PATH):
+    c = from_dict(data_class=Env, data=yaml.safe_load(open(_ENV_MOUNT_PATH)))
+  else:
+    with open(_ENV_CODE_PATH) as f:
+      full_nl_config = yaml.safe_load(f.read())
+      c = from_dict(data_class=Env, data=full_nl_config['nl']['env'])
+  logging.debug(json.dumps(asdict(c), indent=2))
```
keyurva, line 168: nit: Add a short log message prefix so we know what's getting logged?

nl_server/config_reader.py
```diff
@@ -0,0 +1,214 @@
...
+  Merges the catalog and env into a server config.
+  """
+  server_config = ServerConfig(
+      version=catalog.version,
+      default_indexes=env.default_indexes,
+      enable_reranking=env.enable_reranking,
+      indexes={},
+      models={},
+  )
+  server_config.default_indexes = [
```
keyurva, line 183: Instead of mutating `server_config.default_indexes` here, consider creating the final list at the outset and setting it in the `ServerConfig` constructor itself.

nl_server/custom_dc_env.yaml
```diff
@@ -0,0 +1,10 @@
+# TODO: move this to deploy folder. This is here for the easy of
```
keyurva, line 1: nit:s/easy/ease

nl_server/custom_dc_env.yaml
```diff
@@ -0,0 +1,10 @@
...
+# custom DC loading for now.
+
+default_indexes:
```
keyurva, line 4: Can you add a short comment for the 3 properties here - `default_indexes`, `enabled_indexes`, `enable_reranking`?

nl_server/routes.py
```diff
@@ -117,36 +97,38 @@ def detect_verbs():
...
+@bp.route('/api/server_config/', methods=['GET'])
 def embeddings_version_map():
-  return json.dumps(asdict(current_app.config[config.CATALOG_KEY]))
+  r: Registry = current_app.config[REGISTRY_KEY]
+  server_config = r.server_config()
+  return json.dumps(asdict(server_config))
 
 
 @bp.route('/api/load/', methods=['GET'])
 def load():
```
keyurva, line 112: Can you ensure with this change, a hot reload of embeddings works as expected?

keyurva (commented):
Loving all the refactoring!

nl_server/config_reader.py
shifucun: good idea!

shifucun (commented):
thanks for the review!

keyurva (approved):
Thanks for the updates!

nl_server/config.py
```diff
@@ -103,95 +90,25 @@ class Catalog:
...
-# Parse the input `embeddings.yaml` dict representation into Catalog
-# object.
-#
-def parse(catalog_dict: Dict[str, any], vertex_ai_model_info: Dict[str, any],
-          reranking_enabled: bool) -> Catalog:
-  if catalog_dict['version'] == 1:
-    return parse_v1(catalog_dict, vertex_ai_model_info, reranking_enabled)
-  else:
-    raise AssertionError('Could not parse catalog: unsupported version.')
+  def from_dict(d):
```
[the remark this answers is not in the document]
shifucun: can not annotate since this is in the class definition..

nl_server/config.py
```diff
@@ -103,95 +90,25 @@ class Catalog:
...
-          prediction_endpoint_id=vertex_ai_model_info[model_name]
-          ['prediction_endpoint_id'],
-          location=vertex_ai_model_info[model_name]['location'])
-    else:
-      raise AssertionError(
-          'Error parsing information for model {model_name}: unsupported type {model_type}'
-      )
-  return Catalog(indexes=indexes, models=models)
+@dataclass(kw_only=True)
+class ServerConfig:
```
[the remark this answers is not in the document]
shifucun: Added comments and a diagram for the relation.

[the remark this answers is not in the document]
shifucun: Done

nl_server/config_reader.py
shifucun: I really think the yaml conversion should happen offline during import or system update, not at runtime..

nl_server/routes.py
```diff
@@ -43,31 +41,7 @@
...
-                                               [query]).get(query)
-    if result and result.svs:
-      default_embeddings_loaded = True
-      return 'OK', 200
-    else:
-      logging.warning(f'Health Check Failed: query "{query}" failed!')
-  else:
-    logging.warning('Health Check Failed: Default index not yet loaded!')
-  return 'Service Unavailable', 500
+  return 'NL Server is healthy', 200
```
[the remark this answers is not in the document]
shifucun: Done

nl_server/config_reader.py
```diff
@@ -0,0 +1,215 @@
...
+  # Only add enabled indexes
+  indexes = {}
+  for index_name in env.enabled_indexes:
+    if index_name not in catalog.indexes:
+      logging.warning('Index %s not found in catalog', index_name)
+      continue
+    index_config = catalog.indexes[index_name]
+    indexes[index_name] = index_config
+    model_name = index_config.model
+    models[model_name] = catalog.models[model_name]
```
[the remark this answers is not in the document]
pradh: What would be nice is to tag some types of log entries as ALERT entries, and alert if we ever see these.   But for future.

nl_server/config_reader.py
```diff
@@ -0,0 +1,215 @@
...
+
+  # Add vertex AI model info
+  for model_name in models:
+    if model_name in env.vertex_ai_models:
+      m = models[model_name]
+      v = env.vertex_ai_models[model_name]
+      m_dict, v_dict = asdict(m), asdict(v)
+      merged = _merge_dicts(m_dict, v_dict)
+      models[model_name] = VertexAIModelConfig(**merged)
```
pradh, line 202: i find it slight odd that the same dict starts with a dict and gets turned into ModelConfig.  should we have separate dicts with clearer types?

nl_server/config_reader.py
```diff
@@ -0,0 +1,215 @@
...
+      logging.warning('Index %s not found in catalog', index_name)
+      continue
+    index_config = catalog.indexes[index_name]
+    indexes[index_name] = index_config
+    model_name = index_config.model
+    models[model_name] = catalog.models[model_name]
+
+  # Add vertex AI model info
+  for model_name in models:
+    if model_name in env.vertex_ai_models:
```
pradh, line 197: If models[model_name] is VERTEXAI and we don't have an entry in vertex_ai_models, should we log error?

nl_server/config_reader.py
```diff
@@ -0,0 +1,215 @@
...
+
+
+def get_server_config(catalog: Catalog, env: Env) -> ServerConfig:
+  """
+  Merges the catalog and env into a server config.
+  """
+  default_indexes = [x for x in env.default_indexes if x in catalog.indexes]
+
+  # Add reranking models
+  models = {}
```
pradh, line 178: Could you add type info?

pradh: Or, start with empty ServerConfig, and fill it up, so types are clear

nl_server/config_reader.py
```diff
@@ -0,0 +1,215 @@
...
+
+  server_config = ServerConfig(
+      version=catalog.version,
+      default_indexes=default_indexes,
+      enable_reranking=env.enable_reranking,
+      indexes=indexes,
+      models=models,
+  )
+
+  config_str = json.dumps(asdict(server_config), indent=2)
```
pradh, line 212: Given multiple places we do {asdict, dumps, log} I think you could use a log_asdict helper?

`def log_asdict(msg_prefix: str, obj: Any)`

nl_server/__init__.py
```diff
@@ -1,4 +1,4 @@
-# Copyright 2023 Google LLC
+# Copyright 2024 Google LLC
```
pradh, line 1: Can you do a search for embeddings.yaml, and ensure references are fixed?

nl_server/routes.py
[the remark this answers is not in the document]
pradh: Thanks!

nl_server/config_reader.py
```diff
@@ -0,0 +1,215 @@
...
+from nl_server.config import ModelUsage
+from nl_server.config import ServerConfig
+from nl_server.config import StoreType
+from nl_server.config import VertexAIIndexConfig
+from nl_server.config import VertexAIModelConfig
+from shared.lib import custom_dc_util
+from shared.lib import gcs
+
+# catalog paths
+_CATALOG_CODE_PATH = Path(__file__).parent / '../deploy/nl/catalog.yaml'
```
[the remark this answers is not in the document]
shifucun: Done

nl_server/routes.py
```diff
@@ -83,29 +57,35 @@ def search_vars():
...
-  if not idx:
-    idx = default_index_type
-
-  registry: Registry = current_app.config[config.REGISTRY_KEY]
-
   skip_topics = False
   if request.args.get('skip_topics'):
     skip_topics = True
 
+  r: Registry = current_app.config[REGISTRY_KEY]
```
pradh, line 64: nit: `r` is easy to miss with github codebrowsing, consider `reg` or `registry`.

nl_server/routes.py
```diff
@@ -117,36 +97,38 @@ def detect_verbs():
...
+  return json.dumps(r.get_attribute_model().detect_verbs(query.strip()))
 
 
-@bp.route('/api/embeddings_version_map/', methods=['GET'])
+@bp.route('/api/server_config/', methods=['GET'])
 def embeddings_version_map():
-  return json.dumps(asdict(current_app.config[config.CATALOG_KEY]))
+  r: Registry = current_app.config[REGISTRY_KEY]
+  server_config = r.server_config()
+  return json.dumps(asdict(server_config))
```
pradh, line 108: woohoo!

nl_server/config_reader.py
```diff
@@ -0,0 +1,215 @@
...
+_CATALOG_MOUNT_PATH = '/datacommons/nl/catalog.yaml'
+_CATALOG_USER_PATH_SUFFIX = 'datacommons/nl/custom_catalog.yaml'
```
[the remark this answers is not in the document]
shifucun: Done

nl_server/routes.py
```diff
@@ -117,36 +97,38 @@ def detect_verbs():
...
+
+def _get_indexes(r: Registry, idx_types: List[str]) -> List[Embeddings]:
+  embeddings: List[Embeddings] = []
+  for idx in idx_types:
+    try:
+      emb = r.get_index(idx)
+      if emb:
+        embeddings.append(emb)
+    except Exception as e:
+      logging.warning(f'Failed to load index {idx}: {e}')
```
pradh, line 130: Why do we need this try-except?  get_index seems to do a `.get()` on the dict?

nl_server/config_reader.py
```diff
@@ -0,0 +1,215 @@
...
+
+# env paths
+_ENV_CODE_PATH = (Path(__file__).parent /
+                  '../deploy/helm_charts/envs/autopush.yaml')
+_ENV_MOUNT_PATH = '/datacommons/nl/env.yaml'
+_ENV_USER_PATH = Path(__file__).parent / 'custom_dc_env.yaml'
+
+_SUPPORTED_VERSIONS = set([1])
```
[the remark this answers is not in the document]
shifucun: Done

nl_server/config_reader.py
```diff
@@ -0,0 +1,215 @@
...
+      if this_ver not in _SUPPORTED_VERSIONS:
+        raise ValueError(f'Unsupported version: {this_ver} in file {p}')
+      catalog.version = this_ver
+
+      indexes = catalog.indexes
+      models = catalog.models
+
+      # TODO: when another version is added, extract separate version read as
+      # a function.
+      if this_ver == 1:
```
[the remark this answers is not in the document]
shifucun: Done

nl_server/config_reader.py
```diff
@@ -0,0 +1,215 @@
...
+    with open(p) as f:
+      partial_catalog = yaml.safe_load(f.read())
+      # read version
+      ver = catalog.version
+      this_ver = partial_catalog['version']
+      if (ver and this_ver and ver != this_ver):
+        raise ValueError(
+            f'Inconsistent version in config file {p}: {ver} and {this_ver}')
+
+      if this_ver not in _SUPPORTED_VERSIONS:
```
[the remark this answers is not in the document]
shifucun: it would like it to be string, but it was number. Updated to string now, also in the catalog.yaml as well.

nl_server/config_reader.py
```diff
@@ -0,0 +1,215 @@
...
+          # read models
+          for model_name, model_config in partial_catalog['models'].items():
+            model_type = model_config['type']
+            match model_type:
+              case ModelType.LOCAL:
+                models[model_name] = LocalModelConfig(**model_config)
+              case ModelType.VERTEXAI:
+                models[model_name] = VertexAIModelConfig(**model_config)
+              case _:
+                raise ValueError(f'Unknown model type: {model_type}')
```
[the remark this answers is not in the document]
shifucun: that's a very nice catch!!

nl_server/config_reader.py
```diff
@@ -0,0 +1,215 @@
...
+  return catalog
+
+
+def read_env() -> Env:
+  """Reads the envs
+  """
+  # TODO: Make this generic by checking a user provided config file path
+  if custom_dc_util.is_custom_dc():
+    c = Env.from_dict(yaml.safe_load(open(_ENV_USER_PATH)))
+  elif os.path.exists(_ENV_MOUNT_PATH):
```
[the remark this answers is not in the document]
shifucun: added comments. this is in GKE deployment mounted volume

nl_server/config_reader.py
shifucun: aha, i moved that to this formation from last round of comment. I added typing for the struct now.

nl_server/config_reader.py
```diff
@@ -0,0 +1,215 @@
...
+      c = Env.from_dict(full_nl_config['nl']['env'])
+  logging.debug('NL Env:\n%s', json.dumps(asdict(c), indent=2))
+  return c
+
+
+def get_server_config(catalog: Catalog, env: Env) -> ServerConfig:
+  """
+  Merges the catalog and env into a server config.
+  """
+  default_indexes = [x for x in env.default_indexes if x in catalog.indexes]
```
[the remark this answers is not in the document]
shifucun: This is actually a case in custom DC..

so the user has an env that claims to use default [custom_ft, medium_ft] but the index has not been loaded. Once they built the custom_ft and reload the new index, both the default indexes can be used now.

nl_server/config_reader.py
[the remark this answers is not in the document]
shifucun: I tend to crash here, since the input yaml is invalid?

nl_server/config_reader.py
shifucun: good point. Now checking very vertex model from models

nl_server/config_reader.py
shifucun: the value is always a model config, so they were both converted to dict, merged, and then converted back to the class object. Added comment on preferring the env values.

nl_server/config_reader.py
shifucun: Done

nl_server/__init__.py
shifucun: all in the build custom dc tool comments. I will update them when I get to change that tool.

nl_server/routes.py
shifucun: updated to reg, not using registry to avoid override the module name.

nl_server/routes.py
shifucun: seems to be a wrong change. reverted back.

pradh (approved):
Thanks for the changes!   This is a great refactor, should really help make things less brittle and more readable!

nl_server/config_reader.py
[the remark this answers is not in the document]
shifucun: okay, this is indeed a bit implicit though. So the custom DC really should also upload their env.yaml (which is now a checked in file). Once the import tool does that, the check here can be more tight.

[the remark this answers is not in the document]
pradh: that'd be helpful!

nl_server/config_reader.py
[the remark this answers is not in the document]
shifucun: ok, added explicit raise here.

shifucun:
Thank you both for this very detailed review!

nl_server/__init__.py
```diff
@@ -43,20 +43,34 @@ def create_app():
...
+    # Below is a safe check to ensure that the model and embedding is loaded.
+    # TODO: uncomment this code when fixing the crash issue during test.
+
+    # server_config = reg.server_config()
+    # idx_type = server_config.default_indexes[0]
+    # embeddings = reg.get_index(idx_type)
+    # query = server_config.indexes[idx_type].healthcheck_query
+    # result = search.search_vars([embeddings], [query]).get(query)
+    # if not result or not result.svs:
+    #   raise Exception(f'Registry does not have default index {idx_type}')
```
shifucun, line 67: this is causing crash during test. This is likely due to some threading issue during tests. comment out this for now and will come back to this next week.

PR #4240, opened 2024-05-15
NL: avoid NULL access in healthz

pradh:
Custom DC crashes here.  This likely won't fix it, but we might know more...

### DIFF
--- a/nl_server/routes.py
+++ b/nl_server/routes.py
@@ -37,10 +37,11 @@ bp = Blueprint('main', __name__, url_prefix='/')
 def healthz():
   nl_embeddings = current_app.config[config.NL_EMBEDDINGS_KEY].get_index(
       config.DEFAULT_INDEX_TYPE)
-  result: VarCandidates = search.search_vars(
-      [nl_embeddings], ['life expectancy'])['life expectancy']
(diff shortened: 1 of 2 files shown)


nl_server/routes.py
```diff
@@ -37,10 +37,11 @@
 def healthz():
   nl_embeddings = current_app.config[config.NL_EMBEDDINGS_KEY].get_index(
       config.DEFAULT_INDEX_TYPE)
-  result: VarCandidates = search.search_vars(
-      [nl_embeddings], ['life expectancy'])['life expectancy']
-  if result.svs and 'Expectancy' in result.svs[0]:
-    return 'OK', 200
+  if nl_embeddings:
```
keyurva, line 40: Is there a place we can log a warning if the default model is not loaded?

keyurva (approved):
Thanks.

nl_server/routes.py
pradh: If the model isn't loaded we throw an exception all the way up (which is what happened in the first case).  In this case, the model gets loaded, but oddly the healthz happens before that, though, I don't see what in this flow changed to make that happen.

PR #4249, opened 2024-05-17
Round to 5 decimal for embedding score to prevent flakiness

shifucun:


nl_server/search.py
```diff
@@ -89,7 +89,7 @@ def _rank_vars(candidates: EmbeddingsResult,
         sv2sentences[dcid].add(c.sentence)
 
   for sv, score in sorted(sv2score.items(),
-                          key=lambda item: (-item[1], item[0])):
+                          key=lambda item: (-round(item[1], 6), item[0])):
```
pradh, line 92: Maybe have TODO to truncate score using model parameter?

pradh (approved):
Nice!

nl_server/search.py
pradh: Oh also leave a comment why we do it

shifucun: Done

nl_server/search.py
```diff
@@ -88,8 +88,11 @@ def _rank_vars(candidates: EmbeddingsResult,
             dvars.SentenceScore(sentence=c.sentence, score=c.score))
         sv2sentences[dcid].add(c.sentence)
 
+  # TODO: truncate the score based on model parameters from yaml
+  # Same model would produce different scores after certain decimals, so we want
+  # to round to 6 decimal places to make the score and rank stable.
   for sv, score in sorted(sv2score.items(),
-                          key=lambda item: (-item[1], item[0])):
+                          key=lambda item: (-int(item[1] * 1e6), item[0])):
```
pradh, line 95: curious why multiply and turn to int?

shifucun: revert back to round. The tests were failing and I thought it's the float comparing issue. But looks like the original sort did not really take effect. Looks like if I create a new variable of the sorted result it works..

PR #4251, opened 2024-05-17
[Custom DC] Fix Disappearing Stat Var bug on Map and Scatter tool

juliawu:
This PR fixes some bugs in the map and scatter tool on Custom DC

## Bug 1: Disappearing stat vars

When a user selects a place + child place type in the dropdowns, the custom variables sometimes disappear from the sidebar. For example, when using the default OECD preview data, the OECD variables disappear from the sidebar when selecting "Asia" and "Country" in the dropdowns.

This happens because the custom variables don't have enough geographic coverage to meet the 10 place minimum set in PR #2314.

This PR sets this minimum to 1 place for custom DC (so that variables with no places are still filtered out) and leaves the 10 place minimum unchanged on main DC.

Screenshot 2024-05-17 at 4 06 45 PM

Screenshot 2024-05-28 at 3 03 55 PM

Screenshot 2024-05-28 at 3 04 25 PM

Screenshot 2024-05-28 at 3 04 50 PM

## Bug 2: ResizeObserver error

This PR also fixes a console error on the scatter tool encountered while debugging:
Screenshot 2024-05-17 at 4 31 53 PM


server/templates/custom_dc/custom/base.html
```diff
@@ -46,6 +46,9 @@
   {% if OVERRIDE_CSS_PATH %}
   <link href="{{ OVERRIDE_CSS_PATH }}" rel="stylesheet">
   {% endif %}
+  <script>
+    globalThis.isCustomDC = {{ config['CUSTOM']|int }};
```
dwnoble, line 50: this approach will require existing custom DC users to add this line in order to see the fix. is there a way we can push the fix to them requiring them to change any code?

static/js/tools/map/stat_var_chooser.tsx
```diff
@@ -114,10 +114,11 @@ export function StatVarChooser(props: StatVarChooserProps): JSX.Element {
       selectSV={(svDcid) =>
         selectStatVar(dateCtx, statVar, display, placeInfo, svDcid)
       }
-      numEntitiesExistence={Math.min(
-        NUM_ENTITIES_EXISTENCE,
-        samplePlaces.length
-      )}
+      numEntitiesExistence={
```
dwnoble, line 117: Can you extract this logic to a function and add some comments explaining the isCustomDC vs NUM_ENTITIES_EXISTENCE check

server/templates/custom_dc/custom/base.html
juliawu: Instead of setting a value in base.html, switched to using a config variable in app_env/*.py files, and setting globalThis on the tools pages themselves, instead of the template which could be overridden.

static/js/tools/map/stat_var_chooser.tsx
juliawu: Good idea. Extracted to a helper function and added an explaining comment.

juliawu:
Thanks for the review! Because Custom DC now uses the new visualization tools by default, I also fixed this bug in the new visualization tools. Client (npm) tests have been updated to reflect the new default numEntitiesExistence = 1 when the flask config variable is not set.

beets (commented):
thanks for the fix (and sorry for the delay!)

it might be easier to reason about the setting if it were the default value to filter (1 for all tools, 10 for main dc in map and scatter). then the value can be easily overriden via config.

short of that, could we update the config var name from USE_STAT_VAR_FILTERING to FILTER_LOW_GEO_COVERAGE_STAT_VAR (or a more concise version of that)

beets (approved):
thanks for the update (and the concise flag name)

juliawu:
Thanks for the reviews!

For our records: took beet's advice and switched to setting the filter value in the config, and renamed the variable name to "MIN_STAT_VAR_GEO_COVERAGE"

PR #4252, opened 2024-05-20
Add generic GCS download functions and use them in NL server/tools

shifucun:
Added 3 generic GCS functions:

* download_blob() which downloads a blob(file or folder) given bucket, blob name and local path.
* download_blob_by_path() which is a wrapper of download_blob() that takes gcs path
* maybe_download() which downloads the blob under 'local_path_prefix' and keeps the gcs path structure, and skip the download if already exists.

Use these functions throughout NL apps and remove unused functions.


shared/lib/gcs.py
```diff
@@ -16,81 +16,107 @@
...
+    local_path: The local path to download the blob to.
+  """
+  if not is_gcs_path(gcs_path):
+    raise ValueError(f"Invalid GCS path: {gcs_path}")
+  bucket_name, blob_name = get_path_parts(gcs_path)
+  return download_blob(bucket_name, blob_name, local_path, use_anonymous_client)
+
+
+def maybe_download(gcs_path: str,
+                   local_path_prefix='/tmp',
```
keyurva, line 98: nit: Any reason why `local_path_prefix` here and `local_path` in the other methods? Perhaps being more explicit and calling it something like `local_parent_dir` may be clearer.

keyurva (approved):
Very cool simplification!

shared/lib/gcs.py
shifucun: trying to be consistent with the naming, so local_path(*) means the download destination. I guess `local_path_root` maybe better here?

keyurva: That works.

shared/lib/gcs.py
```diff
@@ -16,81 +16,107 @@
...
+    local_path_prefix: The local path prefix to download the folder to.
+    use_anonymous_client: Whether to use anonymous client to download the file.
+  Returns:
+    The local path of the downloaded file/folder.
+  """
+  if not is_gcs_path(gcs_path):
+    raise ValueError(f"Invalid GCS path: {gcs_path}")
+  bucket_name, blob_name = get_path_parts(gcs_path)
+  local_path = os.path.join(local_path_prefix, bucket_name, blob_name)
+  if os.path.exists(local_path):
```
pradh, line 118: When running locally, if the gcs_path is a directory, the below can be a useful check because after reboot, I've noticed the files in /tmp/ are deleted but directory exists.

```
  if os.path.exists(local_path) and len(os.listdir(local_path)) > 0:
    # When running locally, we may already have downloaded the path.
    # But sometimes after restart, the directories in `/tmp` become
    # empty, so ensure that's not the case.
    return local_path
```

shared/lib/gcs.py
shifucun: Fixed

shared/lib/gcs.py
```diff
@@ -16,81 +16,107 @@
...
+    return local_path
+  if download_blob_by_path(gcs_path, local_path, use_anonymous_client):
+    return local_path
```
pradh, line 121: Why not use download_blob here given you have bucket_name and blob_name?

shared/lib/gcs.py
```diff
@@ -16,81 +16,107 @@
...
-  return download_file(bucket, filename, use_anonymous_client)
 
-
-def _get_local_path(bucketname: str, filename: str) -> str:
-  return os.path.join(TEMP_DIR, bucketname, filename)
+  bucket = storage_client.bucket(bucket_name)
+  blobs = bucket.list_blobs(prefix=blob_name)
+  count = 0
+  for blob in blobs:
+    if blob.name.endswith("/"):
```
pradh, line 58: Comment what this means?

shared/lib/gcs.py
```diff
@@ -16,81 +16,107 @@
...
 
 
-#
-# Downloads the `filename` from GCS to TEMP_DIR
-# and returns its path.
-#
-def download_file(bucket: str,
-                  filename: str,
-                  use_anonymous_client: bool = False) -> str:
+def download_blob(bucket_name: str,
```
pradh, line 36: This is a nice general function!

shared/lib/gcs_test.py
```diff
@@ -0,0 +1,77 @@
...
+from pathlib import Path
+import unittest
+
+from parameterized import parameterized
+import pytest
+
+import shared.lib.gcs as gcs
+
+
+class TestGCSFunctions(unittest.TestCase):
```
pradh, line 26: Add some comments on what each test is testing?

server/lib/nl/common/bad_words.py
```diff
@@ -74,8 +74,8 @@ class BannedWords:
 # Loads a list of bad words from a text file.
 #
 def load_bad_words() -> BannedWords:
-  local_file = gcs.download_file(bucket=GLOBAL_CONFIG_BUCKET,
-                                 filename=BAD_WORDS_FILE)
+  local_file = gcs.maybe_download(
+      f'gs://{GLOBAL_CONFIG_BUCKET}/{BAD_WORDS_FILE}')
```
pradh, line 78: Instead of hardcoding `gs://`, should we have a `make_path(bucket, suffix)` helper?

pradh (commented):
Very nice refactor!!

shared/lib/gcs.py
shifucun: added check

shared/lib/gcs.py
pradh: If download_blob_by_path is no longer used, can get rid of it?

pradh (approved):

shared/lib/gcs.py
```diff
@@ -16,81 +16,115 @@
...
+    local_path_root: The local root path to download the gcs resources to.
+    use_anonymous_client: Whether to use anonymous client to download the file.
+  Returns:
+    The local path of the downloaded file/folder.
+  """
+  if not is_gcs_path(gcs_path):
+    raise ValueError(f"Invalid GCS path: {gcs_path}")
+  bucket_name, blob_name = get_path_parts(gcs_path)
+  local_path = os.path.join(local_path_root, bucket_name, blob_name)
+  if os.path.exists(local_path) and len(os.listdir(local_path)) > 0:
```
pradh, line 123: Oh, one quick check, does `os.listdir(file_path)` work?  If not might need another condition

pradh, line 123: Oh, one quick check, does `os.listdir(file_path)` work?  If not might need another condition

shifucun: yes, need to check is a dir first.

shared/lib/gcs.py
shifucun: This would be a general function when downloading gcs folder to a destination folder without keeping the original path, more like a regular `cp`

shifucun:
Thanks for the review!

PR #4262, opened 2024-05-22
Update index to remove some unwanted stat vars for embeddings index.

shifucun:


pradh (approved):
Might also be worth checking if this is part of topic?

PR #4263, opened 2024-05-22
Improve / fix naming in NL server

shifucun:
This PR aims to greatly disambiguate code logic by using distinct naming for various object. Currently there are 10+ names like embedding_xxx for class, object, dict and files with a lot of clashes.

With this change:

- Use `catalog` to replace `embeddings_config` object and represent the all available (may not be used) model/index config
- Use `registry` to replace `embedddings_map` object which holds all the model/client objects
- Use `env` to replace `embeddings_spec` which is the additional config per environment
- Make all the field naming the same across yaml file and dataclass object. This will help for parsing and error checking.
- Use A and A_dict to differentiate dataclass object vs yaml dict.
- Move attribute model part of registry


deploy/helm_charts/envs/worldbank.yaml
```diff
@@ -49,9 +49,11 @@ serviceGroups:
...
 nl:
   enabled: true
-  embeddingsSpec:
-    defaultIndex: "medium_ft"
-    enabledIndexes: ["medium_ft", "sdg_ft", "undata_ft"]
+  env:
+    default_indexes:
+      - medium_ft
+    enabled_indexes:
+      - medium_ft
```
pradh, line 56: intend to remove others?

nl_server/loader.py
```diff
@@ -23,23 +22,22 @@
...
 _CUSTOM_EMBEDDINGS_YAML_PATH = 'datacommons/nl/custom_embeddings.yaml'
-_EMBEDDINGS_SPEC_PATH: str = '/datacommons/nl/embeddings_spec.json'
+_ENV_PATH: str = '/datacommons/nl/env.yaml'
 _LOCAL_ENV_VALUES_PATH: str = f'{Path(__file__).parent.parent}/deploy/helm_charts/envs/autopush.yaml'
 
 
 @dataclass
-class EmbeddingsSpec:
-  default_index: str
+class Env:
```
pradh, line 39: Add class comments?

nl_server/config.py
```diff
@@ -23,10 +22,9 @@
 CUSTOM_DC_INDEX: str = 'custom_ft'
 
 # App Config constants.
-ATTRIBUTE_MODEL_KEY: str = 'ATTRIBUTE_MODEL'
-NL_EMBEDDINGS_KEY: str = 'NL_EMBEDDINGS'
-NL_EMBEDDINGS_VERSION_KEY: str = 'NL_EMBEDDINGS_VERSION_MAP'
-EMBEDDINGS_SPEC_KEY: str = 'EMBEDDINGS_SPEC'
+REGISTRY_KEY: str = 'NL_REGISTRY'
+CATALOG_KEY: str = 'NL_CATALOG'
+ENV_KEY: str = 'NL_ENV'
```
pradh, line 27: A one-line comment on what each is?

nl_server/registry.py
```diff
@@ -33,57 +34,59 @@
 
 
 #
-# A map from specific index to embeddings stores and models.
+# A class to hold embeddings stores and models.
```
pradh, line 37: Is there a more precise definition, and distinguish it from Env and Catalog?

nl_server/config.py
```diff
@@ -48,54 +46,55 @@ class ModelUsage(str, Enum):
...
+  location: str = None
+  index_endpoint_root: str = None
+  index_endpoint: str = None
+  index_id: str = None
 
 
-# Defines one embeddings index config.
-@dataclass
-class EmbeddingsConfig:
+# This represents the full catalog of models and indexes.
```
pradh, line 94: ... a subset of them may not be enabled by the Env.

nl_server/routes.py
```diff
@@ -47,13 +48,12 @@ def healthz():
   if default_embeddings_loaded:
     return 'OK', 200
 
-  default_index_type = current_app.config[
-      config.EMBEDDINGS_SPEC_KEY].default_index
+  default_index_type = current_app.config[config.ENV_KEY].default_indexes[0]
```
pradh, line 51: [0] may not be safe?  LHS has a check before access.

pradh (approved):
Beautiful change!  Thank you for breaking it up so that it is easy to review!

deploy/helm_charts/envs/worldbank.yaml
shifucun: I intended to only keep medium_ft for these instances. For world bank, do you think we should keep sdg and undata?

nl_server/config.py
shifucun: Done

nl_server/config.py
shifucun: Done

nl_server/loader.py
shifucun: Done

nl_server/registry.py
shifucun: Updated

nl_server/routes.py
shifucun: good point, updated the check.

shifucun:
thanks for the review!

deploy/helm_charts/envs/worldbank.yaml
pradh: Since worldbank is part of un, maybe keep it there?

shifucun: ok

nl_server/routes.py
```diff
@@ -48,10 +48,12 @@ def healthz():
   if default_embeddings_loaded:
     return 'OK', 200
 
-  default_index_type = current_app.config[config.ENV_KEY].default_indexes[0]
-  if not default_index_type:
-    logging.warning('Health Check Failed: Default index name empty!')
+  nl_env = current_app.config[config.ENV_KEY]
+  if not nl_env.default_indexes:
+    logging.error('Health Check Failed: Default index name empty!')
```
pradh, line 53: There are warnings below, so change them all or not?

(original rational for warning was I thought this can happen while booting up, i think i saw this crash in custom dc - https://github.com/datacommonsorg/website/pull/4240)

pradh (approved):

nl_server/routes.py
shifucun: Oh. Let me update here in next PR. I hope we don't need to make a real prediction call in /healthz then all logic here can be simplified.

pradh: ah, makes sense!

tools/nl/embeddings/build_custom_dc_embeddings.py
```diff
@@ -142,9 +143,9 @@ def generate_embeddings_yaml(model_info: utils.ModelConfig,
       "version": 1,
       "indexes": {
           "custom_ft": {
-              "embeddings": embeddings_csv_handler.abspath(),
+              "embeddings_path": embeddings_csv_handler.abspath(),
```
keyurva, line 146: It'll be cool if at some point, changes to the yaml format don't require changes to the custom embeddings tool.

keyurva (approved):
Always a fan of using good names - thanks!

tools/nl/embeddings/build_custom_dc_embeddings.py
shifucun: I think update the generated yaml file should do that, so running code that generate v10 yaml file with <name>_v10.yaml. and the server would look for _v10.yaml and not read _v9.yaml (which can not parse).

PR #4264, opened 2024-05-22
Fix bug in complex query detection

pradh:
This fixes the LLM fallback regression with "What are the sources of greenhouse gas emissions in the US" as a result of some score shift in multi-var match with the new model.

Please see regression test for more details.


shifucun (approved):
very cool!

PR #4274, opened 2024-05-24
Clean up gke deployment config and process

shifucun:
There are two gke yaml files for each env but the one-time setup one is mostly redundant. The region and number nodes can be later on updated and not accurately reflected here. store related fields are all constant too.

This PR moves ip and domain to the helm yaml file so there is one place to refer.

Also enforces the use of region in all the commands. In the future, when we would like to support zone, it's easier to add.


dwnoble (approved):
Nice! Collapsing the config into a single yaml file is a huge help

shifucun:
thanks!

PR #4280, opened 2024-05-29
Setup the bare bone of gemma eval tool

shifucun:


server/routes/nl/html.py
```diff
@@ -37,3 +37,10 @@ def eval_page():
     return render_template('/eval_embeddings.html',
                            server_config=json.dumps(server_config),
                            eval_golden=json.dumps(json.load(f)))
+
+
+@bp.route('/eval/gemma')
```
pradh, line 42: nit: do we want to call out gemma, or just say `rig` ?

because, the eval is for rig, but it could be used with any model

pradh (approved):
Nice!!

PR #4281, opened 2024-05-29
Shard NL integration tests to speed up PR test

shifucun:


run_test.sh
```diff
@@ -233,9 +233,15 @@ while [[ "$#" -gt 0 ]]; do
         shift 1
         ;;
     --explore)
-        echo --explore "### Running explore page integration tests"
-        run_integration_test explore_test.py
-        shift 1
+        if [[ -n "$2" && "$2" != -* ]]; then
```
pradh, line 236: what does `-*` mean? :)

run_test.sh
```diff
@@ -173,7 +173,7 @@ function run_integration_test {
   export TEST_MODE=test
   export ENABLE_EVAL_TOOL=false
 
-  python3 -m pytest -vv --reruns 2 server/integration_tests/$1
+  python3 -m pytest -vv --reruns 2 server/integration_tests/$1 -k $2
```
pradh, line 176: comment about `-k $2` ?

pradh (approved):
Very nice!!

run_test.sh
shifucun: Done

run_test.sh
shifucun: this is to check the argument is not like "-p" as an option (from llm..)

shifucun (commented):
thanks!

PR #4282, opened 2024-05-29
Add modal with links to stat vars for chart tiles

beets:
Adds a link to show metadata to each stat var tile, opening a modal with links for each to the stat var explorer.
Fetches the names of the stat var when the modal opens, using the stat var dcid as a fallback before the name is returned.

Fetches stat var names:
image

Updates with per capita:
image

Narrow screen, long string:
image

Scrolls:
image


static/js/tools/shared/tile_metadata_modal.tsx
```diff
@@ -0,0 +1,138 @@
...
+          </ModalHeader>
+          <div className="modal-subtitle">
+            Select a variable from the list to see its details. The links below
+            open the Statistical Variable Explorer in a new tab.
+          </div>
+          <ModalBody>
+            <div className="metadata-modal-links">
+              {statVarNames.length
+                ? statVarNames.map((dcidName, i) =>
+                    buildMetadataRow(dcidName[0], dcidName[1], i)
```
dwnoble, line 118: alternative: `<MetadataRow dcid={dcidName[0]} name={dcidName[1]} key={i} />`

static/js/tools/shared/tile_metadata_modal.tsx
```diff
@@ -0,0 +1,138 @@
...
+
+interface TileMetadataModalPropType {
+  statVarSpecs: StatVarSpec[];
+}
+
+type Dcid = string;
+type Name = string;
+type DcidNameTuple = [Dcid, Name];
+
+function buildMetadataRow(dcid: Dcid, name: Name, key: number): JSX.Element {
```
dwnoble, line 38: Why not make this a proper component `MetadataRow` and refer to it like `<MetadataRow dcid={...} name={...} />` below? I find components easier to read since they have named parameters. you could also drop "key" as an explicit parameter if it's a component

static/js/tools/shared/tile_metadata_modal.tsx
```diff
@@ -0,0 +1,138 @@
...
+import { StatVarSpec } from "../../shared/types";
+import { datacommonsClient } from "../../utils/datacommons_client";
+
+const SV_EXPLORER_REDIRECT_PREFIX = "/tools/statvar#sv=";
+
+interface TileMetadataModalPropType {
+  statVarSpecs: StatVarSpec[];
+}
+
+type Dcid = string;
```
dwnoble, line 34: Why not just define `dcid: string, name: string` below?

static/css/shared/_metadata_modal.scss
```diff
@@ -0,0 +1,105 @@
...
+    font-weight: 500;
+    line-height: 1.5;
+  }
+
+  .modal-subtitle {
+    color: var(--gm-3-sys-light-on-surface);
+    padding: 0 24px 12px;
+
+    /* body/medium */
+    font-family: "Google Sans Text";
```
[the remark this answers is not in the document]
beets: I needed an override for Google Sans Text (rather than the current default Google Sans). I'll update to $headings-font-family for now. Perhaps either of those should be the default sans-serif vs headings font as a follow-on update.

static/js/tools/shared/tile_metadata_modal.tsx
beets: thanks - good point. updated.

static/css/shared/_metadata_modal.scss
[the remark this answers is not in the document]
dwnoble: Good call-right now we have both "Google Sans", "Google Sans Text" in $font-family-sans-serif  - we should probably just have Google Sans Text in `$font-family-sans-serif` and Google Sans in `$headings-font-family`

static/js/tools/shared/tile_metadata_modal.tsx
beets: done!

beets (commented):
thanks for the comments! ptal

dwnoble:
looks like the modal isn't showing up right for the web components - i don't have a great system for testing this, so i resort to opening a local HTML file or jsfiddle like: https://jsfiddle.net/4xt5yp7u/

i think the fix is to set the `Modal`s container "container" prop to the web component's root element like https://github.com/datacommonsorg/website/blob/ee1dae1b7cb3652c31294edd262cb770161a4338/static/js/place/chart_embed.tsx#L409

dwnoble:
last thing-i found myself trying to close the modal by clicking outside the modal area (rather than the close button)- can you add a close hook for clicking outside the modal?

docs/developer_guide.md
```diff
@@ -128,6 +133,14 @@ to be brought up locally (in a separate process):
 
 By default the NL server runs on port 6060.
 
+If you run into problems starting the server, try running these commands before restarting the server:
+```bash
+./run_test.sh --setup_python
+rm -rf ~/.datacommons
+rm -rf /tmp/datcom-nl-models
+rm -rf /tmp/datcom-models-dev
```
juliawu, line 141: Sorry, I might have given you the wrong command before. I think this should be `rm -rf /tmp/datcom-nl-models-dev`

static/js/tools/shared/tile_metadata_modal.tsx
juliawu: Is there a missing commit that didn't get pushed? I don't think I see the update.

juliawu (commented):
Wonderful!

docs/developer_guide.md
beets: ah yes - thanks for catching that! fixed.

static/css/shared/_metadata_modal.scss
```diff
@@ -0,0 +1,105 @@
+/**
+ * Copyright 2023 Google LLC
```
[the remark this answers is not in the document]
beets: fixed

static/css/shared/_metadata_modal.scss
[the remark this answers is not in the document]
beets: sounds good - will do this in a follow up

static/js/tools/shared/tile_metadata_modal.tsx
beets: pushed -- rookie mistake :)

beets (commented):
thanks for the reviews and pointers to testing web components. lots more plumbing and tested the components on the playground. ptal

beets:
> last thing-i found myself trying to close the modal by clicking outside the modal area (rather than the close button)- can you add a close hook for clicking outside the modal?

fixed! needed to pass along a toggle function

beets:
> looks like the modal isn't showing up right for the web components - i don't have a great system for testing this, so i resort to opening a local HTML file or jsfiddle like: https://jsfiddle.net/4xt5yp7u/
> 
> i think the fix is to set the `Modal`s "container" prop to the web component's root element like
> 
> https://github.com/datacommonsorg/website/blob/ee1dae1b7cb3652c31294edd262cb770161a4338/static/js/place/chart_embed.tsx#L409

thanks for the pointer -- that did the trick!

juliawu (approved):
Very nice!

beets:
thanks for the reviews!

PR #4284, opened 2024-05-29
Fix bug in NL feedback

juliawu:
This PR fixes the issue where clicking the NL feedback icons in the charts no longer appears to do anything. This happens because the feedback box got hidden by the chart container for being "out of bounds". 

Before:
Screenshot 2024-05-29 at 4 45 27 PM

After:

Screenshot 2024-05-29 at 4 44 47 PM


dwnoble (approved):
Thanks for the quick fix!

PR #4286, opened 2024-05-30
[RIG EVAL] Show per question eval status/result and disable re-eval (for now)

shifucun:


static/js/apps/eval_rig/feedback_form.tsx
```diff
@@ -35,7 +34,7 @@ export interface EvalInfo {
   dcStat: string;
 }
 
-interface Response extends EvalInfo {
+export interface Response extends EvalInfo {
```
chejennifer, line 37: curious why exporting this interface because I don't see Response used in any other files?

static/js/apps/eval_rig/feedback_form.tsx
```diff
@@ -84,103 +92,84 @@ export function FeedbackForm(props: FeedbackFormProps): JSX.Element {
...
+              disabled={completed}
+            />
+          </div>
 
-        <button type="submit" disabled={Object.values(response).includes("")}>
-          Submit
-        </button>
-      </fieldset>
-    </form>
+          <button type="submit" disabled={Object.values(response).includes("")}>
```
chejennifer, line 168: nit: maybe only include submit button if not `completed`?

chejennifer (approved):
nice!

static/js/apps/eval_rig/feedback_form.tsx
shifucun: it's used in data_store for typedef

static/js/apps/eval_rig/feedback_form.tsx
shifucun: good idea!

shifucun (commented):
thanks for review!

PR #4288, opened 2024-05-30
Speed up NL cloudbuild even more

shifucun:
Previous test split reduces the time from ~10m to ~6m. This tries to split the 2 most long running test. The whole run should then be around 4-5min.

image


pradh (approved):
Verry nice!

PR #4289, opened 2024-05-31
[RIG EVAL] Save rater feedback to google sheet

shifucun:
Also:
- put google sheet client (doc) in context
- Remove try catch so errors should be handled by caller

### DIFF
--- a/static/js/apps/eval_rig/constants.tsx
+++ b/static/js/apps/eval_rig/constants.tsx
@@ -29,7 +29,9 @@ export const DC_RESPONSE_COL = "dc_response";
 export const LLM_STAT_COL = "llm_stat";
 export const DC_STAT_COL = "dc_stat";
 
-export const DC_QUESTION_FEEDBACK_COL = "dc_question_feedback";
-export const DC_RESPONSE_FEEDBACK_COL = "dc_response_feedback";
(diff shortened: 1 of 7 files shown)


static/js/apps/eval_rig/constants.tsx
```diff
@@ -27,3 +28,8 @@ export const DC_QUESTION_COL = "dc_question";
 export const DC_RESPONSE_COL = "dc_response";
 export const LLM_STAT_COL = "llm_stat";
 export const DC_STAT_COL = "dc_stat";
+
+export const DC_QUESTION_FEEDBACK_COL = "dc_question_feedback";
+export const DC_RESPONSE_FEEDBACK_COL = "dc_response_feedback";
+export const LLM_STAT_FEEDBACK_COL = "llm_stat_feedback";
+export const DC_STAT_FEEDBACK_COL = "dc_stat_feedback";
```
pradh, line 35: super nit: maybe construct it: `DC_QUESTION_COL + FB_SUFFIX`, etc.

pradh (approved):
Nice!

static/js/apps/eval_rig/constants.tsx
shifucun: good idea

PR #4291, opened 2024-05-31
[RIG EVAL] add eval list

chejennifer:
https://github.com/datacommonsorg/website/assets/69875368/e0974e8e-234e-4150-a6b4-7d20d2125039

### DIFF
--- a/static/js/apps/eval_rig/app.tsx
+++ b/static/js/apps/eval_rig/app.tsx
@@ -155,16 +155,14 @@ export function App(props: AppPropType): JSX.Element {
         )}
         {user && allQuery[selectedQuery] && allCall[selectedQuery] && (
           <>
-            <EvalList
-              sheetId={props.sheetId}
-              user={user}
-              queries={allQuery}
-              calls={allCall}
-              onQuerySelected={(q) => setSelectedQuery(q.id)}
-            />
             <AppContext.Provider
               value={{ doc, sheetId: props.sheetId, userEmail: user.email }}
             >
+              <EvalList
+                queries={allQuery}
+                calls={allCall}
+                onQuerySelected={(q) => setSelectedQuery(q.id)}
+              />
               <QuerySection
                 query={allQuery[selectedQuery]}
                 calls={allCall[selectedQuery]}
--- a/static/js/apps/eval_rig/app.tsx
+++ b/static/js/apps/eval_rig/app.tsx
@@ -28,6 +28,7 @@ import {
   USER_COL,
 } from "./constants";
 import { AppContext } from "./context";
+import { EvalList } from "./eval_list";
 import { DcCall } from "./eval_section";
 import { Query, QuerySection } from "./query_section";
 
@@ -43,6 +44,7 @@ export function App(props: AppPropType): JSX.Element {
   const [doc, setDoc] = useState<GoogleSpreadsheet>();
   const [allQuery, setAllQuery] = useState<Record<string, Query>>({});
   const [allCall, setAllCall] = useState<Record<string, DcCall[]>>({});
+  const [selectedQuery, setSelectedQuery] = useState("1");
 
   async function loadHeader(doc: GoogleSpreadsheet): Promise<HeaderInfo> {
     const result: HeaderInfo = {};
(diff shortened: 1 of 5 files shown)


static/js/apps/eval_rig/app.tsx
```diff
@@ -151,12 +153,24 @@ export function App(props: AppPropType): JSX.Element {
...
         )}
-        {user && allQuery["1"] && allCall["1"] && (
-          <AppContext.Provider
-            value={{ doc, sheetId: props.sheetId, userEmail: user.email }}
-          >
-            <QuerySection query={allQuery["1"]} calls={allCall["1"]} />
-          </AppContext.Provider>
+        {user && allQuery[selectedQuery] && allCall[selectedQuery] && (
+          <>
+            <EvalList
```
shifucun, line 158: put <EvalList> within <AppContext.Provider> as well so sheetId and user can be obtained from that.

static/js/apps/eval_rig/eval_list.tsx
```diff
@@ -0,0 +1,125 @@
...
+  calls: Record<string, DcCall[]>;
+  onQuerySelected: (query: Query) => void;
+}
+
+export function EvalList(props: EvalListPropType): JSX.Element {
+  const [userEvalsOnly, setUserEvalsOnly] = useState(false);
+  const [modalOpen, setModalOpen] = useState(false);
+  const [queryCompletionStatus, setQueryCompletionStatus] = useState({});
+
+  const userLdap = props.user.email.split("@")[0];
```
shifucun, line 40: i wonder if we should use full email as user id since we want this to be used externally or by 3P raters.

static/js/apps/eval_rig/eval_list.tsx
```diff
@@ -0,0 +1,125 @@
...
+
+  const openModal = () => {
+    setModalOpen(true);
+    const existPromises = orderedQueries.map((query) => {
+      return Promise.all(
+        props.calls[query.id].map((call) => {
+          return getCallExistence(props.sheetId, query.id, call.id);
+        })
+      );
+    });
```
shifucun, line 53: is there a way to count the collections under query.id in firestore? then can just do one query and check if the count is the same as the total count?

shifucun (commented):
very nice!!

static/js/apps/eval_rig/app.tsx
chejennifer: good idea, done

static/js/apps/eval_rig/eval_list.tsx
chejennifer: unfortunately not, I've tried that and I just get 0 ..

shifucun: I wonder if this would work https://firebase.google.com/docs/firestore/query-data/aggregation-queries#use_the_count_aggregation

shifucun (approved):

static/js/apps/eval_rig/data_store.tsx
```diff
@@ -91,3 +98,29 @@ export async function saveToSheet(
...
+    "sheets",
+    sheetId,
+    "queries",
+    queryId,
+    "calls",
+    callId,
+    "responses"
+  );
+  const snapshot = await getCountFromServer(collectionRef);
+  return !!snapshot.data().count;
```
shifucun, line 125: similar to this, but use path (db, "sheets", sheetId, "queries", queryId, "calls") would get the count of calls under a query?

chejennifer: updated

static/js/apps/eval_rig/data_store.tsx
```diff
@@ -37,13 +38,28 @@ import {
...
+  sheetId: string,
+  queryId: string,
+  callId: string
+): Promise<void> {
+  const docRef = doc(
+    collection(db, "sheets", sheetId, "queries", queryId, "calls"),
+    callId
+  );
+  // Need to set fields in a document in order to create the document
+  setDoc(docRef, { dummy: "dummy" });
```
shifucun, line 52: nit: maybe set it {id: callId} ?

shifucun (approved):
very nice!

static/js/apps/eval_rig/data_store.tsx
shifucun: If there is already a call entry with response, when we allow multiple evals and another response is added, would this still be okay?

i am not sure if setDoc is a pure reset (we don't want that), or just adding a new field (desired)

chejennifer: done

shifucun: reading through documentation, sub-collection and document are relatively independent, so this is ok!

PR #4295, opened 2024-06-03
[Place Page Revamp] Add a new development endpoint

juliawu:
Adds a new endpoint to hold the revamped place pages while in development.

In particular, changes include:
* Renaming _nl_search_bar.scss to without the leading underscore to allow `@use` importing
* A starter `.js` file that renders the nl search bar at the top of the page
* A starter `.html` template that renders the place name, type, and links to parent places
* A starter `.scss` file for styling
* A new /dev-place blueprint and endpoint in flask to hold changes
* Update the link generation helper functions to also use the new endpoint

An example:

```/dev-place/geoId/667000```
Screenshot 2024-06-03 at 12 28 09 PM


beets (commented):
lgtm in general. 

i'd lean towards making a new directory, rather than using things within /place. that will make it easier to switch over once dev is completed. it'll also help to avoid any unintended dependencies

juliawu:
> lgtm in general.
> 
> i'd lean towards making a new directory, rather than using things within /place. that will make it easier to switch over once dev is completed. it'll also help to avoid any unintended dependencies

Good idea, switched to using a new directory "/dev_place"

beets (approved):
thanks for the update!

PR #4299, opened 2024-06-04
Update variable count on homepage

juliawu:
Updates the stat var count on our homepage from 450k to 260k. The original 450k included some biomedical stat vars that are not yet accessible to the public.

Before:
image

After:
image

### DIFF
--- a/server/templates/static/homepage.html
+++ b/server/templates/static/homepage.html
@@ -74,7 +74,7 @@
       </article>
       <article>
         <p>
-          Over 240B data points across 450k variables from the
+          Over 240B data points across 260k variables from the
           <a href="https://www.who.int/" target="_blank">World Health
           Organization</a>, <a href="https://www.cdc.gov/index.htm"
           target="_blank">Centers for Disease Control and Prevention</a>, <a


kmoscoe (approved):

premr (approved):
Lgtm

server/templates/static/homepage.html
```diff
@@ -74,7 +74,7 @@ <h4>Data from</h4>
       </article>
       <article>
         <p>
-          Over 240B data points across 450k variables from the
+          Over 240B data points across 260k variables from the
```
pradh, line 77: nit: (can feel free to ignore or slide into another PR) `k` -> `K` for symmetry with `B` for billion?

pradh (commented):
Thanks for the change!

server/templates/static/homepage.html
kmoscoe: +1 :-)

juliawu: Great call out! Put up https://github.com/datacommonsorg/website/pull/4308

PR #4304, opened 2024-06-04
[RIG EVAL] update flow of feedback

chejennifer:
This updates the feedback flow so that first user gives feedback on the overall query answer & then goes into giving feedback for each call. 

In an attempt to keep components more coherent, this pr includes some refactoring. For the feedback section, we have the following components:
- QueryFeedback
- CallFeedback
- FeedbackNavigation

https://github.com/datacommonsorg/website/assets/69875368/5964edd5-0173-4e40-b6b4-e78e00e8223b


static/js/apps/eval_rig/data_store.tsx
```diff
@@ -34,10 +35,48 @@ import {
...
 
+// Gets the document path for a specific query or call
+export function getPath(
+  sheetId: string,
+  queryId?: number,
+  callId?: number
+): string[] {
+  const path = ["sheets", sheetId];
+  if (queryId) {
+    path.push(...["queries", String(queryId)]);
```
shifucun, line 51: in what condition queryId and/or callId is not set?

static/js/apps/eval_rig/data_store.tsx
```diff
@@ -34,10 +35,48 @@ import {
...
+  return path;
+}
+
+// Sets a field in a doc at the specified path
+export async function setField(
+  path: string[],
+  fieldKey: string,
+  fieldValue: string
+): Promise<void> {
+  const docRef = doc(db, path.shift(), ...path);
```
shifucun, line 65: why need to do path.shift() here?

static/js/apps/eval_rig/data_store.tsx
```diff
@@ -34,10 +35,48 @@ import {
...
+  setDoc(docRef, { [fieldKey]: fieldValue }, { merge: true });
+}
+
+// Gets a field from a doc at the specified path
+export async function getField(
+  path: string[],
+  fieldKey: string
+): Promise<string> {
+  const docRef = doc(db, path.shift(), ...path);
```
shifucun, line 74: maybe wrap the path conversion into a helper function?

static/js/apps/eval_rig/data_store.tsx
chejennifer: this is for getting the path to a document in firestore, so if we want to get the path to a query document, we wouldn't set a callId

static/js/apps/eval_rig/eval_list.tsx
```diff
@@ -39,17 +40,28 @@ export function EvalList(): JSX.Element {
...
-      .then((results) => {
+    Promise.all([callCountPromises, queryFeedbackPromises])
+      .then(([callCountResults, queryFeedbackResults]) => {
         orderedQueries.forEach((query, i) => {
-          // A query might not have any calls.
-          const calls = allCall[query.id] || {};
-          const completed = results[i] === Object.keys(calls).length;
-          queryCompletionStatus[query.id] = completed;
+          // If a query is marked as hallucination, it won't have any feedback
+          // on its calls and is considered completed.
```
shifucun, line 56: this is tricky!

shifucun (approved):
nice, the component separation is very good

static/js/apps/eval_rig/data_store.tsx
shifucun: should queryId always be set?

static/js/apps/eval_rig/data_store.tsx
chejennifer: ah, originally because looking at the doc() method, it takes a path and additional path segments. But updated getpath to just get the full path with slashes & now here, we just pass in path.

shifucun (approved):

PR #4307, opened 2024-06-04
Option to disable stop-words removal for monovars

pradh:
image

### DIFF
--- a/server/lib/nl/detection/types.py
+++ b/server/lib/nl/detection/types.py
@@ -441,5 +441,5 @@ class DetectionArgs:
   reranker: str
   # Enable entities/triples detection (e.g., for biomed DC)
   allow_triples: bool
-  # Remove stop-words for query detection
+  # Include stop-words for query detection
   include_stop_words: bool
--- a/server/lib/nl/detection/types.py
+++ b/server/lib/nl/detection/types.py
@@ -441,5 +441,5 @@ class DetectionArgs:
   reranker: str
   # Enable entities/triples detection (e.g., for biomed DC)
   allow_triples: bool
-  # Disable stop-word removal for query detection
-  disable_stop_words: bool
+  # Remove stop-words for query detection
+  include_stop_words: bool
--- a/server/lib/nl/detection/types.py
+++ b/server/lib/nl/detection/types.py
@@ -429,3 +429,17 @@ class Detection:
   classifications: List[NLClassifier]
   llm_resp: Dict = field(default_factory=dict)
   detector: ActualDetectorType = ActualDetectorType.HybridHeuristic
+
+
+@dataclass
+class DetectionArgs:
+  # Name of the embeddings index type to override ENV default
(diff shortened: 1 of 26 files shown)


server/lib/nl/detection/types.py
```diff
@@ -428,4 +428,18 @@ class Detection:
...
+  # Name of the embeddings index type to override ENV default
+  embeddings_index_type: str
+  # Query mode. e.g., `strict`, `toolformer`
+  mode: str
+  # Reranker model name
+  reranker: str
+  # Enable entities/triples detection (e.g., for biomed DC)
+  allow_triples: bool
+  # Disable stop-word removal for query detection
+  disable_stop_words: bool
```
shifucun, line 445: nit: include_stop_words? (currently should be: disable stop words removal)

shifucun (commented):
DetectionArgs makes code much simpler! Jus one naming comment.

server/lib/nl/detection/types.py
pradh: Done.

pradh:
Thanks for the review!

server/lib/nl/explore/params.py
```diff
@@ -42,6 +42,8 @@ class Params(str, Enum):
   #    Ex, if multiple verbs present, treat as action query and do not fulfill.
   MODE = 'mode'
   CLIENT = 'client'
+  # If set then we don't strip stop-words for mono-variable match.
+  DISABLE_STOP_WORDS = 'disableStopWords'
```
shifucun, line 46: also update the url param?

shifucun (approved):
thanks!

server/lib/nl/detection/variable.py
```diff
@@ -60,8 +57,11 @@ def detect_vars(orig_query: str,
   # the potential areas for improvement. For now, this removal blanket removes
   # any words in ALL_STOP_WORDS which includes contained_in places and their
   # plurals and any other query attribution/classification trigger words.
-  query_monovar = shared_utils.remove_stop_words(orig_query,
-                                                 query_util.ALL_STOP_WORDS)
+  if dargs.include_stop_words:
```
chejennifer, line 60: slightly confused, is this argument supposed to strip stop words or include stop words? because seems like it's stripping stop words here?

chejennifer (commented):
thanks!

server/lib/nl/detection/variable.py
pradh: Thanks, yes, when I renamed it per Bo's suggestion, I messed it up.

Fixed now, updated screenshot with new param.

server/lib/nl/explore/params.py
pradh: Done.

shifucun (approved):
nice! Now we can try something more interesting!

chejennifer (approved):
thanks for updating!

PR #4314, opened 2024-06-05
Try to fix webdriver tests flakiness

shifucun:
A few things changed here:

* Build a new webdriver docker image with newer chrome webdriver also new linux patches.
* Do not use cloud logging for webdriver tests.
* Remove the warmup requests since they are very limited, never updated and do no contribute much. They incur extra threading in flask.
* Add back skipped webdriver tests.

### DIFF
--- a/web_app.py
+++ b/web_app.py
@@ -11,47 +11,15 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-"""Main entry module specified in app.yaml.
-
-This module contains the request handler codes and the main app.
+"""Main entry for website Flask app.
 """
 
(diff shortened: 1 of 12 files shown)


web_app.py
```diff
@@ -11,48 +11,16 @@
...
-import threading
-import time
-
-import requests
 
 from server.__init__ import create_app
 
 app = create_app()
 
-WARM_UP_ENDPOINTS = [
```
chejennifer, line 30: curious why we're removing warm up endpoints?

chejennifer (approved):
thanks for the update! Just curious if the warm up endpoints were meant to be deleted

chejennifer:
> thanks for the update! Just curious if the warm up endpoints were meant to be deleted

Nevermind, I see the explanation in the PR description

PR #4315, opened 2024-06-05
Fix more stat var name typos

juliawu:
Fixes a few more typos in SV display name overrides.


chejennifer (approved):
thanks for the fixes!

PR #4316, opened 2024-06-05
[bard] nodejs changes

chejennifer:
- for nodejs/query, stop using strict mode
- pass heuristic detector & sv threshold as arguments
- make relatedQuestions out of related topics


static/nodejs_server/query.ts
```diff
@@ -365,6 +400,6 @@ export async function getQueryResult(
       total: getElapsedTime(startTime, endTime),
     },
   };
-
-  return { charts: processedResults, debug };
+  const relatedQuestions = getRelatedQuestions(relatedThings, place);
```
pradh, line 403: should we flag protect this?

static/src/server.ts
```diff
@@ -306,7 +304,8 @@ app.get("/nodejs/query", (req: Request, res: Response) => {
   const apikey = (req.query.apikey as string) || "";
   const urlRoot = `${protocol}://${host}`;
   const client = (req.query.client as string) || BARD_CLIENT_URL_PARAM;
-  const mode = (req.query.mode as string) || DEFAULT_NL_MODE;
+  const mode = (req.query.mode as string) || "";
+  const svThreshold = (req.query.svThreshold as string) || "";
```
pradh, line 308: varThreshold

static/nodejs_server/query.ts
```diff
@@ -248,13 +285,10 @@ export async function getQueryResult(
...
   try {
-    nlResp = await axios
-      // Set "mode=strict" to use heuristic detector, disable using default place,
-      // use a higher SV threshold and avoid multi-verb queries
-      .post(
-        `${apiRoot}/api/explore/detect-and-fulfill?q=${query}&mode=${mode}&client=${client}`,
-        {}
-      );
+    nlResp = await axios.post(
+      `${apiRoot}/api/explore/detect-and-fulfill?q=${query}&mode=${mode}&client=${client}&detector=${QUERY_DETECTOR}&svThreshold=${svThreshold}`,
```
pradh, line 289: What does client do in NL server?

static/nodejs_server/query.ts
chejennifer: Since it's returned as a separate field in the result, wondering what the reason for flag protecting it is?

pradh: This is just a prototype that we may change evolve to doing heavier-weight things, like using an LLM.  Having a flag set lets bard caller control that.  And if the flag is set, we don't need to do related-thing handling either.

static/nodejs_server/query.ts
chejennifer: hm, looking through the editors, looks like client was added in this pr: https://github.com/datacommonsorg/website/pull/3785 

tracing through the code though, I'm not sure it's still being used ..

static/src/server.ts
chejennifer: done

static/nodejs_server/query.ts
chejennifer: sure, can flag protect this for now, updated

static/nodejs_server/query.ts
pradh: oh lol, i added it!  Ah, yes, this goes into the logs.

static/nodejs_server/query.ts
```diff
@@ -248,13 +286,10 @@ export async function getQueryResult(
...
   try {
-    nlResp = await axios
-      // Set "mode=strict" to use heuristic detector, disable using default place,
-      // use a higher SV threshold and avoid multi-verb queries
-      .post(
-        `${apiRoot}/api/explore/detect-and-fulfill?q=${query}&mode=${mode}&client=${client}`,
-        {}
-      );
+    nlResp = await axios.post(
+      `${apiRoot}/api/explore/detect-and-fulfill?q=${query}&mode=${mode}&client=${client}&detector=${QUERY_DETECTOR}&varThreshold=${varThreshold}`,
```
pradh, line 290: TODO: do we want to control related-thing generation?

pradh (approved):
Thanks!

static/nodejs_server/query.ts
chejennifer: added todo

PR #4319, opened 2024-06-05
Several CSV export fixes

dwnoble:
* fixed issue where place page CSV export would always pull from datacommons.org instead of the local hostname
* fixed issue where CSV export would always pull from datacommons.org even when an apiRoot was specified
* fixed issue where CSV export on a bar chart web component with a specified parentPlace and childPlaceType would throw an error

## Place page
before 
place-before
after
place-after

## bar chart csv export
before
bar-chart-error

after
bar-chart-success

## setting apiRoot on web components:
before
api-root-error

after
api-root-success


packages/client/src/data_commons_client.ts
```diff
@@ -78,7 +78,7 @@ class DataCommonsClient {
       ? p.apiRoot.replace(/\/$/, "")
       : "https://datacommons.org";
     this.webClient = new DataCommonsWebClient({
-      apiRoot: this.apiRoot,
+      apiRoot: p.apiRoot,
```
beets, line 81: hm, i don't follow this change. isn't this equivalent to the current logic, unless i'm missing something subtle about the "this" object in a constructor? if that's the case, could we do something like this:

```
const apiRoot = p.apiRoot ? ...
this.apiRoot = apiRoot
this.webClient = new DataCommonsWebClient({
  apiRoot: apiRoot
})
```

beets (approved):
thanks for the fix!

packages/client/src/data_commons_client.ts
dwnoble: i added a comment to help clarify, as this line is what was causing the place page bug.

backing up, I want a way to both specify a specific hostname for the website api root, or to use whatever the current hostname is. this way a web component user could point at a particular web api root, but the *.datacommons.org website itself ( or custom DCs) can just default to the current hostname.

I'm not totally in love with this solution, but the current behavior is: 

* `apiRoot parameter = "" or undefined` -> sets internal apiRoot to `"https://datacommons.org"`
* `apiRoot parameter = "/"` -> sets internal apiRoot to "", which will use the current hostname
* `apiRoot parameter = anything else (e.g., "https://dev.datacommons.org")  -> internal hostname = param hostname

so, i when initializing the `DataCommonsClient` with apiRoot = `"/"`, it will set it's internal apiRoot to `""`, but i want to initialize its member `DataCommonsWebClient` with the apiRoot = "/" that was originally passed in (lest it get set to the default datacommons.org value).

Happy to entertain other solutions, as this seems pretty nuanced.

packages/client/src/utils.ts
```diff
@@ -310,3 +312,20 @@ export function isDateInRange(
...
+ * DataCommonsWebClient
+ *
+ * - Defaults to "https://datacommons.org" if apiRoot is undefined or ""
+ * - If apiRoot is "/", returns an empty string that indicates we should use the
+ * current hostname as the website API root.
+ * - Otherwise, uses the website api root specified with trailing slashes
+ * stripped
+ *
+ * @param apiRoot Data Commons Website API root
+ * @returns
```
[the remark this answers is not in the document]
dwnoble: done

juliawu (approved):

packages/client/src/data_commons_client.ts
beets: The comments and helper makes it clearer, thanks for the update and explanation!

beets (approved):

PR #4322, opened 2024-06-06
[rig eval] update order of options

chejennifer:
In the feedback. update order of options so that it goes negative to positive.

Screenshot 2024-06-06 at 9 17 10 AM

Also, small change to the css for keeping the navigation buttons at the bottom of the page


static/css/eval_rig.scss
```diff
@@ -152,7 +153,6 @@ label {
   }
 
   .feedback-nav-section {
-    margin-top: auto;
```
shifucun, line 155: I added this to fix the position of buttons.

chejennifer: yes, the flex-grow I added accomplishes the same thing, but keeps the border above the buttons at the same height, otherwise it keeps moving when we switch pages

shifucun: I see, i missed the flex-grow change.

shifucun (approved):

PR #4326, opened 2024-06-07
Added lazy loading for map component in explore search results page

dwnoble:
Improves load time of queries like https://datacommons.org/explore/#q=Show+me+unemployment+rate+vs+white+population+in+US+counties


shifucun (approved):

pradh:
very nice!

PR #4330, opened 2024-06-08
Update bard deployment configs

shifucun:


deploy/helm_charts/dc_website/templates/single_cluster.yaml
```diff
@@ -135,7 +135,7 @@ metadata:
     {{- if .Values.ingress.internal }}
     ingress.gcp.kubernetes.io/pre-shared-cert: "website-ilb"
     kubernetes.io/ingress.regional-static-ip-name: "website-ilb-ip"
-    kubernetes.io/spec.ingressClassName: "gce-internal"
+    kubernetes.io/ingress.class: "gce-internal"
```
shifucun, line 138: this was a bad change from me that resulted in the load balancer becoming external. that might cause the nil pointer error we saw initially.

dwnoble (approved):

PR #4331, opened 2024-06-09
NL Model usage update: 1)custom DC dockerfile download model using gsutil; 2) model map object consolidation

shifucun:
Instead of having a download mode in build_custom_dc_embeddings tool, download the model and embeddings directly in the docker image.

* Since we won't upgrade the sentence transformer based model, can just hardcode the only version in dockerfile
* Updated the base image to include gcloud

This is a step towards consolidating the build embeddings tool.

Part of this change consolidates the embedding and reranking model object in registry into one map.


build/web_compose/Dockerfile
```diff
@@ -118,12 +118,12 @@ COPY import/. /workspace/import/
 COPY tools/nl/embeddings/. /workspace/tools/nl/embeddings/
 
 # Download model and embeddings
-ENV IS_CUSTOM_DC=true
-RUN python -m tools.nl.embeddings.build_custom_dc_embeddings --mode=download
+RUN mkdir -p /tmp/datcom-nl-models
+RUN gsutil cp -R gs://datcom-nl-models/ft_final_v20230717230459.all-MiniLM-L6-v2/ /tmp/datcom-nl-models/
+RUN gsutil cp gs://datcom-nl-models/embeddings_medium_2024_05_09_18_01_32.ft_final_v20230717230459.all-MiniLM-L6-v2.csv /tmp/datcom-nl-models/
```
chejennifer, line 123: since embeddings version can change, should we be reading this from embeddings.yaml instead?

nl_server/registry.py
```diff
@@ -82,16 +93,7 @@ def _load_models(self, models: dict[str, ModelConfig]):
...
-          if model_config.usage == ModelUsage.EMBEDDINGS:
-            model = VertexAIEmbeddingsModel(model_config)
-            self.name_to_emb_model[model_name] = model
-          elif model_config.usage == ModelUsage.RERANKING:
-            model = VertexAIRerankingModel(model_config)
-            self.name_to_rank_model[model_name] = model
-        elif model_config.type == ModelType.LOCAL:
-          model = LocalSentenceTransformerModel(model_config)
-          self.name_to_emb_model[model_name] = model
+        self.name_to_emb_model[model_name] = create_model(model_config)
```
chejennifer, line 96: if `model_config.usage == ModelUsage.RERANKING`, model needs to be added to `self.name_to_rank_model` instead

chejennifer (commented):
nice!

nl_server/registry.py
shifucun: nice catch. updated to keep only one map now.

nl_server/registry.py
```diff
@@ -53,14 +63,14 @@ def __init__(self, server_config: ServerConfig):
...
   def get_attribute_model(self) -> AttributeModel:
     return self._attribute_model
 
-  def get_embedding_model(self, model_name: str) -> EmbeddingsModel:
-    return self.name_to_emb_model.get(model_name)
+  def get_model(
+      self,
+      model_name: str) -> tuple[(EmbeddingsModel | RerankingModel, ModelUsage)]:
+    return (self.name_to_model.get(model_name),
+            self._server_config.models.get(model_name).usage)
```
pradh, line 73: Hmm, wonder why not keep the original API, but throw exception if get_reranking_model matches an embedding model and vice versa?

shifucun: yea, that's better.

nl_server/routes.py
```diff
@@ -24,6 +24,7 @@
 
 from nl_server import registry
 from nl_server import search
+from nl_server.config import ModelUsage
```
chejennifer, line 27: don't need this anymore?

chejennifer (approved):
nice!

nl_server/routes.py
shifucun: removed

pradh (approved):

PR #4334, opened 2024-06-11
Add TOOLFORMER_RAG mode

pradh:
Add a `mode=toolformer_rag` where we do not limit the number of places in time-delta timelines and ranking tables.

Also in toolformer modes, avoid return multiple charts in `basic`, and only return what the query asks for.


server/lib/nl/common/constants.py
```diff
@@ -842,6 +842,9 @@
 # Max number of answer places
 MAX_ANSWER_PLACES = 10
 
+# This is used in TOOLFORMER_TABLE mode.  Set to allow "counties in US".
+ABSOLUTE_MAX_PLACES_FOR_TABLES = 3250
```
shifucun, line 846: why not making this a bit larger to allow any new county?

server/lib/nl/detection/types.py
```diff
@@ -435,7 +435,7 @@ class Detection:
 class DetectionArgs:
   # Name of the embeddings index type to override ENV default
   embeddings_index_type: str
-  # Query mode. e.g., `strict`, `toolformer`
+  # Query mode. e.g., `strict`, `toolformer_point`
```
shifucun, line 438: what is "point"?

server/lib/nl/explore/params.py
```diff
@@ -70,9 +70,15 @@ class DCNames(str, Enum):
...
   # NOTE: This mode is incompatible with LLM detector
   STRICT = 'strict'
-  # This is a special mode to be used for toolformer experiments.
-  # This mode does not detect topics and has a sv score threshold of 0.8.
-  TOOLFORMER = 'toolformer'
+  # These are special modes used for toolformer experiments.
+
+  # The point mode returns exact variable's values and used by RIG.
+  # It does not detect topics and has a higher sv score threshold of 0.8.
+  TOOLFORMER_POINT = 'toolformer_point'
```
shifucun, line 77: given these two mode are not only diff by point/table, should they be called TOOLFORMER_RIG and TOOLFORMER_RAG

server/lib/nl/fulfillment/basic.py
```diff
@@ -63,10 +63,13 @@ def populate(state: PopulateState, chart_vars: ChartVars, places: List[Place],
                             [p.dcid for p in places])
     return False
 
-  if chart_vars.source_topic != PROJECTED_TEMP_TOPIC:
+  if (chart_vars.source_topic != PROJECTED_TEMP_TOPIC and
+      not params.is_toolformer_mode(state.uttr.mode)):
```
shifucun, line 67: this is very subtle. For easier tracking, can they be consolidated? Or some comment about PROJECTED_TEMP_TOPIC?

I feel it would be nice to have one file with ALL the condition checks, so can find the logic flow in one pass. Maybe something for later.

server/lib/nl/fulfillment/utils.py
```diff
@@ -270,3 +269,11 @@ def get_places_as_string(places: List[str]) -> str:
     return places[0]
   else:
     return ', '.join(places[0:len(places) - 1]) + f' and {places[-1]}'
+
+
+def get_max_ans_places(places: List[Place], uttr: Utterance) -> List[Place]:
```
shifucun, line 274: For later, I wonder if we can simplify logic branching based on some first order flags, and have them in one giant file or function...

That function / object itself can even be plotted to reason about the decision tree...

shifucun (approved):

shifucun:
I feel we need a DecisionTree class to hold the reasoning flow and by looking at that, can decide on what happened and why. But def something for later.

server/lib/nl/detection/types.py
chejennifer, line 438: also add toolformer_table to this list?

server/lib/nl/explore/params.py
chejennifer: +1

chejennifer (approved):
nice!

server/lib/nl/common/constants.py
pradh: I think the real number is more like 3143 or something? :)

server/lib/nl/detection/types.py
pradh: renaming to rig.

pradh: it was just e.g, but sure

server/lib/nl/explore/params.py
pradh: good idea

server/lib/nl/fulfillment/basic.py
pradh: Added some comments.

I'm not sure how to do that, considering there are different fulfillers for different scenarios and fallback across them.  But as a general point, yes, this could use simplification!

server/lib/nl/fulfillment/utils.py
pradh: Yeah, I think we can collect all the flags in one place.  I wonder if the decision tree might need to be built by propagating state in a context rather than putting all the decision tree code in one place (it is very hard to navigate that and would make things not very modular).  But yes, def something to improve on.

pradh (commented):
Thanks for the reviews!

PR #4335, opened 2024-06-11
Allow per question feedback even the whole question is hallucination

shifucun:


pradh (approved):

chejennifer (commented):
if allowing per question feedback when whole question is hallucination, do we still want to mark the question as completed when the question is marked as hallucination? If not, update here: https://github.com/datacommonsorg/website/blob/17d68b742e5b4cd26adc08d57020f02baee268bd/static/js/apps/eval_rig/eval_list.tsx#L55-L59

shifucun:
> if allowing per question feedback when whole question is hallucination, do we still want to mark the question as completed when the question is marked as hallucination? If not, update here:
> 
> https://github.com/datacommonsorg/website/blob/17d68b742e5b4cd26adc08d57020f02baee268bd/static/js/apps/eval_rig/eval_list.tsx#L55-L59

I see. Removed that check now.

chejennifer (approved):
thanks

PR #4339, opened 2024-06-11
Sv differ fixes

chejennifer:
- update to_vars to use a non-deprecated endpoint
- add nodejs queries that can be used by sv differ tool
- update reading catalog for sv differ

### DIFF
--- a/nl_server/config_reader.py
+++ b/nl_server/config_reader.py
@@ -98,7 +98,7 @@ def merge_vertex_ai_configs(m: ModelConfig,
   return VertexAIModelConfig(**merged)
 
 
-def read_catalog() -> Catalog:
+def read_catalog(catalog_dict: Dict = None) -> Catalog:
   """
   Reads the catalog from the config files and merges them together.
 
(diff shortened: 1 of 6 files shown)


nl_server/config_reader.py
```diff
@@ -98,7 +98,7 @@ def merge_vertex_ai_configs(m: ModelConfig,
   return VertexAIModelConfig(**merged)
 
 
-def read_catalog() -> Catalog:
+def read_catalog(catalog_dict: Dict = None) -> Catalog:
```
shifucun, line 101: I like this change and we should eventually make this 

`read_catalog(catalog_paths: [] = DEFAULT_PATHS, catalog_dict: Dict = None)`

Right now the custom DC path prevents the path to be constant, and really that should be passed in as the catalog_dict string value. 

Can you add a TODO for this to remind me of the change.

shifucun (approved):
nice update!

nl_server/config_reader.py
chejennifer: added TODO

chejennifer:
thanks for the review!

PR #4340, opened 2024-06-12
add base_uae_mem to bard yaml config of enabled indexes

chejennifer:


pradh (approved):
Did it fail on the instance?  We should have a unit test for env `.yamls`?

chejennifer:
> Did it fail on the instance? We should have a unit test for env `.yamls`?

it failed when I tried to deploy. But yes, unit test for yamls will be good!

PR #4342, opened 2024-06-12
Consolidate embeddings build by distilling custom DC treatment as a standalone step

shifucun:
Handle GCS and local files by a FileManager class and this can be applied in base DC as well.

All custom DC process is handled by one function `save_custom_dc_artifacts`. This way, consolidating the two build.py files will be easily achieved later.


tools/nl/embeddings/build_custom_dc_embeddings.py
```diff
@@ -14,117 +14,140 @@
 """Build embeddings for custom DCs."""
 
 from dataclasses import asdict
+import os
```
pradh, line 17: Is there any unit test to validate this (esp in Keyur's absence I'm not sure of this flow)?

pradh (approved):
Nice simplification!

tools/nl/embeddings/build_custom_dc_embeddings.py
shifucun: I have manually tested this in local and docker env.

PR #4343, opened 2024-06-12
Log and display NL index types as a list

shifucun:


pradh (approved):
Thanks for the change!

PR #4345, opened 2024-06-12
Update data fetch for line charts to POST instead of GET

chejennifer:
The number of entities we are getting data for can be quite large for TOOLFORMER_RAG mode & makes the url too long for a GET request, so use POST instead


server/routes/shared_api/observation/series.py
```diff
@@ -42,17 +42,23 @@
 bp = Blueprint("series", __name__, url_prefix='/api/observations/series')
 
 
-@bp.route('', strict_slashes=False)
+@bp.route('', strict_slashes=False, methods=['GET', 'POST'])
```
shifucun, line 45: use `request.method == 'POST'` to check method and branch logic?

chejennifer: done

server/routes/shared_api/observation/series.py
```diff
@@ -42,17 +42,24 @@
...
 def series():
   """Handler to get preferred time series given multiple stat vars and entities."""
-  entities = list(filter(lambda x: x != "", request.args.getlist('entities')))
-  variables = list(filter(lambda x: x != "", request.args.getlist('variables')))
+  if request.method == 'POST':
+    entities = request.json.get('entities')
+    variables = request.json.get('variables')
+    facet_ids = request.json.get('facetIds')
+  else:
+    entities = list(filter(lambda x: x != "", request.args.getlist('entities')))
```
shifucun, line 54: nit: make this long expression a function..

shifucun (approved):

server/routes/shared_api/observation/series.py
chejennifer: done

server/routes/shared_api/observation/series.py
```diff
@@ -42,26 +43,36 @@
 bp = Blueprint("series", __name__, url_prefix='/api/observations/series')
 
 
-@bp.route('', strict_slashes=False)
+# Filters a list for non empty values
+def _get_filtered_arg_list(arg_list: List[str]) -> List[str]:
+  return list(filter(lambda x: x != "", arg_list))
```
shifucun, line 48: maybe for future PR: you can use `request` in this function directly and this function can just pass in the arg name?

chejennifer: added todo

pradh: +1.  also use constants for field-names when we do that

pradh (approved):
Nice!

To confirm, The "how has life expectancy changed over time in countries of the world" query works locally now?

chejennifer:
> Nice!
> 
> To confirm, The "how has life expectancy changed over time in countries of the world" query works locally now?

Yes, works locally now!

PR #4347, opened 2024-06-12
Fixed intersection-observer build issue

dwnoble:
Fixes autopush build failure by moving intersection-observer from a dev dependency to a regular dependency

(see: https://github.com/datacommonsorg/website/blob/912b2b881b8bbb9d8b043c89924256d3c67e8aa1/build/web_server/Dockerfile#L42)


chejennifer (approved):
thanks for fixing!

PR #4348, opened 2024-06-12
[Build Embeddings] Save index config, md5 and add test to check integrity

shifucun:
- Removed old preindex/duplicate files and keep all input files simply under 'input' directory.
- Re-built all indexes to generate the index_config.yaml and md5sum.txt in the embeddings output gcs folder.


tools/nl/embeddings/integrity_test.py
```diff
@@ -0,0 +1,49 @@
...
+        embeddings_dir = index_config.embeddings_path
+      else:
+        raise ValueError('Unsupported store type: %s' % index_config.store_type)
+      md5sum_path = os.path.join(embeddings_dir, 'md5sum.txt')
+      if gcs.is_gcs_path(md5sum_path):
+        md5sum_path = gcs.maybe_download(md5sum_path)
+      with open(md5sum_path) as f:
+        got_md5sum = f.read().strip()
+      with open(os.path.join(index_config.source_path, '_preindex.csv')) as f:
+        expected_md5sum = hashlib.md5(f.read().encode('utf-8')).hexdigest()
```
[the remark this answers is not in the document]
shifucun: good idea. fixed.

deploy/nl/catalog.yaml
```diff
@@ -59,52 +59,52 @@ indexes:
     healthcheck_query: "Life expectancy"
   sdg_ft:
     store_type: MEMORY
-    source_path: ../../tools/nl/embeddings/data/curated_input/sdg
-    embeddings_path: gs://datcom-nl-models/sdg_ft-final-v20230717230459-all-MiniLM-L6-v2_2024_06_20_23_15_52/embeddings.csv
+    source_path: ../../tools/nl/embeddings/input/sdg
+    embeddings_path: gs://datcom-nl-models/sdg_ft_2024_06_24_23_45_46/embeddings.csv
```
pradh, line 63: Very nice!

tools/nl/embeddings/integrity_test.py
```diff
@@ -0,0 +1,49 @@
...
+
+from nl_server import config_reader
+from shared.lib import gcs
+from tools.nl.embeddings import utils
+
+
+class TestIntegrity(unittest.TestCase):
+  """Check that the index in catalog.yaml is up to date.
+
+  Read the md5sum of each index index and check it against the _preindex.csv in
```
pradh, line 26: 1 index

tools/nl/embeddings/integrity_test.py
```diff
@@ -0,0 +1,49 @@
...
+      else:
+        raise ValueError('Unsupported store type: %s' % index_config.store_type)
+      md5sum_path = os.path.join(embeddings_dir, 'md5sum.txt')
+      if gcs.is_gcs_path(md5sum_path):
+        md5sum_path = gcs.maybe_download(md5sum_path)
+      with open(md5sum_path) as f:
+        got_md5sum = f.read().strip()
+      expected_md5sum = utils.get_md5sum(
+          os.path.join(index_config.source_path, '_preindex.csv'))
+      self.assertEqual(got_md5sum, expected_md5sum)
```
pradh, line 49: What use-case do you anticipate for md5 hash?  Sanity-check against someone checking in a wrong path?  Or actual content shift of some form?

If the former, this LG.  But if latter, I slightly wonder if we should just put the whole _prefix.csv and compare, so the diffs can make sense?

pradh (approved):
Nice!

tools/nl/embeddings/integrity_test.py
shifucun: How I start this is finding that different indexes using the same descriptions might not be updated at the same time.
For example, after changing base descriptions, the lancedb and mistral index are not updated. 
This was a problem (but not anymore) when undata and ilo also relies on base/sdg description.

Right now, can make sure the _preindex of a folder is indeed what is in the actual embeddings.csv for all the indexes. And this is necessary for increment update..

tools/nl/embeddings/integrity_test.py
shifucun: fixed

tools/nl/embeddings/integrity_test.py
shifucun: checking in the _preindex.csv also works but i am a bit worried that the gcs download of many ( potentially large) files can make the test heavy.

PR #4351, opened 2024-06-13
[NL param] Use a list of index types everywhere

shifucun:
NL server now allows a list of indexes. With this change, we don't need to merge index during embedding build time, but rather include them in the runtime.

This allows not replicate the same embeddings in multiple indexes and makes embedding build atomic for each source folder.

### DIFF
--- a/server/lib/nl/detection/llm_detector.py
+++ b/server/lib/nl/detection/llm_detector.py
@@ -149,7 +149,7 @@ def detect(query: str, prev_utterance: utterance.Utterance,
 
   query_detection_debug_logs["llm_response"] = llm_resp
   query_detection_debug_logs["query_transformations"] = {
-      "sv_detection_query_index_types": [dargs.embeddings_index_types]
+      "sv_detection_query_index_types": dargs.embeddings_index_types
   }
 
   # SV Detection.
--- a/server/lib/nl/detection/llm_detector.py
+++ b/server/lib/nl/detection/llm_detector.py
@@ -149,7 +149,7 @@ def detect(query: str, prev_utterance: utterance.Utterance,
 
   query_detection_debug_logs["llm_response"] = llm_resp
   query_detection_debug_logs["query_transformations"] = {
-      "sv_detection_query_index_types": [dargs.embeddings_index_type]
+      "sv_detection_query_index_types": [dargs.embeddings_index_types]
   }
 
   # SV Detection.
--- a/server/lib/nl/explore/params.py
+++ b/server/lib/nl/explore/params.py
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 from enum import Enum
-from typing import Dict
+from typing import Dict, List
 
 from server.lib.nl.detection.types import DetectionArgs
 from shared.lib import constants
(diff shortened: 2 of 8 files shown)


server/lib/nl/detection/llm_detector.py
```diff
@@ -149,7 +149,7 @@ def detect(query: str, prev_utterance: utterance.Utterance,
 
   query_detection_debug_logs["llm_response"] = llm_resp
   query_detection_debug_logs["query_transformations"] = {
-      "sv_detection_query_index_types": [dargs.embeddings_index_type]
+      "sv_detection_query_index_types": [dargs.embeddings_index_types]
```
pradh, line 152: do we need a list in a list?

server/lib/nl/explore/params.py
```diff
@@ -124,18 +124,18 @@ def is_bio(insight_ctx: Dict) -> bool:
   return insight_ctx.get(Params.DC.value) == DCNames.BIO_DC.value
 
 
-def dc_to_embedding_type(dc: str, embeddings_type: str) -> str:
+def dc_to_embedding_types(dc: str, embeddings_types: List[str]) -> str:
```
pradh, line 127: return `List[str]`?

server/routes/explore/helpers.py
```diff
@@ -96,7 +96,8 @@ def parse_query_and_detect(request: Dict, backend: str, client: str,
   i18n = i18n_str and i18n_str.lower() == 'true'
 
   # Index-type default is in nl_server.
-  embeddings_index_type = request.args.get(params.Params.INDEX.value, '')
+  idx_param_str = request.args.get(params.Params.INDEX.value, '')
+  embeddings_index_types = idx_param_str.split(',')
```
pradh, line 100: fwiw, maybe do a strip() on the split values?

server/services/datacommons.py
```diff
@@ -336,9 +336,13 @@ def resolve(nodes, prop):
...
 
 
-def nl_search_vars(queries, index_type, skip_topics=False, reranker=''):
+def nl_search_vars(queries,
+                   index_types: List[str],
+                   skip_topics=False,
+                   reranker=''):
   """Search sv from NL server."""
-  url = f'{current_app.config["NL_ROOT"]}/api/search_vars?idx={index_type}'
+  idx_params = ','.join(index_types)
```
pradh, line 344: Oh I see, so https://github.com/datacommonsorg/website/blob/8d67e1c1592d36cc796c93ed67b75abbbed2e37d/nl_server/routes.py#L92 is already supported.

pradh (commented):
Very nice!

Should we have an integration test with 2 `idx=` entries and do a correlation?

pradh (approved):
Approving in case I don't get to it in the evening.

dwnoble (approved):
Nice!

server/routes/nl/api.py
```diff
@@ -42,9 +42,9 @@ def search_vector():
...
   query = request.args.get('query')
-  index = request.args.get('index')
+  idx = request.args.get('idx')
   if not query:
     flask.abort(400, 'Must provde a `query`')
-  if not index:
-    flask.abort(400, 'Must provde an `index`')
-  return dc.nl_search_vars([query], index)
+  if not idx:
+    flask.abort(400, 'Must provde an `idx`')
```
chejennifer, line 49: nit: provide

also in line 47

server/routes/nl/api.py
```diff
@@ -42,9 +42,9 @@ def search_vector():
...
+  return dc.nl_search_vars([query], idx.split(','))
```
chejennifer, line 50: nit: I wonder if instead of using ',' as a delimiter, could we do args.getList('idx') & pass the idx as a list argument

server/routes/explore/helpers.py
```diff
@@ -96,7 +96,8 @@ def parse_query_and_detect(request: Dict, backend: str, client: str,
   i18n = i18n_str and i18n_str.lower() == 'true'
 
   # Index-type default is in nl_server.
-  embeddings_index_type = request.args.get(params.Params.INDEX.value, '')
+  idx_param_str = request.args.get(params.Params.INDEX.value, '')
```
chejennifer, line 99: nit: should we do request.args.getList & have the clients pass the values as a list

chejennifer (approved):
nice!

shifucun (commented):
thanks for the review!

server/routes/nl/api.py
shifucun: In this case i think it's doable, but as we use "," delim in the other places, might just use the same convention here.

shifucun:
> Very nice!
> 
> Should we have an integration test with 2 `idx=` entries and do a correlation?

I will add one after cleaning up all current indexes, since they mostly include some kind of base index

PR #4352, opened 2024-06-14
[NL Embeddings] Modularize indexes by input folders; Merge curated input and alternative input

shifucun:
Since all the sv description csv have the same columns (dcid, sentence), the alternative csvs can be moved to the primary folder now.

For existing embeddings indexes that are union of several folders, we can use the union from runtime by specifying multiple default_indexes or use more than one indexes when `dc=` param is set.


deploy/nl/catalog.yaml
```diff
@@ -58,27 +58,22 @@ indexes:
...
-    embeddings_path: gs://datcom-nl-models/embeddings_undata_2024_03_20_11_01_12.ft_final_v20230717230459.all-MiniLM-L6-v2.csv
+    embeddings_path: gs://datcom-nl-models/embeddings_undata_2024_06_13_18_31_43.ft_final_v20230717230459.all-MiniLM-L6-v2.csv
     model: ft-final-v20230717230459-all-MiniLM-L6-v2
     healthcheck_query: "Hunger"
   undata_ilo_ft:
     store_type: MEMORY
-    embeddings_path: gs://datcom-nl-models/embeddings_undata_ilo_2024_05_15_11_18_05.ft_final_v20230717230459.all-MiniLM-L6-v2.csv
-    model: ft-final-v20230717230459-all-MiniLM-L6-v2
-    healthcheck_query: "Employment"
-  undata_dev_ft:
```
shifucun, line 74: @dwnoble I removed undata_dev_ft as it is the union of undata_ft and un_ilo_ft

server/integration_tests/test_data/e2e_triple/whatistheprevalenceofheartdiseaseincalifornia/chart_config.json
```diff
@@ -416,494 +416,14 @@
               }
             ],
             "title": "Diabetes in Counties of California"
-          },
-          {
```
chejennifer, line 420: curious why we lost this?

chejennifer (commented):
this is very nice, thank you!

deploy/nl/catalog.yaml
dwnoble: Nice! With this change, how do i call the /detect-and-fulfill endpoint to fetch both undata and undata_ilo? Currently the request looks like

```
POST /api/explore/detect-and-fulfill?q=employment%20in%20africa&client=ui_query
{
  "contextHistory":[],
  "dc":"undata_ilo",
  "disableExploreMore":""
}
``

shifucun: this should still work since you have "dc" param set!

server/integration_tests/test_data/e2e_triple/whatistheprevalenceofheartdiseaseincalifornia/chart_config.json
shifucun: This is due to the update of stat var descriptions. and these variables get even lower score. Since these are mostly unrelated variables, I would say this is a win?

chejennifer (approved):

deploy/nl/catalog.yaml
dwnoble: Got it - so the the undata_dev dc param still exists, it's now just defined by the combination of undata_ilo_ft and undata_ft 👍

dwnoble (approved):

deploy/nl/catalog.yaml
shifucun: yes, so no need to build undata_dev index anymore

pradh (commented):
Nice cleanup!

PR #4353, opened 2024-06-14
[bard] Deployment updates

chejennifer:
- download nl files in the nl server docker image (this allows for increasing the number of nl server workers)
- in the bard deployment config, increase number of nl server workers & deploy mixer-node in a high cpu node pool
- improve load testing output by also tracking nl debug timing information


build/nl_server/Dockerfile
```diff
@@ -31,6 +31,13 @@ RUN pip3 install --no-cache-dir -r /workspace/nl_server/requirements.txt
 COPY nl_server/. /workspace/nl_server/
 COPY shared/. /workspace/shared/
 
+# Download nl files from gcs
+COPY deploy/nl/catalog.yaml .
+COPY build/nl_server/requirements.txt /workspace/build/nl_server/requirements.txt
+COPY build/nl_server/download_nl_files.py .
+RUN pip3 install -r /workspace/build/nl_server/requirements.txt
+RUN python3 download_nl_files.py
```
shifucun, line 39: wonder how much larger is the new docker image?

build/nl_server/download_nl_files.py
```diff
@@ -0,0 +1,48 @@
...
+    if model_info.type != ModelType.LOCAL:
+      continue
+    gcs.maybe_download(model_info.gcs_folder, use_anonymous_client=True)
+
+  # Download all the indexes that are MEMORY or LANCEDB store type
+  for index_info in catalog.indexes.values():
+    if not index_info.store_type in [StoreType.MEMORY, StoreType.LANCEDB]:
+      continue
+    if gcs.is_gcs_path(index_info.embeddings_path):
+      gcs.maybe_download(index_info.embeddings_path, use_anonymous_client=True)
```
shifucun, line 44: I did some more research on /tmp folder and i feel the way we have been using it is dangerous due the ttl of /tmp. I wonder if we should not use /tmp at all for saving and using model/embeddings.

tools/nl/loadtest/load.py
```diff
@@ -48,12 +48,16 @@ def _run_query(query):
...
       debug_timing = response.json().get('debug', {}).get('timing', {})
-      return {
+      result = {
           _RESULT_KEY_CODE: str(response.status_code),
           _RESULT_KEY_NL_TIME: debug_timing.get('getNlResult'),
           _RESULT_KEY_TOTAL_TIME: debug_timing.get('total'),
-          _RESULT_KEY_TILE_TIME: debug_timing.get('getTileResults')
+          _RESULT_KEY_TILE_TIME: debug_timing.get('getTileResults'),
       }
+      result.update(response.json().get('debug',
```
shifucun, line 57: why there are two 'debug' key here?

build/nl_server/Dockerfile
chejennifer: about 300MB larger: https://screenshot.googleplex.com/9k9jurE7fquXmKe

tools/nl/loadtest/load.py
chejennifer: that's just the format of the API response

build/nl_server/download_nl_files.py
chejennifer: updated so that in docker image, we download to nl_server/data, but locally we download to /tmp

nl_server/embeddings.py
```diff
@@ -16,10 +16,23 @@
...
+
+
+# Get the root folder path to download files to
+def get_download_root() -> str:
+  # When running on docker, use the docker path
+  if os.path.exists(DOCKER_DATA_FOLDER_PATH):
+    return DOCKER_DATA_FOLDER_PATH
+  # otherwise, use tmp folder
+  else:
+    return '/tmp'
```
shifucun, line 34: this is not really embedding related const and functions. Shall we have a new nl_server/cache.py and put these code there?

shifucun: also suggest to change function name to get_cache_root()

build/nl_server/download_nl_files.py
```diff
@@ -0,0 +1,49 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Function to download all the nl files in catalog.yaml from gcs"""
+
+from nl_server import config_reader
+from nl_server.config import ModelType
+from nl_server.config import StoreType
+from nl_server.embeddings import DOCKER_DATA_FOLDER_PATH
```
shifucun, line 20: can we use the new function get_download_root() to get the path here?

build/nl_server/Dockerfile
```diff
@@ -31,6 +31,13 @@ RUN pip3 install --no-cache-dir -r /workspace/nl_server/requirements.txt
 COPY nl_server/. /workspace/nl_server/
 COPY shared/. /workspace/shared/
 
+# Download nl files from gcs
+COPY deploy/nl/catalog.yaml .
+COPY build/nl_server/requirements.txt /workspace/build/nl_server/requirements.txt
+COPY build/nl_server/download_nl_files.py .
```
shifucun, line 37: can you copy these files to keep the git file path structure?

That way, you can use the get_download_root() function and will not break due to files movement in future PRs.

chejennifer: sorry, I'm not sure what you mean by copy to keep the git file path structure?

build/nl_server/download_nl_files.py
chejennifer: I think it makes more sense to use the DOCKER_DATA_FOLDER_PATH here because the idea is that: get_download_root() will determine whether to use DOCKER_DATA_FOLDER_PATH or /tmp based on the existence of DOCKER_DATA_FOLDER_PATH .. and at this point in the docker build, that folder won't exist yet

nl_server/embeddings.py
chejennifer: done

build/nl_server/Dockerfile
shifucun: I mean 
`COPY build/nl_server/download_nl_files.py /workspace/build/nl_server/download_nl_files.py`

build/nl_server/download_nl_files.py
shifucun: oh, i see, this is to get that path ready..

build/nl_server/Dockerfile
chejennifer: hm, I tried making that update but getting an error in the next line when trying to run the file .. I think because download_nl_files.py file needs to be at the same folder level as nl_server

shifucun (approved):
thanks for the update!

chejennifer:
thanks for the review!

PR #4354, opened 2024-06-14
[nl] add param to skip getting related things

chejennifer:


static/nodejs_server/query.ts
```diff
@@ -294,6 +294,7 @@ export async function getQueryResult(
     idx,
     mode,
     varThreshold,
+    skipRelatedThings: wantRelatedQuestions ? "" : "True",
```
shifucun, line 297: nit: use "true"

static/nodejs_server/query.ts
```diff
@@ -330,6 +331,7 @@ export async function getQueryResult(
       enclosedPlaceType;
   }
   const relatedThings = nlResp.data["relatedThings"] || {};
+  console.log(relatedThings);
```
shifucun, line 334: remove logging?

server/routes/explore/api.py
```diff
@@ -118,6 +118,8 @@ def detect_and_fulfill():
       Params.EXP_MORE_DISABLED.value] = request.get_json().get(
           Params.EXP_MORE_DISABLED, "")
   utterance.insight_ctx[Params.DC.value] = dc_name
+  utterance.insight_ctx[Params.SKIP_RELATED_THINGS] = request.args.get(
```
shifucun, line 121: should here set the value to be a boolean while checking `lower(Params.SKIP_RELATED_THINGS.value) == 'true'`?

chejennifer: done

static/nodejs_server/query.ts
chejennifer: done

static/nodejs_server/query.ts
chejennifer: done

shifucun (approved):

server/routes/explore/api.py
pradh: nit: use lower()?

PR #4355, opened 2024-06-14
A diff tool to show 2 website page in 2 iframes for comparison

shifucun:
Can enter full url in domain field, or enter the path part in the Path field

image


static/js/apps/diff/app.tsx
```diff
@@ -0,0 +1,115 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import React, { useRef, useState } from "react";
+
+const INIT_URL = [
```
chejennifer, line 19: add a comment about what's in this list & why there are 3 strings? 

instead of a list, maybe better to have an object? because as I read the rest of the code, I think the first 2 strings are domains that are being compared and the third string is the part of the url after the domain? but that was a little confusing

static/js/apps/diff/app.tsx
```diff
@@ -0,0 +1,115 @@
...
+  const rest = `${urlObj.pathname}${urlObj.search}${urlObj.hash}`;
+  return { domain, rest };
+}
+
+export function App(): JSX.Element {
+  const ref = useRef<HTMLInputElement>(null);
+  const iframeRef1 = useRef<HTMLIFrameElement>(null);
+  const iframeRef2 = useRef<HTMLIFrameElement>(null);
+  const [url, setUrl] = useState<string[]>(INIT_URL);
+  const [tmpUrl, setTmpUrl] = useState<string[]>(INIT_URL);
```
chejennifer, line 37: nit: inputUrl? to be more specific about the use of this state

static/js/apps/diff/app.tsx
```diff
@@ -0,0 +1,115 @@
...
+
+  const onKeyDown = (e) => {
+    if (e.key === "Enter") {
+      {
+        const { domain, rest } = splitUrl(tmpUrl[0]);
+        if (rest !== "/") {
```
chejennifer, line 43: hm, so what happens when `rest === "/"`? did you just want to use the previous rest?

static/js/apps/diff/app.tsx
```diff
@@ -0,0 +1,115 @@
...
+        />
+      </div>
+      <label>
+        <div className="label-text">Path (url without domain)</div>
+      </label>
+      <input
+        type="text"
+        ref={ref}
+        value={tmpUrl[2]}
+        placeholder="Enter Path or full URL"
```
chejennifer, line 95: the label in line 89 says "url without domain", so should this just be "Enter path"?

static/js/apps/diff/app.tsx
```diff
@@ -0,0 +1,115 @@
...
+        onChange={(e) => {
+          setTmpUrl([tmpUrl[0], tmpUrl[1], e.target.value]);
+        }}
+        onKeyDown={onKeyDown}
+      />
+      <div className="iframe-container">
+        <iframe
+          ref={iframeRef1}
+          id="iframe1"
+          src={`${url[0]}${url[2]}`}
```
chejennifer, line 105: hm, will you need to add a "/" between the two strings? or is it assumed that either the domain will end in "/" or the path will start with "/"

chejennifer (commented):
this is very cool!

static/js/apps/diff/app.tsx
shifucun: updated to an interface

static/js/apps/diff/app.tsx
shifucun: Done

static/js/apps/diff/app.tsx
shifucun: yea, in that case, the domain is not updated and no extra path is provided, so no action needs to take.

static/js/apps/diff/app.tsx
shifucun: good catch

static/js/apps/diff/app.tsx
shifucun: The path has to start with /. I added that to the place holder text.

shifucun (commented):
thanks for the review!

chejennifer (approved):
thanks for the updates!

pradh (approved):
Very nice!  Look forward to using it!

PR #4359, opened 2024-06-17
Added bard staging environment

dwnoble:
- Added bard staging environment
- Fixed bug in `gke/get_storage_permission.sh` where the project environment name was being used for the robot service account instead of the project_id. Before: `website-robot@bard_staging.iam.gserviceaccount.com` (inconsistent with the account created by `create_robot_account.sh` script . After : `website-robot@datcom-bard-staging.iam.gserviceaccount.com`


deploy/helm_charts/envs/bard_staging.yaml
```diff
@@ -0,0 +1,93 @@
...
+    replicas: 3
+    nodePool: "default-pool"
+  default:
+    replicas: 3
+    nodePool: "default-pool"
+
+svg:
+  blocklistFile: ["dc/g/Uncategorized", "oecd/g/OECD"]
+
+cronTesting:
```
chejennifer, line 88: we probably don't need cronTesting for staging instance?

chejennifer (approved):
thank. you for setting this up!

deploy/helm_charts/envs/bard_staging.yaml
dwnoble: removed

PR #4363, opened 2024-06-18
[rag eval] read eval type from sheet

chejennifer:


static/js/apps/eval_retrieval_generation/app.tsx
```diff
@@ -143,6 +148,34 @@ export function App(props: AppPropType): JSX.Element {
...
+        const metadataKey = sheet.getCell(i, header[METADATA_KEY_COL]).value;
+        if (metadataKey === METADATA_KEY_TYPE) {
+          setEvalType(
+            sheet.getCell(i, header[METADATA_VAL_COL]).value as EvalType
+          );
+          return;
+        }
+      }
+      // by default, set eval type as RIG
+      setEvalType(EvalType.RIG);
```
shifucun, line 175: maybe raise exception just in case this field is not populated? (old doc, or new type..)

shifucun (approved):
nice!! and feel like there is a mini framework now!

static/js/apps/eval_retrieval_generation/app.tsx
chejennifer: added as an alert

PR #4364, opened 2024-06-18
[NL Eval Tool] Don't set default query id to 1

shifucun:


pradh (approved):
Thanks for the quick fix!

PR #4365, opened 2024-06-18
DC website compare tool navigation support

shifucun:
With this, when clicking in one side of the iframe, the other iframe can be updated correspondingly based on the url path.

### DIFF
--- a/server/templates/base.html
+++ b/server/templates/base.html
@@ -77,13 +77,13 @@
     // Post the iframe's URL to the parent
     window.addEventListener('message', function(event) {
       // MUST verify the origin to be autopush DC
-      // TODO: remove localhost after verification
-      if (event.origin === 'https://autopush.datacommons.org' || event.origin === 'http://localhost:8080') {
-        console.log('Message received from parent:', event.data);
+      if (event.origin === 'https://autopush.datacommons.org') {
         // Send the iframe's URL back to the parent
         if (event.data === 'Request URL') {
-          console.log('Message from parent:', event.data);
-          event.source.postMessage({ type: 'URLResponse', url: window.location.href }, event.origin);
+          event.source.postMessage(
+            { type: 'URLResponse', url: window.location.href },
+            event.origin
+          );
         }
       }
     });
--- a/server/templates/base.html
+++ b/server/templates/base.html
@@ -74,8 +74,10 @@
     globalThis.STAT_VAR_HIERARCHY_CONFIG = {{ config['STAT_VAR_HIERARCHY_CONFIG'] | tojson }};
   </script>
   <script>
-    // Post the iframe's URL to the parent
+    // This script is to allow embedding the website as an iframe and allow the
+    // parent to request the iframe's URL. This is used by the /dev/diff tool.
     window.addEventListener('message', function(event) {
+      // The parent should send a postMessage event.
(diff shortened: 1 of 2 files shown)


server/templates/base.html
```diff
@@ -73,6 +73,21 @@
     globalThis.isCustomDC = {{ config['CUSTOM']|int }};
     globalThis.STAT_VAR_HIERARCHY_CONFIG = {{ config['STAT_VAR_HIERARCHY_CONFIG'] | tojson }};
   </script>
+  <script>
+    // Post the iframe's URL to the parent
```
dwnoble, line 77: What does this block of code do?

server/templates/base.html
```diff
@@ -73,6 +73,21 @@
...
+    window.addEventListener('message', function(event) {
+      // MUST verify the origin to be autopush DC
+      if (event.origin === 'https://autopush.datacommons.org') {
```
dwnoble, line 80: What's the check for autopush.datacommons.org for?

static/js/apps/diff/app.tsx
```diff
@@ -42,20 +42,35 @@ export function App(): JSX.Element {
   const [url, setUrl] = useState<AppUrl>(INIT_URL);
   const [inputUrl, setInputUrl] = useState<AppUrl>(INIT_URL);
 
+  useEffect(() => {
```
dwnoble, line 45: I'm not sure the event passing is necessary here. If the goal is to keep the two iframes in sync, could you just set a local react state variable in this component from the iframe's onLoad and have the the relevant iframe re-render based on that state?

server/templates/base.html
shifucun: This block is to pass the iframe site url back to the parent

server/templates/base.html
shifucun: basically only allow autopush to host this comparator app

static/js/apps/diff/app.tsx
shifucun: I tried that but it's not possible due to cross origin check. Googled around and this seems to be a popular way of doing it.

static/js/apps/diff/app.tsx
```diff
@@ -42,20 +42,35 @@ export function App(): JSX.Element {
...
+    const handleMessage = (event) => {
+      if (event.data.type === "URLResponse") {
+        const iframeURL = event.data.url;
+        const { domain, path } = splitUrl(iframeURL);
+        console.log(domain, path);
```
chejennifer, line 50: remove logging?

chejennifer (approved):
this is cool!

static/js/apps/diff/app.tsx
shifucun: Done

server/templates/base.html
dwnoble: is there  a way to extract this out of base.html to a ts/tsx file? if not, can you add some comments giving context as to what the intent is of the code (e.g., this supports the /dev/diff tool so navigating between the two sides i synchronized)

shifucun: Done

dwnoble (approved):

shifucun:
thanks for review!

PR #4372, opened 2024-06-20
[rag eval] add concept of feedback stage

chejennifer:
- Add a new FeedbackStage enum to be used when deciding what to display in the query section and feedback sections
- Changes to make RIG eval tool work with this new concept
- only change made for RAG was displaying the rag calls in the query section when feedback stage is the CALLS stage. Actual navigation for RAG will come later

### DIFF
--- a/static/js/apps/eval_retrieval_generation/util.ts
+++ b/static/js/apps/eval_retrieval_generation/util.ts
@@ -16,6 +16,8 @@
 
 import _ from "lodash";
 
+import { EvalType, FeedbackStage } from "./types";
+
 const HTTP_PATTERN = /https:\/\/[^\s]+/g;
 const LONG_SPACES = "&nbsp;&nbsp;&nbsp;&nbsp;";
 const TABLE_DIVIDER_PATTERN = /[--][-]+/g;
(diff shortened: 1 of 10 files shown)


static/js/apps/eval_retrieval_generation/util.ts
```diff
@@ -66,3 +68,12 @@ export function processTableText(text: string): string {
   // created
   return text.replace(TABLE_DIVIDER_PATTERN, tableDivider);
 }
+
+// Get the first feedback stage to show for an eval type
+export function getFirstFeedbackStage(evalType: EvalType): FeedbackStage {
```
shifucun, line 73: curious will there be function like getSecondFeedbackStage later?

shifucun (approved):
very clean and structured!

static/js/apps/eval_retrieval_generation/util.ts
chejennifer: Some ideas I'm playing with:
- getNextFeedbackStage(evalType, feedbackStage) & getPrevFeedbackStage(evalType, feedbackStage)
- turn feedbackNavigation into RigFeedbackNavigation and RagFeedbackNavigation and they each have the control over setting the feedback stage
- have a config that is a list of stages and as next or prev is clicked, we move through the list of stages

chejennifer:
thanks for the review!

static/js/apps/eval_retrieval_generation/util.ts
shifucun: I see. So sounds like this needs more work.

PR #4374, opened 2024-06-20
Avoid place fallback for mode=toolformer_*

pradh:
Data Gemma use-cases expect exact place's results, and when we don't have data, don't try to do place fallback.

**Before**
image

**After**
image


chejennifer (approved):
nice

PR #4382, opened 2024-06-21
fixed caching issue for POST requests to /api/observations/series

dwnoble:
Before: POST request for a stat var returns the the wrong result due to caching
Requested: Count_Person
Screenshot 2024-06-21 at 3 34 31 PM
Got: SDG variable
Screenshot 2024-06-21 at 3 34 40 PM

After cache key correction:
Screenshot 2024-06-21 at 3 38 24 PM


server/lib/util.py
```diff
@@ -650,8 +650,11 @@ def post_body_cache_key():
   Builds flask cache key for POST requests using the request path and
   JSON-encoded post body
```
beets, line 651: please update comment since this is now used for GET too.

beets (approved):
could also we file a bug to make a test for this? thanks!

server/lib/util.py
dwnoble: Done!

dwnoble:
> could also we file a bug to make a test for this? thanks!

Added tracking bug here: b/348713165

PR #4385, opened 2024-06-24
[rag eval] some small bug fixes

chejennifer:
- fix some table rendering issues
- add stage indicator
- update overall options for RAG to be: irrelevant, somewhat relevant, relevant
- update dc response options for RAG to be: does not match question, matches question
- handle when there are no dc calls


shifucun (approved):

shifucun:
does this fix the next/prev navigation issue?

chejennifer:
> does this fix the next/prev navigation issue?

it should fix that issue for when there are no calls in a query

PR #4389, opened 2024-06-24
Keep only one build_embedding tool and move all custom DC logic/path logic to admin/html.py

shifucun:
Major changes:

- Put most custom DC specific constants, path, logic into admin/html.py. As this is the "offline import manager", it makes sense to let it decide on all these params and pass to scripts in different stages.
- Make build_embeddings.py a general tool that can be purely self contained.
- Support catalog_dict and catalog_paths when reading catalog.
- In custom DC docker, put catalog.yaml as the same path as the GKE deployment, to unify the paths.


server/routes/admin/html.py
```diff
@@ -25,8 +26,14 @@
 from flask import render_template
 from flask import request
 
+from shared.lib import constants
+
 bp = Blueprint('admin', __name__, url_prefix='/admin')
 
+_CUSTOM_EMBEDDING_INDEX = 'user_all_minilm_mem'
```
keyurva, line 33: Consider adding comments on what these are and why hardcoding is ok.

server/routes/admin/html.py
```diff
@@ -72,11 +98,13 @@ def load_data():
...
-      'tools.nl.embeddings.build_custom_dc_embeddings',
-      '--input_dir',
-      f'{nl_dir}',
+      'tools.nl.embeddings.build_embeddings',
+      '--embeddings_name',
+      _CUSTOM_EMBEDDING_INDEX,
       '--output_dir',
       f'{nl_embeddings_dir}',
+      '--catalog',
+      json.dumps(custom_catalog_dict),
```
keyurva, line 107: Do we want the offline script to also specify the catalog dict? Would it be simpler to pass just the `source_path` and `embeddings_path` and have the embeddings tool create the full dict internally?

tools/nl/embeddings/build_custom_dc_embeddings.py
keyurva, line 1: @ajaits - if you are able to check in your offline script PR before this one, we can update the script in this PR to use the other tool. If not, your script will need to be adapted to use the other tool.

server/routes/admin/html.py
shifucun: done

server/routes/admin/html.py
shifucun: In a regular flow (like adding base DC embedding index, for example, undata), we would:

1. create the entry in catalog.yaml manually
2. run build_embeddings tool to create embedding csv (w/o pre-defined output path).
3. use the catalog.yaml to bring up nl server.

The process here actually mimics that flow.


In the old way of producing / using custom_catalog.yaml, the yaml was not really used as a config (which should be input), but rather an output.. that makes the build embeddings tool not generalizable (an effect of deviating the natural process mentioned above). That also makes the embeddings_path and embeddings source csv path spread across nl server, website(admin) and build_embeddings tool. Now this file would be the source of truth and provide arguments for all the rest.

Also after migration to the script, we should have a step of creating this catalog.yaml as a pre-step before running all the stages below as well.

tools/nl/embeddings/build_custom_dc_embeddings.py
shifucun: I have 2 more PRs to follow and would like to get this in asap. I can help you update the tool script. @ajaits

shifucun:
thanks for the review!

server/routes/admin/html.py
keyurva: OK. Let's get this PR and @ajaits's PR in and we can work on the script changes following that.

keyurva (approved):
Thanks for the consolidation!

nl_server/config_reader.py
```diff
@@ -44,17 +44,9 @@ def _path_from_current_file(rel_path: str) -> str:
...
-# catalog from checked in file
-_CATALOG_CODE_PATH = _path_from_current_file('../deploy/nl/catalog.yaml')
-# catalog from mounted file, this is for GKE deployments
-_CATALOG_MOUNT_PATH = '/datacommons/nl/catalog.yaml'
-# catalog from user provided file, this is for custom DCs
-# TODO: use absolute path set by user.
-_CATALOG_USER_PATH_SUFFIX = 'datacommons/nl/custom_catalog.yaml'
-
-# env paths
+# Default catalog paths to load from
```
pradh, line 47: what are these default paths are for different envs?

deploy/nl/catalog.yaml
```diff
@@ -59,43 +59,43 @@ indexes:
     healthcheck_query: "Life expectancy"
```
pradh, line 59: What about the path for medium_ft?  Wonder when we can get rid of it?

pradh (commented):
Very nice!

nl_server/config_reader.py
shifucun: updated in follow up PR

deploy/nl/catalog.yaml
shifucun: medium_ft does not have a path now or does not worth to update in anyway.

Now that the embeddings pipeline are in shape, we should plan migrate custom DC to vertex AI based model as well.

PR #4393, opened 2024-06-25
Added 'part' selectors to 'show metadata' link

dwnoble:
Gives web component users ability to hide and style the 'show metadata' link


juliawu (approved):
Nice! Let's add these parts to the documentation on the docsite as well. (We can time the submit with the next website release?)

PR #4396, opened 2024-06-25
remove redirect for missing trailing slashes on explore links

beets:
the redirect from /explore landing pages caused a downgrade from https to http.
also fix a few http:// links to docsite.

follow up to #4392


shifucun (approved):
thanks for the fix!!

PR #4400, opened 2024-06-26
Support query transforms with SV differ

pradh:
Diff for medium_ft with vs. without stop-words removal:  https://storage.mtls.cloud.google.com/datcom-embedding-diffs/shanth_medium_ft_2024_06_25_21_00_30.html

The canonical query in this case has stop-words.


tools/nl/svindex_differ/prep/to_vars.py
```diff
@@ -25,6 +25,7 @@
 flags.DEFINE_string('queryset', '', 'Full path to queryset CSV')
 flags.DEFINE_string('queryset_vars', '', 'Full path to output queryset CSV')
 flags.DEFINE_string('endpoint', 'https://dev.datacommons.org', 'URL endpoint')
+flags.DEFINE_bool('strip_stopwords', True, 'Strip stop-words')
```
shifucun, line 28: wonder why still need to strip stop words in prep? Would there be just one queryset, and the diff tool would strip or not based on its flag?

pradh: Good Q.  Dropping this support in the PR.

But for now, keeping the original query-set too because that has ~600 odd queries while the file with stop-words has ~700.  Meaning there are going to be a bunch of dups.  When we switch to not dropping stop-words then we can delete the other query-set?

For now, to run the diffs based on stop-words, follow the directions in the README.

shifucun (approved):
thanks for getting this support!

pradh:
Thanks for the review!

PR #4401, opened 2024-06-26
Support Incremental embeddings build

shifucun:
Load saved embeddings vector from catalog.yaml and only compute text that are not seen.

Also:
- Use a SentenceObject to hold dcid, text and vector
- Added a bunch of tests


tools/nl/embeddings/file_manager.py
```diff
@@ -0,0 +1,75 @@
+# Copyright 2024 Google LLC
```
shifucun, line 1: This entire file is copied over without change.

tools/nl/embeddings/utils.py
```diff
@@ -124,84 +75,123 @@ def get_model(catalog: Catalog, env: Env, model_name: str) -> EmbeddingsModel:
...
+    sentences = []
+    for _, row in df.iterrows():
+      dcid = row['dcid']
+      sentence = row['sentence']
+      vector = row.drop(labels=['dcid', 'sentence']).astype(float).tolist()
+      sentences.append(SentenceObject(text=sentence, dcid=dcid, vector=vector))
+    return sentences
+  except Exception as e:
+    logging.error(e)
+    return []
```
pradh, line 93: Do we want to continue on Exception (when does it happen?) vs. just throw up to user

tools/nl/embeddings/utils.py
```diff
@@ -124,84 +75,123 @@ def get_model(catalog: Catalog, env: Env, model_name: str) -> EmbeddingsModel:
...
-  """
-  Compute embeddings for a list of text strings.
-  """
+
+def retrieve_embeddings(
+    model: EmbeddingsModel,
+    target_sentences: List[SentenceObject],
+    saved_sentences: List[SentenceObject],
+) -> List[SentenceObject]:
+  """Compute embeddings for a list of sentence objects"""
```
pradh, line 139: Describe the args and return value some more-- what is target and saved?  Do the input SentenceObjects have vector?

tools/nl/embeddings/utils.py
```diff
@@ -124,84 +75,123 @@ def get_model(catalog: Catalog, env: Env, model_name: str) -> EmbeddingsModel:
   return model
 
 
-def build_and_save_preindex(fm: FileManager) -> Tuple[List[str], List[str]]:
+def get_saved_embeddings(embeddings_path: str) -> List[SentenceObject]:
```
pradh, line 78: hmm, why not just "load_embeddings" ?

tools/nl/embeddings/build_embeddings.py
```diff
@@ -54,18 +54,20 @@ def main(_):
...
-  texts, dcids = utils.build_and_save_preindex(fm)
+  target_sentences = utils.build_and_save_preindex(fm)
 
-  # Compute embeddings
-  embeddings = utils.compute_embeddings(texts, model)
+  # Load saved embeddings from previous run.
+  saved_sentences = utils.get_saved_embeddings(index_config.embeddings_path)
+
+  # Retrieve embeddings
+  sentences = utils.retrieve_embeddings(model, target_sentences,
```
pradh, line 63: compute_or_retrieve_embeddings?

tools/nl/embeddings/build_embeddings.py
```diff
@@ -54,18 +54,20 @@ def main(_):
   fm = utils.FileManager(input_dir, output_dir)
 
   # Build and save preindex
-  texts, dcids = utils.build_and_save_preindex(fm)
+  target_sentences = utils.build_and_save_preindex(fm)
 
-  # Compute embeddings
-  embeddings = utils.compute_embeddings(texts, model)
+  # Load saved embeddings from previous run.
+  saved_sentences = utils.get_saved_embeddings(index_config.embeddings_path)
```
pradh, line 60: wonder if this should be "precomputed_embeddings"?

tools/nl/embeddings/build_embeddings.py
```diff
@@ -54,18 +54,20 @@ def main(_):
   fm = utils.FileManager(input_dir, output_dir)
 
   # Build and save preindex
-  texts, dcids = utils.build_and_save_preindex(fm)
+  target_sentences = utils.build_and_save_preindex(fm)
```
pradh, line 57: sentences_to_compute?

tools/nl/embeddings/build_embeddings.py
pradh, line 63: should this be final_embeddings?

pradh (commented):
The flow is nicely abstracted but somewhat hard to read.  Maybe add more comments and clarify naming (left some suggestions)

tools/nl/embeddings/utils.py
shifucun: changed to `load_existing_embeddings`

tools/nl/embeddings/utils.py
shifucun: Want to continue with [] list, since this is helper list, so should not block the main flow. Worst case is to provide no help..(with empty list)

tools/nl/embeddings/utils.py
shifucun: done

shifucun (commented):
thanks for review!

tools/nl/embeddings/utils.py
```diff
@@ -111,87 +116,95 @@ def build_and_save_preindex(fm: FileManager) -> List[SentenceObject]:
...
-  filtered_sentences = [
-      x for x in target_sentences if x.text not in saved_text2sentence
+  # Find preindex that are not in the existing embeddings
+  existing_embeddings_map = {x.preindex.text: x for x in existing_embeddings}
+  filtered_preindexes = [
+      x for x in preindexes if x.text not in existing_embeddings_map
   ]
-  logging.info("Filtered sentences to compute embeddings with model: %d",
-               len(filtered_sentences))
+  logging.info("Size of preindexes to compute embeddings with model: %d",
```
pradh, line 161: nit:  For readability, instead of looping over preindexes twice, how about loop once and do:

```
# Prepare a list of preindexes for computation
result: list[Embedding] = []
preindexes_to_compute: list[Preindex] = []
for p in preindexes:
  if x.text in existing_embeddings_map:
    # Save computation, add to result
    result.append(...)
  else:
    # Compute embedding
    preindexes_to_compute.append(p)
```

pradh (approved):
Looks much better!

tools/nl/embeddings/utils.py
shifucun: yes, this is better.

PR #4402, opened 2024-06-26
Build embeddings with the load data script.

keyurva:
* Ran the script with the floret dataset and here are the build embeddings [results](https://pantheon.corp.google.com/storage/browser/datcom-floret-dev-resources/temp/nl).

### DIFF
--- a/custom_dc/load_data.sh
+++ b/custom_dc/load_data.sh
@@ -81,10 +81,6 @@ function setup_python {
         "https://download.pytorch.org/whl/cpu"
       echo_log "Installing Python requirements from $embeddings_req"
       run_cmd pip3 install -r "$embeddings_req"
-      # TODO: remove install once embeddings doesn't need nl_server/requirements.txt
-      # nlserver_req="$WEBSITE_DIR/nl_server/requirements.txt"
-      # echo_log "Installing Python requirements from $nlserver_req"
-      # run_cmd pip3 install -r "$nlserver_req"
     fi
   fi
 }
--- a/custom_dc/load_data.sh
+++ b/custom_dc/load_data.sh
@@ -82,9 +82,9 @@ function setup_python {
       echo_log "Installing Python requirements from $embeddings_req"
       run_cmd pip3 install -r "$embeddings_req"
       # TODO: remove install once embeddings doesn't need nl_server/requirements.txt
-      nlserver_req="$WEBSITE_DIR/nl_server/requirements.txt"
-      echo_log "Installing Python requirements from $nlserver_req"
-      run_cmd pip3 install -r "$nlserver_req"
+      # nlserver_req="$WEBSITE_DIR/nl_server/requirements.txt"
+      # echo_log "Installing Python requirements from $nlserver_req"
+      # run_cmd pip3 install -r "$nlserver_req"
     fi
   fi
 }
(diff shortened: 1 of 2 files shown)


tools/nl/embeddings/requirements.txt
```diff
@@ -6,4 +6,9 @@ google-cloud-storage==2.15.0
...
 parameterized==0.8.1
 sentence-transformers==2.2.2
-torchvision==0.17.2
\ No newline at end of file
+torchvision==0.17.2
+# Downloading the named-entity recognition (NER) library spacy and the large EN model
+# using the guidelines here: https://spacy.io/usage/models#production
+# TODO: try using the large model
+-f https://github.com/explosion/spacy-models/releases/download/en_core_web_sm-3.7.1/en_core_web_sm-3.7.1-py3-none-any.whl
+en_core_web_sm==3.7.1
```
keyurva, line 14: @shifucun - Needed to add this to the embeddings requirements to eliminate installing the nl_server requirements. Let me know if this is ok.

custom_dc/load_data.sh
```diff
@@ -223,9 +219,45 @@ function generate_embeddings {
...
+      "gcs_folder": "$CUSTOM_MODEL_PATH",
+      "score_threshold": 0.5
+    }
+  }
+}
+EOF
+)
+  local start_ts=$(date +%s)
+  set -x
+  python -m tools.nl.embeddings.build_embeddings \
```
keyurva, line 252: @ajaits - The `run_cmd` function did not work for argument values with spaces so calling the script directly here. Let me know if there's a better way or we can go with this?

custom_dc/load_data.sh
```diff
@@ -223,9 +219,45 @@ function generate_embeddings {
...
+    --embeddings_name "$CUSTOM_EMBEDDING_INDEX" \
+    --output_dir "$NL_EMBEDDINGS_DIR" \
+    --catalog "$CUSTOM_CATALOG_DICT" >> $LOG 2>&1
+  set +x
+  status=$?
```
ajaits, line 257: move this before set, right after the python command so we get the status of python (not set).

keyurva: Thanks - done.

custom_dc/load_data.sh
ajaits: is there are no spaces within CUSTOM_CATALOG_DICT, you can try:
cmd=(python -m tools.nl.embeddings.build_embeddings)
cmd+=(--embeddings_name "$CUSTOM_EMBEDDING_INDEX")
cmd+=(--output_dir "$NL_EMBEDDINGS_DIR")
cmd+=(--catalog "$(echo $CUSTOM_CATALOG_DICT | sed -e 's/ //g')")
run_cmd ${cmd[@]}

tools/nl/embeddings/requirements.txt
shifucun: I was even thinking get rid of the lib here and make a shared requirements.txt. Duplicating the lib makes the version diverge very easily. Saw an bug before due to this.

shifucun:
Added @pradh to also take a look. We had some offline discussion about maintaining and testing the script. Would like to get more opinions.

PR #4406, opened 2024-06-26
[rag eval] update wording in calls feedback, add another question to overall feedback

chejennifer:
- update wording in calls feedback
Screenshot 2024-06-26 at 3 17 43 PM

- add another question to overall feedback
Screenshot 2024-06-26 at 3 17 23 PM


static/js/apps/eval_retrieval_generation/overall_feedback.tsx
```diff
@@ -45,80 +46,118 @@ const LOADING_CONTAINER_ID = "form-container";
...
 const QUESTION_CONFIG: Record<string, Record<string, QuestionConfig>> = {
   [FeedbackStage.OVERALL_QUESTIONS]: {
     [EvalType.RAG]: {
+      questions: [
+        {
+          firestoreKey: QUERY_OVERALL_QUESTIONS_KEY,
+          question:
+            "Do the generated questions seem sufficient to answer the query?",
+          responseOptions: {
+            [OVERALL_QUESTIONS_OPTION_HAS_MISSING]: "Missing obvious questions",
```
shifucun, line 66: why the option needs a key and value constant?

chejennifer: the key is the option we had set in google sheets & value is the way we display it to users

static/js/apps/eval_retrieval_generation/overall_feedback.tsx
```diff
@@ -45,80 +46,118 @@ const LOADING_CONTAINER_ID = "form-container";
...
+            [QUERY_OVERALL_OPTION_SOMEWHAT_RELEVANT]: "Somewhat relevant",
+            [QUERY_OVERALL_OPTION_RELEVANT]: "Relevant",
+          },
+        },
+        {
+          firestoreKey: QUERY_OVERALL_ANS_ACCURACY_KEY,
+          question: "What is the quality of the answer?",
+          responseOptions: {
+            [QUERY_OVERALL_OPTION_HALLUCINATION]:
+              "Poor narrative and/or with hallucinations",
```
pradh, line 105: Trying to think if we should distinguish this from the relevant question.  Otherwise it feels confusing. @shifucun

What do you think of:

For relevant answers, how is the overall answer quality?
- Bad (Poor narrative, no stats references, etc)
- Average
- Good (Well written answer with appropriate statistics)

pradh (commented):
Thanks!

static/js/apps/eval_retrieval_generation/overall_feedback.tsx
shifucun: to clarify, this is an overall evaluation without other conditions?

pradh: hmm, is it though - i.e., do we need a further classification for an answer that is irrelevant?

chejennifer: updated according to offline discussion

static/js/apps/eval_retrieval_generation/overall_feedback.tsx
```diff
@@ -45,80 +44,109 @@ const LOADING_CONTAINER_ID = "form-container";
...
+      questions: [
+        {
+          firestoreKey: QUERY_OVERALL_ANS_KEY,
+          question: "How well does the LLM answer the user query??",
+          responseOptions: {
+            [QUERY_OVERALL_OPTION_IRRELEVANT]: "Does not answer the query",
+            [QUERY_OVERALL_OPTION_PARTIALLY_COMPLETE]:
+              "Partially answers the query",
+            [QUERY_OVERALL_OPTION_OK]: "Fully answers the query",
+          },
```
pradh, line 97: Should we use the existing keys?  Its harder to map across eval results if the keys change, esp since the questions are approx the same as before?

pradh (approved):

static/js/apps/eval_retrieval_generation/overall_feedback.tsx
chejennifer: used existing keys for the "does not answer" and "fully answers" & added a new key for partial answer

pradh: My Q was why not use Partially relevant for "partial answer" ?

static/js/apps/eval_retrieval_generation/overall_feedback.tsx
```diff
@@ -45,80 +43,109 @@ import { EvalType, FeedbackStage } from "./types";
...
-        [QUERY_OVERALL_OPTION_OK]: "No obvious factual inaccuracies",
-      },
     },
     [EvalType.RAG]: {
+      questions: [
+        {
+          firestoreKey: QUERY_OVERALL_ANS_KEY,
+          question: "How well does the LLM answer the user query??",
+          responseOptions: {
+            [QUERY_OVERALL_OPTION_HALLUCINATION]: "Does not answer the query",
```
pradh, line 92: Also this is not hallucination right?

static/js/apps/eval_retrieval_generation/overall_feedback.tsx
chejennifer: OH, I see what you mean, I misunderstood & thought you wanted to use existing keys from RIG. Updated to the keys we originally created for RAG

chejennifer:
thanks for the reviews

PR #4407, opened 2024-06-27
[rag eval] fix crashing question, add loading, fix answer showing instead of questions

chejennifer:
- fix crashing when a query doesn't have an answer
- add loading when fetching data for the feedback section
- update reading either answer or questions in the LHS

### DIFF
--- a/static/js/apps/eval_retrieval_generation/call_feedback.tsx
+++ b/static/js/apps/eval_retrieval_generation/call_feedback.tsx
@@ -121,7 +121,7 @@ export function CallFeedback(): JSX.Element {
         }
       })
       .catch((e) => {
-        console.log(e);
+        alert(e);
         setEvalInfo(null);
       })
       .finally(() => {
--- a/static/js/apps/eval_retrieval_generation/call_feedback.tsx
+++ b/static/js/apps/eval_retrieval_generation/call_feedback.tsx
@@ -27,6 +27,8 @@ import {
   DC_RESPONSE_COL,
   DC_RESPONSE_FEEDBACK_COL,
   DC_STAT_COL,
+  FEEDBACK_FORM_ID,
+  FEEDBACK_PANE_ID,
   LLM_STAT_COL,
   LLM_STAT_FEEDBACK_COL,
 } from "./constants";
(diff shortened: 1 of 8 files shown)


static/js/apps/eval_retrieval_generation/call_feedback.tsx
```diff
@@ -103,19 +104,29 @@ export function CallFeedback(): JSX.Element {
...
+          setEvalInfo({
+            dcResponse: `${row.get(DC_RESPONSE_COL)}${tableResponse}`,
+            dcStat: row.get(DC_STAT_COL),
+            llmStat: row.get(LLM_STAT_COL),
+            question: row.get(DC_QUESTION_COL),
+          });
+        }
+      })
+      .catch((e) => {
+        console.log(e);
```
shifucun, line 124: alert here?

static/js/apps/eval_retrieval_generation/feedback_navigation.tsx
```diff
@@ -156,8 +156,8 @@ export function FeedbackNavigation(
       while (!(targetId in userQuery)) {
         targetId -= 1;
       }
-      setSessionQueryId(targetId);
       setFeedbackStage(getFirstFeedbackStage(evalType));
+      setSessionQueryId(targetId);
```
shifucun, line 160: why the order here matters?

chejennifer: I think it has to do with the pane listening on both feedback stage and session query id changes and I'm not sure it'll be better, but just switched it to the same order as eval list choosing a new query

static/js/apps/eval_retrieval_generation/call_feedback.tsx
chejennifer: done

static/js/apps/eval_retrieval_generation/feedback_navigation.tsx
shifucun: i think react setState is applied async and in batch, so the order here should not matter.

shifucun (approved):
thanks!

chejennifer:
thanks for the review!

PR #4409, opened 2024-06-27
[nl] prune topic vars for rig

chejennifer:
Allow topics for rig but prune the svs so that we only keep the ones that were also separately detected on their own.

Only keep pruned topics if there were no svs returned (which, I'm slightly conflicted if we need this piece)


server/lib/nl/fulfillment/fulfiller.py
```diff
@@ -233,3 +239,50 @@ def _prune_non_country_special_dc_vars(state: PopulateState):
...
   state.chart_vars_map = pruned_chart_vars_map
+
+
+#
+# Prune topic chart vars for toolformer_rig mode
+#
+def _prune_topic_vars_for_rig(state: PopulateState):
+  # Set of all svs that were detected
+  detected_svs = set([
+      sv for sv in state.uttr.detection.svs_detected.single_sv.svs
```
pradh, line 250: Do we need some threshold here, maybe the default threshold?

server/lib/nl/fulfillment/fulfiller.py
```diff
@@ -131,6 +134,9 @@ def fulfill(uttr: Utterance) -> PopulateState:
   if params.is_special_dc(state.uttr.insight_ctx):
     _prune_non_country_special_dc_vars(state)
 
+  if state.uttr.mode == QueryMode.TOOLFORMER_RIG:
+    _prune_topic_vars_for_rig(state)
```
pradh, line 138: Do this for non-multi var case only?

server/lib/nl/fulfillment/fulfiller.py
```diff
@@ -233,3 +239,50 @@ def _prune_non_country_special_dc_vars(state: PopulateState):
...
+#
+# Prune topic chart vars for toolformer_rig mode
+#
+def _prune_topic_vars_for_rig(state: PopulateState):
+  # Set of all svs that were detected
+  detected_svs = set([
+      sv for sv in state.uttr.detection.svs_detected.single_sv.svs
+      if not sv.startswith(_TOPIC_PREFIX)
+  ])
+  # skip topics if there are any non topic vars detected
```
pradh, line 253: add rationale?

pradh: hmm, I'm not sure about this.  for example, topics might be at the top and the vars after that might only be poor matches, esp compared to the variables in the t opic.

server/lib/nl/fulfillment/fulfiller.py
```diff
@@ -233,3 +239,50 @@ def _prune_non_country_special_dc_vars(state: PopulateState):
                              list(dropped_vars))
 
   state.chart_vars_map = pruned_chart_vars_map
+
+
+#
+# Prune topic chart vars for toolformer_rig mode
```
pradh, line 245: Add comments for rationale?

server/integration_tests/test_data/e2e_toolformer_rig_mode/numberofhouseholdsunderpovertylineinseattle/chart_config.json
```diff
@@ -0,0 +1,904 @@
...
+    "categories": [
+      {
+        "blocks": [
+          {
+            "columns": [
+              {
+                "tiles": [
+                  {
+                    "statVarKey": [
+                      "Count_Household_HouseholderAge65OrMoreYears_FamilyHousehold_BelowPovertyLevelInThePast12Months"
```
pradh, line 13: This is not a good match?

server/lib/nl/fulfillment/fulfiller.py
chejennifer: I was going back and forth on this.

One example scenario I was thinking was:
- topic A has variables sv1, sv2 (listed in that order)
- sv1 and sv2 are both separately detected as well
- the order of matching is topicA, sv2, sv1
- if we keep topicA with pruned topics, sv1 will be returned as first chart instead of sv2

& I was thinking for your example, since that's the current behavior, by skipping topics when svs are detected, we won't be doing any worse than currently & slightly better with the addition of topics when no single svs are returned

wdyt (also not sure how likely my example scenario is)

server/integration_tests/test_data/e2e_toolformer_rig_mode/numberofhouseholdsunderpovertylineinseattle/chart_config.json
chejennifer: I was using this as an example of skipping adding topics & unfortunately that's the highest score match .. https://screenshot.googleplex.com/AdTpyVqHPv48PDo

server/lib/nl/detection/variable.py
```diff
@@ -77,9 +77,8 @@ def detect_vars(orig_query: str, debug_logs: Dict,
   # 2. Lookup embeddings with both single-var and multi-var queries.
   #
   # Make API call to the NL models/embeddings server.
-  skip_topics = dargs.mode == params.QueryMode.TOOLFORMER_RIG
-  resp = dc.nl_search_vars(all_queries, dargs.embeddings_index_types,
-                           skip_topics, dargs.reranker)
+  resp = dc.nl_search_vars(all_queries, dargs.embeddings_index_types, False,
```
pradh, line 80: skip_topics might no longer be needed at all, check?

server/lib/nl/fulfillment/fulfiller.py
pradh: Actually how about the following logic instead:

1. Collect all SVs part of the topic chart_vars.
2. Collect all non-topic SVs in chart_vars.
3. Walk down the matched vars list:
    a. If topic, skip
    b. Else, if part of (1) and not part of (2), introduce chart_vars entry.

This keeps the original order of the SVs faithfully.

server/lib/nl/fulfillment/fulfiller.py
chejennifer: sorry, not sure what you want me to check here .. there's nothing about multivar or not?

pradh: just coz, we don't want to arbitrarily prune only LHS or RHS?  for multi-var its always 2 svs

server/lib/nl/fulfillment/fulfiller.py
chejennifer: done

server/lib/nl/fulfillment/fulfiller.py
chejennifer: i this this should be handled with the new logic

server/lib/nl/fulfillment/fulfiller.py
chejennifer: i like this, updated

server/lib/nl/fulfillment/fulfiller.py
chejennifer: done

server/integration_tests/test_data/e2e_toolformer_rig_mode/numberofhouseholdsunderpovertylineinseattle/chart_config.json
chejennifer: actually this test is not needed anymore

server/lib/nl/detection/variable.py
chejennifer: removed

server/services/datacommons.py
```diff
@@ -336,15 +336,10 @@ def resolve(nodes, prop):
...
-def nl_search_vars(queries,
-                   index_types: List[str],
-                   skip_topics=False,
-                   reranker=''):
+def nl_search_vars(queries, index_types: List[str], reranker=''):
   """Search sv from NL server."""
   idx_params = ','.join(index_types)
   url = f'{current_app.config["NL_ROOT"]}/api/search_vars?idx={idx_params}'
-  if skip_topics:
-    url = f'{url}&skip_topics={skip_topics}'
```
pradh, line 347: TODO:  cleanup skip_topics argument in nl_server?

server/lib/nl/fulfillment/fulfiller.py
pradh: Hmmm, multi-var uses MultiSV and its very different.  So I'm not sure if this type of handling even makes sense?

server/lib/nl/fulfillment/fulfiller.py
```diff
@@ -233,3 +240,56 @@ def _prune_non_country_special_dc_vars(state: PopulateState):
...
+
+  # add chart vars for detected svs with a score above model threshold & part
+  # of a topic that was in the original chart vars
+  added_svs = set()
+  for i, sv in enumerate(state.uttr.detection.svs_detected.single_sv.svs):
+    # skip topic svs
+    if sv.startswith(_TOPIC_PREFIX):
+      continue
+    # skip if score is below default threshold
+    if state.uttr.detection.svs_detected.single_sv.scores[
```
pradh, line 277: nit: use tmp vars to avoid long strings?

server/lib/nl/fulfillment/fulfiller.py
```diff
@@ -233,3 +240,56 @@ def _prune_non_country_special_dc_vars(state: PopulateState):
...
+        i] < state.uttr.detection.svs_detected.model_threshold:
```
pradh, line 278: nit: maybe for readability:

`for sv, score in zip(svs, scores):`

server/lib/nl/fulfillment/fulfiller.py
```diff
@@ -233,3 +240,56 @@ def _prune_non_country_special_dc_vars(state: PopulateState):
...
+      continue
+    # skip if not part of a topic that was detected
+    if not sv in topic_svs:
+      continue
+    updated_chart_vars_map[sv] = updated_chart_vars_map.get(sv, [])
+    updated_chart_vars_map[sv].append(
+        ChartVars(svs=[sv], orig_sv_map={sv: [sv]}))
```
pradh, line 285: Hmmm, we don't want to add it if it is already there?

Concretely, should we avoid re-adding SVs that are already there in chart_vars.svs?

pradh (commented):
Thanks for the edits!

server/services/datacommons.py
chejennifer: done

server/lib/nl/fulfillment/fulfiller.py
chejennifer: updated as we discussed offline

server/lib/nl/fulfillment/fulfiller.py
chejennifer: done

server/lib/nl/fulfillment/fulfiller.py
chejennifer: oh that's nice, done

server/lib/nl/fulfillment/fulfiller.py
chejennifer: as discussed offline, don't add if sv is already a key in chart vars

pradh (approved):
Thanks!

server/lib/nl/fulfillment/fulfiller.py
```diff
@@ -131,13 +131,12 @@ def fulfill(uttr: Utterance) -> PopulateState:
       state.chart_vars_map = topic.compute_chart_vars(state)
   else:
     state.chart_vars_map = topic.compute_chart_vars(state)
+    if state.uttr.mode == QueryMode.TOOLFORMER_RIG:
```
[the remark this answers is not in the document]
chejennifer: done

chejennifer:
thanks for the review!

PR #4410, opened 2024-06-27
Make a shared nl requirements.txt that can be imported from nl server and nl tools

shifucun:


nl_server/model/create.py
keyurva, line 1: Confirming if this file only requires `nl_requirements.txt` and not `nl_server/requirements.txt`. If that's the case, is there a way to ensure that it remains that way moving forward? If not, consider adding a comment to the file to avoid dependence on the latter.

keyurva (approved):
Thanks!

nl_server/model/create.py
shifucun: I think it's hard to ensure that.. It's fair to say files under nl_server/ would need nl_server/requirements.txt.

The pip install happens in shell script, so should rely on e2e tests (including the shell script) to catch the pip package issues.

keyurva: That makes this brittle - any change that requires nl server requirements can break the tools.

Longer term does it make sense to create a `nl_shared` package that can be shared between nl server and nl tools?

shifucun: During coding time, there is almost no way to know if a used library will need any new packages than existing requirements.txt (same for cc or java, but that would error out in compile time, say, a system lib is missing).

The safest way is just to use nl_server/requirements.txt. And here we feel that's too big and want to optimize that. 

I don't know if this optimization can be tested anyway by a mechanism like c++ compiling, but during coding time.

shifucun: Would be good to see a solid solution for this though! Worth some exploration

PR #4412, opened 2024-06-27
Add a direct link to the required version of protoc

hqpho:
When I was onboarding, I found the versioning for protoc confusing, and it also took me a while to figure out where I could download the required version. Adding a direct link for convenience for future readers/contributors.


shifucun (approved):
thanks!

PR #4414, opened 2024-06-27
[nodejs] update csv table header for toolformer_rag mode, add website commit in debug info

chejennifer:
- update csv table headers for mode=toolformer_rag:
  - line chart: "label" -> "date" e.g., https://paste.googleplex.com/4545373437952000
  - bar chart: "label" -> "variable" e.g., https://paste.googleplex.com/6171709243916288
  - map_chart: "label" -> "place"

- read website commit from environment and return it as part of debug_info


static/nodejs_server/tiles/bar_tile.ts
```diff
@@ -118,8 +125,10 @@ export async function getBarTileResult(
     ) {
       legend = chartData.dataGroup[0].value.map((dp) => dp.label);
     }
+    const csvLabelHeader =
```
shifucun, line 128: why not using this in all modes?

chejennifer: I worry that changing the header will affect bard because bard reads the csv and then generates text ..

shifucun (approved):

static/js/utils/chart_csv_utils.ts
```diff
@@ -191,8 +195,8 @@ export function mapDataToCsv(layerData: MapLayerData[]): string {
     }
   });
   const header = hasVariable
-    ? ["label", "variable", "data"]
-    : ["label", "data"];
+    ? ["place", "variable", "data"]
```
pradh, line 198: Are we not gating this on TOOLFORMER_RAG?

static/js/utils/chart_csv_utils.ts
```diff
@@ -58,8 +61,9 @@ export function dataGroupsToCsv(dataGroups: DataGroup[]): string {
     const dates = dg.value.map((dp) => dp.label);
     allLabels = new Set([...Array.from(allLabels), ...dates]);
   }
+  const labelHeader = labelHeaderOverride || "label";
```
pradh, line 64: use constant for "label"?

pradh (approved):
Thanks!

static/js/utils/chart_csv_utils.ts
chejennifer: i dont think this needs to be gated because map tiles in nodejs server is the only time this function is used & we don't return map tiles for bard

static/js/utils/chart_csv_utils.ts
chejennifer: done

chejennifer:
thanks for the reviews!

PR #4416, opened 2024-06-28
[Stop Words] Introduce stop words exception list and add "how many", "number of" to this list 

shifucun:
The larger embeddings model can understand semantics better (instead of token match). So we should try to preserve stop words. Including stop words would be a big change and we should do this in a controlled way.

This introduces an exception list that are related to stop words and should be kept. 

With this, the query sentence will be "number of asian" instead of "number asian". Turns out this boosts the matching and the score a lot!

https://storage.mtls.cloud.google.com/datcom-embedding-diffs/boxu_base_uae_mem_2024_07_01_13_41_51.html


server/lib/nl/detection/place.py
```diff
@@ -302,6 +305,9 @@ def _get_stripped_string(query_parts: List[QueryPart],
       # if any of the dcids in the query part is resolved, consider the query part
       # resolved and add the original string to resolved_things_str
       resolved_dcids_str.append(p.substr)
+      if len(stripped_query_parts) > 0:
+        stripped_query_parts[-1] = re.sub(_PLACE_PREPOSITION_REGEX, "",
+                                          stripped_query_parts[-1])
```
pradh, line 310: Why are doing any stop-word related things here?  This just strips out resolved place names?

shared/lib/constants.py
```diff
@@ -151,6 +150,10 @@
     'across',
 }
 
+# For a stop words, if it is preceeded or trailed by certain words, don't trim
+# it.
+STOP_WORDS_EXCEPTION = {'how': ['', 'many'], 'of': ['number', '']}
```
pradh, line 155: This feels hard to use.  Is it worth just listing phrases to keep?  "how many", "number of"

server/lib/nl/detection/place.py
shifucun: Now I try to preserve "in", but only remove if it s before a place like this case. Concretely:

For query like "people in the USA", here removes "USA" and left with "people in the" and I want to remove "in the";

shared/lib/utils.py
```diff
@@ -65,9 +65,20 @@ def remove_stop_words(input_str: str, stop_words: Set[str]) -> str:
   # Example: if looking for "cat" in sentence "cat is a catty animal. i love a cat  but not cats"
   # the words "citty" and "cats" will not be matched.
   input_str = input_str.lower()
-  for words in stop_words:
+  for word in stop_words:
     # Using regex based replacements.
-    input_str = re.sub(rf"\b{words}\b", "", input_str)
+    ex = constants.STOP_WORDS_EXCEPTION.get(word, ['', ''])
```
pradh, line 70: Add some comment what is going on?  And maybe some focussed tests?

server/lib/nl/detection/place.py
shifucun: Or should we have another pass on the queryset list and process the stop words there?

pradh: It is very subtle and feels error prone!  What do you think of introducing a new string altogether with place names replaced with "$_PLACE" or something.  Then downstream in the stop-words replacement, add "in the $_PLACE" as replacement string.

chejennifer (approved):
nice!

shared/lib/utils.py
```diff
@@ -53,6 +58,27 @@ def _add_classification_heuristics(set_strings: Set[str]) -> None:
...
+  for placeholder, exception in placeholder_map.items():
+    text = re.sub(re.escape(exception), placeholder, text)
+  return text
+
+
+# Function to remove words in remove_list but not protected exceptions
+def remove_words(text, remove_list):
+  remove_pattern = r'\b(?:' + '|'.join(
+      re.escape(word) for word in remove_list) + r')\b'
+  return re.sub(remove_pattern, '', text)
```
pradh, line 72: This logic is different from LHS, in the interest of changing only the necessary things, should we restore LHS behavior?  If this RHS improves on it, then add clarifying comments?

shared/lib/utils.py
```diff
@@ -65,13 +91,18 @@ def remove_stop_words(input_str: str, stop_words: Set[str]) -> str:
...
   input_str = input_str.lower()
-  for words in stop_words:
-    # Using regex based replacements.
-    input_str = re.sub(rf"\b{words}\b", "", input_str)
-    # Also replace multiple spaces with a single space.
-    input_str = re.sub(r" +", " ", input_str)
 
-  # Return after removing the beginning and trailing white spaces.
+  # Protect exceptions
+  input_str = protect_exceptions(input_str, _PLACEHOLDER_MAP)
```
pradh, line 96: nit: would `replace_exceptions_with_placeholders` and `replace_placeholders_with_exceptions` be more readable?

pradh (commented):
Useful to have this!

How does the svindex_differ diffs look?

shared/lib/utils.py
shifucun: done

shared/lib/utils.py
```diff
@@ -53,6 +58,27 @@ def _add_classification_heuristics(set_strings: Set[str]) -> None:
       ]
 
 
+# Function to replace exceptions with placeholders
+def protect_exceptions(text, placeholder_map):
```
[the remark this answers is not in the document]
shifucun: done

shared/lib/utils.py
shifucun: done

pradh (approved):
LG!  Pending svindex diffs

server/integration_tests/test_data/explore_strict_default_place/numberofheadlessdriversincalifornia/chart_config.json
```diff
@@ -9,16 +9,75 @@
               {
                 "tiles": [
                   {
-                    "type": "PLACE_OVERVIEW"
+                    "statVarKey": [
```
shifucun, line 12: this is a substantial diff

shifucun:
> https://storage.mtls.cloud.google.com/datcom-embedding-diffs/boxu_base_uae_mem_2024_07_01_13_41_51.html

This is the diff https://storage.mtls.cloud.google.com/datcom-embedding-diffs/boxu_base_uae_mem_2024_07_01_13_41_51.html

shifucun:
The svindex diff shows a few good improvement and some loss(ish); A major behavior change is to boost stat vars (with "number of") higher than the topics!

shifucun:
Updated dc/topic/Disabilities to reflect it's about population. Updated golden based on all the latest changes.

pradh (approved):
Thanks for the change and the diffs!

PR #4420, opened 2024-06-28
Update RAG eval tool counter layout for easier interaction

shifucun:
image

### DIFF
--- a/static/css/eval_retrieval_generation.scss
+++ b/static/css/eval_retrieval_generation.scss
@@ -32,7 +32,8 @@ footer {
   flex-direction: column;
 }
 
-.sign-in, .google-sheet-link {
+.sign-in,
+.google-sheet-link {
   width: fit-content;
 }
 
@@ -45,7 +46,7 @@ footer {
 
 .table-pane {
   overflow: auto;
-  height: 30%;
+  height: 50%;
   padding: 11px 24px;
   width: 100%;
   border-top: solid 1px var(--gm-3-sys-light-outline-variant);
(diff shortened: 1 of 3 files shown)


static/css/eval_retrieval_generation.scss
```diff
@@ -188,13 +189,12 @@ label {
 
   #form-container {
     border-top: solid 1px var(--gm-3-sys-light-outline-variant);
-    gap: 16px;
```
chejennifer, line 191: curious why you're removing this, how do the other feedback panes look with this change?

static/css/eval_retrieval_generation.scss
```diff
@@ -188,13 +189,12 @@ label {
...
     padding: 16px 24px;
     flex-grow: 1;
     overflow: auto;
 
     .question-section .title {
-      font-size: 11px;
+      font-size: 16px;
```
chejennifer, line 197: why change the question title font size? these were specs from adriana's mocks

chejennifer (approved):
nice!

static/css/eval_retrieval_generation.scss
shifucun: reverted

static/css/eval_retrieval_generation.scss
shifucun: looks like this is block by some other css.

shifucun (commented):
thanks for review! updated the heigh handling per offline discussion

PR #4424, opened 2024-06-30
Update NL embeddings eval playground

shifucun:
Remove golden stat vars and eval scores since these are not practical and reliable.

Updated this into a playground that can upload sv descriptions and queries on the fly and check stat var matches. The matches can highlight stat vars that have override descriptions.

image


static/js/apps/eval_embeddings/app.tsx
```diff
@@ -15,134 +15,139 @@
...
-        </p>
-        <p>
-          Each row is in the form of
-          &quot;stat_var_dcid,description1;description2;...&quot;
-        </p>
-        <form onSubmit={handleCustomDescription}>
-          <textarea name="override" />
+        <h3>
+          Enter or upload stat var descriptions (each row like
+          &quot;dcid,desc1;desc2&quot;)
```
pradh, line 102: To be safe against `,` within desc1 and desc2, maybe do:

`dcid,"desc1;desc2"`

?

static/js/apps/eval_embeddings/app.tsx
```diff
@@ -155,37 +160,28 @@ const fetchEmbeddings = async (lines: string[], modelNames: string[]) => {
     }
```
pradh, line 160: To confirm, is this tolerant of additional commas in desc1, desc2?

`dcid1,desc1 with, comma;desc2 with, comma`

static/js/apps/eval_embeddings/main.ts
```diff
@@ -30,22 +30,19 @@ window.onload = () => {
...
+    if (EVAL_INDEXES.has(idx)) {
+      indexes[idx] = serverConfig["indexes"][idx];
     }
   }
   ReactDOM.render(
     React.createElement(App, {
-      evalGolden,
-      index2model,
+      indexes,
+      models: serverConfig["models"],
```
pradh, line 45: [for future] I wonder if its helpful to control the models with url param?  e.g., it seems "medium_ft" is not necessary

pradh (approved):
1. Wonder if we should have the query textbox before the  NL description textbox?  Because I can imagine this playground is helpful to try queries to see the variable matches for different models, as well, optionally, override/add NL descriptions?
2. Wonder if it might be good have a few lines to articulate what to do with the tool?

This playground can be used explore the variable matches given a few queries for different models.  Additionally, you can also override or add NL descriptions for variables.

static/js/apps/eval_embeddings/app.tsx
shifucun: very good point. use a real csv parser now

shifucun:
> 1. Wonder if we should have the query textbox before the  NL description textbox?  Because I can imagine this playground is helpful to try queries to see the variable matches for different models, as well, optionally, override/add NL descriptions?
> 2. Wonder if it might be good have a few lines to articulate what to do with the tool?
> 
> This playground can be used explore the variable matches given a few queries for different models. Additionally, you can also override or add NL descriptions for variables.

Swapped the order. I put the suggestions as text area instruction explicitly now.

PR #4432, opened 2024-07-02
US NTSB Crash NL embeddings

hareesh-ms:
NL embeddings for US NTSB crash dataset.

diff : https://storage.mtls.cloud.google.com/datcom-embedding-diffs/hareeshms_base_uae_mem_2024_07_02_11_49_29.html


tools/nl/embeddings/input/base/sheets_svs.csv
```diff
@@ -2370,6 +2370,10 @@ Count_Person_InLaborForce_ResidesInCollegeOrUniversityStudentHousing,Number of L
 Count_Person_InLaborForce_ResidesInGroupQuarters,Number of Labor Force Participants reside in Group Quarters
 Count_Person_InLaborForce_ResidesInNoninstitutionalizedGroupQuarters,Number of Labor Force Participants reside in Noninstitutionalized Group Quarters
 Count_Person_IsInternetUser_PerCapita,percentage of internet users
+Count_Person_InvolvedInCrash_Motorists,Number of motorists involved in crash
+Count_Person_InvolvedInCrash_MotorVehicleOccupant,Number of motor vehicle occupants involved in crash
+Count_Person_InvolvedInCrash_NonMotorists,Number of non-motorists involved in crash
```
shifucun, line 2375: right now we don't add "non xxx" and "other xxx" variables to the index since they are vague when matching query (or even match the opposite meaning)

tools/nl/embeddings/input/base/sheets_svs.csv
```diff
@@ -2760,6 +2764,10 @@ Count_ThunderstormWindEvent,Number of thunderstorm wind events
 Count_TornadoEvent,Number of tornado events
 Count_TropicalStormEvent,Number of tropical storm events
 Count_UnemploymentInsuranceClaim_StateUnemploymentInsurance,Number of state unemployment insurance claims
+Count_Vehicle_InvolvedInCrash_InTransport,Number of vehicles crashed in transport 
+Count_VehicleCrashIncident_NationalHighway,Number of crash incidents occurred on the national highway
```
shifucun, line 2768: "number of vehicle crash incidents..." same as below

tools/nl/embeddings/input/base/sheets_svs.csv
```diff
@@ -2760,6 +2764,10 @@ Count_ThunderstormWindEvent,Number of thunderstorm wind events
 Count_TornadoEvent,Number of tornado events
 Count_TropicalStormEvent,Number of tropical storm events
 Count_UnemploymentInsuranceClaim_StateUnemploymentInsurance,Number of state unemployment insurance claims
+Count_Vehicle_InvolvedInCrash_InTransport,Number of vehicles crashed in transport 
+Count_VehicleCrashIncident_NationalHighway,Number of crash incidents occurred on the national highway
+Count_VehicleCrashIncident_StateHighway,Number of crash incidents on state highway
+Count_VehicleCrashIncident_USHighway,Number of crash incidents on US highway
```
shifucun, line 2770: US highway is a bit vague in the meaning and maybe very hard to match actual query. Looking at these 3 stat vars, I wonder if there are aggregate stat var?

For example, the most likely query is "how many highway car crashes happened in 2020", which stat vars do we want to match here?

tools/nl/embeddings/input/base/sheets_svs.csv
hareesh-ms: Since its counting the vehicles in-transport, shall we add sentence as "Number of vehicles in-transport involved in crash incidents"

shifucun: oh, i did not meant to include line 2767, that line looks good

pradh:
How about we add to a new CSV file?   Like say based on date:  2024_q3.csv?

tools/nl/embeddings/input/base/sheets_svs.csv
hareesh-ms: The raw dataset we processed is at a crash granularity which we aggregated. Looking at the data the USHighway or StateHighway may or may not be part of the National Highway Network. The US highway seems to be an older system as per https://en.wikipedia.org/wiki/Numbered_highways_in_the_United_States. Shall we remove the USHighway from the index and retain both National highway and State Highway.  Is the question "how many highway car crashes happened in 2020" returning both NationalHighway, StateHighway numbers a desirable behavior ?

hareesh-ms:
Added the NL sentences to a new csv file and modified the NL sentences.

tools/nl/embeddings/input/base/2024_q3.csv
```diff
@@ -0,0 +1,8 @@
+dcid,sentence
+Count_Person_InvolvedInCrash_Motorists,Number of motorists involved in crash
+Count_Person_InvolvedInCrash_MotorVehicleOccupant,Number of motor vehicle occupants involved in crash
+Count_Vehicle_InvolvedInCrash_InTransport,Number of vehicles in motion involved in crash
+Count_VehicleCrashIncident_CollisionCrash_HeadOnCollision,Number of headon vehicle collisions
```
shifucun, line 5: head-on

tools/nl/embeddings/input/base/2024_q3.csv
```diff
@@ -0,0 +1,8 @@
+dcid,sentence
+Count_Person_InvolvedInCrash_Motorists,Number of motorists involved in crash
+Count_Person_InvolvedInCrash_MotorVehicleOccupant,Number of motor vehicle occupants involved in crash
+Count_Vehicle_InvolvedInCrash_InTransport,Number of vehicles in motion involved in crash
```
shifucun, line 4: I see, so "in transport" does not mean being transported from place A to B. 

How about just say "moving vehicles"?

tools/nl/embeddings/input/base/2024_q3.csv
shifucun, line 5: a side question, is there distinction between "collision" and "crash" in schema? And shall we unify / distinct them in description?

pradh (commented):
@hareesh-ms - sorry for the delay.  Is it ready for a review?  I had some of the same comments as Bo.

tools/nl/embeddings/input/base/2024_q3.csv
hareesh-ms: As per definition crash seems to be more severe with damage or fatalities, hence will stick with crash for all.

PR #4435, opened 2024-07-02
Support sv description yaml file; Convert base DC sdg topics csv into yaml

shifucun:


tools/nl/embeddings/input/base/convert.py
```diff
@@ -0,0 +1,38 @@
...
+            offset=0)  # sequence and offset have their default values here
+yaml.preserve_quotes = True
+yaml.width = 800  # this is the output line width after which wrapping occurs
+
+
+def main(_):
+  file_name = FLAGS.input_file
+  sv2texts: Dict[str, List[str]] = {}
+
+  with open(file_name) as f:
```
pradh, line 27: nit: should we support vice versa too?

tools/nl/embeddings/input/base/convert.py
```diff
@@ -0,0 +1,38 @@
+import csv
```
pradh, line 1: nit: Should convert.py live elsewhere?

pradh (commented):
Very nice!

tools/nl/embeddings/input/base/convert.py
shifucun: Did not mean to check in this.. I will let it live for now and remove after converting all the csvs to yaml?

tools/nl/embeddings/input/base/convert.py
shifucun: This is just to convert current input csv to input yaml. Do you mean to generate the preindex as yaml too?

shifucun (commented):
Added indent to text as well

tools/nl/embeddings/input/base/convert.py
pradh: I think its still useful to have because we have scripts that generate csvs for NL descriptions (eg for UN), so for future imports it could still be relevant.

tools/nl/embeddings/input/base/convert.py
pradh: I was thinking in case we need to curate in sheets, but never mind, we probably won't go the other way, but mostly only csv -> yaml.

shifucun: yea, i plan to update the google sheets tool to read yaml later.

tools/nl/embeddings/input/base/convert.py
shifucun: move this as a new tool now

pradh (approved):

shifucun:
thanks for review!

PR #4436, opened 2024-07-02
[eval] set up components to be used by sxs evals

chejennifer:
update components that will be shared with sxs evals to take props instead of using context


static/js/apps/eval_retrieval_generation/table_pane.tsx
```diff
@@ -47,27 +48,28 @@ function getTableTrigger(tableInfo: TableInfo, opened: boolean): JSX.Element {
...
 
   useEffect(() => {
-    if (_.isEmpty(allCall[sessionQueryId])) {
+    if (_.isEmpty(props.calls)) {
       setTables([]);
       return;
     }
-    const sheet = doc.sheetsByTitle[DC_CALL_SHEET];
-    const tableIds = Object.keys(allCall[sessionQueryId]).sort(
+    const sheet = props.doc.sheetsByTitle[DC_CALL_SHEET];
```
shifucun, line 64: why not use doc context, esp it's constant through the session?

chejennifer: This is going to be shared component that a new app can use so thought it would be easier to just pass as a prop so the new app doesn’t have to create a whole context

shifucun: Given the app state complexity, I am a bit worried this can cause confusion and bugs, given the same data does exist as context. Looks like there are 3 scenarios:

1 pure static global var: doc_id, user etc.
2 child component that does not change state (table pane)
3 child component that changes parent state

For 1, i think it's okay to not using context, but might be do it consistently across?
For 2, feel like using prop instead of state is cleaner
For 3, need to use context

Whichever decision, can do in later PR.

shifucun (approved):

chejennifer:
thanks for the review!

static/js/apps/eval_retrieval_generation/table_pane.tsx
shifucun: Agree. Given the complexity here, consistency is very important.

PR #4440, opened 2024-07-03
[gemma sxs] create new app

chejennifer:
- add plumbing for new sxs eval app at: /eval/retrieval_generation_sxs?sheetIdLeft=&sheetIdRight=&queryId=
- re-use eval components to display the LHS and RHS

Screenshot 2024-07-03 at 2 10 32 PM


server/routes/nl/html.py
```diff
@@ -55,3 +54,16 @@ def eval_retrieval_generation():
                             sheet_id=_TEST_SHEET_ID),
                     code=302)
   return render_template('/eval_retrieval_generation.html', sheet_id=sheet_id)
+
+
+@bp.route('/eval/retrieval_generation_sxs')
+def eval_retrieval_generation_sxs():
+  if os.environ.get('FLASK_ENV') not in ['local', 'autopush']:
+    flask.abort(404)
+  sheet_id_a = request.args.get('sheetIdA', '')
```
shifucun, line 63: since the url param is user visible, how about something like "leftSheetId", "rightSheetId"?

chejennifer: We will be randomizing which sheet is on the left and which sheet is on the right for each query .. but if you have other suggestions for url param, happy to use them

shifucun: I see. Then this is good!

shifucun (approved):

PR #4441, opened 2024-07-03
Increased fulfill stat var results limit for undata nl indexes

dwnoble:
- Increased stat vars returned from 50 to 200
- Updated check to allow for increased limit on the SDG index


pradh (approved):
Did you test out a few queries?  How does it look?

dwnoble:
Performance isn't bad - response time is about ~1.5s for larger queries. The main use case is to return all "varToTopics" mappings for the SDG index, which tops out at about ~100 variables

PR #4442, opened 2024-07-04
Updated un staging tables

dwnoble:
- Removed sdg table since it's now in covered in `schema_2024_07_03_12_17_40` and `country_2024_07_03_10_18_47`
- Updated ilo tables

### DIFF
--- a/deploy/helm_charts/envs/unsdg_staging.yaml
+++ b/deploy/helm_charts/envs/unsdg_staging.yaml
@@ -67,7 +67,7 @@ kgStoreConfig:
     project: datcom-un-staging
     instance: dc-graph
     tables:
-      - ilo_2024_06_24_09_46_00
-      - sdg_2024_06_25_10_21_43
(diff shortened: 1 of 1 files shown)


pradh (approved):

deploy/helm_charts/envs/unsdg_staging.yaml
```diff
@@ -67,7 +67,7 @@ kgStoreConfig:
     project: datcom-un-staging
     instance: dc-graph
     tables:
-      - ilo_2024_06_24_09_46_00
-      - sdg_2024_06_25_10_21_43
+      - ilo_2024_07_02_14_25_51
```
pradh, line 70: To confirm, this is assuming you will use "deploy_latest.sh" to deploy, since the latest country version may not be checked into bigtable versions yaml, right?

dwnoble: That's right

PR #4029, opened 2024-03-12
Refactor memcache module and clean up redis / file cache logic

shifucun:
Also created Redis cache for eval tool model prediction caching (not for other flask view functions).

PR #4444, opened 2024-07-04
Added script to clear autopush redis cache

dwnoble:


shifucun:
I don't think we should add any cache to autopush, otherwise it will not be "autopush"

juliawu (approved):
Approval for adding a script, but agree with @shifucun that we should discuss whether we should have caching on autopush. Does the cache clear cache each time autopush is redeployed or does that need to happen manually?

dwnoble:
Sorry missed bo's comment from last week. Looks like caching was added to autopush here: https://github.com/datacommonsorg/website/pull/4029 but it's only for NL models, not for views.

PR #4448, opened 2024-07-08
[nodejs] add highlight field to highlight tile response

chejennifer:
e.g., what is the percentage of urban population in Tamil Nadu? https://paste.googleplex.com/4946781240819712


pradh (approved):
Thanks!

PR #4451, opened 2024-07-08
Fixed css bug that was hiding the stanford custom dc navbar

dwnoble:
## Before:
Screenshot 2024-07-08 at 3 30 32 PM

## After:
Screenshot 2024-07-08 at 3 31 43 PM


juliawu (approved):
Thank you for the fix!

PR #4457, opened 2024-07-10
sv index updates

chejennifer:
- remove temperature, min/max temperature, houseless X scheduledCaste/scheduledTribe variables
- add descriptions: include "children" for Count_Person_18OrLessYears_NoHealthInsurance, include "native american" for Count_Person_AmericanIndianOrAlaskaNativeAlone
- replace Annual_Emissions_CarbonDioxide_Biogenic with Annual_Emissions_CarbonDioxide_NonBiogenic

svindex diffs: https://storage.mtls.cloud.google.com/datcom-embedding-diffs/chejennifer_base_uae_mem_2024_07_10_10_15_51.html


tools/nl/embeddings/input/base/sheets_svs.csv
```diff
@@ -3225,7 +3214,6 @@ MedianMothersAge_BirthEvent,Median Mother's Age at Birth
 Min_Humidity_RelativeHumidity,Minimum Relative Humidity
 Min_Rainfall,Minimal rainfall
 Min_Snowfall,minimal snowfall
-Min_Temperature,minimal Temperature
```
pradh, line 3228: We might need to check if any topics have this too and update those?  (e.g., dc/topic/Temperatures)

pradh (commented):
Very nice!

tools/nl/embeddings/input/base/sheets_svs.csv
chejennifer: yes it's in the topic .. which won't show up in RIG, but in regular DC it will. Did we want the variables completely gone? & is it just the min and max we want completely gone or maybe we can just remove the topic because it has min/max/mean & child topics (child topics are already in the index)

pradh: Yeah, I think we wanted the variables taken out for now (because these imports are weird), so that's independent of RIG.

chejennifer: done

diffs (looks ok to me): https://storage.mtls.cloud.google.com/datcom-embedding-diffs/chejennifer_base_uae_mem_2024_07_11_08_45_22.html

pradh (approved):
Thanks!

PR #4458, opened 2024-07-10
[SxS Eval] Update how left/right are determined and add baseline eval type 

hqpho:
Also:
- Make all inputs to left/right picker available via context
- Clean up some unused stuff
- Factor out duplicated template section into its own component


static/js/apps/eval_retrieval_generation/sxs/left_right_picker.ts
```diff
@@ -0,0 +1,25 @@
+import sdbm from "sdbm";
```
chejennifer, line 1: add the copyright comment

nit: maybe can just make this a general utils.ts file, so we can later add more util functions?

static/js/apps/eval_retrieval_generation/sxs/left_right_picker.ts
```diff
@@ -0,0 +1,25 @@
...
+
+import { DocInfo } from "../types";
+
+/**
+ * Result is stable based on inputs.
```
chejennifer, line 6: generally the function comment should describe what the function is doing. So maybe also add something like "Gets the docInfo to show in the left and right panes."

static/js/apps/eval_retrieval_generation/sxs/query_with_tables.tsx
```diff
@@ -0,0 +1,33 @@
+import React, { useContext } from "react";
```
chejennifer, line 1: add copyright comment & a file level comment describing the component

static/js/apps/eval_retrieval_generation/sxs/app.tsx
```diff
@@ -84,41 +84,21 @@ export function App(props: AppPropType): JSX.Element {
...
-                query={docInfo.left.allQuery[queryId]}
-              />
-              {docInfo.left.evalType === EvalType.RAG && (
-                <TablePane
-                  doc={docInfo.left.doc}
-                  calls={docInfo.left.allCall[queryId]}
-                />
-              )}
-            </div>
+            <QueryWithTables docInfo={left}></QueryWithTables>
```
chejennifer, line 97: nit: can just do `<QueryWithTables docInfo={left} />`

static/js/apps/eval_retrieval_generation/sxs/app.tsx
```diff
@@ -84,41 +84,21 @@ export function App(props: AppPropType): JSX.Element {
       )}
 
       {user && <p>Signed in as {user.email}</p>}
-      {user && !docInfo && <p>Loading query...</p>}
-      {docInfo && (
-        <>
+      {user && !docInfos && <p>Loading query...</p>}
+      {docInfos && (
+        <AppContext.Provider
```
chejennifer, line 89: this app context doesn't seem necessary since none of the child components are actually accessing this context?

static/js/apps/eval_retrieval_generation/sxs/app.tsx
```diff
@@ -32,11 +33,14 @@ interface AppPropType {
 
 export function App(props: AppPropType): JSX.Element {
   const [user, setUser] = useState<User | null>(null);
-  const [docInfo, setDocInfo] = useState<{ left: DocInfo; right: DocInfo }>(
-    null
+  const [docInfos, setDocInfos] = useState<{ a: DocInfo; b: DocInfo }>(null);
+  const { setSessionQueryId, sessionQueryId } = useContext(SessionContext);
+  const { left, right } = getLeftAndRight(
```
chejennifer, line 38: nit: can we make the variable naming a little more informational, maybe leftDocInfo/rightDocInfo

chejennifer (commented):
nice!

static/js/apps/eval_retrieval_generation/sxs/left_right_picker.ts
hqpho: Added copyright comment. Prefer not to make a general utils file based on principles in go/no-utils.

static/js/apps/eval_retrieval_generation/sxs/left_right_picker.ts
hqpho: Revised comment.

static/js/apps/eval_retrieval_generation/sxs/query_with_tables.tsx
hqpho: Done.

static/js/apps/eval_retrieval_generation/sxs/app.tsx
hqpho: Done, thanks!

static/js/apps/eval_retrieval_generation/sxs/app.tsx
hqpho: Hm yeah I thought it might be necessary in the future but now that I've started working on the feedback component I agree it can be removed.

static/js/apps/eval_retrieval_generation/sxs/app.tsx
hqpho: Done.

chejennifer (approved):
thanks for the updates!

static/js/apps/eval_retrieval_generation/sxs/left_right_picker.ts
chejennifer: oh, never seen go/no-utils, but good to know!!

static/js/apps/eval_retrieval_generation/sxs/left_right_picker.ts
```diff
@@ -0,0 +1,43 @@
...
+  docInfoB: DocInfo | null,
+  queryId: number
+): { leftDocInfo: DocInfo; rightDocInfo: DocInfo } {
+  if (!docInfoA || !docInfoB) {
+    return { leftDocInfo: docInfoA, rightDocInfo: docInfoB };
+  }
+  const inputs =
+    `${docInfoA.doc.spreadsheetId} ${docInfoB.doc.spreadsheetId}` +
+    ` ${sessionId} ${queryId}`;
+  if (sdbm(inputs) % 2 === 0) {
```
pradh, line 38: I was imagining we'd have a prng seeded with sdbm(inputs).  So raters don't get used to expecting every alternate result to be one way.  But we can start with this, sure...

static/js/apps/eval_retrieval_generation/sxs/left_right_picker.ts
pradh: Ok for this PR.  But you'll find that we have utils. My feeling is, it is sometimes ok, since naming is hard and we also don't want to have a ton of files.

And there's go/no-no-utils, so I'd think of these as opinions, sort of.

pradh (approved):

static/js/apps/eval_retrieval_generation/sxs/left_right_picker.ts
hqpho: Naively I thought I could use sdbm as a sort of prng, but after some manual testing it gives similar results for similar inputs so will layer in an actual prng as well in a follow-up!

static/js/apps/eval_retrieval_generation/sxs/left_right_picker.ts
hqpho: TIL about go/no-no-utils! and also the MOGHOUC: "massive organically grown hairball of unrelated code"

PR #4459, opened 2024-07-10
Add step in GKE setup for granting VertexAI permissions 

chejennifer:


chejennifer:
clsoing because script to do this added in https://github.com/datacommonsorg/website/pull/4479

PR #4463, opened 2024-07-11
Fixed csv download and show metadata link in local and custom DC to use local API path

dwnoble:
Before:
Screenshot 2024-07-10 at 5 17 34 PM

Screenshot 2024-07-10 at 5 17 24 PM

After:
Screenshot 2024-07-10 at 5 16 27 PM
Screenshot 2024-07-10 at 5 16 38 PM


static/js/utils/data_commons_client.ts
```diff
@@ -31,3 +31,12 @@ export const defaultDataCommonsWebClient = new DataCommonsWebClient({
 export const defaultDataCommonsClient = new DataCommonsClient({
   apiRoot: DEFAULT_CLIENT_API_ROOT,
 });
+
```
juliawu, line 34: supernit: Could you add a quick docstring for this exported function? Would be helpful to clarify why we need this function and can't just use `new DataCommonsClient({apiRoot})` directly.

juliawu (approved):
Thanks for the fix!

static/js/utils/data_commons_client.ts
dwnoble: done!

keyurva (approved):
Thanks Dan!

PR #4464, opened 2024-07-11
Make base/sheets_svs.csv a proper csv

pradh:
So it can be viewed as a [table on github](https://github.com/datacommonsorg/website/assets/4375037/6f355c58-89da-4708-99dc-eb1061ef771c) and loaded into BQ, etc.


chejennifer (approved):
thanks!

PR #4465, opened 2024-07-11
Move custom DC tests out of server/webdriver/tests

hqpho:
Otherwise they get run as part of ./run_test.sh -w


dwnoble (approved):
Thanks Hannah!

PR #4467, opened 2024-07-11
Load custom dc catalog at startup.

keyurva:
* Saves custom catalog when building embeddings in `build_embeddings.py`. ([example](https://screenshot.googleplex.com/AJBmDYSE5GXtNPx))
* Loads the saved catalog (if any) at server startup.
* Initializes build embeddings logs so they are captured by the web admin. (similar to [this](https://github.com/datacommonsorg/import/blob/69b0e58ecef656e4706e67b4a47104eec4cf3138/simple/stats/main.py#L54-L60))

### DIFF
--- a/nl_server/config_reader.py
+++ b/nl_server/config_reader.py
@@ -237,3 +237,35 @@ def get_server_config(catalog: Catalog, env: Env) -> ServerConfig:
   )
   _log_asdict(server_config, 'server config')
   return server_config
+
+
+def maybe_load_custom_catalog() -> Dict:
+  """
+  Loads the custom DC catalog and returns it as a dict if running in custom DC mode and if it exists.
+  Returns None otherwise.
+  """
+  if not custom_dc_util.is_custom_dc():
+    return None
+
+  custom_catalog_path = custom_dc_util.get_custom_catalog_path()
+  if not custom_catalog_path:
+    return None
+
+  if gcs.is_gcs_path(custom_catalog_path):
+    local_path = gcs.maybe_download(custom_catalog_path)
+    if not local_path:
+      logging.info("Custom catalog not found and will not be loaded: %s",
+                   custom_catalog_path)
+      return None
+  else:
+    local_path = custom_catalog_path
+    if not os.path.exists(custom_catalog_path):
+      logging.info("Custom catalog not found and will not be loaded: %s",
+                   custom_catalog_path)
+      return None
+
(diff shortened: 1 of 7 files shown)


tools/nl/embeddings/build_embeddings.py
```diff
@@ -32,19 +36,33 @@
                     'A dict of user provided embeddings/model catalog')
 
 
+def _init_logger():
```
shifucun, line 39: this should be set in admin/html.py instead of here?

tools/nl/embeddings/build_embeddings.py
```diff
@@ -71,13 +89,22 @@ def main(_):
...
+  logging.info("Saved embeddings.")
 
   # Save index config
   utils.save_index_config(fm, index_config)
+  logging.info("Saved index config.")
+
+  # Save custom catalog
+  if custom_catalog:
+    utils.save_custom_catalog(fm, custom_catalog)
+    logging.info("Saved custom catalog.")
```
shifucun, line 101: I don't think this tools should save the catalog, this is the input?

tools/nl/embeddings/file_manager.py
```diff
@@ -65,12 +67,17 @@ def preindex_csv_path(self):
   def index_config_path(self):
     return os.path.join(self._local_output_dir, _INDEX_CONFIG_YAML)
 
+  def custom_catalog_path(self):
+    return os.path.join(self._local_output_dir, _CUSTOM_CATALOG_YAML)
+
```
shifucun, line 72: I don't think we should have any custom dc specific logic in this file

nl_server/config_reader.py
```diff
@@ -237,3 +237,35 @@ def get_server_config(catalog: Catalog, env: Env) -> ServerConfig:
   )
   _log_asdict(server_config, 'server config')
   return server_config
+
+
+def maybe_load_custom_catalog() -> Dict:
```
shifucun, line 242: we should have any custom dc specific logic in here

tools/nl/embeddings/build_embeddings.py
keyurva: That does not work because admin/html.py forks build embeddings in the new process.

tools/nl/embeddings/build_embeddings.py
keyurva: For the server to be able to load at startup, it needs to be persisted somewhere. Do you have an alternative in mind?

PR #4468, opened 2024-07-11
[Eval SxS] Feedback footer with navigation; Firestore writes

hqpho:
- Add a feedback component and style it as a footer.
- Add previous/next buttons that navigate between queries.
- Write to Firestore when navigating previous/next. Just write static values for now.

Also fix randomization of left/right.


static/js/apps/eval_retrieval_generation/sxs/left_right_picker.ts
```diff
@@ -35,7 +37,8 @@ export function getLeftAndRight(
   const inputs =
     `${docInfoA.doc.spreadsheetId} ${docInfoB.doc.spreadsheetId}` +
     ` ${sessionId} ${queryId}`;
-  if (sdbm(inputs) % 2 === 0) {
+  const seededRand = new Srand(sdbm(md5(inputs)));
+  if (Math.floor(seededRand.random() * 2) === 0) {
```
pradh, line 41: Add a comment why we do this?

Would it be simpler to use `intInRange` method?

Also, wonder if there was a specific reason you chose this vs. others (like say https://npm.io/package/pure-rand released more recently and has more downloads)?

static/css/eval_retrieval_generation.scss
```diff
@@ -462,7 +477,7 @@ label {
   display: flex;
   flex-direction: column;
   width: 50%;
-  height: 100%;
+  height: 70%;
```
chejennifer, line 480: can you check how this looks with the original RIG/RAG eval tools

pradh (commented):
Thanks!

static/js/apps/eval_retrieval_generation/sxs/app.tsx
```diff
@@ -24,23 +24,38 @@ import { getDocInfo } from "../util";
...
+import { SxsFeedback } from "./sxs_feedback";
 
 interface AppPropType {
   sessionId: string;
   sheetIdA: string;
   sheetIdB: string;
 }
 
+function getSortedQueryIds(docInfos: { a: DocInfo; b: DocInfo }) {
+  const idsA = Object.keys(docInfos?.a?.allQuery || []);
```
chejennifer, line 36: should be `|| {}` since allQuery is an object?

static/js/apps/eval_retrieval_generation/sxs/app.tsx
```diff
@@ -24,23 +24,38 @@ import { getDocInfo } from "../util";
...
+    .sort((a, b) => a - b);
+}
+
 export function App(props: AppPropType): JSX.Element {
   const [user, setUser] = useState<User | null>(null);
   const [docInfos, setDocInfos] = useState<{ a: DocInfo; b: DocInfo }>(null);
   const { setSessionQueryId, sessionQueryId } = useContext(SessionContext);
+  const sortedQueryIds = getSortedQueryIds(docInfos);
+  if (!sessionQueryId) {
+    setSessionQueryId(sortedQueryIds[0]);
```
chejennifer, line 50: check that sortedQueryIds is not empty first before accessing the first item?

static/js/apps/eval_retrieval_generation/sxs/navigation.tsx
```diff
@@ -0,0 +1,145 @@
...
+  PREV_QUERY = "PREV_QUERY",
+  NEXT_QUERY = "NEXT_QUERY",
+  FINISH = "FINISH",
+  EMPTY = "EMPTY",
+}
+
+interface ButtonConfig {
+  text: string;
+  themeClassName: string;
+  icon?: string;
```
chejennifer, line 35: looks like none of the button configs set this field so can get rid of it?

static/js/apps/eval_retrieval_generation/sxs/navigation.tsx
```diff
@@ -0,0 +1,145 @@
...
+    <div className="feedback-nav-section">
+      <div className="nav-buttons">
+        {prevButtonType !== ButtonType.EMPTY && (
+          <Button
+            onClick={getOnClickFunction(prevButtonType)}
+            className={BUTTON_CONFIGS[prevButtonType].themeClassName}
+          >
+            <div>
+              {BUTTON_CONFIGS[prevButtonType].icon && (
+                <span className="material-icons-outlined">
```
chejennifer, line 119: doesn't look like there are any icons being used?

static/js/apps/eval_retrieval_generation/sxs/data_store.tsx
```diff
@@ -0,0 +1,78 @@
...
+    "sheets",
+    `${firstSheetId}:${secondSheetId}`,
+    "queries",
+    queryId.toString(),
+    "ratings",
+    String(sessionId)
+  );
+}
+
+/**
```
chejennifer, line 46: wonder if these 2 functions are even necessary because we can just:

- update getRatingRef to getRatingRefPath
- use getAllFields and setFields from here: https://github.com/datacommonsorg/website/blob/master/static/js/apps/eval_retrieval_generation/data_store.tsx#L63-L77

chejennifer (commented):
nice

static/css/eval_retrieval_generation.scss
hqpho: I think .sxs-pane only gets used in SxS tool.

static/js/apps/eval_retrieval_generation/sxs/data_store.tsx
hqpho: I forked them to avoid using `{merge: true}` and also because I worry it might be unexpected for SxS data store methods to rely on regular eval tool data store methods, since some regular store methods make assumptions about data being for regular eval tool (e.g. saveToStore assumes a regular eval path).

static/js/apps/eval_retrieval_generation/sxs/left_right_picker.ts
hqpho: I just used what I could find, happy to switch to a package with better support! Will add a comment too.

static/js/apps/eval_retrieval_generation/sxs/app.tsx
hqpho: ah true, changed!

static/js/apps/eval_retrieval_generation/sxs/navigation.tsx
hqpho: done

static/js/apps/eval_retrieval_generation/sxs/app.tsx
hqpho: done

static/css/eval_retrieval_generation.scss
chejennifer: ah, yes, did not see this was under sxs

static/js/apps/eval_retrieval_generation/sxs/data_store.tsx
chejennifer: ok makes sense!

chejennifer (approved):
nice!

PR #4469, opened 2024-07-12
Switch from "fork" to "spawn" for multiprocessing in tests

pradh:
Some kind of mysterious cloud build platform change broke "fork" start method for python `multiprocessing`.  Generally, "spawn" is apparently a safer alternative, since it doesn't inherit parent state into child (sharing state can mess with locks, and mess up behavior, perhaps like we see now) -- [doc](https://docs.python.org/3/library/multiprocessing.html).

However, "spawn" is [finicky about pickling of methods and objects](https://stackoverflow.com/questions/72766345/attributeerror-cant-pickle-local-object-in-multiprocessing) that get passed to the `run()` method.

So, this change does three things to switch over to "spawn" mode:
1.  Fix the `multiprocessing.Process` call we make to start NL server in a way that makes spawn happy
2. The Web server gets started by an external library (`flask_testing`) in a way that makes spawn unhappy.  So, stop using that (generally unmaintained) library, and instead start web server in our code.
3. Given our support for spawn, we no longer set fork, so avoid the mac-os specific setting.

### DIFF
--- a/server/integration_tests/__init__.py
+++ b/server/integration_tests/__init__.py
@@ -12,6 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from shared.lib.test_setup import set_up_macos_for_tests
+from shared.lib.test_setup import set_up_multiprocessing_for_tests
 
-set_up_macos_for_tests()
+set_up_multiprocessing_for_tests()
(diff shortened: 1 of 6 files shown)


pradh:
This still makes Web Driver unhappy!

PR #4472, opened 2024-07-12
[Eval SxS] Actually get and set ratings; Improve layout

hqpho:
- Add radio buttons for preference and a text area for reason
- Fetch and show existing ratings, correcting left/right orientation if necessary
- Save rating when navigating if it has been updated
- Make panes scroll independently
- Display query ID and question only once
- Make footer take up only as much room as it should

[Screencast](https://github.com/user-attachments/assets/1f7cb506-eacf-408a-8219-84b9be59165c)


static/js/apps/eval_retrieval_generation/sxs/query_id_and_question.tsx
```diff
@@ -0,0 +1,45 @@
...
+import { QuerySection } from "../query_section";
+import { TablePane } from "../table_pane";
+import { DocInfo, EvalType, FeedbackStage } from "../types";
+import { SessionContext } from "./context";
+
+interface QueryIdAndQuestionPropType {
+  docInfo: DocInfo;
+}
+
+export function QueryIdAndQuestion(
```
chejennifer, line 32: there isn't really any logic or rendering logic, does this need to be its own component

static/js/apps/eval_retrieval_generation/sxs/data_store.tsx
```diff
@@ -43,6 +44,38 @@ function getRatingRef(
   );
 }
 
+/**
+ * Saves a rater's input to Firestore only if it is different from any existing
+ * stored rating. Overwrites any previous data.
+ */
+export async function saveRatingIfChanged(
```
chejennifer, line 51: I don't think this function is necessary and we don't need the extra read because from the component that calls this function: 
- we've already retrieved the stored rating beforehand (can save that in the state or have a bit that says whether rating has changed or something)
- then from the component can call saveRatingToStore only if there is a change

alternatively can just always write to the document since it seems like you're just saving the most recent version anyways so it doesn't matter if it's equivalent or not?

static/js/apps/eval_retrieval_generation/sxs/rating_util.ts
```diff
@@ -0,0 +1,62 @@
...
+    case SxsPreference.RIGHT:
+      return SxsPreference.LEFT;
+    case SxsPreference.NEUTRAL:
+      return SxsPreference.NEUTRAL;
+    default:
+      throw "All preference values should be handled";
+  }
+}
+
+function equal(r1: Rating | null, r2: Rating | null) {
```
chejennifer, line 52: can just use `_.isEqual` (https://lodash.com/docs/#isEqual)?

static/js/apps/eval_retrieval_generation/sxs/rating_util.ts
```diff
@@ -0,0 +1,62 @@
...
+  if (!rating) return null;
+  return {
+    leftSheetId: rating.rightSheetId,
+    preference: getReversed(rating.preference),
+    reason: rating.reason,
+    rightSheetId: rating.leftSheetId,
+  };
+}
+
+function getReversed(preference: SxsPreference) {
```
chejennifer, line 39: include return type?

chejennifer, line 39: nit: hard to tell the difference between getReversed and getFlipped by just reading function names. Maybe `getReversedRating` and `geReversedPreference`

chejennifer (commented):
nice

static/js/apps/eval_retrieval_generation/sxs/query_id_and_question.tsx
hqpho: no, it was one of those situations where by the time i realized that I had already made it its own component and i wasn't sure it was worth the effort to undo it ¯\_(ツ)_/¯

static/js/apps/eval_retrieval_generation/sxs/rating_util.ts
hqpho: done (Codacy wanted it too)

hqpho: Renamed

chejennifer (approved):
thanks for updating!

PR #4473, opened 2024-07-12
Load custom dc embeddings at server startup.

keyurva:
* `custom_catalog.yaml` will be created by the simple importer with https://github.com/datacommonsorg/import/pull/332
  + This website PR will only be submitted after the importer one has been submitted and the module is merged with website.
* The yaml generated above is accepted as an additional catalog path by both the build embeddings tool and the nl server at startup.
* The build embeddings tool generates embeddings based on this yaml.
* The nl server downloads embeddings based on this yaml at startup.


nl_server/flask.py
```diff
@@ -51,7 +53,8 @@ def create_app():
   try:
     # Build the registry before creating the Flask app to make sure all resources
     # are loaded.
-    reg = registry.build()
+    additional_catalog_path = os.environ.get(constants.ADDITIONAL_CATALOG_PATH)
```
pradh, line 56: If there's no other use, inline constant?

tools/nl/embeddings/build_embeddings.py
```diff
@@ -31,19 +36,44 @@
 flags.DEFINE_string('catalog', '',
                     'A dict of user provided embeddings/model catalog')
```
pradh, line 37: Can we deprecate the --catalog?

nl_server/config_reader.py
```diff
@@ -94,13 +96,26 @@ def merge_vertex_ai_configs(m: ModelConfig,
...
-    if os.path.exists(p):
+
+    # If gcs path, download it locally.
+    if gcs.is_gcs_path(p):
+      gcs_path = p
+      logging.info('Loading catalog from gcs path: %s', gcs_path)
+      p = gcs.maybe_download(gcs_path)
+      if not p:
+        logging.info('GCS catalog path not found and will be skipped: %s',
+                     gcs_path)
```
pradh, line 116: I wonder if should we have this gcs.maybe_download in "admin.py"?  So it downloads locally and NL server or build_embeddings need not?

tools/nl/embeddings/build_embeddings.py
```diff
@@ -31,19 +36,44 @@
...
+
 
 def main(_):
+  _init_logger()
+
   # Get embeddings_name
   embeddings_name = FLAGS.embeddings_name
   output_dir = FLAGS.output_dir
+
+  # If additional catalog path is provided and embeddings name is not specified, extract the embeddings name from the catalog.
```
pradh, line 61: Can this too happen upstream in "admin.py".  So the interface of this tool is:  catalog_path and embeddings_name.

Specifically, the decision to pick the first index-name in the additional-catalog-file is a very custom-dc specific decision that's not generally applicable?

tools/nl/embeddings/utils.py
```diff
@@ -223,3 +223,14 @@ def save_embeddings_lancedb(local_dir: str, embeddings: List[Embedding]):
 def save_index_config(fm: FileManager, index_config: IndexConfig):
   with open(fm.index_config_path(), 'w') as f:
     yaml.dump(asdict(index_config), f)
+
+
+def get_additional_catalog_embeddings_name(additional_catalog_path: str) -> str:
```
pradh, line 228: hmm, what is additional_catalog specific about this function?

should it be:  `get_first_index_name()`.   Also, this should probably be in admin.py?

nl_server/config_reader.py
keyurva: It would need to be downloaded by nl server at startup.

nl_server/flask.py
keyurva: Done.

tools/nl/embeddings/build_embeddings.py
keyurva: Removed.

tools/nl/embeddings/build_embeddings.py
keyurva: The separation would be nice. However, once we have data docker, admin/html.py will be deprecated and the tool will be called by a shell script in docker.

tools/nl/embeddings/utils.py
keyurva: Renamed.

See other comment on admin/html.py.

nl_server/config_reader.py
```diff
@@ -94,13 +96,26 @@ def merge_vertex_ai_configs(m: ModelConfig,
...
     all_paths = []
-    if os.path.exists(p):
+
+    # If gcs path, download it locally.
+    if gcs.is_gcs_path(p):
+      gcs_path = p
+      logging.info('Loading catalog from gcs path: %s', gcs_path)
+      p = gcs.maybe_download(gcs_path)
+      if not p:
+        logging.info('GCS catalog path not found and will be skipped: %s',
```
pradh, line 115: nit: error ?

tools/nl/embeddings/build_embeddings.py
pradh: Hmm, if it will have this fixed name - https://github.com/datacommonsorg/import/pull/332/files#diff-5ea0ff46203687a6da5298b11caf748eeb36522e1ee79ec018d18bdee6693369R25 - wonder if the caller (html.py or that script) should pass that in?  Anyway the file there is constants shared with this repo.

keyurva: That's a good point. Updated code to pass embeddings name from admin (and in the future from docker). Also removed code that extracted the first name from the catalog.

nl_server/config_reader.py
keyurva: Done.

tools/nl/embeddings/build_embeddings.py
pradh: Much better, thanks!

pradh (approved):

keyurva:
Thanks for the review.

PR #4475, opened 2024-07-12
[toolformer] replace "global" with "world" in queries in toolformer mode

chejennifer:
ran goldens without the `if (params.is_toolformer_mode(dargs.mode))` & there were no diffs


pradh (approved):
Thanks!

PR #4476, opened 2024-07-12
add women in parliament to nl index

chejennifer:
no diffs: https://storage.mtls.cloud.google.com/datcom-embedding-diffs/chejennifer_base_uae_mem_2024_07_12_16_30_34.html


tools/nl/embeddings/input/base/sheets_svs.csv
```diff
@@ -3602,3 +3602,5 @@ worldBank/4_1_SHARE_RE_IN_ELECTRICITY,Renewable electricity share of total elect
 worldBank/EG_ELC_ACCS_RU_ZS,percentage of rural population with access to electricity
 worldBank/EG_ELC_ACCS_UR_ZS,percentage of urban population with access to electricity
 worldBank/EG_ELC_ACCS_ZS,percentage of population with access to electricity
+sdg/SG_GEN_PARL.SEX--F,proportion of seats held by women in national parliaments
+sdg/SG_GEN_PARLN.SEX--F,number of seats held by women in national parliaments
```
pradh, line 3606: nit:  I wonder if we should skip number?  the pc denominator is not valid, and we have some risk of number coming in the way of proportion result.

pradh (approved):
Thanks!

tools/nl/embeddings/input/base/sheets_svs.csv
chejennifer: that makes sense, updated & no new diffs

PR #4478, opened 2024-07-15
update goldens after topic cache update

chejennifer:
commute time topic updated with mean commute time replacing total commute time


pradh (approved):
Thanks!

PR #4481, opened 2024-07-15
Add CDC SV explorer sanity test.

keyurva:
* Added a simple sanity test for the SV explorer page.
* Updated the home page test to look for the homepage element.


server/webdriver/cdc_tests/autopush_test.py
```diff
@@ -40,14 +44,24 @@ def tearDown(self):
...
+
+    svg_container_elem = find_elem(self.driver, By.CLASS_NAME,
+                                   'stat-var-hierarchy-container')
+    self.assertIsNotNone(svg_container_elem, 'SVG container element not found')
+
+    custom_svg_elem = wait_elem(svg_container_elem, By.XPATH,
+                                '//*[contains(text(), "OECD")]')
+    self.assertIsNotNone(custom_svg_elem, 'Custom SVG (OECD) element not found')
+
+    custom_svg_elem.click()
```
hqpho, line 63: Optional: Add a comment either inline here or on the function def that sums up the test, e.g. `"""Tests that clicking a custom SVG in the SV explorer works as expected"""`.

hqpho (approved):

server/webdriver/cdc_tests/autopush_test.py
keyurva: Done.

server/webdriver/cdc_tests/cdc_tests_utils.py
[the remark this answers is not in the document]
keyurva: That's a good point. Added them to `base_utils` since we use that file for `create_driver` already.

PR #4483, opened 2024-07-15
Show DC stats in tooltips for RIG

hqpho:
For RIG answers:
  - Don't show inline DC stats. Instead show DC stat with label in a tooltip when an LLM stat is hovered over.
  - When an LLM stat in SxS UI doesn't have an associated DC stat, don't highlight it or show a tooltip. (For regular eval UI, still highlight but don't add a tooltip.)
 
Other changes:
 - Change "Why?" free-text form field label to "Comments (optional)"
 - Show "Loading answer..." when fetching answers, so answers are never out of sync with the feedback form.
 - Fetch all call data when initially loading each spreadsheet
 - Don't limit some sheet calls by column (fetch more data including possibly extraneous data in exchange for making fewer fetches)
 - Minor naming updates for readability

[Screencast](https://github.com/user-attachments/assets/a7ca0359-ee0c-40ff-99eb-b332cae61bff)


chejennifer (commented):
nice, code looks good, but can we try out the tooltip content changes we talked about in the chat?

hqpho:
> nice, code looks good, but can we try out the tooltip content changes we talked about in the chat?

yep, I'm working on them; they're non-trivial since the detailed call info isn't in DocInfo. Thanks for the preliminary review!

static/js/apps/eval_retrieval_generation/sxs/answer_with_tables.tsx
```diff
@@ -31,6 +31,32 @@ interface AnswerWithTablesPropType {
 export function AnswerWithTables(props: AnswerWithTablesPropType): JSX.Element {
   const { sessionQueryId } = useContext(SessionContext);
 
+  const onAnswerChange = () => {
```
chejennifer, line 34: since we want the same tooltips for the regular RIG eval tool too, I think it'll be easier to just update `processText` util function to take text AND an optional allCalls object & just populate the tooltip there & when doing this, can probably make getting the tooltip html a helper method for readability

static/js/apps/eval_retrieval_generation/util.ts
```diff
@@ -206,13 +201,21 @@ function getEvalType(
...
-    return Promise.all([
-      getQueries(doc, allHeader),
-      getCalls(doc, allHeader),
-      getEvalType(doc, allHeader),
-    ]).then(([allQuery, allCall, evalType]) => {
+  return getHeader(doc)
+    .then((allHeader) => {
+      return Promise.all([
+        Promise.resolve(allHeader),
+        getEvalType(doc, allHeader),
```
chejennifer, line 208: curious what the point of doing a separate promise.all here is when we getQueries and getCalls don't need to wait for evalType?

chejennifer (commented):
nice!

static/js/apps/eval_retrieval_generation/types.tsx
```diff
@@ -33,11 +33,19 @@ export interface Query {
...
 }
 
-// Key is the call id, value is the row index in the sheet.
-export type DcCall = Record<number, number>;
+export interface DcCallInfo {
+  rowIndex: number;
+  question: string;
+  llmStat: string;
+  dcResponse: string;
+  dcStat: string;
```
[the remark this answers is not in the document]
hqpho: I compared the sheet you shared with a copy of the sheet with all but 5 queries and their associated calls removed, and I don't notice any lag! Chrome task manager shows reasonable tab memory usage too.

static/js/apps/eval_retrieval_generation/sxs/answer_with_tables.tsx
hqpho: Right, forgot we wanted the tooltips for regular eval too! Updated both this and the CSS to make tooltips work for both sxs and regular. [Demo](https://github.com/user-attachments/assets/10cac027-2b58-4fa3-8c97-72405c0fb05c)

static/js/apps/eval_retrieval_generation/util.ts
hqpho: ah, this is a holdover from when I was passing evalType into getCalls! revised.

static/js/apps/eval_retrieval_generation/types.tsx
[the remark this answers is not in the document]
chejennifer: nice!

static/js/apps/eval_retrieval_generation/util.ts
```diff
@@ -78,6 +91,20 @@ export const processText = (text: string): string => {
   );
 };
 
+function getTooltipHtml(
+  dcStat: string,
+  callId: string,
+  calls?: DcCalls
+): string {
+  const dcResponse: string = calls ? calls[callId]?.dcResponse ?? "" : "";
```
chejennifer, line 99: nit: instead of trying to write all of this in one line with lots of ? and ??, probably more readable with: 

```
let dcResponse: string = "";
if (calls) {
    dcResponse = calls[callId]?.dcResponse || "";
}
```

chejennifer (approved):
thanks for the updates!

static/js/apps/eval_retrieval_generation/util.ts
hqpho: done!

PR #4485, opened 2024-07-15
Add Custom DC NL sanity test.

keyurva:
* Adds a simple sanity test that asserts that the detect-and-fulfill response includes a Custom DC SV.
* This PR will only be submitted after this one: https://github.com/datacommonsorg/website/pull/4473
  + Without that PR, custom embeddings will not load on startup, which will cause this test to fail.
  + Ironically, in the future, the failure of this test will be an indicator of a Custom DC NL regression.

### DIFF
--- a/server/webdriver/cdc_tests/autopush_test.py
+++ b/server/webdriver/cdc_tests/autopush_test.py
@@ -13,12 +13,10 @@
 # limitations under the License.
 
 import unittest
-import urllib
-import urllib.request
 
-import requests
 from selenium.webdriver.common.by import By
 
+from server.integration_tests.explore_test import ExploreTest
 from server.webdriver import shared
 from server.webdriver.base_utils import create_driver
 from server.webdriver.base_utils import find_elem
@@ -69,15 +67,10 @@ class CdcAutopushWebdriverTest(unittest.TestCase):
                          'Custom SV (Average Annual Wage) element not found')
 
 
-class CdcAutopushNLTest(unittest.TestCase):
+class CdcAutopushNLTest(ExploreTest):
 
-  def test_detect_and_fulfill(self):
-    """Tests that the detect-and-fulfill endpoint returns a Custom DC SV in the response."""
-    query = "Average annual wages in Europe"
-    url = f'{CDC_AUTOPUSH_URL}/api/explore/detect-and-fulfill?q={query}'
-    response = requests.post(url, json={
-        'contextHistory': [],
-        'dc': '',
-    }).json()
-    svs = response.get('context', [{}])[0].get('svs', [])
-    assert 'average_annual_wage' in svs, 'Custom SV (average_annual_wage) not in detect-and-fulfill response'
+  def get_server_url(self):
+    return CDC_AUTOPUSH_URL
+
+  def test_cdc_nl(self):
+    self.run_detect_and_fulfill('cdc_nl', ['gender wage gap in europe'])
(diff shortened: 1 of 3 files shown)


server/webdriver/cdc_tests/autopush_test.py
```diff
@@ -27,7 +28,7 @@
 CDC_AUTOPUSH_URL = 'https://dc-autopush-kqb7thiuka-uc.a.run.app'
 
 
-class CdcAutopushTest(unittest.TestCase):
+class CdcAutopushWebdriverTest(unittest.TestCase):
```
keyurva, line 31: @hqpho - With this PR, CDC tests now include non-webdriver tests as well. Wondering if we should still keep them in a `webdriver` folder or move them out of it?

server/webdriver/cdc_tests/autopush_test.py
```diff
@@ -66,3 +67,17 @@ def test_statvar_explorer(self):
...
+  def test_detect_and_fulfill(self):
+    """Tests that the detect-and-fulfill endpoint returns a Custom DC SV in the response."""
+    query = "Average annual wages in Europe"
+    url = f'{CDC_AUTOPUSH_URL}/api/explore/detect-and-fulfill?q={query}'
+    response = requests.post(url, json={
+        'contextHistory': [],
+        'dc': '',
+    }).json()
+    svs = response.get('context', [{}])[0].get('svs', [])
+    assert 'average_annual_wage' in svs, 'Custom SV (average_annual_wage) not in detect-and-fulfill response'
```
keyurva, line 83: @pradh - this only tests that a custom SV was returned and is not testing NL functionality exhaustively like `explore_test`. Should it? Also, any other tests that may be useful as a NL sanity check?

pradh: I think it should, and I feel we should use the exact same tests, albeit with a different URL?

server/webdriver/cdc_tests/autopush_test.py
hqpho: I see other NL tests are in server/integration_tests. Maybe it makes sense for all these tests to be in integration_tests? IMO that's a decent description of them!

server/webdriver/cdc_tests/autopush_test.py
keyurva: SG. Done.

server/webdriver/cdc_tests/autopush_test.py
keyurva: That works. Let's do this in a follow-on PR.

server/webdriver/cdc_tests/autopush_test.py
```diff
@@ -66,3 +65,12 @@ def test_statvar_explorer(self):
...
                          'Custom SV (Average Annual Wage) element not found')
+
+
+class CdcAutopushNLTest(ExploreTest):
+
+  def get_server_url(self):
+    return CDC_AUTOPUSH_URL
+
+  def test_cdc_nl(self):
+    self.run_detect_and_fulfill('cdc_nl', ['gender wage gap in europe'])
```
pradh, line 76: This is very nice!

As a followup in a future PR, if we can share the test-case queries, that would be wonderful.

pradh (approved):

keyurva:
Thanks for the review.

PR #4487, opened 2024-07-16
Allow subset of pattern exclusions from heuristics

pradh:
Currently, words like "youngest" etc are added to stop-words.  Add a set of exception patterns per heuristic type.  Start off with toolformer-only.


chejennifer (approved):
nice!

PR #4493, opened 2024-07-17
[timeline] fix ratio bug

chejennifer:
there was a bug in picking the denominator observation where when comparing if next denom is better, it was comparing with the earliest denominator & not the previous denominator. This caused weird behavior for per capita timelines

autopush: https://screenshot.googleplex.com/6gh9Nyjr7kg638C
local: https://screenshot.googleplex.com/BHUsQPMih5X27Hy

### DIFF
--- a/static/js/tools/shared_util.ts
+++ b/static/js/tools/shared_util.ts
@@ -185,13 +185,14 @@ export function computeRatio(
   let j = 0; // denominator position
   for (let i = 0; i < num.length; i++) {
     const numDate = Date.parse(num[i].date);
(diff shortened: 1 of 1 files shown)


dwnoble (approved):
Good find, thanks Jennifer!

pradh:
Nice find!  Does the timeline in the local screenshot still go above 100%?

chejennifer:
> Nice find! Does the timeline in the local screenshot still go above 100%?

Nope, does not go above 100%!

PR #4494, opened 2024-07-17
Add frozendict to nl_server

pradh:
https://github.com/datacommonsorg/website/pull/4487 introduced frozendict import in shared lib.  The NL server requirements was missing it ([Web server requirements has it](https://github.com/datacommonsorg/website/blob/e868bcc8a374fb72106d52cf7fd55a9d8a58f7bb/server/requirements.txt#L9)).


chejennifer (approved):
thanks!

PR #4496, opened 2024-07-18
Add ESLint warning for missing return type.

hqpho:
Codacy flags this but I would like to find out about it before waiting for checks to run.


chejennifer (approved):
nice

PR #4497, opened 2024-07-18
[SxS] only show tables that are used in answer

chejennifer:
for RAG answers, the table pane should only show tables actually used in the answer

autopush: https://screenshot.googleplex.com/J5CqEUu87ptoKCd
local: https://screenshot.googleplex.com/8ygR9UTWk7HrCu6


pradh (approved):
Thank you!  I'll defer to Hannah for js review!

hqpho (approved):

PR #4500, opened 2024-07-18
[SEO Experiment] Add plumbing to read experiment pages from GCS

juliawu:
To run our SEO experiments, this PR:

* Adds logic to the place pages for the server to read a jinja template from GCS.
* Starts a folder in server config to hold experiment template files.
* Checks in a sample template for Egypt with matching CSS.
* Adds a script to sync GCS bucket contents with local config folder.

Note: Templates for the other places in the experiment group will come in a follow up PR. Until those pages are completed, reading from GCS is only enabled in autopush, dev, and locally (i.e. not staging and not prod).

Screenshot 2024-07-18 at 12 32 54 PM


scripts/sync_seo_experiment_templates_to_gcs.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
+# Copyright 2023 Google LLC
```
dwnoble, line 2: Nit: 2024

server/routes/place/html.py
```diff
@@ -66,6 +67,49 @@
...
+                                       filename)
+      gcs.download_blob("datcom-website-config",
+                        os.path.join(SEO_EXPERIMENT_HTML_GCS_DIR, filename),
+                        template_filepath)
+
+    with open(template_filepath, 'r', errors='ignore') as f:
+      experiment_template = f.read()
+
+  except Exception as e:
+    logging.info(
```
dwnoble, line 108: Should this be logging.error? I dont think we have a standard way to do logging-what do you think?

server/routes/place/html.py
```diff
@@ -66,6 +67,49 @@
...
+
+    if current_app.config['LOCAL']:
+      # Load template from local path
+      template_filepath = os.path.join(current_app.root_path,
+                                       SEO_EXPERIMENT_HTML_LOCAL_DIR, filename)
+    else:
+      # Load template from GCS
+      template_filepath = os.path.join("/tmp/seo_experiment/template/",
+                                       filename)
+      gcs.download_blob("datcom-website-config",
```
dwnoble, line 100: Nit: extract bucket to a constant

dwnoble: Instead of downloading the template to a temporary directory, why not read directly from GCS to a string in memory?

server/routes/place/html.py
```diff
@@ -66,6 +67,49 @@
...
+  Will load files from GCS unless server is running locally, in which case
+  will load from server/config/seo_experiments/html_templates/active instead.
+
+  If a file cannot be found or read, will return empty string instead.
+  """
+  experiment_template = ""
+  try:
+    filename = SEO_EXPERIMENT_DCID_TO_HTML[dcid]
+
+    if current_app.config['LOCAL']:
```
dwnoble, line 92: How would we test loading templates through GCS when running locally?

server/routes/place/html.py
```diff
@@ -274,9 +318,52 @@ def place(place_dcid=None):
   # Block pages from being indexed if not on the main DC domain. This prevents
   # crawlers from indexing dev or custom DC versions of the place pages.
   block_indexing = not is_canonical_domain(flask.request.base_url)
-  logging.info(f"flask.requests.base_url is {flask.request.base_url}")
-  logging.info(f"Block indexing on place pages? {block_indexing}")
 
+  # Use SEO experiment templates on English overview pages for places in
+  # the experiment group.
+  use_experiment = place_dcid in SEO_EXPERIMENT_DCID_TO_HTML.keys(
```
dwnoble, line 324: Nit: extract all of the logic to see if the experiment is enabled to a function. then you wont need two flags use_experiment/is_experiment_enabled_flask_env

scripts/sync_seo_experiment_templates_to_gcs.sh
```diff
@@ -0,0 +1,28 @@
...
+# Will set the contents of the GCS bucket to match what's in local HEAD.
+#
+# To Use:
+#  (0) Set up server/config/seo_expermients/html_templates to contain the
+#      HTML templates in the hierarchy desired.
+#  (1) Make sure you're running from root with a clean HEAD
+#  (2) Make sure you've signed into authenticated to gcloud using 
+#          `gcloud auth application-default login`
+#  (3) Run `./scripts/sync_seo_experiment_templates_to_gcs.sh`
+
```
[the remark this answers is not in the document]
juliawu: Added a note about the GCS bucket name and the project ID in a comment.

server/config/seo_experiments/README.md
```diff
@@ -0,0 +1,19 @@
...
+
+The production website reads not from this folder, but from a copy of this
+folder stored in GCS. If changes are made to this folder, the changes can be
+synced to GCS using `./scripts/sync_seo_experiment_templates_to_gcs.sh` from
+the root directory.
+
+File structure:
+```text
+config/seo_experiments/
+├─ html_templates/
```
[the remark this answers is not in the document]
juliawu: Good catch! Removed.

server/routes/place/html.py
juliawu: +1 to using error. Will use error from here on out for exception/error catching.

server/config/seo_experiments/html_templates/active/Egypt.html
```diff
@@ -0,0 +1,407 @@
+{#
+Copyright 2020 Google LLC
```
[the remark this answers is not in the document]
juliawu: fixed.

server/routes/place/html.py
juliawu: Good idea. Extracted the bucket name to a constant (reusing already available constant), and added a new function to read directly from GCS to a text string.

server/routes/place/html.py
juliawu: Good idea, extracted out this logic to a function.

server/routes/place/html.py
juliawu: Switched to defaulting to reading from GCS so we can also test the loading templates path, and instead added an optional `useLocalTemplates=true` param to the request to enable reading files locally when I'm developing new templates.

server/routes/place/html.py
```diff
@@ -66,6 +68,52 @@
...
+      with open(template_filepath, 'r', errors='ignore') as f:
+        experiment_template = f.read()
+
+    else:
+      # Load template from GCS
+      gcs_filepath = gcs.make_path(
+          GLOBAL_CONFIG_BUCKET,
+          os.path.join(SEO_EXPERIMENT_HTML_GCS_DIR, filename))
+      output = gcs.read_to_string(gcs_filepath)
+      print(type(output))
```
dwnoble, line 108: nit: remove

dwnoble (approved):
Thanks Julia!

PR #4503, opened 2024-07-18
[tiles] fix fraction digits in highlight tile

chejennifer:
bug was that all the highlight tile values were showing 1 digit after the decimal. This was because:

- we always defaulted numFractionDigits to 1
- before PR https://github.com/datacommonsorg/website/pull/4417 we actually weren't setting numFractionDigits field correctly in the highlight tile & this change here caused us to set numFractionDigits correctly, but for default cases that meant setting it to 1: https://github.com/datacommonsorg/website/pull/4417/files#diff-a7d41f3b232c03e93d2230c267b7465828c783406c2be4a4998e0b9e68df1d31R212

To fix this,
- default numFractionDigits to undefined. This is safe in the other tiles because highlight tile is the only tile that uses numFractionDigits

screenshot diff that found this bug: https://screenshot.googleplex.com/4yiW2b8d6WGPZ9f
localhost: https://screenshot.googleplex.com/9cYvM2Xdx4ydo6C


dwnoble (approved):
Thanks Jennifer!

PR #4505, opened 2024-07-18
[nl] Fix PC stop words regex

chejennifer:
fix the regex used for detecting rate/rates

- dropped the word boundary because both classification detection and stop word removal (both cases where the stop words are used) add their own word boundary & having the word boundary in the stop word regex causes problems
- use negative lookbehind instead of negative lookahead because we should be looking for all cases of rate that are not preceded by special words like literacy or mortality


pradh:
Sorry, I don't understand the problem here.  Could you elaborate with an example?

pradh:
> Sorry, I don't understand the problem here. Could you elaborate with an example?

Maybe add some demonstrative test cases here: https://github.com/datacommonsorg/website/blob/master/shared/tests/lib/utils_test.py#L108 ?

chejennifer:
> Sorry, I don't understand the problem here. Could you elaborate with an example?

original per capita change intention:
- recognize rate/rates as a per capita query ONLY if it's not one of the rate words to skip
    - examples that should be recognized as per capita: obesity rate, rates of uninsured, etc
    - examples that should not be recognized as per capita: birth rate, mortality rates, etc

original per capita change bug:
- rate/rates NEVER recognized (e.g., https://screenshot.googleplex.com/A5e69bnvmjE4EDW)

pradh:
> > Sorry, I don't understand the problem here. Could you elaborate with an example?
> 
> original per capita change intention:
> 
> * recognize rate/rates as a per capita query only if it's not one of the rate words to skip
>   
>   * examples that should be recognized as per capita: obesity rate, rates of uninsured, etc
>   * examples that should not be recognized as per capita: birth rate, mortality rates, etc
> 
> original per capita change bug:
> 
> * rate/rates NEVER recognized (e.g., https://screenshot.googleplex.com/A5e69bnvmjE4EDW)
> 
> & yes, will add test cases

As just another thought, instead of trying to use such subtle regex, would it work to just add these combos to STOP_WORDS_EXCLUSIONS https://github.com/datacommonsorg/website/blob/ec4b48e9d6746241e8db791bd04ee13d64456a4e/shared/lib/constants.py#L154-L155 ?  And keep just `rate(s)?` in the PER_CAPITA?

pradh:
> STOP_WORDS_EXCLUSIONS

hmm, that means we won't strip "rate" out of "employment rate" as stop-word.  but we will consider the query as per-capita. which restores the original behavior.

chejennifer:
> > STOP_WORDS_EXCLUSIONS
> 
> hmm, that means we won't strip "rate" out of "employment rate" as stop-word. but we will consider the query as per-capita. which restores the original behavior.

yeah, I think we'll need more logic change if we use STOP_WORDS_EXCLUSIONS ..

shared/lib/constants.py
```diff
@@ -281,8 +291,8 @@
             "per capita",
             "percapita",
             "per person",
-            # remove "rate" or "rates" if is not followed by certain words (used as one metric)
-            f"\brate(s)?\b(?!\s*{_RATE_WORDS_TO_SKIP}\s+rate(s)?)",
+            # remove "rate" or "rates" if is not preceded by certain words (used as one metric)
+            f"rates?{_RATE_WORDS_LOOK_BEHINDS}",
```
pradh, line 295: Why no `\b`?  Also, why no `(s)?`?

shared/tests/lib/utils_test.py
```diff
@@ -110,6 +110,26 @@ def test_query_remove_stop_words(self, query, expected):
...
+      ],
+      # unemployment rate is a special case we don't want to strip
+      ["unemployment rate in palo alto", "unemployment rate palo alto"],
+      # following are all cases where rate/rates should be stripped
+      ["rate of unemployment in palo alto", "unemployment palo alto"],
+      [
+          "what are the rates of uninsured people in california",
+          "uninsured people california"
+      ],
+      ["obesity rate", "obesity"]
```
pradh, line 126: nit:  Should we add these queries to the non-toolformer case too to see diffs?

shared/lib/constants.py
chejennifer: no `\b` because these words are used in classifier and removing stop words and:
- it was having trouble getting recognized in the heuristic classifier when used with the additional regex logic there: https://github.com/datacommonsorg/website/blob/ec4b48e9d6746241e8db791bd04ee13d64456a4e/server/lib/nl/detection/heuristic_classifiers.py#L270
- when removing stop words, the logic there already adds '\b' anyways: https://github.com/datacommonsorg/website/blob/ec4b48e9d6746241e8db791bd04ee13d64456a4e/shared/lib/utils.py#L79

updated `(s)?` to just `s?` because it's the same thing, only difference is whether or not `s` is in a capturing group and there's no need for `s` to be a capture group

shared/tests/lib/utils_test.py
chejennifer: done

pradh (approved):
Thanks for the changes!  This is too subtle and really needs a revisit I think!

chejennifer:
thanks for the reviews, I'll add a buganizer to revisit this!

PR #4509, opened 2024-07-19
Consolidate CDC env file.

keyurva:
* Consolidates CDC env vars into a single file (`env.list`).
* See [go/cdc-env-file-consolidation](http://go/cdc-env-file-consolidation) for more details.
* As part of the consolidation `sqlite_env.list` and `cloudsql_env.list` need to be removed.
  + We'll remove them from the code once @kmoscoe has removed their mention from the docs.

* `Dockerfile` changes:
  + `lancedb` deps were moved to a different requirements file a few weeks ago. Adapting Dockerfile to that move.
  + Uses `npm ci` vs `npm install` for faster builds. See [this thread](https://stackoverflow.com/a/74068335) for more details.
* The Dockerfile changes are not related to env consolidation so I can do it in a separate PR if needed. LMK.
* Built docker image (`gcr.io/datcom-ci/datacommons-website-compose:keyurs-test`) with these changes and verified in the cloud [here](https://screenshot.googleplex.com/AdiWt7P4umvBuX4).

### DIFF
--- a/build/web_compose/compose.sh
+++ b/build/web_compose/compose.sh
@@ -19,20 +19,46 @@ export TOKENIZERS_PARALLELISM=false
 # https://github.com/UKPLab/sentence-transformers/issues/1318#issuecomment-1084731111
 export OMP_NUM_THREADS=1
 
+# If OUTPUT_DIR is not specified and the deprecated GCS_DATA_PATH is, use that as OUTPUT_DIR.
+if [[ $OUTPUT_DIR == "" && $GCS_DATA_PATH != "" ]]; then
+    echo "GCS Data Path: $GCS_DATA_PATH"
+    echo "GCS_DATA_PATH is deprecated and will be removed in the future. Use OUTPUT_DIR instead."
+    export OUTPUT_DIR=$GCS_DATA_PATH
+fi
+
+# Set OUTPUT_DIR to same as INPUT_DIR if not specified.
+if [[ $OUTPUT_DIR == "" ]]; then
+    if [[ $INPUT_DIR == "" ]]; then
+        echo "INPUT_DIR not specified."
+        exit 1
+    fi
+    export OUTPUT_DIR=$INPUT_DIR
+fi
+
+# Check for required variables.
+
+if [[ $DC_API_KEY == "" ]]; then
+  echo "DC_API_KEY not specified."
+  exit 1
+fi
+
+if [[ $MAPS_API_KEY == "" ]]; then
+  echo "MAPS_API_KEY not specified."
+  exit 1
+fi
+
+echo "OUTPUT_DIR=$OUTPUT_DIR"
+
+export IS_CUSTOM_DC=true
+export USER_DATA_PATH=$OUTPUT_DIR
+export ADDITIONAL_CATALOG_PATH=$USER_DATA_PATH/datacommons/nl/embeddings/custom_catalog.yaml
+
 if [[ $USE_SQLITE == "true" ]]; then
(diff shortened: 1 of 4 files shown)


custom_dc/env.list
```diff
@@ -1,7 +1,20 @@
...
 DC_API_KEY=
 # API key for accessing Maps and Places APIs for exploration tools.
 MAPS_API_KEY=
+
+### Directories ###
+
+# Input directory in Google Cloud Storage or local drive, containing files to be processed.
+INPUT_DIR=
+# Output directory in Google Cloud Storage or local drive for generated csv files.
+# Leave blank if same as INPUT_DIR.
```
hqpho, line 13: could maybe mark this as an optional param?

build/web_compose/compose.sh
```diff
@@ -19,21 +19,47 @@ export TOKENIZERS_PARALLELISM=false
...
+if [[ $OUTPUT_DIR == "" && $GCS_DATA_PATH != "" ]]; then
+    echo "GCS Data Path: $GCS_DATA_PATH"
+    echo "GCS_DATA_PATH is deprecated and will be removed in the future. Use OUTPUT_DIR instead."
+    export OUTPUT_DIR=$GCS_DATA_PATH
 fi
 
-nginx -c /workspace/nginx.conf
+# Set OUTPUT_DIR to same as INPUT_DIR if not specified.
+if [[ $OUTPUT_DIR == "" ]]; then
+    if [[ $INPUT_DIR == "" ]]; then
```
hqpho, line 31: Should we do this check even if OUTPUT_DIR is set?

hqpho (approved):

custom_dc/env.list
```diff
@@ -1,7 +1,20 @@
+### API Keys ###
```
kmoscoe, line 1: Keys > keys

custom_dc/env.list
```diff
@@ -1,7 +1,20 @@
...
+
 # API key for accessing the base Data Commons API.
 DC_API_KEY=
 # API key for accessing Maps and Places APIs for exploration tools.
 MAPS_API_KEY=
+
+### Directories ###
+
+# Input directory in Google Cloud Storage or local drive, containing files to be processed.
```
kmoscoe, line 10: Please add a line about how to specify the directory format for GCS, i.e. gs://<GCS bucket>/<directory>

build/web_compose/compose.sh
keyurva: Probably not. This docker ("services docker") is about serving data from the output directory so if it is set, that's the one we want to use.

custom_dc/env.list
keyurva: Done.

custom_dc/env.list
keyurva: Done.

custom_dc/env.list
keyurva: This one is tricky. This is actually one of the most important variables for the services docker.

Ideally, I would've liked to do `OUTPUT_DIR=$INPUT_DIR` but docker does not apply substitution on these variables. ([ref](https://stackoverflow.com/a/45923906))

I'm wondering if it's best to make it required instead and if both are the same, require that both be set to the same value.

Thoughts?

keyurva: Merging it as-is for now. Can do a follow-up if we decide to make it required.

keyurva (commented):
Thanks for the reviews.

PR #4514, opened 2024-07-24
Update nodejs query test goldens

juliawu:
Updates the goldens for our nodejs_query_differ test. Looks like there's been some data updates since we last updated these goldens.


juliawu:
@pradh @chejennifer could you look through and double check that these diffs look correct?

chejennifer (approved):
diffs make sense to me!

PR #4516, opened 2024-07-25
Update bad words python test

juliawu:
This PR removes the assertion that in the bad words test, in `multi` mode, lines should not contain spaces. This assertion breaks when creating cross products that include place names like "united states of america" or "united kingdom". 

This PR also updates the "headless drivers" NL test case to look for the correct detection type.

This should unblock the python and nl test failures seen in unrelated, already approved PRs.


pradh (approved):
Nice find!

PR #4517, opened 2024-07-25
Remove *_env.list files.

keyurva:
Now that the [doc](https://github.com/datacommonsorg/docsite/pull/456) has been updated to use the consolidated `env.list`, this PR removes the other `*_env.list` files.


kmoscoe (approved):
Could you just use this PR to add a comment to the INPUT/OUTPUT_DIR variables: "If it is a local directory, you must specify an absolute path"

PR #4518, opened 2024-07-25
Add absolute path comments in env.list.

keyurva:


kmoscoe (approved):
Nit: change "should be in the form" to "must be in the form" for both options.

PR #4519, opened 2024-07-25
Update .dockerignore to fix docker build

juliawu:
Updates .dockerignore to exclude the sanity.py file (in **/tests) to fix a cron-testing docker build failure.


keyurva (approved):
Thanks!

PR #4520, opened 2024-07-25
[CDC Autopush] Update scripts to accommodate data docker

hqpho:
- Split out submodule update into its own script so it can be run as a prerequisite step for both data and service docker build steps.
  - Make sure to get website hash before making a temp commit.
  - Write image label made of combined commit hashes to a temp file so other scripts can use it.
- Rename service docker build and deploy scripts to distinguish them from data docker build and (eventually) deploy scripts.
- Make some minor edits recommended by Shellcheck and shell-format VSCode extensions.

This PR will temporarily break custom DC autopush until I update the compose autopush cloudbuild config in the deployment repo. Planned update: https://paste.googleplex.com/6181631964741632


scripts/update_submods_for_cdc_autopush.sh
```diff
@@ -14,35 +14,37 @@
...
 image_label="${website_rev}-${mixer_rev}-${import_rev}"
 
-# Build a new image and push it to Container Registry, tagging it as latest
-docker build -f build/web_compose/Dockerfile \
-          --tag gcr.io/datcom-ci/datacommons-website-compose:${image_label} \
-          --tag gcr.io/datcom-ci/datacommons-website-compose:latest \
-          .
-docker push gcr.io/datcom-ci/datacommons-website-compose:${image_label}
-docker push gcr.io/datcom-ci/datacommons-website-compose:latest
+echo "$image_label" >"$image_label_path"
```
keyurva, line 50: Is it possible to have this script only echo the image label the end and have the cloud build yaml write it to a temp file? The other scripts can then just accept the label as an argument with the cloud build yaml again reading from the temp file instead of the scripts.

hqpho: I think this would also require getting rid of `set -x` which prints all commands as the script runs for ease of debugging logs. Is there a reason not to use a file to store this value? It's temporary only for each run of the trigger.

keyurva: Can we redirect just the last echo to a file? `./test.sh | tail -1 > test.txt`

The reason to use standard arguments vs files is to keep the scripts independent of cloud build specifics as much as possible. It arguably keeps the scripts simpler (this is of course subjective and YMMV!) and allows us to run in other contexts easily.

hqpho: Makes sense! Updated, and also updated [planned autopush cloudbuild changes](https://paste.googleplex.com/6181631964741632).

scripts/build_cdc_data_and_tag_latest.sh
```diff
@@ -14,36 +14,26 @@
...
-git commit --allow-empty -am "DO NOT PUSH: Temp commit to update pinned submod versions (empty if submods are already up-to-date)"
-
-website_rev="$(git rev-parse --short HEAD)"
-# Even though data docker does not use mixer, we use it to generate the tag so that the tags are consistent between data and services docker images.
-mixer_rev="$(git rev-parse --short HEAD:mixer)"
-import_rev="$(git rev-parse --short HEAD:import)"
-image_label="${website_rev}-${mixer_rev}-${import_rev}"
+image_label=$1
+if [[ $image_label = "" ]]; then
+  echo "Expected positional argument with image label."
```
keyurva, line 29: nit: May be also `echo "Usage: ./scripts/build_cdc_data_and_tag_latest.sh $IMAGE_LABEL"` here and elsewhere?

scripts/build_cdc_data_and_tag_latest.sh
```diff
@@ -14,36 +14,26 @@
...
+  exit 1
+fi
 
 # Build a new image and push it to Container Registry, tagging it as latest
 docker build -f build/cdc_data/Dockerfile \
-          --tag gcr.io/datcom-ci/datacommons-data:${image_label} \
-          --tag gcr.io/datcom-ci/datacommons-data:latest \
-          .
-docker push gcr.io/datcom-ci/datacommons-data:${image_label}
+  --tag "gcr.io/datcom-ci/datacommons-data:${image_label}" \
```
keyurva, line 35: Curious where quotes are needed here?

keyurva (approved):
Very cool!

scripts/update_submods_for_cdc_autopush.sh
keyurva: In the cloudbuild file, should the path be `workspace/...` or `/workspace/...`?

scripts/build_cdc_data_and_tag_latest.sh
hqpho: They're recommended by Shellcheck: https://www.shellcheck.net/wiki/SC2086

scripts/build_cdc_data_and_tag_latest.sh
hqpho: Done, here and in the service build script.

scripts/update_submods_for_cdc_autopush.sh
hqpho: You're right it should be `/workspace/...`! I'll update it before pushing that change. Thanks!!

PR #4524, opened 2024-07-26
sv index updates to handle global mortality/death rates

chejennifer:
- remove global from topic descriptions
- mortality rate/death rate -> mortalities/deaths
- remove "mortality rate" and "death rate" from Per Capita exclusions because the exclusion list is all forms of "x rate" that shows up as sv descriptions in the index

sv diffs (looks ok to me): https://storage.mtls.cloud.google.com/datcom-embedding-diffs/chejennifer_base_uae_mem_2024_07_26_13_31_00.html


tools/nl/embeddings/input/base/sheets_svs.csv
```diff
@@ -3389,7 +3389,7 @@ sdg/ER_PTD_TERR,Average proportion of Terrestrial Key Biodiversity Areas covered
 sdg/ER_RSK_LST,Red List Index
 sdg/SH_FPL_MTMM.AGE--Y15T49__SEX--F,Proportion of females of reproductive age who have their need for family planning satisfied with modern methods
 sdg/SH_STA_ANEM.AGE--Y15T49__SEX--F,Proportion of females aged 15-49 years with anaemia
-sdg/SH_STA_MORT.SEX--F,Maternal mortality ratio
+sdg/SH_STA_MORT.SEX--F,Maternal mortalities
```
pradh, line 3392: Wonder about plural for "mortality"?  (Makes sense for "deaths")

tools/nl/embeddings/input/base/main_topics.csv
```diff
@@ -340,20 +340,20 @@ dc/topic/VeteransIncome,Veterans Income;Veterans Income;income of veterans;veter
 dc/topic/Visibility,Visibility;Visibility
 dc/topic/Volcano,Volcano;Volcano
 dc/topic/Voting,Voting;Voting
-dc/topic/WHOChildImmunization,Global Child Immunization;Child immunization rates;WHO child immunization
```
pradh, line 343: Are we removing "immunization rates" too?

tools/nl/embeddings/input/base/main_topics.csv
```diff
@@ -54,7 +54,7 @@ dc/topic/CommuteModeByOccupation,Commute Mode By Occupation;Commute Mode By Occu
 dc/topic/CommuteTime,Commute Time;Commute Time;time spend during commute;time spent traveling to work;time spent travelling from work
 dc/topic/Crime,Crime;Crime
 dc/topic/Cyclone,Cyclone;Cyclone
-dc/topic/Debt,Debt;Debt;Global debt
+dc/topic/Debt,Debt;Debt
```
pradh, line 57: Delete dup?

shared/lib/constants.py
```diff
@@ -18,8 +18,8 @@
 import frozendict
 
 _RATE_WORDS_TO_SKIP = [
-    "birth", "change", "death", "exchange", "fertility", "literacy",
-    "mortality", "participation", "unemployment", "withdrawal"
+    "birth", "change", "exchange", "fertility", "literacy", "participation",
```
pradh, line 21: This could use a useful comment.  Something like, or better than:

"These are words that come before `rate`, which together mean a specific metric (like "fertility rate").  So if "fertility rate" appears in the query, we will not infer PerCapita heuristic.  Note that we expect to find the full phrase with the word `rate` in the NL index description."

pradh (commented):
Nice!

pradh (commented):
Why does dc/topic/GlobalHealth come before dc/topic/Health?  Is it the capital `H` vs. small `h`?

tools/nl/embeddings/input/base/sheets_svs.csv
chejennifer: updated to not plural

tools/nl/embeddings/input/base/main_topics.csv
chejennifer: yes because the "rate" causes it to show up for "global literacy rate"

tools/nl/embeddings/input/base/main_topics.csv
chejennifer: done

shared/lib/constants.py
chejennifer: done

server/integration_tests/test_data/demo_feb2023/query_6/chart_config.json
```diff
@@ -745,554 +745,6 @@
...
             "title": "Mental Health Deaths by Age"
-          },
-          {
-            "columns": [
-              {
-                "tiles": [
-                  {
-                    "statVarKey": [
-                      "Count_Person_WithDisability_Female",
-                      "Count_Person_WithDisability_Male"
```
pradh, line 756: why do we lose this?

server/integration_tests/test_data/e2e_no_default_place_specified/chart_config.json
```diff
@@ -327,65 +327,6 @@
             "denom": "Count_Person",
             "title": "Population: Female, Enrolled in School in States of United States of America"
           },
-          {
-            "columns": [
-              {
-                "tiles": [
-                  {
-                    "statVarKey": [
-                      "Count_Death_Female_AsAFractionOf_Count_Person_Female"
```
pradh, line 336: Is this a good thing we lost this?

pradh (approved):
Nice!

Its hard to tell from some of the integration-test diffs if they are OK, have you looked at them closely?  e.g., the SDG ones

server/integration_tests/test_data/demo_feb2023/query_6/chart_config.json
chejennifer: this one is interesting actually because it looks like these are stat vars from opening dc/topic/HealthEquity: https://screenshot.googleplex.com/9odTVhUvd6a5iUL

but dc/topic/HealthEquity has a score lower than 0.7 (the score threshold for uae model).

I think in this change, dc/topic/HealthEquity no longer gets opened because there is a new topic (dc/topic/WHOHealthBehavior) that comes before it

server/integration_tests/test_data/e2e_no_default_place_specified/chart_config.json
chejennifer: the query is "female population" & since this information about the female population got replace by other information about the female population, this is ok?

server/integration_tests/test_data/demo_feb2023/query_6/chart_config.json
pradh: Ah this is the sort of scenario I was worried about.  Though, in this case, the loss is ok it feels like?

server/integration_tests/test_data/e2e_no_default_place_specified/chart_config.json
pradh: yup, agreed.

server/integration_tests/test_data/demo_feb2023/query_6/chart_config.json
chejennifer: yeah I think the loss feels ok here! but are we even supposed to be opening topics that are lower than the score threshold?

pradh: hmm, wait, no, we shouldn't be doing that... let me check.

pradh: Could that be from rounding? https://screenshot.googleplex.com/7iJYjZ6StxfjAmK

chejennifer: oh maybe .. I'll take a look

pradh: Actually, could it be because HealthEquity is part of Health?  https://source.corp.google.com/piper///depot/google3/third_party/datacommons/schema/stat_vars/topics_health_insights.mcf;l=146?q=dc%2Ftopic%2FHealthEquity%20f:mcf

chejennifer: oh, actually, that makes sense

PR #4527, opened 2024-07-29
[CDC Stable] Util for finding and tagging a stable candidate from Custom DC service images

hqpho:


scripts/find_and_tag_cdc_stable_candidate.sh
```diff
@@ -0,0 +1,62 @@
...
+
+set -e
+
+###### BEGIN CONFIG ######
+
+# Tag of the website commit that we'll attempt to find a matching CDC image for.
+WEBSITE_PROD_TAG="prod"
+
+# Tag to be added to the CDC image that is found.
+STABLE_CANDIDATE_TAG="stable-candidate"
```
keyurva, line 30: We eventually want to tag as "stable". Is "stable-candidate" temporary until we have tagging working fully?

scripts/find_and_tag_cdc_stable_candidate.sh
```diff
@@ -0,0 +1,62 @@
...
+# which image the stable-candidate tag applies to.
+
+# Usage: ./scripts/find_and_tag_cdc_stable_candidate.sh
+
+set -e
+
+###### BEGIN CONFIG ######
+
+# Tag of the website commit that we'll attempt to find a matching CDC image for.
+WEBSITE_PROD_TAG="prod"
```
keyurva, line 27: Can the script take the website tag as a parameter? Also, does it need to be a tag or can we get it to work with a commit sha as well?

scripts/find_and_tag_cdc_stable_candidate.sh
hqpho: The way I envisioned the workflow was (1) automatically tag stable-candidate (2) verify tests pass against it (3) manually change the tag from stable-candidate to stable. Thoughts?

scripts/find_and_tag_cdc_stable_candidate.sh
hqpho: Sure, both are doable, though it's more work and maintenance to have the script be able to do more things. What will the standard workflow be?

keyurva: The tag would either be based on a prod tag or on a specific commit. It'd be good to support both.

scripts/find_and_tag_cdc_stable_candidate.sh
keyurva: I'm thinking that since autopush sanity would've already kicked in, we'd know before hand if the version had any issues. WDYT?

hqpho: It would have kicked in, but it might not have succeeded. I think it'd be good to have a manual verification of test results before finalizing that the image is stable.

scripts/find_and_tag_cdc_stable_candidate.sh
hqpho: Ok, working on adding this now.

scripts/find_and_tag_cdc_stable_candidate.sh
keyurva: SG. 

A couple questions:
* Are you thinking we'd have a cloud run instance pointing to stable-candidate and use that for testing?
* We need to tag the git commit with the `customdc_stable` tag as well. Would we be tagging it with `customdc_stable_candidate` first and update to `customdc_stable` after the manual verification?

dwnoble:
@hqpho Looking through my old PRs- is this still relevant?

hqpho:
> @hqpho Looking through my old PRs- is this still relevant?

Converted back to a draft since this needs to be rethought! Planning to edit [this doc](https://docs.google.com/document/d/1dY8WbX7o2HLY8ct_gXo6w5g9iZVcEJlo-e3hybpzI-Q/r/0-xFxRPRLr8XmzsnWc8QKJFw/edit) first and then revisit this PR.

dwnoble:
sounds good! thanks @hqpho

dwnoble (approved):

PR #4529, opened 2024-07-29
Update GlobalHealth topic description

chejennifer:
- in PR https://github.com/datacommonsorg/website/pull/4524, removed "Global" from all topic descriptions, however this caused losses for queries like "Health+conditions+vs+median+age+in+Alameda+County" and "Most+common+medical+conditions+in+US" because GlobalHealth as a topic got ranked higher than Health and HealthConditions
-  here we revert that global change & instead replace the word "Global" with "World", which is a word that can not be overindexed because it would be removed from the query (it is a place) and we confirmed that "global" in the query does not prefer "world" from SV description

sv diffs: https://storage.mtls.cloud.google.com/datcom-embedding-diffs/chejennifer_base_uae_mem_2024_07_29_15_21_46.html


pradh:
> * here we revert that global change & instead replace the word "Global" with "World", which is a word that can not be overindexed because it would be removed from the query (it is a place)

Also.... "And we confirmed that "global" in the query does not prefer "world" from SV description"?

server/integration_tests/test_data/sdg/query_3/chart_config.json
```diff
@@ -10,9 +10,9 @@
                 "tiles": [
                   {
                     "statVarKey": [
-                      "WHO/M_Est_smk_daily"
+                      "Count_Death"
```
pradh, line 13: Nice, reverted

pradh (approved):
Nice!

chejennifer:
thanks for the review!

PR #4531, opened 2024-07-30
[Eval UI] Tweak table parsing

hqpho:
- Require table divider to be at least three dashes long. This prevents sequences of 1 or 2 dashes in values from getting parsed as dividers.
- Allow any character other than a pipe in a table header value.


chejennifer (approved):
this regex is nice, thanks for the fix!

PR #4534, opened 2024-08-01
Fix map tool stat var search with many places

hqpho:
Fixes b/356689537 where variable search in the new map tool when plotting countries on earth resulted in an error "Request line is too large".

Since the error only reproduces when running the website server with Gunicorn, also added a mode to run_server.sh for ease of reproing.

I've moved all request params to the request body, but if we want to only move places (maybe for the sake of analytics), I can modify this.

### DIFF
--- a/run_server.sh
+++ b/run_server.sh
@@ -22,7 +22,6 @@ function cleanup {
 }
 trap cleanup SIGINT
 
-
 source .env/bin/activate
 
 PORT=8080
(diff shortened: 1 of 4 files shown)


dwnoble:
This may be a separate issue, but now when when i search for a stat var from place=Earth, child place type=Country, i get a mixer timeout: 

http://localhost:8080/tools/visualization#visType%3Dmap%26place%3DEarth%26placeType%3DState

dwnoble (approved):

PR #4537, opened 2024-08-05
Update cache key for stat var search POSTs

hqpho:
Quick fix to follow up https://github.com/datacommonsorg/website/pull/4534 and unblock the website release. For the future it would be nice to wrap @cache.cached in a custom decorator that we can use everywhere which takes care of passing default params and making sure post body is in the cache key.


dwnoble (approved):
Thanks Hannah!

PR #4538, opened 2024-08-05
Fix misplaced comment.

kmoscoe:


kmoscoe:
Also, should we move ADMIN_SECRET to the deprecated section?

keyurva:
~ Also, should we move ADMIN_SECRET to the deprecated section?

We should but probably wait until we have published the data docker docs?

keyurva (approved):
Thanks for fixing this.

PR #4539, opened 2024-08-06
Update redirects.json

premr:
Adding two redirects for datacommons.org/link/video and datacommons.org/link/form for the two pager


dwnoble (approved):
Thanks prem!

Side note: we should add a better way to test these redirects locally

premr:
Should we add Natalie's broken URL test to this file.somehow?

dwnoble:
I'm thinking:

(1) It looks like there's no easy way to test if the redirect works until the redirects file is deployed to our production bucket. i'll submit a follow-up PR with a way to test redirects locally.
(2) if there's typo in redirects.json, it'll break all redirects and give the user a 500 error (but won't impact anything else in the site). I'll add a way to handle gracefully and just redirect to '/' (datacommons.org) if there's an error in the redirects file.
(3) Testing for broken URLs: @n-h-diaz do you think the broken URL test would help here?

dwnoble:
Regardless @premr i was able to test this one and it's good to merge!

n-h-diaz:
> I'm thinking:
> 
> (1) It looks like there's no easy way to test if the redirect works until the redirects file is deployed to our production bucket. i'll submit a follow-up PR with a way to test redirects locally. (2) if there's typo in redirects.json, it'll break all redirects and give the user a 500 error (but won't impact anything else in the site). I'll add a way to handle gracefully and just redirect to '/' (datacommons.org) if there's an error in the redirects file. (3) Testing for broken URLs: @n-h-diaz do you think the broken URL test would help here?

Oh hmm the test I added for schema just tests for 200 status codes, not sure if that would work here since it requires being deployed first / tested locally. The test is more for checking if urls have broken over time though, so if you'd like I can add these urls once they've been submitted

PR #4541, opened 2024-08-06
remove world from mortality topic descriptions

chejennifer:
"world" was being overindexed in queries like "global population". 

This is a short term fix. Longer term fix will be to have a set of negative variables that require a higher score threshold to be returned

sv diffs: https://storage.mtls.cloud.google.com/datcom-embedding-diffs/chejennifer_base_uae_mem_2024_08_06_16_04_57.html


pradh (approved):
Thanks for the fix!

PR #4545, opened 2024-08-08
Use async when loading Google maps APIs

juliawu:
This PR adds `loading=async` to the call to load the Google Maps API, as per [Google's Documentation](https://developers.google.com/maps/documentation/javascript/overview#Loading_the_Maps_API). 

This removes the following console warning from pages with maps calls:

Screenshot 2024-08-08 at 9 50 54 AM


dwnoble (approved):
Thanks Julia!

PR #4546, opened 2024-08-08
Update RIG UI to show footnotes if no tooltips

chejennifer:
If there are no tooltips shown, go back to old UI of showing footnotes

Screenshot 2024-08-08 at 3 45 40 PM


hqpho:
Just to clarify, does this mean for SxS you've done some extra processing on RIG answers? If so what's that processing?

hqpho (approved):

chejennifer:
> Just to clarify, does this mean for SxS you've done some extra processing on RIG answers? If so what's that processing?

In the sheets that I feed in for the new SxS evals, I've stripped the dc call annotation (__DC__#1(...)) and replaced it with either the dc stat or the llm stat

PR #4547, opened 2024-08-11
Create smaller, faster building services docker.

keyurva:
* This PR creates the full multi-stage services image.
* The new image is more than 50% smaller to download - [1 GB](https://screenshot.googleplex.com/5qum85dSxSxUieq) vs [2.3 GB](https://screenshot.googleplex.com/767mwgPEqk3es35)
   + On-disk decompressed size is [3.77 GB vs 7.69 GB](https://screenshot.googleplex.com/Bxw7rvUJcdJXmJd)
* It is also about 3x faster to build in the cloud - [6 min](https://screenshot.googleplex.com/747AAifhDVKGUi3) vs [18 min](https://screenshot.googleplex.com/3ta43Yd9yUEwy2z)
* The new image bundles the embeddings in the image thereby improving the startup time (20 sec vs 30 sec on my machine).
* Verified by running it both locally and in the [cloud](https://screenshot.googleplex.com/6bEjV3Eer4sJCvb).
* Also disabled the web admin in the new services image so data docker is the only option to load data.
  + Will remove the admin code once the doc is updated and we've completely deprecated the older image.
* Updated the runtime image as well to be multi-stage.
* Next steps
   + Auto build and publish the new image - cc @hqpho 
   + Update doc to use new image - cc @kmoscoe 
   + Remove web admin

### DIFF
--- a/build/cdc_services/Dockerfile
+++ b/build/cdc_services/Dockerfile
@@ -112,7 +112,7 @@ WORKDIR /workspace/static
 RUN npm run-script build
 
 
-# #### Stage 3: Download base dc model and embeddings from GCS. ####
+# #### Stage 4: Download base dc model and embeddings from GCS. ####
 FROM --platform=linux/amd64 google/cloud-sdk:slim as download
 
 # Copy model and embeddings.
--- a/build/cdc_services/Dockerfile
+++ b/build/cdc_services/Dockerfile
@@ -127,7 +127,7 @@ FROM gcr.io/datcom-ci/datacommons-services-runtime:2024-08-08 as runner
 # Env defaults.
 ENV WEBSITE_MIXER_API_ROOT=http://127.0.0.1:8081 \
     ENV_PREFIX=Compose \
-    ENABLE_ADMIN=true \
+    ENABLE_ADMIN=false \
     DEBUG=false \
     USE_SQLITE=false \
     USE_CLOUDSQL=false \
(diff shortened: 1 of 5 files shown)


build/cdc_services_runtime/Dockerfile
dwnoble, line 1: Why not have this image be another build stage direclty in cdc_services?

build/cdc_services/Dockerfile
```diff
@@ -48,13 +48,91 @@ RUN protoc \
...
 # #### Final stage: Runtime env. ####
-FROM gcr.io/datcom-ci/datacommons-services-runtime:2024-08-06 as runner
+FROM gcr.io/datcom-ci/datacommons-services-runtime:2024-08-08 as runner
 
-# Env that should not change
+# Env defaults.
 ENV WEBSITE_MIXER_API_ROOT=http://127.0.0.1:8081 \
     ENV_PREFIX=Compose \
-    ENABLE_ADMIN=true
+    ENABLE_ADMIN=false \
```
dwnoble, line 130: Should we keep this as ENABLE_ADMIN=true until we're ready to update the docs with the CDC data container use?

dwnoble (approved):

dwnoble:
Really great improvements! Thanks Keyur!

build/cdc_services/Dockerfile
keyurva: The current web-compose image continues to support admin. We'll add references to this image in the doc only after the data docker doc updates are in.

build/cdc_services_runtime/Dockerfile
keyurva: I eventually want to make builds for partners with their custom changes faster as well. To do that I want to discuss a layered approach with you and Bo and a separate runtime image should help in that regard.

Let's discuss later this week.

PR #4551, opened 2024-08-13
Build cdc services image with docker buildkit enabled.

keyurva:
* To fix [build failure](https://screenshot.googleplex.com/9sGn944m5eJmqVc).

### DIFF
--- a/scripts/build_cdc_services_and_tag_latest.sh
+++ b/scripts/build_cdc_services_and_tag_latest.sh
@@ -32,7 +32,7 @@ if [[ $image_label = "" ]]; then
 fi
 
 # Build a new image and push it to Container Registry, tagging it as latest
-docker build -f build/cdc_services/Dockerfile \
+DOCKER_BUILDKIT=1 docker build -f build/cdc_services/Dockerfile \
   --tag "gcr.io/datcom-ci/datacommons-services:${image_label}" \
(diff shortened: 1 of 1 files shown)


scripts/build_cdc_services_and_tag_latest.sh
```diff
@@ -32,7 +32,7 @@ if [[ $image_label = "" ]]; then
 fi
 
 # Build a new image and push it to Container Registry, tagging it as latest
-docker build -f build/cdc_services/Dockerfile \
+DOCKER_BUILDKIT=1 docker build -f build/cdc_services/Dockerfile \
```
hqpho, line 35: Nit: Do we want a line break between DOCKER_BUILDKIT=1 and docker build? Or does that not mean the same thing?

hqpho (approved):

scripts/build_cdc_services_and_tag_latest.sh
keyurva: I ran into this with the [manual build](https://github.com/datacommonsorg/website/blob/fc155d929b80ae2484e7621335c5a9a8f76fdd3b/build/ci/cloudbuild.push_cdc_services_image.yaml#L28) where it didn't work if it was not on the same line.

keyurva (commented):
Thanks for the quick review!

PR #4553, opened 2024-08-14
Run chmod as a separate step when building docker image.

keyurva:
* When building the new services image locally, @kmoscoe ran into an error that said buildkit was required when building this image.
* The only statement in the Dockerfile that requires buildkit is [this one](https://github.com/datacommonsorg/website/blob/2f0779132258148b217aa3e9b9356e2fd48e2a01/build/cdc_services/Dockerfile#L156) where chmod is run as part of COPY.
* Separating it out into its own command gets rid of this requirement so partners don't need to worry about enabling it.
* However, I've still kept it enabled for both [manual](https://github.com/datacommonsorg/website/blob/2f0779132258148b217aa3e9b9356e2fd48e2a01/build/ci/cloudbuild.push_cdc_services_image.yaml#L28) and [auto](https://github.com/datacommonsorg/website/blob/2f0779132258148b217aa3e9b9356e2fd48e2a01/scripts/build_cdc_services_and_tag_latest.sh#L35) builds since it turns out that building with buildkit enabled is much faster.
  + The services build takes[ 11 min](https://screenshot.googleplex.com/BsGTg99qQoPZbSQ) with buildkit disabled and [5 min](https://screenshot.googleplex.com/NcaB2mWoPvywPyf) with it enabled.

### DIFF
--- a/build/cdc_services/Dockerfile
+++ b/build/cdc_services/Dockerfile
@@ -153,7 +153,10 @@ COPY --from=static /workspace/server ./server
 COPY --from=download /tmp/datcom-nl-models /tmp/datcom-nl-models
 
 # Copy executable script.
-COPY --chmod=0755 build/cdc_services/run.sh .
+COPY build/cdc_services/run.sh .
+
+# Make script executable.
+RUN chmod +x run.sh
 
 # Add virtual env to the path.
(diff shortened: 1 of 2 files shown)


keyurva (commented):
Thanks for the review!

PR #4554, opened 2024-08-14
Use url from env for running cdc tests.

keyurva:
* To be able to run the tests against different CDC instances.


server/webdriver/cdc_tests/autopush_test.py
hqpho, line 1: Now that this file isn't just for autopush, should we rename it?

hqpho (changes requested):

server/webdriver/cdc_tests/autopush_test.py
keyurva: Good point. Done.

hqpho (approved):

PR #4557, opened 2024-08-16
Create initial client to fetch dc api keys for a given project.

keyurva:


tools/apigee/apigee.py
```diff
@@ -0,0 +1,90 @@
...
+    return key_string
+
+  async def fetch_dc_api_key_name(self, project_id: str) -> str:
+    response = await self.http_get(
+        f"{_API_KEYS_BASE_URL}/v2/projects/{project_id}/locations/global/keys")
+    for key in response.get("keys", []):
+      key_name = key.get("name")
+      if not key_name:
+        continue
+      for api_target in key.get("restrictions", {}).get("apiTargets", []):
```
hqpho, line 61: Should we also return any keys that don't have restrictions?

keyurva: Discussed offline. Updated to keep the restriction but return all keys with the restriction instead of just one.

hqpho (approved):

PR #4561, opened 2024-08-19
[nl] update bar sorting

chejennifer:
updating sorting bars in a bar chart to not have a default order if there is more than 1 sv because for some svpg, we want to show it in a specific order

e.g., we want age distribution to be in the order of the ranges
prod: https://screenshot.googleplex.com/BVyKS3ZWk9pW8vk
local: https://screenshot.googleplex.com/BPdhLpgGMfDx8tq


server/lib/nl/config_builder/bar.py
```diff
@@ -118,10 +118,14 @@ def multiple_place_bar_block(column,
...
-  sort_order = None if is_special_dc(state.uttr.insight_ctx) \
-    else BarTileSpec.DESCENDING
+  sort_order = None
+
+  # No default sort_order for special DC since UN wants the order to be as
+  # provided in variable groupings. Also no default sort order if there is more
+  # than one sv because for cases like age ranges, we also want the order to be
+  # what's provided in the variable groupings.
+  if not is_special_dc(state.uttr.insight_ctx) and cspec.svs == 1:
+    sort_order = BarTileSpec.DESCENDING
```
beets, line 128: Could we instead disable the sorting for cases where there are numbers in the SV (e.g. age ranges, number of rooms, income ranges etc). This seems like a regression for other cases, such as modes of commute or job families  which are more useful when they are sorted.

chejennifer: updated to only cases where there is a range in the SVs

chejennifer:
thanks for the review!

PR #4562, opened 2024-08-20
Read projects from Google Sheet and write dc keys back to it.

keyurva:
* The keys are written to a `keys` column in the same sheet. [[Example]](https://docs.google.com/spreadsheets/d/1hxOmST5kLtil8RFfZcwYgeRvSATvvPa3EkduPpvNosc/edit?gid=193373547#gid=193373547&range=F:F)
* Once the keys are in the sheet, the next step is to import them into apigee (next PR).

### DIFF
--- a/tools/apigee/apigee.py
+++ b/tools/apigee/apigee.py
@@ -40,10 +40,14 @@ _APIGEE_ORGANIZATION = os.environ.get("APIGEE_ORGANIZATION")
 # The URL to the Google Sheet with details of partners to be migrated.
 _SHEETS_URL = os.environ.get("SHEETS_URL")
 # The name of the worksheet at the SHEETS_URL with details of partners to be migrated.
+# Worksheet should have the following named columns:
+# project_id, developer_email, developer_first_name, developer_last_name, org_name, keys
 _WORKSHEET_NAME = os.environ.get("WORKSHEET_NAME")
 
 assert _DC_API_TARGET, "'DC_API_TARGET' env variable not specified"
 assert _APIGEE_ORGANIZATION, "'APIGEE_ORGANIZATION' env variable not specified"
+assert _SHEETS_URL, "'SHEETS_URL' env variable not specified"
+assert _WORKSHEET_NAME, "'WORKSHEET_NAME' env variable not specified"
 
 
 class CloudApiClient:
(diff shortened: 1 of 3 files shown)


tools/apigee/apigee.py
```diff
@@ -36,16 +37,23 @@
 # The apigee organization to migrate the keys to.
 # e.g. datcom-apigee
 _APIGEE_ORGANIZATION = os.environ.get("APIGEE_ORGANIZATION")
+# The URL to the Google Sheet with details of partners to be migrated.
+_SHEETS_URL = os.environ.get("SHEETS_URL")
+# The name of the worksheet at the SHEETS_URL with details of partners to be migrated.
+_WORKSHEET_NAME = os.environ.get("WORKSHEET_NAME")
```
hqpho, line 43: Maybe add asserts if these aren't set like we have for _DC_API_TARGET and _APIGEE_ORGANIZATION?

Also, consider adding a comment about expected sheet format (worksheet should have a project_id column)

tools/apigee/apigee.py
```diff
@@ -156,11 +171,81 @@ async def http_post(self, url: str, data: dict = None) -> dict:
...
 
 
+class SheetsClient:
+
+  def __init__(self, cloud_client: CloudApiClient, sheets_url: str,
+               worksheet_name: str) -> None:
+    self.cloud_client = cloud_client
+    self.sheets_url = sheets_url
+    self.worksheet_name = worksheet_name
+    # gs = gspread.authorize(self.cloud_client.credentials)
```
hqpho, line 181: rm?

tools/apigee/apigee.py
```diff
@@ -105,10 +113,17 @@ async def create_developer(self, email: str, first_name: str,
...
+      self, project_ids: list[str]) -> dict[str, list[str]]:
+    futures = [self.get_dc_api_keys(project_id) for project_id in project_ids]
+    keys = dict(await asyncio.gather(*futures))
+    return keys
+
+  async def get_dc_api_keys(self, project_id: str) -> tuple[str, list[str]]:
     key_names = await self.fetch_dc_api_key_names(project_id)
     futures = [self.fetch_dc_api_key_string(key_name) for key_name in key_names]
-    return list(filter(lambda x: x, await asyncio.gather(*futures)))
+    keys = list(filter(lambda x: x, await asyncio.gather(*futures)))
```
hqpho, line 125: Any risk that this results in enough concurrent requests to cause quota issues?

tools/apigee/apigee.py
keyurva: Done.

tools/apigee/apigee.py
keyurva: It is guarded by the semaphore so we're ok.

tools/apigee/apigee.py
keyurva: Done.

PR #4563, opened 2024-08-20
Added disableEntityLink option to bar chart web component

dwnoble:
(UN ask)

Adds `disableEntityLink` option to bar chart web components to remove the entity link from x-axis:

9Yy4WfAEkGGD5PY

### DIFF
--- a/static/js/chart/draw_bar.ts
+++ b/static/js/chart/draw_bar.ts
@@ -317,7 +317,7 @@ export function drawStackBarChart(
     chartHeight,
     x,
     false,
-    labelToLink,
+    options?.disableEntityLink ? null : labelToLink,
     null,
     options?.apiRoot
   );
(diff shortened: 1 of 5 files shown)


static/js/chart/draw_bar.ts
```diff
@@ -317,7 +317,7 @@ export function drawStackBarChart(
     chartHeight,
     x,
     false,
-    labelToLink,
+    options?.disableEntityLink ? null : labelToLink,
```
keyurva, line 320: I'm not super familiar with this code but do we have tests for these functions / components? Would be good to add tests if we don't.

keyurva (approved):
Thanks Dan.

chejennifer (approved):
nice!

static/js/chart/draw_bar.ts
dwnoble: We dont 😬  - testing web components doesnt work out of the box with our webdriver setup, and adding support has been on my todo list

PR #4566, opened 2024-08-21
Disable obs browser pages for some custom DCs

n-h-diaz:
Updates 
* climate_trace
* custom 
* feedingamerica
* iitm
* unsdg

This is to remove dependency on BQ (and is currently broken for many instances currently), so we don't have to keep as many BQ versions

Verified locally for each instance, but wanted to double check that this won't negatively impact any prod instance?


n-h-diaz:
Updated with a new field in the config based on recommendations and verified locally. Let me know if this seems okay (or if there's a better/preferred way to do this)!

server/app_env/_base.py
```diff
@@ -91,3 +91,6 @@ class Config:
   MIN_STAT_VAR_GEO_COVERAGE = 10
   # NL Bad words file.
   BAD_WORDS_FILE = 'nl_bad_words.txt'
+  # Whether to disable BigQuery from instance. This is primarily used for
+  # accessing the observation browser pages.
+  DISABLE_BQ = False
```
keyurva, line 96: Is it possible to make it disabled by default and enable it only for main dc?

Also, if we go with that, consider naming it `ENABLE_BQ`.

n-h-diaz: Good idea, thanks! I switched to ENABLE_BQ and only added for main dc + test environments

server/app_env/_base.py
```diff
@@ -91,3 +91,6 @@ class Config:
   MIN_STAT_VAR_GEO_COVERAGE = 10
   # NL Bad words file.
   BAD_WORDS_FILE = 'nl_bad_words.txt'
+  # Whether to enable BigQuery from instance. This is primarily used for
```
keyurva, line 94: nit: The comment does not look right. Can you correct it?

server/app_env/feedingamerica.py
```diff
@@ -26,4 +26,4 @@ class Config(_base.Config):
 class LocalConfig(local.Config):
   CUSTOM = True
   NAME = "Feeding America"
-  SHOW_DISASTER = False
\ No newline at end of file
+  SHOW_DISASTER = False
```
keyurva, line 29: nit: This file can be left as is.

keyurva (approved):
Very cool - thanks Natalie!

server/app_env/_base.py
n-h-diaz: Done

server/app_env/feedingamerica.py
n-h-diaz: Done

PR #4568, opened 2024-08-21
Add support for importing keys into apigee.

keyurva:
* Also, added the ability to run the script in multiple modes.
* Did the migration for all staging keys.
  - Keys along with the created apps and developers are now in [apigee](https://screenshot.googleplex.com/8cBkUrb6pz69npe).
  - Sheet [updated](https://screenshot.googleplex.com/4d68tetAc3iQmxV) to reflect the details.
  - Note that none of the staging keys have any developer details, so the script defaults developer and app info.
* Working with @ajaits on some spreadsheet issues with prod keys. Will migrate prod once those are resolved.

### DIFF
--- a/tools/apigee/apigee.py
+++ b/tools/apigee/apigee.py
@@ -13,11 +13,15 @@
 # limitations under the License.
 
 import asyncio
+from collections import defaultdict
 from enum import StrEnum
+from functools import wraps
 import json
 import logging
 import os
+import random
 import re
+import string
 
 from absl import app
 from absl import flags
@@ -44,6 +48,7 @@ _API_KEYS_BASE_URL = "https://apikeys.googleapis.com"
 _APIGEE_BASE_URL = "https://apigee.googleapis.com"
 _HTTPX_LIMITS = Limits(max_keepalive_connections=5, max_connections=10)
 _HTTP_RESOURCE_EXISTS_CODE = 409
+_SECRET_CHARS = string.ascii_letters + string.digits
 
 # The DC API target of the keys being migrated.
 # e.g. api.datacommons.org
@@ -51,6 +56,9 @@ _DC_API_TARGET = os.environ.get("DC_API_TARGET")
 # The apigee organization to migrate the keys to.
 # e.g. datcom-apigee
 _APIGEE_ORGANIZATION = os.environ.get("APIGEE_ORGANIZATION")
+# The apigee api product that the migrated keys should be associated to.
+# e.g. api, api-staging
+_APIGEE_API_PRODUCT = os.environ.get("APIGEE_API_PRODUCT")
 # The URL to the Google Sheet with details of partners to be migrated.
 _SHEETS_URL = os.environ.get("SHEETS_URL")
 # The name of the worksheet at the SHEETS_URL with details of partners to be migrated.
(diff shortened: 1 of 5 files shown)


tools/apigee/apigee.py
```diff
@@ -44,6 +58,11 @@
 # project_id, developer_email, developer_first_name, developer_last_name, org_name, keys
 _WORKSHEET_NAME = os.environ.get("WORKSHEET_NAME")
 
+_DEFAULT_DEVELOPER_EMAIL = "datcom-core@google.com"
```
keyurva, line 61: @hqpho - Please comment if these defaults make sense or suggest alternatives.

tools/apigee/apigee.py
```diff
@@ -64,15 +83,66 @@ def __init__(self, dc_api_target: str, apigee_organization: str) -> None:
...
+      data["migrated_developer_last_name"] = developer_last_name
+
+      return key
+    except HTTPStatusError as hse:
+      logging.error(
+          "Error importing key to apigee: %s (%s),\nRequest:\n%s,\nResponse:\n%s",
+          key, str(hse), hse.request.content, hse.response.content)
+      return ""
+
+  async def create_key(self, developer_email: str, app_name: str,
```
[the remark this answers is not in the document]
keyurva: Added a separate `associate_key` method to do this.

tools/apigee/apigee.py
```diff
@@ -64,15 +83,66 @@ def __init__(self, dc_api_target: str, apigee_organization: str) -> None:
...
                        key: str) -> str:
     request = {"consumerKey": key, "consumerSecret": key}
```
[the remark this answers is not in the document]
keyurva: Generating a random secret now.

tools/apigee/apigee.py
hqpho: The defaults are what we'll use if we don't have an email address for a key, right? Rather than just "Apigee Migrated", maybe we can include something in the name that makes it clear that this is a key we didn't find an owner for? Like "Owner Unknown" or "Unknown Org"?

tools/apigee/apigee.py
```diff
@@ -188,27 +315,68 @@ def __init__(self, cloud_client: CloudApiClient, sheets_url: str,
...
     self._insert_project_keys(data, project_id_2_keys)
 
     # Write data with keys back to the sheet.
     self._write_sheet_data(data)
 
+  async def import_keys(self):
+    """
+    Imports DC API keys into apigee.
+
+    Reads the sheet that should already include they keys,
```
hqpho, line 344: Nit: they -> the

tools/apigee/staging.env
```diff
@@ -1,4 +1,5 @@
 DC_API_TARGET=staging.api.datacommons.org
 APIGEE_ORGANIZATION=datcom-apigee-dev
+APIGEE_API_PRODUCT=api-staging
```
hqpho, line 3: can update this to `datacommons-api-staging`

tools/apigee/apigee.py
```diff
@@ -154,8 +255,22 @@ async def fetch_dc_api_key_names(self, project_id: str) -> list[str]:
       logging.warning("No DC API key name found for project: %s", project_id)
```
hqpho, line 255: Optional: When no keys are found, put something in the sheet about it so there is an easy way to tell between a row that hasn't been exported and a row that was successfully exported with no keys.

hqpho (approved):

tools/apigee/apigee.py
keyurva: Done.

tools/apigee/apigee.py
keyurva: When there are no keys, nothing is exported and an empty `migrated_keys` column is an indicator that it wasn't.

tools/apigee/apigee.py
keyurva: Done.

tools/apigee/staging.env
keyurva: Done. Updated prod to `datacommons-api` as well.

PR #4572, opened 2024-08-22
Added terraform variable google_analytics_tag_id for enabling Google Analytics

dwnoble:
- Added terraform variable `google_analytics_tag_id` for enabling Google Analytics
- The Google Analytics Tag ID can now be passed as an environment variable to Custom Data Commons. Previously, users had to set the variable in their `server/app_env/*.py` config file


kmoscoe:
Problem is the name of the variable in base.html is GA_ACCOUNT. Don't they
need to be the same?

dwnoble:
> Problem is the name of the variable in base.html is GA_ACCOUNT. Don't they need to be the same?

@kmoscoe main.tf takes care of translating between google_analytics_tag_id and GA_ACCOUNT. Maybe in a follow up we can rename GA_ACCOUNT to GOOGLE_ANALYTICS_TAG_ID to be more precise

server/__init__.py
```diff
@@ -450,8 +450,13 @@ def log_unhandled(e):
       app.logger.error('Error thrown for request: %s\nerror: %s', request.url,
                        e)
 
+  # Attempt to retrieve the Google Analytics Tag ID (GA_ACCOUNT):
+  # 1. First, check the environment variables for 'GA_ACCOUNT'.
+  # 2. If not found, fallback to the application configuration ('GA_ACCOUNT' in app.config).
+  ga_account = os.environ.get('GA_ACCOUNT', app.config['GA_ACCOUNT'])
```
keyurva, line 456: Can we add `GA_ACCOUNT` to `env.list` as well in the [optional section](https://github.com/datacommonsorg/website/blob/40d3ac2be9d508b60ebee4b16e3d3eff82e341f8/custom_dc/env.list#L37)?

@kmoscoe - can we document this variable once this PR is in?

keyurva (approved):
Thanks Dan. We should continue moving all config properties from python to env like this.

server/__init__.py
keyurva: Yes, let's do that!

kmoscoe:
Yes please!  Let's add it to env.list | sample and rename it. I will prepare a doc PR for that.

server/__init__.py
dwnoble: added

dwnoble:
> Yes please! Let's add it to env.list | sample and rename it. I will prepare a doc PR for that.

Done, and renamed GA_ACCOUNT to GOOGLE_ANALYTICS_TAG_ID

build/cdc/dev/.env.sample
```diff
@@ -38,5 +38,8 @@ CLOUDSQL_INSTANCE=
 # Use custom data commons environment.
 FLASK_ENV=custom
 
+# Optional: Google Analytics Tag ID
```
keyurva, line 41: Can we put the same comment in here as it is in `env.list`?

> (Optional) Set to your Google Tag ID to enable Google Analytics. See https://support.google.com/analytics/answer/9539598

server/__init__.py
```diff
@@ -450,8 +450,13 @@ def log_unhandled(e):
       app.logger.error('Error thrown for request: %s\nerror: %s', request.url,
                        e)
 
+  # Attempt to retrieve the Google Analytics Tag ID (GOOGLE_ANALYTICS_TAG_ID):
+  # 1. First, check the environment variables for 'GOOGLE_ANALYTICS_TAG_ID'.
+  # 2. If not found, fallback to the application configuration ('GOOGLE_ANALYTICS_TAG_ID' in app.config).
+  ga_account = os.environ.get('GOOGLE_ANALYTICS_TAG_ID', app.config['GOOGLE_ANALYTICS_TAG_ID'])
```
keyurva, line 456: The env var name looks good but let's not change the python config name else it will break for users who are using it today.

I'd recommend marking the python property as deprecated and asking them to use the env variable instead.

build/cdc/dev/.env.sample
dwnoble: done

server/__init__.py
dwnoble: Updated

server/__init__.py
```diff
@@ -450,8 +450,18 @@ def log_unhandled(e):
       app.logger.error('Error thrown for request: %s\nerror: %s', request.url,
                        e)
 
+  # Attempt to retrieve the Google Analytics Tag ID (GOOGLE_ANALYTICS_TAG_ID):
+  # 1. First, check the environment variables for 'GOOGLE_ANALYTICS_TAG_ID'.
+  # 2. If not found, fallback to the application configuration ('GOOGLE_ANALYTICS_TAG_ID' in app.config).
+  # 3. If still not found, fallback to the deprecated application configuration ('GA_ACCOUNT' in app.config).
+  config_deprecated_ga_account = app.config['GA_ACCOUNT']
+  if config_deprecated_ga_account:
+    logging.warn("Use of GA_ACCOUNT is deprecated. Use GOOGLE_ANALYTICS_TAG_ID instead.")
```
keyurva, line 459: nit: s/Use GOOGLE_ANALYTICS_TAG_ID/Use env variable GOOGLE_ANALYTICS_TAG_ID

server/__init__.py
```diff
@@ -450,8 +450,18 @@ def log_unhandled(e):
...
+  config_google_analytics_tag_id = app.config['GOOGLE_ANALYTICS_TAG_ID']
```
keyurva, line 460: Any reason to support this as a python property at all? We can just say that the `GA_ACCOUNT` property is deprecated and they should use the `GOOGLE_ANALYTICS_TAG_ID` env variable instead.

keyurva (approved):
Just think about whether we need to support the renamed python property or direct them to always use the env variable.

server/__init__.py
dwnoble: Updated

server/__init__.py
dwnoble: I like keeping the config variable for now for better backwards compatibility- this way our various config.py files can switch over to `GOOGLE_ANALYTICS_TAG_ID=...`. I think in a future iteration we can completely deprecate the config.py construct and move all options to environment variables.

PR #4573, opened 2024-08-23
Add custom_dc/env.list to .gitignore

wittrock:
The instructions for running a custom datacommons include creating a file (`custom_dc/env.list`) which is likely to contain API keys. Add it to .gitignore to avoid accidental checkins.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -71,6 +71,7 @@ experimental/sdg-static/datacommons/nl_interface.min.css
 
 # Custom DC data
 dc-data/
+custom_dc/env.list
 
 # Topic cache
 gen_ordered_list_for_topics.mcf


dwnoble:
Hey @wittrock , I like this change to avoid accidentally checking in API keys, but i think it's also helpful to have a template of what variables are available. What do you think about renaming "env.list" to "env.list.sample" and ignoring "env.list" (like you have)?

Then we'd need to update the instructions here: https://docs.datacommons.org/custom_dc/custom_data.html (https://github.com/datacommonsorg/docsite) to tell the user to copy the env.list.sample file to the env.list file

wdyt @keyurva @kmoscoe

keyurva:
I like Dan's idea. To not break existing users, can we do it in 3 steps:

1. Create a PR with `env.list` duplicated as `env.list.sample`.
2. Update the doc that uses the latter.
3. Create a PR that removes `env.list` and adds it to `.gitignore`.

Thoughts?

kmoscoe:
Yes, I like Dan's approach too, where they create a copy of the sample
instead of directly editing it.

kmoscoe:
And yes, add it to .git ignore. I wonder if we should also add all the
generated files to .git ignore. I was quite surprised to create a PR that
added 32 generated files to it... !

wittrock:
Done, created :
1. #4577 to add env.list.sample and 
1. https://github.com/datacommonsorg/docsite/pull/488 to update the docs

Those two should merge first, then this one.

wittrock:
Also split out my setup instructions change into #4578

wittrock:
I think this can merge now! PTAL

keyurva:
> I think this can merge now! PTAL

Yep. Retriggered tests and enabled auto merge.

keyurva (approved):

PR #4576, opened 2024-08-23
Set default value for FLASK_ENV.

keyurva:
* The `FLASK_ENV` value seldom changes, so setting it to a default value alleviates partners from having to provide one themselves.
* Ran into an issue with a partner yesterday where this was not set and it resulted in a startup error.
* Setting a default will get rid of this error.

### DIFF
--- a/build/cdc_services/Dockerfile
+++ b/build/cdc_services/Dockerfile
@@ -127,7 +127,7 @@ FROM gcr.io/datcom-ci/datacommons-services-runtime:2024-08-08 as runner
 # Env defaults.
 ENV WEBSITE_MIXER_API_ROOT=http://127.0.0.1:8081 \
     ENV_PREFIX=Compose \
-    FLASK_ENV=compose \
+    FLASK_ENV=custom \
     ENABLE_ADMIN=false \
     DEBUG=false \
     USE_SQLITE=false \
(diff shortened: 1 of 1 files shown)


build/cdc_services/Dockerfile
```diff
@@ -127,6 +127,7 @@ FROM gcr.io/datcom-ci/datacommons-services-runtime:2024-08-08 as runner
 # Env defaults.
 ENV WEBSITE_MIXER_API_ROOT=http://127.0.0.1:8081 \
     ENV_PREFIX=Compose \
+    FLASK_ENV=compose \
```
dwnoble, line 130: Should this be "custom"?

keyurva: Duh - yes! Done - thanks for catching it.

dwnoble: Also with the env variable set in the dockerfile, would users actually be able to override it?

keyurva: Yes. https://screenshot.googleplex.com/4XbKDwniTRgAsWm

dwnoble: Looks like overriding should work (https://stackoverflow.com/questions/40334751/dockerfile-overriding-env-variable)

dwnoble (approved):

build/cdc_services/Dockerfile
dwnoble: awesome thanks!

PR #4582, opened 2024-08-26
Removed 'Analyze this data in BigQuery' button from visualization tools

dwnoble:
## Before
before1
after2

## After
after1
before2


n-h-diaz:
Thanks for doing this! Could you wait on submitting this for now? I think the usage dashboard I was looking at earlier wasn't updated/out-of-date, so want to double check with Prem

n-h-diaz:
> Thanks for doing this! Could you wait on submitting this for now? I think the usage dashboard I was looking at earlier wasn't updated/out-of-date, so want to double check with Prem

Just checked with Prem and looks like we're good to disable for now, thanks!

dwnoble:
> > Thanks for doing this! Could you wait on submitting this for now? I think the usage dashboard I was looking at earlier wasn't updated/out-of-date, so want to double check with Prem
> 
> Just checked with Prem and looks like we're good to disable for now, thanks!

Sounds good-thanks Natalie!

chejennifer (approved):
nice!

PR #4588, opened 2024-09-03
Feature/home page react (#1)

jenibelle08:
## Description

This PR converts the home page to React. The conversion keeps the functionality and appearance of the home page unchanged, but contains substantial changes to how these pages are composed. This is an interim step to implementing the redesign of the home page.

*This PR is not intended for merge into master until the completion of the homepage revamp*

## Notes

In addition to the conversion of the content of home page, this PR also converts the header and footer of the site to React, also in anticipation of the home page redesign. 

New React apps were added in the homepage (rendered by `homepage/main.ts`) and the base template (rendered in `base/main.ts`). 

To facilitate different modules independently rendering components on load, the `window.onLoad` assignments have in all modules been converted to `window.addEventListener` (without this change, only the last assigned onLoad handler is applied).

We set up a system of standardized transferring of routing and translation label information from the Jinja templates to the TypeScript. The routing and translations are now rendered into respective elements in an auxiliary template that is included by the base template. These can then be read by any TypeScript module (such as a `main.ts`) and then passed through to to the relevant apps. Individual pages can still independently read their own attributes past up through the templates.

## Considerations

Previously, the menu content in the headers and footers were contained directly in the templates. The move to rendering with React components presents an opportunity separate this content from the actual implementation of the header and footer, with the goal of making it easier for forks to override the content without having to make any changes to the component that implements it.

We have tentatively set up a process by which the header and footer content are sent through from the backend as JSON (in a similar way to how the partners and topics are sent through to the homepage). 

The goal here will be to set up an easy way for a fork to provide a custom version of the header or footer in their own backend (for example, with the commonly used `_custom_dc` tag in the filename), to override in the fork without affecting the core. This could be done either in a separate PR, or following feedback and discussion, in this PR.

### Future Fork Customization

Before this project is completed, we will need to consider the potential fork customization options for the new React structure of the home page (as we are no longer dealing with a scenario where a significant amount of the material to be customized lies in the template).

We can consider three levels of customization that a fork might want to implement:
* Content: This is the simplest customization, referred to above, where they may which to change the available menus and associated text.
* Style: In this scenario, a fork may which to keep the structure of the components intact, but style the components (changing fonts, colors, and the layout to the degree that the structure allows).
* Components: Here, a fork may choose to remove, edit or otherwise substantially change the components themselves.

These will be considered as development of the homepage revamping progresses.


static/js/apps/base/main.ts
```diff
@@ -0,0 +1,79 @@
...
+    document.getElementById("metadata-base").dataset.header
+  ) as HeaderMenu[];
+  const footerMenu = JSON.parse(
+    document.getElementById("metadata-base").dataset.footer
+  ) as FooterMenu[];
+
+  const name = document.getElementById("metadata-base").dataset.name;
+  const logoPath = document.getElementById("metadata-base").dataset.logoPath;
+  const hideFullFooter =
+    document.getElementById("metadata-base").dataset.hideFullFooter === "true";
```
dwnoble, line 45: This might need to get lowercased-- im seeing the full footer in other pages in the app: 

in http://localhost:8080/tools/statvar

```
<div id="metadata-base" class="d-none"
...
  data-hide-full-footer="True"
...
/>
```

server/templates/base.html
```diff
@@ -94,190 +94,31 @@
...
       {% block content %}
       {% endblock %}
     </main>
-
-    <footer id="main-footer">
-      {% if not is_hide_full_footer %}
-      <div class="container top-footer">
-        <div class="row">
-          <section class="col-12 col-sm-6 col-md-4">
-            {# TRANSLATORS: The label for a collection of exploration tools. #}
```
dwnoble, line 192: It's important to keep all of the `{#  TRANSLATORS: ... #}` notes when migrating the `{% trans %}...{% endtrans %}` tags

These tags get extracted by pybabel tool and wind up in the translation files: https://github.com/datacommonsorg/website/blob/f5e8e87c2291d87dfa37a3a887f01d7ff28d6467/server/i18n/all.pot#L77

References:
https://github.com/datacommonsorg/website/blob/master/docs/localization.md
https://github.com/datacommonsorg/website/blob/f5e8e87c2291d87dfa37a3a887f01d7ff28d6467/scripts/extract_messages.sh#L37

server/templates/auxiliary/labels.html
```diff
@@ -0,0 +1,56 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-#}
+
+<div id="metadata-labels" class="d-none">
+  <div data-label="Data Commons" data-value="{% trans %}Data Commons{% endtrans %}"></div>
```
dwnoble, line 18: Make sure to include `{# TRANSLATORS #}` comments (see my other comment about this)

server/templates/base.html
```diff
@@ -94,190 +94,31 @@
...
+  data-header="{{ HEADER_MENU }}"
+  data-footer="{{ FOOTER_MENU }}"
+  data-logo-path="{{ LOGO_PATH }}"
+  data-name="{{ NAME }}"
+  data-show-disaster="{{ SHOW_DISASTER }}"
+  data-show-sustainability="{{ SHOW_SUSTAINABILITY }}"
+  data-hide-full-footer="{{ is_hide_full_footer }}"
+  data-hide-sub-footer="{{ is_hide_sub_footer }}"
+  data-subfooter-extra="{{ subfooter_extra_content|safe }}"
+  data-brand-logo-light="{{ brand_logo_light }}"
```
dwnoble, line 107: Nit: maybe delete this flag in a follow-on pr - i dont think it's actually used anymore

static/js/apps/base/utilities/utilities.ts
```diff
@@ -0,0 +1,82 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { Labels, Routes } from "../../../shared/types/general";
+
+export const resolveHref = (href: string, routes: Routes): string => {
```
dwnoble, line 19: Nit: add a comment describing this function

static/js/apps/base/utilities/utilities.ts
```diff
@@ -0,0 +1,82 @@
...
+  if (match) {
+    const routeKey = match[1];
+    const resolvedRoute = routes[routeKey] || "";
+    return href.replace(regex, resolvedRoute);
+  } else {
+    return routes[href] || href;
+  }
+};
+
+export const slugify = (text: string): string => {
```
dwnoble, line 32: Nit: add a comment describing this function

static/js/apps/base/utilities/utilities.ts
```diff
@@ -0,0 +1,82 @@
...
+  return routes;
+};
+
+export const getLabels = (elementId = "metadata-labels"): Labels => {
+  const labelElements = document.getElementById(elementId)?.children;
+  const labels: Labels = new Proxy(
+    {},
+    {
+      get: (target, prop): string => {
+        return prop in target ? target[prop] : (prop as string);
```
dwnoble, line 69: A typo when fetching a label might be hard to detect. Could you add a log message if the label isn't found?

static/js/apps/base/utilities/utilities.ts
```diff
@@ -0,0 +1,82 @@
...
+      routes[routeTag] = element.getAttribute("data-value");
+    });
+  }
+
+  return routes;
+};
+
+export const getLabels = (elementId = "metadata-labels"): Labels => {
+  const labelElements = document.getElementById(elementId)?.children;
+  const labels: Labels = new Proxy(
```
dwnoble, line 65: Should this be `Proxy<Labels>`?

static/js/apps/base/utilities/utilities.ts
```diff
@@ -0,0 +1,82 @@
...
+    .replace(/--+/g, "-");
+};
+
+export const getRoutes = (elementId = "metadata-routes"): Routes => {
+  const routeElements = document.getElementById(elementId)?.children;
+  const routes: Routes = new Proxy(
+    {},
+    {
+      get: (target, prop): string => {
+        return prop in target ? target[prop] : (prop as string);
```
dwnoble, line 48: i think this should throw an error if the route isn't found - does that sound right?

dwnoble (commented):
Thanks Jennifer! The behavior looks overall consistent with the existing homepage. I did see the other pages are loading the full footer though:
Screenshot 2024-09-03 at 4 57 29 PM

should be:

Screenshot 2024-09-03 at 4 57 29 PM


I'll let @beets  weigh in as well

static/js/apps/base/utilities/utilities.ts
```diff
@@ -0,0 +1,82 @@
...
+};
+
+export const slugify = (text: string): string => {
+  return text
+    .toString()
+    .toLowerCase()
+    .trim()
+    .replace(/\s+/g, "-")
+    .replace(/[^\w-]+/g, "")
+    .replace(/--+/g, "-");
```
beets, line 39: curious when we encounter this string?

static/js/apps/base/components/HeaderBar.tsx
beets, line 1: please use snake case for these longer filenames e.g. header_bar.tsx

beets (commented):
thanks for sending this change. in general, it's much easier for us if these pr's are small :)

the changes look reasonable and looks like it's at parity - even the e.g. ?hl=fr parameter works! i didn't review the homepage components too closely since we are about to change those to the new content.

for the footer, it looks like we are doing away with the full footer and only keeping the bottom 4 links. so hopefully a lot of it can be simplified as well.

while we are adding code, can we please add file/module comments, and comments for exported functions. tests would also be great for the exported functions.

static/js/apps/base/utilities/utilities.ts
beets: +1 -- i'd prefer to fail loudly in dev for errors like that (and in getLabels)

beets: I see - thanks for explaining the reasoning behind the logic. In that case, how about moving all the routing logic up into jinja and keeping the routes in the JSON config. The ts should only include fully formed URLs (with `{tools.visualization}#visType=timeline` expanded out as well), removing the need for regex's and onload complexity?

static/js/apps/homepage/components/learn_more.tsx
```diff
@@ -0,0 +1,72 @@
...
+          <div className="learn-more-content">
+            <h3>Learn More</h3>
+            <p>
+              Data forms the foundation of science, policy, and journalism, but
+              its full potential is often limited. Data Commons addresses this
+              by offering cloud-based APIs to access and integrate cleaned
+              datasets, boosting their usability across different domains.
+            </p>
+            <div className="learn-more-links">
+              <a href={resolveHref("static.about", routes)}>
```
dwnoble, line 48: "{static.about}"

static/js/apps/homepage/components/tools.tsx
```diff
@@ -0,0 +1,95 @@
...
+                routes
+              )}
+              id="timeline-button"
+              className="tool-button"
+            >
+              <div className="tool-icon"></div>
+              <span>Timelines</span>
+            </a>
+            <a
+              href={resolveHref("place.place", routes)}
```
dwnoble, line 73: "{place.place}"

static/js/apps/homepage/components/tools.tsx
```diff
@@ -0,0 +1,95 @@
...
+              id="place-button"
+              className="tool-button"
+            >
+              <div className="tool-icon"></div>
+              <span>Place explorer</span>
+            </a>
+            <a
+              href={resolveHref("tools.download", routes)}
```
dwnoble, line 81: "{tools.download}"

static/js/apps/base/components/footer.tsx
```diff
@@ -0,0 +1,133 @@
...
+                width="74"
+                height="25"
+                src={
+                  brandLogoLight
+                    ? "/images/google-logo-reverse.svg"
+                    : "/images/google-logo.svg"
+                }
+                alt="Google logo"
+              />
+              <img
```
dwnoble, line 104: Remove extra logo

static/js/apps/base/utilities/utilities.ts
beets: thanks - could you add a todo here to revisit this after the revamp?

static/js/apps/base/utilities/utilities.ts
```diff
@@ -0,0 +1,82 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { Labels, Routes } from "../../../shared/types/general";
+
+export const resolveHref = (href: string, routes: Routes): string => {
+  const regex = /{([^}]+)}/;
```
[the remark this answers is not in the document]
beets: sounds good - could you add a todo here to revisit this after we get the bulk of the revamp done?

static/js/apps/homepage/app_v2.tsx
```diff
@@ -0,0 +1,90 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * Main component for Version 2 of the homepage.
```
beets, line 18: please add a todo to fold this into app.tsx (dropping the v2) when ready.

beets (approved):
thanks for addressing the comments. i left a few comments - they are mainly to add todo's in code so it's easy to keep track of tasks we are deferring for now in the interest of time. we'll have to do another pr of all the changes when we merge to the main branch, so we'll take a closer look at everything then. i'll approve now to unblock progress - please merge after the todo's are added.

dwnoble:
Thanks @nick-next and @jenibelle08!

PR #4591, opened 2024-09-04
Rough version of a check on the NL response with the redirection.

gmechali:


gmechali:
PR WIP - The test is still not passing, just sharing out now to see if this code structure is up to standards (for the short term fix), and if this is the kind of test to add - though I need to fix it.
Successfully ran locally, see screencast: https://screencast.googleplex.com/cast/NDUyOTkxMjUyMTE2Mjc1MnwwMDA3NjU1ZC1lYQ

PR #4592, opened 2024-09-05
Clean up existing docs for running website locally.

gmechali:
Adds `set -e` to run_npm script to exit on failure. 
Adds instructions for installing python-setuptools for python 3.12 and over since preinstallation ended with version 3.11.


docs/developer_guide.md
```diff
@@ -45,6 +45,15 @@ website and mixer changes.
   ./run_test.sh --setup_python
   ```
 
+- Python 3.12
+
+  You need to run the following command:
+  ```brew install python-setuptools
```
dwnoble, line 51: should be:

(Replace ' with ` below, otherwise the comment itself renders the markdown)

'''bash
brew install python-setuptools
'''

docs/developer_guide.md
```diff
@@ -45,6 +45,15 @@ website and mixer changes.
...
+  ```
+  or for linux:
+  ```pip install python-setuptools
```
dwnoble, line 54: ditto

docs/developer_guide.md
```diff
@@ -45,6 +45,15 @@ website and mixer changes.
   ./run_test.sh --setup_python
   ```
 
+- Python 3.12
+
+  You need to run the following command:
```
dwnoble, line 50: Add a note that this is for mac users

docs/developer_guide.md
```diff
@@ -45,6 +45,15 @@ website and mixer changes.
   ./run_test.sh --setup_python
   ```
 
+- Python 3.12
```
dwnoble, line 48: This section header seems at odds with the above `Python 3.11` section that reads: "Confirm the Python3 version is 3.11.x. Otherwise install/upgrade your Python and confirm the version". 

Can you make it more cohesive?

docs/developer_guide.md
gmechali: done.

docs/developer_guide.md
gmechali: done.

docs/developer_guide.md
gmechali: done.

docs/developer_guide.md
gmechali: So I think the python 3.11 guidance should just be that we need you to have AT LEAST python 3.11.
But if you have 3.12, then you also need to do this. 
I changed the wording to reflect that but let me know if you prefer to change it to just tell them to use 3.12 and always follow that additional setup.

dwnoble: I agree using Python 3.11+ sounds to me!

dwnoble (approved):

PR #4593, opened 2024-09-05
Client side redirect for place explorer + hides unnecessary components

gmechali:
PR WIP - The test is still not passing, just sharing out now to see if this code structure is up to standards (for the short term fix), and if this is the kind of test to add - though I need to fix it.
Successfully ran locally, see screencast: https://screencast.googleplex.com/cast/NDUyOTkxMjUyMTE2Mjc1MnwwMDA3NjU1ZC1lYQ


static/js/apps/explore/app.tsx
```diff
@@ -214,6 +216,13 @@ export function App(props: { isDemo: boolean }): JSX.Element {
       pageMetadata.pageConfig &&
       pageMetadata.pageConfig.categories
     ) {
+      // If the response is a single PLACE_OVERVIEW tile, redirect to the place explorer.
+      if (isPlaceOverviewOnly(pageMetadata)) {
+        console.log('Redirecting to place explorer');
```
dwnoble, line 221: Remove log statement

static/js/apps/explore/app.tsx
```diff
@@ -214,6 +216,13 @@ export function App(props: { isDemo: boolean }): JSX.Element {
...
+        let url = 'place/';
```
dwnoble, line 222: Add a leading slash to the url. Otherwise, if i start from `http://localhost:8080/explore/` (with the trailing slash), the redirect takes me to `/explore/place/<dcid>`.

static/js/apps/explore/app.tsx
```diff
@@ -214,6 +216,13 @@ export function App(props: { isDemo: boolean }): JSX.Element {
...
+        url += pageMetadata.pageConfig.metadata["placeDcid"][0];
```
dwnoble, line 223: To avoid a possible array out of bounds, can you use `pageMetadata.place.dcid` instead of  `pageMetadata.pageConfig.metadata["placeDcid"][0]` ?  The former is the "primary place" detected in the query, which should always be the same as the place overview

dwnoble: Prefer using const when possible. What do you think about:

```
const placeDcid = pageMetadata.place.dcid;
const url = `/place/${placeDcid}`;
```

static/js/apps/explore/app.tsx
```diff
@@ -214,6 +216,13 @@ export function App(props: { isDemo: boolean }): JSX.Element {
...
+        window.open(url, "_self");
```
dwnoble, line 224: Use `window.location.replace(url);` instead. `window.open` adds a new record to the browser history, so hitting the back button takes the user back to a page where they'll just get redirected again

static/js/components/subject_page/block.tsx
```diff
@@ -666,7 +666,8 @@ function renderTiles(
           </p>
         );
       case "PLACE_OVERVIEW":
-        return <PlaceOverviewTile key={id} place={place} />;
+        // TODO(gmechali): Switch to server-side redirection
+        return tiles.length == 1 ? <></> : <PlaceOverviewTile key={id} place={place} />;
```
dwnoble, line 670: I'm seeing this message flash (for < 1s) 
Screenshot 2024-09-05 at 3 46 20 PM

when visiting: 
http://localhost:8080/explore/#q=United+states+of+america&client=ui_query

static/js/apps/explore/app.tsx
gmechali: Done.

static/js/apps/explore/app.tsx
gmechali: hm interesting, im not seeing that happening when I run local though. Still done.

static/js/apps/explore/app.tsx
gmechali: Done.

static/js/components/subject_page/block.tsx
gmechali: hmm interesting, im not seeing this at all. 
https://screencast.googleplex.com/cast/NjUyNDY1MzM0MTI0NTQ0MHw2ODkzN2MwMS01YQ

static/js/apps/explore/app.tsx
gmechali: Done - for the record the back and forward buttons still seem to work with window.open :)

static/js/apps/explore/app.tsx
dwnoble: Hm could be because im running in the `./run_cdc_dev_docker.sh` setup

static/js/apps/explore/app.tsx
dwnoble: If i started from localhost:8080/ and ran a query, the back button seemed to work, but if i started from localhost:8080/explore/ and ran a query, then i'd see the intermediate page with window.open 🤷

static/js/components/subject_page/block.tsx
dwnoble: I'm seeing it if i start from /explore : https://screencast.googleplex.com/cast/NDU0NzMwODc0OTMyNDI4OHwzZTI5NGYxMi1kMw

static/js/apps/explore/app.tsx
gmechali: ah okay I had misunderstood your comments about starting from explore/, so I didnt try that. it looks like it works now with this approach though!

static/js/components/subject_page/block.tsx
gmechali: So I wasnt able to see the first message you mentioned, but I did see the second one ("See relevant statistics based on the previous query."). That's now taken care of.

dwnoble (approved):
Thank you!

gmechali:
This PR should be good to go, all the tests passed. Dan, leaving it open for now in case you have anything to add here, otherwise will submit this afternoon!

dwnoble (approved):

PR #4594, opened 2024-09-05
Updating NL goldens

beets:
Part of the mixer release -- updated goldens based on staging mixer.

I see a lot of new bio dcid's (entities and properties) being matched -- that doesn't seem to be expected. There doesn't seem to be any UI impact though.

PR #4595, opened 2024-09-06
Update NL test goldens

beets:
Undoes transient error - see discussion in #4594 for context.


chejennifer (approved):
thanks for updating!

PR #4598, opened 2024-09-07
New desktop header

jenibelle08:
### Description

This PR is an initial version of the desktop menu for the homepage revamp. The mobile menu is temporarily removed, so that the desktop menu can be reviewed in isolation. 

*This PR is not intended for merge into master until the completion of the homepage revamp*

### Notes

There is still some refactoring and cleanup to do (particularly in the CSS) that will be done as part of a subsequent larger PR that includes the mobile menu. However, the functionality of the desktop can be previewed.


static/js/apps/base/components/menu_desktop_rich_menu.tsx
```diff
@@ -0,0 +1,72 @@
...
+  open: boolean;
+}
+
+const MenuDesktopRichMenu = ({
+  menuItem,
+  routes,
+  open,
+}: MenuDesktopRichMenuProps): ReactElement => {
+  return (
+    <div className={`rich-menu-content ${open ? "open" : ""}`}>
```
beets, line 39: this looks like the only difference between desktop and mobile -- can we fold both into the same component (ok to add a todo for now).

beets (approved):
thanks for this - it's nice to see it working across the site. i added some comments - they are mostly very minor and have to do with naming. could you add them as todo's and we can merge this.

server/config/base/header_v2.json
```diff
@@ -0,0 +1,203 @@
+[
```
dwnoble, line 1: We'll have to think about a good way to add translations to these strings (in a follow-up pr)

static/js/apps/base/components/header_bar/menu_desktop_rich_menu.tsx
```diff
@@ -0,0 +1,74 @@
+/**
```
dwnoble, line 1: (For a follow-up PR): When scrolling down the page with the menu open, the menu sticks to the top.

dwnoble (approved):
Looks great!

PR #4602, opened 2024-09-11
Follow-up to header, footer and page revamp

nick-next:
### Description

This PR implements the P0 follow-up items from the triage list.

Of particular note, the revamp of the header and footer had caused an issue some of the older existing pages that were relying on the older HTML structure (and overriding that structure themselves). In order to solve this, we reworked our structure for the new header and footer to make it compatible with those older pages.


dwnoble (approved):

beets (approved):
thanks for these fixes!

PR #4608, opened 2024-09-11
Menu behavior, copy updates and bounce fix.

nick-next:
### Description

This PR addresses items from the Triage list, as well as some accessibility and menu behavior improvements:

* Copy updates
* Keyboard behavior on the menu improved (including a fix to how the mobile menu was handling tab order
* A fix to bouncing content relating to the loading in of the logo
* Removal of the Safari artifact around the video in the hero

### Notes

I added a TODO on the server side to determine the size of the logo automatically rather than have it given explicitly. It is possible to do (and easy to do via Pillow for raster images like .pngs) but since some logos are SVGs, which Pillow cannot handle and so might require a different library, I put this aside for a follow-up after the revamp. For now, the logo width can be supplied explicitly and this will remove the bounce.


beets (approved):
thanks for these fixes!

PR #4614, opened 2024-09-11
Create cloudbuild to run website sanity in the cloud.

keyurva:


shifucun (approved):

beets:
thanks for adding this! could we add instructions to [go/dcrelease](https://goto.google.com/dcrelease) so we check this on staging / prod during each release?

PR #4617, opened 2024-09-12
Update DataGemma paper link

shifucun:


server/routes/redirects/redirects.json
```diff
@@ -13,5 +13,5 @@
   "dc-coverage-aa2": "/tools/visualization#visType%3Dmap%26place%3DEarth%26placeType%3DCountry%26sv%3D%7B%22dcid%22%3A%22Count_Variable_AdministrativeArea2%22%7D",
   "video": "https://www.youtube.com/watch?v=O6iVsS-RDYI",
   "form": "https://forms.gle/C816TLMyp5fBdDt67",
-  "DataGemmaPaper": "https://datacommons.org"
+  "DataGemmaPaper": "https://docs.datacommons.org/papers/DataGemma-FullPaper.pdf"
```
kmoscoe, line 16: I think Prem wanted you to use the "redirect" URL https://datacommons.org/link/DataGemmaPaper so that we don't need to update this when the paper goes to Arxiv.

shifucun: Right, the link here is the actual url. This mapping is to setup the redirect.

kmoscoe: Oh, I see!  Got it.

kmoscoe (approved):

server/routes/redirects/redirects.json
shifucun: thanks!

PR #4620, opened 2024-09-13
Add more comments to file

kmoscoe:
Added note re GCS output dirs and moved NL model info from doc to here.

### DIFF
--- a/custom_dc/env.list.sample
+++ b/custom_dc/env.list.sample
@@ -12,9 +12,10 @@ MAPS_API_KEY=
 # Local paths must be absolute paths.
 INPUT_DIR=
 # Output directory in Google Cloud Storage (GCS) or local drive for generated csv files.
-# GCS paths must be of the form gs://<BUCKET>/<FOLDER>.
+# GCS paths must be of the form gs://<BUCKET>/<FOLDER>. If you specify a GCS path, be 
+# sure to set USE_CLOUDSQL=true.
 # Local paths must be absolute paths.
-# Can be same as INPUT_DIR.
+# Can be the same as INPUT_DIR.
 OUTPUT_DIR=
 
 ### DB parameters ###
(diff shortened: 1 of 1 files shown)


custom_dc/env.list.sample
```diff
@@ -48,7 +49,8 @@ GOOGLE_ANALYTICS_TAG_ID=
 
 ### Variables that rarely need to be changed ###
 
-# Enable embeddings generation and natural language querying.
+# Enable embeddings generation and natural language querying. If you find that container startup
```
keyurva, line 52: Let's not add the additional comment. They lose NL if they set it to false.

custom_dc/env.list.sample
```diff
@@ -12,9 +12,10 @@ MAPS_API_KEY=
 # Local paths must be absolute paths.
 INPUT_DIR=
 # Output directory in Google Cloud Storage (GCS) or local drive for generated csv files.
-# GCS paths must be of the form gs://<BUCKET>/<FOLDER>.
+# GCS paths must be of the form gs://<BUCKET>/<FOLDER>. If you specify a GCS path, be 
+# sure to set USE_CLOUDSQL=true.
```
keyurva, line 16: They need to set more than USE_CLOUDSQL. So how about: If you specify a GCS path, you need to use Cloud SQL.

custom_dc/env.list.sample
```diff
@@ -12,9 +12,10 @@ MAPS_API_KEY=
 # Local paths must be absolute paths.
 INPUT_DIR=
 # Output directory in Google Cloud Storage (GCS) or local drive for generated csv files.
-# GCS paths must be of the form gs://<BUCKET>/<FOLDER>.
+# GCS paths must be of the form gs://<BUCKET>/<FOLDER>. If you specify a GCS path, be
```
keyurva, line 15: nit: Perhaps start the additional sentence on a new line.

custom_dc/env.list.sample
kmoscoe: SG.

custom_dc/env.list.sample
kmoscoe: Will do.

custom_dc/env.list.sample
kmoscoe: Good point, I will actually point to relevant parameters section.

PR #4632, opened 2024-09-17
Makes url flag optional in screenshot generation script.

gmechali:
This PR fixes the run_website_screenshot.sh which today does not pass in a URL. The URL defaults to the domain, which is properly passed in from the script.


chejennifer (approved):
thanks!

PR #4635, opened 2024-09-18
Fix place redirection to NOT redirect for unfulfilled queries.

gmechali:
Queries with valid identified SVs but unfulfilled data also started redirecting to place page since monday (autopush only). This PR takes into account the svSource before performing the redirection.

However, since having the place overview tile as the only tile should remain a valid use case, this PR undoes a change to hide place overview when it's the only tile, which has the side effect of flashing the tile before redirection (view screencast: 
https://screencast.googleplex.com/cast/NTM3MjU4NDQzMzU0OTMxMnw1NjE3MGFjYi1iOA).
Resolving this from client-side redirect while still supporting the place overview tile seems quite hacky. Leaving the flashing map for now. Will follow up.


dwnoble (approved):
Good catch!

PR #4641, opened 2024-09-24
Sv explorer redirect

gmechali:
Fix Redirection on the autocomplete selection from the SV Explorer and Map Explorer (same component):

SV Explorer:
Before: https://screencast.googleplex.com/cast/NTg2NzQ1ODIxNTgwNDkyOHxlNmY4Y2RiMy0wOQ
After: https://screencast.googleplex.com/cast/NTI2ODYzMjQ2NzY2ODk5MnxmNDg0Mzk2NC05Nw
Map Explorer:
Before: https://screencast.googleplex.com/cast/NTI5MDk5NzE2NzY4NTYzMnw4OTAxMWQxMC02ZA
After: https://screencast.googleplex.com/cast/NTI2ODYzMjQ2NzY2ODk5MnxmNDg0Mzk2NC05Nw


beets:
could we trigger the "onclick" handler for that stat var instead? then it could work for timeline too and wouldn't rely on the url format of those specific tools

gmechali:
Yes sounds good. It looks like we just needed to trigger the the togglePath method in the handler. I modified it and put it in a new PR. closing this one.

PR #4642, opened 2024-09-25
US NHTSA FARS Crash NL

saanikaaa:
[SV Index Differ](https://storage.mtls.cloud.google.com/datcom-embedding-diffs/chharish_base_uae_mem_2024_09_25_05_23_28.html)

### DIFF
--- a/tools/nl/embeddings/input/base/sheets_svs.csv
+++ b/tools/nl/embeddings/input/base/sheets_svs.csv
@@ -3606,3 +3606,31 @@ sdg/SG_GEN_PARL.SEX--F,proportion of seats held by women in national parliaments
 Count_Person_Upto18Years,children population count
 Count_Person_18OrMoreYears,adult population count
 Count_Person_65OrMoreYears,senior population count
+Count_VehicleCrashIncident_CountyRoad,number of crash on county road
+Count_VehicleCrashIncident_Dark,number of crash in the dark;number of crash at night
+Count_VehicleCrashIncident_Daylight,number of crash in daylight;number of crash in day
+Count_VehicleCrashIncident_InterstateRoad,number of crash on interstate road
+Count_VehicleCrashIncident_LocalStreet,number of crash on local street
+Count_VehicleCrashIncident_OnRoadside,number of crash on roadside
+Count_VehicleCrashIncident_OnRoadway,number of crash on roadway
+Count_VehicleCrashIncident_RainOrMist,number of crash in rain or mist
+Count_VehicleCrashIncident_SleetOrHail,number of crash in sleet or hail
+Count_VehicleCrashIncident_Snow,number of crash in snow
+Count_Vehicle_InvolvedInCrash_InTransport,number of vehicles in transport involved in crash
+Count_VehicleCrashIncident_InvolvedInCrash_SchoolBus,number of school bus crash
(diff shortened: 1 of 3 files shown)


hareesh-ms:
Is the step to update the integration golden files executed by running ./run_test.sh -g ?

tools/nl/embeddings/input/base/sheets_svs.csv
```diff
@@ -3605,4 +3605,32 @@ worldBank/EG_ELC_ACCS_ZS,percentage of population with access to electricity
 sdg/SG_GEN_PARL.SEX--F,proportion of seats held by women in national parliaments
 Count_Person_Upto18Years,children population count
 Count_Person_18OrMoreYears,adult population count
-Count_Person_65OrMoreYears,senior population count
\ No newline at end of file
+Count_Person_65OrMoreYears,senior population count
+Count_VehicleCrashIncident_CountyRoad,number of crash on county road;number of motor vehicle accident on county road;number of road accident on county road
```
shifucun, line 3609: A few high level points:

1. Only add high level stat vars in this PR and restrict the addition to less than 10. For example, `Count_VehicleCrashIncident`, `Count_MortalityEvent_VehicleCrashIncident` are good ones. `Count_VehicleCrashIncident_YIntersection` is not a good candidate.
2. Try add one description for each stat var. Only add additional one if the description states very differently from the primary one. In this example, I would prefer "accident" over "crash". Description like `road accident` is very vague and not accurate.
3. Also add a test query that triggers the new stat var

tools/nl/embeddings/input/base/sheets_svs.csv
```diff
@@ -3605,4 +3605,12 @@ worldBank/EG_ELC_ACCS_ZS,percentage of population with access to electricity
...
 Count_Person_18OrMoreYears,adult population count
-Count_Person_65OrMoreYears,senior population count
\ No newline at end of file
+Count_Person_65OrMoreYears,senior population count
+Count_VehicleCrashIncident_Dark,number of vehicle accident in the dark
+Count_VehicleCrashIncident_Daylight,number of vehicle accident in the daylight
+Count_VehicleCrashIncident_RainOrMist,number of vehicle accident in rain or mist
+Count_VehicleCrashIncident_SleetOrHail,number of vehicle accident in sleet or hail
+Count_VehicleCrashIncident_Snow,number of vehicle accident in snow
+Count_VehicleCrashIncident_InvolvedInCrash_SchoolBus,number of school bus accident
```
chejennifer, line 3614: to be consistent with the rest of the descriptions, maybe something like "number of vehicle accident with school bus" or "number of vehicle accident involving school bus"

tools/nl/embeddings/input/base/sheets_svs.csv
saanikaaa: 1. Restrict the addition to less than 10.
2. Modified to one description.  Updated "accident" over "crash".
3. Added test query.

tools/nl/embeddings/input/base/sheets_svs.csv
saanikaaa: Modified to "number of vehicle accident involving school bus"

chejennifer (commented):
thanks for updating! The link to the sv differ results seems to be broken, can you please fix that? And please also fix the merge conflicts

PR #4644, opened 2024-09-26
Re-introduce message for topic selected from previous query.

gmechali:
Jennifer caught regressions from goldens. This message is still necessary for queries that use the previous query's topic.


chejennifer (approved):
thanks for fixing!

PR #4661, opened 2024-10-10
Custom DC Testing

gmechali:
Factor out base dc tests for homepage, download tool, map and browser to be shared across base dc + custom dc.

In the future, all tests to be shared across Base DC + Custom DC should be added to these shared_tests/ files, while tests specific to CDC or Base DC can remain under their respective directory files.

Limiting to those 4 test files for now, will do the rest in a follow up.
Part of Fixit: b/368385125


server/webdriver/cdc_tests/autopush/download_test.py
```diff
@@ -0,0 +1,44 @@
...
+from server.webdriver.base import WebdriverBaseTest
+from server.webdriver.cdc_tests.base import CDC_AUTOPUSH_URL
+from server.webdriver.shared_tests.download_test import DownloadTestMixin
+
+
+# Class to test the download tool for Custom DC
+class TestDownload(DownloadTestMixin, WebdriverBaseTest):
+  DATACOMMONS_STRING = "Custom Data Commons"
+
+  def get_server_url(self):
```
keyurva, line 26: Perhaps create a `CdcAutopushTestBase` (or similarly named) class that provides these common fields and methods and have all CDC tests extend it?

server/webdriver/cdc_tests/autopush/download_test.py
```diff
@@ -0,0 +1,44 @@
...
+import tempfile
+
+from server.webdriver.base import WebdriverBaseTest
+from server.webdriver.cdc_tests.base import CDC_AUTOPUSH_URL
+from server.webdriver.shared_tests.download_test import DownloadTestMixin
+
+
+# Class to test the download tool for Custom DC
+class TestDownload(DownloadTestMixin, WebdriverBaseTest):
+  DATACOMMONS_STRING = "Custom Data Commons"
```
keyurva, line 24: nit: instance fields are typically `lower_snake_case`. Also is there a more expressive name for this?

keyurva (commented):
This is cool - in that it bases both base and custom dc tests on a common core.

It however makes this a much bigger change and we should add FE stakeholders as reviewers - @dwnoble @beets

server/webdriver/cdc_tests/autopush/download_test.py
gmechali: Done!

server/webdriver/cdc_tests/autopush/download_test.py
gmechali: Done. I changed to dc_title_string, but open to suggestions!

gmechali:
So when I looked at what tests to add to make the CDC tests more robust, it was basically that we should add nearly all tests that we have on base dc +/- a couple. some functionality isnt in CDC, and we would want to add some tests such as checking for custom SVs.
But then it seemed silly to just replicate the tests, especially since it would require us to dupe every test in the future or risk CDC and base dc tests getting out of sync again. This way, we would control which tests are shared vs just Base vs just CDC. 

Let me know what you all think about this, and if I should continue with the other tests! I think this only ~30% of all tests.

dwnoble:
I like the idea of extracting these tests to the shared_tests directory. But, it looks like now we still need some boilerplate files in both `webdriver/cdc_test/autopush` and in the `webdriver/tests` to pick up the tests in `webdriver/shared_tests`. This seems a little error-prone to me. Would it be possible to turn the shared_test mixins to proper tests, and then run the tests from the two environments something like:

```
python3 -m pytest server/webdriver/cdc_tests/ server/webdriver/shared_tests/
```

```
python3 -m pytest server/webdriver/tests/ server/webdriver/shared_tests/
```

server/webdriver/cdc_tests/base.py
```diff
@@ -0,0 +1,16 @@
+# Copyright 2023 Google LLC
```
dwnoble, line 1: Nit: Use 2024 across the board for new files

dwnoble (approved):
After chatting offline, let's keep the current approach with the shared_tests mixin files

keyurva (approved):
Thanks Gabe! The shared tests will improve coverage and alignment with base dc.

beets (commented):
thanks for the update and improving our test coverage!

server/webdriver/base_dc_webdriver.py
```diff
@@ -0,0 +1,19 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from server.webdriver.base import WebdriverBaseTest
+
+
+class BaseDcWebdriverTest(WebdriverBaseTest):
+  """Base class for CDC Autopush webdriver tests."""
```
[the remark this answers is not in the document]
gmechali: ah yes thank you! Fixed in the follow up CL which im about to send out!

PR #4666, opened 2024-10-14
Fixes bug that sets the query parameter as placeholder only.

gmechali:
Likely introduced when with my autocomplete FE PR. The result of this bug was that once on the explore page, you would see the previous query in the input bar, but whenever you clicked on it, the query would disappear as it was only set as a placeholder, not the input text.
With this PR, we bring it back to the same state as prod.


dwnoble (approved):
Thanks for the fix!

PR #4667, opened 2024-10-14
Modify the autocomplete processing logic to improve results

gmechali:
Negates the "score", best match has lowest score. This allows us to sort with the first results being the best matches, but preserves the Google Maps ordering for same score responses.
Cleans up the logic a bit to remove dupes at the end. The new sorting requires computing all scores, so this deduping at the end doesnt lead to useless processing.
When computing score, take into account the ordering of the words.
This significantly improves the results from my testing.

### DIFF
--- a/server/routes/shared_api/autocomplete/helpers.py
+++ b/server/routes/shared_api/autocomplete/helpers.py
@@ -23,7 +23,6 @@ import requests
 MAPS_API_URL = "https://maps.googleapis.com/maps/api/place/autocomplete/json?"
 MIN_CHARACTERS_PER_QUERY = 3
 MAX_NUM_OF_QUERIES = 4
-RESPONSE_COUNT_LIMIT = 20
 DISPLAYED_RESPONSE_COUNT_LIMIT = 5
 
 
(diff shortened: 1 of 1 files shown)


server/routes/shared_api/autocomplete/helpers.py
```diff
@@ -71,97 +70,72 @@ def execute_maps_request(query: str, language: str) -> Dict:
...
-          # print(bm)
-          duplicates[pred['place_id']] = bm
-        else:
-          # print("We're just getting our first dupe.")
-          duplicates[pred['place_id']] = query
-
-      if len(responses) >= RESPONSE_COUNT_LIMIT:
-        # prevent new loop to iterate through next answer.
-        break
+      result_pred = {}
```
dwnoble, line 117: Consider making this a dataclass in a follow-up

dwnoble (approved):
Nice!

server/routes/shared_api/autocomplete/helpers.py
gmechali: Yes still on my todo list!

PR #4669, opened 2024-10-15
Complete test parity for Custom DC and Base DC

gmechali:
Follow up from: https://github.com/datacommonsorg/website/pull/4661

This PR replicates the same logic we added in PR above to share test code between base DC and custom DC, for all remaining test files.
There are currently 57 tests that run for Base DC. 
Now 51 of those also run for Custom DC.


server/webdriver/shared_tests/stat_var_hierarchy_test.py
```diff
@@ -0,0 +1,91 @@
...
+    """Test the stat var filtering based on place selection."""
+    self.driver.get(self.url_ + MAP_URL)
+
+    # Wait until stat var hierarchy is present
+    element_present = EC.presence_of_element_located(
+        (By.ID, 'hierarchy-section'))
+    WebDriverWait(self.driver, self.TIMEOUT_SEC).until(element_present)
+
+    # Get the count of the first category
+    agriculture_category = self.driver.find_element(
```
gmechali, line 39: FYI this is the only test I had to modify. It previously just checked the first grouping of statVars, which on Custom DC, was the OECD group. Then when we added the filter, it removed that group since there was no data for the filter, so it compared the SV counts from 2 different groupings.
I changed the test so it's always looking at the same grouping of SVs.

dwnoble (approved):
Thanks for migrating these!

run_cdc_tests.sh
```diff
@@ -20,4 +20,4 @@ set -e
 source .env/bin/activate
 export FLASK_ENV=webdriver
 
-python3 -m pytest server/webdriver/cdc_tests/
+python3 -m pytest --reruns 2 server/webdriver/cdc_tests/
```
keyurva, line 23: Curious about the 2 reruns. Is this because we have more tests now and this mitigates some of the flakiness?

keyurva (approved):
Thanks for bringing test parity to custom dc!

run_cdc_tests.sh
gmechali: Yes it's also mainly the type of tests. There are a few tests that have some flakiness (think it was the scatter visualization ,and those for which we enter the settings manually). They're also the ones that tend to flake for base dc.
Base DC tests have a 2 reruns already enabled. So since it's the same tests and there was some flakiness, I matched the reruns.

PR #4670, opened 2024-10-15
Add Typing to autocomplete API + Fix bug where comma affects the scoring, moving down a better match

gmechali:
Adds an AutocompleteResult and AutoCompleteApiResponse type.
Fixes bug identified by Dan in sheet (https://docs.google.com/spreadsheets/d/1VQwmMeGqFJkpNDXrSUtNeQ2jpl7PP4LZ-xfARaM8HPc/edit?resourcekey=0-kCZy9U9hA0U-UEAp0qMMsw&gid=0#gid=0) that downgrades better matches because of the comma. In our processing an exact match is valued more highly, but we only split on spaces. This PR also splits on spaces + commas.
In the examples of California, the states now have the same score as "California City" and therefore the Google Maps Prediction API ordering is preserved.


dwnoble (approved):
Nice!

PR #4675, opened 2024-10-18
Fix Autopush FR test

gmechali:
The test string contained the data's date "2022". With the updates to our data, this was bound to start failing.


keyurva (approved):

gmechali:
Jennifer already submitted the fix. Closing.

beets (commented):
This is still a good change to have - but we should not include the changes to update import, mixer and the pb2 files.

PR #4678, opened 2024-10-21
Adds logging for autocomplete responses.

gmechali:
Logs the response count for autocompletion. Staging is not showing any responses. Would like to better understand where the breakdown is occurring.


server/routes/shared_api/autocomplete/helpers.py
```diff
@@ -69,6 +70,7 @@ def execute_maps_request(query: str, language: str) -> Dict:
       'language': language
   }
   response = requests.post(MAPS_API_URL + urlencode(request_obj), json={})
+  logging.info("Executed Google Maps Prediction API request.")
```
chejennifer, line 73: wonder if we need this because if the request fails, this should throw an error & you have another logging in place to count the number of predictions

server/routes/shared_api/autocomplete/autocomplete.py
```diff
@@ -51,6 +52,7 @@ def autocomplete():
   place_id_to_dcid = []
   if place_ids:
     place_id_to_dcid = json.loads(findplacedcid(place_ids).data)
+  logging.info("Found %d place ID to DCID mappings.", len(place_id_to_dcid))
```
chejennifer, line 55: should we also add a keyword like "place_autocomplete" or something at the beginning of each logging message so we can filter logs

gmechali: done.

server/routes/shared_api/autocomplete/helpers.py
gmechali: done.

PR #4682, opened 2024-10-21
Exit cdc_services/run.sh when any background process exits

hqpho:
This makes startup errors in Mixer or NL servers more obvious.

Bug: b/374820494
Reference: https://docs.docker.com/engine/containers/multi-service_container/#use-a-wrapper-script

### DIFF
--- a/build/cdc_services/run.sh
+++ b/build/cdc_services/run.sh
@@ -88,7 +88,7 @@ else
 fi
 
 # Wait for any process to exit
-wait
+wait -n
 
 # Exit with status of process that exited first
 exit $?


build/cdc_services/run.sh
```diff
@@ -88,7 +88,7 @@ else
 fi
 
 # Wait for any process to exit
-wait
+wait -n
```
keyurva, line 91: In the happy case, we want this to block until all services have started. Checking if this continues to be the case after this change?

hqpho: Confirmed! The behavior of this script is the same in the happy case - confirmed by using a bind mount with datacommons-services:stable to replace run.sh with the new version and comparing the command output to the output without the bind mount.

keyurva: Thank you!

keyurva (approved):

PR #4683, opened 2024-10-22
Accessibility Updates

nick-nlb:
## Description

This PR implements improvements following an accessibility sweep of the home page, and includes the following fixes and improvements:

 * significant reworking of the carousel's tab handling; it now handles switching between slides by tab graciously regardless of the content.
 * addition of aria labels and corrections to existing aria labels.
 * the media text component can now take an optional alt prop in lieu of (or in addition to) the title, with at least one of them being required.
 * The headers in the rich menus are adjusted to better match the hierarchy.
 * tab outlines are fixed for the main menu.
 * Ids that could be duplicated have been removed. In the case of the search bar, we could not remove the id at this stage (because it is tied to downstream functionality) but we are now selectively rendering the mobile and desktop menu to prevent the duplicate id. To facilitate this selective rendering, we have added a breakpoints hook that allows us to apply Bootstrap-like media queries on the TypeScript side. This breakpoints hook may proof to be a generally useful utility.

### DIFF
--- a/static/js/apps/base/components/header_bar/header_bar.tsx
+++ b/static/js/apps/base/components/header_bar/header_bar.tsx
@@ -66,9 +66,7 @@ const HeaderBar = ({
             labels={labels}
             routes={routes}
           />
-          {showHeaderSearchBar && up("lg") && (
-            <HeaderBarSearch inputId={"query-search-input-desktop"} />
-          )}
+          {showHeaderSearchBar && up("lg") && <HeaderBarSearch />}
           <MenuDesktop menu={menu} labels={labels} routes={routes} />
         </div>
         <div className="navbar-menu-mobile">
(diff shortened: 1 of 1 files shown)


static/js/shared/hooks/breakpoints.tsx
```diff
@@ -0,0 +1,115 @@
...
+/*
+ For all comparison functions (up, down, only), the given size ("md", "lg", etc.) refers to
+ the screen size range, not just a single breakpoint. This aligns with the equivalent Bootstrap media queries.
+
+ - `up("md")`: Returns true if the window is at least in the "md" screen size range (768px or larger by default).
+ - `down("md")`: Returns true if the window is in the "md" screen size range or smaller (i.e., smaller than the "lg" screen size).
+ - `only("md")`: Returns true if the window is within the "md" screen size range (768px or larger and smaller than 1068px).
+ */
+
+interface BreakpointInterface<B extends Breakpoints = Breakpoints> {
```
dwnoble, line 31: Very helpful!

dwnoble (approved):

PR #4685, opened 2024-10-22
update nodejs goldens

chejennifer:
goldens needed to be updated because of a bunch of recent data updates (data diffs can be seen here: https://github.com/datacommonsorg/mixer/pull/1438, https://github.com/datacommonsorg/mixer/pull/1439)


beets (approved):
thanks for the update

could you add some commentary to the pr description for why the goldens required to be updated (we should be logging these for future reference)

beets:
thanks for updating the description!

PR #4690, opened 2024-10-25
Improves Typo recognition for autocomplete

gmechali:
This PR modifies the scoring algorithm for place autocomplete to count a small score for non-exact matches, to account for one typo.
With these changes, we will favor "San Diego" over "Dieppe" for the query "Sna Die".
Prod: https://screenshot.googleplex.com/Bsx2BbyLZArbQuX
Local with this change: https://screenshot.googleplex.com/9jHqKb2uHJLz37k

Note that "Sne Die" will still go back to "Dieppe" because that's 2 typos, so San Diego is out even if it was returned by google Maps predictions: https://screenshot.googleplex.com/9LViJoVFni3Lui6

Typo check done as a bag of letters with at most off by one. We do this check on top of the Google Maps Predictions which already take into account typo correction. This part is just to choose the best prediction from google maps.

Doing this as part of gaps identified in place autocomplete: https://docs.google.com/document/d/15RVckX9ck5eyyhBHW8Nb9lmxPBDPMIeLbax14HbN-GI/edit?tab=t.0


server/routes/shared_api/autocomplete/helpers.py
```diff
@@ -73,6 +73,39 @@ def execute_maps_request(query: str, language: str) -> Dict:
...
+    dict: A dictionary where keys are letters and values are their counts.
+    """
+  bag = {}
+  for char in text.lower():
+    if char.isalpha():
+      bag[char] = bag.get(char, 0) + 1
+  return bag
+
+
+def off_by_one_letter(str1_word: str, name_word: str) -> bool:
```
dwnoble, line 88: Have you looked into [Levenshtein distance](https://en.wikipedia.org/wiki/Levenshtein_distance) as another approach? Levenshtein distance takes letter order into account. not sure if that's helpful.

server/routes/shared_api/autocomplete/helpers.py
```diff
@@ -125,6 +166,7 @@ def predict(queries: List[str], lang: str) -> List[ScoredPrediction]:
   all_responses.sort(key=get_score)
   logging.info("[Place_Autocomplete] Received %d total place predictions.",
                len(all_responses))
+  print(all_responses)
```
dwnoble, line 169: remove print statement

server/routes/shared_api/autocomplete/helpers.py
dwnoble, line 88: Can you add a unit test for this function?

dwnoble (approved):

server/routes/shared_api/autocomplete/helpers.py
gmechali: Done.

server/routes/shared_api/autocomplete/helpers.py
gmechali: Added a TODO to look into it. I ll take another look at it before we talk about launching this :)

gmechali: Done.

PR #4692, opened 2024-10-29
Fix nl server start up in webdriver tests

chejennifer:
Recently webdriver tests have sometimes failed due to NL server start up issues. 

This, I believe has to do with the fact that we use multiprocessing for webdriver tests & when we start up nl server, we only download the folder for a model if it's not already there. So one process may have started downloading the folder for a model & another process sees that the folder for the model is already there and attempts to use a file from the folder that hasn't completed downloading

Here, I add another step in the webdriver cloudbuild to first download the model files before starting the webdriver tests

Testing:

- These changes do not trigger webdriver tests, so I added a print statement to trigger the tests on a previous commit & tests passed: https://pantheon.corp.google.com/cloud-build/builds;region=global/5ba92206-1df7-4dfc-9f79-7e7d04d595bf;step=2?e=13803378&inv=1&invt=AbgLkA&mods=-monitoring_api_staging&project=datcom-ci
- This change also affects the nl_server docker image, so I tested building the image and pushing the image to dev and it seems to work.


dwnoble (approved):
Nice!

beets (approved):
nice, thanks jennifer!

PR #4693, opened 2024-10-30
Adds Continents to AutoComplete

gmechali:
This PR implements a hack to add autocomplete for continents. We always fake that Continents came back in the response from the Google Maps Predictions API response, and hardcode the Place ID to DCID.

The Place IDs were fetched from here: https://developers.google.com/maps/documentation/places/web-service/place-id

For one word continent, we just add a matching for the single word query. 
For two word continents (S|N America), we add matching for the single word query, and two word query if applicable.

We then compute the match score, and remove all continents with positive scores (negative score means a better match).

Note that we prepend the continents in order to give the continents an edge over other places that have the same score.
I also added a boost to scoring if the query matches the response from the start in order to give "America" an edge over "North America" when typing "Ameri".


server/routes/shared_api/autocomplete/helpers.py
```diff
@@ -181,6 +259,29 @@ def predict(queries: List[str], lang: str) -> List[ScoredPrediction]:
   return responses
 
 
+def fetch_place_id_to_dcid(
+    prediction_responses: List[ScoredPrediction]) -> Dict:
```
dwnoble, line 263: Can the response type be more precise? (is `Dict[str, str]`right?)

server/routes/shared_api/autocomplete/helpers.py
```diff
@@ -22,11 +22,41 @@
...
 from server.routes.shared_api.autocomplete.types import ScoredPrediction
+from server.routes.shared_api.place import findplacedcid
 
 MAPS_API_URL = "https://maps.googleapis.com/maps/api/place/autocomplete/json?"
 MIN_CHARACTERS_PER_QUERY = 3
 MAX_NUM_OF_QUERIES = 4
 DISPLAYED_RESPONSE_COUNT_LIMIT = 5
+TWO_WORD_CONTINENTS = [{
+    'description': 'North America',
+    'place_id': 'ChIJnXKOaXELs1IRgqNhl4MoExM'
```
dwnoble, line 33: Can drop place_ids entirely from from these constants? Looks like they're only used in the internal data structure wrangling

server/routes/shared_api/autocomplete/helpers.py
```diff
@@ -22,11 +22,41 @@
 import requests
 
 from server.routes.shared_api.autocomplete.types import ScoredPrediction
+from server.routes.shared_api.place import findplacedcid
 
 MAPS_API_URL = "https://maps.googleapis.com/maps/api/place/autocomplete/json?"
 MIN_CHARACTERS_PER_QUERY = 3
 MAX_NUM_OF_QUERIES = 4
 DISPLAYED_RESPONSE_COUNT_LIMIT = 5
+TWO_WORD_CONTINENTS = [{
```
dwnoble, line 31: I could see us wanting to match other places at some point that aren't accounted for by the Google Maps API. Maybe instead of calling this `CONTINENTS`, it could be something like `ADDITIONAL_PLACES` ?

server/routes/shared_api/autocomplete/helpers.py
```diff
@@ -22,11 +22,41 @@
...
+    'description': 'Oceania',
+    'place_id': 'ChIJQbA4_Cu8QW4RbuvrxISzaks'
+}, {
+    'description': 'Africa',
+    'place_id': 'ChIJ1fWMlApsoBARs_CQnslwghA'
+}, {
+    'description': 'Asia',
+    'place_id': 'ChIJV-jLJIrxYzYRWfSg0_xrQak'
+}] + TWO_WORD_CONTINENTS
+CONTINENT_PLACE_ID_TO_DCID = {
```
dwnoble, line 51: Can we drop this mapping entirely and just have place names + DCIDS?

server/routes/shared_api/autocomplete/helpers.py
dwnoble, line 31: Can we just have a single list of additional places instead of one-word and two-word places?

server/routes/shared_api/autocomplete/helpers.py
gmechali: yes and done.

server/routes/shared_api/autocomplete/helpers.py
gmechali: So I specifically named it continents because I think we shouldn't be doing this for anything else. This is just a stopgap because we can't launch place autocomplete without continents. But I think we should use whatever we will for StatVars for everything custom in the future.
I can still do the renaming, but this is pretty hacky, so I do want to emphasize we shouldnt make this a precedent for throwing in other things :)

server/routes/shared_api/autocomplete/helpers.py
gmechali: We can - doing it this was just plugs in better in the existing code - I just add those responses and modify the place id to dcid map, and all the other logic is the same for fetching the score, extracting the data into the response etc.
 
That's why I did this way, but I can also make it more custom logic for continents if you prefer and bypass that intermediary. I just worry that it ll be a bigger divergence and will be more code on a separate branch.

server/routes/shared_api/autocomplete/helpers.py
gmechali: The reason I did that is to account the continent hacking properly.
We want to evaluate "Africa" only against one word queries, while "North america" should be evaluated against both one word queries and two word queries in order to account for the queries "North Ame" and "Nort".

If I just have one list of continents, then Id need to do some processing on it to see if there's a space to see if I add it to the two word query too. This seemed a bit cleaner to me, but lmk if you disagree.

server/routes/shared_api/autocomplete/helpers.py
gmechali: I think this is the same as the comment two comments down. closing this one - keeping the other open for the discussion.

dwnoble (approved):

PR #4699, opened 2024-11-01
Refactor continent hack into custom places, and add 'world' and 'earth'

gmechali:
Adds World and Earth to the autocomplete.
Refactors the continent hack into a "Custom places" hack where we no longer rely on the place ID intermediary.


dwnoble (approved):
Thanka Gabe!

PR #4700, opened 2024-11-02
Knowledge graph

nick-nlb:
## Description

This PR implements the redesign of the knowledge graph (browser landing) page.

It consists of both a conversion of the page itself to a React app with components (along with the movement of the source data that populates the page into json, similar to comparable pages), and a reworking of the page itself into the new layout.

It includes a general component (the intro) that will be used in other similar redesigns, as well as components specific to the knowledge graph page.


dwnoble (approved):
Thank you!

PR #4705, opened 2024-11-04
Fix size of Custom DC default logo

hqpho:
The height attribute has not been applied since https://github.com/datacommonsorg/website/pull/4683 changed `#main-header-logo` to `.main-header-logo`.


keyurva (approved):
Thanks for fixing this!

PR #4709, opened 2024-11-05
Improvements to header

nick-nlb:
## Description

This PR follows up from previous changes to the autocomplete, in order to fix the margins around the box in some views.

The header search bar now displays in the correct place in all views, including with the autocomplete.

Some additional refactoring of the styles was done, including the beginning of a more complete separation of the new home page styles from the template custom styles.

As an example of one of the adjustments: 

Before: 

before

After:

Screenshot 2024-11-04 at 8 01 06 PM


dwnoble (approved):

PR #4714, opened 2024-11-05
Add wait for loading before SV group click in timeline test

hqpho:
Hopefully this will resolve custom DC autopush test failures.


server/webdriver/shared_tests/timeline_test.py
```diff
@@ -127,6 +127,7 @@ def test_check_statvar_and_uncheck(self):
     self.assertEqual(len(charts), 0)
 
     # Expand the Demographics section of the stat var hierarchy.
+    shared.wait_for_loading(self.driver)
```
gmechali, line 130: Did that work to remove the flakiness?

otherwise you can do 
charts_present = EC.presence_of_element_located((By.CLASS, 'node-title'))
WebDriverWait(self.driver, self.TIMEOUT_SEC).until(charts_present)

Which will specifically wait until that class is present - or the timeout.

gmechali (approved):

PR #4715, opened 2024-11-05
[bio] some fulfillment improvements

chejennifer:
- when multiple entities are recognized but only some entities have data for the property, should only show the entities with data (e.g., [autopush](https://screenshot.googleplex.com/48VaUAxjZtXYMgC) vs [local](https://screenshot.googleplex.com/Brsnuj4DSNmpJSN))
- fix bug in deciding when to show low confidence message. Previously did not consider property score


clincoln8 (approved):

beets (approved):

chejennifer:
thanks for the reviews!

PR #4717, opened 2024-11-06
Explore landing Update

nick-nlb:
## Description

This PR implements the redesign of the explorer landing page page.

The first part of the PR converts the page to a  React app with components followed by a reworking of the page into the new layout.

The link content for the stat var chips have been added to the same `TopicConfig` as the general and comparison query links.

The link chip component has been generalized to contain a variant for the stat var chips.


static/js/components/content/link_chips.tsx
```diff
@@ -38,28 +38,34 @@ export interface LinkChip {
...
+            <li key={linkChip.id} className={`chip-item ${variant}`}>
               <a
                 href={linkChip.url}
                 onClick={(): void => {
                   triggerGAEvent(GA_EVENT_HOMEPAGE_CLICK, {
-                    [GA_PARAM_ID]: `topic ${linkChip.id}`,
+                    [GA_PARAM_ID]: `${section} ${linkChip.id}`,
                     [GA_PARAM_URL]: linkChip.url,
                   });
                   window.location.href = linkChip.url;
```
dwnoble, line 71: It looks like command or control clicking on these links doesn't open the link in a new tab. Can you look into fixing that? We have some GA logging here, but im not sure that's strictly necessary - GA might automatically log all link clicks regardless.

PR #4718, opened 2024-11-06
Truncate golden CosineScores to 6 decimals to reduce noisy diffs.

clincoln8:
Reduces CosineScores to 6 decimals to reduce the diffs that are generated when updating the integration_test goldens.

This is to reduce noise in reviews and should not fundamentally affect the actual tests.


server/integration_tests/explore_test.py
```diff
@@ -120,12 +120,22 @@ def handle_response(self,
...
                                 key=lambda x: (-resp_var_to_score.get(x, 0), x))
       resp['variables'] = sorted_variables
 
+    # Truncate CosineScores to 6 decimals to reduce noisy diffs.
+    for candidate in dbg['sv_matching']['MultiSV'].get('Candidates', []):
+      for part in candidate.get('Parts', []):
+        for i, score in enumerate(part.get('CosineScore', [])):
+          part['CosineScore'][i] = float("{:.6f}".format(score))
+    for i, score in enumerate(dbg['props_matching'].get('CosineScore', [])):
+      dbg['props_matching']['CosineScore'][i] = float("{:.6f}".format(score))
```
chejennifer, line 137: nit: instead of re-writing this formatting code, could we have a function to format_scores which takes a list of scores and returns a list of formatted scores, so that if we ever want to change how we format the scores, we do it in one place and won't run the risk of missing a spot?

chejennifer (commented):
thanks for updating!

server/integration_tests/explore_test.py
clincoln8: Agreed! I reformatted to use helper functions, is that better?

server/integration_tests/explore_test.py
```diff
@@ -120,12 +120,22 @@ def handle_response(self,
     if 'variables' in resp:
       resp_var_to_score = {}
       for i, sv in enumerate(dbg['sv_matching']['SV']):
-        score = dbg['sv_matching']['CosineScore'][i]
-        resp_var_to_score[sv] = float("{:.6f}".format(score))
+        truncated_score = _format_score(dbg['sv_matching']['CosineScore'][i])
+        resp_var_to_score[sv] = truncated_score
+        dbg['sv_matching']['CosineScore'][i] = truncated_score
```
chejennifer, line 125: maybe we can move this part outside of the for loop & just have `dbg['sv_matching']['CosineScore'] = _format_scores(dbg['sv_matching'].get('CosineScore', [])`

& that way we just need one of the two helper functions & can read score from this formatted list

clincoln8: done!

PR #4719, opened 2024-11-06
Webdriver test fixes and improvements

hqpho:
- Remove some landing page example links that don't work anymore
- Add parallelization for Custom DC autopush tests
- Add some waiting for elements to be ready in tests that were failing on Cloud Build

PR #4720, opened 2024-11-07
Replace CO2 emissions example links

hqpho:
The previous links were removed in https://github.com/datacommonsorg/website/pull/4719


beets (approved):
thanks for updating these. the link for the new viz tool doesn't work for me - does it work locally for you? if so - lgtm

hqpho:
> thanks for updating these. the link for the new viz tool doesn't work for me - does it work locally for you? if so - lgtm

Yes, both new links work locally!  
[Screen recording](https://github.com/user-attachments/assets/482480b6-7a28-4499-bcbb-196066616e9a)

PR #4722, opened 2024-11-07
[NL] Disable llm fallback for entity based queries.

clincoln8:
Disabling the LLM fallback for entity based queries for now since the prompt is tailored to place+statvar queries.


server/lib/nl/detection/llm_fallback.py
```diff
@@ -122,6 +126,10 @@ def _has_no_entity(d: Detection) -> bool:
   return not d.places_detected or not d.places_detected.entities_found
 
 
+def _has_entity(d: Detection) -> bool:
```
chejennifer, line 129: nit: I don't think we need this helper since there's not really any logic here

chejennifer (approved):
thanks!

server/lib/nl/detection/llm_fallback.py
clincoln8: The double negative is a little rough for readability -- I updated the code to only have _has_entity to avoid this issue. Is that okay?

clincoln8: I'll merge this for now so that the logic is in autopush. I can change the readability part if you prefer the _has_no_entity with the double negative as a follow up!

PR #4724, opened 2024-11-08
Update goldens.

keyurva:


ajaits (approved):
deletions are due to deprecated who imports.

server/integration_tests/test_data/detection_api_multivar/howarefactorslikeobesity,bloodpressureandasthmaimpactedbyclimatechange/debug_info.json
```diff
@@ -73,7 +73,7 @@
                 0.746364,
                 0.740165,
                 0.735123,
-                0.713099
+                0.7131
```
chejennifer, line 76: curious why this score got truncated to 4 digits @clincoln8

chejennifer (approved):
bio changes are also expected from data update

server/integration_tests/test_data/detection_api_multivar/howarefactorslikeobesity,bloodpressureandasthmaimpactedbyclimatechange/debug_info.json
clincoln8: These scores are written as floats, not strings so trailing 0's are not preserved.

So in this case, the original truncated number must have been "0.713100", but since it's a float the last two 0's are dropped.

We could write them as strings if you prefer to keep all 6 digits.

PR #4725, opened 2024-11-08
fix sanity test

chejennifer:
looks like the class name for topic items got changed in this PR causing sanity tests to fail: https://github.com/datacommonsorg/website/pull/4673/files#diff-37059329da755a92aa218821e795bf1cb89ad26682ceb8f6d5dddc84a49bb645L43


keyurva (approved):
Thanks for fixing this!

PR #4731, opened 2024-11-12
Desensitize hiding the autocomplete results on scroll.

gmechali:
Only hides the autocomplete results after 15% of window width has been scrolled through. 
https://screencast.googleplex.com/cast/NjMyMDgzNjY5MTYyMzkzNnxmNjAxODU2Yy0xZQ

### DIFF
--- a/static/js/components/nl_search_bar/auto_complete_input.tsx
+++ b/static/js/components/nl_search_bar/auto_complete_input.tsx
@@ -79,22 +79,40 @@ export function AutoCompleteInput(
   const [inputActive, setInputActive] = useState(false);
   const [lastAutoCompleteSelection, setLastAutoCompleteSelection] =
     useState("");
+  // Used to reduce sensitivity to scrolling for autocomplete result dismissal.
+  const [lastScrollYOnTrigger, setLastScrollYOnTrigger] = useState(0);
+  const [lastScrollY, setLastScrollY] = useState(0);
 
   const isHeaderBar = props.barType == "header";
   let lang = "";
 
   useEffect(() => {
     // One time initialization of event listener to clear suggested results on scroll.
-    // It allows the user to navigate through the page without being annoyed by the results.
     window.addEventListener("scroll", () => {
-      if (results.placeResults) {
-        setResults({ placeResults: [], svResults: [] });
-      }
+      setLastScrollY(window.scrollY);
     });
 
     const urlParams = new URLSearchParams(window.location.search);
     lang = urlParams.has("hl") ? urlParams.get("hl") : "en";
   }, []);
+
+  // Whenever any of the scrollY states change, recompute to see if we need to hide the results.
+  // We only hide the results when the user has scrolled past 15% of the window height.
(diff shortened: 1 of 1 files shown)


static/js/components/nl_search_bar/auto_complete_input.tsx
```diff
@@ -79,23 +79,41 @@ export function AutoCompleteInput(
   const [inputActive, setInputActive] = useState(false);
   const [lastAutoCompleteSelection, setLastAutoCompleteSelection] =
     useState("");
+  // Used to reduce sensitivity to scrolling for autocomplete result dismissal.
```
chejennifer, line 82: can you add what each of these states are actually keeping track of

static/js/components/nl_search_bar/auto_complete_input.tsx
```diff
@@ -79,23 +79,41 @@ export function AutoCompleteInput(
...
+      results.placeResults.length > 0 &&
+      Math.abs(lastScrollY - lastScrollYOnTrigger) > window.outerHeight * 0.15
+    ) {
+      setResults({ placeResults: [], svResults: [] });
+    }
+  }, [
+    lastScrollY,
+    setLastScrollY,
+    lastScrollYOnTrigger,
+    setLastScrollYOnTrigger,
```
chejennifer, line 114: nit: since the two set methods are not used here, don't need to include them in the dependencies

static/js/components/nl_search_bar/auto_complete_input.tsx
```diff
@@ -79,23 +79,41 @@ export function AutoCompleteInput(
...
-      }
+      setLastScrollY(window.scrollY);
     });
 
     const urlParams = new URLSearchParams(window.location.search);
     lang = urlParams.has("hl") ? urlParams.get("hl") : "en";
   }, []);
 
+  // Whenever any of the scrollY states change, recompute to see if we need to hide the results.
+  // We only hide the results when the user has scrolled past 15% of the window height.
```
chejennifer, line 100: maybe add a note that this is when user has scrolled 15% of the window height from where they were when the last autocomplete triggered

chejennifer (commented):
nice, just some nits on adding more documentation

static/js/components/nl_search_bar/auto_complete_input.tsx
gmechali: done.

static/js/components/nl_search_bar/auto_complete_input.tsx
gmechali: done.

static/js/components/nl_search_bar/auto_complete_input.tsx
gmechali: I think that's the next line, just clarified the wording that it's since the autocomplete trigger.

chejennifer (approved):
thanks for updating!

PR #4732, opened 2024-11-13
Place redirect smoothing

nick-nlb:
## Description

This PR updates aspects of the flow, layout and styling of the redirect process for when a the search results give a "place overview" and the svSource is "unfulfilled", with the goal of removing redirect flashes and FOUCs.

The following changes were made to make the process visually smoother:
* In the event that the a page is poised to redirect (i.e., the above criteria is true), the loading state is maintained. This prevents the intermediate "Place Overview" page from displaying just before the redirect, removing the first flash.
* In the place page that follows (for example, `place/country/CAN`), the layout was adjusted so that a FOUC was removed while the content was being loaded in.
* The page overview now fades in when it loads, softening the appearance of the load.
* A small spinner was added onto this page to give more of a loading sense.

### DIFF
--- a/server/templates/place.html
+++ b/server/templates/place.html
@@ -69,7 +69,11 @@ limitations under the License.
       <div id="place-summary" class="row-col place-summary-container">{{ place_summary }}</div>
       <div id="main-pane" class="row"></div>
       {# TRANSLATORS: A message shown on the page while the content is loading. #}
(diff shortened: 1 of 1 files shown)


server/templates/place.html
```diff
@@ -69,7 +69,7 @@ <h3 id="locale" data-lc="{{ locale }}"></h3>
       <div id="place-summary" class="row-col place-summary-container">{{ place_summary }}</div>
       <div id="main-pane" class="row"></div>
       {# TRANSLATORS: A message shown on the page while the content is loading. #}
-      <div id="page-loading" className="mt-4">{% trans %}Loading...{% endtrans %}</div>
+      <div id="page-loading" className="mt-4">{% trans %}<svg xmlns="http://www.w3.org/2000/svg" height="24px" viewBox="0 -960 960 960" width="24px" fill="#5f6368"><path d="M480-80q-82 0-155-31.5t-127.5-86Q143-252 111.5-325T80-480q0-83 31.5-155.5t86-127Q252-817 325-848.5T480-880q17 0 28.5 11.5T520-840q0 17-11.5 28.5T480-800q-133 0-226.5 93.5T160-480q0 133 93.5 226.5T480-160q133 0 226.5-93.5T800-480q0-17 11.5-28.5T840-520q17 0 28.5 11.5T880-480q0 82-31.5 155t-86 127.5q-54.5 54.5-127 86T480-80Z"/></svg><p>Loading</p>{% endtrans %}</div>
```
dwnoble, line 72: Just wrap the {% trans %}  tag around the actual text, in this case the word "Loading"

dwnoble: Can you add comment describing the SVG and give a source reference where it came from?

dwnoble (approved):
Thank you!

gmechali (approved):
Nice! Thanks for getting this working!

PR #4735, opened 2024-11-13
Improve custom DC stable release PR creation script

hqpho:
- Fix head branch ref
- Add options to make rerunning quicker
- Don't directly create the PR, open web browser instead
- Seed the PR description with instructions and reference links


n-h-diaz (approved):
Thanks for making the updates! (And for helping me debug earlier 😃 )

PR #4740, opened 2024-11-18
Global header search

nick-nlb:
### Description

The purpose of this PR is to have the search bar in the header on all pages. 

Previously, pages by default did not have the search in the header, and had to be flagged in order to provide the search bar. Now, this flag is reversed and pages will contain the header search by default.

The search bar has been removed from the explore page, and the more complicated search functionality provided in that page (such as the ability to programmatically change the current query and provide a debug modal) is now available to the universal search bar in the header.

Related layout updates have been made to ensure that the explore page displays correctly with the internal search now gone, and the debug "bug" button is now aligned nicely (with the global search at the top). The debug modal has also been revamped.

### Notes

When the search bar was inline in the explore page, that page could interact and share state with its inline search. This allowed the explore page to provide and programmatically update the query string as well as to provide debugging information, among other things.

We now have to communicate that information from the explore page all the way to the header. Normally, we might consider options like React context, or lifting of state. 

However, in this case, the header and the explore page are in totally separate React apps with separate mount points, and that limits our ability to use more traditional methods to communicate this data.

The solution implemented was to create a global "Query Store", that can be updated by a component in one app, and then have those changes provided to a component in another app. The query store provides a hook to facilitate this communication. This allows these otherwise completely independent applications to communicate with each other in a reactive way.

### DIFF
--- a/static/css/core.scss
+++ b/static/css/core.scss
@@ -626,7 +626,7 @@ section {
     border: none;
     border-color: var(--gm-3-ref-neutral-neutral-40);
 
-    .search-icon {
+    .search-bar-icon {
       font-size: 24px;
       line-height: 1;
     }
--- a/static/css/explore.scss
+++ b/static/css/explore.scss
@@ -275,10 +275,9 @@
     position: relative;
   }
 
-  .debug-info-toggle {
-    top: -0.5rem;
-    right: 1.5rem;
-    padding: 0;
+  .debug-info-open {
+    top: 20px!important;
+    right: -10px!important;
   }
 }
 
--- a/static/js/components/nl_search_bar/auto_complete_input.tsx
+++ b/static/js/components/nl_search_bar/auto_complete_input.tsx
@@ -314,7 +314,7 @@ export function AutoCompleteInput(
         >
           <InputGroup className="search-bar-content">
             {isHeaderBar && (
-              <span className="search-icon">
+              <span className="search-bar-icon">
                 <Search />
               </span>
             )}


gmechali (approved):
Thanks Nick! Looks great, just curious what kind of testing you've done on this?
Also please wait for Dan's review still as I'm still newish to the DC codebase, so a secondary reviewer is more prudent.

static/js/shared/stores/query_store_hook.tsx
```diff
@@ -0,0 +1,107 @@
...
+
+export const useQueryStore = (): QueryStoreData => {
+  const [queryString, setQueryStringState] = useState<string>("");
+  const [placeholder, setPlaceholderState] = useState<string>(
+    "Enter a question to explore"
+  );
+  const [queryResult, setQueryResultState] = useState<QueryResult>(null);
+  const [debugData, setDebugDataState] = useState<any>({});
+
+  useEffect(() => {
```
dwnoble, line 45: Can you add a comment describing the useEffect?

static/js/shared/stores/query_store_hook.tsx
```diff
@@ -0,0 +1,107 @@
...
+    const queryStore = (globalThis.queryStore as QueryStore) || null;
+
+    const handleStoreUpdate = (
+      store: typeof queryStore,
+      changeType:
```
dwnoble, line 50: Can you reuse the ChangeType type here?

static/js/shared/stores/query_store_hook.tsx
```diff
@@ -0,0 +1,107 @@
...
+  queryString: string;
+  placeholder: string;
+  queryResult: QueryResult | null;
+  debugData: any;
+  setQueryString: (query: string) => void;
+  setQueryResult: (result: QueryResult) => void;
+  setDebugData: (data: any) => void;
+}
+
+export const useQueryStore = (): QueryStoreData => {
```
dwnoble, line 37: Nice job building out this hook!

static/js/types/app/explore_types.ts
```diff
@@ -62,7 +61,7 @@ export interface SVScores {
 export interface SentenceScore {
   sentence: string;
   score: number;
-  rerankScore: number;
+  rerank_score: number;
```
dwnoble, line 64: Good catch

server/templates/base.html
```diff
@@ -129,6 +131,7 @@
 {% include 'metadata/labels.html' %}
 {% include 'metadata/routes.html' %}
 
+<script src={{url_for('static', filename='queryStore.js', t=config['VERSION'])}} async></script>
```
dwnoble, line 134: Can you add this to all of the `server/templates/custom_dc/**/base.html` files?

dwnoble (changes requested):
Thanks Nick & Pablo- i like this approach.

I'm seeing an error when i run in "custom DC" mode (`./run_server.sh -m -e custom`):

Screenshot 2024-11-26 at 3 47 18 PM

This might be a little more involved to fix. some potential options:

(1) in custom dc mode, keep the search bar inline on the explore page
(2) update custom dc headers (in `server/templates/custom_dc/**/base.html`) to include a search bar. Maybe this means attaching the `HeaderBarSearch` to some new ID in the custom DC navbar, or it could be a pure-flask inlined search bar

What do you think?

dwnoble:
Also i saw a small difference in padding between the search bar in the visualization tools and the knowledge graph:

header-visualization-tools
header-knowledge-graph

dwnoble:
> > Thanks Nick & Pablo- i like this approach.
> > I'm seeing an error when i run in "custom DC" mode (`./run_server.sh -m -e custom`):
> > This might be a little more involved to fix. some potential options:
> > (1) in custom dc mode, keep the search bar inline on the explore page (2) update custom dc headers (in `server/templates/custom_dc/**/base.html`) to include a search bar. Maybe this means attaching the `HeaderBarSearch` to some new ID in the custom DC navbar, or it could be a pure-flask inlined search bar
> > What do you think?
> 
> Thank you for catching that! The error itself is fixed by including the store in the various base templates, but that leaves the question of how to implement the search bar in explore pages that don't use the standard header.
> 
> I would go with option 1 for this PR, because having the inline explorer search as an option (tied to the flag) will allow for all cases, including ones where a custom header implementation doesn't have it (and will give us the flexibility to use it if we do want to later).
> 
> It actually won't be a big task to restore the inline explorer search. It already does work with the header store functionality and so we really would just need to restore it and then tie its display to the flag that determines if the header search displays (if the header search flag is false, we display the inline search).
> 
> However, it could be nice to also do option 2 as a separate PR, to move towards having the header bar search as the default standard!
> 
> Let me know if you are happy with option 1 and I will implement that. All the other comments addressed!

Option 1 sounds good to me! And agreed that option 2 can come later, but i dont think it's urgent

dwnoble (approved):
Looks great! Thanks Nick.

One final request: can you remove the navbar search from the /place/* page? The place page has a separate search when you click "Change Place", and we don't want to show users two search bars.

(This will be moot in a few weeks, as we're planning on rolling out a new, re-designed version of the place page that can have the top search bar added back in)

dwnoble:
> Looks great! Thanks Nick.
> 
> One final request: can you remove the navbar search from the /place/* page? The place page has a separate search when you click "Change Place", and we don't want to show users two search bars.
> 
> (This will be moot in a few weeks, as we're planning on rolling out a new, re-designed version of the place page that can have the top search bar added back in)

Sorry for being wishy washy on this- after a little more internal discussion, can you **leave the navbar search box on the place page in place**.  We'll separately remove the "change place" search box altogether.

PR #4745, opened 2024-11-19
Remove "In Arcs" header when there are no arcs to display.

clincoln8:
One side effect of this change is that the footer does not stick to the bottom of the page for short pages. This is non blocking for now might may be addressed in follow up changes.

Before:
image

After:
image


beets (approved):
thanks for the fix!

PR #4749, opened 2024-11-20
Add alternate health check route for website server

hqpho:
Routes ending with "z" are reserved by Cloud Run: https://cloud.google.com/run/docs/known-issues#ah


keyurva (approved):
"super" :)

PR #4751, opened 2024-11-21
Adds a list of words to skip Autocomplete trigger and abort old requests

gmechali:
The list of words can grow later, for now we're just taking the keywords from the queries in the homepage.
Skipping the old requests became a requirement because requests for TELL would return faster than the previous one for TEL, so we still saw results from "TEL".

This PR fixes both of these things.


beets (approved):
thanks for these fixes!

dwnoble (approved):

PR #4756, opened 2024-11-25
Emotion and homepage updates

nick-nlb:
## Description

This PR contains three related updates:

### Homepage redesign

The homepage redesign is implemented based on the Homepage Improvements Figma ([link to figma](https://www.figma.com/design/G3D7geGkxHC23lom91oFkE/DC-Homescreen?node-id=701-2802&node-type=frame&t=8C10qesTLIeOxDYI-0)).

### Emotion

The core homepage, about page and build page components have been reworked into emotion, with localized styles defined within the component itself. A corresponding emotion theme (which will continue to be expanded on) is provided.

### Component rework

In doing the emotion conversion, the components themselves have been significantly refactored and reworked into:
* elements: these are small components such as chips, boxes, that can be used either independently or as part of a set within a larger component
* content: these are content components that typically represent a block on the page. These include things like hero components, chip sets, brick walls, and other re-usable pieces of content. Note that these components are generic and do not contain page specific content. Page specific content is provided to the components when they are used.
* layout: there are two layout components: Section and Separator. Section will wrap a page block and provide the padding and background color of that section (these are adjustable within the block). Separator allows extra padding to be added to the page in fixed amounts, with an optional border.

The page content itself will always be located within that page’s `app` directory. So all about page content is located in `static/js/apps/about/components`. These app-specific components make use of the generic content components and providing them with the actual content.

### Merging Note

Once the changes are approved we will apply updates to the relevant webdriver tests.

### DIFF
--- a/static/js/apps/homepage/app.tsx
+++ b/static/js/apps/homepage/app.tsx
@@ -36,7 +36,7 @@ import theme from "../../theme/theme";
 import { BuildYourOwn } from "./components/build_your_own";
 import { HomeHero } from "./components/home_hero";
 import { SampleQuestions } from "./components/sample_questions";
-import Tools from "./components/tools";
+import { Tools } from "./components/tools";
 import { UnitedNations } from "./components/united_nations";
 
 interface AppProps {
--- a/static/js/apps/homepage/components/tools.tsx
+++ b/static/js/apps/homepage/components/tools.tsx
@@ -36,7 +36,7 @@ interface ToolsProps {
   routes: Record<string, string>;
 }
 
-const Tools = ({ routes }: ToolsProps): ReactElement => {
+export const Tools = ({ routes }: ToolsProps): ReactElement => {
   const theme = useTheme();
   return (
     <>
@@ -123,5 +123,3 @@ const Tools = ({ routes }: ToolsProps): ReactElement => {
     </>
   );
 };
-
-export default Tools;
--- a/static/js/components/content/brick_wall.tsx
+++ b/static/js/components/content/brick_wall.tsx
@@ -38,10 +38,10 @@ const patterns = [
 export const BrickWall = ({ title, bricks }: BrickWallProps): ReactElement => {
   const theme = useTheme();
 
-  // This function spreads the bricks among the columns, effectively generating the mortar pattern based on the pattern set
-  // given above. Each column is represented by an object containing the bricks, and a patternEntry array that gives the corresponding
-  // pattern type for the brick at that index (i.e., does the brick take up two spaces or one). This is used in the component to
-  // correctly style the bricks.
+  // This function spreads the bricks among the columns, effectively generating the mortar pattern based on the
+  // pattern set given above. Each column is represented by an object containing the bricks, and a patternEntry
+  // array that gives the corresponding pattern type for the brick at that index (i.e., does the brick take up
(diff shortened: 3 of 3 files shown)


static/js/apps/about/components/who_can_use.tsx
```diff
@@ -0,0 +1,46 @@
...
+export const WhoCanUse = (): ReactElement => {
+  return (
+    <MediaText
+      mediaType="video"
+      mediaSource="O6iVsS-RDYI"
+      header="Who can use it"
+      headerComponent="h3"
+    >
+      <p>
+        Data Commons can be accessed by anyone here at Datacommons.org.
```
dwnoble, line 36: Looks like the figma has an href around Datacommons.org here, though it's not in the current live version. Could you add the link?

static/js/apps/about/components/splash_quote.tsx
```diff
@@ -0,0 +1,33 @@
...
+import Quote from "../../../components/content/quote";
+
+/**
+ * A component to display the splash quote section at the top of about page
+ */
+
+export const SplashQuote = (): ReactElement => {
+  return (
+    <Quote
+      quote="Every moment around the world people and organizations are generating data that can be extraordinarily useful and I think we have to find the way to harness that to solve problems.
```
dwnoble, line 28: I noticed some slight differences in padding & responsive behavior between the new & old quote:

static/js/apps/build/components/build_hero.tsx
```diff
@@ -0,0 +1,65 @@
...
+          >
+            Learn how
+          </a>
+        </h2>
+      </HeroColumns.Left>
+      <HeroColumns.Right>
+        <div>
+          <h4>Build and deploy your own</h4>
+          <p>
+            Launch your own Data Commons and customize it with your own data
```
dwnoble, line 45: should be "own data **to** better engage"

static/js/apps/build/components/build_hero.tsx
```diff
@@ -0,0 +1,65 @@
...
+            href="https://docs.datacommons.org/custom_dc?utm_source=buildpage_hero"
+            title="Build your own Data Commons"
+          >
+            Learn how
+          </a>
+        </h2>
+      </HeroColumns.Left>
+      <HeroColumns.Right>
+        <div>
+          <h4>Build and deploy your own</h4>
```
dwnoble, line 43: Looks like the font sizes are slightly different here compared to what's in production:

Screenshot 2024-12-06 at 10 28 57 PM

static/js/apps/homepage/app.tsx
```diff
@@ -56,24 +59,41 @@ export function App({
...
+      </Section>
+
+      <Section variant="small">
+        <Tools routes={routes} />
+      </Section>
+
+      <Separator border={false} variant="small" />
+
+      <Section variant="large" colorVariant="dark">
+        <BuildYourOwn routes={routes} />
```
dwnoble, line 85: Looks like the figma has some copy updates for this page:

Screenshot 2024-12-06 at 10 43 57 PM

dwnoble (commented):
Pablo and Nick,

Thanks for these changes-they look great. Left a few comments with some minor style & copy callouts, and one stylistic question about `styled` vs inline `css` in emotion

dwnoble (approved):
Thanks Nick & Pablo!

PR #4770, opened 2024-12-04
Reformat NodeJS failure email subject

hqpho:
Right now, failures in different environments are lumped into the same thread in the Google Groups UI. b/253630058 explains that this is because anything in square brackets at the start of the subject line is ignored.


chejennifer (approved):
good find, thanks for fixing!

PR #4771, opened 2024-12-05
Only show the Google Map widget on Place pages for places contained within the US

gmechali:
The Geometry coordinates for detailed Google maps boundaries is only available for places contained in the USA (not even the USA country), so we only want to show that component for the places contained within the USA.


dwnoble (approved):
Thanks Gabe!

PR #4774, opened 2024-12-06
Improve "place in" box on dev place pages

juliawu:
## 1) Fix long list of child places on continent place pages

On continent place pages, the list of child places will contain places that are more than one level down in containment (e.g, showing municipalities, not just countries). This PR adds a child place type entry for continents to the child place fetch logic.

### Before:

Screenshot 2024-12-05 at 6 18 26 PM

### After:

Screenshot 2024-12-05 at 6 18 33 PM

## 2) Add a "show more" toggle when the list of child places is long

Sometimes the list of child places is very long, which overwhelms the page. This PR also adds a "show more" toggle when the number of child places is greater than 15.

### Long lists are truncated initially
Screenshot 2024-12-09 at 9 57 08 AM

### Show full list after clicking "show more"
Screenshot 2024-12-05 at 6 12 25 PM

### Short lists don't show a toggle
Screenshot 2024-12-05 at 6 11 53 PM

## 3) Replace place search with link to knowledge graph on current place pages

In preparation for the changes made in PR #4740, this PR also removes the place search bar to avoid showing users double search bars, and instead shows the DCID and a link to the knowledge graph entry, fixing a longstanding feature regression.

Before:
Screenshot 2024-12-09 at 11 53 21 AM

After:
Screenshot 2024-12-09 at 11 47 12 AM


static/js/place/dev_place_main.tsx
```diff
@@ -395,24 +395,46 @@ const RelatedPlaces = (props: {
   place: NamedTypedPlace;
   childPlaces: NamedTypedPlace[];
 }) => {
+  const [isCollapsed, setIsCollapsed] = useState(true);
   const { place, childPlaces } = props;
   if (!childPlaces || childPlaces.length === 0) {
     return null;
   }
+  const NUM_PLACES = 15;
```
gmechali, line 403: curious - how did we choose 15?

static/js/place/dev_place_main.tsx
```diff
@@ -395,24 +395,46 @@ const RelatedPlaces = (props: {
...
       <div className="item-list-container">
         <div className="item-list-inner">
-          {childPlaces.map((place) => (
-            <div key={place.dcid} className="item-list-item">
-              <a className="item-list-text" href={`/dev-place/${place.dcid}`}>
-                {place.name}
-              </a>
-            </div>
-          ))}
+          {childPlaces
```
gmechali, line 414: I might argue this is starting to add a lot of logic within the render block. 
could we maybe store the slice in a different variable called trimmedChildPlaces and then just toggle between trimmedChildPlaces and allChildPlaces?

gmechali (approved):
Looks very neat!! 
Did you figure out why we kept hovering between the long list and the short list of child places when looking at the europe page?

static/js/place/dev_place_main.tsx
juliawu: Good call out. Stored the slice in a different variable and now just toggle between the lists which looks much cleaner now.

beets:
Just from the screenshots -- looks like we are returning Yugoslavia. We should be filtering "deprecated" countries..

dwnoble:
> Just from the screenshots -- looks like we are returning Yugoslavia. We should be filtering "deprecated" countries..

Agreed-this is already on our task list

static/js/place/dev_place_main.tsx
```diff
@@ -413,6 +423,16 @@ const RelatedPlaces = (props: {
           ))}
         </div>
       </div>
+      {showToggle && (
+        <div className="show-more-toggle" onClick={toggleShowMore}>
+          <span className="show-more-toggle-text">
+            {isCollapsed ? "Show more" : "Show less"}
```
dwnoble, line 429: Can you add the count for show more? e.g., "Show 25 more"

dwnoble (approved):

static/js/place/dev_place_main.tsx
juliawu: Admittedly a bit arbitrarily -- this looks good to my eyes for now. My thought is to use `NUM_PLACES = 15` to unblock experimental release before the code freeze, and we can decide on a better solution after.

static/js/place/dev_place_main.tsx
juliawu: Great idea, done! Also updated screenshots in the PR description.

gmechali (approved):

static/js/place/dev_place_main.tsx
dwnoble: After thinking about this more, i think a better experience might be to make this responsive and to show more items when the page is larger, and fewer for smaller screen sizes. Wdyt @miss-o-soup ?

juliawu: Sounds good, I'll discuss with @miss-o-soup and add responsive behavior in a follow-up PR, I don't think this needs to block release.

PR #4775, opened 2024-12-06
Create script to generate topic cache from mcf.

keyurva:
* The MCF is the latest MCF provided by the UN.
* The topic cache json and sentence CSV are generated by the script.

### DIFF
--- a/tools/sdg/topics/README.md
+++ b/tools/sdg/topics/README.md
@@ -39,3 +39,16 @@ To run it:
 export DC_API_KEY=<AUTOPUSH KEY>
 python3 enum_topics.py
 ```
+
+
+## Generate Topic Cache
+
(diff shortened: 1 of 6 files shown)


dwnoble (commented):
Thanks Keyur!

Can you remove the sheets_svs.csv file? I think it does have some issues so i'd rather keep the oder version that we know produces a good topic cache output

keyurva:
The old one is the one that will continue to be used. The `sheets_svs.csv` that will be used is this one: https://github.com/datacommonsorg/website/blob/732f6da6d568e2efd8e7e210796b818b10980163/tools/nl/embeddings/input/sdg/sheets_svs.csv and not the one from this PR. It will need to be copied to the embeddings directory if it were to be used.

dwnoble (approved):
Got it-that makes sense.

Thanks keyur!

PR #4779, opened 2024-12-09
Filters out deprecated places from the fetch_places helper method.

gmechali:
Modifies the fetch_places helper to also get the DissolutionDate and filter out these dissolved places from the responses

No more Yugoslavia: https://screenshot.googleplex.com/B5axwNbChiP5Y3N


server/routes/dev_place/utils.py
```diff
@@ -245,6 +245,10 @@ def fetch_places(place_dcids: List[str], locale=DEFAULT_LOCALE) -> List[Place]:
     name_with_locale = select_string_with_locale(place_name_with_languages_strs,
                                                  locale=locale)
 
+    if place_props.get('dissolutionDate', []):
+      # skip places that have a dissolution date.
+      continue
```
gmechali, line 250: An alternative approach could have been to just add a "dissolved" boolean on the Place and do the filtering in related_places. I thought this was better but lmk what you think .

dwnoble: i like adding the dissolved attribute and filtering in related_places so this function doesn't have any side effects

dwnoble (approved):

PR #4780, opened 2024-12-09
Fixed padding on dev place page

dwnoble:
* Fixed dev place page to have `padding-top: 48px;`
* Removed extra tag with id="main" which was adding to the padding

## Before
Screenshot 2024-12-09 at 3 22 20 PM

## After
Screenshot 2024-12-09 at 3 20 07 PM


juliawu (approved):
Nice!

PR #4781, opened 2024-12-09
Added cities with population > 1M to dev place page experiment group

dwnoble:
Generated group using ~10% of all world cities with population > 1M at random using bigquery:

```
SELECT * FROM (
  SELECT 
    DISTINCT(svo.observation_about) as placeDcid, place.name, COUNT(*) as count
  FROM `datcom-store.dc_kg_latest.StatVarObservation`  svo
  JOIN `datcom-store.dc_kg_latest.Place` place
    ON svo.observation_about = place.id
  WHERE
    CAST(svo.value AS FLOAT64) > 1000000 AND 
    svo.variable_measured = "Count_Person" AND
    place.type = "City" AND
    place.name IS NOT null
  GROUP BY
    svo.observation_about, place.id, place.name
  ORDER BY RAND() DESC
  LIMIT 70
) ORDER BY placeDcid ASC
```


juliawu (approved):

dwnoble:
Re-opening this for the new place page rollout

gmechali (approved):

PR #4782, opened 2024-12-09
Add "embed this chart" modal to map tiles

juliawu:
This PR adds an "embed this chart" modal to the footer of the map tile that gives users HTML they can use to embed that chart as a web component.

Screenshots:

Screenshot 2024-12-09 at 3 42 21 PM

Screenshot 2024-12-09 at 3 42 30 PM

More concretely, this PR:
* Adds a new modal to show web component HTML
* Adds helper function and test to generate the web component source code for map tiles
* Adds the "material symbols outlined" fonts to web component shadow and light roots so that the `code` icon can be used, and DRYs the logic used
* Updates the styling of the icon buttons so we can have an emphasized (blue) and default (white) style
* Updates the styling of modal textareas to match figma mocks


miss-o-soup:
This is looking good, thank you!
Some small nit-picks are that the blue "Close" button should not have a hairline/border, or if keeping should be the same color as the blue button. Same for the modal dialog box, there shouldn't be a border/hairline, or if keeping it should be white.
Wondering if this button implementation and dialog can be generalized to be re-used as a component. Thank you!

juliawu:
> This is looking good, thank you! Some small nit-picks are that the blue "Close" button should not have a hairline/border, or if keeping should be the same color as the blue button. Same for the modal dialog box, there shouldn't be a border/hairline, or if keeping it should be white. Wondering if this button implementation and dialog can be generalized to be re-used as a component. Thank you!

Updated to remove the border on the blue button and modal dialog box. Screenshot also updated.

Button impelmentation is already a re-usable component! Styling for the modal is also shared (I inherited a lot from the work @beets did for the "see metadata" modal). Getting the download modal styling to match is a future fix-it friday item.

static/js/components/tiles/map_tile.tsx
```diff
@@ -749,3 +751,94 @@ function getExploreLink(props: MapTilePropType): {
...
+  // Generate title with replacement strings filled in
+  const replacementStrings =
+    mapChartData && getReplacementStrings(props, mapChartData);
+  const header = getChartTitle(props.title, replacementStrings);
+
+  let sourceCode = `<script src="https://datacommons.org/datacommons.js"></script>
+<datacommons-map
+\theader="${header}"
+\tchildPlaceTypes="${enclosedPlaceTypes}"
+\tparentPlaces="${places}"
```
dwnoble, line 813: Can you use childPlaceType & parentPlace (singular) if there's only a single parent place? Right now i dont think we actually use multiple parent places anywhere, and using multiple places isn't documented in the web component map api

static/js/tools/shared/tile_code_modal.tsx
```diff
@@ -0,0 +1,96 @@
...
+            </a>
+            .
+          </div>
+          <ModalBody>
+            <textarea
+              className="modal-textarea"
+              readOnly
+              value={props.sourceCode}
+            >
+              {props.sourceCode}
```
dwnoble, line 78: Does it work to just set value and not the innerHTML here?

static/js/tools/shared/tile_code_modal.tsx
```diff
@@ -0,0 +1,96 @@
...
+          container={props.containerRef?.current}
+          toggle={toggleModal}
+          className="metadata-modal modal-dialog-centered modal-lg"
+        >
+          <ModalHeader toggle={toggleModal} close={<></>}>
+            Embed this chart
+          </ModalHeader>
+          <div className="modal-subtitle">
+            Use the following code to embed this chart on your website. For{" "}
+            documentation, see our{" "}
```
dwnoble, line 62: how about this alternate wording?

"Embed this chart on your website using the code below. For instructions and customization options, see our [Web Component Documentation](https://docs.datacommons.org/api/web_components/)."

static/library/index.ts
```diff
@@ -56,27 +57,8 @@ globalThis.datacommons = {
...
  * directly to the parent page
  */
-function loadStyles() {
-  const googleSansEl = document.createElement("link");
-  googleSansEl.href =
-    "https://fonts.googleapis.com/css2?family=Google+Sans:wght@300;400;500;700&family=Google+Sans+Text:wght@300;400;500;700&display=swap";
-  googleSansEl.rel = "stylesheet";
-  const materialIconsEl = document.createElement("link");
-  materialIconsEl.href =
-    "https://fonts.googleapis.com/icon?family=Material+Icons&display=block";
```
dwnoble, line 70: Does dropping `&display=block` here & in the sheet below have any side effects?

static/js/tools/shared/tile_code_modal.tsx
```diff
@@ -0,0 +1,96 @@
...
+              href="https://docs.datacommons.org/api/web_components/"
+              rel="noreferrer"
+              target="_blank"
+            >
+              web component documentation
+            </a>
+            .
+          </div>
+          <ModalBody>
+            <textarea
```
dwnoble, line 73: Can you combine this with the textarea in chart_embed.tsx to a reusable component?

static/css/shared/modal.scss
```diff
@@ -20,16 +20,20 @@
...
 
 .modal textarea {
-  background: #efefef;
-  border-radius: 3px;
+  background: #fff;
+  border-radius: 8px;
   border: $border;
   display: flex;
-  font-family: monospace;
+  font-family: "Google Sans Text", sans-serif;
```
dwnoble, line 30: Use css variable here - might need to verify which one, but maybe `font-family: var(--dc-font-family, base.$font-family-sans-serif);`

PR #4786, opened 2024-12-11
Reorganize custom DC Terraform script for readability

kmoscoe:
- Groups together variables for each service, and adds a heading for each group
- Reorders services to follow workflow sequence (e.g. Redis comes at the end, if needed)
- Moves API hostname and protocol to locals.tf so users can't change it


dwnoble (commented):
I get this error when running terraform plan:

```
$ terraform plan
╷
│ Error: Reference to undeclared input variable
│ 
│   on locals.tf line 30, in locals:
│   30:   dc_api_root = "${var.dc_api_protocol}://${var.dc_api_hostname}"
│ 
│ An input variable with the name "dc_api_protocol" has not been declared. This variable can be declared with a variable "dc_api_protocol" {} block.
╵
╷
│ Error: Reference to undeclared input variable
│ 
│   on locals.tf line 30, in locals:
│   30:   dc_api_root = "${var.dc_api_protocol}://${var.dc_api_hostname}"
│ 
│ An input variable with the name "dc_api_hostname" has not been declared. This variable can be declared with a variable "dc_api_hostname" {} block.
╵
```

Looks like this is because `dc_api_hostname` was removed from the variables.tf. I think the options are:
(1)  we keep the `dc_api_hostname` defined, but recommend users don't change it. Down the road, we may want to point the variable at "autopush.api.datacommons.org for our own internal testing instances.
(2) Add "api.datacommons.org" inline in locals.tf

I'm inclined to do (1) - wdyt @kmoscoe ?

kmoscoe:
> I get this error when running terraform plan:
> 
> ```
> $ terraform plan
> ╷
> │ Error: Reference to undeclared input variable
> │ 
> │   on locals.tf line 30, in locals:
> │   30:   dc_api_root = "${var.dc_api_protocol}://${var.dc_api_hostname}"
> │ 
> │ An input variable with the name "dc_api_protocol" has not been declared. This variable can be declared with a variable "dc_api_protocol" {} block.
> ╵
> ╷
> │ Error: Reference to undeclared input variable
> │ 
> │   on locals.tf line 30, in locals:
> │   30:   dc_api_root = "${var.dc_api_protocol}://${var.dc_api_hostname}"
> │ 
> │ An input variable with the name "dc_api_hostname" has not been declared. This variable can be declared with a variable "dc_api_hostname" {} block.
> ╵
> ```
> 
> Looks like this is because `dc_api_hostname` was removed from the variables.tf. I think the options are: (1) we keep the `dc_api_hostname` defined, but recommend users don't change it. Down the road, we may want to point the variable at "autopush.api.datacommons.org for our own internal testing instances. (2) Add "api.datacommons.org" inline in locals.tf
> 
> I'm inclined to do (1) - wdyt @kmoscoe ?

Can we not just remove it altogether from locals.tf (I should have done that; sorry, my bad) or is that locals.tf file also used by the base Data Commons?

dwnoble:
The`dc_api_root` variable from locals.tf is used in the rest of the terraform deployment to point at the main data commons instance.

kmoscoe:
> The`dc_api_root` variable from locals.tf is used in the rest of the terraform deployment to point at the main data commons instance.

OK, in that case, I'd prefer to do option #1.

dwnoble:
Sounds good to me

kmoscoe:
Sorry, I meant option 2!

kmoscoe:
> Sorry, I meant option 2!

OK done!

dwnoble (approved):

PR #4791, opened 2024-12-12
Topics to questions

nick-nlb:
## Description

This PR appends question marks to the end of topics that are worded as questions. These appear on the topics pages, such as  [http://localhost:8080/explore/economics](http://localhost:8080/explore/economics) . 

There are also minor copy fixes to some of the questions.

### Screenshot
questions

### DIFF
--- a/static/js/apps/explore_landing/topics.json
+++ b/static/js/apps/explore_landing/topics.json
@@ -38,29 +38,29 @@
             "url": "q=Tell+me+about+economic+equity+in+California"
           },
           {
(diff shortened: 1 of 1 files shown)


miss-o-soup:
This change looks great to me, thank you!
We are looking out for the "?" in all the question-tiles like the one here in all the topic-pages and also home page, right?
Thank you!

miss-o-soup (approved):

dwnoble (approved):

PR #4792, opened 2024-12-12
Place revamp - minor bug fixes

gmechali:
Sorts the parent places according to the parent place type in both old and new place page.
Before: https://screenshot.googleplex.com/73zetb4iETK6D7p
After: https://screenshot.googleplex.com/4KQ4baSMN64ZNS6

Adds source tile to the bottom of the key demographics in the overview:
https://screenshot.googleplex.com/57gfbASDD62Jx5y

Replaces the source tile from "Source" to "Sources" when there are multiple.
Before: https://screenshot.googleplex.com/8i94e76fpnWRStE
After: https://screenshot.googleplex.com/HfKwv3zrbDsHxK9

### DIFF
--- a/server/routes/dev_place/utils.py
+++ b/server/routes/dev_place/utils.py
@@ -27,7 +27,7 @@ import server.routes.shared_api.place as place_api
 from server.services import datacommons as dc
 
 # Parent place types to include in listing of containing places at top of page
-PARENT_PLACE_TYPES_TO_HIGHLIGHT = {
+PARENT_PLACE_TYPES_TO_HIGHLIGHT = [
     'County',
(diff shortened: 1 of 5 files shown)


juliawu (approved):
Paging @miss-o-soup for styling. I think the font-size of the sources on the overview tile in the mocks is a little smaller than what you have in the screenshot. Should we downsize?

Otherwise, LGTM! Thanks for this, very clean.

gmechali:
Ah good point! Resolved!
https://screenshot.googleplex.com/9CKXGryzqMUjSGy

dwnoble (approved):

PR #4794, opened 2024-12-12
Fixed sanity test issue to account for homepage redesign

dwnoble:
The sanity tests started failing after [migrating the homepage to use emotion](https://github.com/datacommonsorg/website/pull/4756)

* Added validation to run_website_sanity.sh to error gracefully if no $domain arg is passed in
* Allow setting the protocol along with $domain arg to allow testing from localhost
* Added `data-testid` attribute to LinkChip, LinkBox, and QueryLink element to allow test cases to select dom elements without a fixed class name
* Updated sanity.py test to use `data-testid` attribute to select elements

### DIFF
--- a/run_website_sanity.sh
+++ b/run_website_sanity.sh
@@ -17,8 +17,22 @@ set -e
 
 # DC Instance domain like: "dev.datacommons.org", "datacommons.org"
 domain=$1
+# Check if the domain is set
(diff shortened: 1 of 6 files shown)


gmechali (approved):
Thanks for taking this on!

keyurva (approved):
Thanks for the fix, Dan!

PR #4796, opened 2024-12-13
Keep the user's query in the NL search bar when landing on the place page

dwnoble:
Demo

[Keep query in the search bar when landing on the place page.webm](https://github.com/user-attachments/assets/12da3e1f-f01a-44f0-b87e-de68f3adbb3d)

### DIFF
--- a/server/webdriver/shared_tests/place_explorer_test.py
+++ b/server/webdriver/shared_tests/place_explorer_test.py
@@ -180,10 +180,14 @@ class PlaceExplorerTestMixin():
     # Assert 200 HTTP code: successful page load.
     self.assertEqual(shared.safe_url_open(self.driver.current_url), 200)
 
-    # Assert page title is correct.
+    # Assert page title is correct, and that the query string is set in the url.
     WebDriverWait(self.driver, self.TIMEOUT_SEC).until(
         EC.title_contains('United States of America'))
-    self.assertTrue("place/country/USA" in self.driver.current_url)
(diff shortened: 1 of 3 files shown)


gmechali:
We also do redirection directly from the autocomplete result. Should we also add that query there? It might just be a partial query like typing "Unite", select "United States", gets redirected to USA place page.

I might argue at that point it's better not to add the query? Just calling it out in case we do want to add it.

https://github.com/datacommonsorg/website/blob/6bfa441d3c482b3f143f9765273e892c516d5053/static/js/components/nl_search_bar/auto_complete_input.tsx#L291

gmechali (approved):

dwnoble:
> We also do redirection directly from the autocomplete result. Should we also add that query there? It might just be a partial query like typing "Unite", select "United States", gets redirected to USA place page.
> 
> I might argue at that point it's better not to add the query? Just calling it out in case we do want to add it.
> 
> https://github.com/datacommonsorg/website/blob/6bfa441d3c482b3f143f9765273e892c516d5053/static/js/components/nl_search_bar/auto_complete_input.tsx#L291

Good call- ill address this in a follow up since autocomplete isn't turned on yet. From the discussion with @miss-o-soup : 
(1) if the user arrives at the page from a search, we'll show the search that they typed in (which is what this PR adds)
(2) if the user arrives at the place page any other way, we'll show the place name in the search bar instead of the default placeholder text. This should also cover when the user clicks on a result in the autocomplete list

PR #4797, opened 2024-12-13
Moving place revamp test to Base DC only.

gmechali:
The revampled place page experiment should not be tested against CDC autopush. Extracting that test back out to the Based DC tests only.
Once the revampled place page is fully enabled, we should update all the place page tests to be for the new page.

### DIFF
--- a/server/webdriver/base_utils.py
+++ b/server/webdriver/base_utils.py
@@ -25,7 +25,7 @@ DEFAULT_WIDTH = 1200
 def create_driver(preferences=None):
   # These options are needed to run ChromeDriver inside a Docker without a UI.
   chrome_options = Options()
-  chrome_options.add_argument('--headless=new')
+  # chrome_options.add_argument('--headless=new')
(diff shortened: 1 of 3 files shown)


server/webdriver/base_utils.py
```diff
@@ -25,7 +25,7 @@
 def create_driver(preferences=None):
   # These options are needed to run ChromeDriver inside a Docker without a UI.
   chrome_options = Options()
-  chrome_options.add_argument('--headless=new')
+  # chrome_options.add_argument('--headless=new')
```
dwnoble: Was this intentional? Should we just delete the line entirely?

dwnoble (approved):

server/webdriver/base_utils.py
gmechali: Reverted! 
Definitely should keep it because without it, it ends up showing you all the tests that run which is great for debugging but insanely annoying when it keeps taking over your cursor :)

PR #4805, opened 2024-12-18
Updated cdc data Dockerfile to install dependencies before copying over source files

dwnoble:
Previously during local development, code changes would for pip dependencies to be reinstalled. With this change, pip dependencies are only reinstalled if the actual requirements.txt files change.

### DIFF
--- a/build/cdc_data/Dockerfile
+++ b/build/cdc_data/Dockerfile
@@ -14,51 +14,25 @@
 
 
 # #### Stage 1: Download base dc model from GCS. ####
-FROM google/cloud-sdk:slim as model-downloader
+FROM google/cloud-sdk:slim AS model-downloader
 
 # Copy model.
 RUN mkdir -p /tmp/datcom-nl-models \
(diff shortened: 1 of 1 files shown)


build/cdc_data/Dockerfile
```diff
@@ -14,69 +14,71 @@
...
 
 # Copy model.
 RUN mkdir -p /tmp/datcom-nl-models \
     && gsutil -m cp -R gs://datcom-nl-models/ft_final_v20230717230459.all-MiniLM-L6-v2/ /tmp/datcom-nl-models/
 
 
-# #### Stage 2: Copy required files. ####
-FROM python:3.11.4-slim as file-copier
+# #### Stage 2: Install python dependencies. ####
+FROM python:3.11.4-slim AS file-copier
```
keyurva, line 25: Consider renaming the stage to `dependencies-installer` or something similar.

keyurva (approved):
Very cool - this makes the docker approach usable in the development stage!

Can you also run the container in the cloud as a verification step before submitting this PR?

build/cdc_data/Dockerfile
dwnoble: Done!

PR #4811, opened 2024-12-20
Fixed webdriver tests that stopped working due to data changes

dwnoble:
* Updated tests to account for data removed from `GrowthRate_Count_Person` and `Amount_Consumption_Energy_PerCapita` stat vars
* Updated `/api/ranking` endpoint to return a 404 when no data is available for a place rather than a 500

### DIFF
--- a/server/webdriver/shared_tests/ranking_test.py
+++ b/server/webdriver/shared_tests/ranking_test.py
@@ -57,7 +57,7 @@ class RankingTestMixin():
     """Test translations are displayed correctly in hindi, as well as bottom rankings rendered correctly."""
     self.driver.get(
         self.url_ +
-        '/ranking/Count_Person/Country/?h=country%2FIND&unit=kg&hl=hi&bottom=')
+        '/ranking/Count_Person/Country/?h=country%2FIND&hl=hi&bottom=')
 
     title_present = EC.text_to_be_present_in_element(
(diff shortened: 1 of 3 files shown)


server/webdriver/shared_tests/ranking_test.py
```diff
@@ -53,12 +53,11 @@ def test_northamerica_population(self):
...
     self.assertEqual(x_text[-1].text, 'Montserrat')
 
-  def test_energy_consumption_bottom_ranking_hi(self):
+  def test_population_bottom_ranking_hi(self):
     """Test translations are displayed correctly in hindi, as well as bottom rankings rendered correctly."""
     self.driver.get(
         self.url_ +
-        '/ranking/Amount_Consumption_Energy_PerCapita/Country/?h=country%2FIND&unit=kg&hl=hi&bottom='
-    )
+        '/ranking/Count_Person/Country/?h=country%2FIND&unit=kg&hl=hi&bottom=')
```
chejennifer, line 60: nit: maybe remove the unit=kg because that doesn't super make sense for count_person?

chejennifer (approved):
thanks for fixing and simplifying the tests!

server/webdriver/shared_tests/ranking_test.py
dwnoble: done!

PR #4814, opened 2024-12-23
Resolves more lint warnings. Down to 221.

gmechali:
More and more return types on functions that did not have it. Down to 281 remaining warnings.


chejennifer (approved):
thank you!!

PR #4817, opened 2024-12-26
Fixed bug where export csv wasn't working on place page 'snapshot' group bar charts

dwnoble:
Context: b/384872751


chejennifer (approved):
thanks for fixing!

PR #4818, opened 2024-12-26
Removed stage from CDC data dockerfile to reduce image size

dwnoble:
* Previously, pip dependencies were installed in Stage 2 and copied to Stage 3, causing them to be stored in two layers and doubling their size in the final image.
* Combined Stages 2 and 3 into a single layer to eliminate the duplication of pip dependencies and to reduce the final image size.

### DIFF
--- a/build/cdc_data/Dockerfile
+++ b/build/cdc_data/Dockerfile
@@ -46,7 +46,12 @@ ARG PIP_NO_CACHE_DIR=1
 # Create a virtual env, add it to path, and install all requirements.
 RUN python -m venv /workspace/venv
 ENV PATH="/workspace/venv/bin:$PATH"
+
+# TODO: Install requirements for embeddings importer and data importer in separate virtual envs.
+# Install embeddings importer requirements.
 RUN pip3 install -r ./import/simple/requirements.txt
+
+# Install data requirements.
 # Remove lancedb - it is not used by custom dc.
 RUN sed -i'' '/lancedb/d' /workspace/nl_requirements.txt \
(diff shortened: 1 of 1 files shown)


build/cdc_data/Dockerfile
```diff
@@ -21,11 +21,17 @@ RUN mkdir -p /tmp/datcom-nl-models \
     && gsutil -m cp -R gs://datcom-nl-models/ft_final_v20230717230459.all-MiniLM-L6-v2/ /tmp/datcom-nl-models/
 
 
-# #### Stage 2: Install python dependencies. ####
-FROM python:3.11.4-slim AS dependencies-installer
+# #### Stage 2: Python runtime. ####
+FROM python:3.11.4-slim AS runner
```
keyurva, line 25: One thing I'm not a big fan of is that there is no separation between the import and embeddings modules (the [original version](https://github.com/datacommonsorg/website/blob/b20ed2b7e57ebc7a9e6a2df7f0bf358c932b1812/build/cdc_data/Dockerfile) did).

Having said that, let's go with this to keep the image size in check. But it'd be good to figure out a way to keep the separation whilst managing conflicting dependencies in the future (perhaps add a TODO).

keyurva (approved):
Thanks Dan!

build/cdc_data/Dockerfile
dwnoble: Agreed we should separate the dependencies for these two.

Looking closer at the previous version, I don't believe that actually separated the dependencies either. It looks like the virtual environments were installed separately, but then combined on top of each other to one common virtual environment at the end.

A better solution may be to create two entirely separate virtual environments for each tool.

I added a todo noting this

PR #4820, opened 2024-12-27
Skip failing custom dc tests.

keyurva:
* Certain place explorer tests are failing on custom DC autopush. This PR skips those tests for now.
* The failures are likely due to recent place explorer page changes that have probably not been ported to custom DC.
* Filed b/386405593 to fix and unskip them subsequently.
* Also filed b/386405338 to run these tests on each PR so they are caught at the outset.

### DIFF
--- a/server/webdriver/cdc_tests/autopush/place_explorer_i18n_test.py
+++ b/server/webdriver/cdc_tests/autopush/place_explorer_i18n_test.py
@@ -11,11 +11,19 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+import unittest
+
 from server.webdriver.cdc_tests.autopush.cdc_base_webdriver import \
     CdcAutopushTestBase
+from server.webdriver.cdc_tests.autopush.cdc_base_webdriver import \
+    SKIPPED_TEST_REASON
 from server.webdriver.shared_tests.place_explorer_i18n_test import \
     PlaceI18nExplorerTestMixin
 
 
 class TestPlaceExplorerI18n(PlaceI18nExplorerTestMixin, CdcAutopushTestBase):
   """Class to test the i18n place explorer page for Custom DC. Tests come from PlaceI18nExplorerTestMixin."""
+
+  @unittest.skip(reason=SKIPPED_TEST_REASON)
+  def test_explorer_redirect_place_explorer_keeps_locale(self):
+    super().test_explorer_redirect_place_explorer_keeps_locale()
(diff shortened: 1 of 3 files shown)


server/webdriver/cdc_tests/autopush/place_explorer_i18n_test.py
```diff
@@ -11,11 +11,19 @@
...
+from server.webdriver.cdc_tests.autopush.cdc_base_webdriver import \
+    SKIPPED_TEST_REASON
 from server.webdriver.shared_tests.place_explorer_i18n_test import \
     PlaceI18nExplorerTestMixin
 
 
 class TestPlaceExplorerI18n(PlaceI18nExplorerTestMixin, CdcAutopushTestBase):
   """Class to test the i18n place explorer page for Custom DC. Tests come from PlaceI18nExplorerTestMixin."""
+
+  @unittest.skip(reason=SKIPPED_TEST_REASON)
```
dwnoble, line 27: Just a quick note: we already have an approach for skipping unit tests in custom DC. You can remove the test from the shared_tests mixin and add it directly to the main DC `/tests/` package instead. Here's an example [example](https://github.com/datacommonsorg/website/blob/62e394b9e4d1acf7e53c18a3946453a8a0a839a7/server/webdriver/tests/place_explorer_test.py#L27)

keyurva: Thanks for the note. In these 2 cases I don't know if we actually want to remove them or fix and keep them in custom DC. I'll work with Gabe in the new year to figure it out.

Skipping them explicitly for now to stop the alerts until then.

gmechali: I think what is done here is better because if this test passes on base dc we should keep it! We definitely should fix this test for custom dc. I like the way keyur did it here! 
Also I think we ve had enough instances of breaking auto push cdc with these webdriver tests. We should really try to have these run on a local cdc as well.

dwnoble: Sounds good—skipping for now makes sense

dwnoble: @gmechali agreed. is testing against autopush helpful, or should these tests only run against a local cdc?

dwnoble (approved):

server/webdriver/cdc_tests/autopush/place_explorer_i18n_test.py
gmechali: I'd argue it's still useful since it allows us to test what's deployed in autopush and catch bugs pre releases
But yeah local would probably be a bit more valuable since we d catch it even earlier

keyurva: Thanks both for chiming in!

PR #4823, opened 2025-01-02
Fix default place name and add continents to place experiment

gmechali:
The place page name is selected as the nameWithLanguage property for the current language and fallsback to the name property first, then the dcid. However a bug in the implementation made it only fallback to the name property if it didn't have ANY nameWithLanguage values.
See datacommons.org/browser/wikidataId/Q1133254 for a place that has nameWithLanguage but only for one language.
With this PR, the name property is correctly used as the default.

Additionally, Continent place pages work on the dev place page although not super well yet. This PR adds all continents to the place experiments since it represents an improvement from the status quo. More improvements to come.

Place name fix: https://screenshot.googleplex.com/93nHvgo4VHqTh8d

### DIFF
--- a/run_servers.sh
+++ b/run_servers.sh
@@ -30,6 +30,14 @@ if [[ "$1" == "--verbose" ]]; then
   VERBOSE=true
 fi
 
+export FLASK_ENV="${FLASK_ENV:-integration_test}"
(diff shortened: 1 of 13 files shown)


dwnoble (approved):
Nice find

PR #4824, opened 2025-01-02
Add child places for all allowed place types and sort them

gmechali:
Some places did not have a name which caused an error when we tried to get the name property. See https://autopush.datacommons.org/browser/nuts/LVZZ

Discussed whether to keep the restriction for related place charts on only one child_place_type and we decided to drop the requirement. The old place page didnt have it, and the new one would show significantly changed child places. Some countries use EurostatNuts1 while others use AA1. Now we show children of different type and dedupe them.

See Latvia:
Old Place Page: https://screenshot.googleplex.com/BjiMRDVEqQpXEZK
New Place Page before: https://screenshot.googleplex.com/3L6ot765jxM4AjT
New place page after: https://screenshot.googleplex.com/A4kBUEaGitVkum4

Sources tile no longer shown on load:
Before: https://screenshot.googleplex.com/8RGHgDKNTBQKoYD
After: https://screenshot.googleplex.com/9L6AA9XWVYgEgjW

### DIFF
--- a/server/routes/dev_place/api.py
+++ b/server/routes/dev_place/api.py
@@ -156,6 +156,7 @@ def related_places(place_dcid: str):
   seen_dcids = set(
   )  # Keep track of seen DCIDs to prevent dupes but keep ordering.
 
+  # TODO(gmechali): Refactor this into async calls.
   for child_place_type in ordered_child_place_types:
     for dcid in place_utils.fetch_child_place_dcids(place,
                                                     child_place_type,
--- a/server/routes/dev_place/api.py
+++ b/server/routes/dev_place/api.py
@@ -138,6 +138,7 @@ def related_places(place_dcid: str):
   - Place details (name, type, etc.)
   - Lists of nearby, similar, and child places
   """
+  print("We just a call in dev place api relalted places.")
   # Fetch place info
   place = place_utils.fetch_place(place_dcid, locale=g.locale)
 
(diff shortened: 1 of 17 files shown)


dwnoble (approved):

server/routes/dev_place/utils.py
```diff
@@ -370,7 +367,7 @@ def chart_config_to_overview_charts(chart_config, child_place_type: str):
 ])
 
 
-def get_child_place_type(place: Place) -> str | None:
+def get_child_place_types(place: Place) -> list[str]:
   """
   Determines the primary child place type for a given place.
```
dwnoble, line 372: Update comment

server/routes/dev_place/api.py
```diff
@@ -146,10 +148,22 @@ def related_places(place_dcid: str):
...
+  ordered_child_place_types = place_utils.get_child_place_types(place)
+  primary_child_place_type = ordered_child_place_types[
+      0] if ordered_child_place_types else None
+
+  child_place_dcids = []
+  seen_dcids = set(
+  )  # Keep track of seen DCIDs to prevent dupes but keep ordering.
+
+  for child_place_type in ordered_child_place_types:
+    for dcid in place_utils.fetch_child_place_dcids(place,
```
dwnoble, line 160: We should eventually refactor these calls to run asynchronously. maybe using https://flask.palletsprojects.com/en/stable/async-await/

server/routes/dev_place/utils.py
gmechali: Done.

server/routes/dev_place/api.py
gmechali: Added a todo.

PR #4825, opened 2025-01-02
Update NL Goldens

gmechali:
Tests started failing because our NL goldens hardcode exact year that rely on the real date. So tests like "Population over the last 5 years" now excludes the year 2020. 

A couple of the python unit tests also did the same thing, now they have the date mocked.


chejennifer (approved):
thanks for updating!

dwnoble (approved):

PR #4826, opened 2025-01-03
Topics tab refactoring

gmechali:
Refactor the Topics menu to use a shared TopicItem for each of the buttons.

Only show the Topic button for categories for which we know we have data. We add a second call for the OVERVIEW chart config even when we're on a different category because we need it to identify which topics we have.

France: https://screenshot.googleplex.com/X4McSQ5qax7f9w7
USA: https://screenshot.googleplex.com/BUJQV8rxhfvvT2e

### DIFF
--- a/server/routes/dev_place/utils.py
+++ b/server/routes/dev_place/utils.py
@@ -68,8 +68,9 @@ def get_parent_places(dcid: str) -> List[Place]:
       dcid, [])
   all_parents = []
   for parent in parents_resp:
-    all_parents.append(
-        Place(dcid=parent['dcid'], name=parent['name'], types=parent['type']))
+    if 'type' in parent and 'name' in parent and 'type' in parent:
+      all_parents.append(
+          Place(dcid=parent['dcid'], name=parent['name'], types=parent['type']))
 
   return all_parents
 
@@ -369,24 +370,24 @@ DEFAULT_CHILD_PLACE_TYPES = set([
 ])
 
 
-def get_child_place_type(place: Place) -> str | None:
+def get_child_place_types(place: Place) -> list[str]:
   """
-  Determines the primary child place type for a given place.
+  Determines the child place types for a given place.
 
   This function uses custom matching rules and fallback hierarchies to decide
-  the most relevant child place type for a given place.
+  the ordered child place types for a given place.
 
   Args:
       place (Place): The place object to analyze.
 
   Returns:
-      str | None: The primary child place type for the given place, or None if no match is found.
+      list[str]: The child place types for the given place, or empty list if no match is found.
   """
   # Attempt to directly match a child place type using the custom expressions.
   for f in PLACE_MATCH_EXPRESSIONS:
     matched_child_place_type = f(place)
     if matched_child_place_type:
-      return matched_child_place_type
+      return [matched_child_place_type]
 
   # Fetch all possible child places for the given place using its containment property.
   raw_property_values_response = fetch.raw_property_values(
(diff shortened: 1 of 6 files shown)


server/routes/dev_place/utils.py
```diff
@@ -128,6 +130,7 @@ def get_place_type_with_parent_places_links(dcid: str) -> str:
   return ''
 
 
+@cache.cached(timeout=TIMEOUT, query_string=True)
```
dwnoble, line 133: I think this needs to be @cache.memoize() to include the arguments in the cache key: https://flask-caching.readthedocs.io/en/latest/#memoization

Also-you can enable caching locally for testing by setting `USE_MEMCACHE = True` [here](https://github.com/datacommonsorg/website/blob/bb7b11fecd82d66f23ada004a11be067a029495a/server/app_env/local.py#L22)

static/js/place/dev_place_main.tsx
```diff
@@ -184,6 +184,61 @@ const PlaceHeader = (props: {
...
+  const { category, selectedCategory, forceDevPlaces, place } = props;
+
+  const createHref = (
+    category: string,
+    forceDevPlaces: boolean,
+    place: NamedTypedPlace
+  ): string => {
+    const isOverview = category === "Overview";
+
+    let href = `/place/${place.dcid}`;
```
dwnoble, line 212: Using URLSearchParams might make this more clear. Something like:

```
const href = `/place/${place.dcid}`;
const params = new URLSearchParams();

if (!isOverview) {
  params.set("category", category);
}
if (forceDevPlaces) {
  params.set("force_dev_places", "true");
}

const queryString = params.toString();
if (queryString) {
  return `${href}?${queryString}`;
} else {
  return href;
}

return `/place/${place.dcid}${queryString ? `?${queryString}` : ""}`;
```

static/js/place/dev_place_main.tsx
```diff
@@ -586,16 +573,30 @@ export const DevPlaceMain = (): React.JSX.Element => {
...
       setChildPlaces(relatedPlacesApiResponse.childPlaces);
       setParentPlaces(relatedPlacesApiResponse.parentPlaces);
-      setPageConfig(pageConfig);
+      setPageConfig(config);
     })();
   }, [place]);
 
+  useEffect(() => {
+    if (placeChartsApiResponse && placeChartsApiResponse.charts) {
+      const categories = new Set<string>();
```
dwnoble, line 588: Is this categories set used somewhere?

static/js/place/dev_place_main.tsx
```diff
@@ -586,16 +573,30 @@ export const DevPlaceMain = (): React.JSX.Element => {
...
+      placeChartsApiResponse.charts.forEach((chart) => {
+        categories.add(chart.category);
+      });
+      setCategories(
+        ["Overview"].concat(
+          Object.values(placeChartsApiResponse.translatedCategoryStrings)
```
dwnoble, line 594: I suspect this categories list should store both the english & translated values (one to use on the page text, and one to use in the URL param for the place page)

dwnoble: Also- we should add "Overview": "<translated overview value>" to the translatedCategoryStrings entry in the API response.

server/routes/dev_place/utils.py
gmechali: Done.

static/js/place/dev_place_main.tsx
gmechali: Done - definitely cleans it up a bit, thanks!

static/js/place/dev_place_main.tsx
gmechali: Oh i was using that before but then switched to using the translatedCategories directly, so I think we're good now. Removed.

static/js/place/dev_place_main.tsx
gmechali: ah I see. Are the translation implemented yet? I tried using it with different HL params and it didnt do anything. I assume we'll need a bit more refactoring here, for example the overview topic will need a translation too, so it should come from the backend probably. 
I added a todo to figure this part out.

dwnoble: The translations are implemented, but i think there's some extra wiring we need to do to get the frontend to pass the language to the API.

FR example in autopush: https://autopush.datacommons.org/api/dev-place/charts/country/LVA?category=Overview&hl=fr

dwnoble (approved):

static/js/place/dev_place_main.tsx
gmechali: ack will see what's up with those in follow up.

dwnoble: sounds good!

PR #4829, opened 2025-01-07
update webdriver test

chejennifer:
fix a webdriver test that broke with new population data. 

Update the check to just look for values along the y axis instead of checking the actual value which can change


hqpho (approved):

hqpho:
FYI @keyurva This should also fix the recent CDC autopush test failures

PR #4830, opened 2025-01-07
Migrate /v1/bulk/observation-dates/linked to /v2/observation

n-h-diaz:


server/lib/util.py
```diff
@@ -880,8 +918,8 @@ def fetch_highest_coverage(variables: List[str],
   if entities is not None:
     series_dates_response = get_series_dates_from_entities(entities, variables)
   else:
-    series_dates_response = dc.get_series_dates(parent_entity, child_type,
```
dwnoble, line 883: With this change i think it's safe to remove the `get_series_dates` method entirely

dwnoble (approved):

PR #4831, opened 2025-01-07
WIP - Getting the Place page to show bar charts and ranking charts

gmechali:
See example: https://screenshot.googleplex.com/3kJS7mS2rtkjLkY

This is still WIP - Translations not working, and some of the code remains hacky. Sharing now for early feedback.

### DIFF
--- a/server/config/chart_config/economics.json
+++ b/server/config/chart_config/economics.json
@@ -8,7 +8,7 @@
       "Count_Person_Employed"
     ],
     "isOverview": true,
-    "isRankingChart": true,
+    "isRankingChart": false,
     "relatedChart": {
       "titleId": "CHART_TITLE-Percentage_employed",
       "title": "Percentage of people employed",
@@ -27,7 +27,7 @@
     "statsVars": [
       "Count_Person_InLaborForce"
     ],
-    "isRankingChart": true,
+    "isRankingChart": false,
     "relatedChart": {
       "titleId": "CHART_TITLE-Labor_force_rate",
       "title": "Labor force participation rate",
@@ -47,7 +47,7 @@
       "Count_Person_15To64Years_InLaborForce_AsFractionOf_Count_Person_15To64Years"
     ],
     "isOverview": true,
-    "isRankingChart": true,
+    "isRankingChart": false,
     "unit": "%",
     "scaling": 100,
     "relatedChart": {
@@ -65,7 +65,7 @@
     "unit": "%",
     "isOverview": true,
     "isChoropleth": true,
-    "isRankingChart": true,
+    "isRankingChart": false,
     "relatedChart": {
       "scale": false
     }
@@ -131,7 +131,7 @@
       "Count_UnemploymentInsuranceClaim_StateUnemploymentInsurance"
     ],
     "isOverview": true,
-    "isRankingChart": true,
+    "isRankingChart": false,
     "relatedChart": {
       "titleId": "CHART_TITLE-State_unemployment_claims_rate",
       "title": "State unemployment insurance claims rate",
@@ -152,7 +152,7 @@
     ],
     "unit": "%",
     "isOverview": true,
-    "isRankingChart": true,
+    "isRankingChart": false,
     "relatedChart": {
       "scale": false
     }
@@ -195,9 +195,34 @@
     ],
     "unit": "$",
     "isOverview": true,
-    "isRankingChart": true,
+    "isRankingChart": false,
     "relatedChart": {
       "scale": false
+    }
+  },{
+    "category": "Economics",
+    "titleId": "CHART_TITLE-Gdp",
+    "title": "Gross domestic product",
+    "description": "Gross domestic product, see https://en.wikipedia.org/wiki/Gross_domestic_product",
+    "statsVars": [
+      "Amount_EconomicActivity_GrossDomesticProduction_Nominal"
+    ],
+    "unit": "$",
+    "isOverview": true,
+    "isRankingChart": false
+  },{
+    "category": "Economics",
+    "titleId": "CHART_TITLE-Gdp-Other-Countries",
+    "title": "Gross domestic product Other Countries",
+    "description": "Gross domestic product other countries, see https://en.wikipedia.org/wiki/Gross_domestic_product",
+    "statsVars": [
+      "Amount_EconomicActivity_GrossDomesticProduction_Nominal"
+    ],
+    "unit": "$",
+    "isOverview": false,
+    "isRankingChart": false,
+    "relatedChart": {
+      "scale": true
     }
   },
   {
@@ -209,7 +234,7 @@
       "GrowthRate_Amount_EconomicActivity_GrossDomesticProduction"
     ],
     "unit": "%",
-    "isRankingChart": true,
+    "isRankingChart": false,
     "relatedChart": {
       "scale": false
     }
@@ -223,7 +248,7 @@
       "Amount_EconomicActivity_GrossNationalIncome_PurchasingPowerParity_PerCapita"
     ],
     "unit": "$",
-    "isRankingChart": true,
+    "isRankingChart": false,
     "relatedChart": {
       "scale": false
     }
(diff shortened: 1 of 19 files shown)


static/js/types/subject_page_proto_types.ts
```diff
@@ -66,6 +66,8 @@ export interface RankingTileSpec {
 
   showMultiColumn: boolean;
   showHighestLowest?: boolean;
+
+  showContainedInParent?: boolean;
```
dwnoble, line 70: I think these types correspond to `server/config/subject_page_proto.proto` - it might make sense to attach this metadata somewhere else on the place page config so the chart protos don't have to change

gmechali: In that case, how about I use the "showComparisonPlaces": "SIMILAR" | "CHILD"
That I already added for the bar chart configs? 

But I can add a "SIMILAR_IN_PARENT" value?
The bar charts, would choose countries like Germany/China/India when on the US place page.
For these rankings we want the ones that are in the same continent - so I think it just translates to using the SIMILAR countries but only the ones in the parent. 

Lmk if you prefer a different approach!  i ll implement it right now!

packages/client/src/data_commons_web_client_types.ts
```diff
@@ -155,6 +182,9 @@ export interface Chart {
   topicDcids: string[];
   denominator?: string; // Optional
   unit?: string; // Optional
+  rankingTileSpec?: RankingTileSpec; // Optional
+  barTileSpec?: BarTileSpec; // Optional
+  comparisonPlaceType?: string; // Optional
```
dwnoble, line 187: Nit: make this an enum instead of a string. `comparisonPlaceType?: "SIMILAR" | "SIMILAR_IN_PARENT"`

dwnoble: Also- when i read "comparisonPlaceType" it makes me think of a placeType like "Country" or "State". What do you think of renaming this to something like "comparisonRelationshipType" or "comparisonPlacesRelationshipType"?

static/js/place/dev_place_main.tsx
```diff
@@ -81,7 +133,7 @@ function placeChartsApiResponsesToPageConfig(
       const charts = chartsByCategory[categoryName];
 
       const tiles: TileConfig[] = charts.map((chart) => {
-        return {
+        const chartConfiguration = {
```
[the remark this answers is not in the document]
gmechali: Done.

static/js/place/dev_place_main.tsx
```diff
@@ -530,10 +599,23 @@ export const DevPlaceMain = (): React.JSX.Element => {
...
   const hasNoCharts =
     place && pageConfig && pageConfig.categories.length == 0 && !isLoading;
+  const placeTypesForCharts = new Set<string>([
+    "Continent",
+    "Country",
+    "AdministrativeArea1",
+    "EurostatNUTS1",
+  ]);
+
+  function filterForCharts(places: NamedTypedPlace[]): NamedTypedPlace[] {
```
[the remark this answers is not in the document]
gmechali: This might be something still to fix - there's always a lot of parents, and we don't really care about all of them. I have this to find the parent that we care about and pass those through. 
Later in the stack, we just pick the first parent in the list if we need it. 

For example, place page for france, we would only pass in parent places of the types in  placeTypesForCharts, which will be Europe only. 
Then when we see a chart for SIMILAR_IN_PARENT, we would fetch the children of Europe. 

We already have this: PARENT_PLACE_TYPES_TO_HIGHLIGHT in the utils which is probably the right list to use.

static/js/types/subject_page_proto_types.ts
```diff
@@ -165,6 +165,7 @@ export interface TileConfig {
   type: string;
   statVarKey: string[];
   comparisonPlaces?: string[];
+  comparisonPlaceType?: string;
```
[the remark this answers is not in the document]
gmechali: I think we definitely need to keep this, because we put that field in the tile config to have access to it in block.tsx to determine what to pass in the <Ranking tag.

server/config/chart_config/economics.json
```diff
@@ -199,6 +199,50 @@
...
+    "chartType": "BAR",
+    "comparisonPlaceType": "SIMILAR",
+    "barTileSpec": {
+      "maxPlaces": 5
+    }
+  },
+  {
+    "category": "Economics",
+    "titleId": "CHART_TITLE-Gdp-other-countries-ranking",
+    "title": "Gross domestic product Other Countries ranking",
```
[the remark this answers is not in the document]
gmechali: I kept it in the title to be able to differentiate the charts easily, but I removed it from the translated message.

static/js/place/dev_place_main.tsx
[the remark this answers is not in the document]
gmechali: Copied over that list, and added function comment.

server/config/chart_config/economics.json
```diff
@@ -199,6 +199,50 @@
...
+    "statsVars": [
+      "Amount_EconomicActivity_GrossDomesticProduction_Nominal"
+    ],
+    "unit": "$",
+    "isOverview": true,
+    "isRankingChart": true
+  },{
+    "category": "Economics",
+    "titleId": "CHART_TITLE-Gdp-other-countries",
+    "title": "Gross domestic product Other Countries",
```
dwnoble, line 216: Keep the title generic here, since this chart will show up for other place types besides countries.

The strategy used by the old place page is to use titles like "Gross domestic product", and injecting the "other countries" part using the react i18n library here: https://github.com/datacommonsorg/website/blob/78fbb8b2111916015fefd7dee487f24c8eabdba2/static/js/place/chart_block.tsx#L121

static/js/types/subject_page_proto_types.ts
```diff
@@ -165,6 +165,7 @@ export interface TileConfig {
   type: string;
   statVarKey: string[];
   comparisonPlaces?: string[];
+  comparisonPlacesRelationshipType?: string;
```
dwnoble, line 168: Changes here should also be added to `server/config/subject_page.proto` (also, when changes are made to that proto, after running run_server.sh, some new python type definitions will get generated

static/js/components/subject_page/block.tsx
```diff
@@ -386,6 +391,61 @@ export function Block(props: BlockPropType): JSX.Element {
...
+/**
+ * Find the correct place to consider the parent place for the current component. This will depend on the current
+ * place, parent places and the comparisonPlacesRelationshipType.
+ *
+ * @param comparisonPlacesRelationshipType String representing the type of comparison places.
+ * @param parentPlaces List of all possible parent places. We usually pick the first one.
+ * @param place The current place to evaluate.
+ * @returns The place dcid for the place to consider the parent, it can be a parentPlace, or the current place.
+ */
+function getParentPlaceDcid(
```
dwnoble, line 403: I think this approach could lead to a bug where the title of the chart doesnt match its contents.

working example:
Place = iceland
parent place = europe
comparisonPlacesRelationshipType = similar in parent
chart title = "Population: other countries"

This works because both getParentPlaceDcid & and getEnclosedPlace type will select Europe & Country respectively

Potentially buggy case
Place = [46528](https://autopush.datacommons.org/place/zip/46528) (random zip code in indiana)
parent place = [Elkhart County](https://datacommons.org/place/geoId/18039)
comparisonPlacesRelationshipType = similar in parent
chart title = "Population: other zip codes in indiana"

We don't have "ZipCodeTabulatedArea" accounted for in the `placeTypeLevelUp` structure, so we'd end up seeing a ranking chart titled "Population: other zip codes in indiana" but the data would be for other "cities in the zip code of 46528"
place=[46528](https://autopush.datacommons.org/place/zip/46528)
enclosedPlaceType="City"

dwnoble: Looking at the subject page proto definition, we have an existing field for [placeDcidOverride / place_dcid_override](https://github.com/datacommonsorg/website/blob/78fbb8b2111916015fefd7dee487f24c8eabdba2/server/config/subject_page.proto#L351)

If we added another field for "enclosed_place_type_override", then we could do the logic of figuring out if we can properly "uplevel" the parent/enclosed place relationship when we're building the tile, and we can figure out the place title then.

What do you think?

PR #4835, opened 2025-01-10
~80 additional lint fixes

gmechali:
Fixed some react hook dependency arrays, some lint errors to skip and some return types to add.

Down to 99 lint warnings.


static/js/components/tiles/gauge_tile.tsx
```diff
@@ -151,7 +151,18 @@ export function GaugeTile(props: GaugeTilePropType): JSX.Element {
   );
 }
 
-const fetchData = async (props: GaugeTilePropType) => {
+const fetchData = async (
+  props: GaugeTilePropType
+): Promise<{
```
chejennifer, line 156: nit: for these large objects as return types, maybe should define a type for them & in fact, for this one, can just use GaugeChartData

import
chejennifer, line 1: did you mean to include this? and the one for mixer?

chejennifer (approved):
thanks for fixing!

static/js/components/tiles/gauge_tile.tsx
gmechali: So I used GaugeChartData here.

Right now most of these large objects that we should name I think I skipped them (they're in the remaining 99 lints). 
Some of the ones in this PR are a bit large but I think they're always only ever used once / twice and are very custom so I didnt think these deserved to be typed properly. Lmk if you think there are some in this PR that should be!

import
gmechali: Fixed now!

PR #4836, opened 2025-01-10
Skip empty embeddings in custom DC

keyurva:
* Some custom DCs may not have SVs or topics, hence the embeddings file will be empty.
* Prior to this fix, starting up a service with such files failed.
* With this fix, in custom DC mode, we'll log a warning in such instances and skip that index.
* In base DC mode, it will continue to fail.


chejennifer (approved):
nice!

PR #4838, opened 2025-01-13
[Place Page Revamp] Refactors place page into using the new json format for the home page.

gmechali:
Uses blocks to display the charts in the place page - adds the per capita checkbox when needed.
Uses a pattern of matching Line charts with highlight charts, Maps with rankings charts, and uses bar charts for comparisons.

See screencast showing current behavior: https://screencast.googleplex.com/cast/NjA5MzQyODE3MjUyMTQ3Mnw5ODZjN2RiNS1mZg
Still some bugs to resolve, and more charts need to be configured but this sets up the foundations.


static/js/utils/app/visualization_utils.tsx
```diff
@@ -331,7 +331,7 @@ export function getHash(
 ): string {
   const params = {
     [URL_PARAMS.VIS_TYPE]: visType,
-    [URL_PARAMS.PLACE]: places.join(PARAM_VALUE_SEP),
+    [URL_PARAMS.PLACE]: places ? places.join(PARAM_VALUE_SEP) : "",
```
dwnoble, line 334: What's this for?

server/routes/dev_place/utils.py
```diff
@@ -148,13 +152,15 @@ def filter_chart_config_by_place_dcid(chart_config: List[Dict],
...
-    if config.get("isChoropleth"):
-      map_stat_var_dcids.extend(config["statsVars"])
-    else:
-      non_map_stat_var_dcids.extend(config["statsVars"])
-    denominator = config.get("relatedChart", {}).get("denominator", None)
-    if denominator:
-      non_map_stat_var_dcids.extend(denominator)
+    for block in config["blocks"]:
+      for chart in block["charts"]:
+        if chart.get("type") == "MAP":
```
dwnoble, line 157: I think we need to change this approach to come up with existence checks for 3x sets of variables based on the `placeScope` (instead of whether its type map or not):

3x types of existence checks:
* For the current place
* For child places
* For peer places within the same parent

And below, instead of filtering out the entire chart config, we can just filter out individual "blocks" if they don't have data for their specified scope

server/lib/util.py
```diff
@@ -49,6 +49,11 @@
     "housing",
     "environment",
     "energy",
+    "energy_new",
```
dwnoble, line 52: Would it be easier to separate this out into a new list and read in the new chart configs entirely separately from the old place page chart configs?

server/routes/dev_place/utils.py
gmechali: Yeah so I already implemented but I did it in a different PR:
https://github.com/datacommonsorg/website/pull/4842

Was going to try to phase some of it since it's a lot of changes at once. So this current PR adds a bunch of charts that arent working properly but im hoping that next PR can be submitted as a fast follow :)
Just need to add some tests to it and it should be good to go!

server/routes/dev_place/utils.py
```diff
@@ -279,53 +285,39 @@ def chart_config_to_overview_charts(chart_config, child_place_type: str):
...
-        # Related charts only specify one denominator, so we repeat it for each stat var
-        related_chart.denominator = [
-            page_config_related_chart.get("denominator")
-        ] * len(chart.statisticalVariableDcids)
-      if is_map_chart:
-        related_chart.childPlaceType = child_place_type
-      charts.append(related_chart)
-
-  return charts
+    for block in page_config_item["blocks"]:
```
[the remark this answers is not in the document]
gmechali: Isnt it there already? 
It's the changes in: server/routes/dev_place/types.py or are you referring to something else?

static/js/types/subject_page_proto_types.ts
```diff
@@ -166,6 +166,7 @@ export interface TileConfig {
   statVarKey: string[];
   comparisonPlaces?: string[];
   placeDcidOverride?: string;
+  enclosedPlaceTypeOverride?: string;
```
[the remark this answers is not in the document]
gmechali: Done. Can you just double check I added it in the right place. Also why do we need it there? What's that file for?

static/js/utils/app/visualization_utils.tsx
gmechali: Some error I found in the logs. 
this tries to render before the page is set. There's also the possibility of just skipping the rendering if place isnt set which wasnt dont before.

server/lib/util.py
gmechali: Yeah I went back and forth on this a bit. I'd say the nice part here is that when we're done with the old place page, we just remove these other filenames from the list. 
What's nice about one list means all our tests and stuff still apply, though we had to have it read the new variable field properly. But we get all that with minimal changes. Changing how we read entirely means we may want to duplicate a lot more stuff to have the same effect? So I'm not sure what's better honestly. Settled for this, but I dont mind changing it if you have a strong preference.

server/routes/dev_place/utils.py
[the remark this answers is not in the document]
dwnoble: It looks like there's a definition for the API response & its blocks, but i mean for the `config/chart_config/*_new.json` files themselves

server/routes/dev_place/utils.py
dwnoble: That sounds good!

server/lib/util.py
dwnoble: That's true about needing to duplicate a bunch of logic if we separated these into a different folder. I don't have a strong preference so your approach lgtm!

static/js/types/subject_page_proto_types.ts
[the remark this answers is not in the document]
dwnoble: Looks like it's missing a semicolon at the end of the line in the .proto file, but otherwise is in the right spot!

Can you also run `./run_server.sh` locally to re-generate & check-in the `subject_page_pb2.py` file?

That proto file is used to generate the python representation of the SubjectPageConfig & Tile config classes in `server/config/subject_page_pb2.py`. Those definitions are mainly used in the "topics" page (example: the [topic_page/equity/US_Places.textproto](https://github.com/datacommonsorg/website/blob/3b9ca55d3570250aa8a7d0d4ff3ebdc10f2a615a/server/config/topic_page/equity/US_Places.textproto#L4) file is used to generate this page: https://datacommons.org/topic/equity/geoId/06) and the NL search endpoint to encode [SubjectPageConfig + tile config objects directly into the response](https://github.com/datacommonsorg/website/blob/3b9ca55d3570250aa8a7d0d4ff3ebdc10f2a615a/server/lib/nl/explore/fulfiller_bridge.py#L40)

server/routes/dev_place/utils.py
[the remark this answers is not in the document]
gmechali: Oh I see done. 
I went with naming of ServerChartConfiguration, ServerBlockMetadata, ServerChartMetadata
I liked having the "server" in there to differentiate between what we send back to the FE and the setup from the server configs. And then used block/chart for the appropriate levels with Metadata as the qualifier. 
Lmk if you see a better naming scheme.

static/js/types/subject_page_proto_types.ts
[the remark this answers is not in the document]
gmechali: Sg, thanks for explaining! Added the file.  There's definintely a bunch of changes and even when I dont modify the proto file, when I run server locally, it makes some of these modifications to the subject_page_pb2.py file, so just wanna check that these are expected?

server/routes/dev_place/utils.py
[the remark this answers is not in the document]
gmechali: also since I added the type here it became necessary to follow the snake case naming for the json format in order to appease lint and keep 1-1 conversion with the fields in the json and in the new classes.
for example: titleId --> title_id etc...

dwnoble (approved):

PR #4839, opened 2025-01-14
Update NodeJS query differ goldens

hqpho:


chejennifer (approved):
thanks for updating!

hqpho:
> thanks for updating!

and thank you for the speedy review!!

PR #4843, opened 2025-01-15
Updated terraform configuration to initialize empty input and ouput directories

dwnoble:
New variables:

```
gcs_data_bucket_output_folder = "output"
gcs_data_bucket_input_folder = "input"
```

Renamed `dc_gcs_data_bucket_path_override` to `gcs_data_bucket_name`
Renamed `dc_gcs_data_bucket_location` to `gcs_data_bucket_location`


deploy/terraform-custom-datacommons/modules/variables.tf
```diff
@@ -73,6 +73,18 @@ variable "dc_gcs_data_bucket_path_override" {
   default     = ""
 }
 
+variable "dc_gcs_data_bucket_input_folder" {
```
kmoscoe, line 76: This is great!  I wonder if we need the "dc" prefix, since we don't use it for other services. Also, is it possible to remove "path" from the previous option? It's kind of misleading since it's only the bucket name. I would just call it "gcs_data_bucket_name".

deploy/terraform-custom-datacommons/modules/main.tf
```diff
@@ -360,7 +374,7 @@ resource "google_cloud_run_v2_job" "dc_data_job" {
 
         env {
           name  = "INPUT_DIR"
-          value = "gs://${local.dc_gcs_data_bucket_path}/input"
+          value = "gs://${local.dc_gcs_data_bucket_path}/${var.dc_gcs_data_bucket_input_folder}"
```
kmoscoe, line 377: Do we also need one for OUTPUT_DIR, since we're giving them that variable too?

dwnoble: Already added that! `dc_gcs_data_bucket_output_folder`

deploy/terraform-custom-datacommons/modules/variables.tf
dwnoble: Good call. Removed `dc_` prefix to all of the bucket variables, and renamed the bucket name to `gcs_data_bucket_name`

kmoscoe (approved):

deploy/terraform-custom-datacommons/modules/variables.tf
```diff
@@ -67,13 +67,25 @@ variable "google_analytics_tag_id" {
...
+  default     = "input"
+}
+
+variable "gcs_data_bucket_output_folder" {
+  description = "Output data folder in the GCS data bucket"
+  type        = string
+  default     = "output"
+}
+
+variable "gcs_data_bucket_location" {
```
keyurva, line 88: For my understanding, what's the difference between `gcs_data_bucket_name` and `gcs_data_bucket_location`?

keyurva (approved):

deploy/terraform-custom-datacommons/modules/variables.tf
kmoscoe: It's really the region where it's hosted. For some reason GCS buckets don't match the regions used for SQL and other resources.

PR #4844, opened 2025-01-15
Add flake detection mode for webdriver tests

hqpho:
- Also add a shared helper for clicking elements and use it to deflake a test
- Document all of the above


server/webdriver/shared.py
```diff
@@ -64,6 +64,13 @@ def click_sv_group(driver, svg_name):
       break
 
 
+def click_el(driver, element_locator):
```
gmechali, line 67: It might be good to return the element found as well? to give the caller the option to do something with the element if needed.

gmechali (approved):

server/webdriver/shared_tests/vis_timeline_test.py
```diff
@@ -142,20 +142,21 @@ def test_manually_enter_options(self):
     self.driver.get(self.url_ + TIMELINE_URL.replace('#visType=timeline', ''))
 
     # Click the timeline tab
+    vis_types_clickable = EC.element_to_be_clickable(
```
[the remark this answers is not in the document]
hqpho: Can't because it needs to check for which of the tabs has 'Timeline' on it

server/webdriver/shared.py
hqpho: Good idea, done!

PR #4848, opened 2025-01-16
[Place Page Revamp] Add logic for filtering charts

gmechali:
Implements the fix for removing charts that have no data on either of the cases according to their block scope.
So either for the current place, the child places, or the peer places within the parent.

Modifies the call to relatedPlaces such that it comes BEFORE placeCharts (as opposed to in parallel). We have to do this in order to pass in the parent place dcid to the placeCharts method so that we can verify we have data for peer places within parent.

With this change, all charts that have NO data are removed.

### DIFF
--- a/server/routes/dev_place/utils.py
+++ b/server/routes/dev_place/utils.py
@@ -155,19 +155,22 @@ def place_type_to_highlight(place_types: List[str]) -> str:
 
 
 def get_place_override(parent_places: List[Place]) -> str:
-    place_override = None
-    for place in parent_places:
-        # Python's way to achieve the same logic as your TypeScript code
-        try:
-            lowest_index = min(PARENT_PLACE_TYPES_TO_HIGHLIGHT.index(type) for type in place.types if type in PARENT_PLACE_TYPES_TO_HIGHLIGHT)
-        except ValueError:  # No matching types found
-            lowest_index = float('inf')  # Equivalent to Infinity in this context
+  place_override = None
+  for place in parent_places:
+    # Python's way to achieve the same logic as your TypeScript code
+    try:
+      lowest_index = min(
+          PARENT_PLACE_TYPES_TO_HIGHLIGHT.index(type)
+          for type in place.types
+          if type in PARENT_PLACE_TYPES_TO_HIGHLIGHT)
+    except ValueError:  # No matching types found
+      lowest_index = float('inf')  # Equivalent to Infinity in this context
 
-        if lowest_index != float('inf'):
-            place_override = place
-            break  # Exit the loop once a match is found
+    if lowest_index != float('inf'):
+      place_override = place
+      break  # Exit the loop once a match is found
 
-    return place_override.dcid if place_override else None  # Use None instead of undefined
+  return place_override.dcid if place_override else None  # Use None instead of undefined
 
 
 def filter_chart_configs_for_category(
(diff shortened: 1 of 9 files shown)


static/js/place/dev_place_main.tsx
```diff
@@ -467,16 +468,19 @@ export const DevPlaceMain = (): React.JSX.Element => {
...
-        await Promise.all([
-          defaultDataCommonsWebClient.getPlaceCharts({
-            category,
-            placeDcid: place.dcid,
-          }),
-          defaultDataCommonsWebClient.getRelatedPLaces({
-            placeDcid: place.dcid,
-          }),
-        ]);
+      const relatedPlacesApiResponse =
```
dwnoble, line 471: Instead of making the calls synchronous, would it make sense to add a call to
```
parent_places = place_utils.get_parent_places(place.dcid)
```
 to `def place_charts`?
 
 (Looks like the getPlaceOverride logic would have to move up to the web server there as well)

server/routes/dev_place/utils.py
```diff
@@ -173,48 +190,94 @@ def filter_chart_config_by_place_dcid(
...
+      needs_current_place_data |= block.place_scope == "PLACE"
+      needs_child_data |= block.place_scope == "CHILD_PLACES"
+      needs_peer_places_data |= block.place_scope == "PEER_PLACES_WITHIN_PARENT"
+
+    if needs_child_data:
+      child_places_stat_var_dcids.extend(config.variables)
+    if needs_current_place_data:
+      current_place_stat_var_dcids.extend(config.variables)
+      denominator = config.denominator
+      if denominator:
```
dwnoble, line 211: Should the denominator check be handled separately? Maybe if the place scope doesn't have values for the denominator we can disable the per capita checkbox

server/routes/dev_place/utils.py
```diff
@@ -173,48 +190,94 @@ def filter_chart_config_by_place_dcid(
...
       config for config in chart_config
       # Set intersection to see if this chart has any variables with observations for our place_dcid
       if set(config.variables) & stat_vars_with_observations_set
   ]
+
+  for config in filtered_chart_config:
+    updated_blocks = []
+    for block in config.blocks:
+      if block.place_scope == "CHILD_PLACES":
+        has_child_data = any(var in child_places_stat_vars_with_observations
```
dwnoble, line 262: Can this (& the similar cases below) use the `set` version of `*_stat_vars_with_observations`?

static/js/place/dev_place_main.tsx
dwnoble: Also - it looks like getPlaceOverride can return undefined in some cases- can you make sure to handle that in the place charts API?

server/routes/dev_place/utils.py
gmechali: done.

server/routes/dev_place/utils.py
gmechali: oh yeah that would make sense. Pushing this off to a follow up. Need to think about whether we need this for all the peer/child as well?

static/js/place/dev_place_main.tsx
gmechali: Ah yeah good point. Switched over to the api side call. And yea it should still work if we return None for the place override! tested out with place/Earth, we just wont have any existing data for the peer places within so they get filtered out

dwnoble (approved):

PR #4849, opened 2025-01-16
Added incremental i18n support to new places page

dwnoble:
Changes
* Added `hl=<lang>` param to api calls
* Translated category tab names
* Translations for "Places in" box
* Displays translated variable names

## Before (hl=fr)
i18n_before

## After (hl=fr)
i18n_after


packages/client/src/data_commons_web_client.ts
```diff
@@ -214,11 +216,19 @@ class DataCommonsWebClient {
...
+    if (params.locale) {
+      urlQueryParams.set(LOCALE_PARAM, params.locale);
+    }
+    if (params.category) {
+      urlQueryParams.set("category", params.category);
+    }
     const url = `${this.apiRoot || ""}/api/dev-place/charts/${
       params.placeDcid
-    }${params.category ? "?category=" + params.category : ""}`;
+    }${urlQueryParams.size ? "?" + urlQueryParams.toString() : ""}`;
```
gmechali, line 231: nit: can we just do urlQueryParams.toString() directly? since it would just be an empty string if size is 0.

packages/client/src/data_commons_web_client.ts
```diff
@@ -229,11 +239,16 @@ class DataCommonsWebClient {
    * @param params.placeDcid place dcid to fetch data for
    */
   async getRelatedPLaces(params: {
+    locale?: string;
```
gmechali, line 242: nit: thoughts on putting the optional parameter at the end?

server/routes/dev_place/api.py
```diff
@@ -57,13 +51,15 @@ def place_charts(place_dcid: str):
...
-  # Ensure category is valid
-  place_category = request.args.get("category", OVERVIEW_CATEGORY)
+  # Get parent place DCID
   parent_place_dcid = place_utils.get_place_override(
       place_utils.get_parent_places(place_dcid))
-  if place_category not in CATEGORIES:
+
+  # Ensure category is valid
+  place_category = request.args.get("category", place_utils.OVERVIEW_CATEGORY)
+  if place_category not in place_utils.CATEGORIES:
```
gmechali, line 60: Should we move this above the call to get the parent places? There's no need to execute that call if this is going to fail!

gmechali (approved):

packages/client/src/data_commons_web_client.ts
dwnoble: Good catch. I refactored this to use the `toURLSearchParams` helper & to leave out the ternary

packages/client/src/data_commons_web_client.ts
dwnoble: Done

server/routes/dev_place/api.py
dwnoble: done

PR #4850, opened 2025-01-16
[Place Page Revamp] Optimization on existence check and force include the place.

gmechali:
Optimizes the existence check by always running the existence check on the full chart config, then filtering out the blocks that are for a different category. Previously we ran 2 existence checks, on the full chart config (to know which categories have data) and then on the category only chart config. 

We also remove the bar chart blocks if there is no data for the current place.

And we now pass in a list of DCIDs as peers within the parent which returns peer places within a parent to highlight.
That list is used in the comparisonPlaces param in the TileConfig. Note that we randomize the order of the list on the server side.

https://screencast.googleplex.com/cast/NDk5MDU2ODg4NjE3MzY5NnxkMDVkOWM0YS02Yw


server/routes/dev_place/utils.py
```diff
@@ -117,7 +102,31 @@ def get_place_type_with_parent_places_links(dcid: str) -> str:
...
+  place_type = place_api.api_place_type(dcid)
+  place_type_display_name = place_api.get_place_type_i18n_name(place_type)
+
+  # Get parent places
+  all_parents = get_parent_places(dcid)
+
+  # Get parent places
+  parents_to_include = get_ordered_parents_to_highlight(all_parents)
   parent_dcids = [parent.dcid for parent in parents_to_include]
+  all_parents = get_parent_places(dcid)
```
dwnoble, line 128: Looks like this is called an extra time here

dwnoble (approved):

PR #4856, opened 2025-01-17
Place page style fixes

dwnoble:
Added:
* Updated "Knowledge Graph" link display
* Updated styles in place overview box
* Migrated some styles to emotion
* Updated padding in subject page display
* Moved place overview component to its own file

Left for a future PR:
* Remove overview tab & have page h1 title link to overview
* Add ranking to overview panel
* Update Key Demographics text to match mocks
* Update key demographics dates to only show the year

Before
before

After
after


miss-o-soup:
Looking good to me. Let me know when the updates are in staging so that I can take a closer look.
One important place page to look into is Earth as this gets a lot of traffic. The page will be longer but want to make sure apart from that all the content is there and looks good.

miss-o-soup (approved):
Looking good to me. Let me know when the updates are in staging so that I can take a closer look.
One important place page to look into is Earth as this gets a lot of traffic. The page will be longer but want to make sure apart from that all the content is there and looks good.

static/js/place/util.ts
```diff
@@ -433,6 +466,29 @@ const pluralPlaceTypeMessages = defineMessages({
 });
 
 export const pageMessages = defineMessages({
+  KnowledgeGraph: {
```
gmechali, line 469: No need to change it in this PR, but what do you think about extracting all messages to _messages.ts files in the future? They take up a lot of space in these files that control a lot of logic and kinda bloats it making it harder to read.

static/js/components/subject_page/category.tsx
```diff
@@ -68,8 +68,18 @@ export const Category = memo(function Category(
     <article className="category col-12" id={props.id}>
       {title && (
         <h2 className="block-title">
-          {props.config.url && <a href={props.config.url}>{title}</a>}
-          {!props.config.url && <span>{title}</span>}
+          {!props.config.url ? (
```
gmechali, line 71: Would this ever happen? Should we just make it a non-optional field?

static/js/place/dev_place_main.tsx
```diff
@@ -326,7 +169,7 @@ const RelatedPlaces = (props: {
     return null;
   }
 
-  const NUM_PLACES = 15;
+  const NUM_PLACES = 100;
   const showToggle = childPlaces.length > NUM_PLACES;
```
gmechali, line 173: so we're still keeping the toggle mechanism just raising the threshold?

gmechali (approved):

static/js/place/util.ts
dwnoble: I like that idea- i'll migrate to a _messages.ts file in a follow-up

static/js/components/subject_page/category.tsx
dwnoble: Looks like we support showing a URL in the category section, but it's hard to tell where that's actually used. Here's a reference to setting the url field: https://github.com/datacommonsorg/website/blob/67e8610cc84b73d6654f3b980d92e141b54e1360/static/js/apps/explore/app.tsx#L261

@chejennifer do you recall if the subject page config 
"Category" object ever renders the URL?

static/js/place/dev_place_main.tsx
dwnoble: I figured with moving the related places to the bottom of the page we can show more by default, but i still wanted to have a limit in case some edge case tried to show a huge list. Maybe we can get rid of the limit entirely-wdyt @gmechali ?

static/js/components/subject_page/category.tsx
chejennifer: I tried checking different places where I know the subject page config is used (/explore, /disasters, /topic, and /event) and have not seen any cases of the category object rendering a URL so maybe safe to just remove the url field?

dwnoble: That sounds good - ill keep this behavior for now and add a todo to combine url & link_text to a single field

beets: sorry i don't recall. i did a quick spot check on datacommons.techsoup.org and don't see it used there either

PR #4859, opened 2025-01-17
Make sure submods are initialized before merge

hqpho:
Otherwise the merge script is a no-op on a fresh clone of the repo.


ajaits (approved):
Thanks! this is simpler than 2 separate commands

PR #4861, opened 2025-01-17
Remove lancedb.

keyurva:
* Local development note
   + If your local NL server fails to startup after this change, delete any locally downloaded models prior to startup: `rm -rf /tmp/datcom-nl-models`


chejennifer (approved):
thanks for removing!

PR #4863, opened 2025-01-17
Add logging to understand flaky python test

chejennifer:
The python cloudbuild has been seeing transient errors having to do with embeddings file being empty ([example](https://pantheon.corp.google.com/cloud-build/builds;region=global/eaf1afc5-d2cd-4625-b650-2ed1bf7dc6a6;step=1?e=13803378&inv=1&invt=AbnIfg&mods=-monitoring_api_staging&project=datcom-ci)). However, I have not been able to trigger the error after re-building the cloudbuild multiple times: https://screenshot.googleplex.com/6TBMmvmyLCUT8fA

Adding an extra line in the logging to use to debug if the transient error shows up again.

### DIFF
--- a/nl_server/store/memory.py
+++ b/nl_server/store/memory.py
@@ -66,8 +66,8 @@ class MemoryEmbeddingsStore(EmbeddingsStore):
     logging.info('Loading embeddings file: %s', embeddings_path)
     try:
       ds = load_dataset('csv', data_files=embeddings_path)
-    except:
-      error_str = f'No embedding could be loaded for {embeddings_path}'
+    except Exception as e:
+      error_str = f'No embedding could be loaded for {embeddings_path}: {e}'
       logging.error(error_str)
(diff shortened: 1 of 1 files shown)


nl_server/store/memory.py
```diff
@@ -67,9 +67,9 @@ def __init__(self, idx_info: MemoryIndexConfig) -> None:
     try:
       ds = load_dataset('csv', data_files=embeddings_path)
     except:
```
keyurva, line 69: For troubleshooting, knowing what the original exception can also help. So consider capturing and logging that as well:

```
  except Exception as e:
    raise Exception(f'No embedding could be loaded for {embeddings_path}: {e}')
```

You can also do exception chaining but it requires the upstream code to log the cause so ok to go with just logging it here.

keyurva (approved):

PR #4864, opened 2025-01-18
Data Sources

nick-nlb:
## Description

This PR implements the tabbed data sources pages.

To facilitate this, we have developed a generic tab component that is used here and can be used elsewhere.
The tab component can optionally be “routed” or “standard”. We are using routed tabs for the data sources.
With routed tabs, when you select a tab (such as “demographics”) then “demographics” will appear in the url. For example, with the base page “/data”, the demographics tab will be “/data/demographics”. If the user arrives directly on the page “/data/demographics” (for example, from a link or a bookmark), then the demographics tab will be selected by default.

Standard tabs (available within the tab component, but not used here), work very similarly. However, the URLs are not altered, and the tabs are handled entirely by internal state. These could be useful if, for example, you have multiple tab components on a page, where it may not make sense to put the tab on the URL.

### Tab Design

The look and feel of the tab design was inspired by Material Design and tab sets found on Google pages. Following this, we made the tabs responsive by adding navigation arrows dynamically when the tab set container can no longer contain all the tabs. Users can navigate the tabs by either clicking those arrows, or on a mobile device by swiping. 

### Notes

The data used from the data sources come from the following pages in the docs:
https://docs.datacommons.org/datasets/Energy.html (where “Energy” can be swapped out with “Health”, etc). In order to get this information in a form that can be used in the frontend, we have provided a Python script that parses the relevant doc pages and creates the json that we then use.

This measure may be temporary. In the future, the same scripts that generate the doc pages (located outside of this repository) could also generate the json used by the data page implemented here. However, pending the that implementation, the provided script here can update the content of these data pages.

The script does not need to be run to test or merge this PR. It has already been run and the output has been committed. 

We only need to run it if we wish to update the data based on changes to the docs. Instructions on running the script are provided in a README file in the same directory (/tools/data_sources/README.md).

### DIFF
--- a/static/js/apps/data/components/data_source_details.tsx
+++ b/static/js/apps/data/components/data_source_details.tsx
@@ -118,6 +118,9 @@ export const DataSourceDetails = ({
                   @media (max-width: ${theme.breakpoints.md}px) {
                     margin-left: ${theme.spacing.md}px;
                   }
+                  code {
+                    color: ${theme.colors.text.secondary.base};
+                  }
                 `}
               >
                 <h4
--- a/static/js/apps/data/data_sources.json
+++ b/static/js/apps/data/data_sources.json
@@ -1273,7 +1273,7 @@
           {
             "label": "National Emissions Inventory",
             "url": "https://www.epa.gov/air-emissions-inventories/national-emissions-inventory-nei",
-            "description": "The National Emissions Inventory (NEI) is a comprehensive and detailed estimate of air emissions of criteria pollutants, hazardous pollutants and greenhouse gases from 188 onroad air emission sources (mobile sources such as highway and border crossing vehicles electricity measurement), 248 nonroad air emissions sources (mobile sources such as off-highway vehicle gasoline measurement), 703 nonpoint air emissions sources (such as industrial processes, oil and gas exploration and production) and 5818 point air emissions sources (such as chemical and organic solvent evaporation measurement) at the US county level."
+            "description": "The National Emissions Inventory (NEI) is a comprehensive and detailed estimate of air emissions of criteria pollutants, hazardous pollutants and greenhouse gases from 188 on-road air emission sources (mobile sources such as highway and border crossing vehicles electricity measurement), 248 non-road air emissions sources (mobile sources such as off-highway vehicle gasoline measurement), 703 non-point air emissions sources (such as industrial processes, oil and gas exploration and production) and 5818 point air emissions sources (such as chemical and organic solvent evaporation measurement) at the US county level."
           },
           {
             "label": "Superfund Sites",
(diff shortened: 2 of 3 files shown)


static/js/apps/data/main.ts
```diff
@@ -0,0 +1,40 @@
...
+
+import { loadLocaleData } from "../../i18n/i18n";
+import { App } from "./app";
+
+window.addEventListener("load", (): void => {
+  loadLocaleData("en", [import("../../i18n/compiled-lang/en/units.json")]).then(
+    () => {
+      renderPage();
+    }
+  );
```
beets, line 32: the units.json strings might not be needed for data sources

tools/data_sources/README.md
```diff
@@ -0,0 +1,31 @@
...
+
+## Purpose
+
+This script is designed to scrape the existing data sources pages (external to)
+the primary site, and convert those data sources to json in a format that 
+is usable by the data sources page in the primary Data Commons site.
+
+## Notes
+
+This script may be temporary. The external data sources pages are populated
```
beets, line 11: could you please update this to a TODO for us to migrate into updating this page directly - thanks!

static/js/apps/data/components/data_sources.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+interface DataSourcesProps {
+  dataSources: DataSourceTopic[];
+}
+
+export const DataSources = ({
+  dataSources,
+}: DataSourcesProps): ReactElement => {
+  const dataSourceTabs = dataSources.map((topic) => ({
+    value: topic.slug,
```
beets, line 54: could we use "slugify" here, rather than having it hardcoded in json (and easier to be set incorrectly there)

beets: alternatively, perhaps the property "value" is too vague, and should be set to "route" or something more indicative of it's eventual use

static/js/components/elements/tabs/routed_tabs.tsx
```diff
@@ -0,0 +1,94 @@
...
+interface RoutedTabsInterface {
+  //the currently active tab value
+  activeTab: string;
+  //a function to handle tab changes
+  onTabChange: (newValue: string | number) => void;
+}
+
+export function useRoutedTabs({
+  enabled = true,
+  tabValues,
```
beets, line 44: could we throw an error in developer mode if these are not set to unique strings which are uri compliant?

beets (commented):
thanks for setting this up! it looks good and cleanly implemented. appreciate the script to update the json.

one comment on slugify, and doing it outside the json within code. is it set up in the json since it's used for routing (rather than in data_sources.tsx). i initially was concerned about potentially introducing errors in the json, but perhaps we need to add more error handling in the component (e.g. if they end up being non-unique, or are set to strings which are incompatible for routing).

beets (approved):
thank you for the updates! i'm excited to see these changes go out :)

please fix any new linter issues you find as we're working on reducing those to 0.

beets:
could you also please run the importer one more time before submitting as there have been some recent changes to the source files -- thanks!

PR #4866, opened 2025-01-20
[Place Page Revamp] Filter out bar charts unless we have data for 2+ peers within parents

gmechali:
Removes charts like this: [autopush](https://screenshot.googleplex.com/ocLuP7AYsis8bNW) [local](https://screenshot.googleplex.com/53yFY92Fr95om3r).

Only keeping charts if we have 2 or more places to compare it with.
For that, I fetch the peers within parent list and trim to the first 15 (that's how many we use for now), and check that at least 2 have data for each sv. We rely on the cache of fetch_peers_within_parent to guarantee that the same list will be returned from the related_places API call and in the filter_charts call within the api call to fetch charts.


server/routes/dev_place/utils.py
```diff
@@ -731,6 +741,24 @@ def fetch_nearby_place_dcids(place: Place, locale=DEFAULT_LOCALE) -> List[str]:
   return nearby_place_dcids
 
 
+@cache.memoize(timeout=TIMEOUT)
+def fetch_peer_places_within(place_dcid: str, place_types: list[str]):
```
dwnoble, line 745: Nit: add docstring & return type annotation

server/routes/dev_place/utils.py
```diff
@@ -731,6 +741,24 @@ def fetch_nearby_place_dcids(place: Place, locale=DEFAULT_LOCALE) -> List[str]:
...
+  parent_places = get_parent_places(place_dcid)
+  parents_to_highlight = get_ordered_parents_to_highlight(parent_places)
```
dwnoble, line 747: this may be outside the scope of this PR, but can you expand on the docstring with a few concrete examples to the various `*_to_highlight` methods? I find myself having to pause when im reading the code to remember what we meant by "highlight"

server/routes/dev_place/utils.py
```diff
@@ -266,14 +267,23 @@ def filter_chart_config_by_place_dcid(
           stat_var_dcid, {}).get("byEntity", {})
   ])
 
-  # find stat vars that have data for our peer places.
+  # find stat vars that have data for our peer places. We only want to keep
+  # these stat vars if there is data for more than one place.
+  peer_places_within = fetch_peer_places_within(place_dcid, place_type)[:15]
   peer_places_obs_point_response = dc.obs_point_within(
```
dwnoble, line 273: It looks like the place_dcid and place_type used here in the call to obs_point_within don't exactly line up with the parent & child type used in `fetch_peer_places_within`. Can you take a look?

I think we should make sure the two call use the same parent+childType combo, or maybe we can just pass the list of peer_places_within directly and call `dc.obs_point` (without the within)

server/routes/dev_place/utils.py
```diff
@@ -731,6 +741,24 @@ def fetch_nearby_place_dcids(place: Place, locale=DEFAULT_LOCALE) -> List[str]:
...
+  parents_to_highlight = get_ordered_parents_to_highlight(parent_places)
+
+  peers_within_parent = []
+  if (parents_to_highlight):
+    peers_within_parent = fetch_child_place_dcids(
+        parents_to_highlight[0], place_type_to_highlight(place_types))
+    random.shuffle(peers_within_parent)
+
+  # Remove place_dcid from the list
+  if place_dcid in peers_within_parent:
```
dwnoble, line 756: Does removing the current place from this list mean we could show a bar chart that has data for peers but not for the current place?

gmechali: no - we force it in the comparison places elsewhere. This is just to ensure we dont return the current place as a peer because that would be inaccurate. For now we're only using it in the bar charts, but down the road it may be an assumption we forget about. I think the correct approach is not to consider the current place as a peer.

But elsewhere we definitely validate that we only show the bar chart if we have data for the current place + that the current place must be in the list.

dwnoble: Got it- does that mean we only need to check for existence of data for at least one peer place instead of two peer places? Then we'd always be showing at least 2 bars. The logic as is i think ensures we have at least 3 bars in the bar chart.

server/routes/dev_place/utils.py
gmechali: Yeah so basically it's because our KG has a lot of data about parents but some don't make sense to use in that kind of representation.
So the string in the header saying "Arizona - State in USA, NorthAmerica" makes use of that for the parent places to highlight.
Or in this case, when we decide what are the peers, we take the parent to highlight, and find the child type to highlight.

The reason we need it is that some of the parent types don't make much sense to the user, certainly not in geographic terms. For example for France, one of the containedIn arcs is for UNGeoRegion WesternEurope, which would be pretty redundant to show
"France in WesternEurope, Europe". It's also in the OECD.

So because of that we define which types we want to highlight both going up and down.
Up is for the subheader string, and to find peers within that parent.
Down is to decide what peer type to choose.

Updated some comments.

server/routes/dev_place/utils.py
gmechali: I think this is working properly. Im not sure im following where you're seeing the discrepancy. could you elaborate?

For example place: USA
When we want to see if we have data for the peers, we call obs_point_within(NorthAmerica, Country, ...)
And we fetch_peer_places_within(USA, Country). That method in turn fetches the parent to highlight which will be NorthAmerica, and it will search for Countries within it.

I think maybe the confusion was from the name? fetch_peer_places_within should be named fetch_peer_places_within_parent - and then you pass in the current place, to fetch it's peers.

Do you agree about the logic being accurate? ( I can do the renaming)

server/routes/dev_place/utils.py
gmechali: So I did that intentionally tbh to only show peer comparisons when we have data for more than 1 peer, since the bar charts look more logical for comparisons when we have multiple charts to compare to (at least in the context of showing peer places). But I can revert it to just 1 other if you prefer.

server/routes/dev_place/utils.py
gmechali: Done.

server/routes/dev_place/utils.py
dwnoble: That makes sense- thank you!

server/routes/dev_place/utils.py
dwnoble: Having the main place + 2+ peers makes sense- i just wanted to make sure it was deliberate.

server/routes/dev_place/utils.py
dwnoble: I think im getting thrown off because the logic for getting the parent place looks slightly different.

For `parent_place_dcid`:

```
  # Get parent place DCID
  parent_place_dcid = place_utils.get_place_override(
      place_utils.get_parent_places(place_dcid))
```

Compared to `fetch_peer_places_within`: 
```
  parent_places = get_parent_places(place_dcid)

  # Only consider peers within the first parent of a type to highlight.
  # For example, from country/FRA, we want to show the peers within the Continent Europe.
  # as opposed to peers within the UNGeoRegion WesternEurope, or within the UNData region OECD.
  parents_to_highlight = get_ordered_parents_to_highlight(parent_places)

  # Then the parent place dcid is p arents_to_highlight[0]
```

Is it possible to unify these two flows for selecting parents?

And an aside: do we need to fetch `peer_places_within` in this check at all? would it be enough just to look at the `peer_places_obs_point_response` and see if it has the current place + at least two other places?

dwnoble (commented):
Thats gabe! lmk what you think about unifying the logic for getting parent_place_dcid - otherwise lgtm!

server/routes/dev_place/utils.py
gmechali: So on the aside first, yes I think we do need to check the peer places. Otherwise you can end up in a position, where we have data for 10 places in the parent, but none that are in the list passed in as the peers. So your chart wouldn't include any peer data. And it would basically render the existence check useless since we confirm that it has peer data, but not for the peers we're going to show. So I do think is a significant improvement to check it for the peer_places_within.

And regarding the two different paths you're highlighting. yeah you're definitely right they can be unified. I just made the method get_place_override call the one that orders all the parents, and return the first one. Should be more clear now! 

Also in the process, I just added the city in places to highlight since I was debugging on a city and it should have shown more :)

dwnoble: ahh that makes sense about not having dat for the sample of places we're showing. And the new get_place_override looks good! thank you

dwnoble (approved):

PR #4869, opened 2025-01-21
[Rollout Infra] Start reading the GCS feature flags and expose them in a global Var

gmechali:
Uploaded feature_flags.json in datcom-website-[ENV]-resources GCS buckets in format as:
`
[{
    "name": "autocomplete",
    "enabled": true
},
{
    "name": "dev_place_experiment",
    "enabled": false
},
{
    "name": "dev_place_ga",
    "enabled": false
}]
`
On server startup, it ingests this file, and exposes the feature status through api/features.
Also adds an api/features/{feature_name} endpoint to expose whether it's enabled.

Starts reading the feature flags for place autocomplete, place page experiment rollout and place page GA rollout.

The feature flags were uploaded as:
- autocomplete: Enabled for all environments
- dev place experiment: Enabled for autopush
- dev place GA rollout: Enabled for autopush

This implementation is as outlined in https://docs.google.com/document/d/19CqgKt6aIbFs5v2zhm3TagDfPJF-K_SpH-9VPP9FeRw/edit?resourcekey=0-qM2XvR7dYp4j0jXshva04g&tab=t.0

In a followup, I will add the fallback flag values, ingest them in the load_feature_flags method, and write the script to auto update the GCS buckets. See PR - https://github.com/datacommonsorg/website/pull/4871


server/lib/util.py
```diff
@@ -394,6 +396,47 @@ def get_nl_no_percapita_vars():
     return nopc_vars
 
 
+def get_feature_flag_bucket_name() -> str:
+  """Returns the bucket name containing the feature flags."""
+  env_for_bucket = os.environ.get('FLASK_ENV')
+  if env_for_bucket in ['local', 'integration_test', 'test', 'webdriver']:
+    env_for_bucket = 'autopush'
+  elif env_for_bucket == 'production':
```
dwnoble, line 404: We should also think about what to do for the 'staging' environment and various legacy custom data commons (e.g., 'feedingamerica', and others defined in server/app_env/*.py)

For staging, i think we should read the 'prod' bucket.
For the legacy custom dcs, we can either read the prod bucket (but that assumes the right read permissions are in place), or maybe we can just read a local feature flag file. Another option could be we keep the prod feature flag file checked into this repo, and for all other environments beyond what you have listed here, we just read that locally checked in version? wdyt?

server/lib/util.py
```diff
@@ -394,6 +396,47 @@ def get_nl_no_percapita_vars():
     return nopc_vars
 
 
+def get_feature_flag_bucket_name() -> str:
+  """Returns the bucket name containing the feature flags."""
+  env_for_bucket = os.environ.get('FLASK_ENV')
+  if env_for_bucket in ['local', 'integration_test', 'test', 'webdriver']:
```
dwnoble, line 402: What do you think about `local` reading the feature flag file from the local filesystem for testing? That said, it would also be nice to be able to read the autopush or production feature flag file in locally as well for testing.

server/lib/util.py
```diff
@@ -394,6 +396,47 @@ def get_nl_no_percapita_vars():
...
+  """Loads the feature flags into app config."""
+  storage_client = storage.Client()
+  bucket_name = get_feature_flag_bucket_name()
+  try:
+    bucket = storage_client.get_bucket(bucket_name)
+  except NotFound:
+    logging.error("Bucket not found: " + bucket_name)
+    return {}
+
+  blob = bucket.get_blob("feature_flags.json")
```
dwnoble, line 419: this might need a try/except block too

server/lib/util.py
```diff
@@ -394,6 +396,47 @@ def get_nl_no_percapita_vars():
...
+      data = json.loads(blob.download_as_bytes())
+    except json.JSONDecodeError:
+      logging.warning("Loading feature flags failed to decode JSON.")
+    except TypeError:
+      logging.warning("Loading feature flags encountered a TypeError.")
+  else:
+    logging.warning("Feature flag file not found in the bucket.")
+
+  # Create the dictionary using a dictionary comprehension
+  feature_flag_dict = {
```
dwnoble, line 432: Why not just store the feature flag configuration as a object/dictionary in json?

server/routes/place/html.py
```diff
@@ -302,6 +303,9 @@ def is_seo_experiment_enabled(place_dcid: str, category: str,
 def is_dev_place_experiment_enabled(place_dcid: str, locale: str,
                                     request_args: MultiDict[str, str]) -> bool:
   """Determine if dev place experiment should be enabled for the page"""
+  if not is_feature_enabled("dev_place_experiment"):
```
dwnoble, line 306: Nit: define these strings as constants

server/lib/shared.py
```diff
@@ -243,3 +245,9 @@ def is_percapita_relevant(sv_dcid: str, nopc_svs: Set[str]) -> bool:
     if skip_phrase in sv_dcid:
       return False
   return True
+
+
+def is_feature_enabled(feature_name: str) -> bool:
+  """Returns whether the feature with `feature_name` is enabled."""
+  feature_flags = current_app.config['FEATURE_FLAGS']
+  return feature_name in feature_flags and feature_flags[feature_name]
```
dwnoble, line 253: Alternative: 
`return feature_flags.get(feature_name, False)`

static/js/components/nl_search_bar/nl_search_bar_header_inline.tsx
```diff
@@ -32,16 +33,22 @@ const NlSearchBarHeaderInline = ({
...
   const urlParams = new URLSearchParams(window.location.search);
   const isAutopushEnv = window.location.hostname == "autopush.datacommons.org";
-  const enableAutoComplete =
-    isAutopushEnv ||
-    (urlParams.has("ac_on") && urlParams.get("ac_on") == "true");
+
+  useEffect(() => {
+    setAutoCompleteEnabled(
+      isFeatureEnabled("autocomplete") ||
+        isAutopushEnv ||
```
dwnoble, line 43: is this necessary if the feature flag is turned on in the autopush config?

server/lib/util.py
gmechali: I would argue that staging should still read its own bucket, but that we should aim to keep the staging flags and prod flags in sync. They should only differ if we need to test something in staging before launching it to prod. So temporary diffs. 
Do you agree?

For custom dc, I agree with your take that it should read the local file. If custom DC customers want to support launches on the fly, they can make their own bucket, store their flag and update the target. I ll leave that conversation for the following PR where I actually add the local files!

server/lib/util.py
gmechali: Totally agree. Lets leave that for the follow up PR which adds the local file!

server/lib/util.py
gmechali: So I did load it up in json but doing this just plays it a little safer since we now know that if we have a feature flag, it must have a name and enabled.
so in case the file that was read in had bad data this would ensure it doesn't get added to the flags.

static/js/components/nl_search_bar/nl_search_bar_header_inline.tsx
gmechali: Definitely something to cleanup. I just didnt want to regress anything (and in this case I mean disable autocomplete in autopush), if this doesnt end up working entirely as we expected.

server/lib/util.py
gmechali: so actually if you look at the method implementation, it already nests the read call within a try/catch and returns None if there's a failure. 
So this should never raise any exception.

server/routes/place/html.py
gmechali: Done. I moved this method into a feature_flags.py file and defined all the flags there.

gmechali:
> Thanks Gabe!
> 
> Last thing (can be in a follow up)- can you add some instructions, either to the readme or inline in one of the feature_flags files on how to enable feature flags and how to update the feature flag files (including what gcp account they're in)

Yeah absolutely, i ll do that after im done implementing. And will clean up the doc too!

PR #4870, opened 2025-01-21
[Place Page Revamp] Stop respecting the is_overview attribute from place page configs for continents since we only have places_within data.

gmechali:
Local: https://screenshot.googleplex.com/9ok8mDJPfGZYXkj

This PR stops respecting the is_overview attribute from place page configs for continents because we do not have continent level data. We decided to only show place data, or peer places data from the place page overview. 
For continents, we will show all blocks available.


server/webdriver/base_utils.py
```diff
@@ -30,7 +30,7 @@
 def create_driver(preferences=None):
   # These options are needed to run ChromeDriver inside a Docker without a UI.
   chrome_options = Options()
-  chrome_options.add_argument('--headless=new')
+  # chrome_options.add_argument('--headless=new')
```
dwnoble, line 33: What's this for?

dwnoble (commented):
I could see this problem of not having "is_overview" data, but having other data in the category available popping up for other places as well. Is there a good way to generalize it? Maybe we can check for "is_overview" data, and if nothing comes back we just show non-is_overview data in the overview page?

gmechali:
> I could see this problem of not having "is_overview" data, but having other data data in the category available popping up for other places as well. Is there a good way to generalize it? Maybe we can check for "is_overview" data, and if nothing comes back we just show non-is_overview data in the overview page?

Yeah good point - Fixed it. It was pretty simple to just fall back to the full chart if we dont have any data in the overview.

server/webdriver/base_utils.py
gmechali: oops undone.

server/routes/dev_place/utils.py
```diff
@@ -194,18 +194,26 @@ def get_place_override(places: List[Place]) -> str:
 def filter_chart_configs_for_category(
     place_category: str, chart_config: List[ServerChartConfiguration]
 ) -> List[ServerChartConfiguration]:
-  """Only returns the appropriate"""
+  """
+  Only returns the appropriate charts for the current category. Note that we do not
+  respect the is_overview filter for continents since we do not have continent level data.
```
dwnoble, line 199: Update this comment

PR #4871, opened 2025-01-21
[Rollout Infra] Adds the fallback flags, creates the script to upload to GCS and modifies the loading logic.

gmechali:
This PR modifies the logic for reading the feature flags such that it falls back to local files if for some reason the GCS bucket file is not found, if there's a permission error, etc. 

We also modify the logic on the 3 existing flags (autocomplete, Place page experiment, Place page GA) to represent what is needed.

We create the local files to be used as fallback, and from which the GCS files will get updated. Note that we enabled Autocomplete on all environment (for the next release). But kept Place page experiment and GA disabled for now.

Creates the script to update the GCS files from the local files.
For production, we also diff the file in the staging GCS bucket in order to ensure it has been verified.

Lastly the script is able to trigger a Kubernetes restart.

This has all been tested end to end.


scripts/update_gcs_feature_flags.sh
```diff
@@ -0,0 +1,108 @@
+#!/bin/bash
+# Copyright 2024 Google LLC
```
dwnoble, line 2: 2025- happy new year!

scripts/update_gcs_feature_flags.sh
```diff
@@ -0,0 +1,108 @@
...
+if [[ ! " ${valid_environments[@]} " =~ " ${environment} " ]]; then
+  echo "Error: Invalid environment '$environment'."
+  exit 1
+fi
+
+# Construct the filename (e.g., dev.json, staging.json)
+file="${environment}.json"
+
+# Validate the JSON file
+if ! python -m json.tool "server/config/feature_flag_configs/${file}" &> /dev/null; then
```
dwnoble, line 54: This fails for me because i only have python3 and not python in my path

```
./scripts/update_gcs_feature_flags.sh: line 54: python: command not found

```

server/lib/util.py
```diff
@@ -424,6 +425,32 @@ def load_feature_flags():
   else:
     logging.warning("Feature flag file not found in the bucket.")
 
+  return data
+
+
+def load_fallback_feature_flags(environment: str):
```
dwnoble, line 431: This fails for me in the `custom` or other custom dc environments like `unsdg_staging`

```
FileNotFoundError: [Errno 2] No such file or directory: 'website/server/config/feature_flag_configs/unsdg_staging.json'
```

dwnoble: Maybe: if a feature flag environment file isnt found, we can just return the production file?

gmechali: Good point - this should be addressed now. Skipping straight to local files and defaulting to autopush for anything we dont know.

gmechali: oh wait I should definitely doproduction. ok done.

scripts/update_gcs_feature_flags.sh
gmechali: Fair - fixed.

scripts/update_gcs_feature_flags.sh
gmechali: done.

dwnoble (approved):

PR #4876, opened 2025-01-22
2025-01-22 Custom DC stable release

hqpho:
# Highlights
- Bug fix for data imports failing if the GCS input path was not initialized with an empty blob
- Bug fix for custom variables not showing up in custom Stat Var Explorer
- Bug fix for services container not starting up if embeddings file is empty
- Improvements to data load log format
- Services container no longer creates SQL tables if they don't exist (the data management container should be used to create the tables instead)
- Web admin removed (was deprecated after data management container launch)


keyurva (approved):
nit in the description: 

> Bug fix for data imports ~failed~ *failing*(?) if the GCS input path was not initialized with an empty blob

PR #4879, opened 2025-01-23
[Place Page Revamp] Removes Overview from topics, Adds link to overview from Title, Adds world to parent places to highlight.

gmechali:
* Adds world to list of places to highlight.
* Makes the place name in the title a link back to the overview page
* Removes the overview from the topics
* Makes the tests check the ordering of the topics.

### DIFF
--- a/static/js/place/dev_place_main.tsx
+++ b/static/js/place/dev_place_main.tsx
@@ -58,7 +58,9 @@ const PlaceHeader = (props: {
       <div className="place-info">
         <h1>
           <span>
-            {place.name}
+            <a className="place-info-link" href={`/place/${place.dcid}`}>
+              {place.name}
+            </a>
             {category != "Overview" ? ` • ${category}` : ""}{" "}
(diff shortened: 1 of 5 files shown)


static/js/place/dev_place_main.tsx
```diff
@@ -58,7 +58,9 @@ const PlaceHeader = (props: {
       <div className="place-info">
         <h1>
           <span>
-            {place.name}
+            <a className="place-info-link" href={`/place/${place.dcid}`}>
+              {place.name}
+            </a>
             {category != "Overview" ? ` • ${category}` : ""}{" "}
```
dwnoble, line 64: Do we need this line anymore?

dwnoble (approved):

static/js/place/dev_place_main.tsx
gmechali: Discussed offline. all good.

PR #4883, opened 2025-01-24
[Place Page Revamp] Remove the chart configs when the blocks are empty.

gmechali:
Fixes guernsey topics: https://screenshot.googleplex.com/4M3tcdjeMurDjga
Also allows Zip codes to be used as children of cities, see LA: https://screenshot.googleplex.com/5ZTyvSuwfPuaEUD


dwnoble (approved):
Nice addition with the zip codes

PR #4875, opened 2025-01-22
Update place page summaries

juliawu:
(1) Updates the data values used in place page summaries to more recent data.
(2) Cleans up the summary generation code, adding a new utility function.

PR #4884, opened 2025-01-24
Updated place page key demographics table display and fetch path

dwnoble:
- Updated key demographics variable names to match figma mocks
- Added localization support to display values, units, and dates
- Moved overview table data fetch path from the client side to the server side (new endpoint: `/api/dev-place/overview-table/<place-dcid>` for a performance boost.
  - @juliawu - this also changes the data fetching strategy a bit. Can you take a look and see if my comment from https://github.com/datacommonsorg/website/pull/4875  still applies?

Before:
overview_before

After: 
overview_after


packages/client/src/data_commons_web_client.ts
```diff
@@ -248,6 +249,20 @@ class DataCommonsWebClient {
     const response = await fetch(url);
     return (await response.json()) as RelatedPlacesApiResponse;
   }
+
+  async getPlaceOverviewTable(params: {
```
gmechali, line 253: nit: add doc

packages/client/src/data_commons_web_client_types.ts
```diff
@@ -201,3 +201,19 @@ export interface RelatedPlacesApiResponse {
   parentPlaces: Place[];
   peersWithinParent: string[];
 }
+
+export interface OverviewTableDataRow {
+  date: string;
```
gmechali, line 206: Do we always get the data from the same date?

server/routes/dev_place/api.py
```diff
@@ -186,3 +187,17 @@ def related_places(place_dcid: str):
...
+
+
+@bp.route('/overview-table/<path:place_dcid>')
+@log_execution_time
+@cache.cached(timeout=TIMEOUT, query_string=True)
+def overview_table(place_dcid: str):
+  """
+  Fetches and returns overview table data for the specified place.
+  """
+  # Fetch overview table data
```
gmechali, line 199: nit: this comment seems unnecessary given the function comment right above

server/routes/dev_place/types.py
```diff
@@ -118,3 +118,24 @@ class ServerChartConfiguration:
...
+
+@dataclass
+class OverviewTableDataRow:
+  """
+  A single row of overview table data for a place.
+  """
+  date: str
+  name: str
+  provenanceUrl: str
+  unit: str
```
gmechali, line 131: should this be optional str?

server/routes/dev_place/utils.py
```diff
@@ -62,6 +63,27 @@
 PLACE_TYPE_IN_PARENT_PLACES_STR = '%(placeType)s in %(parentPlaces)s'
 NEIGHBORING_PLACES_IN_PARENT_PLACE_STR = 'Neighboring %(placeType)s in %(parentPlace)s'
 
+# Variables to include in overview table
+PLACE_OVERVIEW_TABLE_VARIABLES = [
```
gmechali, line 67: Can be left for a follow up - but I think this would be nice to have:
1. in a different file that does everything for messages.
2. in a typed list

static/js/place/dev_place_main.tsx
```diff
@@ -380,13 +391,16 @@ export const DevPlaceMain = (): React.JSX.Element => {
...
         />
-        {isOverview && categories.length > 0 && placeSummary != "" && (
-          <PlaceOverview
-            place={place}
-            placeSummary={placeSummary}
-            parentPlaces={parentPlaces}
-          />
-        )}
+        {isOverview &&
+          placeOverviewTableApiResponse &&
```
gmechali, line 395: Instead of conditioning the visibility of the PlaceOVerview component on the response coming back, should we move that check to inside PlaceOverview and just condition visibility of the table on it?

server/i18n/es/LC_MESSAGES/all.po
```diff
@@ -749,3 +749,15 @@ msgstr "Pueblo"
 
 msgid "singular_zip_code"
 msgstr "Código postal"
+
+msgid "VARIABLE_NAME-Count_Person"
```
gmechali, line 753: For all these strings were those updated manually? I only updated the english one in my previous PR.

packages/client/src/data_commons_web_client.ts
dwnoble: done

packages/client/src/data_commons_web_client_types.ts
dwnoble: Not necessarily - we get the latest observation from each variable which could be different:
Screenshot 2025-01-24 at 1 09 29 PM

server/routes/dev_place/api.py
dwnoble: Removed

server/routes/dev_place/types.py
dwnoble: Yes-added

server/routes/dev_place/utils.py
dwnoble: Agreed about adding better typings and moving messages to another file. I added some basic typings for now, but we can organize the messages better in a follow-up

server/i18n/es/LC_MESSAGES/all.po
dwnoble: I did update them manually and i think you're right i should have only done the english one. Reverting the others

dwnoble: After some offline discussion, ill keep the english strings in these files for now as they show better fallback text

gmechali (approved):

PR #4887, opened 2025-01-24
topic_cache.py improvements 

megaz:
Error related to `memberList` not in node but not specifying the node dcid which makes troubleshooting challenging. 

Screenshot 2025-01-24 at 9 13 48 AM

`topic_cache._load_nodes` assumes that nodes contain either relevantVariableList or memberList.  

- Enhanced validation check with additional logging
- Added a unit test for missing unit test coverage.


keyurva (approved):
Thanks for the fix and the tests, Zahir! Much appreciated.

dwnoble:
Thanks @megaz ! can you run `./run_test.sh -f` on this PR to fix the lint messages

PR #4892, opened 2025-01-24
[custom dc] Added resource to run the schema init job on terraform apply operations before deploying the web service

dwnoble:
Requires "gcr.io/datcom-ci/datacommons-data:latest" to run since the input bucket won't have a config file yet.

I'll switch it back to gcr.io/datcom-ci/datacommons-data:stable after the next release.

### DIFF
--- a/deploy/terraform-custom-datacommons/modules/main.tf
+++ b/deploy/terraform-custom-datacommons/modules/main.tf
@@ -413,12 +413,8 @@ resource "null_resource" "run_db_init" {
       gcloud run jobs execute ${var.namespace}-datacommons-data-job \
         --update-env-vars DATA_RUN_MODE=schemaupdate \
         --region=${var.region} \
-        --project=${var.project_id}
-
-      # 2) Wait for the job to complete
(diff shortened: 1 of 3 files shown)


hqpho (approved):
Nice, love that this makes schema update a no-brainer as well!

Thanks for keeping this terraform config maintained! Would love to discuss what we can do by way of integration testing before stable release so you're not having to babysit it and fast-follow fixes.

dwnoble:
> Nice, love that this makes schema update a no-brainer as well!
> 
> Thanks for keeping this terraform config maintained! Would love to discuss what we can do by way of integration testing before stable release so you're not having to babysit it and fast-follow fixes.

Not a problem! Maybe we should migrate the custom DC autopush releases to use terraform?

keyurva (approved):
Thanks Dan!

With this change, if users run into schema errors, could we just tell them "Run the terraform"?

dwnoble:
> Thanks Dan!
> 
> With this change, if users run into schema errors, could we just tell them "Run the terraform"?

Yes that is right!

PR #4894, opened 2025-01-26
Added geo existence check to place page chart filtering logic

dwnoble:
* Adds geo data existence check for child places & peer places map
* Sequentially checks geoJsonCoordinatesDP3 (lowest-resolution) and falls back to geoJsonCoordinatesDP2 if no geoJsonCoordinatesDP3 is available.
* This check adds a good amount of latency to the chart filtering process (the response time is around ~4s total for California)

Before
before-ggy

After
after-ggy


gmechali:
LGTM -  but do you think it's worth the additional latency that you call out? 
Perhaps something we could try is to hide the geo existence check behind a feature flag so that if we start seeing some other metrics drop (and we fear it could be latency related), we can turn it off easily?

gmechali (approved):

PR #4896, opened 2025-01-27
Change sample questions

gmechali:
Which country has the most old people? --> Doesn't show very good data to highlight since we mainly only have US data, until the 5th chart which is life expectancy and that sv is highlighted in one of thehealth question. 
I changed that question to "Tell me about age distribution in the USA."

Change GDP highlight from iran to new zealand.

I also replaced the language highlight from Tamil to Spanish since in the US, the Tamil speaker count is so much lower that the data is a bit less significant. Max state was at 0.3% of population compared to multiple states with 20%+ Spanish speakers.
I couldnt get any data for Tamil speakers by indian state.

I also removed immunization in Pakistan since we don't have that data and I couldnt find any other country for which we had immunization rates. Perhaps that was from one of the datasets we had to remove for licensing?


dwnoble:
This LGTM, but @miss-o-soup can you take a look to see if these question changes sound good to you?

miss-o-soup:
Yes! - Thank you, yes for updating a the sample questions in the homepage to showcase data we do have ;-)

miss-o-soup:
It looks good, and thank you for taking care of this.
I had two small nit pics:


1) Can we transform one of the sample questions to end with a “?” This means:

FROM
Tell me about age distribution in the USA.
TO
What is the age distribution in the USA?

—

2) We wanted to make sure we have more global questions (not only US centric)
So here is an idea to update the questions:

FROM
"What is the prevalence of asthma across California counties?"
TO:
“How has life expectancy changed in the world?”

FROM:
 "What are the rates of child immunization in Pakistan?",
TO:
“Have mortality rates decreased over the years in India?”

miss-o-soup (approved):
It looks good, and thank you for taking care of this.
I had two small nit pics:

Can we transform one of the sample questions to end with a “?” This means:
FROM
Tell me about age distribution in the USA.
TO
What is the age distribution in the USA?

—

We wanted to make sure we have more global questions (not only US centric)
So here is an idea to update the questions:
FROM
"What is the prevalence of asthma across California counties?"
TO:
“How has life expectancy changed in the world?”

FROM:
"What are the rates of child immunization in Pakistan?",
TO:
“Have mortality rates decreased over the years in India?”

gmechali:
> It looks good, and thank you for taking care of this. I had two small nit pics:
> 
> 1. Can we transform one of the sample questions to end with a “?” This means:
> 
> FROM Tell me about age distribution in the USA. TO What is the age distribution in the USA?
> 
> —
> 
> 2. We wanted to make sure we have more global questions (not only US centric)
>    So here is an idea to update the questions:
> 
> FROM "What is the prevalence of asthma across California counties?" TO: “How has life expectancy changed in the world?”
> 
> FROM: "What are the rates of child immunization in Pakistan?", TO: “Have mortality rates decreased over the years in India?”

1. Done.

2. So I kept the california counties because we already have a question in Health about Life Expectancy in Africa. On the india mortality, added it.

PR #4897, opened 2025-01-27
Resolve circular import on test helper

gmechali:


chejennifer (commented):
I think base.py also uses the TIMEOUT from shared.py, should update here as well: https://github.com/datacommonsorg/website/blob/master/server/webdriver/base.py#L29

gmechali:
> I think base.py also uses the TIMEOUT from shared.py, should update here as well: https://github.com/datacommonsorg/website/blob/master/server/webdriver/base.py#L29

oh good find. Looked for others and there doesn't seem to be any more

chejennifer (approved):

PR #4898, opened 2025-01-27
[Place Page Revamp] Fix 2 bug bash bugs, adds dates to charts and removes bar charts for just one place.

gmechali:
Two fixes from place page bugbash:

- Adds the Date to the chart titles. https://screenshot.googleplex.com/6ymCjdtaW2PrAPN
- Refactors the way to compute and store which Stat vars we have data for such that we store how many places we have per stat var as 1, or 2. 2 implies 2+. And we only show bar charts when there are 2+ places.

The stat var existence check refactoring results in removing charts such as this: https://screenshot.googleplex.com/k3vzNSefFjpXeMo from the NorthAmerica page.
Or charts such as this: https://screenshot.googleplex.com/AGrK7cAnakg5h8b From the Asia page.


server/routes/dev_place/utils.py
```diff
@@ -244,6 +244,32 @@ def filter_chart_configs_for_category(
   return filtered_chart_config
 
 
+def count_places_per_stat_var(obs_point_response,
+                              stat_var_dcids: list[str],
+                              places_to_consider: list[str] = None):
```
dwnoble, line 249: Nit: add return type

static/js/place/util.ts
```diff
@@ -223,14 +225,19 @@ export function placeChartsApiResponsesToPageConfig(
...
+          if (!block_title) {
+            block_title = block.title;
+          }
+
           const title = block.title;
           const tileConfig: TileConfig = {
             /** Highlight charts use title as description */
             description: title,
-            title,
+            title: title + DATE_STR,
```
dwnoble: Looks like line charts don't support the ${date} replacement string: 

B5Ld3PGBeP5R8jB

server/routes/dev_place/utils.py
gmechali: done.

static/js/place/util.ts
gmechali: ah good catch, I was just looking at continent place pages. 

I just added the fix to support dates in the line charts.
https://screenshot.googleplex.com/65FPQWbmXAFBdvB

dwnoble: Looks like that line chart is showing 2022 instead of 2023

static/js/components/tiles/line_tile.tsx
```diff
@@ -237,12 +237,35 @@ function getPlaceDcids(props: LineTilePropType): string[] {
     : [props.place.dcid];
 }
 
+/**
+ * Returns the latest year found in the chart data.
+ *
+ * @param chartData Line chart data
+ * @returns Latest year with data.
+ */
+const getLatestDate = (chartData: LineChartData): string | null => {
```
dwnoble, line 246: This logic is slightly different from the logic used in the other tiles: https://github.com/datacommonsorg/website/blob/9d25f5a6dd468e667a6719497a9e28b47adcedf7/static/js/utils/string_utils.ts#L94

the `formatDate` call helps formats dates like `2024-11` as `Nov, 2024`

To be honest i think we can just drop dates from line charts entirely for now

static/js/place/util.ts
gmechali: I hadn't updated the screenshot but I had fixed that :)
https://screenshot.googleplex.com/9irPUUrr4cqYQ9S

static/js/components/tiles/line_tile.tsx
gmechali: So I think what I have now is a bit simpler to follow. I manager to use the time, not label which is definitely better.
Lmk if you just want me to remove it? But I'd say it seems to be working well. Im always seeing the same year and highlight tile.

dwnoble: Got it- keeping what you have sounds good. Can you add a todo to unify the way the line chart & other charts do date formatting in the title though down the road?

dwnoble (approved):

static/js/components/tiles/line_tile.tsx
gmechali: done.

PR #4899, opened 2025-01-28
Chart config fixes.

dwnoble:
* Temporary fix: Fixed issue where country/TLS was showing low-resolution geometries. Added an override to always show `geoJsonCoordinateDP1` level detail. 
* Temp fix: disabled per capita option on "Housing units by date built" because its denominator is the number of housing units built
* Replaced "Fertility Rate" child places bar chart with map
* Replaced "Gini Index" neighboring places map with bar chart
* Removed redundant per-capita option on "Electricity generation per capita"
* Removed redundant per capita on "Life expectancy (years)"

Before
before

After
after


server/config/chart_config/demographics_new.json
```diff
@@ -71,10 +67,12 @@
       },
       {
         "place_scope": "CHILD_PLACES",
-        "is_overview": true,
```
gmechali, line 74: Why this change? Fertility rate is on the old overview page.

server/config/chart_config/demographics_new.json
```diff
@@ -10,9 +10,7 @@
       "Count_Person_NotAUSCitizen",
       "Count_Person_USCitizenBornAbroadOfAmericanParents"
     ],
-    "denominator": [
-      "Count_Person"
```
gmechali, line 14: just FYI - if we leave denominator as [], and non_dividable = false, then it ll default denominator to be count_person.
I still left it explicit in the charts, but just letting you know

server/routes/shared_api/choropleth.py
```diff
@@ -83,6 +83,13 @@
 MULTIPOLYGON_GEOJSON_TYPE = "MultiPolygon"
 POLYGON_GEOJSON_TYPE = "Polygon"
 
+# Override the choropleth display level map for special cases
+# TODO: Remove this once we have a better way to handle special cases
```
gmechali, line 87: why are these special cases? what makes them so?

server/config/chart_config/demographics_new.json
dwnoble: Got it - in this case denominator wasn't actualyl removed, it looks like my editor ran a code formatter

server/config/chart_config/demographics_new.json
dwnoble: Whoops good catch. adding that back

server/routes/shared_api/choropleth.py
dwnoble: Updated the comment to clarify

PR #4904, opened 2025-01-28
[Place Page Revamp] Add the more charts button only when there are more charts.

gmechali:
We already do the existence check on every category and on the overview page so this PR just adds some logic to count the number of blocks for each category in the Overview charts, and in the per category charts.

We also set a boolean field on the Category object to determine if the category `hasMoreCharts` and only if so, we will show the more charts page.

For example, for East Timor: 
* We need moreCharts for the Economics page https://screenshot.googleplex.com/6hAbDTjwAXtMTun
* But not the energy page: https://screenshot.googleplex.com/7qc44kS8m3QSUVv


server/routes/dev_place/utils.py
```diff
@@ -785,6 +786,38 @@ def translate_chart_config(
   return translated_chart_config
 
 
+def get_categories_with_more_charts(
+    category: str, categories: List[Category],
+    existing_chart_config: List[ServerChartConfiguration],
+    existing_chart_config_for_category: List[ServerChartConfiguration]
+) -> List[Category]:
+  """Returns"""
```
dwnoble, line 794: update comment

dwnoble (approved):

server/routes/dev_place/utils.py
gmechali: done.

PR #4910, opened 2025-01-29
Improved place page geo existence check to only check if a place has geo data, and not to actually fetch the geo data

dwnoble:
Previously, we performed the geo check by fetching the geojson value: `http://api.datacommons.org/v2/node?key=<key>&nodes=country/CAN&property=->geoJsonCoordinatesDP1`
Now we just check if the geoJsonCoordinatesDP* property exists without fetching it : `http://api.datacommons.org/v2/node?key=<key>&nodes=country/CAN&property=->`


server/routes/dev_place/utils.py
```diff
@@ -451,20 +451,19 @@ def check_geo_data_exists(place_dcid: str, child_place_type: str) -> bool:
...
-    }
-    if bool(filtered_child_place_geojson_data):
-      return True
+  # Fetch properties for each child place.
+  place_properties = dc.v2node(child_places, "->").get("data", {})
+  # Check if geo data exists for at least one of the child places.
+  for _place_dcid in place_properties.keys():
+    properties = place_properties[_place_dcid].get("properties", [])
+    for property in properties:
+      if property in geo_properties:
```
gmechali, line 465: just checking. - it's guarantyeed to only have the property if there are values? Never an empty list as the value?

dwnoble: I checked with @n-h-diaz , and it looks like we can count on the field value to always be present if the property is listed here

PR #4912, opened 2025-01-30
[Place Page Revamp] Refactor place page code & use asyncio calls to parallelize expensive operations

gmechali:
With some of the recent place page work, the latency on the api calls has reached levels above what we should. This PR adds parallelization to our expensive operations with the related_places, and chats endpoints for the revamped place page.

This also refactors the code significantly to reuse code across different parts of the functions, and create more granular helper functions.
In a follow up PR I will add thorough testing on all these helpers and the main api endpoints. 

See Latency with this PR applied: https://paste.googleplex.com/4715161129320448
See latency without this PR: https://paste.googleplex.com/5451439923789824
Essentially, this leads to an approximately 40-50% speed up in the charts endpoint, and 30-40% for the related places endpoint.


server/requirements.txt
```diff
@@ -6,6 +6,7 @@ Flask-Babel==2.0.0
 Flask-Caching==2.0.1
 flask_cors==5.0.0
 flask_testing==0.8.1
+Flask[async]
```
dwnoble, line 9: I think with Flask[async] you can remove the `Flask==2.3.2` line. Can you also pin `Flask[async]` to version 2.3.2?

dwnoble (approved):

server/requirements.txt
gmechali: ah I see, done.

PR #4919, opened 2025-01-31
Fix flaky embeddings tests

chejennifer:
Through testing, it looks like the problem with the flaky embeddings tests arises when running the tests in parallel & seems to have to do with reading the downloaded files in parallel. Removed the flag that runs the tests in parallel and have tested and seen that this doesn't impact the timing of the tests.


tools/nl/download_nl_files/download_nl_files.py
```diff
@@ -64,6 +64,8 @@ def main(_):
       gcs.maybe_download(index_info.embeddings_path,
                          cache_root,
                          use_anonymous_client=True)
+    path = index_info.embeddings_path
+    print(f'finished downloading {path}')
```
clincoln8, line 68: Do we still need this print statement since we have [this log](https://github.com/datacommonsorg/website/blob/c774b61835bc018856b5ee7ebf3b5f6b425c68f8/shared/lib/gcs.py#L50)?

clincoln8 (approved):
Thanks Jennifer!

tools/nl/download_nl_files/download_nl_files.py
chejennifer: you're right, updated

PR #4922, opened 2025-02-03
[Place Page Revamp] Fix one chart and add API Tests

gmechali:
Fixes one chart metadata to remove scaling on units that don't need it.

Adds significant test coverage across our utils for the new place page. Creates helpers for mocking data. More testing to come.


server/routes/dev_place/utils.py
```diff
@@ -442,6 +443,17 @@ async def fetch_and_process_stats():
   return valid_chart_configs
 
 
+@cache.memoize(timeout=TIMEOUT)
+async def memoized_filter_chart_config_for_data_existence(
```
dwnoble, line 447: Is this used anywhere?

gmechali: ah that was my bad. this is supposed to be called from the API since we still want the memoization to happen. I moved this into a separate method because the cache.memoize annotation was causing some issues with the tests starting up since we don't have the cache initialization in this. I tried messing with it but couldn't figure it out. I can try again, but having the memoized method, and the implemetnation separate worked fine. just a little code duplication.

But the function there is now pretty well tested.

PR #4923, opened 2025-02-03
Enabled Cross-Origin Resource Sharing (CORS) for Custom DC mixer api endpoint (/core/*)

dwnoble:
This change enables cross-origin requests for the Custom DC mixer api endpoint. This allows customers like the UN to make api requests from their static site. 
(The website api endpoint already has CORS enabled)

### DIFF
--- a/build/cdc_services/nginx.conf
+++ b/build/cdc_services/nginx.conf
@@ -12,6 +12,18 @@ http {
 
     # Make mixer api available at /core/api/
     location /core/api/ {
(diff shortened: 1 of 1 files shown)


keyurva (approved):

hqpho:
As discussed offline, would prefer an allowlist mechanism. But defer to yours and Keyur's judgment on whether the risk is high enough to warrant the extra work.

hqpho (approved):

PR #4938, opened 2025-02-07
Added locale parameter & language loading to web components

dwnoble:


gmechali (approved):
Any tests worth adding here?

PR #4942, opened 2025-02-10
DataGemma UI cleanup

chejennifer:
cleaning up the DataGemma UI a little according to asks from adriana

image

### DIFF
--- a/server/templates/dev/datagemma.html
+++ b/server/templates/dev/datagemma.html
@@ -22,6 +22,17 @@
     
     {% block head %}
     <link rel="stylesheet" href="https://fonts.googleapis.com/icon?family=Material+Icons">
+    <style>
+          #main {
+               height: 100vh;
+               display: flex;
+               flex-direction: column;
+          }
+          .main-content {
+               width: 100%;
+               flex-grow: 1;
+          }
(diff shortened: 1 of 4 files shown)


server/templates/dev/datagemma.html
```diff
@@ -22,6 +22,17 @@
     
     {% block head %}
     <link rel="stylesheet" href="https://fonts.googleapis.com/icon?family=Material+Icons">
+    <style>
+          #main {
```
dwnoble, line 26: Could you move this block to an scss file?

chejennifer: in the previous PR you had asked me to use emotion instead of adding a new scss file .. do we prefer an scss file in this case?

dwnoble: I think since the styles apply to the base.html file it's okay to have them in scss (unless it's possible to make equivalent changes in emotion against just the components in the data gemma components)

chejennifer: gotcha, updated

dwnoble (approved):

chejennifer:
thanks for the review!

PR #4943, opened 2025-02-10
[Place Page Revamp]Defaults to 8 bar chart items on mobile width, and 15 on desktop

gmechali:
Minor bug bashes before the full launch
Reduce the default bar chart places to 8 for mobile. https://screenshot.googleplex.com/UZqSFHZTJRtHsS2

Creates a tooltip component to hold the AI summary tooltip.
https://screenshot.googleplex.com/SKasKh8w2u5DhzB


static/js/place/dev_place_overview.tsx
```diff
@@ -184,6 +185,10 @@ export const PlaceOverview = (props: {
       >
         <LocationCity />
         <span>{intl.formatMessage(pageMessages.SummaryOverview)}</span>
+        <InfoTooltipComponent
+          iconPath="../../images/info_spark.svg"
```
dwnoble, line 189: The icon is showing up a little bit pixelated looking for me- are you seeing that? is there a way to make it sharper?

dwnoble: Also- can you migrate this so the svg is wrapped directly in a component like this: https://github.com/datacommonsorg/website/blob/master/static/js/components/elements/icons/arrow_back.tsx

and could you add a note in the component where the icon came from

static/js/place/util.ts
```diff
@@ -36,6 +36,11 @@ import {
   TileConfig,
 } from "../types/subject_page_proto_types";
 
+const MAX_WIDTH_MOBILE = 768;
```
dwnoble, line 39: use `BREAKPOINTS.sm` from theme.ts here

static/js/place/util.ts
```diff
@@ -284,7 +296,9 @@ export function placeChartsApiResponsesToPageConfig(
               rankingCount: maxPlacesCount,
             };
           } else if (tileConfig.type === "BAR") {
-            maxPlacesCount = chart.maxPlaces ? chart.maxPlaces : 15;
+            maxPlacesCount = chart.maxPlaces
```
dwnoble, line 299: Looks like this works when i refresh the page, but the number of bars stays the same if i just resize the window

static/js/shared/components.tsx
```diff
@@ -14,7 +14,8 @@
  * limitations under the License.
  */
 
-import React from "react";
+import { useTheme } from "@emotion/react";
+import React, { useEffect, useRef, useState } from "react";
 import styled from "styled-components";
```
dwnoble, line 19: Update this to import styled from "@emotion/styled"

static/js/shared/components.tsx
```diff
@@ -42,3 +43,60 @@ export const IconPlaceholder = (props: {
...
+
+export const InfoTooltipComponent = (props: {
+  iconPath: string;
+  description: string;
+}): React.JSX.Element => {
+  const [isVisible, setIsVisible] = useState(false);
+  const containerRef = useRef<HTMLDivElement>(null);
+  const theme = useTheme();
+
+  const InfoTooltip = styled.div`
```
dwnoble, line 55: I generally define these helper components outside of the main component, and access the theme using:

```
const InfoTooltip = styled.div`
   ${p => p.theme.typography.text.sm}
   ...
`
```

But i realize this may be personal preference. What do you think?

static/js/theme/theme.ts
```diff
@@ -62,6 +62,7 @@ const DC_BLUE = "hsl(217, 90%, 43%)";
 const DC_BLUE_PILL_TEXT = "hsl(217, 90%, 15%)";
 const DC_BLUE_PILL_BCKG = "hsl(218, 92%, 95%)";
 const DC_BLUE_PILL_BCKG_HOVER = "hsl(204, 90%, 80%)";
+const DC_BLUE_BCKG_LIGHT = "hsl(217, 91%, 91%)";
```
dwnoble, line 65: Is this used anywhere?

static/js/theme/theme.ts
```diff
@@ -114,7 +115,7 @@ const theme: Theme = {
       secondary: {
         base: DC_BONE,
         dark: DC_GRAY,
-        light: DC_GRAY_LIGHT,
+        light: DC_BLUE_WHITE,
```
dwnoble, line 118: Was this change intentional? it will have some side effects in other pages

static/js/shared/components.tsx
```diff
@@ -42,3 +43,60 @@ export const IconPlaceholder = (props: {
...
+
+    document.addEventListener("mousedown", handleClickOutside);
+    return () => {
+      document.removeEventListener("mousedown", handleClickOutside);
+    };
+  }, [isVisible]);
+
+  return (
+    <InfoTooltipContainerStyled ref={containerRef}>
+      <img onClick={handleButtonClick} src={props.iconPath} />
```
dwnoble, line 98: The click behavior is a little awkward for me - can this be an onhover?

static/js/shared/components.tsx
```diff
@@ -42,3 +43,60 @@ export const IconPlaceholder = (props: {
...
+    };
+
+    document.addEventListener("mousedown", handleClickOutside);
+    return () => {
+      document.removeEventListener("mousedown", handleClickOutside);
+    };
+  }, [isVisible]);
+
+  return (
+    <InfoTooltipContainerStyled ref={containerRef}>
```
dwnoble, line 97: The tooltip is getting cut off for me on smaller screen sizes. Can you look into migrating this to use reactstrap's [Tooltip or UncontrolledTooltip](https://reactstrap.github.io/?path=/docs/components-tooltip--uncontrolled) class & overriding the styling?


Example in our app: https://github.com/datacommonsorg/website/blob/413be26e8e806ac8c40694aca4b4f70213610f90/static/js/components/subject_page/block.tsx#L296C1-L297C1

gmechali: okay so I took a serious stab at it but I think the customizations don't work as well. I couldn't get the tooltip to use our styling fully. Gemini also tells you if you want to customize it you should just use your own. I see your point on the small screen but there is not concept of hover on mobile, and afaict it has to be very small for it to get cut off right? 
I could also make it an even smaller box but it does look better with that width tbh.

static/js/shared/components.tsx
gmechali: Changed to hover.

static/js/theme/theme.ts
gmechali: nope not anymore removed.

static/js/theme/theme.ts
gmechali: DC_GRAY_LIGHT is very dark tbh and it seemed to me like the better secondary light background.
I couldn't find any instances that used that theme. Searching for background.secondary.light yields nothing, even when including the null safety operand. Am I missing some usages for it?

static/js/place/dev_place_overview.tsx
dwnoble: That SG - let's go with it

static/js/place/util.ts
dwnoble: That sounds reasonable

static/js/shared/components.tsx
dwnoble: Got it re:reactstrap.

For mobile, i think the accessible thing to do is get the tooltip to display on "touchstart" as well as on hover.

I think preventing the tooltip from getting cut off by the page is important. With that in mind and with the mobile touch vs hover interaction, if it's easier I'd be OK introducing a new 3rd party library like [react-tippy](https://tvkhoa.github.io/testlib/), [react-popper](https://github.com/floating-ui/react-popper), or [floating-ui](https://github.com/floating-ui/floating-ui). What do you think?

static/js/components/elements/icons/info_spark.tsx
```diff
@@ -0,0 +1,37 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/*
+ * Info Spark icon
+ */
+
+import React, { ReactElement } from "react";
+
```
dwnoble, line 22: Could you add a link to the figma where this element came from? (or @miss-o-soup if you have a better link can you paste it here)

static/js/place/dev_place_overview.tsx
```diff
@@ -159,6 +161,7 @@ export const PlaceOverview = (props: {
     parentPlaces.map((place) => place.dcid)
   );
   const theme = useTheme();
+  console.log("Themeeee" + JSON.stringify(theme));
```
dwnoble, line 164: remove extra logging

static/js/shared/components.tsx
```diff
@@ -42,3 +43,54 @@ export const IconPlaceholder = (props: {
...
+  margin: ${(p) => p.theme.spacing.xs}px;
+`;
+
+export const InfoTooltipComponent = (props: {
+  icon: React.JSX.Element;
+  description: string;
+  theme: Theme;
+}): React.JSX.Element => {
+  const [isVisible, setIsVisible] = useState(false);
+  console.log(props.theme?.colors?.background?.secondary?.light);
```
dwnoble, line 73: Remove extra logging

static/js/place/dev_place_overview.tsx
```diff
@@ -184,6 +187,11 @@ export const PlaceOverview = (props: {
       >
         <LocationCity />
         <span>{intl.formatMessage(pageMessages.SummaryOverview)}</span>
+        <InfoTooltipComponent
```
dwnoble, line 190: In the mocks im seeing this icon show up right after the text rather than in the header: https://screenshot.googleplex.com/4bFmZxmkx9qppGi

dwnoble: it looks like the icon is showing up even if there's no summary: http://localhost:8080/place/geoId/06?force_dev_places=true&hl=fr

static/js/shared/components.tsx
```diff
@@ -42,3 +43,54 @@ export const IconPlaceholder = (props: {
...
+  border-radius: ${(p) => p.theme.radius.secondary.borderRadius};
+  padding: ${(p) => p.theme.spacing?.md}px ${(p) => p.theme.spacing.lg}px;
+  z-index: 1;
+  box-shadow: ${(p) => p.theme.elevation.secondary.boxShadow};
+`;
+
+const InfoTooltipContainerStyled = styled.div<{ theme?: Theme }>`
+  position: relative;
+  display: inline-block;
+  margin: ${(p) => p.theme.spacing.xs}px;
```
dwnoble, line 64: The margin looks like it's making the whole line a bit bigger - would it make sense to remove the margin? The display & position styles may not be necessary as well

static/js/theme/theme.ts
dwnoble: Whoops you are right- i must have been looking at background.primary.light.

the change sgtm then!

static/js/components/elements/icons/info_spark.tsx
gmechali: Here is the figma link: https://www.figma.com/design/bK4jZkyB0FqfMhA1AUZVuu/2025-%E2%80%A2-Data-Commons?node-id=3808-980&t=eOQVjYoNHT0h1OSF-0

in offline convo with Adriana, we agreed that next to the title is preferred.

static/js/place/dev_place_overview.tsx
gmechali: Offline convo with Adriana to move it next to the header.

added the conditional to only show if there's a summary.

static/js/shared/components.tsx
gmechali: removed the margin, but the display and position are definitely needed to ensure the hover is in the right place.

static/js/shared/components.tsx
gmechali: https://screencast.googleplex.com/cast/NjQ1NzUyODAxODc5NjU0NHxhMDEyNDhiMy0xNA

Here's a test on mobile. I just used the CSS to get it to work. Let me know if you want some different changes to this.

static/js/i18n/i18n_place_messages.ts
```diff
@@ -158,6 +158,13 @@ export const pageMessages = defineMessages({
   SummaryOverview: {
     id: "summary_overview",
     defaultMessage: "Summary overview",
+    description:
+      "Tooltip text for the Summary overview section for the current place. This text explains that we use AI to generate the Summary Overview text.",
```
beets, line 162: i don't see AI in the in the actual message. the description should be about the defaultMessage, and a screenshot would be helpful for this (screenshot would have to be in g3)

ah, actually looks like the description for both got mixed up. could you update?

beets (approved):
lgtm - just one small update to make with string descriptions

static/js/i18n/i18n_place_messages.ts
gmechali: ohh yeah I messed it up, I shouldn't have modified this one, just added the new one below. Done!

PR #4945, opened 2025-02-10
Enable datagemma based on an env variable.

keyurva:
* The current approach configures this variable in code. Making it an env variable makes it configurable at runtime.
* Also, custom dcs [inherit](https://github.com/ONEcampaign/one-datacommons-website/blob/master/server/app_env/one.py#L29) from some of the base configs, so enabling it in base level configs like `local.Config` is problematic for custom dcs.
* Another approach could be to explicitly disable it for custom DCs. However, we want to slowly phase out python based configs over time so that they can be a pure runtime artifact. So implementing this as an env variable.

### DIFF
--- a/deploy/helm_charts/envs/autopush.yaml
+++ b/deploy/helm_charts/envs/autopush.yaml
@@ -29,6 +29,8 @@ website:
   nodePool: "default-pool"
   evalTool:
     enabled: true
+  dataGemma:
+    enabled: true
   redis:
     enabled: true
     configFile: |
--- a/server/__init__.py
+++ b/server/__init__.py
@@ -75,6 +75,13 @@ def _get_api_key(env_keys=[], gcp_project='', gcp_path=''):
   return ''
 
 
+def _enable_datagemma() -> bool:
+  """Returns whether to enable the DataGemma UI for this instance. 
+  This UI should only be enabled for internal instances.
+  """
+  return os.environ.get('ENABLE_DATAGEMMA') == 'true'
+
+
 def register_routes_base_dc(app):
   # apply the blueprints for all apps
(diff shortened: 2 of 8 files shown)


server/__init__.py
```diff
@@ -75,6 +75,13 @@ def _get_api_key(env_keys=[], gcp_project='', gcp_path=''):
   return ''
 
 
+def _enable_datagemma() -> bool:
+  """Returns whether to enable the DataGemma UI for this instance. 
+  This UI should only be enabled for internal instances.
+  """
+  return os.environ.get('ENABLE_DATAGEMMA') == 'true'
```
chejennifer, line 82: since we want this enabled locally and on autopush, I think we also need to:
- set the environment variable in run_server.sh script
- add the environment variable to here: https://github.com/datacommonsorg/website/blob/master/deploy/helm_charts/dc_website/templates/deployment.yaml#L129-L143
- set that environment variable for autopush: https://github.com/datacommonsorg/website/blob/master/deploy/helm_charts/envs/autopush.yaml

keyurva: Thanks. Done.

deploy/helm_charts/envs/autopush.yaml
```diff
@@ -29,6 +29,8 @@ website:
   nodePool: "default-pool"
   evalTool:
     enabled: true
+  dataGemma:
+    enabled: true
```
chejennifer, line 33: also set a default value for all other environments here: https://github.com/datacommonsorg/website/blob/master/deploy/helm_charts/dc_website/values.yaml#L42

keyurva: Done.

PR #4949, opened 2025-02-11
Upgrade transformers  to 4.48.0

rohitkumarbhagat:
Context:https://github.com/datacommonsorg/website/security/dependabot/201


chejennifer:
since this change didn't trigger the nl tests to run on cloudbuild, can you run them locally to make sure this update is ok?

PR #4951, opened 2025-02-12
Added inline search bar to /explore page in custom dc

dwnoble:
Before
Screenshot 2025-02-11 at 11 40 52 PM

After

Screenshot 2025-02-11 at 11 41 20 PM


server/templates/custom_dc/custom/base.html
```diff
@@ -23,13 +23,16 @@
 
   Optional variables:
   locale: string, value for html lang attr
+  is_hide_header_search_bar: boolean, if true, the header will not contain a search bar. Default false.
```
gmechali, line 26: How about using a feature for this? So you just have to add the flag to the config enable it everywhere except the custom dc file

keyurva (approved):
Thanks for the quick fix, Dan!

server/templates/custom_dc/custom/base.html
dwnoble: I'll push this out as is, but to gabe's point it might be useful having a separate menu of GA-ready features that custom DC users can pick from (like having the search bar in the header vs inline, or turning the place page on or off)

PR #4956, opened 2025-02-13
Feature Flags prevent duplicates and validate same count in all files.

gmechali:
Adds a unit test to validate the count of feature flags is the same in all files. For now lets enforce the same flag count across all files. 
Also ensure there are no duplicates in any file


clincoln8 (approved):
Thank you!!

PR #4960, opened 2025-02-17
[Place Page Revamp] Stop settings html links directly

gmechali:
This approach will more safely create the links in the header, we can get rid of one util test and instead just translate the parent places in the related API response.

Fix bug on the peer place type.


server/routes/dev_place/utils.py
```diff
@@ -96,15 +96,15 @@ def get_parent_places(dcid: str) -> List[Place]:
...
     if 'type' in parent and 'name' in parent and 'type' in parent:
-      all_parents.append(
-          Place(dcid=parent['dcid'],
-                name=parent['name'],
-                types=[parent['type']]))
+      all_parent_dcids.append(parent['dcid'])
 
-  return all_parents
+  all_parents = fetch_places(all_parent_dcids, locale)
+  parents_to_include = get_ordered_by_place_type_to_highlight(all_parents)
```
beets, line 105: since we already have type, should the list get ordered/filtered before fetching the localized names?

beets (approved):
i reviewed it even though it's still marked as a draft (iirc there's a setting to mark new pr's as draft  that can be bypassed).

server/routes/dev_place/utils.py
gmechali: Done.

PR #4964, opened 2025-02-19
[Feature Flags] Fix production rollout script.

gmechali:
Adds the second command for the us west1 cluster for production. 
Fixes doc to put the right command.


clincoln8 (approved):
Thanks for the fix!

PR #4969, opened 2025-02-20
Use api key helper during app creation.

clincoln8:
Migrate each instance of fetching api keys to use a single helper function.
Add a check to ensure that DC_API_KEY was provided as an environment variable when running in LITE mode.


server/__init__.py
```diff
@@ -73,9 +71,11 @@ def _get_api_key(env_keys=[], gcp_project='', gcp_path=''):
       secret_response = secret_client.access_secret_version(name=secret_name)
       return secret_response.payload.data.decode('UTF-8').replace('\n', '')
     except NotFound:
+      logging.warning(f'No key found at {gcp_project}:{gcp_path}')
       return ''
 
   # If key is not found, return an empty string
+  logging.warning(f'No key found for project: {gcp_project}; path:{gcp_path}')
```
chejennifer, line 78: this is not necessarily true because this is also a case of the key not found in the environment

server/__init__.py
```diff
@@ -73,9 +71,12 @@
       secret_response = secret_client.access_secret_version(name=secret_name)
       return secret_response.payload.data.decode('UTF-8').replace('\n', '')
     except NotFound:
+      logging.warning(f'No key found at {gcp_project}:{gcp_path}')
```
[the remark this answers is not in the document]
clincoln8: @chejennifer Is it okay to log this? Since SECRET_PROJECT is set in our code, it doesn't seem like it's sensitive information.

chejennifer (approved):
thanks for completing the todo!!

server/__init__.py
clincoln8: Updated the log to include the environment variables that were also checked.

server/__init__.py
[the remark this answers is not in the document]
chejennifer: I think it's ok to log, we can dismiss this alert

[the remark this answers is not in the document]
chejennifer: I feel like it's ok to log, but maybe instead we can just say 'no key found at {gcp_path} for this GCP project'

server/__init__.py
```diff
@@ -56,9 +56,7 @@ def _get_api_key(env_keys=[], gcp_project='', gcp_path=''):
 
   Returns:
       API key if it exists, otherwise an empty string.
-  
-  TODO: use this method everywhere else in this file
-  """
+    """
```
gmechali, line 59: ultra nit: might be nice to take this helper method into a helper file, to keep this a bit emptier. 
Obviously you're not introducing this so no pressure to make the change now :) but at least a TODO would be good!

gmechali (approved):
Wow using a helper there makes a big difference! THank you!

server/__init__.py
clincoln8: I see that send_email.py ([link](https://github.com/datacommonsorg/website/blob/4d28f8c8dfea466e65b1b2293119a4e17fb9427b/tools/send_email/send_email.py#L67)) and runner.py ([link](https://github.com/datacommonsorg/website/blob/4d28f8c8dfea466e65b1b2293119a4e17fb9427b/server/webdriver/screenshot/runner.py#L77)) reuse the same logic!

Do you have an idea for where this helper file should live and whether I should have send_email and runner use the same function? Thanks!

server/__init__.py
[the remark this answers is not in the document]
clincoln8: Done, thanks for the suggestion!

server/__init__.py
clincoln8: Chatted offline, will move to website/shared/lib/utils.py in follow up PR today!

PR #4972, opened 2025-02-21
Move get_api_key helper to shared lib and refactor direct calls to secretmanager.

clincoln8:
- Splits _get_api_key helper into get_gcp_secret and get_api_key. The latter first checks if api key is present as an environment variable and if not, then calls get_gcp_secret to fetch the key from GCP.

- Migrates the two other secretmanager accesses in the website repo to use the get_gcp_secret helper in the shared lib utils.


gmechali (approved):
LGTM - I assume you did run local to confirm everything still works, but just calling it out in case :)

gmechali:
oh I see, could you just try running the screenshot tester? It should be enough :)
https://github.com/datacommonsorg/website/blob/c774b61835bc018856b5ee7ebf3b5f6b425c68f8/run_website_screenshot.sh

clincoln8:
I'll close this for now since I'm still not clear on how to test the screenshot runner and there's not a huge gain from getting this submitted.

PR #4974, opened 2025-02-25
[Dynamic Placeholders] Modifies the Placeholder in the search bar to type sample questions

gmechali:
We are adding this feature in order to provide users inspiration for the type of questions they can ask. We use show the characters typed in order to create some movement on the page and help attract users' attention.

In order to get this to work, I had to make some modifications such that the placeholder gets stored in the QueryStore, so that we can safely override it in the input, then return to it after we have cycled through a few questions.

This PR only gets this feature to work on the homepage, new place page and old place page. The explore page always has a query set (not placeholder). We do NOT override a query in the search bar to show this. 

This also adds google analytics tracking, such that all future queries will show whether they were executed while this feature was on or off.

We added 2 feature flags for this, one for a GA rollout, and one for an experiment setup with the experimental rollout set at 20% of pages.

We reused the sample queries already displayed in the home page tiles.

Here is a screencast of how it works:
https://screencast.googleplex.com/cast/NTU4OTMxNDQzOTIxNzE1MnxkZGIzZDAwNC1kYg

Note we added a couple seconds delay before it starts, then we cycle through 5 questions from the sample questions. Then we stop and revert to the original placeholder.


static/js/i18n/i18n_place_messages.ts
```diff
@@ -204,4 +204,10 @@ export const pageMessages = defineMessages({
     description:
       'Used for the place page subheader. Shows the place type of the current place, and a list of all parent places. For example, for California, it shows "State in {USA, North America, World}".',
   },
+  categoryInPlace: {
```
gmechali, line 207: This is unrelated change but since i was doing a bunch of placeholder work and adriana asked for this, I just added it to do "Economics in India" when on the india economics page.

static/js/apps/base/components/header_bar/header_bar_search.tsx
```diff
@@ -59,6 +65,13 @@ const HeaderBarSearch = ({
   // Otherwise, use the query string from the query store.
   const initialValue = searchBarHashMode ? queryString : urlQuery;
 
+  // Initialize whether to let the placeholder show dynamic examples.
+  const EXPERIMENT_ROLLOUT_RATIO = 0.2;
```
dwnoble, line 69: Is the experiment rollout reset on page refresh? I think ideally we should partition users in a way that if the user refreshes the page the experiment is still either enabled or disabled. Maybe we can use their IP address?

static/js/components/nl_search_bar/dynamic_placeholder_helper.ts
```diff
@@ -0,0 +1,117 @@
...
+  const sampleQuestions = metadataContainer?.dataset?.sampleQuestions
+    ? JSON.parse(metadataContainer.dataset.sampleQuestions).flatMap(
+        (category) => category.questions
+      ) ?? []
+    : [];
+  return sampleQuestions;
+};
+
+/* Start typing the sample questions through the Input's placeholder attribute while inactive. */
+export const cycleSampleQuestions = (
```
dwnoble, line 75: Im seeing a slight delay when i load the page - the holder is "Enter a question to explore" for a second or two before switching to the "Try searching ..." placeholder -is that intentional?

gmechali: Yes! I put a 5 second delay before it starts. I did that because it looked kinda weird when I just land on a page and it immediately started doing that. I liked it more with the little delay.

static/js/apps/base/components/header_bar/header_bar_search.tsx
gmechali: Yes it does refresh on every page. 
totally agree it would be nice to keep it uniform across users, but Im pretty sure there were objections last time we talked about using IP to split up users for experiments.
And we dont have the framework for it right now, if that gets approved happy to move in that direction next.

However, I dont think it's worth delaying this for it? And when you think of what this feature is, it really doesn't impact users if they go to the next page and "lose" this feature. 

It's not like changing charts, or enabling a button for something.

static/css/core.scss
```diff
@@ -660,13 +665,19 @@ section {
...
         font-family: var(--font-family-sans-serif);
-        font-size: 16px;
         font-style: normal;
         font-weight: 400;
         line-height: 24px;
         white-space: nowrap;
         overflow: hidden;
         text-overflow: ellipsis;
+
+        @include media-breakpoint-down(lg) {
```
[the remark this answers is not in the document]
gmechali: https://screencast.googleplex.com/cast/NTA0Mjc4Njk5Nzg5NTE2OHxhNmI3MzE2NS1iYg

Obviously it just picked the longest questions this time lol. But it's not amazing given the text ends up being too long for the bar, that's why I moved the magnifier icon to the left, and reduced the font size (which looked too big on mobile anyways).
Lmk if you have some thoughts on how else to improve it. TBH I think we should put in effort to allow it to expand and stuff because even the explore page on mobile doesn't have a good experience for that...
https://screenshot.googleplex.com/Bbet2CsTeLqtJtm Would be good to wrap it - and I think we should treat that separate from this placeholder issue?

static/js/apps/base/components/header_bar/header_bar_search.tsx
dwnoble: Yeah that Yeah i think shrinking the font is a decent solution. Agreed any sort of wrapping can be separate.

Another thought: in mobile we could just pick from a pool of shorter questions

static/css/core.scss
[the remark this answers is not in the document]
dwnoble: Yeah that Yeah i think shrinking the font is a decent solution. Agreed any sort of wrapping can be separate.

Another thought: in mobile we could just pick from a pool of shorter questions

static/js/components/nl_search_bar/dynamic_placeholder_helper.ts
dwnoble: I kind of like it starting right away like https://www.thrillophilia.com/, but let's push this one out to autopush to get some more feedback

dwnoble (approved):
lgtm! thanks Gabe

PR #4975, opened 2025-02-25
Allow v2node to fetch more pages with nextToken.

clincoln8:


server/services/datacommons.py
```diff
@@ -263,6 +263,55 @@ def v2node(nodes, prop):
   })
 
 
+def _merge_paged_response(result, paged_response):
```
chejennifer, line 266: name this specifically for node, maybe something like merge_node_v2_response?

server/services/datacommons.py
```diff
@@ -263,6 +263,55 @@ def v2node(nodes, prop):
...
+  result = {}
+  next_token = ''
+  while True:
+    url = get_service_url('/v2/node')
+    response = post(url, {
+        'nodes': sorted(nodes),
+        'property': prop,
+        'nextToken': next_token
+    })
+    if max_pages == 1:
```
chejennifer, line 304: instead of specifically returning response when max_pages == 1, maybe _merge_paged_response() can just return response if result is empty?

server/services/datacommons.py
```diff
@@ -263,6 +263,55 @@ def v2node(nodes, prop):
...
+    if 'properties' in paged_response['data'][dcid]:
+      result['data'][dcid].setdefault('properties', []).extend(
+          paged_response['data'][dcid].get('properties', []))
+
+  result['nextToken'] = paged_response.get('nextToken', '')
+  if not result['nextToken']:
+    del result['nextToken']
+
+
+def v2node_paginated(nodes, prop, max_pages=1):
```
chejennifer, line 285: since this function is for fetching paginated v2node, maybe default for max_pages should be None so when this function does get called, user has to consciously decide to only fetch a specific number of pages?

server/services/datacommons.py
```diff
@@ -263,6 +263,55 @@ def v2node(nodes, prop):
   })
 
 
+def _merge_paged_response(result, paged_response):
+  for dcid in paged_response.get('data', []):
```
chejennifer, line 267: it looks like data is an object, so default should be {}?

server/services/datacommons.py
```diff
@@ -263,6 +263,55 @@ def v2node(nodes, prop):
...
+    # Initialize dcid in data even when no arcs or properties are returned
+    result.setdefault('data', {}).setdefault(dcid, {})
```
chejennifer, line 269: nit: maybe use a variable to hold this so can just reuse it in the rest of the code & maybe break apart the setdefault from extend so that it's a little more readable. something like:

```
node_result = result.setdefault('data', {}).setdefault(dcid, {})
....
prop_nodes = node_result.setdefault('arcs', {}).setdefault(prop, {}).setdefault('nodes', [])
prop_nodes.extend(...)
...

File: server/tests/services/datacommons_test.py [hunk 6]
```diff
@@ -0,0 +1,93 @@
...
+
+
+class TestServiceDataCommonsV2NodePaginated(unittest.TestCase):
+
+  @mock.patch('server.services.datacommons.post')
+  def test_termination_condition_max_pages_fetched(self, mock_post):
+    response_with_next_token = get_json('v2node_response_with_next_token')
+    mock_post.side_effect = [response_with_next_token, response_with_next_token]
+
+    v2node_paginated(['dc/1', 'dc/2'], '->{property1,property2}', max_pages=2)
```
chejennifer, line 41: assert what the response should look like?

server/tests/services/datacommons_test.py
```diff
@@ -0,0 +1,93 @@
...
+  with open(filepath, "r") as f:
+    return json.load(f)
+
+
+class TestServiceDataCommonsV2NodePaginated(unittest.TestCase):
+
+  @mock.patch('server.services.datacommons.post')
+  def test_termination_condition_max_pages_fetched(self, mock_post):
+    response_with_next_token = get_json('v2node_response_with_next_token')
+    mock_post.side_effect = [response_with_next_token, response_with_next_token]
```
chejennifer, line 39: should also check that the request going to the mock post request is correct by only returning mocked response  for the specific request like: https://github.com/datacommonsorg/website/blob/master/server/tests/routes/api/series_test.py#L214-L245

chejennifer (commented):
nice & thank you for adding tests!

server/services/datacommons.py
chejennifer: I see what you mean, that makes sense to me

server/services/datacommons.py
clincoln8: Ah sorry, this was from before when I was directly editing the v2node function. Now that I'm just creating a new one, I can delete this condition.

Thanks for calling it out!!

server/services/datacommons.py
clincoln8: added, thanks for the tip!

server/tests/services/datacommons_test.py
clincoln8: added input checks in side_effect functions. Is it okay to asserts like how I did it?

server/tests/services/datacommons_test.py
clincoln8: added

clincoln8 (commented):
thanks for the tips!

chejennifer (approved):
thanks for updating!

PR #4976, opened 2025-02-26
Add script to create a new feature flag in all environments and create biomed_nl flag.

clincoln8:


server/config/feature_flag_configs/local.json
```diff
@@ -1,25 +1,33 @@
...
     "deprecated": true,
     "enabled": true,
     "owner": "gmechali",
     "description": "Feature flag to enable the scroll to top button on the explore and place pages."
-}]
\ No newline at end of file
+  },
+  {
+    "name": "biomed_nl_routes",
+    "enabled": false,
```
chejennifer, line 29: I think we want this at least enabled for local?

server/config/feature_flag_configs/autopush.json
```diff
@@ -1,25 +1,33 @@
...
     "deprecated": true,
     "enabled": true,
     "owner": "gmechali",
     "description": "Feature flag to enable the scroll to top button on the explore and place pages."
-}]
\ No newline at end of file
+  },
+  {
+    "name": "biomed_nl_routes",
+    "enabled": false,
```
chejennifer, line 29: we could also enable it in autopush so as we're developing we can see the changes as they get into autopush?

scripts/create_feature_flag.sh
```diff
@@ -0,0 +1,55 @@
...
+# Construct the JSON object
+new_entry='
+{
+  "name": "'"$flag_name"'",
+  "enabled": false,
+  "owner": "'"$owner"'",
+  "description": "'"$description"'"
+}'
+
+environments=(local dev autopush staging production custom)
```
chejennifer, line 42: instead of hardcoding this, we can just iterate through every file in server/config/feature_flag_configs?

docs/feature_flags.md
```diff
@@ -34,30 +37,34 @@ Where `<environment>` is one of:
...
+* The script can be modified to upload different files or perform additional actions after the deployment.
+
+## Adding a New Flag
+
+1. **Add flag in flag configurations**: Check the flags in the Github master branch [`server/config/feature_flag_configs`](https://github.com/datacommonsorg/website/tree/master/server/config/feature_flag_configs). The flag must be added to all environments.
+    * TIP: Use [this script](https://github.com/datacommonsorg/website/tree/master/scripts/create_feature_flag.sh) to add a new flag to each config file.
+    > ```
+    > ./scripts/create_feature_flag.sh <flag_name> <owner> <description>
+    > ```
 2. **Define flag in server layer**: Add your flag constant to [feature_flags.py](https://github.com/datacommonsorg/website/blob/master/server/lib/feature_flags.py#L19) helper file for use in the API layer.
```
chejennifer, line 66: should we also do steps 2 and 3 to define the new flag? or are you doing that in a separate PR?

scripts/create_feature_flag.sh
```diff
@@ -0,0 +1,55 @@
...
+# ./scripts/create_feature_flag.sh <flag_name> <owner> <description>
+
+# Check if the correct number of arguments is provided
+if [ $# -ne 3 ]; then
+    echo "Usage: $0 <flag_name> <owner> <description>"
+    exit 1
+fi
+
+# Get the arguments
+flag_name=$1
```
gmechali, line 29: Could you just prompt the user for this instead of taking it as a flag. It allows us to put some more context, and these descriptions will be multi word, so it can get annoying to put it all in the command line.

scripts/create_feature_flag.sh
gmechali: +1, we never know how this list may grow

server/config/feature_flag_configs/autopush.json
gmechali: for the record when I've been adding a feature, I always just enable it in all environments except staging, prod & custom

docs/feature_flags.md
gmechali: I'd say those are fine to leave out because you may only need the flag in the server layer or FE layer, and you should only add your flag if you need it. So i'd argue against adding that to script?

gmechali (approved):

docs/feature_flags.md
chejennifer: oh I don't mean adding it to the script, I just mean doing these steps for the biomed nl flag

server/config/feature_flag_configs/autopush.json
clincoln8: I'll keep it off dev for now and then enable it when it's ready. I've enabled for local and autopush.

server/config/feature_flag_configs/local.json
clincoln8: enabled for local and autopush

docs/feature_flags.md
clincoln8: I've added the biomed flag to the server layer.

scripts/create_feature_flag.sh
clincoln8: Changed to prompt the user for input instead of taking command line params.

scripts/create_feature_flag.sh
clincoln8: Changed script to update all json files in the feature_flag_configs dir.

clincoln8 (commented):
Thanks both for the review!

server/tests/lib/util_test.py
```diff
@@ -1776,12 +1776,13 @@ def test_fetch_highest_coverage_with_no_observation_dates(
 
 
 class TestFeatureFlagsTest(unittest.TestCase):
-  FEATURE_FLAG_COUNT = 6
```
clincoln8, line 1779: @gmechali Just a FYI for this modification. Let me know if this might have unintended consequences.

server/lib/feature_flags.py
```diff
@@ -18,8 +18,9 @@
 AUTOCOMPLETE_FEATURE_FLAG = 'autocomplete'
 PLACE_PAGE_EXPERIMENT_FEATURE_FLAG = 'dev_place_experiment'
 PLACE_PAGE_GA_FEATURE_FLAG = 'dev_place_ga'
+BIOMED_NL_ROUTES_FEATURE_FLAG = 'biomed_nl_routes'
```
chejennifer, line 21: sorry, just catching this now, but why make the flag specifically just for routes? can we do 'biomed_nl', so that it's a flag for the full feature, including the client piece

clincoln8: I'm planning on conditionally registering the html and api routes based on this flag. Is that the same a a flag for the full feature?

chejennifer: yes, when you register the html and api routes, the full feature should show up. I just meant, we can just name the flag "biomed_nl" instead of "biomed_nl_routes"?

chejennifer (approved):

PR #4977, opened 2025-02-26
Prints the feature flags in the version page

gmechali:
At the bottom of the version page, include the list of feature flags and their status.
See screenshot: https://screenshot.googleplex.com/5NJRjTht3Z3iCPA

This also disables the scroll_to_top feature flag in autopush. The flag is no longer used since the following PR made it to production:
https://github.com/datacommonsorg/website/pull/4957

Once I submit this PR, I will just validate the feature is still working in autopush. And then I will delete the flag from all environments


clincoln8 (approved):
Thanks!!

PR #4978, opened 2025-02-26
[Feature Flags] Create a CloudBuild action to only trigger on feature flag updates and upload the new flags to GCS

gmechali:
This PR refactors the script to remove the prompts such that it can be run programmatically from cloudbuild. 
It also adds a cloud build action to upload to the appropriate GCS buckets when the files have been updated.


scripts/update_gcs_feature_flags.sh
```diff
@@ -169,11 +175,13 @@ fi
...
 if [[ "$environment" == "production" ]]; then
-    read -p "Have you validated these feature flags in staging? (yes/no) " -n 1 -r
-    echo
-    if [[ ! $REPLY =~ ^[Yy]$ ]]; then
-      echo "Aborting deployment to production."
-      exit 1
+    if [[ -z "$skip_staging_prompt" || "$skip_staging_prompt" != "true" ]]; then
+        read -p "Have you validated these feature flags in staging? (yes/no) " -n 1 -r
+        echo
+        if [[ ! $REPLY =~ ^[Yy]$ ]]; then
```
clincoln8, line 181: nit: Should the prompt display the options "(y/n)" instead of "(yes/no)" if this condition is checking for "y" or "Y"? 
Or is it common practice to assume that for "(yes/no)" options you just put "y"?

build/ci/cloudbuild.update_feature_flags.yaml
```diff
@@ -0,0 +1,39 @@
...
+      - |
+        modified_file="${_MODIFIED_FILE}"
+        filename=$(basename "$modified_file")
+        filename_without_extension="${filename%.*}"
+        should_restart="false"
+        skip_staging_prompt="true"
+        ./scripts/update_gcs_feature_flags.sh "$filename_without_extension" "$skip_staging_prompt" "$should_restart"
+
+options:
+  substitution_option: 'ALLOW_LOOSE'
```
clincoln8, line 29: What does this do/mean?

clincoln8 (approved):
Thanks!!

Once this trigger is active, we should update the feature flags readme with more info about this.

scripts/update_gcs_feature_flags.sh
gmechali: This was entirely written by Gemini :)
It looks like a custom practice, that you just check the first letter tbh.

build/ci/cloudbuild.update_feature_flags.yaml
gmechali: https://cloud.google.com/build/docs/configuring-builds/substitute-variable-values#using_user-defined_substitutions

Gemini added it, I looked into it but it just seems to be for more flexibility when running.

PR #4979, opened 2025-02-26
Reduce flake in cdc test that blocked autopush

gmechali:
Adds the ability to run a single test file at a time in cdc test. 

I think this change deflaked a test a little but not too sure why...


hqpho (approved):
"autiopush" -> "autopush" in PR title

PR #4981, opened 2025-02-26
Enable dev place GA in autopush and local

gmechali:
Fixes python tests and webdriver tests to allow keeping the tests disabled.


server/tests/utils.py
```diff
@@ -0,0 +1,22 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Helpers for tests"""
+
+from typing import List
+
+
+def mock_feature_flags(app, flags: List[str], enabled: bool):
```
clincoln8, line 19: maybe add a note that this should never be used in non-test code? Just in case!

clincoln8 (approved):

server/tests/utils.py
gmechali: done - it's also in the /tests/ directory so hopefull it shouldnt be a problem :)

PR #4984, opened 2025-02-27
Add --debug to website helm upgrade command.

clincoln8:


gmechali (approved):

clincoln8:
If the output become to large -- we can revert this change.

keyurva (approved):
Thanks!

PR #4986, opened 2025-02-27
[NL] add url params for topic limit and subtopic sv limits

chejennifer:
Parameterize the limits for setting max topics that get opened in fulfill and  the max number of svs from subtopics that get processed when processing topics

### DIFF
--- a/server/lib/nl/explore/params.py
+++ b/server/lib/nl/explore/params.py
@@ -53,8 +53,8 @@ class Params(str, Enum):
   SKIP_RELATED_THINGS = 'skipRelatedThings'
   # The default place to use if no place detected in query
   DEFAULT_PLACE = 'default_place'
-  # The maximum number of SVs to open for subtopics when processing topics
-  MAX_SUBTOPIC_SVS = 'maxSubtopicSvs'
+  # The maximum number of SVs to open for subtopics within a topic
+  MAX_TOPIC_SVS = 'maxTopicSvs'
   # The maximum number of topics to process when generating a result
   MAX_TOPICS = 'maxTopics'
 
--- a/server/lib/nl/explore/params.py
+++ b/server/lib/nl/explore/params.py
@@ -51,6 +51,12 @@ class Params(str, Enum):
   RERANKER = 'reranker'
   # If set, then don't get related things
   SKIP_RELATED_THINGS = 'skipRelatedThings'
+  # The default place to use if no place detected in query
+  DEFAULT_PLACE = 'default_place'
+  # The maximum number of SVs to open for subtopics when processing topics
+  MAX_SUBTOPIC_SVS = 'maxSubtopicSvs'
+  # The maximum number of topics to process when generating a result
+  MAX_TOPICS = 'maxTopics'
 
 
 class DCNames(str, Enum):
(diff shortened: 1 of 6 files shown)


server/lib/nl/explore/params.py
```diff
@@ -51,6 +51,12 @@ class Params(str, Enum):
   RERANKER = 'reranker'
   # If set, then don't get related things
   SKIP_RELATED_THINGS = 'skipRelatedThings'
+  # The default place to use if no place detected in query
+  DEFAULT_PLACE = 'default_place'
+  # The maximum number of SVs to open for subtopics when processing topics
+  MAX_SUBTOPIC_SVS = 'maxSubtopicSvs'
```
keyurva, line 57: nit: From a user perspective, is the "sub" part relevant or just saying "max variables per topic" enough? If so, consider renaming to `maxTopicVariables`.

server/lib/nl/explore/topic.py
```diff
@@ -301,6 +299,10 @@ def _direct_chart_vars(svs: List[str], svpgs: List[str], source_topic: str,
 
 
 def _max_subtopic_sv_limit(state: ftypes.PopulateState) -> int:
+  # If there was a limit specified in the insight context, use that limit.
+  if state.uttr.insight_ctx.get(Params.MAX_SUBTOPIC_SVS) != None:
+    return state.insight_ctx[Params.MAX_SUBTOPIC_SVS]
```
keyurva, line 304: Should we have a cap on these values? e.g. 100 for topics, 500 for SVs?

server/routes/explore/helpers.py
```diff
@@ -257,6 +259,26 @@ def parse_query_and_detect(request: Dict, backend: str, client: str,
...
+    Args:
+      request: the original api request
+      utterance: the utterance generated for the request
+      dc_name: the value to use for the parameter "dc"
+  """
+  utterance.insight_ctx[params.Params.DC.value] = dc_name
+  for p in [params.Params.MAX_SUBTOPIC_SVS, params.Params.MAX_TOPICS]:
+    param_val = request.args.get(p, None)
+    if param_val:
+      param_val = int(param_val)
```
keyurva, line 278: Should we handle errors if non-numeric values are specified?

keyurva (commented):
Thanks!

server/lib/nl/explore/topic.py
chejennifer: I don't actually think this is necessary because we process topics one at a time & after we process each topic, we check how many variables have been processed & stop if we've reached the max_svs (which is just always 500)

server/lib/nl/explore/params.py
chejennifer: good point, updated!

server/routes/explore/helpers.py
chejennifer: done

server/lib/nl/explore/topic.py
keyurva: Ah, yes! Just confirming that even if someone sends a request with `?maxTopicSvs=10000` or `?maxTopics=10000`, once it has processed 500 SVs, it will return. Is that correct?

keyurva (approved):

server/lib/nl/explore/topic.py
chejennifer: correct!

chejennifer:
thanks for the review!

PR #4994, opened 2025-03-04
[biomed_nl] Add BMDC graph traversal logic.

clincoln8:


server/routes/experiments/biomed_nl/traversal.py
```diff
@@ -0,0 +1,536 @@
...
+                               key=lambda item: item[1],
+                               reverse=True)[:num_top_properties]
+    ]
+
+    self.path_store.filter_by_properties(top_props)
+
+    return top_props
+
+  def select_paths(self):
+    '''Selects paths to purpose using the Gemini model based on a prompt.
```
chejennifer, line 470: what does paths to purpose mean?

server/routes/experiments/biomed_nl/traversal.py
```diff
@@ -0,0 +1,536 @@
...
+
+    Returns:
+        True if the traversal should terminate, False otherwise.
```
chejennifer, line 473: this is slightly confusing because from the method description, I would assume this is for selecting the paths and returning those paths..

server/routes/experiments/biomed_nl/traversal.py
```diff
@@ -0,0 +1,536 @@
...
+      # This branch only occurs if there has been an error with Gemini.
+      return True
+
+    paths = [
+        line.strip()
+        for line in response.text.split('\n')
+        if bool(re.match(r"^path\d+:", line.strip()))
+    ]
+
+    self.selected_paths = PathStore.from_selected_paths(paths, self.path_store)
```
chejennifer, line 507: curious why we need to save this instead of just returning it?

server/tests/routes/experiments/biomed_nl/traversal_test.py
```diff
@@ -0,0 +1,572 @@
...
+# limitations under the License.
+
+import unittest
+from unittest.mock import MagicMock
+from unittest.mock import patch
+
+from deepdiff import DeepDiff
+
+# Assuming the code from the original file is in a file named 'your_module.py'
+# and fetch and utils are mocked appropriately
```
chejennifer, line 22: what do you mean by code from the original file?

server/tests/routes/experiments/biomed_nl/traversal_test.py
```diff
@@ -0,0 +1,572 @@
...
+    }
+    assert DeepDiff(path_store.path_store,
+                    expected_path_store,
+                    ignore_order=True) == {}
+
+  @patch('server.lib.fetch.triples')
+  def test_traversal_without_sampling(self, mock_fetch):
+    mock_fetch.side_effect = fetch_triples_side_effect
+
+    def null_sample(_):
```
chejennifer, line 213: curious why using this instead of using an input which will cause sampling to not happen OR if sampling always happens then we don't need this test case

chejennifer (commented):
nice!

server/routes/experiments/biomed_nl/traversal.py
```diff
@@ -0,0 +1,536 @@
...
+    '''
+
+    all_prop_dcids = {
+        prop for paths in self.path_store.values() for path in paths
+        for prop in Path.parse(path, include_incoming_types=False)
+    }
+
+    descriptions = fetch.property_values(all_prop_dcids, 'description')
+
+    merged_descriptions = {}
```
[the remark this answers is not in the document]
clincoln8: Merged might not be the right term here. Essentially I need to transform the raw/fetched descriptions in two ways.
1. if no description is available (which is common), then set the description to just the property's dcid.
2. Add the property's dcid to the description by prepending '{property} means'. This is because the embeddings require the dcid to be in the embeddings string for it to work.

I've refactored this to just update the existing self.property_descriptions so there's no extra object.

server/routes/experiments/biomed_nl/traversal.py
```diff
@@ -0,0 +1,536 @@
...
+      else:
+        merged_descriptions[prop] = '. '.join([
+            f"{prop} means {description}"
+            for description in descriptions.get(prop)
+        ])
+
+    self.property_descriptions = merged_descriptions
+
+  def get_property_descriptions(self):
+    if not self.property_descriptions:
```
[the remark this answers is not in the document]
clincoln8: good point! reduced the get and fetch functions down to one method. 
We first check to see if there are any props that are not in property_descriptions already. If yes --> fetch, if no --> return the prefetched descriptions

server/routes/experiments/biomed_nl/traversal.py
```diff
@@ -0,0 +1,536 @@
...
+
+    Args:
+        selected_properties: A list of property names to filter by.
+    '''
+
+    filtered_path_store = {}
+    for start_dcid, paths in self.path_store.items():
+      for path, next_dcids in paths.items():
+        # If the path contains any of the selected_properties, then keep it.
+        if any(property in Path.parse(path, include_incoming_types=False)
```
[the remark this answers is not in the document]
clincoln8: updated

server/routes/experiments/biomed_nl/traversal.py
```diff
@@ -0,0 +1,536 @@
...
+
+    Returns:
+        A list of the top properties that were selected for filtering.
+    '''
+
+    property_descriptions = self.path_store.get_property_descriptions()
+
+    # TODO: add a description to description in DC KG.
+    property_descriptions['description'] = (
+        'The description describes the entity by its characteristics or '
```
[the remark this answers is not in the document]
clincoln8: done

server/routes/experiments/biomed_nl/traversal.py
clincoln8: typo -- I meant paths to _pursue_. Thanks for catching!!

server/routes/experiments/biomed_nl/traversal.py
```diff
@@ -0,0 +1,536 @@
...
+
+  def fetch_property_descriptions(self):
+    '''Fetches and stores descriptions for properties in the path store.
+
+    Updates `self.property_descriptions` with descriptions retrieved from
+    DC KG.  If a description isn't found, the property name itself is used as
+    the description.
+    '''
+
+    all_prop_dcids = {
```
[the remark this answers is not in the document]
clincoln8: done

server/routes/experiments/biomed_nl/traversal.py
clincoln8: updated this so selected_paths are saved in the path_store. do you think that's an improvement?

server/routes/experiments/biomed_nl/traversal.py
clincoln8: updated the method name and doc string to be a little more consistent. Does that help?

server/tests/routes/experiments/biomed_nl/traversal_test.py
clincoln8: this is to indicate that the focus of this test is on the functionality of traversal and that the sampling logic is tested elsewhere in the file. Otherwise there would be more mocking and more convoluted inputs/outputs. 

I can add the extra mocks for the sampling part if you think it's best to do so! Let me know!

server/routes/experiments/biomed_nl/traversal.py
```diff
@@ -0,0 +1,536 @@
...
+      for path, next_dcids in paths.items():
+        # If the path contains any of the selected_properties, then keep it.
+        if any(property in Path.parse(path, include_incoming_types=False)
+               for property in selected_properties):
+          filtered_path_store.setdefault(start_dcid, {})[path] = next_dcids
+
+    self.path_store = filtered_path_store
+
+
+class PathFinder:
```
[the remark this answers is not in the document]
clincoln8: done!

chejennifer (approved):
thanks for the updates! Just a few more nits

PR #5001, opened 2025-03-07
[Place Page Revamp] Fix bug in place type ordering to only keep values in the list

gmechali:


clincoln8 (approved):
thanks for the quick fix!

PR #5010, opened 2025-03-10
Updated scatter plot link

dwnoble:
Due to some recent datasets being removed, the scatter plot tool sample link "Literate population per capita vs schools per capita for states in india" wasn't working.

This change updates that link to use Literate populatoin per capita vs Population: Below Poverty Level in The Past 12 Months, which does have data.


clincoln8 (approved):
Thanks for the update!

PR #5013, opened 2025-03-11
[Feature Flags] Pass variable through the workspace

gmechali:
Replicating what is described in https://cloud.google.com/build/docs/configuring-builds/pass-data-between-steps#passing_data_using_workspaces


clincoln8 (approved):
I tested the individual steps locally on master (and echoing the update_gcs_feature_flag script params instead of actually running the script) -- it seems to work!!

the git diff command picked up that the staging file was modified between head and previous head and the second stage was able to get the file name from the flag. 

Hopefully passing through /workspace will work!

PR #5021, opened 2025-03-11
[Place Page] Reverting test to use logic from legacy place page

gmechali:
Custom DC is still using the old place page, so autopush tests started failing since this part of the test assumed it would use the new place page.


clincoln8 (approved):
thanks!!

PR #5025, opened 2025-03-12
[Feature Flags] Sort feature flag upload stages and make staging/prod check for enabled flags only

gmechali:
Sort the stages as autopush, dev, staging, prod.

Change the staging/prod check such that we allow diffs between the files as long as no flag is enabled in production but disbled in staging.
For example, I can enable 2 flags in staging, then only enable one of them in production.

I tested the script by running the compare_staging_prod check while forcing it to use autopush as the source. It correctly identified the incorrectly configured flags.


clincoln8 (approved):
Thanks Gabe!

PR #5026, opened 2025-03-12
Added new place page i18n strings

dwnoble:
* New translations from cl/735615367
* Instructions: https://github.com/datacommonsorg/website/blob/master/docs/localization.md


beets (approved):
are there any screenshots or tests that need to be updated with these new strings?

dwnoble:
> are there any screenshots or tests that need to be updated with these new strings?

Updated with some test case fixes and a new screenshot test

PR #5029, opened 2025-03-15
I18n fixes

dwnoble:
* Fixed issue in `labels.html` that was preventing the navbar translations from showing up
* Updated place page i18n tests to use new place page
* Migrated place page loading indicator from flask to react, added "ProgressActivty" icon & "Loadsing" component to show localized loading indicator
* Fixed issue preventing ranking tile footnote translation from showing up
* Fixed issue where the place page category name wasn't showing up localized in the title. (For example, the title was showing up as `"日本 • Equity"` instead of `"日本 • 資産"` for the japan equity page)
* Localized place page links
* Added missing translations for "see per capita", date ranges, and the default NL search query placeholder.
* Renamed i18n_tile_messages.ts to i18n_messages.ts because these translations apply to more than just tiles
* Removed unused CHART_TITLE-CHART_CATEGORY-Overview translation`

Before:
autopush-before

After:
38gefaDuQQg2CmX


server/templates/dev_place.html
```diff
@@ -54,11 +54,6 @@
     <div id="plage-page-main" class="container">
       <div id="place-page-content" class="page-content-container">
       </div>
-      <div id="page-loading" class="mt-4" style="display:none;">
```
beets, line 57: how are we planning to show a loading message moving forward?

server/webdriver/shared_tests/place_explorer_i18n_test.py
```diff
@@ -25,43 +28,69 @@ class PlaceI18nExplorerTestMixin():
   def test_japan_in_japanese(self):
     """Test translations from various sources are displayed correctly."""
 
-    start_url = self.url_ + '/place/country/JPN?hl=ja&disable_dev_places=true'
```
beets, line 28: should we fork these, so the old pages are tested until we complete the migration?

i think at least one test for i18n for the old pages should be kept

static/js/i18n/i18n_place_messages.ts
```diff
@@ -200,14 +200,32 @@ export const pageMessages = defineMessages({
...
+    id: "place_not_found",
+    defaultMessage: 'Place "{placeDcid}" not found.',
+    description:
+      'Text to display on a 404 page when a place ID is not found. Example: "Place "dcid:geoId/04" not found.". Please leave the {placeDcid} variable unchanged in the resulting translation.',
+  },
+  categoryNotFound: {
+    id: "category_not_found",
+    defaultMessage: 'Category not found for place "{place}".',
+    description:
+      'Text to display on a 404 page when a category is not found. Example: "Category not found for place "Austria".". Please leave the {place} variable unchanged in the resulting translation.',
```
beets, line 223: should we mention what category was not found? something like: "No economics data found for Austria" ?

beets: Hmm, I don't see this being referenced anywhere..

beets (approved):
A few comments but overall looks good. Thanks for adding the extra error handling too.

server/templates/dev_place.html
beets: Ah, I see the new component below

static/js/i18n/i18n_place_messages.ts
dwnoble: Whoops meant to remove this. removed

PR #5033, opened 2025-03-17
[biomed_nl] Use structured outputs for entity recognition and final response prompts, handle typos and fetching too much data

clincoln8:
Example of structured output from final response:

image
image


chejennifer (commented):
Started taking a look at this PR and realized there's some changes to the prompts and steps in the flow that are beyond just formatting. Before I continue reviewing, have you run the original biomed queries against these changes to check for regressions?

clincoln8 (commented):
Overall, this change seems reasonable in terms of response quality. Compared the colab, there's now 11 improvements in responses and 5 regressions and one timeout. Most the regressions are more minor or actually related to traversal which I didn't change the prompt for.

See https://docs.google.com/spreadsheets/d/1uvL990dvqhes7FoFf-iPuiwc2qGt042K69FYuygZVIc/edit?gid=1335743036#gid=1335743036

Also, adding entity chips to the UI will help the cases that don't respond accurately/completely.

server/routes/experiments/biomed_nl/traversal.py
```diff
@@ -586,21 +593,97 @@ class PathFinder:
...
+    prompt = utils.PARSE_QUERY_PROMPT.format(QUERY=self.raw_query)
+
+    gemini_response = self.gemini.models.generate_content(
+        model=self.gemini_model_str,
+        contents=prompt,
+        config={
+            'response_mime_type': 'application/json',
+            'response_schema': PathFinder.ParseQueryResponse,
+        })
+    parsed_response = PathFinder.ParseQueryResponse(
```
chejennifer, line 660: why is parsing the query response going through PathFinder when this response doesn't have to do with path finding. maybe ParseQueryResponse should be a util function

server/routes/experiments/biomed_nl/traversal.py
```diff
@@ -586,21 +593,97 @@ class PathFinder:
...
+        **json.loads(gemini_response.text))
+    self.query_type = parsed_response.query_type
+
+    # Only traverse from one starting point
```
chejennifer, line 664: nit: "Take the first entity because we only want to traverse from one starting point", the original comment makes me think we're starting the traversal here

server/routes/experiments/biomed_nl/traversal.py
```diff
@@ -586,21 +593,97 @@ class PathFinder:
...
+    for entity in parsed_response.entities:
+      str_to_raw[entity.sanitized_str] = entity.raw_str
+      for synonym in entity.synonyms:
+        str_to_raw[synonym] = entity.raw_str
+    entities_to_dcids, entities_to_recognized_types = recognize_entities_from_query(
+        ' '.join((str_to_raw.keys())))
+
+    start_dcids = set()
+    for entity, dcids in entities_to_dcids.items():
+      if entity in start_strs:
```
chejennifer, line 679: add a comment that we are only adding the dcids of the start entity

server/routes/experiments/biomed_nl/traversal.py
```diff
@@ -814,7 +896,9 @@ def get_traversed_entity_info(self):
 
           # Only fetch triples for a given dcid one time.
           fetch_dcids = [dcid for dcid in next_dcids if dcid not in entity_info]
-          if fetch_dcids:
+          is_not_last_hop = index < (len(properties_in_path) - 1)
+          should_fetch_hop = is_not_last_hop or len(properties_in_path) < 3
```
chejennifer, line 900: make 3 a constant?

chejennifer (commented):
nice!

server/routes/experiments/biomed_nl/api.py
```diff
@@ -21,76 +22,126 @@
...
     traversed_entity_info = path_finder.get_traversed_entity_info()
-    # TODO: add error handling.
 
-    final_prompt = utils.FINAL_PROMPT.format(
-        sentence=query, json_str=utils.format_dict(traversed_entity_info))
+  except Exception as e:
+    logging.error(f'[biomed_nl]: {e}', exc_info=True)
+    if path_finder.start_dcids:
+      response[
+          'answer'] = f'{path_finder.start_entity_name}: {", ".join(path_finder.start_dcids)}'
```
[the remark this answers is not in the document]
clincoln8: will include in follow up pr!

server/routes/experiments/biomed_nl/api.py
```diff
@@ -21,76 +22,126 @@
...
+    # TODO: add error handling.
+    final_prompt = utils.FINAL_RESPONSE_PROMPT.format(
+        QUERY=query, ENTITY_INFO=utils.format_dict(traversed_entity_info))
 
     input_tokens = gemini_client.models.count_tokens(
         contents=final_prompt, model=GEMINI_PRO).total_tokens
-    gemini_response = None
-    if input_tokens < GEMINI_PRO_TOKEN_LIMIT:
+
+    if input_tokens > GEMINI_PRO_TOKEN_LIMIT * 0.75:
```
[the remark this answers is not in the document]
clincoln8: moved

server/routes/experiments/biomed_nl/traversal.py
clincoln8: discussed offline -- the name ParseQueryResponse was confusing, updated to be more clear

server/routes/experiments/biomed_nl/traversal.py
clincoln8: added

server/routes/experiments/biomed_nl/traversal.py
clincoln8: updated

server/routes/experiments/biomed_nl/traversal.py
clincoln8: done

server/routes/experiments/biomed_nl/api.py
```diff
@@ -21,76 +22,145 @@
...
-      return response
-
-    response['answer'] = gemini_response.text
-    response['debug'] += '\n' + utils.format_dict(
-        path_finder.path_store.get_paths_from_start(only_selected_paths=True))
+      should_include_fallback = True
 
+    # TODO: return fallback for traversal error + timeouts and uncertain final responses
+    if should_include_fallback:
+      _get_fallback_response(query, response, path_finder,
```
chejennifer, line 157: you would want to either set this to the response or return this right?

chejennifer (commented):
thanks for updating, just one more thing

server/routes/experiments/biomed_nl/api.py
clincoln8: I updated the function name to append, so that we append this to the response.

chejennifer (approved):

PR #5037, opened 2025-03-19
A script for automating local CDC runs

kmoscoe:
This is a convenience script for end users and us to run different Docker commands, with different scenarios, in a local dev environment. This will greatly simplify the initial user experience (and documentation!) By default it runs both Docker containers, using the stable CDC release. It offers options for:
- 5 different workflows covering build, run, and upload
- Selecting between stable and latest releases
- Building and running a custom image
- Building and uploading a custom image to GCP Artifact registry
- Running data container in schema update mode only
- Checking various GCP credentials for actions that require it
It reads options from both the env.list file and from the command line.
Complete documentation is given in a help here document at the top of the file.

I've tested all the following combinations (it's impossible to go through all of them!)

Valid input
=============
./run_local_docker.sh  : should start both 'stable ' containers
./run_local_docker.sh --container service --release latest : should start only service container with 'latest' release
./run_local_docker.sh --image my-datacommons:dev --container service : should start only service container with custom-built image
./run_local_docker.sh --actions build_run --image my-datacommons:dev --container service : should build a custom image and start only the service container with it
./run_local_docker.sh --image my-datacommons:dev : should run 'stable' data container, and custom service container
./run_local_docker.sh --actions build --image my-datacommons:dev : should build a custom image and stop
./run_local_docker.sh --actions build_upload --image my-datacommons:dev : should warn about missing package, build a custom image, create a package with the same name, and upload it to Artifact Registry repo, no containers run
./run_local_docker.sh --actions upload --image my-datacommons:dev  : should should warn about missing package, create a package with the same name as a previously built image, and upload it to Artifact Registry repo, no containers run
./run_local_docker.sh --actions upload --image my-datacommons:dev --package my-datacommons:latest : should create a package with a new name from an existing image, and and upload it to Artifact Registry repo, no containers run
./run_local_docker.sh -e custom_dc/examples/env.list  : Should start both 'stable' containers, using the dataset specified in the examples dir
./run_local_docker.sh --schema_update -- should run both stable containerss, with data container in schema update actions

Invalid input
=============
./run_local_docker.sh garbage input : should print an error and exit
./run_local_docker.sh --actions build myimage : should print an error and exit
./run_local_docker.sh --container service --image latest : should print an error and exit
./run_local_docker.sh --build_run --container service : should print an error and exit (handled by getopt)
./run_local_docker.sh --image my-datacommons:dev --container data : should print an error and exit
./run_local_docker.sh --actions upload --image my-datacommons:dev --container service : should silently ignore container option, give info about missing package option, package and upload custom image, and run no containers
./run_local_docker.sh --actions build  : should print an error and exit
./run_local_docker.sh --actions build --image my-datacommons:dev --schema_update : should silently ignore invalid schema update option and just build image
./run_local_docker.sh --actions build --image my-datacommons:dev --container all  : should silently ignore container option, build image and just build image


run_local_docker.sh
```diff
@@ -0,0 +1,254 @@
...
+    -v $INPUT_DIR:$INPUT_DIR \
+    -v $OUTPUT_DIR:$OUTPUT_DIR \
+    gcr.io/datcom-ci/datacommons-data:$RELEASE
+  fi
+}
+
+run_service() {
+  if [ -n "$CUSTOM_IMAGE" ]; then
+    echo -e "\n${GREEN}Starting Docker services container with custom image ${YELLOW}${CUSTOM_IMAGE}...${NC}\n"
+    docker run -it \
```
ajaits, line 100: can we skip `-t` so this script can be run in background mode too?

run_local_docker.sh
```diff
@@ -0,0 +1,254 @@
...
+  docker build --tag $CUSTOM_IMAGE \
+  -f build/cdc_services/Dockerfile .
+}
+
+run_data() {
+  if [ $SCHEMA_UPDATE == true ]; then
+    echo -e "\n${GREEN}Starting Docker data container in schema update mode...${NC}\n"
+    docker run -it \
+    --env-file $PWD/custom_dc/env.list \
+    -e DATA_UPDATE_MODE=schemaupdate \
```
ajaits, line 83: since this is the only extra option, can be simplified with a common call to docker:
```
$UPDATE_OPTS=""
if [ $SCHEMA_UPDATE == true ]; then
  echo -e "\n${GREEN}Starting Docker data container in schema update mode...${NC}\n"
  UPDTE_OPTS="-e DATA_UPDATE_MODE=schemaupdate"
else
  echo -e "\n${GREEN}Starting Docker data container...${NC}\n"
fi
docker run -it \
    --env-file $PWD/custom_dc/env.list \
  $UPDATE_OPTS \ 
   -v $INPUT_DIR:$INPUT_DIR \
    -v $OUTPUT_DIR:$OUTPUT_DIR \
    gcr.io/datcom-ci/datacommons-data:$RELEASE
```

run_local_docker.sh
```diff
@@ -0,0 +1,254 @@
...
+  echo -e "\n${GREEN}Starting Docker build...${NC}\n"
+  docker build --tag $CUSTOM_IMAGE \
+  -f build/cdc_services/Dockerfile .
+}
+
+run_data() {
+  if [ $SCHEMA_UPDATE == true ]; then
+    echo -e "\n${GREEN}Starting Docker data container in schema update mode...${NC}\n"
+    docker run -it \
+    --env-file $PWD/custom_dc/env.list \
```
ajaits, line 82: safer to use quotes around variables that may have spaces like $PWD
`--env-file "$PWD/custom_dc/env.list" `

run_local_docker.sh
```diff
@@ -0,0 +1,254 @@
...
+# Check that env.list file exists
+if [ ! -f "custom_dc/env.list" ]; then
+  echo -e "Error: Configuration file ${YELLOW}env.list${NC} not found."
+  echo -e "1. Copy custom_dc/env.list.sample and save as custom_dc/env.list."
+  echo -e "2. Set all necessary environment variables." 
+  echo -e "For details, see comments in the file or refer to https://docs.datacommons.org/custom_dc/quickstart.html#env-vars\n"
+  exit 1
+fi
+
+# Read required variables from env.list file
```
ajaits, line 130: why not source the env.list to get all variables?
`source custom_dc/env.list`

run_local_docker.sh
```diff
@@ -0,0 +1,254 @@
...
+
+# Usage: ./run_local_docker.sh [--mode|-m <running mode>] ([--release|-r stable|latest] | 
+#        [--custom_image|-c <image name and tag>]) [--schema_update|-s]
+#
+# If no options are set, the defaults are --mode run_all --release stable
+#
+# Options:
+#   --mode|-m <running mode>:
+#       Optional. Available options are:
+#       * run_all (default): Run data and services containers.
```
ajaits, line 25: would it be easier to keep build and run as separate options to reduce combinations?

`--build --run <all|service|none>`

run_local_docker.sh
```diff
@@ -0,0 +1,254 @@
...
+#       Use this if you haven't made any changes to your data but just want
+#       to pick up the latest code.
+#   ./run_local_docker.sh --mode build_and_run_all --custom_image my-datacommons:dev
+#       This builds a custom image and starts all containers.
+#   ./run_local_docker.sh --mode build_only --custom_image my-datacommons:prod
+#       This builds a custom image but does not start any containers. Use this
+#       if you are building a custom image to upload to Google Cloud.
+
+set -e
+
```
ajaits, line 63: you can do a `cd $(dirname $0)` to get to the directory for this script irresctive of where the script is invoked from. This makes it easier for callers.

run_local_docker.sh
kmoscoe: First of all, thank you so much for the review, Ajai!!!

The thing is that I intervene for each new docker command/container to keep giving status updates. I suppose users could skip that if they wanted to.

run_local_docker.sh
kmoscoe: Awesome, this simplifies hugely! Thank you!

run_local_docker.sh
kmoscoe: Will do.

run_local_docker.sh
kmoscoe: Yes, that would be so much better and simpler, but I wasn't sure if @keyurva would be OK with that. Would you, Keyur? It would greatly simplify things to not have to get variable settings from both a file and stdin.

run_local_docker.sh
kmoscoe: Yes, especially since I just added 4 more combinations for some more advanced scenarios. I think that might be a good idea. It will require more input handling, but simplify other things.

run_local_docker.sh
kmoscoe: Great, thank you!

run_local_docker.sh
kmoscoe: Oh, I just realized that you were asking a different question. The reason is that we need very few of all the variables in there, especially not all the GCP ones, so I didn't want to have to process the whole file when it's kind of unnecessary. Do you have a way to do that using 'source'?

Also, env.list doesn't actually allow quotation marks, so we tell them to never put them.

kmoscoe: Done.

run_local_docker.sh
```diff
@@ -0,0 +1,421 @@
...
+#-------------------------------------------------------------
+# Needed for docker run -v option
+if [ -z "$INPUT_DIR" ] || [ -z "$OUTPUT_DIR" ]; then
+  echo -e "${RED}ERROR:${NC} Missing input or output data directories.\nPlease set 'INPUT_DIR' and 'OUTPUT_DIR' in your env.list file.\n" 1>&2
+  exit 1
+fi
+
+# Needed for docker tag and push
+if  [[ ( "$ACTIONS" == *"upload"* ) && ( -z "$GOOGLE_CLOUD_PROJECT" || -z "$GOOGLE_CLOUD_REGION" )]]; then
+  echo -e "${RED}ERROR:${NC} Missing GCP project and region settings.\nPlease set 'GOOGLE_CLOUD_PROJECT' and/or 'GOOGLE_CLOUD_REGION' in your env.list file.\n" 1>&2
```
ajaits, line 377: instead of `echo - e "...\n"`, can do `echo "..."`

ajaits: I see, you need -e for the colour... pls retain that.

run_local_docker.sh
```diff
@@ -0,0 +1,421 @@
...
+  fi
+  schema_update='""'
+  schema_update_text=""
+  if [ "$SCHEMA_UPDATE" == true ]; then
+    schema_update="-e DATA_UPDATE_MODE=schemaupdate"
+    schema_update_text=" in schema update mode"
+  fi
+  echo -e "${GREEN}Starting Docker data container with '$RELEASE' release${schema_update_text}...${NC}\n"
+  docker run -it \
+  --env-file $ENV_FILE \
```
ajaits, line 169: safer to use double quotes for any user supplied argument like input folders. Windows users tend to use spaces in the folder names.

`--env-file "$ENV_FILE"` \

same for all `"$INPUT_DIR"` etc.

run_local_docker.sh
```diff
@@ -0,0 +1,421 @@
...
+      ;;
+    -h | --help)
+      help
+      exit 0
+      shift
+      ;;
+    --) 
+      shift
+      break 
+      ;;
```
ajaits, line 333: May be useful to add a hidden debug option to debug the script:
`-x) set -x; shift;;`


To print help message for any unknown option, you can use 
`*) echo -e "Unknown option '$1'"; help; exit 1;;`

ajaits (approved):
Thanks Kara... you should get shell readability now :)

run_local_docker.sh
kmoscoe: Re the first point, that's a great idea, I'll do that. 
Re the second point, getopt handles this automatically, and no matter what I did I could not disable it to do my own error handling. Supposedly there is a special option, ':' that you put before the option names in the call to getopt to disable its automatic checking. But it didn't work for me. If I use the wildcard as you say, it will catch absolutely everything, including valid options.

run_local_docker.sh
kmoscoe: Even the ones that come from the env.list file itself, like "INPUT_DIR"?

run_local_docker.sh
```diff
@@ -0,0 +1,254 @@
+#!/bin/bash
+# Copyright 2024 Google LLC
```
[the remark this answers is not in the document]
kmoscoe: Argh, this is really odd, because I thought I did...

PR #5038, opened 2025-03-19
Updated flakey map viz test

dwnoble:
Attempting to resolve a flakey test by removing an element check race condition


hqpho (approved):

dwnoble:
> If you wanna verify this fixes the flakiness you can run with `--flake-finder` as described in go/deflake-datcom!

This particular test seems to only be failing in cloud build 🤦

PR #5044, opened 2025-03-21
Moved new place page i18n test out of shared testing folder to avoid CDC test failures

dwnoble:
Moved new place page i18n test out of shared testing folder to avoid CDC test failures  (the new place page is disabled in CDC, which was causing the failure)


clincoln8:
thanks Dan!!

PR #5046, opened 2025-03-21
Updated chart axis labels to use localized links

dwnoble:
Fixed issue where place links on the bar chart axis were always opening up the english, even if the chart was in a different locale

ja-place-links


beets (approved):
thanks for the fix! sorry for the late review

gmechali (approved):
Thanks Dan! Sorry for slow review!

PR #5048, opened 2025-03-21
[biomed_nl] Remove sentence_transformers from website server + UI nits

clincoln8:
Includes:
* re-implementing cosine-score calculation via numpy
* small ui layout changes

after PR:
image

before PR:
image

### DIFF
--- a/server/routes/experiments/biomed_nl/utils.py
+++ b/server/routes/experiments/biomed_nl/utils.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 import json
+
 import numpy as np
 
 PARSE_QUERY_PROMPT = '''
--- a/server/routes/experiments/biomed_nl/utils.py
+++ b/server/routes/experiments/biomed_nl/utils.py
@@ -345,6 +345,11 @@ def cos_sim(embeddings1, embeddings2):
   embeddings1 = np.array(embeddings1)
   embeddings2 = np.array(embeddings2)
 
+  if len(embeddings1.shape) == 1:
+    embeddings1 = embeddings1.reshape(1, -1)
+  if len(embeddings2.shape) == 1:
+    embeddings2 = embeddings2.reshape(1, -1)
+
   # Normalize the embeddings
   embeddings1_norm = embeddings1 / np.linalg.norm(
       embeddings1, axis=1, keepdims=True)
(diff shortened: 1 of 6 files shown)


server/routes/experiments/biomed_nl/utils.py
```diff
@@ -325,3 +327,37 @@ def batch_requested_nodes(nodes, batch_size=MAX_NUM_DCID_PER_V2NODE_REQUEST):
...
+    embeddings2 = embeddings2.reshape(1, -1)
+
+  # Normalize the embeddings
+  embeddings1_norm = embeddings1 / np.linalg.norm(
+      embeddings1, axis=1, keepdims=True)
+  embeddings2_norm = embeddings2 / np.linalg.norm(
+      embeddings2, axis=1, keepdims=True)
+
+  # Calculate the cosine similarity
+  similarity_matrix = np.dot(embeddings1_norm, embeddings2_norm.T)
```
chejennifer, line 361: what is the `.T` at the end of embeddings2_norm for?

chejennifer (commented):
nice, just 1 question

server/routes/experiments/biomed_nl/utils.py
clincoln8: transpose -- here's the original sentence_transformers cos_sim code ([link](https://github.com/UKPLab/sentence-transformers/blob/master/sentence_transformers/util.py#L92-L108)) ; they also take the transpose

it's so that the final similarity scores are in the right order. since we only use this for 1 embedding vs 1 embedding, it doesn't really make a difference because the final output is just a 1x1 "matrix"

clincoln8: would it make more sense to change it to a single vector vs vector comparison? 

i kept the matrix thing just to be consistent with sentence_transformers, but we just need single scores in the current implementation

chejennifer: oh interesting, I was just confused because the `.T` was only added to one of the embeddings, but if that's how sentence_transformers does it, let's just keep that! Let's be consistent with sentence_transformers for now

chejennifer (approved):

server/routes/experiments/biomed_nl/utils.py
clincoln8: sg! thanks for the review!!

PR #5051, opened 2025-03-24
Take 2 at fixing flakey visualization tool test

dwnoble:
- Updated the test to wait for both the highest & lowest ranking charts to load before asserting the lists had text content
- Removed `wait_for_loading` option to avoid a potential race condition when waiting for the place "display name" to show up in the ranking chart

### DIFF
--- a/server/requirements.txt
+++ b/server/requirements.txt
@@ -21,7 +21,7 @@ google-cloud-storage==2.15.0
 google-genai==1.3.0
 gunicorn==22.0.0
 isort==5.10.0
-jinja2==3.1.5
+jinja2==3.1.6
 json5==0.9.14
 langdetect==1.0.9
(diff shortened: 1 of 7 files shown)


clincoln8 (approved):
Thanks Dan! One question -- what was the reasoning behind originally setting wait_for_loading to False for these tests?

dwnoble:
> Thanks Dan! One question -- what was the reasoning behind originally setting wait_for_loading to False for these tests?

I believe the intent was to only wait once for the entire list of places to load, and not to wait for the two subsequent checks when re-ordering by clicking per capita and the sources option.

But, i think this can cause issues because the place "display name" is only loaded dynamically once the place is actually visible in the ranking chart. I updated the pr description with a note about it. cc @hqpho for additional context

PR #5052, opened 2025-03-25
Disable web crawlers (GoogleBot) on dev

clincoln8:
Adds a new config variable (DISABLE_CRAWLERS) which disables all crawlers access to content within the entire site. 
This is set to False by default in the Base Config and only set to True for the Dev Config.

Screenshots of robots.txt when running locally: 

DISABLE_CRAWLERS == False: 
image

DISABLE_CRAWLERS == True:
image

### DIFF
--- a/server/routes/static.py
+++ b/server/routes/static.py
@@ -21,10 +21,12 @@ from flask import Blueprint
 from flask import current_app
 from flask import redirect
 from flask import render_template
-from flask import send_from_directory
 from flask import request
+from flask import send_from_directory
 
+from server.lib.cache import cache
 import server.lib.render as lib_render
+from server.routes import TIMEOUT
 from server.services import datacommons as dc
 
 bp = Blueprint('static', __name__)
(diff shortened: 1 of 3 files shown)


server/routes/static.py
```diff
@@ -115,3 +118,12 @@ def version():
       bigquery=mixer_version.get('bigquery', ''),
       featureFlags=current_app.config.get('FEATURE_FLAGS', []),
       remote_mixer_domain=mixer_version.get('remoteMixerDomain', ''))
+
+
+@bp.route('/robots.txt')
+@cache.cached(timeout=TIMEOUT)
```
clincoln8, line 124: @dwnoble  Is this the correct way to enable caching and is the timeout set to the right amount?

server/routes/static.py
```diff
@@ -115,3 +118,12 @@ def version():
...
+def robots_config():
+  if current_app.config.get('DISABLE_CRAWLERS', False):
+    return "User-agent: *<br>Disallow: /"
```
dwnoble, line 127: Use `\n` instead of `<br>`, and make sure to return the result with content-type: "text/plain"

server/routes/static.py
dwnoble: That looks right to me

clincoln8: thanks!

server/routes/static.py
clincoln8: done:
image

dwnoble (approved):

PR #5057, opened 2025-03-27
[Cleanup] Remove scroll to top flag from all environemnts

gmechali:
All references to the SCROLL_TO_TOP flag have been removed in https://github.com/datacommonsorg/website/pull/4957. This PR has been in 2 prod releases, and the flag has been disabled + marked as deprecated for weeks. We verified we can still see the feature even with it disabled in prod.


clincoln8 (approved):
Should we standardize the PR comment for removing flags in the documentation?

For example "all references to the flags have been removed X prod releases ago" or something similar.

We'd want to make sure that the flag reference removal PR is (1) in prod today and (2) in at least one previous prod releases that it wouldn't be an issue if we had to roll prod back today.

gmechali:
> Should we standardize the PR comment for removing flags in the documentation?
> 
> For example "all references to the flags have been removed X prod releases ago" or something similar.
> 
> We'd want to make sure that the flag reference removal PR is (1) in prod today and (2) in at least one previous prod releases that it wouldn't be an issue if we had to roll prod back today.

Yup fair point, I added some more text around it to confirm a rollback would be safe + we've verified this has no impact.

PR #5059, opened 2025-03-27
[Place Page] Create a Redirection helper to preserve flags. Use that on autocomplete redirection.

gmechali:
https://screencast.googleplex.com/cast/NTc4NjAwMDQ4NjgyNTk4NHw4MzY0YjI1ZC01Yg


static/js/shared/util.ts
```diff
@@ -207,3 +208,43 @@ export function stripPatternFromQuery(query: string, pattern: string): string {
...
+ * @param originUrl Current URL from which to extract URL parameters
+ * @param destinationUrl Desitnation URL to follow
+ * @param overrideParams Parameters to override.
+ */
+export function redirect(
+  originUrl: string,
+  destinationUrl: string,
+  overrideParams: URLSearchParams = new URLSearchParams()
+): void {
+  const originParams = extractFlagsFromUrl(originUrl);
```
dwnoble, line 237: I would only preserve specific query params here (hl & the feature flag params - maybe we can by convention add a prefix to feature flag params?) so we don't end up keeping query params we actually may not want. I could see a case where preserving a url param from another page gives us some unintended side effects when redirecting to the place page

gmechali: Can you expand on what would be an example where we don't want to keep all flags?
It seems a little arbitrary to limit to hl and feature flags, and introducing a convention for feature flags also adds some divergence from just using the feature flag name, which would be annoying.

dwnoble: A user could be on say the "Energy" page for a given place like https://datacommons.org/place/geoId/04?category=Energy 

And searching for a different place could land them on another place, but if we preserve all params, we'd keep  "category=Energy" so they'd see the energy page instead of the Overview. There's also a chance the energy page doesn't exist like in https://autopush.datacommons.org/place/wikidataId/Q1353?q=Delhi,%20India&category=Energy)

dwnoble: And that's fair about the feature flag names. but at the same time we wouldn't want to accidentally use a feature flag name that also exists as a URL param.

gmechali: Modifying it now to just do the HL param. I ll work on a follow up for handling the feature flag override, and will include feature flags then.

dwnoble (approved):

PR #5065, opened 2025-03-27
[biomed_nl] Minor UI fixes, ensure fallback prompt is below token limit

clincoln8:
UI Changes:
* Fix links for citations of the format [#, #] 
  * image
* Add "back to samples" link
  * image
 * Fix extra escaping 
   * image
* Move crashing query to bottom of Sample list for now

Backend:
* Truncate fallback prompt if it exceeds allowed num of input tokens
* Standardize API response output format into a pydantic class


static/js/apps/biomed_nl/app.tsx
```diff
@@ -412,6 +455,19 @@ export function App(): ReactElement {
...
+                      <span
+                        onClick={resetToSampleQueries}
+                        css={css`
+                          color: ${theme.colors.link.primary.base};
+                          cursor: pointer;
+                        `}
+                      >
+                        &larr; Back
+                      </span>
+                      &nbsp;to see sample queries.
```
chejennifer, line 469: nit: I'm not sure how I feel about the color change between back and the rest of the string .. what if we just do "Back to sample queries" all as a link?

server/routes/experiments/biomed_nl/api.py
```diff
@@ -80,14 +89,29 @@ def _append_fallback_response(query, response, path_finder,
       SELECTED_PATHS=utils.format_dict(selected_paths),
       ENTITY_INFO=utils.format_dict(entity_info))
 
+  token_count = gemini_client.models.count_tokens(
+      model=GEMINI_PRO,
+      contents=fallback_prompt,
+  ).total_tokens
+  while token_count > GEMINI_PRO_TOKEN_LIMIT:
+    logging.info('[biomed_nl] truncate fallback prompt')
```
chejennifer, line 97: nit: instead of logging this for every round of the while loop, can we just do it once? maybe get the count before and after the while loop & log if there is a difference

chejennifer (commented):
nice! just some suggestions

server/routes/experiments/biomed_nl/api.py
clincoln8: the loop should really only run one time; if it's running more than that, then I want to know so that I can fix it. I updated to log total number of truncation loops instead of each loop as a warning if it ran more than once.

static/js/apps/biomed_nl/app.tsx
clincoln8: done!

image

chejennifer (approved):
thanks for updating!

PR #5067, opened 2025-03-27
Removed feedback buttons from NL search results page

dwnoble:
Before:
feedback-before

After:
feedback-after


chejennifer (approved):
thanks for removing!

PR #5068, opened 2025-03-27
Updated summary overview text to just summary. 

dwnoble:
Rationale: the wording "Summary overview" for place pages can be seen as redundant, and is leading to translations like (fr) "Aperçu du résumé", or literally, "the summary of the overview"

- Updated "summary overview" text to just "summary"
- Updated translator hint for the summary text tooltip

After: https://screenshot.googleplex.com/ABdRhmHXVzskkbz


static/js/i18n/i18n_place_messages.ts
```diff
@@ -169,16 +169,16 @@ export const pageMessages = defineMessages({
   },
   SummaryOverview: {
     id: "summary_overview",
```
gmechali, line 171: Should we be making a new message for this?

We need them to retranslate it, will this do that?

dwnoble: This should still request a new translation. The benefit of keeping the ID the same is that users in other locales will still see the old "summary overview" translation until the new "summary" translation comes back.

gmechali (approved):

PR #5070, opened 2025-03-27
Added api modal feature flag

dwnoble:
Just adding the feature flag for now


gmechali:
You have a python test to update which just tracks the number of feature flags. Im wondering if we should just kill that test (by that I mean just ensure all files have the same number of flags, but not use the static flag count)

gmechali (approved):

dwnoble:
> You have a python test to update which just tracks the number of feature flags. Im wondering if we should just kill that test (by that I mean just ensure all files have the same number of flags, but not use the static flag count)

That sounds like a good approach - i updated the unit test to check if all environments have the same number of feature flags

PR #5073, opened 2025-03-28
[biomed_nl] Add entity chips to final result, timeout final entity fetching after 3 minutes 

clincoln8:
Examples with answer+ chips:
image
image

Example no answer, only chips:
image

### DIFF
--- a/server/routes/experiments/biomed_nl/entity_recognition.py
+++ b/server/routes/experiments/biomed_nl/entity_recognition.py
@@ -20,7 +20,7 @@ import server.services.datacommons as dc
 MIN_SAMPLES = 3
 
 
-def sample_dcids_by_type(dcids, min_sample_size):
+def sample_dcids_by_type(entity_name, dcids, min_sample_size):
   """Samples a set of DCIDs such that all unique types of the original DCID list
   are present in the sample.
 
@@ -44,7 +44,8 @@ def sample_dcids_by_type(dcids, min_sample_size):
       if 'name' in type_node
   }
 
-  unique_types = list(remaining_unique_types)
+  graph_entities = []
+  skipped_entities = []
 
   sample_dcids = []
   first_pass_skipped = []
@@ -59,12 +60,19 @@ def sample_dcids_by_type(dcids, min_sample_size):
       # needs to be added to the set since they are now represented in the
       # sample set.
       remaining_unique_types.difference_update(dcid_types)
+      graph_entities.append(
+          utils.GraphEntity(dcid=dcid, types=dcid_types, name=entity_name))
       sample_dcids.append(dcid)
     elif not remaining_unique_types:
       sample_dcids.append(dcid)
+      graph_entities.append(
+          utils.GraphEntity(dcid=dcid, types=dcid_types, name=entity_name))
+
       if len(sample_dcids) >= min_sample_size:
         break
     else:
+      skipped_entities.append(
+          utils.GraphEntity(dcid=dcid, types=dcid_types, name=entity_name))
       first_pass_skipped.append(dcid)
 
   # If some dcids were skipped in the first pass to find all unique types, but
@@ -73,8 +81,8 @@ def sample_dcids_by_type(dcids, min_sample_size):
   num_skipped_samples_to_add = min_sample_size - len(sample_dcids)
   if num_skipped_samples_to_add > 0:
     sample_dcids.extend(first_pass_skipped[:num_skipped_samples_to_add])
-
-  return sample_dcids, unique_types
+    graph_entities.extend(skipped_entities[:num_skipped_samples_to_add])
+  return graph_entities
 
 
 def recognize_entities_from_query(query):
@@ -93,8 +101,8 @@ def recognize_entities_from_query(query):
         (strings).
   """
   recognize_response = dc.recognize_entities(query)
-  entities_to_dcids = {}
-  entities_to_recognized_types = {}
+
+  detected_entities = []
   for item in recognize_response:
     queried_entity = item['span']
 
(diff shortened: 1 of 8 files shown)


server/routes/experiments/biomed_nl/utils.py
```diff
@@ -307,6 +308,12 @@
 MAX_NUM_DCID_PER_V2NODE_REQUEST = 100
 
 
+class GraphEntity(BaseModel):
```
chejennifer, line 312: nit: add comments documenting this class?

server/routes/experiments/biomed_nl/entity_recognition.py
```diff
@@ -73,8 +81,8 @@ def sample_dcids_by_type(dcids, min_sample_size):
   num_skipped_samples_to_add = min_sample_size - len(sample_dcids)
   if num_skipped_samples_to_add > 0:
     sample_dcids.extend(first_pass_skipped[:num_skipped_samples_to_add])
-
-  return sample_dcids, unique_types
```
chejennifer, line 77: do we still need to keep sample_dcids now that we have graph_entities?

server/routes/experiments/biomed_nl/utils.py
```diff
@@ -307,6 +309,18 @@
...
+  name: str
+  types: list[str]
+  dcid: str
+
+  def __eq__(self, other):
+    if isinstance(other, GraphEntity):
+      return (self.name == other.name and self.dcid == other.dcid and
+              Counter(self.types) == Counter(other.types))
```
chejennifer, line 320: could there be more than 1 of each type? could just use set(self.types) == set(other.types) if each type should only appear once?

server/routes/experiments/biomed_nl/traversal.py
```diff
@@ -686,21 +693,32 @@ def find_start_and_traversal_type(self):
...
-    for entity, dcids in entities_to_dcids.items():
-      if entity in start_strs:
-        start_dcids.update(dcids)
     raw_to_types = {}
-    for entity, types in entities_to_recognized_types.items():
-      raw_to_types.setdefault(str_to_raw[entity], []).extend(types)
+    for entity in detected_entities:
+      raw_to_types.setdefault(str_to_raw.get(entity.name, entity.name),
+                              []).extend(entity.types)
+      if any(entity.name in start_str for start_str in start_strs):
```
chejennifer, line 707: slightly confused if entity.name is actual name of the entity as it is in the KG or the string in the query that resolved to this entity? Maybe add some comments clarifying this

server/routes/experiments/biomed_nl/traversal.py
```diff
@@ -686,21 +693,32 @@ def find_start_and_traversal_type(self):
...
+        start_dcids.add(entity.dcid)
+
+      if entity.dcid in displayed_entities:
+        combined_types = list(
+            set(displayed_entities[entity.dcid].types) | set(entity.types))
+        displayed_entities[entity.dcid].types = combined_types
```
chejennifer, line 713: could there be multiple detected entities with the same dcid but different name? wouldn't this overwrite the entity's name field if there could be different names and is that ok?

static/js/apps/biomed_nl/app.tsx
```diff
@@ -247,16 +258,90 @@ function formatCitationsInResponse(answer: string): string {
   return annotatedAnswer;
 }
 
+function formatDetectedEntities(entities: GraphEntity[]): JSX.Element {
```
chejennifer, line 261: add a function comment?

static/js/apps/biomed_nl/app.tsx
```diff
@@ -247,16 +258,90 @@ function formatCitationsInResponse(answer: string): string {
...
+      })}
+    </>
+  );
+}
+
 function processApiResponse(response: BiomedNlApiResponse): DisplayedAnswer {
   const formattedAnswer = formatCitationsInResponse(response.answer);
   const feedbackLink = constructFeedbackLink(response);
   const tripleReferences = formatReferences(response.footnotes);
+  const formattedEntites = formatDetectedEntities(response.entities);
```
chejennifer, line 337: nit: formattedEntities

chejennifer (commented):
nice!

server/routes/experiments/biomed_nl/entity_recognition.py
clincoln8: thanks for pointing this out, forgot to delete this after I confirmed the entity approach worked!

server/routes/experiments/biomed_nl/traversal.py
clincoln8: This is the name as it is in the KG. Added a comment!

server/routes/experiments/biomed_nl/traversal.py
clincoln8: Yes! It's basically whichever name it sees first is the name we'll show in the response. It's quicker for now to just show one name for an entity instead of listing all of them, especially since we're also displaying the dcid and types, I think the chips would get a little cramped. 
We can revisit this if it becomes an issue and work out a good way to display multiple names for a single entity.

static/js/apps/biomed_nl/app.tsx
clincoln8: fixed!

static/js/apps/biomed_nl/app.tsx
clincoln8: done

server/routes/experiments/biomed_nl/utils.py
```diff
@@ -307,6 +309,24 @@
 MAX_NUM_DCID_PER_V2NODE_REQUEST = 100
 
 
+class GraphEntity(BaseModel):
+  '''
+  A single node from the Data Commons knowledge graph, represented by it name, types, and dcid.
```
chejennifer, line 314: nit: "represented by its name, types .."

chejennifer (approved):
thanks for updating, just one grammar thing

PR #5075, opened 2025-03-28
[biomed_nl] migrate from hash param to URL param

clincoln8:
Previous url format: 
> https://localhost:8080/experiments/biomed_nl/#q=What%20is%20the%20chembl%20ID%20of%20acetaminophen%3F

then clicking on the citation link would replace the hashParam with the scrollTo:
> https://localhost:8080/experiments/biomed_nl/#biomednl-ref-1

New URL format: 
> http://localhost:8080/experiments/biomed_nl/?query=What+is+the+chembl+ID+of+acetaminophen%3F
clicking on citation: 
> http://localhost:8080/experiments/biomed_nl/?query=What+is+the+chembl+ID+of+acetaminophen%3F#biomednl-ref-1

### DIFF
--- a/static/js/apps/biomed_nl/app.tsx
+++ b/static/js/apps/biomed_nl/app.tsx
@@ -22,7 +22,6 @@
 import { css, ThemeProvider } from "@emotion/react";
 import axios from "axios";
 import _ from "lodash";
-import queryString from "query-string";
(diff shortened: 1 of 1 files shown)


chejennifer (approved):
nice

PR #5076, opened 2025-03-28
Fix 37 eslint warnings

clincoln8:
Random cleanup PR for some eslint warnings: fixes 37 warnings mostly pertaining to missing return types and variable name casing.


static/js/apps/eval_embeddings/app.tsx
```diff
@@ -43,6 +43,12 @@ interface appState {
   embeddings: Record<string, EmbeddingObject[]>;
 }
 
+interface EmbeddingResult {
```
clincoln8, line 46: structure inferred from lines 257-260

PR #5081, opened 2025-03-31
Fix caching issue for robots.txt

clincoln8:
Tested with LocalConfig.USE_MEMCACHE=True with DISABLE_CRAWLERS both True and False to ensure both options work with this new approach.


dwnoble (approved):

dwnoble:
Thanks Christie! we should also add an uptime or similar check to validate /robots.txt doesn't error in staging

PR #5087, opened 2025-04-01
Updated update gcs feature flag script to restart us-west1 cluster

dwnoble:


scripts/update_gcs_feature_flags.sh
```diff
@@ -161,11 +161,13 @@ restart_kubernetes_deployment() {
 
   if [[ "$environment" == "prod" ]]; then
     gcloud container clusters get-credentials website-us-west1 --region us-west1 --project "datcom-website-${environment}"
+    kubectl rollout restart deployment website-app -n website
```
gmechali, line 164: eventually we should migrate this to just list off the environments and restart them

gmechali (approved):

PR #5088, opened 2025-04-01
Fixed bug where place page header link was keeping the category param instead of redirecting to the overview page

dwnoble:
Screenshot fixes this:

place-header

### DIFF
--- a/static/js/place/dev_place_main.tsx
+++ b/static/js/place/dev_place_main.tsx
@@ -73,6 +73,12 @@ const PlaceHeader = (props: {
     );
   });
 
+  const placeHref = createPlacePageCategoryHref(
+    "Overview",
+    forceDevPlaces,
+    place
+  );
+
   return (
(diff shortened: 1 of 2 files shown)


beets (approved):

static/js/place/dev_place_main.tsx
```diff
@@ -87,9 +81,11 @@ const PlaceHeader = (props: {
             {selectedCategory.name === "Overview" ? (
               place.name
             ) : (
-              <a className="place-info-link" href={placeHref}>
-                {place.name}
-              </a>
+              <LocalizedLink
```
gmechali, line 84: Here you're just replicating the logic we built in createPlacePageCategoryHref, I would argue it'd be better to reuse that and just pass in Overview as the destination category.

dwnoble: That's reasonable- updated!

PR #5089, opened 2025-04-01
Fixed issue where places with no 'name' attribute were showing a generic 'no data found' issue

dwnoble:
Before:

err-before

After:
err-after

### DIFF
--- a/static/js/place/dev_place_main.tsx
+++ b/static/js/place/dev_place_main.tsx
@@ -24,6 +24,7 @@ import { ThemeProvider } from "@emotion/react";
 import React, { useEffect, useState } from "react";
 import { RawIntlProvider } from "react-intl";
 
+import styled from "@emotion/styled";
 import { Loading } from "../components/elements/loading";
 import { ScrollToTopButton } from "../components/elements/scroll_to_top_button";
 import { SubjectPageMainPane } from "../components/subject_page/main_pane";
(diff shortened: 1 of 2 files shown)


beets (approved):
interesting -- there's a nameWithLanguage field, but none for EN.

dwnoble:
> interesting -- there's a nameWithLanguage field, but none for EN.

Yeah- maybe in this case we should be setting the `name` attribute to a non-english locale, like `Schönermark@de` for the case of https://datacommons.org/browser/wikidataId/Q995366

beets:
We can also do this by fixing the data so all our places have a name (which
we should have some resolution data for)

dwnoble:
> We can also do this by fixing the data so all our places have a name (which we should have some resolution data for)
> […](#)
> Message ID: ***@***.***>

Agreed

static/js/place/dev_place_main.tsx
```diff
@@ -338,15 +344,15 @@ export const DevPlaceMain = (): React.JSX.Element => {
       console.error("Error loading place page metadata element");
       return;
     }
-    if (
-      pageMetadata.dataset.placeDcid != "" &&
-      pageMetadata.dataset.placeName === ""
-    ) {
+    if (pageMetadata.dataset.placeDcid == "") {
```
gmechali, line 347: nit: we should probably import _ from lodash and use the isEmpty method?

gmechali (approved):

static/js/place/dev_place_main.tsx
dwnoble: done

PR #5093, opened 2025-04-02
Added canonical and alternate link tags to new place page html head

dwnoble:
the canonical and alternate language link tags were dropped when migrating to the new place page. this change adds them back in. This may have been the cause for these indexing errors: https://screenshot.googleplex.com/BNP7krKTihZQZrL

Before: links were added in the HTTP header in the old place page:
Screenshot 2025-04-02 at 4 14 13 PM

Now: these links are added back in, but thi
Screenshot 2025-04-02 at 4 17 39 PM
s time in the html <head> element:

i find this easier to maintain than in the HTTP headers - looks like both [approaches are valid](https://developers.google.com/search/docs/crawling-indexing/consolidate-duplicate-urls), gemini recommends adding the links to the <head> tag for what its worth.


server/routes/place/html.py
```diff
@@ -576,11 +640,14 @@ def dev_place(place_dcid=None):
   else:
     place_summary = ""
 
+  canonical_links = get_canonical_links(place_dcid, place_category)
+  print(canonical_links)
```
gmechali, line 644: nit: remove

server/routes/place/html.py
```diff
@@ -576,11 +640,14 @@ def dev_place(place_dcid=None):
...
   return flask.render_template('dev_place.html',
                                maps_api_key=current_app.config['MAPS_API_KEY'],
                                place_dcid=place_dcid,
                                place_name=place_name,
+                               place_summary=place_summary,
```
gmechali, line 649: why is the place summary removed?

gmechali (approved):

server/routes/place/html.py
dwnoble: done

PR #5094, opened 2025-04-03
Place Data Overview Page

nick-nlb:
## Description

This PR implements an initial version of the place data overview page, as referenced on approximately page 66 of the go/dc-rolling-ux slideshow.

The primary purpose of this PR is to create and prepare the various components that make up the page. However, the initial framework of the page is also created.

### Feature Flag

The availability of the routing for this page is determined by a new feature flag: `place_data_overview`, set to true only in `local`. 

### Routing

The base route is `/data/place/<place dcid>`. Canada can be accessed at `/data/place/country/CAN`

### Data

Pending the availability of actual information, this page currently uses mock data for Canada (as per the dc-rolling-ux example). Other places will not be available (unless they are also added to the mock data). Once the source of available information is determined, this can be replaced with a fetch of the actual data for the page. Typing is provided for this data (which may be altered to fit once the real data is available).

### Components

Some of the basic components (link_chips, section) have been enhanced to allow for the kind of customization required here. As part of this, we've introduce `sx` props - inspired from Material UI (MUI) - that allow consumers of these components to freely override the styles of those components. We've also added a generic box component.

Then there are the specific components used on this page. These should be able to be used with minimal alterations once the real data is available.

### Stat Var Hierarchy

Because of the inclusion of the StatVarHierarchy component, we also need to include SCSS associated with this page, as well as the Material fonts. We could put the StatVarHierarchy on the list of components to be converted to Emotion, which would allow it to be used here and in other places without bringing in these other assets.

## Testing

This page can be viewed at: http://localhost:8080/data/place/country/CAN 

## Screenshots

data_place_country


static/js/apps/data/places/app.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+const placeDataOverview: PlaceDataOverview = placeDataOverviewData;
+
+/**
+ * Application container
+ */
+
+export function App(): ReactElement {
+  const pathname = window.location.pathname;
+  const dcid = pathname.replace(/^\/data\/place\//, "");
```
beets, line 41: could we pass this down to the app through a data object, so that it's agnostic of the page route?

server/routes/data/html.py
```diff
@@ -0,0 +1,25 @@
...
+"""Place Data overview routing"""
+
+from flask import Blueprint
+
+from server.lib import render as lib_render
+
+bp = Blueprint('data', __name__)
+
+
+@bp.route("/data/place/<path:place>")
```
beets, line 23: sorry if this was discussed already, but maybe replace "data" with "overview"? i'll let @dwnoble chime in here too.

server/config/feature_flag_configs/autopush.json
```diff
@@ -40,5 +40,11 @@
     "enabled": false,
     "owner": "dwnoble",
     "description": "Feature flag to show the API modal on web components."
+  },
+  {
+    "name": "place_data_overview",
+    "enabled": false,
```
beets, line 46: we can enable this for autopush

beets (commented):
looks great! i don't have many comments except for the route of the page. thanks for hooking this up so quickly

server/routes/data/html.py
dwnoble: how about data_overview?

static/js/apps/data/places/app.tsx
dwnoble: In other places we attach metadata in the DOM using a `<div id="metadata" data-place-dcid="..." />` tag like: https://github.com/datacommonsorg/website/blob/6b9a24ccffbc570ce6dbcbf9ae15949711031870/server/templates/dev_place.html#L50

static/js/apps/data/places/components/place_data_overview_header.tsx
```diff
@@ -0,0 +1,69 @@
...
+        `}
+      >
+        <h3
+          css={css`
+            ${theme.typography.family.heading};
+            ${theme.typography.heading.sm};
+            margin-left: ${theme.spacing.sm}px;
+          `}
+        >
+          <a href={`/place/${place.dcid}`}>{place.name}</a> · Data Overview
```
dwnoble, line 56: (can be in a future PR): add i18n string references here

dwnoble (approved):
Thank you!

PR #5096, opened 2025-04-03
Removed dev places experiment logic

dwnoble:
I'll remove the actual feature flag in a follow-up


beets (approved):

gmechali (approved):
LGTM - Just make sure not to remove the flags before the release goes out!

PR #5106, opened 2025-04-07
Fixed word wrap in localized headers

dwnoble:
- Bug: b/407053284
- Tested this locally in various locales and screen sizes

Before:
ja-before

After:
ja-after


n-h-diaz (approved):
Thanks!

gmechali (approved):

PR #5107, opened 2025-04-07
Refactor place page routes, various redirect fixes

dwnoble:
- Added a new route to handle the place explorer landing page and google search redirects
- Updated logic flow handle new place page rendering first, and legacy place page by exception
- Added redirect handling to new place page to:
  - strip trailing slashes (`/place/Earth/` redirects to `/place/Earth`)
  - redirect legacy categories to new categories (`?category=Climate` redirects to `?category=Environment`)
  - to redirect legacy 'topic' param to use new 'category' param (we still get quite a bit of traffic using the legacy `topic=` param in prod - @beets any idea why?)
  - Strip out invalid `category=abcdefg` params and redirect to the overview page
- Updated place_test.py redirect test case to use the new dev place page
- Fixed issue where place page "category" wasn't showing up in the title. Now the title reads:
  - `/place/country/USA?category=Demographics`: `<title>アメリカ合衆国 - 人口統計 - Data Commons</title>`
  - `/place/country/USA?hl=ja&category=Demographics`: `<title>United States of America - Demographics - Data Commons</title>`

### DIFF
--- a/server/routes/place/html.py
+++ b/server/routes/place/html.py
@@ -442,6 +442,26 @@ def get_canonical_links(place_dcid: str, place_category: str) -> List[str]:
   return links
 
 
+def redirect_to_place_page(dcid: str, request_args: MultiDict[str, str]):
+  """Redirect to the place page for the given DCID
+
+  Handles redirects from Google Search using old URL format
+  Args:
+    dcid: The DCID of the place to redirect to
+    request_args: The request arguments to forward to the place page
+  Returns:
+    A redirect to the place page
+  """
+  redirect_args = dict(request_args)
+  redirect_args['place_dcid'] = dcid
+  del redirect_args['dcid']
+  url = flask.url_for('place.place',
+                      **redirect_args,
+                      _external=True,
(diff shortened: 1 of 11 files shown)


server/routes/place/html.py
```diff
@@ -514,8 +442,88 @@ def get_canonical_links(place_dcid: str, place_category: str) -> List[str]:
...
+@cache.cached(query_string=True)
+def place_explorer():
+  """Renders the place explorer landing page.
+
+  Also handles redirects from Google Search to individual place pages if the
+  request includes a dcid.
+  """
+  dcid = flask.request.args.get('dcid', None)
+  if dcid:
+    # Handle redirects from Google Search using old URL format
```
gmechali, line 455: could we move this to a simple helper method named similarly to "doSearchRedirect"
That way we can also test that part individually?

gmechali (approved):

server/routes/place/html.py
dwnoble: done!

PR #5108, opened 2025-04-08
Make Place name clickable in multi-place explore pages

gmechali:
Before: https://screenshot.googleplex.com/5sPmWzXSCJRGbKU
After: https://screenshot.googleplex.com/952dKD3PCyhuHNv


static/js/apps/explore/result_header_section.tsx
```diff
@@ -66,25 +66,30 @@ export function ResultHeaderSection(
...
+    const places = props.pageMetadata.places;
+    const numPlaces = places.length;
+
+    if (numPlaces === 0) {
+      return <></>;
+    }
+
+    if (numPlaces === 1) {
+      return (
+        <a className="place-callout-link" href={`/place/${places[0].dcid}`}>
```
dwnoble, line 78: While your at it (or maybe in a follow-up), could you add localized links?

dwnoble (approved):
Nice catch

static/js/apps/explore/result_header_section.tsx
gmechali: ah ok I briefly looked into it. Let me add it in a follow up to close out this PR. I ll file a bug for tracking. But I think there's a little more work to do to support it in order to fetch the translated name.

gmechali: I guess it's just that the explore page isnt internationalized which we already know about. I know it s on our roadmap for the year. So I'll let it be like this for now until we can take on i18n for it!

PR #5114, opened 2025-04-11
[AIMode Attribution] Enable Explore page to inject a highlight chart and support multiple places.

gmechali:
This PR uses the URL Hash Params to add the ability to pass in multiple places, and pass in one statistical variable.
We inject the chart for the one stat var, and issue a second call to fetch the Topic of said SV, then issue another fulfill call to with the associated topic.

Adds support for multiple StatVars, and multiple Topics on the fulfill call. 
Does a very basic merging of the responses.

In this state, here's what the page looks like, while passing in:
* Rhode Island and Arizona
* Median_Income_Household, Area_Farm, Count_Person
* TIMELINE_WITH_HIGHLIGHT for chartType

http://localhost:8080/explore/#sv=Median_Income_Household___Area_Farm___Count_Person&p=geoId%2F48___geoId%2F06&client=ui_related_topic&chartType=TIMELINE_WITH_HIGHLIGHT

https://screencast.googleplex.com/cast/NDg2NDE3MjA1ODczODY4OHwyODllMGExYS0wMQ

Note that the first 3 charts are for each of the StatVars passed in. The remaining come from the 3 topics identified for the statVars
dc/topic/Income, dc/topic/Agriculture, dc/topic/Demographics

Next Up: Allow injecting provenance, unit and timerange. more Code Cleanup. Testing.


static/js/utils/url_utils.ts
```diff
@@ -81,3 +85,72 @@ export function updateHash(params: Record<string, string | string[]>): void {
...
+  }
+  return input;
+}
+
+/**
+ * Extracts all the const parameters from the URL hash parameters within the Explore.
+ * @param hashParams parameters.
+ * @returns string array of all parrams.
+ */
+export function extractUrlHashParams(
```
dwnoble, line 106: Would it be better to return a typed object here instead of a list? Then you could update the calling code to:

```
const {
  query,
  place,
  topic,
  statVar,
  dc,
  idx,
  disableExploreMore,
  detector,
  testMode,
  i18n,
  includeStopWords,
  defaultPlace,
  mode,
  reranker,
  maxTopics,
  maxTopicSvs,
  maxCharts,
  chartType,
} = extractUrlHashParams(hashParams);
```

static/js/apps/explore/app.tsx
```diff
@@ -377,26 +377,66 @@ export function App(props: AppProps): ReactElement {
...
+      let data = {};
+      if (statVar) {
+        let statVars = [];
+        if (statVar.includes(URL_DELIM)) {
+          statVars = toApiList(statVar);
+        } else {
+          statVars = [statVar];
+        }
+
+        data = await axios.get("/api/node/propvals", {
```
dwnoble, line 390: Could you add this call to the web client library? Around here: https://github.com/datacommonsorg/website/blob/d10ba51e349efd88703434eb23a598a2c5f783ec/packages/client/src/data_commons_web_client.ts#L53

static/js/apps/explore/app.tsx
```diff
@@ -302,42 +295,49 @@ export function App(props: AppProps): ReactElement {
     );
   }
 
-  function handleHashChange(): void {
+  async function handleHashChange(): Promise<void> {
```
dwnoble, line 298: This function is getting pretty big- do you see a good way to refactor it any more than you've already done?

Also- could you add some docstrings to the function while you're at it?

static/js/utils/url_utils.ts
gmechali: done.

static/js/apps/explore/app.tsx
gmechali: Done.

static/js/apps/explore/app.tsx
gmechali: I added the logic for the new promise for the highlight, but then cleaned up some of the existing one...so maybe a net neutral in terms of complexity :)

WDYT?

dwnoble (approved):

static/js/apps/explore/app.tsx
dwnoble: that looks good to me!

PR #5116, opened 2025-04-12
Removed legacy place page

dwnoble:
* Removed legacy place page html and css
* Removed legacy place page typescript files, including legacy chart types
* Removed legacy place page python api 
* Removed legacy place page python utilities and the `/api/landingpage/data` api endpoint
* Updated "clicker" cache warming tool to use the new place page `api/dev-place/charts` endpoint
* Removed legacy place page SEO experiment


gmechali (approved):
No more old place pageeee!!!!

PR #5117, opened 2025-04-14
Tooltips

nick-nlb:
## Description

The goal of this PR is to implement a comprehensive and flexible tooltip / popover component, that meets the following qualifications:

- both simple text and complex HTML (i.e. React Nodes) for the trigger.
- both simple text and complex HTML for the content.
- the tooltips should work on both desktop and mobile (and therefore, should also have a “popover” mode, as hover is not a relevant concept in touch).
- an option to use popovers for desktop as well, thus making this both a tooltip and popover component.
- position can be determined by a prop, but it must flip intelligently when there is not enough room to display.
- the user should be able to interact with the content of the tooltip (i.e., the tooltip can supply links, buttons). Therefore, the user must be able to travel from the trigger to the tooltip without the tooltip closing.
- the tooltips should be fully keyboard accessible, opening up on tab and allowing tabbing inside interactive elements inside the tooltip.
- the tooltips should be easy to use and interact with.
- Only one tooltip should be open at a time.
- the tooltips should be highly customizable, with most features being overridable via props (i.e., animations can be turned off, arrows can be added, etc).

## Testing

An existing tooltip has been converted to this new component on the dev_place_overview page, and can be viewed here:

http://localhost:8080/place/country/NZL?q=new+zealand

A more comprehensive set of demonstrations can be found in a temporary section of the home page that will be removed before merging. (see comment below).

http://localhost:8080

### Documentation

Instructions and examples of how to use the tooltip are provided in comments at the top of the component. Additionally, each prop is documented in the interface. As the component itself is quite extensive and has multiple moving parts, internal comments are provided to annotate and explain functionality.

### Demo Panel

We created a temporary section on the home page to demonstrate some of the various options and features of the tooltips. This will be removed before we merge in the page. Perhaps we could look at something like Storybook in the future for this type of UI demonstration.

### Libraries

The tooltip uses Floating-UI for placement (this will be installed when running `run_npm.sh`).

### Tooltip and Trigger Content

With the goal in mind of being flexible, the tooltip can take any component as a child, with the child acting as the trigger. In other words, you can use a string, an icon, a div, a button, a link, etc as the child. Likewise, the tooltip content itself can either be a string, or any arbitrary 

### Mouse Interactivity

We want this tooltip component to be easy and enjoyable to use. Given that the tooltips can be interactive, we made sure that it is easy for the user to travel between the trigger and the tooltip without it closing.

To facilitate this, we created a “bridge zone” between the trigger and the tooltip, where the tooltip does not close. To further make the tooltip easy to use, we created lateral buffer zones (i.e., if the tooltip is above the trigger, the buffer is to the left and right of the trigger) where the tooltip does not close. This allows the user to comfortably travel from the trigger to the tooltip without having to take great care where the mouse goes.

### Keyboard Interactivity

To make the tooltip keyboard accessible, we have implemented functionality that manages the tooltip state when the user tabs onto a tooltip element, and allows the user to tab into the tooltip and back out again.

Popovers can also be opened and closed by keyboard.

### Message Bus

Because of the buffering and bridging, it is possible (but unlikely) for a second tooltip to open while we are still in the buffer zone of the first tooltip. This could happen if two tooltips are in close proximity. To solve this, we have implemented a message bus that allows the most recently opened tooltip to inform other tooltips that they must now close.

### Options

There are numerous options to customize how the tooltip works (such as a popover mode, arrows and placement, among others). These are document in the prop interface.

## Questions

Should the arrow be on by default?

**With arrow**

Screenshot 2025-04-14 at 6 54 36 AM

**Without arrow**

Screenshot 2025-04-14 at 7 00 02 AM

### DIFF
--- a/static/js/apps/homepage/app.tsx
+++ b/static/js/apps/homepage/app.tsx
@@ -34,7 +34,6 @@ import {
 } from "../../shared/types/homepage";
 import theme from "../../theme/theme";
 import { BuildYourOwn } from "./components/build_your_own";
-import { DemoTooltips } from "./components/demo_tooltips";
 import { HomeHero } from "./components/home_hero";
 import { SampleQuestions } from "./components/sample_questions";
 import { Tools } from "./components/tools";
@@ -75,11 +74,6 @@ export function App({
 
       <Section>
         <SampleQuestions sampleQuestions={sampleQuestions} />
-      </Section>
-
-      {/* TODO (nick-next): Remove this before PR */}
-      <Section variant="small">
-        <DemoTooltips />
       </Section>
 
       <Section variant="small">
--- a/static/js/components/elements/tooltip/demo_tooltips.tsx
+++ b/static/js/components/elements/tooltip/demo_tooltips.tsx
@@ -18,16 +18,16 @@
  * A component that demos the tooltips.
  */
 
-// TODO (nick-next): Remove this file before PR is merged into master.
+// TODO (nick-next): Move this to Storybook when ready.
 
 /** @jsxImportSource @emotion/react */
 
 import { css, useTheme } from "@emotion/react";
 import React, { ReactElement } from "react";
 
-import { InfoSpark } from "../../../components/elements/icons/info_spark";
-import { ScatterPlot } from "../../../components/elements/icons/scatter_plot";
-import { Tooltip } from "../../../components/elements/tooltip/tooltip";
+import { InfoSpark } from "../icons/info_spark";
+import { ScatterPlot } from "../icons/scatter_plot";
+import { Tooltip } from "./tooltip";
 
 export const DemoTooltips = (): ReactElement => {
   const theme = useTheme();


static/js/components/elements/tooltip/tooltip.tsx
```diff
@@ -0,0 +1,1157 @@
...
+      </WrapperComponent>
+    );
+  }
+}
+
+/**
+ * The primary exported Tooltip component. This implements the core
+ * tooltip functionality and renders the trigger and tooltip on the page.
+ */
+export const Tooltip = ({
```
dwnoble, line 657: Is there a good way to override the styles of the tooltip?

static/js/components/elements/tooltip/tooltip.tsx
```diff
@@ -0,0 +1,1157 @@
...
+            $showArrow={showArrow}
+            style={{
+              position: strategy,
+              top: y ?? 0,
+              left: x ?? 0,
+            }}
+          >
+            {title}
+
+            {(popoverMode || openAsPopover) && (
```
dwnoble, line 1137: I'm seeing the close button flash in & out on chrome's mobile browser mode - can you take a look? 
[Popover.webm](https://github.com/user-attachments/assets/20756f56-39d5-4851-9d38-9df57c314690)

dwnoble (commented):
Thank you! this looks quite comprehensive

static/js/components/elements/tooltip/tooltip.tsx
dwnoble: Looks great-thanks!

PR #5122, opened 2025-04-23
Removed "dev" references in place page

dwnoble:
- Renamed files with "dev_place" in the name to just "place"
- Renamed website API endpoints from `/api/dev-place/*` to `/api/place/*`


gmechali (approved):
Lint failing, but otherwise lgtm

PR #5123, opened 2025-04-23
Custom dc place page test fix

dwnoble:
A custom dc place page test was failing because it was trying to interact with the top NL search bar, which isn't present in custom dc. I moved this test out of shared and into the tests (non-custom-dc) folder


clincoln8 (approved):
Thanks!

gmechali (approved):

PR #5124, opened 2025-04-24
Refactor, deflake, fix some webdriver tests

hqpho:
- Move place explorer i18n search bar test out of shared since Custom DC doesn't have a search bar on the place page
- Consolidate some UI interactions in complex webdriver tests into helpers that wait for elements as needed


server/webdriver/shared.py
```diff
@@ -60,11 +65,8 @@ def wait_for_loading(driver):
 
 def click_sv_group(driver, svg_name):
   """In the stat var widget, click on the stat var group titled svg_name."""
-  sv_groups = driver.find_elements(By.CLASS_NAME, 'node-title')
-  for group in sv_groups:
-    if svg_name in group.text:
-      group.click()
-      break
+  xpath_selector = f"//span[contains(@class, 'Collapsible__trigger') and .//div[contains(@class, 'node-title')]//span[contains(@class, 'title')][contains(text(), '{svg_name}')]]"
```
gmechali, line 68: Is there a better way to find the element? This is kind of unreadable :)
You can definitely fetch the list of elements with the class 'title' as
sv_groups.find_elems(By.CLASS_NAME, 'title').
Then iterate through those and check if it contains the text you need?

Or I'd say at least keep
sv_groups = driver.find_elements(By.CLASS_NAME, 'node-title')
then use the xpath from sv_groups so it's far shorter and a bit more readable :)

server/webdriver/shared.py
```diff
@@ -149,3 +151,59 @@ def assert_topics(self, driver, path_to_topics, classname, expected_topics):
   # Iterate through the elements and assert their text content
   for item, expected_text in zip(item_list_items, expected_topics):
     self.assertEqual(item.text, expected_text)
+
+
+def search_for_california_counties(self, driver):
```
gmechali, line 156: TBH I dont love adding such a specific helper method. I think the helpers should be generic where we can.
I would argue if you want more helper method, it should be on for a generic search that can hold:

"""
search_box_input = find_elem(driver, by=By.ID, value='ac')
  search_box_input.send_keys(PLACE_SEARCH_CA)

  # Wait until there is at least one result in autocomplete results.
  self.assertIsNotNone(wait_elem(driver, value='pac-item'))
"""

The place type selection, I would leave in the test, and same for the button clicking.
For the _vis version, since the attributes are located differently, I would also just pass in the By, and value as arguments so one search method can handle it all

server/webdriver/shared.py
hqpho: I moved away from the iterating over elements approach to avoid element references becoming stale before click. Simplified the xpath, wdyt?

server/webdriver/shared.py
hqpho: I get it on principle but in practice there are very complicated tests that share logic that had subtly diverged, so I think these helpers are useful. Also updated to have the search term and place type passed in.

gmechali (approved):

dwnoble (approved):

PR #5136, opened 2025-04-29
[Production Excellence] Restart the Kubernetes pod for production feature flag updates.

gmechali:
So far we excluded production to be safe - but this has been running well since January. As discussed during team onsite, we are now automating the rollout for the prod push


clincoln8 (approved):
thanks!

PR #5138, opened 2025-04-29
[Attribution] Allows facet selection from URL parameters in the Explore page for the Highlight result

gmechali:
Adds facet selection to chart injection. You can pass in either of Unit, importName and measurementMethod as a URL parameter to force the selection of facets matching those parameters.
For now, this only adds support for Bar Tiles , Line Tiles and Highlight Tiles. 

Upcoming: Adding webdriver tests. Removing non-exact charts from the highlight component. Improving the facet selection & UX.

See France GDP in [Euros](http://localhost:8080/explore#sv=Amount_EconomicActivity_GrossDomesticProduction_Nominal&p=country%2FFRA&client=ui_related_topic&chartType=TIMELINE_WITH_HIGHLIGHT&unit=Euro) ([screenshot](https://screenshot.googleplex.com/3ehzccodbV7QjV3)), France GDP in [Dollars](http://localhost:8080/explore#sv=Amount_EconomicActivity_GrossDomesticProduction_Nominal&p=country%2FFRA&client=ui_related_topic&chartType=TIMELINE_WITH_HIGHLIGHT&unit=USDollar)([screenshot](https://screenshot.googleplex.com/7NthnRsymBkfdKM))

Italy & France Population comparison as Timeline chart from [default source](http://localhost:8080/explore#sv=Count_Person&p=country%2FFRA___country%2FITA&client=ui_related_topic&chartType=TIMELINE_WITH_HIGHLIGHT) ([screenshot](https://screenshot.googleplex.com/XJ8nwau8HdRVRSR)), which is WDI.
Now see when we force it to pick [WikidataPopulation](http://localhost:8080/explore#sv=Count_Person&p=country%2FFRA___country%2FITA&client=ui_related_topic&chartType=TIMELINE_WITH_HIGHLIGHT&importName=WikidataPopulation) ([screenshot](https://screenshot.googleplex.com/5WpkcydeCg9ycRs))

Italy and France Population comparison as bar chart selecting [Euros](http://localhost:8080/explore#sv=Amount_EconomicActivity_GrossDomesticProduction_Nominal&p=country%2FFRA___country%2FITA&client=ui_related_topic&chartType=BAR_CHART&unit=Euro) ([screenshot](https://screenshot.googleplex.com/9tDXV8AbBeUih7G))


server/routes/shared_api/observation/point.py
```diff
@@ -26,12 +26,63 @@
 bp = Blueprint('point', __name__, url_prefix='/api/observations/point')
 
 
+def _filter_point_for_facets(point_data, facet_ids):
```
dwnoble, line 29: Nit: add typing

server/routes/shared_api/observation/point.py
```diff
@@ -41,9 +92,19 @@ def point():
...
-  # Fetch observations from a specific date or date = 'LATEST'
-  return fetch.point_core(entities, variables, date, False)
+                                  all_facets=False,
+                                  facet_ids=facet_id)
+
+  # If facet_ids are provided, we need to filter the data after fetching
+  # the point data. This is because the fetch.point_core function does not
+  # support filtering by facet_ids directly.
+  all_facets = True if facet_id else False
+  point_data = fetch.point_core(entities, variables, date, all_facets)
```
dwnoble, line 102: does passing facet_id here, then along to `obs_point_within` do the same filtering by facet id?

static/js/apps/explore/app.tsx
```diff
@@ -280,18 +282,22 @@ export function App(props: AppProps): ReactElement {
...
+          .map((place) => place.name)
+          .join(", ")
+          .replace(/, ([^,]*)$/, " and $1")}`;
         if (
           pageMetadata.mainTopics.length == 2 &&
           pageMetadata.mainTopics[0].name &&
           pageMetadata.mainTopics[1].name
         ) {
-          const q = `${pageMetadata.mainTopics[0].name} vs. ${pageMetadata.mainTopics[1].name} in ${pageMetadata.place.name}`;
+          const q = `${pageMetadata.mainTopics[0].name} vs. ${pageMetadata.mainTopics[1].name} ${inPlaces}`;
```
dwnoble, line 296: Can you extract this to the i18n file?

server/routes/shared_api/observation/point.py
```diff
@@ -26,12 +26,63 @@
 bp = Blueprint('point', __name__, url_prefix='/api/observations/point')
 
 
+def _filter_point_for_facets(point_data, facet_ids):
+  """Filter the point data to only include the specified facets.
+  Args:
+    point_data: The point data to filter.
+    facet_ids: The list of facet IDs to include in the filtered data.
```
beets, line 33: could you document what type of ID's are expected here, keeping in mind that we have some implicit IDs now, that we will make proper DCIDs in the near future.

static/js/apps/explore/highlight_result.tsx
```diff
@@ -13,22 +14,29 @@ const PAGE_ID = "highlight-result";
...
-        id={PAGE_ID}
-        place={props.highlightPageMetadata.place}
-        pageConfig={trimCategory(
-          props.highlightPageMetadata.pageConfig,
-          props.maxBlock
-        )}
-        showExploreMore={false}
-      />
+      <div className="highlight-result-title">
+        <h2>Highlight chart!</h2>
```
beets, line 26: please remove before submitting

static/js/constants/app/explore_constants.ts
```diff
@@ -48,6 +48,11 @@ export const URL_HASH_PARAMS = {
   MAX_TOPIC_SVS: "maxTopicSvs",
   MAX_CHARTS: "maxCharts",
   CHART_TYPE: "chartType",
+  IMPORT_NAME: "importName",
+  MEASUREMENT_METHOD: "measurementMethod",
+  OBSERVATION_PERIOD: "observationPeriod",
+  SCALING_FACTOR: "scalingFactor",
```
beets, line 54: should we use some shorter param keys for these? it's easier to make mistakes across apps when they get long  (or during manual testing)

static/js/utils/data_fetch_utils.ts
```diff
@@ -119,6 +120,45 @@ function getProcessedPointResponse(
...
+): Promise<string[] | null> {
+  if (!highlightFacet) {
+    return null;
+  }
+  const facetsResponse = await getFacets(apiRoot, entities, variables);
+  for (const svDcid of Object.keys(facetsResponse)) {
+    const facets = facetsResponse[svDcid];
+    for (const [facetId, f] of Object.entries(facets)) {
+      if (
+        (!_.isEmpty(highlightFacet.importName) &&
```
beets, line 147: should we also do matches on the other facet properties such as scaling factor?

static/js/constants/app/explore_constants.ts
beets: could we also add some additional documentation here, that these are for facet selection etc?

beets (commented):
thanks for adding this! 

overall, it would be great to start documenting what the URL format is for these. what get's a little lost is how to combine multiple facets, if it's possible.

server/routes/shared_api/observation/point.py
gmechali: this doesnt' actually do any filtering. It just takes the boolean for whether you want all facets or not. If not, it just picks a default facet.
I added the filtering on the lines right after this.

static/js/apps/explore/app.tsx
```diff
@@ -280,18 +282,22 @@ export function App(props: AppProps): ReactElement {
       if (
         !userQuery &&
         !_.isEmpty(pageMetadata.mainTopics) &&
-        pageMetadata.place.name
+        pageMetadata.places.length > 0
       ) {
+        const inPlaces = `in ${pageMetadata.places
```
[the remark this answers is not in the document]
gmechali: Done.

static/js/apps/explore/app.tsx
gmechali: done.

static/js/apps/explore/highlight_result.tsx
gmechali: Yes this was just temporary :)
I will add some formatting around it but will do it later!

static/js/constants/app/explore_constants.ts
gmechali: Added comment that it's for facet selection. I do agree about shorter being less error prone but then it also becomes harder to understand acronys.

Would "import", "msrmnt", "obs_per" be what you had in mind? or more like "imp", "mm", "op"?

static/js/utils/data_fetch_utils.ts
gmechali: Absolutely :)
Leaving that for a follow up since the scaling factor wasnt in the response for the ones I tested with. I wasnt sure if it was because it's left out of the facet response, or the data I tested with just didnt have a scaling factor. 

I am planning on adding the scaling factor and observation period in a follow up!

server/routes/shared_api/observation/point.py
gmechali: These are the facet IDs you're referring to, those implicit ones.
This facet_id isn't what's passed in the URL params, but within the data_fetch_utils, we essentially convert from the ImportName, Unit, MM, ..., into the facet ID that matches by calling the API to fetch all facets.

I'm not sure how to describe it other than facet_ids. Can you elaborate on what might be a more clear description?

dwnoble (approved):

static/js/constants/app/explore_constants.ts
beets: i like the latter set: "imp", "mm", "op"
in the past, we have come close to URL limits, and i've made many mispellings myself when i've crafted these url's for testing.

PR #5142, opened 2025-04-30
[Attribution] Re-adds ability to inject chart to highlight, with the query fix. 

gmechali:
https://github.com/datacommonsorg/website/pull/5114 was reverted in https://github.com/datacommonsorg/website/pull/5141 following a bug being noticed where the query was correct processed but the results were no longer displayed. This was a bug tied to the changes in processFulfillData which is now resolved. I made the SetPageMetadata parameter non-optional to ensure it doesn't happen again.

This PR uses the URL Hash Params to add the ability to pass in multiple places, and pass in one statistical variable.
We inject the chart for the one stat var, and issue a second call to fetch the Topic of said SV, then issue another fulfill call to with the associated topic.

Adds support for multiple StatVars, and multiple Topics on the fulfill call.
Does a very basic merging of the responses.

In this state, here's what the page looks like, while passing in:

Rhode Island and Arizona
Median_Income_Household, Area_Farm, Count_Person
TIMELINE_WITH_HIGHLIGHT for chartType
http://localhost:8080/explore/#sv=Median_Income_Household___Area_Farm___Count_Person&p=geoId%2F48___geoId%2F06&client=ui_related_topic&chartType=TIMELINE_WITH_HIGHLIGHT

https://screencast.googleplex.com/cast/NDg2NDE3MjA1ODczODY4OHwyODllMGExYS0wMQ

Note that the first 3 charts are for each of the StatVars passed in. The remaining come from the 3 topics identified for the statVars
dc/topic/Income, dc/topic/Agriculture, dc/topic/Demographics

Next Up: Allow injecting provenance, unit and timerange. more Code Cleanup. Testing.

### DIFF
--- a/static/js/apps/explore/app.tsx
+++ b/static/js/apps/explore/app.tsx
@@ -20,6 +20,7 @@
 
 import { ThemeProvider } from "@emotion/react";
 import axios from "axios";
+import { url } from "inspector";
 import _ from "lodash";
 import queryString from "query-string";
 import React, { ReactElement, useEffect, useRef, useState } from "react";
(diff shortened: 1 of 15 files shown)


static/js/apps/explore/app.tsx
```diff
@@ -352,23 +355,23 @@ export function App(props: AppProps): ReactElement {
...
+        urlHashParams.mode,
+        urlHashParams.reranker,
+        urlHashParams.includeStopWords,
+        urlHashParams.maxTopics,
+        urlHashParams.maxTopicSvs,
+        urlHashParams.maxCharts
       )
         .then((resp) => {
-          processFulfillData(resp, query);
+          processFulfillData(resp, setPageMetadata, query);
```
gmechali, line 374: FYI - the bug was because I had omitted this change. I had made that param optional but it should not be. Fixed now.

dwnoble (approved):

PR #5150, opened 2025-05-06
Metadata Dialog II

nick-nlb:
## PR Background

This PR re-applies the changes from https://github.com/datacommonsorg/website/pull/5147 to fix a TypeScript compilation error in our server (ts-node) environment by importing the type augmentations. Further details are at the bottom of this PR’s description.

This follow-up PR also addresses the need to wrap the web components (i.e., `<datacommons-bar>` with the Emotion theme and cache provider.

## Description

This PR implements the new metadata modal for charts. The new metadata modal content is substantially different than before, where only a list of included stat vars DCIDs was provided. The metadata now retrieved and displayed contains information from the details of each stat var, the details of each source as well as specific details pulled from the knowledge graph via their DCIDs. Because of the complexity of the information being pulled, this will be outlined in detail below.

### Required Information.

Before this PR, only the stat vars and the source URLs were being passed up from each chart to `<TileSources> `component (where the metadata modal is triggered). This is because the old metadata modal only needed the stat var list, and `<TileSources>` then displayed those source URLs in a concatenated string. 

However, in order to retrieve the information now displayed in the new metadata modal, we need: the stat vars, the sources themselves (the dcids of the sources) and a map of which sources are used by which stat var.

This information is available at each chart, but was not being collated and sent up to other components. So the first task of this PR was to retrieve all of that information from each chart, and send it up to where the metadata modals can use it. This accounts for changes in files such as `bar_tile.tsx`, `line_tile.tsx`, and `chart_tile.tsx`, among others. These changes are all to pass the needed facet information (which gives us the sources) to the metadata dialog.

A note that we did apply this to most of the charts, but there were some charts we could see and therefore verify the functionality of within the system. If a chart does not pass the required metadata information up, it will default to use the old stat-var-only metadata modal.

### Refactoring

Because of the increased complexity of these components, the TileSources from `tile_utils` has been moved into its own components and metadata functionality is now in its own location. While working in tile_utils.tsx, we also updated some of the code to remove TypeScript/ESLint warnings. We’ve done this in a few files that we touched.

### The API Calls

The API calls (located in `<TileMetadataModal />` are documented, but to outline the process here:

1. First we retrieve the full stat var names (via the dcids that we have). This was being done in the old modal as well.
2. We then make the necessary calls to get the full category names for each stat vars (i.e., “Demographics”).
3. We then pull the full stat var information for each stat var (that gives us details for periodicity, dates, measurement methods, etc for particular stat var / source combinations)
4. We then pull the full source (provenance) data, in order to get things like the source name and license type.
5. Finally we look up any measurement methods we might have, converting the DCIDs to the description.

This all happens quickly, but in a future PR we may look at combining these calls into a single call. 

### The Metadata Content

The metadata content is then rendered with the labels and field name translated.

#### Periodicity (Publication Cadence).

Periodicity comes as an ISO 8106 duration (i.e. “P1D” for 1 day). To facilitate the conversion of this to natural language, we built a conversion function (which converts it into text that makes sense for the concept of cadence) and we created a set of translation methods (located in `i18n_periodicity_messages` to facilitate translation of these to other languages. This leverages ICU message formats for handling plurals and combinations. These provide a flexible way for translators to convert strings like this into other languages where pluralization is very different. This could be something we use in other places as well.

We also provided a test file for the conversion function (currently only testing the English results).

#### Metadata fields

To help understand where the metadata information comes from, attached is a screenshot of an example metadata field, and corresponding screenshots mapping where the data comes from (with annotations below where relevant).

metadata-example
stat-var
source

3* The url is taken from the `url` attribute of the provenance. Note there are other choices, such as `sourceDataUrl` which is sometimes the same (as in OECDRegionalDemography_LifeExpectancy) but sometimes different (as in EurostatData_Fertility, where the `sourceDataUrl` is a direct download link.)

5* The secondary name of the source is not found in the screenshot, but is taken from "isPartOf[0]->name".

10* This comes directly as a DCID and so is looked up, with the description being used.

12* This is looked up. We take the absoluteName of the last item in the stat var's path (which will be "Demographics", "Economics", etc.

### Citation

The citation is a single, united citation for all sources. The text "with minor processing by Data Commons" is appended to the end of the citation.

The action of copying is indicated to the user by swapping out the icon. We can consider adding in a toast/snackbar component, (such as the ones used in Gmail) to give further feedback to the user.

## Codacy

Codacy produced the following note:

```
expected parameter: 'omission' to have a typedef
```

However adding in the type definition created an opposite linting error because the type definition can be inferred.

Codacy also produced warnings because of the use of `any`. These will not be simple to avoid because they parallel existing use of `any` in the already existing parent function. We could save a fix to this for a separate PR.

## TS-Node Compilation

Until now, all components that import Emotion have been used solely on the frontend. With the application of the dialog component to the chart tiles, `server.ts` needs to import Emotion-based components as part of its import graph. However, unlike the static front-end, the `server.ts` was not including type augmentations, causing the TypeScript to fail to compile. We should note that this is not specific to Emotion. Any augmentation of types in an area of the application touched by `server.ts` will need to be included. In our case, the augmentation of react-inert also needed to be included (or it would also cause a compile error on triggering `server.ts`.

The solution we implemented was to add a typing import file and include it at the top of `server.ts`. This typing import file includes the typing augmentations as compiler directives.


dwnoble (commented):
Thanks for the fixes @nick-next ! I found a few things on this second pass- let me know what you think

dwnoble (approved):
Thanks @nick-next & @pablonoel !

PR #5151, opened 2025-05-06
Update data_sources.json

premr:
Removed the incorrect trailing "[" in front of Kenya CEnsus on Data Sources page


kmoscoe:
I believe this file is auto-generated from g3 source. Have you updated that
first?

beets:
sending a CL for that..

PR #5153, opened 2025-05-06
CDC Autopush Webdriver test fix.

gmechali:
Use the Base component's title string to differentiate between Custom & Base DC.


clincoln8 (approved):
Thanks for the quick fix!!

PR #5155, opened 2025-05-12
[WIP] Add usage metrics for StatVar search

alyssasuiyiguo:


static/js/shared/ga_events.ts
```diff
@@ -206,6 +206,20 @@ export const GA_EVENT_AUTOCOMPLETE_SELECTION_REDIRECTS_TO_PLACE =
  */
 export const GA_EVENT_AUTOCOMPLETE_TRIGGERED = "autocomplete_trigger";
 
+/**
+ * Triggered on start of StatVar search.
+ * Parameters: None
+ */
+export const GA_EVENT_STATVAR_SEARCH_START = "statvar_search_start";
```
gmechali, line 213: nit: could you use the word "triggered" here instead of "start" to match the naming for autocomplete.

static/js/stat_var_hierarchy/stat_var_search.tsx
```diff
@@ -205,6 +209,11 @@ export class StatVarHierarchySearch extends React.Component<
     if (query === "") {
       this.props.onSelectionChange("");
     }
+    // When the user is starting a new search (i.e. previous query is empty),
+    // trigger Google Analytics event
+    if (this.state.query === "") {
```
gmechali, line 214: We should also check that the new query is not empty.

static/js/stat_var_hierarchy/stat_var_search.tsx
```diff
@@ -273,6 +282,11 @@ export class StatVarHierarchySearch extends React.Component<
         }
       }
     }
+    // Trigger Google Analytics event to track a successful search
+    triggerGAEvent(GA_EVENT_STATVAR_SEARCH_SELECTION, {
+      [GA_PARAM_QUERY]: this.state.query,
+      [GA_PARAM_SEARCH_SELECTION_NAME]: displayName,
```
gmechali, line 288: I actually think the DCID is going to be a better thing for us to put here, since it's guaranteed to allow us to find the right entity if we need to!

You can probably just drop "NAME" from GA_PARAM_SEARCH_SELECTION_NAME

gmechali:
Can you add more details in the description explaining how you've tested it?

gmechali:
LGTM - Thanks Alyssa!

gmechali (approved):

PR #5158, opened 2025-05-14
Added feature flag URL override

dwnoble:
Added ability to manually enable feature flags using a new `enable_feature` URL parameter.

From the updated `docs/feature_flags.md`:

Feature flags can be enabled manually on any datacommons.org page using the `enable_feature` URL parameter.

For example, to enable the "autocomplete" feature:
 - Enable: https://datacommons.org/explore?enable_feature=autocomplete

To enable both the autocomplete and metadata_modal features:
 - Disable: https://datacommons.org/explore?enable_feature=autocomplete&enable_feature=metadata_modal
 
These URL overrides take precedence over the environment-specific feature flag settings defined in server/config/feature_flag_configs/<environment>.json

### DIFF
--- a/static/js/shared/feature_flags/util.test.ts
+++ b/static/js/shared/feature_flags/util.test.ts
@@ -50,14 +50,16 @@ describe("isFeatureEnabled", () => {
       },
       writable: true,
     });
+    // Make a copy of the original FEATURE_FLAGS object
+    globalThis.FEATURE_FLAGS = { ...originalFeatureFlags };
   });
 
   afterEach(() => {
     // Restore window.location and FEATURE_FLAGS
     Object.defineProperty(window, "location", {
       value: originalLocation,
-      writable: true,
     });
+    // Restore the original FEATURE_FLAGS object
     globalThis.FEATURE_FLAGS = originalFeatureFlags;
   });
 
--- a/static/js/shared/feature_flags/util.test.ts
+++ b/static/js/shared/feature_flags/util.test.ts
@@ -1,0 +1,83 @@
+/**
+ * Copyright 2025 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { afterEach, beforeEach, describe, expect, test } from "@jest/globals";
+import {
+  DISABLE_FEATURE_FLAG_PREFIX,
+  ENABLE_FEATURE_FLAG_PREFIX,
+  isFeatureEnabled,
+} from "./util";
+
+describe("isFeatureEnabled", () => {
+  const featureName = "testFeature";
+  const enableParam = `${ENABLE_FEATURE_FLAG_PREFIX}${featureName}`;
+  const disableParam = `${DISABLE_FEATURE_FLAG_PREFIX}${featureName}`;
+
+  // Store original window.location and FEATURE_FLAGS
+  const originalLocation = window.location;
+  const originalFeatureFlags = globalThis.FEATURE_FLAGS;
+
+  beforeEach(() => {
+    // Mock window.location
+    Object.defineProperty(window, "location", {
+      value: {
+        ...originalLocation,
+        search: "",
+      },
+      writable: true,
+    });
+  });
+
+  afterEach(() => {
+    // Restore window.location and FEATURE_FLAGS
+    Object.defineProperty(window, "location", {
+      value: originalLocation,
+      writable: true,
+    });
+    globalThis.FEATURE_FLAGS = originalFeatureFlags;
+  });
+
+  test("returns true when enable param is present", () => {
+    window.location.search = `?${enableParam}`;
+    expect(isFeatureEnabled(featureName)).toBe(true);
+  });
+
+  test("returns false when disable param is present", () => {
+    window.location.search = `?${disableParam}`;
+    expect(isFeatureEnabled(featureName)).toBe(false);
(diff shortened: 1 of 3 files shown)


static/js/shared/feature_flags/util.ts
```diff
@@ -14,18 +14,45 @@
...
+ * https://datacommons.org/explore?disable_autocomplete will disable the autocomplete feature.
+ *
  * @param featureName name of feature for which we want status.
  * @returns Bool describing if the feature is enabled
  */
 export function isFeatureEnabled(featureName: string): boolean {
-  if (FEATURE_FLAGS && featureName in FEATURE_FLAGS) {
-    return FEATURE_FLAGS[featureName];
+  // Check URL params for feature flag overrides
+  const urlParams = new URLSearchParams(window.location.search);
```
gmechali, line 44: Instead of this concatenation of the flag, could we just it be in the form of
dc.org/?**feature_flag_name=enabled**

I think it's a little cleaner if we exactly match the feature flag name. You can also support disabled if you want that way.

static/js/shared/feature_flags/util.ts
```diff
@@ -14,18 +14,45 @@
...
+  const enableParam = `${ENABLE_FEATURE_FLAG_PREFIX}${featureName}`;
+  const disableParam = `${DISABLE_FEATURE_FLAG_PREFIX}${featureName}`;
+  if (urlParams.has(enableParam)) {
```
gmechali, line 47: nit: I'd just move this to another helper function for getFeatureOverride that returns the override as enabled or disabled or None. And then it ll keep this code cleaner.

const override = getFeatureOverride(featureName)
if (override != null) {
  return override.value
}

const featureFlags = getFeatureFlags();
  if (featureFlags && featureName in featureFlags) {
    return featureFlags[featureName];
  }
  return false;

static/js/shared/feature_flags/util.test.ts
```diff
@@ -0,0 +1,103 @@
...
+      writable: true,
+    });
+  });
+
+  afterEach(() => {
+    // Restore window.location and FEATURE_FLAGS
+    Object.defineProperty(window, "location", {
+      value: originalLocation,
+      writable: true,
+    });
```
gmechali, line 66: Can this just all be in the beforeEach to reset it?

gmechali (commented):
Big fan of this testing :)

static/js/shared/feature_flags/util.ts
```diff
@@ -14,18 +14,45 @@
...
 
-export const FEATURE_FLAGS = globalThis.FEATURE_FLAGS;
+export function getFeatureFlags() {
+  return globalThis.FEATURE_FLAGS;
+}
+
 export const AUTOCOMPLETE_FEATURE_FLAG = "autocomplete";
 export const METADATA_FEATURE_FLAG = "metadata_modal";
+export const ENABLE_FEATURE_FLAG_PREFIX = "enable_";
+export const DISABLE_FEATURE_FLAG_PREFIX = "disable_";
```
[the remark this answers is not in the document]
dwnoble: That sounds right-ill remove disabling features

static/js/shared/feature_flags/util.ts
dwnoble: The thing i'm worried about with that approach is if we add a feature flag with same name as a valid URL param used by another page. Maybe an alternative approach could be setting the feature flag name as the value to a fixed param? something like

(1) datacommons.org/?enable_feature=autocomplete&enable_feature=metadata_modal
(2) datacommons.org/?feature_flag=autocomplete&feature_flag=metadata_modal
(3) datacommons.org/?enable_features=autopush,metadata_modal

What do you think?

dwnoble: I went with `enable_feature` as the parameter name, and using an approach like #2 to enable multiple features. Happy to adjust it though!

static/js/shared/feature_flags/util.ts
```diff
@@ -14,18 +14,53 @@
...
+}
+
+/**
+ * Helper method to check if a feature is enabled with feature flags.
+ *
+ * Feature flags are set for each environment in
+ * server/config/feature_flag_configs/<environment>.json
+ *
+ * Feature flags can be overridden and enabled or disabled manually
+ * by adding ?enable_<feature_name> or ?disable_<feature_name> to the URL.
```
gmechali, line 46: nit: update this comment to reflect new state of the world.

static/js/shared/feature_flags/util.test.ts
gmechali: I guess I dont understand, why cant you just in the beforeEach have:
Object.defineProperty(window, "location", {
      value: {
        ...originalLocation,
        search: "",
      },
      writable: true,
    });
    globalThis.FEATURE_FLAGS = originalFeatureFlags;

Isnt that the resetting you need to do? So if you do it before each test, then it's enough?

gmechali (approved):

static/js/shared/feature_flags/util.ts
gmechali: That option looks good to me! Thanks for the update!

static/js/shared/feature_flags/util.ts
```diff
@@ -14,18 +14,53 @@
  * limitations under the License.
  */
 
-export const FEATURE_FLAGS = globalThis.FEATURE_FLAGS;
+export function getFeatureFlags() {
```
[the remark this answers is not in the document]
dwnoble: done

static/js/shared/feature_flags/util.ts
dwnoble: done

static/js/shared/feature_flags/util.test.ts
dwnoble: I think we need both because:
In the before:
- For the window.location, we overwrite window.location and make it writeable (by default it's read only)
- For the FEATURE_FLAGS, we make it a copy of the globalThis.FEATURE_FLAGS object (i just added this- it was missing before)

In the after, we reset window.location and globalThis.FEATURE_FLAGS to their original values so there aren't side effects in other, completely unrelated tests. The globalThis.* state and window.location states are preserved between tests

gmechali: ack thanks for the explanation

PR #5159, opened 2025-05-14
Add support for a skip_topics param to the search-vector endpoint.

keyurva:
* The `search-vector` endpoint currently returns a mix of topics and SVs. This change is to support a param that returns only the latter.
* Verified locally by bringing up the website and NL servers and making the `search-vector` call with and without `skip_topics`.


keyurva:
Thanks for the review!

PR #5161, opened 2025-05-15
[FeatureFlags] Update script for adding feature flag

gmechali:
Running the script will add the feature flag to all files except production, then sorts all the entries.

I ran it once to test it, it sorted all the entries and added my test flag. I removed the test flag and kept the sorting. 

This also updates documentation for adding a flag. And lastly, updates our test to validate that all config files have the same number of flags, such that production can have fewer flags.


clincoln8 (approved):
Thanks for this!!

PR #5165, opened 2025-05-16
[Chart Injection] Make the highlight Response only show the highlighted chart + Remove the matching chart from the non-highlight response.

gmechali:
This PR does two main things for improving the chart injection work:

For single chart injection, it would show the proper chart + a chart for the contained in places. This PR removes that part of the code. Whenever we have that chartType URL parameter, we filter out any additional contained_in logic for chart generation.

For all charts, since we duplicated the call to fulfill while restricting to a single stat var, that chart will appear in the highlight chart + the regular response. so I implemented the logic to remove any duplicate blocks from the non-highlight block.

Examples:
Before: https://screenshot.googleplex.com/BtvrNX3hHEGaBjv
After: https://screenshot.googleplex.com/9WF3EDSQLVtEZn9
Note the contained in block is gone, and the duplicate first chart is gone.

### DIFF
--- a/server/lib/nl/fulfillment/basic.py
+++ b/server/lib/nl/fulfillment/basic.py
@@ -76,6 +76,7 @@ def _populate_explore(state: PopulateState, chart_vars: ChartVars,
                       places: List[Place], chart_origin: ChartOriginType,
                       rank: int) -> bool:
   added = False
+  # TODO(gmechali): Consider making is_highlight a part of the utterance.
   is_highlight = bool(state.uttr.insight_ctx.get(params.Params.CHART_TYPE))
 
   # For peer-groups, add multi-line charts.
@@ -103,6 +104,8 @@ def _populate_explore(state: PopulateState, chart_vars: ChartVars,
     if state.place_type:
       # If this is SDG, unless user has asked for ranking, do not return!
       added_child_type_charts = False
+
+      # TODO(gmechali): Refactor this code for more explicit logic.
       if not is_highlight and not is_special_dc or state.ranking_types:
         ranking_orig = state.ranking_types
         if not state.ranking_types:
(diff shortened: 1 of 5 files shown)


server/lib/nl/fulfillment/basic.py
```diff
@@ -102,7 +104,9 @@ def _populate_explore(state: PopulateState, chart_vars: ChartVars,
     if state.place_type:
       # If this is SDG, unless user has asked for ranking, do not return!
       added_child_type_charts = False
-      if not is_special_dc or state.ranking_types:
+
+      # TODO(gmechali): Refactor this code for more explicit logic.
+      if not is_highlight and not is_special_dc or state.ranking_types:
```
dwnoble, line 109: Could you add a comment here explaining when this case is expected?

server/lib/nl/fulfillment/basic.py
```diff
@@ -76,6 +76,8 @@ def _populate_explore(state: PopulateState, chart_vars: ChartVars,
                       places: List[Place], chart_origin: ChartOriginType,
                       rank: int) -> bool:
   added = False
+  # TODO(gmechali): Consider making is_highlight a part of the utterance.
+  is_highlight = bool(state.uttr.insight_ctx.get(params.Params.CHART_TYPE))
```
dwnoble, line 80: Can you explain the logic here in a comment?

dwnoble (approved):

server/lib/nl/fulfillment/basic.py
gmechali: Done.

server/lib/nl/fulfillment/basic.py
gmechali: Done.

PR #5166, opened 2025-05-16
Add usage metrics for StatVar browsing

alyssasuiyiguo:
Adds Google Analytics events to track user browsing StatVar. 

- One event is fired when the user clicks on any toggles in the StatVar hierarchy. Note that this fires both when the toggle is opened OR closed (I thought it would be relevant to track both opening and closing the toggle since they both reflect user browsing, but lmk if you disagree with this). 
- The other change updates the existing event `GA_EVENT_TOOL_STAT_VAR_CLICK` which currently fires whenever a stat var is selected. I added a parameter here to track whether the stat var was selected directly from the hierarchy or through searching. 
   - While testing this change I also noticed that the event is currently also fired whenever the user clears the current search query ([example](http://screencast/cast/NTAyNTIyOTcwODQ2MDAzMnw2Yjg1NjAyOC1kZQ)), I believe because of [this](https://github.com/datacommonsorg/website/blob/master/static/js/stat_var_hierarchy/stat_var_search.tsx#L256) line which calls `onSelectionChange` with an empty string, which down the line ends up calling `togglePath` with `sv` as an empty string. So I also added a check to see if `sv` is non-empty before firing the event. 

Tested by using console logs & manually verifying that code is reached during expected events. Also modified an existing client test to test this change.

### DIFF
--- a/static/js/stat_var_hierarchy/stat_var_hierarchy.tsx
+++ b/static/js/stat_var_hierarchy/stat_var_hierarchy.tsx
@@ -27,7 +27,10 @@ import React from "react";
 import { Context } from "../shared/context";
 import {
   GA_EVENT_TOOL_STAT_VAR_CLICK,
+  GA_PARAM_SOURCE,
   GA_PARAM_STAT_VAR,
+  GA_VALUE_TOOL_STAT_VAR_OPTION_HIERARCHY,
+  GA_VALUE_TOOL_STAT_VAR_OPTION_SEARCH,
   triggerGAEvent,
 } from "../shared/ga_events";
 import {
@@ -344,12 +347,12 @@ export class StatVarHierarchy extends React.Component<
         searchSelectionCleared,
         expandedPath: searchSelectionCleared ? this.state.focusPath : [],
       });
-      this.togglePath(selection, path);
+      this.togglePath(selection, path, true);
     });
   }
 
   // Add or remove a stat var and its path from the state.
-  private togglePath(sv: string, path?: string[]): void {
+  private togglePath(sv: string, path?: string[], searchSelection: boolean = false): void {
     if (sv in this.state.svPath) {
       const tmp = _.cloneDeep(this.state.svPath);
       delete tmp[sv];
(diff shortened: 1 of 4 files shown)


alyssasuiyiguo:
Just had a thought - since the existing event GA_EVENT_TOOL_STAT_VAR_CLICK already tracks when users are selecting a StatVar (and with this change, also tracks whether the user selected the sv through browsing or through search), is it still relevant to keep the new event GA_EVENT_STATVAR_SEARCH_SELECTION from my last [PR](https://github.com/datacommonsorg/website/pull/5155/files#diff-36dccb9e4242a16b7c170d888927cc91c05a38d0861b8bf0a5b5b4399bb000baR269)? I'm thinking it'd be good to still track the user's query (which the `stat_var_search` component has), but I could also just pass that down to `stat_var_hierarchy` and add to the GA_EVENT_TOOL_STAT_VAR_CLICK event

static/js/stat_var_hierarchy/stat_var_hierarchy.tsx
```diff
@@ -360,9 +367,17 @@ export class StatVarHierarchy extends React.Component<
     } else {
       if (this.props.selectSV) {
         this.props.selectSV(sv);
-        triggerGAEvent(GA_EVENT_TOOL_STAT_VAR_CLICK, {
-          [GA_PARAM_STAT_VAR]: sv,
-        });
+        console.log("Stat var selected: ", sv);
+        // Trigger GA event if sv is not empty
+        if (sv) {
+          console.log("triggering GA event");
```
gmechali, line 373: nit: remove the logging, here and above.

static/js/stat_var_hierarchy/stat_var_hierarchy.tsx
```diff
@@ -360,9 +367,17 @@ export class StatVarHierarchy extends React.Component<
     } else {
       if (this.props.selectSV) {
         this.props.selectSV(sv);
-        triggerGAEvent(GA_EVENT_TOOL_STAT_VAR_CLICK, {
-          [GA_PARAM_STAT_VAR]: sv,
-        });
+        console.log("Stat var selected: ", sv);
+        // Trigger GA event if sv is not empty
```
gmechali, line 371: What's the case for when sv is empty? Did you see that ever happening?

static/js/stat_var_hierarchy/stat_var_hierarchy.tsx
```diff
@@ -344,12 +347,16 @@ export class StatVarHierarchy extends React.Component<
...
+      this.togglePath(selection, path, true);
     });
   }
 
   // Add or remove a stat var and its path from the state.
-  private togglePath(sv: string, path?: string[]): void {
+  private togglePath(
+    sv: string,
+    path?: string[],
+    searchSelection = false
```
gmechali, line 358: nit: Add typing on the parameter.
Can you name it like a boolean, so isSearchSelection

static/js/stat_var_hierarchy/stat_var_hierarchy.tsx
```diff
@@ -344,12 +347,16 @@ export class StatVarHierarchy extends React.Component<
         searchSelectionCleared,
         expandedPath: searchSelectionCleared ? this.state.focusPath : [],
       });
-      this.togglePath(selection, path);
+      this.togglePath(selection, path, true);
```
gmechali, line 350: nit: It's often preferred to have a comment inline to improve readability by explaining what these values refer to in arguments in method calls.

this.togglePath(selection, path, /* isSearchSelection=*/ true);

static/js/stat_var_hierarchy/stat_var_hierarchy.tsx
alyssasuiyiguo: Yeah, I noticed it happening whenever clearing the search bar after selecting an sv from search - I recorded a screen recording [here](http://screencast/cast/NTAyNTIyOTcwODQ2MDAzMnw2Yjg1NjAyOC1kZQ)

gmechali (approved):

PR #5172, opened 2025-05-20
Script

kmoscoe:


kmoscoe:
Ajai had asked to remove it so the services could in the background. But it had the effect of making it impossible to stop the servers with Ctrl-C, which was driving me nuts.

keyurva (approved):
Thanks Kara.

PR #5185, opened 2025-05-27
[ProductionExcellence] Mark deprecated flags as such

gmechali:
These flags all correspond to features that have been released in GA and for which we got rid of the code. 

We can remove these flags, however, they must first be removed in staging. 
https://github.com/datacommonsorg/website/pull/5184 Needs to be submitted first.
Then this current PR.
Then https://github.com/datacommonsorg/website/pull/5183

### DIFF
--- a/server/config/feature_flag_configs/production.json
+++ b/server/config/feature_flag_configs/production.json
@@ -9,24 +9,28 @@
     "name": "dev_place_experiment",
     "enabled": true,
     "owner": "gmechali",
+    "deprecated": true,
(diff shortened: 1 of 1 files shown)


gmechali:
> Should these be marked as deprecated in all environments?

I left the other environments out because Im just about to remove all the flags in the other environments in https://github.com/datacommonsorg/website/pull/5183
But I cant submit that PR until this one goes through. THat's why I didn't update the other environments. However, in a world where we just mark them as deprecated, we should have first submitted a PR marking the non-prod environments as deprecated, then a separate PR for the prod env!

PR #5188, opened 2025-05-28
Highlight Metadata Loading

nick-nlb:
## Description

This PR adds the following functionality to the highlight result metadata loading:

* The component now keeps a loading state, and feeds that loading state through to the MetadataSummary component (moved out to its own component now that we've extended this part of the functionality.
* The loading state shows a spinner and loading message (using the shared loader) and then once loaded, displays the metadata. By having the loading, we also prevent a page bump to the components below once the metadata summary has loaded.
* We were previously using the `block.description` to pass the metadata summary down through to the block. I was worried that overloading the description might interfere with actual block descriptions, so I created an extension to the block that adds `metadataSummary`. I didn't add it directly into the interface in `subject_page_proto_types.ts` because this interface represents the shape of what the API returns (and there is a plan to have that autogenerated from the API). 
* We now send the metadataSummary through using this attribute. This way, we also can clearly separate metadata from description for the purposes of the loading state.
* The fetch now no longer occurs when there is no highlightFacet (this was triggering a runtime error error - although one that did not affect the display of the page - when there was no highlight facet).

We will not display an error directly to the user if the metadata fetch fails (but there is a TODO for reporting this error through Google Analytics).

## Testing

The following link will demonstrate the loading state (and then display the metadata as before):

http://localhost:8080/explore#sv=LifeExpectancy_Person&p=country%2FITA___country/FRA&client=ui_related_topic&chartType=TIMELINE_WITH_HIGHLIGHT&unit=Euro&imp=WorldDevelopmentIndicators

This link shows a highlight result that does not have a highlight facet. This will display as before (without a loading state or any metadata and the runtime error now gone).

http://localhost:8080/explore#sv=Count_Person&p=country%2FFRA___country%2FITA&client=ui_related_topic&chartType=TIMELINE_WITH_HIGHLIGHT

### DIFF
--- a/server/tests/lib/util_test.py
+++ b/server/tests/lib/util_test.py
@@ -1796,16 +1796,14 @@ class TestFeatureFlagsTest(unittest.TestCase):
         data = json.load(f)
 
       features = [
-          item.get("name")
-          for item in data
-          if isinstance(item, dict) and "name" in item
+          item.get("name") for item in data if isinstance(item, dict) and
+          "name" in item and not item.get("deprecated", False)
       ]
       duplicate_features = [f for f in features if features.count(f) > 1]
 
       feature_flag_count = len(features)
       feature_flag_counts.append(feature_flag_count)
       self.assertEqual(len(duplicate_features), 0)
-
       if filename == "production.json":
         production_features = set(features)
       else:
--- a/static/js/apps/explore/app.tsx
+++ b/static/js/apps/explore/app.tsx
@@ -228,6 +228,7 @@ export function App(props: AppProps): ReactElement {
     /* eslint-disable-next-line */
     fulfillData: any,
     pageMetadata: SubjectPageMetadata,
+    allowRedirect: boolean,
     userQuery?: string,
     isHighlight?: boolean
   ): void {
@@ -249,7 +250,8 @@ export function App(props: AppProps): ReactElement {
       pageMetadata.pageConfig &&
       pageMetadata.pageConfig.categories
     ) {
-      isPendingRedirect = shouldSkipPlaceOverview(pageMetadata);
+      isPendingRedirect =
+        allowRedirect && shouldSkipPlaceOverview(pageMetadata);
       if (isPendingRedirect) {
         const placeDcid = pageMetadata.place.dcid;
         // If the user has a query, append it to the url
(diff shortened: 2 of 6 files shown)


server/webdriver/shared_tests/explore_test.py
```diff
@@ -95,12 +97,17 @@ def test_highlight_chart_france_italy_gdp_bar_chart(self):
...
 
-    block_description = find_elems(self.driver, value='block-desc')[0]
-    self.assertEqual(
-        block_description.text,
-        'World Bank, World Development Indicators, with minor processing by Data Commons'
+    expected_citation = (
+        "World Bank, World Development Indicators, with minor processing by Data Commons"
     )
 
+    citation_locator = (By.CLASS_NAME, "metadata-summary")
```
gmechali, line 104: The find_elem helper can help simplify this! Could you just have
citation = find_elem(self.driver, value="metadta-summary)
self.assertEqual(citation.text, expected_citation)

static/js/apps/explore/highlight_result.tsx
```diff
@@ -98,24 +117,36 @@ export function HighlightResult(props: HighlightResultProps): ReactElement {
...
+        return;
+      }
+      try {
+        setMetadataLoadingState(true);
+        const { metadata } = await doMetadataFetch(props);
         setMetadataMap(metadata);
-      });
+        setMetadataLoadingState(false);
+      } catch (err) {
+        // TODO (nick-next): Look at using GA to report this error.
```
gmechali, line 134: ultra nit: I think the error reporting should go through cloud logging more than GA :)

static/js/components/subject_page/category.tsx
```diff
@@ -37,7 +35,7 @@ import { StatVarProvider } from "./stat_var_provider";
 
 export interface CategoryPropType {
   id: string;
-  config: CategoryConfig;
+  config: ProcessedCategoryConfig;
```
gmechali, line 38: tbh I'm not sure I love the idea of the different types for SubjectPageConfig and ProcessedX + same for category and block.
Absolutely understand why you don't like my hijacking of the description though...
Could we maybe just add the optional metadata summary to the block directly, and either render the block description or the block metadata summary?

static/js/apps/explore/highlight_result.tsx
dwnoble: @nick-next we talked about logging these errors to GA in our last call, but i think Gabe is right here. we should build a path to route these errors back to GCP cloud logging so we can set up monitoring & alerting on them.

gmechali (approved):
Thanks Nick, looks great!

dwnoble (approved):
Great, thank you nick!

PR #5189, opened 2025-05-28
Metadata Modal Error State

nick-nlb:
## Description

To account for the possibility of unforeseen errors in the metadata fetching, we are now providing error feedback to the user. We catch errors thrown from `fetchMetadataModal` and use those to display an error message in the modal.

metadata_modal_error

## Testing

The easiest way to test the metadata modal error state is to artificially trigger an error by adding a line to throw one in the fetchMetadata function itself.

- Open the file `static/js/tools/shared/metadata/metadata_fetcher.ts`.
- At the top of the function `fetchMetadata` add the following line:
- `throw new Error("test");`
- Once the front-end has built, refresh and load any metadata dialog on any chart and you will see the error state.


dwnoble (commented):
Thanks Nick! Can you add a unit test for the error path?

PR #5193, opened 2025-06-02
Dataset Selector V0

nick-nlb:
## Description

This PR implements V0 of the dataset selector revamping from the [presentation](https://docs.google.com/presentation/d/1ztyZwu8b3YzU1MMAFPI3iC-Hm00CE63rPb7eP6vGZlw/edit?usp=sharing). Note that there are two variations of V0. Here we implement the first one (with the dataset selector on the left). This is slide 6 as of the creation of this PR.

To keep the PRs smaller, especially as they touch very core areas, and related tests, this first PR only implements the reworking of the selector itself (and related aspects of the chart). It does not yet include updates to the dialog.

Included are the following changes:
- The data selector and chart options were previously in a chart footer below the chart. These are now moved above the chart, as per the design, into a new ChartHeader component.
- A new button variant was created for this button
- The button displays the total number of available sources.
- The chart options (where they exist) are now displayed on a single line where space exists, with a background, as per the discussion (now that the options are at the top, stacking those options takes up a lot of space).
- The theme was updated to make the button variants in the theme match the actual button variants we now have (rather than abstract names). 
- TypeScript and lint fixes in the various components touched by this PR.
- The old Facet Selector is still used in the old visual tools as well as the download tool (see comment), and so has been renamed FacetSelectorLegacy and is mostly unchanged.

## Screenshots
chart-no-options

chart-one-option
chart-multiple-options

## Testing

To test these changes, you can any "new" version tool, such as the new Timeline, Scatterplot or Map tool and view any chart.

An example of a chart with the new header is: 

http://localhost:8080/tools/visualization#visType%3Dtimeline%26place%3DgeoId%2F06025%26placeType%3DTown%26sv%3D%7B%22dcid%22%3A%22WithdrawalRate_Water_Irrigation%22%7D___%7B%22dcid%22%3A%22WithdrawalRate_Water_Irrigation_FreshWater_GroundWater%22%7D___%7B%22dcid%22%3A%22WithdrawalRate_Water_Irrigation_SurfaceWater%22%7D

## Notes

The old Data Commons toolset was not updated as part of this PR, as those tools have a different footer that can be updated in a separate PR if desired. Therefore, we still have the old FacetSelector component, now renamed FacetSelectorLegacy.


dwnoble (approved):
Thanks Pablo & Nick!

This looks great.

Can you also add this to the 'old' visualization tool in a follow-up?

PR #5196, opened 2025-06-03
Make Gemini calls parallel for add_metadata.py

alyssasuiyiguo:
Two changes to add_metadata script which pulls additional metadata for the stat vars:
1. Makes the calls to Gemini execute in parallel rather than sequentially, which speeds up the script from taking ~80 minutes to now taking <5 minutes to run. This does mean that 37 calls to Gemini are now all made at the same time though, which may cause resource problems once we run this in cloud. 
2. Modifies JSON output format to compress all constraint properties into one field, represented as a list of strings. Ex.:
`
constraintProperties: {
   "gender": "female",
   "age": "<20"
}
`
now becomes
`
constraintProperties: ["gender: female", "age: <20"]
`
Previously, all constraintProperties were extracted to become top-level attributes, which resulted in not all attributes being indexed by Vertex AI as it has a limit of 50 columns for indexing.


tools/nl/nl_metadata/add_metadata.py
```diff
@@ -216,42 +216,80 @@ def create_generated_sentences(
...
-    for metadata in curr_batch:
-      if metadata.dcid in alt_sentences:
-        metadata.generatedSentences = alt_sentences[metadata.dcid]
-      else:
-        print(
-            f"No alternative sentences generated for DCID {metadata.dcid}. Falling back to empty list."
-        )
-        metadata.generatedSentences = []
-      metadata_with_sentences.append(metadata)
+    tasks_to_parallelize.append(
```
gmechali, line 222: nit: lets call this parallel_tasks

gmechali (approved):
Beautiful! Thanks for doing this!

PR #5211, opened 2025-06-12
Add option to generate metadata in French and Spanish

alyssasuiyiguo:
Adds the ability to generate metadata for stat vars in French and Spanish by passing in a new feature flag "-language" in the script. As part of this change I refactored some functions in the script to return the metadata as a list of dictionaries rather than a list of StatVarMetadata objects. This was done to standardize the format used regardless of whether the user was generating sentences or not, and in different languages, as sticking with StatVarMetadata objects could only support metadata generation in English, and would result in inconsistent output formats depending on whether sentences were being generated by Gemini or not.


tools/nl/nl_metadata/add_metadata.py
```diff
@@ -69,9 +68,28 @@ def extract_flag() -> argparse.Namespace:
       "Whether to generate alternative sentences for the SVs using the Gemini API.",
       type=bool,
       default=False)
+  parser.add_argument(
+    "-language",
+    help="The language to return the metadata results in. Currently supports English, French, and Spanish.",
+    choices=["English", "French", "Spanish"],
```
gmechali, line 74: nit: Can you add a TODO to support passing multiple languages in the future. No need to add support now, but just good to track it here.

tools/nl/nl_metadata/add_metadata.py
```diff
@@ -69,9 +68,28 @@ def extract_flag() -> argparse.Namespace:
...
+    default="English"
+  )
   args = parser.parse_args()
   return args
 
+def get_language_settings(target_language: str) -> str:
+  EXPORTED_SV_FILE = f"tools/nl/nl_metadata/alyssaguo_statvars_{target_language}.json"
+
+  match target_language.lower():
+    case "French":
```
gmechali, line 85: you use .lower() above, but then this isn't in lower case. So how can it ever match?

tools/nl/nl_metadata/add_metadata.py
```diff
@@ -147,59 +165,69 @@ def extract_constraint_properties(new_row: StatVarMetadata,
...
     gemini_client: genai.Client, gemini_config: types.GenerateContentConfig,
-    sv_metadata: List[StatVarMetadata]) -> Dict[str, List[str]]:
+    gemini_prompt: str, sv_metadata: List[SVMetadataDict]) -> List[SVMetadataDict]:
   """
   Calls the Gemini API to generate alternative sentences for a list of SV metadata.
-  Returns the alt sentences as a dictionary mapping DCID to the list of sentences.
+  Returns the alt sentences as a list of dictionaries.
   """
-  prompt_with_metadata = types.Part.from_text(text=(GEMINI_PROMPT +
+  MAX_RETRIES = 5
```
gmechali, line 198: nit move this to a const at the top of the file.

tools/nl/nl_metadata/add_metadata.py
```diff
@@ -69,9 +68,28 @@ def extract_flag() -> argparse.Namespace:
...
+    help="The language to return the metadata results in. Currently supports English, French, and Spanish.",
+    choices=["English", "French", "Spanish"],
+    type=str,
+    default="English"
+  )
   args = parser.parse_args()
   return args
 
+def get_language_settings(target_language: str) -> str:
+  EXPORTED_SV_FILE = f"tools/nl/nl_metadata/alyssaguo_statvars_{target_language}.json"
```
gmechali, line 82: can you keep this as a const above, but just have a EXPORTED_FILE_DIR = 
And EXPORTED_FILENAME_PREFIX = alyssaguo_statvars_

Also, might be time to stop calling it Alyssaguo :)

gmechali (approved):
Just a couple nit comments left and right but looks good! 
I haven't really reviewed the prompt. If you're getting to work with what you need great for now, we can take a closer look at it later!

You also have a few lint errors to fix, so make sure to address that before submitting!

tools/nl/nl_metadata/add_metadata.py
```diff
@@ -147,59 +168,67 @@ def extract_constraint_properties(new_row: StatVarMetadata,
...
+      # Returns a GenerateContentResponse object, where the .text field contains the output from Gemini
+      # Output is formatted as a JSON string representing a Dict mapping DCID to a list of alt sentences
+      response: types.GenerateContentResponse = await gemini_client.aio.models.generate_content(
+          model=GEMINI_MODEL, contents=model_input, config=gemini_config)
+
+      results = json.loads(response.text, strict=False)
+      return results
+    except json.JSONDecodeError as e:
+      print(
+          f"Attempt {attempt + 1} failed. Exception occurred while generating alt sentences for the batch starting at DCID {sv_metadata[0]["dcid"]}. Error decoding Gemini response into JSON: {e}"
```
dwnoble, line 218: I think this line and line 222 are causing the linter to bomb. Looks potentially to the double quotes of "dcid" inside the double quoted format string.

if i do:

```
sv_metadata_dcid = sv_metadata[0]["dcid"]
print(
          f"Attempt {attempt + 1} failed. Exception occurred while generating alt sentences for the batch starting at DCID {sv_metadata_dcid}. Error decoding Gemini response into JSON: {e}"
)
```

it seems to work.

as a long term fix we should look into upgrading our linter from yapf to something like [ruff](https://github.com/astral-sh/ruff) supports more modern python syntax

tools/nl/nl_metadata/sv_types.py
```diff
@@ -30,4 +30,39 @@ class StatVarMetadata:
   name: str | None = None
   populationType: str | None = None
   statType: str | None = None
-  constraintProperties: dict[str, str] = field(default_factory=dict)
+  constraintProperties: list[str] = field(default_factory=list)
+
+type SVMetadataDict = dict[str, str | list[str]]
```
dwnoble, line 35: Looks like this may also be causing a linter error. if i replace this line with 

```
SVMetadataDict: dict[str, str | list[str]] = {}
```

then it works.

Similar to the above, i think this is due to our linter 'yapf' being incompatible with new python syntax. @alyssasuiyiguo could you also try just upgrading yapf in the ./run_test.sh script to see if that does the trick?

PR #5214, opened 2025-06-16
Facet Selector Adjustments

nick-nlb:
## Description

This PR makes the following adjustments to the layout and copy of the facet selector:

- "Close" has become "Cancel" in the facet selector modal.
- The "action" button in the modals now appear on the right. In other words, "Update" appears to the right of "Cancel". This will be the standard going forward.
- The facet selector is now rounded and has a smaller font to better match its surroundings in the chart.

## Notes

The font size here is a special case (to match the design of the existing charts) but is otherwise not part of the current design standard. Therefore we applied it specifically to these components only rather than adding it to the theme.

### Modal buttons

Screenshot 2025-06-16 at 1 36 37 PM

### Facet selector update

Screenshot 2025-06-16 at 2 13 39 PM

### DIFF
--- a/static/css/shared/_subject_page.scss
+++ b/static/css/shared/_subject_page.scss
@@ -77,6 +77,10 @@ article.category {
   border-top: none;
 }
 
+.highlight-result-title .subtopic:last-of-type {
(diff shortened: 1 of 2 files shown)


dwnoble (approved):

PR #5220, opened 2025-06-18
Add option to save Statvar Metadata to cloud storage

alyssasuiyiguo:
Adds a function and command line flag argument to save SV metadata to Google cloud storage.


tools/nl/nl_metadata/add_metadata.py
```diff
@@ -44,6 +45,9 @@
 STAT_VAR_SHEET = "tools/nl/embeddings/input/base/sheets_svs.csv"
 EXPORTED_FILE_DIR = "tools/nl/nl_metadata/"
 EXPORTED_FILENAME_PREFIX = "sv_complete_metadata"
+GCS_PROJECT_ID = "datcom-website-dev"
+GCS_BUCKET = "gmechali_csv_testing"
+GCS_FILE_DIR = "StatVar Metadata"
```
gmechali, line 50: Since this is a directory, I expect that this needs to be in one word or it could cause some weirdness/confusion. how about just statvar_metadata (keep lowercase)

tools/nl/nl_metadata/add_metadata.py
```diff
@@ -277,12 +285,21 @@ async def batch_generate_alt_sentences(
...
   """
   Exports the SV metadata list to a JSON file.
   """
   sv_metadata_df = pd.DataFrame(sv_metadata_list)
   sv_metadata_df.to_json(exported_sv_file, orient="records", lines=True)
+  if shouldSaveToGCS:
+    gcs_client = storage.Client(project=GCS_PROJECT_ID)
+    bucket = gcs_client.bucket(GCS_BUCKET)
+    file_path = f"{GCS_FILE_DIR}/{exported_sv_file}"
+    json_data = sv_metadata_df.to_json(orient="records", lines=True)
```
gmechali, line 298: You're doing this same operation on line 293 but not saving the output. Can we just save it in json_data in line 293?

gmechali (commented):
Nice! 

You will probably need to expand this batching capability further and upload to gcs for each batch. That will require some changes the overall flow of the script. But feel free to leave that for a follow up!

tools/nl/nl_metadata/add_metadata.py
```diff
@@ -277,25 +285,37 @@ async def batch_generate_alt_sentences(
 
 
 def export_to_json(sv_metadata_list: List[dict[str, str | list[str]]],
-                   exported_sv_file: str) -> None:
+                   exported_sv_file: str, shouldSaveToGCS: bool) -> None:
   """
   Exports the SV metadata list to a JSON file.
   """
+  file_path = f"{EXPORTED_FILE_DIR}/{exported_sv_file}"
```
alyssasuiyiguo, line 292: I changed `exported_sv_file` into just the file name and had this function put together the full file path instead of earlier because otherwise the file name that was being saved to GCS was the full file path

gmechali (approved):

PR #5221, opened 2025-06-20
Dataset Selector Modal V2

nick-nlb:
## Description

This PR implements the new facet selector dialog with an enriched metadata. This corresponds to V2 of the dialog from the slide deck (located on [page 16](https://docs.google.com/presentation/d/1ztyZwu8b3YzU1MMAFPI3iC-Hm00CE63rPb7eP6vGZlw/edit?pli=1&slide=id.g347cc273f35_0_120#slide=id.g347cc273f35_0_120) of the slide-deck).

Because each chart in each vis tool collects and prepares the facets to be sent into the facet selector independently (and in different ways across the different charts and vis tools), this PR touches all the different charts across the two versions of the vis tools. However, we made efforts to isolate the changes as much as possible to the aspects of each chart that supplies the facet selector (in other words, this PR should not affect data sent into the charts themselves).

### Architectural of Facet Selector Promises

Before this PR, the facet selector component took in a promise, rather than a list of facets, and then resolved the promise itself. The old vis tools and download tool passed the facet list in as an already resolved promise (with an empty facet list before the fetch had occurred). The new vis tools handled in a promise that was not yet resolved.

These variations complicate handling of loading states. A note that this was always an issue; however it wasn’t noticeable until we introduced the number of facet counts into the button, i.e., `Select a dataset [3]`. The old facet selector just displayed `Edit Sources` even as the facets were still loading, effectively relying on the loading to have finished by the time the user clicked the button. 

The new facet selector needs to know when it is loading so that we only display the button along with the facet count once loading is complete. This is more robust in general.

To do this we wanted to standardize how the information was transmitted and convert it to a more common React pattern. The facet selector now take a list and loading state rather than a promise.

We introduced a custom hook to resolve those promises in the new vis tools. The older tools were structured in such a way that this was not necessary.

### Measuring methods

As discussed, the measuring methods display without a label (i.e., no "Measuring method: " if we have a full description of the measuring method). If we 

### Before

before

### After

after 3

## Feature Flag

The functionality that displays the V2 (enriched-metadata) facet selector dialog is hidden behind the same feature flag that mediates the display of the new rich metadata modal. 

Once that flag is removed and V2 because live, we will remove the old facet selector entirely (this marked as a TODO). In the meantime, the new facet selector is called FacetSelectorRich, the old one FacetSelectorSimple, and FacetSelector uses the feature flag to determine which to display.

## Tests

Because the feature flag causes the facet selector modal to render differently in cdc versus standard charts during testing, the `select_source` functionality in the shared test suite is updated to take multiple possible options to search for, allowing us to provide search options for both the old and the new versions of the selector at the same time in the test.


dwnoble (commented):

The interface looks great- ill take a closer look at the code changes over the next couple of days.

In the meantime, @miss-o-soup some UX questions for you:

(1) When i first opened the metadata modal, the grayed out text made it look like the radio button was disabled. what do you think?
(2) should we hide the 'best available' option if there's only one dataset source
7HDHvkJYEvaCbUA
8DFMrMsUmMpt8fo

miss-o-soup:
Thank you for the notes Dan.

*1)* I agree with you that we need to address the "Best available" option
for dataset sources. While it might seem tempting to hide this option when
only one dataset is available, doing so would only mask the underlying
issue rather than resolve it.

If we hide the "Best available" option in such cases, it creates a
usability problem: there's nothing for the user to actually select, making
the entire selection modal pointless. I've tested this by hiding the "Best
available" option when only one source exists, and it indeed renders the
dataset selection dialog without purpose.

Let's discuss this internally and resolve the root cause of the problem
before adding another UI patch.
[image: BKXtJdj8bzaC7b7.png]

2) To alleviate the disable-impression that Dan mentioned, could we:

a) Increase (a bit) the color contrast of the light gray text? We could
apply this globally to the component itself. I was thinking something
around #666a6c, or something similar that you might have already defined in
your new stylesheet. b) Would it be possible to align the radio button to
the first line of the text vs the middle section? I know this is not what
material does, but since our options are rather tall, it would help
perceptually to align the option with the start of the text in that option.

3) Making the selection more distinct, would we promote the 3th line on
each option (I believe it is the measuring method, is that right? or
whatever that is) to be the headline (first line) of each option? - This
will make each option to select more distinct and easier to decide on.

I have included here an image with the above annotations and also with how
it would end up looking after applying these changes.


[image: BrXzfrSs64sPZDQ.png]
[image: 6R82rT9GsJXsVmf.png]

miss-o-soup:
Adding this note again since the images didn't seem to have linked when replied via email.

Thank you for the notes Dan.

*1)* I agree with you that we need to address the "Best available" option
for dataset sources. While it might seem tempting to hide this option when
only one dataset is available, doing so would only mask the underlying
issue rather than resolve it.

If we hide the "Best available" option in such cases, it creates a
usability problem: there's nothing for the user to actually select, making
the entire selection modal pointless. I've tested this by hiding the "Best
available" option when only one source exists, and it indeed renders the
dataset selection dialog without purpose.

Let's discuss this internally and resolve the root cause of the problem
before adding another UI patch.
BKXtJdj8bzaC7b7

2) To alleviate the disable-impression that Dan mentioned, could we:

a) Increase (a bit) the color contrast of the light gray text? We could
apply this globally to the component itself. I was thinking something
around #666a6c, or something similar that you might have already defined in
your new stylesheet. b) Would it be possible to align the radio button to
the first line of the text vs the middle section? I know this is not what
material does, but since our options are rather tall, it would help
perceptually to align the option with the start of the text in that option.

3) Making the selection more distinct, would we promote the 3th line on
each option (I believe it is the measuring method, is that right? or
whatever that is) to be the headline (first line) of each option? - This
will make each option to select more distinct and easier to decide on.

I have included here an image with the above annotations and also with how
it would end up looking after applying these changes.

BrXzfrSs64sPZDQ
6R82rT9GsJXsVmf

dwnoble:
Thanks @miss-o-soup - i am in agreement for (1), we can try to figure out a solution that gets more to the root cause. I also like moving the radio button to be top-aligned. i think that would make it clear the button is enabled and clickable.

static/js/apps/visualization/vis_type_configs/timeline_config.tsx
```diff
@@ -90,29 +93,42 @@ function groupStatVars(appContext: AppContextType): {
...
-function getFacetSelector(
-  appContext: AppContextType,
-  chartSvInfo: ContextStatVar[]
-): ReactElement {
+interface ChartFacetSelectorProps {
+  appContext: AppContextType;
+  chartSvInfo: ContextStatVar[];
+}
+
+function ChartFacetSelector({
```
dwnoble, line 101: NIce refactor

dwnoble (approved):

PR #5224, opened 2025-06-20
[Productionization Improvements] Github Action to update submodules daily.

gmechali:
This PR adds a new Github Action which will update mixer and import submodules on a daily basis and send a PR for review. 
Leaving as is for now to test the action. In a follow up, I will allow it to automatically submit the code.

### DIFF
--- a/static/js/components/subject_page/block.tsx
+++ b/static/js/components/subject_page/block.tsx
@@ -186,6 +186,7 @@ export function Block(props: BlockPropType): JSX.Element {
   const [overridePlaceTypes, setOverridePlaceTypes] =
     useState<Record<string, NamedTypedPlace>>();
   const [useDenom, setUseDenom] = useState(props.startWithDenom);
+  const [denom, setDenom] = useState<string>("");
   const isEligibleForSnapToHighestCoverage = eligibleForSnapToHighestCoverage(
     props.columns,
     props.statVarProvider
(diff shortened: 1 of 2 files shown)


.github/workflows/submodule-update.yml
```diff
@@ -0,0 +1,79 @@
...
+        run: |
+          git config user.name github-actions[bot]
+          git config user.email github-actions[bot]@users.noreply.github.com
+
+      - name: Update submodules to latest
+        id: update_submodules
+        run: |
+          SUBMODULES_UPDATED=false
+
+          echo "Checking mixer submodule..."
```
dwnoble, line 36: you might be able to make this more concise and collapse the two submodule updates together with something like:
(the downside is we dont get logging for the indivdual submodule updates)
```bash
# update all submodules
git submodule update --remote --merge

# git status --porcelain will be empty if there are no changes
if [[ -z $(git status --porcelain) ]]; then
  echo "No submodule updates found."
  echo "submodules_updated=false" >> "$GITHUB_OUTPUT"
else
  echo "Submodules updated. Committing changes."
  echo "submodules_updated=true" >> "$GITHUB_OUTPUT"
fi
```

.github/workflows/submodule-update.yml
```diff
@@ -0,0 +1,79 @@
...
+          if [ "$IMPORT_STATUS" -ne "0" ]; then
+            echo "import submodule has updates."
+            SUBMODULES_UPDATED=true
+          else
+            echo "import submodule is already up-to-date."
+          fi
+
+          if [ "$SUBMODULES_UPDATED" = "true" ]; then
+            echo "Committing updated submodules..."
+            git add mixer import
```
dwnoble, line 58: I think these two git commands are redundant- the "create pull request step" may handle adding & creating the commit

.github/workflows/submodule-update.yml
gmechali: Done, thanks!

dwnoble (approved):

PR #5226, opened 2025-06-20
[Revert Per Capita Change] 

gmechali:
I noticed this updated logic caused a regression where some queries default to using the per capita value when they should not have. Reverting for now, will find proper way forward later.

Verified fix locally.


dwnoble (approved):

vish-cs:
Merging this to unblock the rollout

PR #5229, opened 2025-06-23
[Explore] Implement the Follow Up Questions Component

javi-vaz:
It is the initial version of the Follow Up Questions Component, displaying up to 10 generated questions at the bottom of the explore page. 

Before: https://screenshot.googleplex.com/87kBdiichniHvtu
After: https://screenshot.googleplex.com/428GmoCvvrzL2ZZ

### DIFF
--- a/static/js/apps/explore/follow_up_questions.tsx
+++ b/static/js/apps/explore/follow_up_questions.tsx
@@ -26,17 +26,18 @@ import {
   URL_HASH_PARAMS,
 } from "../../constants/app/explore_constants";
 import { SubjectPageMetadata } from "../../types/subject_page_types";
+import { getTopics } from "../../utils/app/explore_utils";
 import { getUpdatedHash } from "../../utils/url_utils";
 
 // Number of follow up questions displayed
 const FOLLOW_UP_QUESTIONS_LIMIT = 10;
 
-export interface Question {
+interface Question {
   text: string;
   url: string;
 }
 
-export interface FollowUpQuestionsPropType {
+interface FollowUpQuestionsPropType {
   query: string;
   pageMetadata: SubjectPageMetadata;
 }
@@ -45,12 +46,9 @@ export function FollowUpQuestions(
   props: FollowUpQuestionsPropType
 ): ReactElement {
   // Gets the name of all related topics while removing the Root topic.
-  const relatedTopics = []
-    .concat(props.pageMetadata?.childTopics || [])
-    .concat(props.pageMetadata?.peerTopics || [])
-    .concat(props.pageMetadata?.parentTopics || [])
-    .filter((topic) => topic.dcid != DEFAULT_TOPIC || topic.name)
-    .map((topic) => topic.name)
+  // Empty string can be passed since only the topic name will be used, which is stored in the property `text`.
+  const relatedTopics = getTopics(props.pageMetadata, "")
+    .map((topic) => topic.text)
     .slice(0, FOLLOW_UP_QUESTIONS_LIMIT);
 
   if (_.isEmpty(relatedTopics) || _.isEmpty(props.query)) {
@@ -61,7 +59,7 @@ export function FollowUpQuestions(
   useEffect(() => {
     getFollowUpQuestions(props.query, relatedTopics)
       .then((value) => {
-        setFollowUpQuestions(value.followUpQuestions);
+        setFollowUpQuestions(value);
       })
       .catch(() => {
         setFollowUpQuestions([]);
@@ -94,15 +92,12 @@ export function FollowUpQuestions(
   );
 }
 
-// Full result from /api/explore/follow-up-questions
-interface FollowUpQuestionsResult {
-  followUpQuestions: Question[];
-}
-
+//  Parses request from /api/explore/follow-up-questions by creating a
+//  url with the generated question as the query.
 const getFollowUpQuestions = async (
   query: string,
   relatedTopics: string[]
-): Promise<FollowUpQuestionsResult> => {
+): Promise<Question[]> => {
   const url = "/api/explore/follow-up-questions";
   const body = {
     q: query,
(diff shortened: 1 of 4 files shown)


static/js/apps/explore/result_header_section.tsx
```diff
@@ -57,11 +58,16 @@ export function ResultHeaderSection(
...
         {topicNameStr && <span> • {topicNameStr}</span>}
       </div>
-      {!_.isEmpty(props.pageMetadata.mainTopics) && !_.isEmpty(topicList) && (
-        <div className="explore-topics-box">
-          <ItemList items={topicList} showRelevantTopicLabel={true}></ItemList>
-        </div>
-      )}
+      {!_.isEmpty(props.pageMetadata.mainTopics) &&
+        !_.isEmpty(topicList) &&
+        !showFollowUpQuestions && (
```
chejennifer, line 63: we should pass this through props instead of calling it like this because showFollowUpQuestions uses math.random and I believe constants from a module are initialized when the module is imported, so the number we get from math.random in this file might be different than the one used in the SuccessResult component

So, maybe add `showRelatedTopics` to `ResultHeaderSectionPropType` and use that property here instead

static/js/apps/explore/follow_up_questions.tsx
```diff
@@ -0,0 +1,124 @@
...
+export interface FollowUpQuestionsPropType {
+  query: string;
+  pageMetadata: SubjectPageMetadata;
+}
+
+export function FollowUpQuestions(
+  props: FollowUpQuestionsPropType
+): ReactElement {
+  // Gets the name of all related topics while removing the Root topic.
+  const relatedTopics = []
```
chejennifer, line 48: maybe we can re-use this function instead of duplicating logic: https://github.com/datacommonsorg/website/blob/master/static/js/utils/app/explore_utils.ts#L39

static/js/apps/explore/follow_up_questions.tsx
```diff
@@ -0,0 +1,124 @@
...
+  DEFAULT_TOPIC,
+  URL_HASH_PARAMS,
+} from "../../constants/app/explore_constants";
+import { SubjectPageMetadata } from "../../types/subject_page_types";
+import { getUpdatedHash } from "../../utils/url_utils";
+
+// Number of follow up questions displayed
+const FOLLOW_UP_QUESTIONS_LIMIT = 10;
+
+export interface Question {
```
chejennifer, line 34: is this interface used anywhere outside of this file? if not, we should not export it

static/js/apps/explore/follow_up_questions.tsx
```diff
@@ -0,0 +1,124 @@
...
+  text: string;
+  url: string;
+}
+
+export interface FollowUpQuestionsPropType {
```
chejennifer, line 39: same as here, don't export if it's not being used anywhere else

static/js/apps/explore/follow_up_questions.tsx
```diff
@@ -0,0 +1,124 @@
...
+              );
+            })}
+          </div>
+        </div>
+      )}
+    </div>
+  );
+}
+
+// Full result from /api/explore/follow-up-questions
```
chejennifer, line 97: nit: technically this is not the format that comes back from /api/explore/follow-up-questions because you're adding the URL to the original response

static/js/apps/explore/follow_up_questions.tsx
```diff
@@ -0,0 +1,124 @@
...
+interface FollowUpQuestionsResult {
```
chejennifer, line 98: I also don't think we really need this

`getFollowUpQuestions` can just return `Promise<Question[]>`

chejennifer (commented):
nice!

static/js/apps/explore/follow_up_questions.tsx
javi-vaz: Done.

static/js/apps/explore/follow_up_questions.tsx
javi-vaz: Done.

static/js/apps/explore/follow_up_questions.tsx
javi-vaz: I called the function and then parsed the response to extract only the names of the topics.

static/js/apps/explore/success_result.tsx
```diff
@@ -162,6 +175,7 @@ export function SuccessResult(props: SuccessResultPropType): ReactElement {
                 pageMetadata={props.pageMetadata}
                 placeUrlVal={placeUrlVal}
                 hideRelatedTopics={false}
+                showFollowUpQuestions={showFollowUpQuestions}
```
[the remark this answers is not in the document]
javi-vaz: Done.

static/js/apps/explore/follow_up_questions.tsx
```diff
@@ -0,0 +1,117 @@
...
+              );
+            })}
+          </div>
+        </div>
+      )}
+    </div>
+  );
+}
+
+//  Parses request from /api/explore/follow-up-questions by creating a
```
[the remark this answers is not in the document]
javi-vaz: Done.

static/js/apps/explore/follow_up_questions.tsx
```diff
@@ -0,0 +1,117 @@
...
+  const [followUpQuestions, setFollowUpQuestions] = useState(null);
+  useEffect(() => {
+    getFollowUpQuestions(props.query, relatedTopics)
+      .then((value) => {
+        setFollowUpQuestions(value);
+      })
+      .catch(() => {
+        setFollowUpQuestions([]);
+      });
+  }, []);
```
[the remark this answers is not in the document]
javi-vaz: I added the props.query dependency, and moved the creation of relatedTopics to within the useEffect so it wouldn't continuously be called due to the lack of a stable identity.

Also, I moved the location of the useEffect to fix the other lint issues appearing.

chejennifer (approved):
thanks for updating!

static/js/apps/explore/success_result.tsx
```diff
@@ -53,6 +59,13 @@ import { UserMessage } from "./user_message";
 
 const PAGE_ID = "explore";
 
+const EXPERIMENT_FOLLOW_UP_ROLLOUT_RATIO = 0.2;
+
+const showFollowUpQuestions =
```
clincoln8, line 64: Does this mean that for the experiment, it will be a 20% chance of seeing the follow up questions each time a query is submitted? ie a user may see both questions and topics in a single session when submitting multiple queries? (just want to confirm)

clincoln8 (approved):
Thanks! Can you include a screenshot of the feature in the PR description?

PR #5238, opened 2025-06-25
[Update Submods] Fix potential unrelated histories error.

gmechali:
See: https://github.com/datacommonsorg/website/actions/runs/15880687506/job/44779942452

It seems there is no simple way to get around this but this pretty complex bash script within the yaml. I haven't yet been able to test it.


dwnoble (commented):
Did you try something like this? https://github.com/actions/checkout/issues/125

the error coudl be related to a 'shallow' checkout of the submodules

this all looks very annoying to test, but maybe adding 
adding `submodules: recursive` back in along with `fetch-depth: 0` might be a fix?

gmechali:
Thanks Dan, that actually did resolve it! closing this Pr.

PR #5244, opened 2025-06-26
Add all statvars from BigQuery

alyssasuiyiguo:


tools/nl/nl_metadata/add_metadata.py
```diff
@@ -120,142 +165,70 @@ def split_into_batches(
   return batched_df_list
 
 
-def get_prop_value(prop_data) -> str:
```
alyssasuiyiguo, line 123: Because the metadata now gets pulled from BigQuery, the next 4 functions (`get_prop_value`, `extract_metadata`, `extract_constraint_properties`, `create_sv_metadata`) no longer get called. Should I leave them in?

tools/nl/nl_metadata/add_metadata.py
```diff
@@ -12,25 +12,25 @@
...
+4. Create a new dataframe with the SVs and their full metadata, and export it as a JSON file sv_complete_metadata_{target_language}_{page_number}.json.
+   If the flag --saveToGCS is specified, also save to cloud storage.
 
 To run this script, make a copy of .env.sample and register your data commons and Gemini API keys to DOTENV_FILE_PATH (./.env), then run the script using the command ./add_metadata.py
 """
 import argparse
 import asyncio
 import json
 import os
-from typing import Dict, List
```
alyssasuiyiguo, line 28: I realized I had some inconsistencies between using dict/Dict and list/List so I changed everything to using python's built-in dict/list

tools/nl/nl_metadata/add_metadata.py
```diff
@@ -264,58 +237,58 @@ async def batch_generate_alt_sentences(
       seed=GEMINI_SEED,
       max_output_tokens=GEMINI_MAX_OUTPUT_TOKENS,
       response_mime_type="application/json",
-  )
-  batched_list: List[List[dict[str, str | list[str]]]] = split_into_batches(
+      response_schema=list[StatVarMetadata])
```
alyssasuiyiguo, line 240: One thing to note - since the StatVarMetadata class has all its fields in English, I don't think this response_schema would work if we asked gemini to translate the metadata. I could probably set up alternative StatVarMetadata classes in French and Spanish and pick the one to pass in to response_schema based on the target_language, but I didn't think that was worth doing for now since we've decided to de-prioritize i18n.

tools/nl/nl_metadata/add_metadata.py
```diff
@@ -95,8 +98,50 @@ def extract_flag() -> argparse.Namespace:
...
+  results = query_job.result(page_size=PAGE_SIZE)
+
+  page_number: int = 1
+
+  for page in results.pages:
+    page_svs: list[dict[str, str | list[str]]] = []
+    for statvar in page:
+      # Collect constraint properties
+      constraint_properties = []
+      for i in range(1, 11):
```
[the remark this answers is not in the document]
alyssasuiyiguo: In BigQuery the constraint properties are stored as key-value pairs in separate columns, going from p1, v1 up to p10, v10, so the for loop goes from 1-10 to check if constraint properties are populated in each of these columns. I also added a comment in the code to explain this!

gmechali (approved):

PR #5254, opened 2025-06-30
Data Selector - Button and selection updates

nick-nlb:
## Description

This PR updates now the dataset selector button displays in various situations:

- When there are datasets to choose from (at least one stat var in the chart has multiple facets), we display "Explore other datasets" as the text of the button.
- When there are no datasets to choose from (no stat vars have multiple facets), the text "Displaying all the datasets available" is displayed, and no button is made available.
- If we display the button (because at least one stat var has a choice), any stat vars that do not display a choice display the single facet with a radio button. The "combined" option is now gone in that situation.
- In the download tool, when there are datasets to choose from, we show either "Select datasets" or "Select a dataset", depending on how many stat vars there are. This is the same behaviour as before.
- If there are no datasets to choose from, we do not display the button at all.

## Notes

- While we did update the copy of the button for the old facet selector (the one not behind the feature flag), we didn't update the old facet selector dialog itself. This was intentional to try to keep the changes as light as possible, given that the facet selector will disappear once we go live with the new one.
- This involved a minor webdriver test update because locally a test was failing due to case sensitivity that may not have been failing in other places. This update aligns that test with other tests that have been changed in a similar way recently. 
- We did some refactoring and cleanup of the rich facet selector dialog content - this has made the overall git diff a bit bigger than the changes would otherwise seem.

### No choices available

[Link to chart](http://localhost:8080/tools/visualization#visType%3Dtimeline%26place%3DgeoId%2F06%26placeType%3DCounty%26sv%3D%7B%22dcid%22%3A%22Count_Person_EducationalAttainmentBachelorsDegree%22%2C%22pc%22%3A%221%22%7D___%7B%22dcid%22%3A%22Count_Person_AmbulatoryDifficulty%22%7D
)

displaying all the datasets available

### Choices available

[Link to chart](http://localhost:8080/tools/visualization#visType%3Dtimeline%26place%3DgeoId%2F0606000___geoId%2F0656938%26placeType%3DCensusZipCodeTabulationArea%26sv%3D%7B%22dcid%22%3A%22Count_Person_EducationalAttainmentBachelorsDegree%22%2C%22pc%22%3A%221%22%7D___%7B%22dcid%22%3A%22Count_Person_EducationalAttainmentMastersDegree%22%2C%22pc%22%3A%221%22%7D___%7B%22dcid%22%3A%22Count_Person_EducationalAttainmentDoctorateDegree%22%2C%22pc%22%3A%221%22%7D)

choices available

choices available modal content

### Download tool - One stat var, choices available

download tool - one stat var choices

### Download tool - Multiple stat vars, choices available

download tool - two stat vars, choices

### Download tool - no choices
download tool - no choices

### DIFF
--- a/static/js/tools/timeline/__snapshots__/page.test.tsx.snap
+++ b/static/js/tools/timeline/__snapshots__/page.test.tsx.snap
@@ -2,7 +2,9 @@
 
 exports[`Single place and single stat var 1`] = `
 "<div class=\\"chart-container dc-async-element-holder\\">
-  <div class=\\"css-1w3kops\\"><span class=\\"css-17y0nyh\\">Displaying all the datasets available</span></div>
+  <div class=\\"css-1w3kops\\">
+    <p class=\\"css-15y2wdo\\">Displaying all the datasets available</p>
+  </div>
   <div class=\\"card\\">
     <div class=\\"statVarChipRegion\\">
       <div class=\\"chip\\"><span class=\\"chip-text chip-clickable-text\\">Age</span><button class=\\"chip-action\\"><i class=\\"material-icons\\">cancel</i></button></div>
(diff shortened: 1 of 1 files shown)


miss-o-soup (commented):
These all looks good to me and according to our conversations. I leave the final stamp to Dan in case I missed something. Thank you!

PR #5255, opened 2025-07-01
[Submod Update] Only Run the action on the main repo

gmechali:
Updates the submodule update action to ensure it does not run on forked repositories, and to add the date on the branch name,


gmechali:
The action is configured to only run on Master. Master is what we deploy from, so it's only needed in master. Other branches should pull in master to be in sync

dwnoble (approved):

PR #5269, opened 2025-07-07
Block-level dataset selector (line, highlight, scatter)

nick-nlb:
## Description

This PR implements dataset selection at the level of the blocks used in the NL search.

This first round applies it to the line, highlight and scatter charts. These are the three charts for which there are no changes required in the dataset selector mechanism itself 

These updates are limited to the NL search, and exclude the web components.

### Updates to the tiles

We introduced a per-block cache that memoizes StatVarSpec[] results. This prevents block-level re-renders from
triggering refetches in the child.

This update was prompted by the additional rendering cycles from the facet selector handling. However any similar functionality added into the block would have required this as well.

Although we tried to keep the changes in the block itself (to minimize the breadth of the PR), we did have to update some aspects of the tiles themselves. For example, the highlight and ranking tiles were triggering a fetch on a dependency of [props] - resulting in duplicate calls. We now destructure the props and only trigger the fetch if a relevant prop has changed. 

This overall sensitivity that the tiles have to re-rendering could be hardened in a future PR (thus allowing the tiles to be used in any context without rendering causing unnecessary fetches).

### Selector on a line/highlight pair

Screenshot 2025-07-07 at 9 44 55 AM

### Selector on a scatter plot

Screenshot 2025-07-07 at 9 44 04 AM

### Block with only one dataset available

Screenshot 2025-07-07 at 9 44 34 AM

## Testing

Between the following calls, you will see examples of each of the three chart types involved in this PR:

### Line, highlight

http://localhost:8080/explore#q=life+expectancy+in+sweden&client=ui_query

### Scatter

http://localhost:8080/explore#q=poverty+vs+employment&client=ui_query

### No dataset choices available

http://localhost:8080/explore#q=united+states+statistics&client=ui_query

## Feature Flag

This functionality is behind the same feature flag used for the rich metadata modal and the rich facet selector, and so will show up locally and in autopush.


static/js/components/subject_page/stat_var_spec.tsx
```diff
@@ -0,0 +1,97 @@
...
+ * facet override.
+ *
+ * @param snapToHighestCoverage - Whether to snap to the date with the highest coverage.
+ * @param useDenom - Whether to use a denominator for per capita calculations.
+ * @param denom - The denominator to use.
+ * @param facetOverrides - Any facet overrides.
+ * @param statVarProvider - The stat var provider.
+ * @returns An object containing getStatVarSpec and getSingleStatVarSpec functions.
+ */
+export function useStatVarSpec(
```
dwnoble, line 53: Nice refactor!

miss-o-soup:
Thank you for doing all these updates Nick/Pablo, and thank you for the reviews and note Dan.
Per our offline discussion, let's hide the case of "Displaying all datasets available" in the NL page result pages and keep in the visualization tool use cases.

static/js/components/subject_page/block.tsx
```diff
@@ -199,10 +268,109 @@ export function Block(props: BlockPropType): JSX.Element {
...
+    const allTiles = _.flatten(props.columns.map((c) => c.tiles));
+    const placeDcids = new Set<string>([props.place.dcid]);
+    allTiles.forEach((tile) => {
+      if (tile.placeDcidOverride) {
+        placeDcids.add(tile.placeDcidOverride);
+      }
+      getComparisonPlaces(tile, props.place)?.forEach((p) => placeDcids.add(p));
+    });
+
+    const baseFacets = await getFacets(
```
dwnoble, line 336: I'm seeing an issue with the facet selection on some of the scatter plots: On [this page](http://localhost:8080/explore#q=poverty+vs+employment&client=ui_query) if i select the second ACS facet, im seeing an error:

69CDcQs7VT9E6s8 (2)


It looks like the facet check may only be looking at country level data instead of state-level data.


Also unrelated - how come the text for both these ACS facets are the same?

PR #5271, opened 2025-07-07
Enable experiment flags for Follow Up Questions

javi-vaz:
Enables the follow up questions experiment in all environments except Prod and Custom. 
- Prod requires that Staging have the flag enabled, and it will allow us to test the feature before pushing to Prod.
- Custom will require additional considerations and implementation that will be completed in the future.


gmechali (approved):

clincoln8 (approved):
woohoo!

chejennifer (approved):

PR #5276, opened 2025-07-08
nl curation script for curating matching variables

chejennifer:
This adds a tool for auto-curating variable dcids (including all the variables in the Data Commons knowledge graph) that match queries.


tools/nl/detection_evals/golden_curation/generate_vars.py
```diff
@@ -0,0 +1,370 @@
...
+    list: A list of variable DCIDs associated with the input variable string and SVGs.
+  """
+  var_list = []
+  # query the index for each svg to find all the associated vars for the var string
+  for svg in svg_list:
+    embedding_idx = _SVG_TO_EMBEDDING_IDX.get(svg)
+    idx_param = f'&idx={embedding_idx}' if embedding_idx else ''
+    request_query = f'{var_string} {_DEFAULT_PLACE}'
+    nl_response = requests.post(
+        f'http://localhost:8080/api/explore/detect?q={request_query}{idx_param}',
```
clincoln8, line 233: Is there a benefit to using `/api/explore/detect` over `/api/nl/search-vector`?

tools/nl/detection_evals/golden_curation/README.md
```diff
@@ -0,0 +1,30 @@
...
+
+### Prerequisites
+
+1. In your input folder, you need to have your CSV files with the following columns (there can be multiple rows for the same query if there is disagreement on what the variable string for the query is)
+    - query_num: This is a unique number for each query
+    - query: This is the full query string
+    - var_string: This is the substring of the query that should be used to
+    detect the variable in the query
+    - groups: These are the groups that the variable in the query could belong to. The variable could belong to multiple groups and these groups should be comma separated.
+2. Start a local server with additional embedding indices enabled for all the variable groups
```
clincoln8, line 20: is this a `website` server w/ llm_mode enabled? Can you add some more info about getting it started?
Is it simply adding the catalog + env configs you've linked and then running `./run_server -m` like normal?

tools/nl/detection_evals/golden_curation/generate_vars.py
```diff
@@ -0,0 +1,370 @@
...
+    topic_cache: Dict[str, List[str]]) -> Tuple[Dict[str, List[str]], Set[str]]:
+  """
+  Processes a list of variable names to extract and format topic strings, and identifies statistical variables (SVs) included in the topics.
+
+  Args:
+    var_list (List[str]): A list of variable dcids, some of which may represent topics (identified by a specific prefix).
+    topic_cache (Dict[str, List[str]]): A mapping from topic dcid to their child topics or SVs.
+
+  Returns:
+    Tuple[Dict[str, List[str]], Set[str]]:
```
clincoln8, line 145: It would be helpful to see an example input/output for this function. I'm not exactly sure what "formatted topic strings" are.

tools/nl/detection_evals/golden_curation/generate_vars.py
```diff
@@ -0,0 +1,370 @@
...
+      for dcid in node.get('dcid', []):
+        result[dcid] = children
+  return result
+
+
+def get_topic_strings(
+    var_list: List[str],
+    topic_cache: Dict[str, List[str]]) -> Tuple[Dict[str, List[str]], Set[str]]:
+  """
+  Processes a list of variable names to extract and format topic strings, and identifies statistical variables (SVs) included in the topics.
```
clincoln8, line 138: Add a note in the description that this only "unravels" the topic up to MAX_TOPIC_DEPTH

tools/nl/detection_evals/golden_curation/generate_vars.py
```diff
@@ -0,0 +1,370 @@
...
+                    'The input directory of files with queries to get svs for')
+flags.DEFINE_string(
+    'output_folder', 'output',
+    'The output directory to save results of getting svs for each query')
+flags.DEFINE_string(
+    'topic_cache', 'topic_cache.json',
+    'file path to a copy of the topic cache at server/config/nl_page/topic_cache.json'
+)
+
+_MAX_TOPICS_DEPTH = 3
```
clincoln8, line 36: How did you choose 3 for this?

tools/nl/detection_evals/golden_curation/README.md
```diff
@@ -0,0 +1,30 @@
+# Golden Curation
+
+This folder contains resources and scripts for curating the goldens for nl
+detection evals. These are for auto-curating the goldens that then need to go
+through manual checks.
+
+## Generate Vars
+
+This auto-curates a list of variable dcids (including all the variables in the
+Data Commons knowledge graph) that match a given query.
```
clincoln8, line 10: It might be helpful to add a few more sentences of at a high level what exactly this does to generate the list. ie 

"""
This script performs a simple embedding search on indexes that are specific to a single by top level StatVarGroups. For example, given the statvargroup "Health" would have a single embeddings index containing all statvars within that group.

High Level Workflow:

First read in the input file, transforming it to QueryInfo objects. Then read in the topic cache file.

Given the query ABC with variable_str X and SVgroups: [Y,Z]:

*  for each SVgroup, query an embeddings index associated with that group via locally running website server to get matched stat_vars and sv_topics
    * sv_topics are unraveled using the topic_cache.json file up to MAX_TOPIC_DEPTH  to get individual stat_vars
        * <something about formatting that I don't 100% understand. is this adding the "-->-->" prefixes? I think that's something worth noting>
    * if a stat_var was returned as an indivual result in step1 and also included in an unraveled topic, then we only return it within the topic so as not to duplicate statvars in the response

"""

tools/nl/detection_evals/golden_curation/generate_vars.py
clincoln8: After taking a closer look, I remembered that we added the "-->" prefix for topic hierarchy. I think that's worth detailing here and in the overview in the README. 
It can be described with just a simple example.

clincoln8 (approved):
Thanks for getting this in!! All of my comments are related to documentation. If this is blocking the next PR, feel free to submit this one as is and then follow up with the clarifications in a separate PR.

tools/nl/detection_evals/golden_curation/generate_vars.py
chejennifer: I don't think so, I just chose detect because I didn't know the search-vector endpoint existed before ..

tools/nl/detection_evals/golden_curation/generate_vars.py
chejennifer: Because the stat vars of a peer group will be at atleast level 2 so I wanted one more for good measure

PR #5278, opened 2025-07-09
[Explore] Added the Explanation Feature Flags

javi-vaz:
Added the feature flags to control the roll-out of the Overall Explanation feature and the corresponding Typescript constants.

### DIFF
--- a/server/config/feature_flag_configs/autopush.json
+++ b/server/config/feature_flag_configs/autopush.json
@@ -40,5 +40,17 @@
     "enabled": true,
     "owner": "nick-next",
     "description": "Feature flag to show the complete metadata modal. If not set, the older modal will be used as a fallback."
+  },
+  {
+    "name": "overall_explanation_experiment",
+    "enabled": false,
+    "owner": "javiervazquez",
+    "description": "Enables the overall explanation generated from query and relevant stat vars as an experiment."
+  },
+  {
+    "name": "overall_explanation_general_audience",
+    "enabled": false,
+    "owner": "javiervazquez",
+    "description": "Enables the overall explanation generated from query and relevant stat vars to general audience."
   }
 ]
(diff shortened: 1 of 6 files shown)


javi-vaz:
> I would recommend using the "ga" naming just because it's so long to use general_audience, but I know there was a question of it being confusing with google analytics, so leaving it up to you.
> 
> Make sure to run the lint fix script!

I believe at the end of the discussion Prem had said no more abbreviations. He said to use the long form google_analytics and general_availability.

side note: I barely realized I put audience instead of availability so I'll fix that and the lint.

javi-vaz:
> > > > I would recommend using the "ga" naming just because it's so long to use general_audience, but I know there was a question of it being confusing with google analytics, so leaving it up to you.
> > > > Make sure to run the lint fix script!
> > > 
> > > 
> > > I believe at the end of the discussion Prem had said no more abbreviations. He said to use the long form google_analytics and general_availability.
> > > side note: I barely realized I put audience instead of availability so I'll fix that and the lint.
> > 
> > 
> > Adding to this, I believe we're planning a small PR where we clean up the feature flags and correct the naming convention.
> 
> So that might be a problem. You can't just rename them because the files in cloud storage get updated immediately. while the release happens biweekly (roughly). If you want to rename them, you need a make a new flag, wait for a release, make the new flag control the same behavior as the old flag then deprecate the new flag (and make sure releases go through properly).
> 
> In other words, truly not worth the risk/effort to rename them

That makes sense, so do we just keep the long naming convention going forward?

server/config/feature_flag_configs/autopush.json
```diff
@@ -41,6 +41,18 @@
     "owner": "nick-next",
     "description": "Feature flag to show the complete metadata modal. If not set, the older modal will be used as a fallback."
   },
+  {
+    "name": "page_overview_experiment",
+    "enabled": false,
+    "owner": "javiervazquez",
+    "description": "Enables the page overview generated from query and relevant stat vars as an experiment."
```
clincoln8, line 48: It would be helpful to document what you mean by "experiment" here. 

For example, project 1's description would include "enables a 20% chance of the feature being displayed. Once the feature has been displayed, it will continue to be displayed for the entirety of the user's session." 

If you're not sure of this experiment's mechanics yet, then feel free to submit as is, but consider coming back and making this more descriptive once it has been decided.

clincoln8 (approved):

PR #5279, opened 2025-07-09
[Vis Tools] Add feature flag and hide search bar in the header menu

juliawu:
This is part 1 of many PRs that will standardize the visualization tool for our users. This PR:

* Adds a `standardized_vis_tool` feature flag to hide changes behind
* Removes the search bar in the header of the old tools

Before:
After:

Before:
After:

Before:
After:

### DIFF
--- a/server/config/feature_flag_configs/autopush.json
+++ b/server/config/feature_flag_configs/autopush.json
@@ -40,5 +40,11 @@
     "enabled": true,
     "owner": "nick-next",
     "description": "Feature flag to show the complete metadata modal. If not set, the older modal will be used as a fallback."
+  },
+  {
+    "name": "standardized_vis_tool",
+    "enabled": false,
+    "owner": "juliawu",
+    "description": "Enables standardized visualization tool UI for the /tool/map, /tool/scatter, and /tool/timeline pages"
   }
 ]
--- a/server/lib/feature_flags.py
+++ b/server/lib/feature_flags.py
@@ -18,6 +18,7 @@ from flask import current_app
 AUTOCOMPLETE_FEATURE_FLAG = 'autocomplete'
 BIOMED_NL_FEATURE_FLAG = 'biomed_nl'
 DATA_OVERVIEW_FEATURE_FLAG = 'data_overview'
+STANDARDIZED_VIS_TOOL_FEATURE_FLAG = 'standardized_vis_tool'
 
 
 def is_feature_enabled(feature_name: str, app=None) -> bool:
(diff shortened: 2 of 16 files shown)


server/lib/feature_flags.py
```diff
@@ -18,10 +18,18 @@
...
 
-def is_feature_enabled(feature_name: str, app=None) -> bool:
+def is_feature_enabled(feature_name: str, app=None, request=None) -> bool:
   """Returns whether the feature with `feature_name` is enabled."""
   if not app:
     app = current_app
-  return app.config['FEATURE_FLAGS'].get(feature_name, False)
+  # check if a URL param override for the feature is present
+  url_override_present = False
+  if request:
```
gmechali, line 31: nit: I think it improves readability a bit to extract the override check in its own function!
But nice find that we forgot to build the override on the API feature check!

def is_feature_override_enabled(featureName: string, request = None) -> bool {
  if request is None:
    return false

  return request.args.get(
        FEATURE_FLAG_URL_OVERRIDE_PARAM) == feature_name
}

then you can just do
if is_feature_override_enabled(featre_name, request):
   return true

return app.config['FEATURE_FLAGS'].get(feature_name, False)

server/config/feature_flag_configs/autopush.json
```diff
@@ -40,5 +40,11 @@
     "enabled": true,
     "owner": "nick-next",
     "description": "Feature flag to show the complete metadata modal. If not set, the older modal will be used as a fallback."
+  },
+  {
```
gmechali, line 44: ultra nit: for the future (no need to update) I would recommend just submitting feature flag changes on their own. 

That way if we need to rollback a PR due to the code below, the feature flags arent touched

gmechali (approved):
Just left a few nits but nice!

server/config/feature_flag_configs/autopush.json
juliawu: Sounds good, will do!

static/css/tools/base_tools.scss
```diff
@@ -121,3 +120,53 @@ body {
     text-decoration: none;
   }
 }
+
+/** TODO(juliawu): Once migration to old visualization tools is complete,
+                   use these styles as default and remove the extra
+                   `class="standardized-tool-ui"` from map.html, scatter.html,
+                   and timeline.html. */
+.standardized-tool-ui {
```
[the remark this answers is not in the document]
juliawu: Good catch, this was intended to go in the following PR, removing from this one.

PR #5283, opened 2025-07-10
[Vis Tools] Update header styling on the old tools

juliawu:
Updates the styling of the title on the "old" map, scatter, and timeline tool landing pages to match the mocks at https://docs.google.com/presentation/d/1ztyZwu8b3YzU1MMAFPI3iC-Hm00CE63rPb7eP6vGZlw/edit?slide=id.g3052c79bc05_0_12#slide=id.g3052c79bc05_0_12

Changes made:
* Create a new shared `ToolHeader` component to unify styling between the tools
* Add i18n support to each of the tools -- currently they are hardcoded to the "en" locale.
* Add a helper function to allow setting of `?hl={locale_code}` parameters in the URL, which matches behavior on the rest of the site

Updates to the place selector card and the landing page text will come in subsequent PRs.

Screenshots:
Map Header
Scatter Header
Timelines Header


juliawu:
@miss-o-soup I've omitted mentions of "statistical variable on the left pane" from the headers because the left pane disappears on narrower screens. I think once we add the blue box with "how to get started" instructions, this will be clear for the user anyway.


Screenshot 2025-07-10 at 1 21 40 PM

server/templates/tools/map.html
```diff
@@ -33,7 +33,11 @@
 
 {% block content %}
   <div id="metadata" class="d-none" data-footer="{{ config['MAP_TOOL_FOOTER'] }}"></div>
+  {% if use_standardized_ui %}
+  <div id="main-pane" class="standardized-vis-tool"></div>
```
gmechali, line 37: Is there any reason why we can't do this via the ts fragments?

static/css/tools/base_tools.scss
```diff
@@ -121,3 +120,53 @@ body {
     text-decoration: none;
   }
 }
+
+/** TODO(juliawu): Once migration to old visualization tools is complete,
+                   use these styles as default and remove the extra
+                   `class="standardized-vis-tool"` from map.html, scatter.html,
+                   and timeline.html. */
+.standardized-vis-tool {
```
gmechali, line 128: I know the plan is to use emotion styling now. I think* if we move the class change to be in the ts, we should be able to control this CSS there too.

static/js/i18n/i18n.tsx
```diff
@@ -407,9 +407,21 @@ function formatDate(dateString: string, locale?: string): string {
   return date.toLocaleDateString(locale || intl.locale, options);
 }
 
+/**
+ * Fetch locale from the URL
+ *
+ * Used to get the current locale set by a ?hl= url parameter.
+ * If the parameter is not present, defaults to the "en" locale.
+ */
+function getLocaleFromUrl(): string {
```
gmechali, line 416: I wonder if this is better in static/js/utils/url_utils.ts so that it can be grouped in all things related to readying URL hash params, as opposed to i18n which only applies to this specific flag? 

Undecided tbh...wdyt?

static/js/tools/map/app.tsx
```diff
@@ -55,7 +64,25 @@ function App(): ReactElement {
       <div id="plot-container" className={ASYNC_ELEMENT_HOLDER_CLASS}>
         <Container fluid={true}>
           <Row>
-            <Title />
+            {useStandardizedUi ? (
+              <ToolHeader
+                title={intl.formatMessage({
+                  id: "map_visualization_tool_name",
```
gmechali, line 70: https://github.com/datacommonsorg/website/tree/b04b2de9a3acd3c7cdce7a53e574757737048469/static/js/i18n

A bunch of the messages today are in this messages files. Could you make a new one for i18n_vistool_messages?

Here and in all other files!

static/js/tools/scatter/app.tsx
```diff
@@ -53,6 +59,10 @@ function App(): ReactElement {
   const showInfo = !showChart && !showChooseStatVarMessage;
   const [isSvModalOpen, updateSvModalOpen] = useState(false);
   const toggleSvModalCallback = (): void => updateSvModalOpen(!isSvModalOpen);
+  const useStandardizedUi = isFeatureEnabled(
+    STANDARDIZED_VIS_TOOL_FEATURE_FLAG
+  );
+  console.log(useStandardizedUi);
```
gmechali, line 65: nit: remove logging!

static/js/tools/scatter/app.tsx
```diff
@@ -63,12 +73,31 @@ function App(): ReactElement {
...
+                    description:
+                      "a description of what our scatter plot explorer tool is used for",
+                  })}
+                  switchToolsUrl="/tools/visualization#visType%3Dscatter"
+                />
+              ) : (
+                <div className="app-header">
+                  <h1 className="mb-4">Scatter Plot Explorer</h1>
+                  <a href="/tools/visualization#visType%3Dscatter">
+                    Go back to the new Data Commons
```
gmechali, line 97: Should this say go back to the new XX tool?

dwnoble: And can you extract this to the i18n file?

static/js/tools/shared/tool_header.tsx
```diff
@@ -0,0 +1,53 @@
+/**
+ * Copyright 2022 Google LLC
```
dwnoble, line 2: Nit: 2025

static/js/tools/scatter/app.tsx
juliawu: Thanks for the catch! Removed

static/js/tools/shared/tool_header.tsx
juliawu: Thanks for the catch! Updated.

server/templates/tools/map.html
juliawu: Good idea, moved this to the ts fragments

static/css/tools/base_tools.scss
juliawu: Originally didn't want to have mixed styling (some in emotion, some in .scss file), but I think you're right that the logic is cleaner if we use emotion. Swapped to using emotion styling.

static/js/i18n/i18n.tsx
juliawu: I think I like it in url_utils better. Moved.

static/js/tools/map/app.tsx
juliawu: Good idea. Extracted to a new message file.

static/js/tools/scatter/app.tsx
juliawu: This logic is for what's currently in prod.

@gmechali Agreed, I've updated that text for all the tools to make things more clear.

@dwnoble I think adding i18n to the old code is out of scope for this PR, those changes will be superseded anyway once we roll out the new styling to prod. Let me know if you still want me to add 18n for this section.

static/css/tools/base_tools.scss
gmechali: oh perfect!

dwnoble (approved):

PR #5284, opened 2025-07-10
2025-07-10 Custom DC stable release

dwnoble:
# Highlights
* Fixed a bug in Custom DC data imports where generated stat var grouping with long IDs would cause the import to fail
* Fixed a bug where Custom DC imports with stat var groupings disabled would cause NL search to fail

Note: this is an off-cycle release, and it not in sync with the datacommons.org production image

Verified changes by
* Verified "URLS to check" locally from go/dconcall 
* Verified locally that the data import changes work
* Verified with the UN that the data import changes work in cloud run
* Verified with the UN that there are no breaking changes to their undata site.


clincoln8 (approved):
Thanks Dan!

PR #5292, opened 2025-07-14
Connect VAI app to Statvar explorer

alyssasuiyiguo:
Connects the Vertex AI application for stat var search to the stat var explorer. This change is controlled by the feature flag `vai_for_statvar_search` introduced in [this PR](https://github.com/datacommonsorg/website/pull/5291). When enabled, the search will only take place if no sources are selected for the search - in the case that they are, the search falls back onto the old stat var search endpoint. 

[Screen recording of UI changes ](http://screencast/cast/NDY1MjI1NTAyOTQ5Mzc2MHwyNjM1YWYwNi1iOA)

Screenshot of UI changes: I used the results count styling for reference, and re-used the loading component for when the changes are loading in. 
image
image

Codacy errors:
image

- "page_two" token for server test case was erroneously flagged as a possible hard-coded password
- Was getting an error from not giving limit a "number" typedef in `search_utils.tsx`, but adding in the typedef was causing a lint error due to the type being easily inferred from the default value.


server/routes/shared_api/stats.py
```diff
@@ -31,6 +35,32 @@
...
+) -> discoveryengine.services.search_service.pagers.SearchPager:
+  """Search statvars using Vertex AI search application."""
+  search_request = discoveryengine.SearchRequest(
+      serving_config=serving_config,
+      query=query,
+      page_token=page_token,
+      page_size=100,
+      spell_correction_spec=discoveryengine.SearchRequest.SpellCorrectionSpec(
+          mode=discoveryengine.SearchRequest.SpellCorrectionSpec.Mode.AUTO),
+      relevance_threshold=discoveryengine.SearchRequest.RelevanceThreshold.LOW)
```
alyssasuiyiguo, line 58: I ultimately chose to set the relevance threshold to `LOW` instead of `MEDIUM` because there were some queries that were returning no stat vars once the threshold was set to `MEDIUM`. For instance, searching "crime" or "robberies" both returned no stat vars even though we have a Statvar called "Robberies", and we have 8637 statvars under "Crime" in the hierarchy. The tradeoff is that for some inquiries there's a lot more noise in the results (ex. "Population" with a relevance threshold of `LOW` returns pretty bad results), but I personally think it'd be better to have a lower threshold so that for most queries users can manually scroll through the results to find the most relevant one, rather than having a higher threshold so that for some queries no search results get returned at all.

static/js/stat_var_hierarchy/stat_var_search.tsx
```diff
@@ -148,12 +157,12 @@ export class StatVarHierarchySearch extends React.Component<
...
               {!_.isEmpty(this.state.svResults) && (
                 <div className="sv-search-results">
-                  {this.state.svResults.map((sv) => {
+                  {this.state.svResults.map((sv, index) => {
                     return (
                       <div
                         className="search-result-value"
                         onClick={this.onResultSelected(sv.dcid)}
-                        key={sv.dcid}
+                        key={`${sv.dcid}-${index}`}
```
alyssasuiyiguo, line 165: This change in key was required because the VAI pagination was at times returning duplicate search results across different pages (which is a known bug with Google cloud search previously filed on Buganizer). Adding the relevance threshold to our search request should filter out most irrelevant results to prevent duplicates, but as an extra precaution I added the index to the key here to prevent React warnings from children having duplicate keys.

server/routes/shared_api/stats.py
```diff
@@ -105,13 +135,39 @@ def stat_var_property():
...
+
+  if is_vai_enabled and len(places) == 0:
+    statVars = []
+    page_token = None
+    while len(statVars) < limit:
+      search_results = search_vertexai(query, page_token)
+      for response in search_results.results:
+        dcid = response.document.struct_data.get("dcid")
+        name = response.document.struct_data.get("name")
+        if dcid and name:
```
gmechali, line 159: Can you change this to be:
if not name or not dcid"
  log --> "Tjere's a problem with " name + dcid.
  continue

statVars.append(....)
...

alyssasuiyiguo:
> Love the testing!
> 
> Thanks for making that so thorough. Overall this looks great to me.
> 
> Could you just add to the PR description a screencast demoing the feature! https://screencast.googleplex.com/

Yup I have one already! http://screencast/cast/NDY1MjI1NTAyOTQ5Mzc2MHwyNjM1YWYwNi1iOA

gmechali (approved):
Looks great, thanks Alyssa, exciting to see this land!

PR #5295, opened 2025-07-14
Updated slider web component to subscribe to events on its own publish channel

dwnoble:
This fixes an issue the UN raised where they were manually triggering events to update the slider state, but the slider wasn't actually getting updated 

Issue: https://screencast.googleplex.com/cast/NTYxNzE1MDAzNDE4MjE0NHxkNWZhMDYzNC04NA

After this change, the slider will listen to events on its own publish channel, making it possible to stay in sync with other sliders or with manually triggered events.

Demo: https://screencast.googleplex.com/cast/NTk3MzQ2NTE1MTU2OTkyMHw5ZWFjNDVlZS1lMw
JS Fiddle for local testing: https://jsfiddle.net/dkh8tw4p/1/

NOTE: I tried adding a unit test for this functionality using https://open-wc.org/docs/testing/testing-package/ , but apparently our "jest" configuration is incompatible with "ES6" module imports.

It's a relatively sizable change to fix our jest configuration, so i'm leaving that for a follow-up PR. Some options are:
1. Update jest configuration (this may have other side effects): https://stackoverflow.com/questions/35756479/does-jest-support-es6-import-export
2. Migrate from jest to vitest https://www.speakeasy.com/blog/vitest-vs-jest


static/library/slider_component.ts
```diff
@@ -326,6 +326,11 @@ export class DatacommonsSliderComponent extends LitElement {
   @state()
   private _value: number;
 
+  /**
+   * Flag to prevent self-triggering when dispatching events
+   */
+  private _isDispatching = false;
```
gmechali, line 332: nit: add bool typedef for codacy.

dwnoble: done

PR #5296, opened 2025-07-15
Skip mixer cache based on header

rohitkumarbhagat:
Modified cache behavior to aid in debugging mixer interactions. 

The same would be used to create uptime tests for website mixer

- We can't have direct uptime checks for website mixer, as endpoints are  not exposed publicly.
- Website uptime routing to website mixers would be needed. Disabling cache per request would help achieve that.

### DIFF
--- a/server/lib/cache.py
+++ b/server/lib/cache.py
@@ -15,6 +15,7 @@
 import os
 from pathlib import Path
 
+from flask import request
 from flask_caching import Cache
 
 import server.lib.config as lib_config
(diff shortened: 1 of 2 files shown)


server/lib/cache.py
```diff
@@ -65,3 +67,19 @@
 else:
   # For some instance with fast updated data, we may not want to use memcache.
   cache = Cache(config={'CACHE_TYPE': 'NullCache'})
+
+
+def should_skip_cache():
```
dwnoble, line 72: Could you add a test case for this decorator?

server/lib/cache.py
```diff
@@ -65,3 +67,19 @@
...
+  """Check if cache should be skipped based on request header.
+  
+  Returns:
+    True if X-Skip-Cache header is set to 'true', False otherwise.
+    Always returns False on any error to ensure caching remains functional.
+  """
+  try:
+    skip_cache_header = request.headers.get('X-Skip-Cache')
```
dwnoble, line 80: Nit: lowercase the value to support "True" or "TRUE"

server/lib/cache.py
```diff
@@ -65,3 +67,19 @@
...
+    return skip_cache_header is not None and skip_cache_header == 'true'
+  except Exception as e:
+    logging.debug("Error checking X-Skip-Cache header.", exc_info=True)
```
dwnoble, line 83: `debug` logs don't get logged in prod- maybe we should use "warning" instead?

dwnoble (approved):

PR #5314, opened 2025-07-21
[Explore] Update the Follow Up Questions component display logic.

javi-vaz:
- Avoid rendering the component when there are no related topics.
- Change the limit of questions shown from 10 to 6.


clincoln8 (approved):
Thanks Javi!

PR #5315, opened 2025-07-21
[Explore] Add component view to the Header Related Topics

javi-vaz:
- Added a GA event trigger to the initial view of the Related Topics in the Result Header
- Modified the FOLLOW_UP_QUESTIONS_VIEW event to RELATED_TOPICS_VIEW with a support for different modes (Follow Up Questions and Header Related Topics).


gmechali (approved):
Just checking, did you verify that behavior works? (i.e. by logging something when the component comes in view?)

clincoln8 (approved):
Awesome, thanks!

PR #5316, opened 2025-07-21
[Production Excellence] Screenshot diff testing with Percy

gmechali:
V0 of a script to take Snapshots using percy on different environments. 

More URLs to be added. This works on production but is failing in staging due to the redirection from the login that then enforces a Content Security Policy that is stricter, and prevents snapshotting the DOM.


dwnoble (approved):

gmechali:
Hey Dan, made some pretty big changes here to inclde all the urls from nl.json and page.json for the scripts. Removed the login since it's looking like we're going that route, added a requirements.txt etc. 

Please take another look!  thanks!

gmechali:
> Lgtm, but was the import update intentional

Resolved!

PR #5324, opened 2025-07-22
[Explore] Add Render GA Event to Follow Up Questions

javi-vaz:
Adds a GA event to track when the Follow Up Question component successfully renders the generated questions.
The parameters for the event are:
- Page Source
- Component


static/js/shared/ga_events.ts
```diff
@@ -291,6 +301,7 @@ export const GA_VALUE_TOOL_CHART_OPTION_SHOW_POPULATION_LOG =
 export const GA_VALUE_TOOL_CHART_OPTION_EDIT_SOURCES = "edit sources";
 export const GA_VALUE_TOOL_CHART_OPTION_FILTER_BY_POPULATION =
   "filter by population";
+export const GA_VALUE_PAGE_SOURCE_EXPLORE = "explore";
```
gmechali, line 304: GA uses the string values, so between this and the one below, it's the same thing. No need to duplicate it!

static/js/shared/ga_events.ts
```diff
@@ -34,6 +34,14 @@ export function triggerGAEvent(
  */
 export const GA_EVENT_PAGE_VIEW = "page_view";
 
+/**
+ * Triggered when a component to track is rendered.
+ * Parameters:
+ *  page_source : "explore",
+ *  component: "related_topics_generated_questions"
+ */
+export const GA_EVENT_COMPONENT_IMPRESSION = "component_impression";
```
gmechali, line 43: Are you basically trying to make this generic, where we can have an event "component_impression" sliced by which component it is? I do like that idea. 

If that's the case, can you move the onComponentRender function into this file and make it take a parameter for the component.

gmechali (changes requested):

static/js/shared/ga_events.ts
javi-vaz: That's true, is there a guideline for reusing values for different events? Like as long as the literal matches or should it conceptually refer to the same thing?

static/js/shared/ga_events.ts
javi-vaz: Sounds good, I'll move it over! The page should also be a parameter right? In case a component appears is reused in multiple pages?

static/js/shared/ga_events.ts
gmechali: If the string is the same, then either it's for the same thing, or the string isn't specific enough.
In this case, I think you can just ername this PAGE_EXPLORE = "explore"
and use the PAGE_EXPLORE everywhere.

clincoln8 (approved):

gmechali (approved):

PR #5328, opened 2025-07-22
[Disclaimers] Update the Disclamers on About page and fix redirections

gmechali:
Implement the redirection from /link/about to /about using the redirection.json.
This approach would previously reload the GCS file on every URL, I refactored it to load it once at server start up time.


beets (approved):
thank you for the quick change!

PR #5330, opened 2025-07-23
[Explore] Add new client type for generated questions

javi-vaz:
Adds a new client type to help track which queries succeed or fail from the Follow Up Questions component. This helps compile the types of failing queries Gemini would generate which can then be used to fine tune the prompt.


gmechali (approved):

clincoln8 (approved):
Thanks!

PR #5331, opened 2025-07-23
[Explore] Update prompt to generate more diverse questions

javi-vaz:
Changed the prompt to generate a variety of questions. This prompt was tested to have a 98% successful response rate with query context.


server/lib/nl/explore/gemini_prompts.py
```diff
@@ -49,19 +49,18 @@
...
 Avoid any other words apart from the generated follow up questions. Your entire response should only be the questions themselves, with each question on a new line.
-
 Avoid asking for data within a specific year or range of years. The questions should be timeless.
+Each question must be simple and focus on a single variable.
+Avoid questions that ask for a metric about places that meet a certain condition.
 
-Avoid questions that ask about correlations, relationships, or comparisons between the original question and the given topics. Each question must be simple and focus on a single variable.
-For example, the question "How does income equity compare to wealth equity in countries with the lowest Gini index?" should instead be a simpler version such as "What is the income equity across countries?"
+When generating the questions, have them be extremely varied. Have diversity in the words used.
+For inspiration, have the following question categories: ["Ranking","Maps","Comparison","Correlation","TimeDelta","Increase","Decrease"]
```
gmechali, line 63: I really think the best thing to do here is going to be give examples.

"For example, if you see topics such as GDP or Unemployment rate about the place \"California\", think to generate questions like \"How does California's GDP rank compared to other US States?\", or \"Has life expectancy in California increased in the past 20 years?\""

server/lib/nl/explore/gemini_prompts.py
```diff
@@ -49,19 +49,18 @@
 The original research question from the user  is: {initial_query}.
 The follow up questions should be based on the following list of topics and statistical variables for the same location.
 RELATED TOPICS START: {related_topics}. RELATED TOPICS END.
-If no related topics are given, do not return anything.
+CRUCIALLY, if no related topics are given, do not return anything.
```
gmechali, line 52: Kinda nit, but I hope that if no topics are returned we don't even call Gemini :)

javi-vaz: Yes, that is correct, we don't even render the component if there are no related topics.:)

server/lib/nl/explore/gemini_prompts.py
```diff
@@ -49,19 +49,18 @@
...
 
 Generate only one question per topic in the related topics list. Make them succinct, general, and simple in relation to the topic provided.
 
 Avoid any other words apart from the generated follow up questions. Your entire response should only be the questions themselves, with each question on a new line.
-
 Avoid asking for data within a specific year or range of years. The questions should be timeless.
+Each question must be simple and focus on a single variable.
```
gmechali, line 58: Also we should literally explain the use case here, telling Gemini these questions will be surfaced to users for additional questions to explore to view more data. And that we should make them engaging to users!

clincoln8 (approved):
The validation results seem promising enough for the experiment. Feel free to submit this as is for the staging cut/experiment and then continue iterating based on Gabe's comments!

javi-vaz:
To include this iteration's improvements in the experiment, we'll submit the PR as is and in future updates we'll address the feedback received.

PR #5342, opened 2025-07-25
[Explore] Fix redirect link to avoid missed context on first click

javi-vaz:
tldr: Changed redirect link to match the NL search bar to preserve context

The redirect link in the Follow Up Questions component would send users to `/explore/` instead of `/explore` used by the NL search bar. This treats the query as a new page and restarts the context which leads to more erroneous calls because it won't have a past place source. While it is functional in the following clicks, having a higher error rate in the first click is detrimental to the user experience.


clincoln8 (approved):
Thanks for the fix, great catch!!

PR #5344, opened 2025-07-27
Block-level dataset selector (bar)

nick-nlb:
## Description

This PR implements facet selection on the NL pages for bar charts. 

This is the first time that bar tiles have had facet selection applied to them (as they are not part of the vis tool set). This required the facet selector dialog to be updated in order to allow a "grouped facet selection" mode, whereby we show a single facet selection list for all stat vars (rather than having each stat vars selection list chosen individually, as we would for other charts).

The PR also hardens some more general aspects of the block-level facet selection by:
- ensuring that the facet selector only appears when all tiles are compatible (i.e., line/highlight, but not line/bar) - Although I have never observed this situation, we now make sure that if it does happen, a faulty facet selector will not appear. 
- making the determination of whether to use places or enclosed places for facet selection more rigorous, using the precise logic of the tile type. This prevents an incorrect facet list from ever showing for a chart.

We have also added a debugging marker that will tell us if a chart has inconsistent facets. This flag only shows if in "debug mode" (using the same logic as the debug icon used for search results).

This PR also contains improvements to the tooltip and dialog component and some incidental updates to the JSDoc comments in data_fetch_utils to make them consistent with their signatures (this file is not otherwise changed but I noticed these when looking at it).

## Notes

To facilitate the new "grouped facet selection" mode, we refactored the facet selection dialog, breaking it up into smaller components that can be shared among the two modes, and in some cases converting JSX-returning functions into full React components.

Many of the files touched in this PR were only changed to update the import for the new location of the facet selector. 

### Bar Chart Data Fetching

During an earlier iteration of the development of this PR, we were planning to allow stat vars to have their facets chosen individually (rather than grouping as we are now). To do this, we needed to be able to supply different facets for each bar (stat var). However, the bar tile did not allow for data fetching with stat-var-specific facets. 

To that end, we updated the `fetchData`, allowing individual stat vars to have their own facets, doing so in a way that groups the calls by facet in order to be as efficient as possible. 

Since then, the specifications have changed, and we no longer have individual facets per stat var. However, we have kept in this change in the PR, in anticipation of this being needed later.

### All facets consistent

Screenshot 2025-07-26 at 8 44 49 PM

### Some facets inconsistent

Screenshot 2025-07-27 at 2 35 38 PM

### Debug Flag

Screenshot 2025-07-26 at 8 52 13 PM

## Testing

A page that has a bar chart with inconsistent facet choices (the bar chart at the top).

http://localhost:8080/explore#q=age+distribution+in+the+United+States

A page that has bar charts with the same facet choices (the bar chart at the top):

http://localhost:8080/explore#q=jobs%20in%20texas


static/js/shared/facet_selector/facet_selector_rich.tsx
```diff
@@ -0,0 +1,344 @@
...
+      </Button>
+      {showInconsistentFacetFlag && (
+        <div
+          css={css`
+            display: inline-flex;
+            padding-left: ${theme.spacing.xs}px;
+            transform: translateY(-1px);
+          `}
+        >
+          <DebugFlag message="This chart’s facet choices are not consistent across all statistical variables." />
```
dwnoble, line 190: Could you add some language here to make it clear to our team that this flag is only shown in debug mode? Maybe just something like "[debug] This chart's..."

static/js/components/subject_page/block.tsx
```diff
@@ -136,11 +136,18 @@ export interface BlockPropType {
...
+  new Set(["LINE", "HIGHLIGHT"]),
+  new Set(["SCATTER"]),
+  new Set(["BAR"]),
+];
+
+const FACET_ELIGIBLE_TILES = new Set(
+  _.flatten(FACET_ELIGIBLE_TILE_GROUPS.map((group) => Array.from(group)))
+);
+
+const FACET_GROUPING_ELIGIBLE_TILES = new Set(["BAR"]);
```
dwnoble, line 150: Could you add a test case to ensure the facet selector shows up for bar charts?

dwnoble (commented):
@nick-next & @pablonoel - this looks great, thanks for the addition and for the refactor.

static/js/components/subject_page/block.tsx
dwnoble: Thanks Nick!

PR #5351, opened 2025-07-28
[Vis Tools] Map vis tool selective display

nick-nlb:
## Description

This PR updates the old map vis tool so that auxiliary information (instructions, etc) appear conditionally based on whether the chart is ready to start loading and display.

## Notes

We are now conditionally rendering the `<ChartLoader />` component, preventing it from mounting until the required inputs (place and stat var) are selected. 

While `<ChartLoader />` is robust enough to render nothing when idle, this prevents it from mounting entirely, avoiding the need to activate its complex rendering lifecycle (for performance) until ready.

### Before

Screenshot 2025-07-28 at 4 46 36 PM

### After

Screenshot 2025-07-28 at 4 52 42 PM

### DIFF
--- a/static/js/tools/map/app.tsx
+++ b/static/js/tools/map/app.tsx
@@ -86,7 +86,11 @@ function App(): ReactElement {
           </Row>
           {showInstructions && (
             <Row>
-              {useStandardizedUi ? <VisToolInstructionsBox /> : <Info />}
+              {useStandardizedUi ? (
+                <VisToolInstructionsBox toolType="map" />
+              ) : (
+                <Info />
+              )}
             </Row>
           )}
           {showChart && (


static/js/tools/map/app.tsx
```diff
@@ -79,10 +85,16 @@ function App(): ReactElement {
...
           </Row>
-          <Row>{useStandardizedUi ? <VisToolInstructionsBox /> : <Info />}</Row>
-          <Row id="chart-row">
-            <ChartLoader />
-          </Row>
+          {showInstructions && (
+            <Row>
+              {useStandardizedUi ? <VisToolInstructionsBox /> : <Info />}
+            </Row>
+          )}
```
juliawu, line 92: You'll probably get merge conflicts with these lines I just submitted: https://github.com/datacommonsorg/website/blob/dc528274a51601af92e68f8353b0f71c12ddb14d/static/js/tools/map/app.tsx#L85-L91

Between the two, I prefer yours both for readability and so the `showInstructions` logic also gates the prod facing `<Info>` component.

juliawu (approved):
Thanks for the fix!

PR #5359, opened 2025-08-01
[Explore] Initial implementation of Page Overview component

javi-vaz:
The initial version of the Page Overview Component, displaying an introduction to the charts shown in the Explore page. Additionally, blue substrings of the introduction are links that scroll to the specific charts the text is mentioning (see After screenshot).  

Before: https://screenshot.googleplex.com/xkKaBHauscB2h5x
After: https://screenshot.googleplex.com/8FrRXR5YSubLfFE

The contents of the introduction is generated by Gemini 2.5 Flash-Lite, and it also marks the text with `<` and `>` to indicate it should link to a chart From there, the links are created if the substring matches to a corresponding chart to avoid incorrect links. If the syntax of the marking is incorrect (ie incomplete pairs of angle brackets), then no link is created and the regular text is shown without angle brackets.


static/js/apps/explore/page_overview.tsx
```diff
@@ -0,0 +1,174 @@
...
+        statVarIdx.natural_language,
+        statVarIdx.stat_var_title
+      );
+    }
+    const titleToCatNBlockIndex = new Map<string, ChartIndex>();
+    for (const statVarChart of statVars) {
+      titleToCatNBlockIndex.set(statVarChart.title, statVarChart);
+    }
+
+    const splitOverview = preprocessedOverview.split(SPLIT_LINKS);
```
clincoln8, line 116: TIL including adding capture group to regex split delimiter causes it to appear in the output!

nit: add a comment explaining this, (ie the delimiters or more specifically, annotated stat var strings are included in the final array) because it might not be intuitive for most readers (unless I'm just out of the loop)

static/js/apps/explore/page_overview.tsx
```diff
@@ -0,0 +1,174 @@
...
+    for (const statVarIdx of preprocessedStatVarIndices) {
+      substringsToStatVarIndex.set(
+        statVarIdx.natural_language,
+        statVarIdx.stat_var_title
+      );
+    }
+    const titleToCatNBlockIndex = new Map<string, ChartIndex>();
+    for (const statVarChart of statVars) {
+      titleToCatNBlockIndex.set(statVarChart.title, statVarChart);
+    }
```
clincoln8, line 114: consider renaming; another option might be statVarTitleToChartConfig or statVarTitleToPageLocation
```suggestion
    const statVarTitleToChartLocation = new Map(
  statVarChartLocations(statVarChart => [
    statVarChart.title,
   statVarChart
  ])
);
```

static/js/apps/explore/page_overview.tsx
```diff
@@ -0,0 +1,174 @@
...
+    // Create Maps to speed up string look ups
+    const preprocessedStatVarIndices = resp.data.stat_var_links;
+    const substringsToStatVarIndex = new Map<string, string>();
+    for (const statVarIdx of preprocessedStatVarIndices) {
+      substringsToStatVarIndex.set(
+        statVarIdx.natural_language,
+        statVarIdx.stat_var_title
+      );
+    }
```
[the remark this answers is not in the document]
javi-vaz: I went with `statVarOverviewExcerptsToTitle` to indicate that they're pieces of the overview that refer to stat vars. Let me know if you're still inclined for `statVarRawStringToTitle`.

static/js/apps/explore/page_overview.tsx
javi-vaz: Done. I added it through a map function which I believe is what you're referring to.

static/js/apps/explore/page_overview.tsx
javi-vaz: Done.

static/js/apps/explore/page_overview.tsx
```diff
@@ -0,0 +1,174 @@
...
+    const titleToCatNBlockIndex = new Map<string, ChartIndex>();
+    for (const statVarChart of statVars) {
+      titleToCatNBlockIndex.set(statVarChart.title, statVarChart);
+    }
+
+    const splitOverview = preprocessedOverview.split(SPLIT_LINKS);
+    return splitOverview.map((part, index) => {
```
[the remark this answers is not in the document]
javi-vaz: There appeared to be no warning when running the lint fixer, is that what you're referring to when 'running npm' or is it something else?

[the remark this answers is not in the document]
gmechali: It's when you run ./run_tests.sh -f 

This will autofix linting errors it knows how to fix, but will just call out warnings for others. This will definitely be added, however we do already have nearly 100 warnings. We took it down last year from like 900. Not letting it get back up is the new goal :)
+1 to christie's point, please add the types on all variables and function parameters!

gmechali (approved):

PR #5363, opened 2025-08-02
Designate smoke tests for Custom DC

hqpho:
The full suite of tests against autopush are frequently flaky and also somewhat redundant now that tests run before merge on each website PR. This will allow us to keep just a handful of them to verify basic functionality in Custom DC autopush after deployment.


dwnoble (commented):
Thanks Hannah!

When will we be running the smoke tests?

hqpho:
> Thanks Hannah!
> 
> When will we be running the smoke tests?

My plan is to modify the Custom DC autopush build trigger (in the deployment repo) to run just the smoke tests instead of all the tests against Custom DC autopush after each autopush deployment.

hqpho:
> Got it. Is it necessary to run tests at all considering the custom DC tests should have already passed by this point? or are there cases where we might still run into failing tests on the custom dc autopush deploy?

The autopush setup has actual custom data loaded so it's slightly different from the pre-commit run, so I think it's worth keeping these tests around. With this change, hopefully failing tests will be more signal than noise.

dwnoble:
That sounds good to me. Thank you!

dwnoble (approved):

PR #5367, opened 2025-08-04
Block-level dataset selector (map)

nick-nlb:
## Description

This PR adds the facet selector to the map and ranking charts in the natural language search results.

Because we are able to now do both facet selection and “Snap to date highest coverage”, the final specification is that the “Snap to date with highest coverage” stays where it is in the header of the chart, and the facet selector appears next to it.

### API

In order to allow facet selection and snapping to highest coverage to work together, we had to update the API. Previously, the function used determined the date of highest coverage was determining that date before applying the requested facet filtering (i.e, using all facets rather than the requested ones to determine that date). This meant that the actual facet we filtered for often would not have data for that particular year, causing an error.

The API update involves the following:

1. The proximate cause of the error in `_get_recent_date_counts` is now hardened so that a lack of data will no longer trigger an error. 
2. Most importantly, the filtering of the series_dates_response by facet is now done in (separately in a new function) and we now look for the date of highest coverage only in the requested facets. If no facets are selected, the result is functionality identical to before.

### Highest Coverage in Front End

The checkbox to allow snapping to highest coverage is selectively enabled or disabled based on the results of the function `highestCoverageDatesEqualLatestDates`, which makes two calls to the API. These calls needed to be updated to allow for the sending of the selected facet, and so they were updated to use the `getPointWithin` from `data_fetch_utils.ts` as these allow for passing in facets.

### Refactoring

The `map_config.tsx` file is refactored to use the new utility functions to fetch facet choices. This parallels similar refactoring we did to line and scatter in earlier PRs.
 
### Tooltip

The tooltip was updated to the new tooltip component (and the related messages moved into one of the i18n messages files).

## Screenshot

Screenshot 2025-08-04 at 11 57 54 AM

## Tests

Two new web driver tests have been added with this PR:
- a test to verify that the facet selector appears on map charts, it contains the expected number of options, and that changing a facet changes the source.
- a test to verify that snapping to date of highest coverage updates the date range used as expected.

## Testing

The following page contains a range of map charts with various different combinations of scenarios:
- http://localhost:8080/explore#q=age+distribution+in+the+united+states+by+state

As an example of how snapping-to-highest-coverage can vary across facets, in the first chart (Population: 1-4 Years in States of United States (2021)):
- with no facets selected, snap-to-highest-coverage can be toggled. 
- with the second facet choice (American Community Survey (ACS), 2010-2021), it cannot be toggled.
- with the first option (Wonder: Mortality, Underlying Cause Of Death) it can be toggled (and you can see how a certain date, not the latest, has the most coverage).

### DIFF
--- a/server/webdriver/tests/explore_test.py
+++ b/server/webdriver/tests/explore_test.py
@@ -83,9 +83,9 @@ class TestExplorePage(ExplorePageTestMixin, BaseDcWebdriverTest):
                          "Could not find the 'Age Distribution' chart block.")
 
     original_source_text = find_elem(chart_block, By.CLASS_NAME, 'sources').text
-    self.assertEqual(
-        original_source_text,
-        'Sources: data.census.gov, census.gov, data.census.gov • Show metadata')
+    self.assertIn('census.gov', original_source_text)
+    self.assertIn('data.census.gov', original_source_text)
+    self.assertIn('Show metadata', original_source_text)
 
     # Click on the button to open the facet selector modal
     facet_button = find_elem(chart_block, By.CLASS_NAME,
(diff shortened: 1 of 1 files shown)


server/webdriver/tests/explore_test.py
```diff
@@ -106,10 +107,123 @@ def test_bar_select_different_facet(self):
...
+    )
+
+    static_header_text = "Population: 1-4 Years in States of United States"
+    wait_for_text(chart_block, static_header_text, By.TAG_NAME, 'h4')
+    initial_header = find_elem(chart_block, By.TAG_NAME, 'h4')
+
+    if not initial_header:
+      self.fail("The chart's inner header (h4) did not appear.")
+
+    self.assertIn("(2021)", initial_header.text,
```
dwnoble, line 207: This test might break as we get new data ingested. It's also not a great solution, but could you change this test to look for a patterns like `/\(\d+\)/` and `/\(\d+ to \d+\)/` respectively? (unless you have any other creative testing strategies which im totally open to)

dwnoble (commented):
@nick-next & @pablonoel this looks really great. Just clicking around a little has  helped me understand the breadth of data available.

server/webdriver/tests/explore_test.py
dwnoble: what do you think

PR #5370, opened 2025-08-06
[Vis Tools] Propogate query parameters in map tool URLs

juliawu:
The map tool currently clears out query parameters from the URL whenever a place/place type/stat var is selected. This results in feature flag url overrides disappearing whenever the user interacts with the tool. 

This PR updates the hash parameter update logic to keep any existing query parameters.

Testing strategy:

1. Spin up a server with `./run_server.sh` and go to `localhost:8080/tools/map?disable_feature=standardized_vis_tool` in your browser.
2. Enter a place, place type, and a stat var into the tool.
3. See that the `disable_feature=standardized_vis_tool` is propagated in the URL.

### DIFF
--- a/static/js/tools/map/app.tsx
+++ b/static/js/tools/map/app.tsx
@@ -148,17 +148,20 @@ function updateHash(context: ContextType): void {
   let hash = updateHashStatVar("", context.statVar.value);
   hash = updateHashPlaceInfo(hash, context.placeInfo.value);
   hash = updateHashDisplay(hash, context.display.value);
-  // leaflet flag is part of the search arguments instead of hash, so need to
-  // update that separately
-  // TODO: forward along all args and then append hash in the url.
-  let args = "";
+  const args = new URLSearchParams(location.search);
   if (context.display.value.allowLeaflet) {
-    args += `?${ALLOW_LEAFLET_URL_ARG}=1`;
+    // leaflet flag is part of the search arguments instead of hash, so need to
+    // update that separately
+    args.set(`${ALLOW_LEAFLET_URL_ARG}`, "1");
   }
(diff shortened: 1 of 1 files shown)


dwnoble (approved):
Thanks Julia!

gmechali:
Could we try to reuse a util from this file?
https://github.com/datacommonsorg/website/blob/8ee5498a64793930a782c9cf1d611f31ca81e34d/static/js/utils/url_utils.ts#L27
Or create one that we can reuse in the future!

juliawu:
> Could we try to reuse a util from this file?
> 
> https://github.com/datacommonsorg/website/blob/8ee5498a64793930a782c9cf1d611f31ca81e34d/static/js/utils/url_utils.ts#L27
> 
> 
> Or create one that we can reuse in the future!

Sounds good to me. Actually, all 3 tools could use a revamp of their url update logic. We are already planning to move to query params and away from hash parameters as part of the project. If it's alright with you, I'll merge this change for now so it's not so annoying to debug map tool changes and leave switching to url utils for a later PR when I switch urls to query parameters.

gmechali:
> > Could we try to reuse a util from this file?
> > https://github.com/datacommonsorg/website/blob/8ee5498a64793930a782c9cf1d611f31ca81e34d/static/js/utils/url_utils.ts#L27
> > 
> > Or create one that we can reuse in the future!
> 
> Sounds good to me. Actually, all 3 tools could use a revamp of their url update logic. We are already planning to move to query params and away from hash parameters as part of the project. If it's alright with you, I'll merge this change for now so it's not so annoying to debug map tool changes and leave switching to url utils for a later PR when I switch urls to query parameters.

sounds good, thank you!

PR #5371, opened 2025-08-06
Fix Custom DC Webdriver Test Errors

juliawu:
`flask_cdc_webdriver_test` tests have been failing with
```
RuntimeError: Failed to import transformers.models.auto.modeling_auto because of the following error:
Failed to import transformers.generation.utils because of the following error:
module 'torch' has no attribute 'uint64'
```

The fix:
1. Pin `numpy==1.26.4` to fix the version conflict error @javi-vaz mentions in the comments. This is the latest numpy version before they moved to v2.0.0.
2. Upgrade torch to `torch==2.8.0`, the latest version, which includes `uint64`.
3. Upgrade torchvision to `torchvision==0.23`, the latest version, because the previously used v0.17.2 required torch to be pinned at v2.2.2 (the version giving us issues).

### DIFF
--- a/nl_requirements.txt
+++ b/nl_requirements.txt
@@ -7,7 +7,7 @@ pandas==2.1.1
 scikit-learn==1.5.0
 sentence-transformers==2.2.2
 spacy==3.7.4
-torchvision==0.17.2
+torchvision==0.23.0
 # TODO: this is pinned because latest huggingface_hub is not compatible with
 # sentence-transformers v2.2.2. Look into upgrading sentence-transformers to
 # v2.3.0 or newer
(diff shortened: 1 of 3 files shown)


javi-vaz:
Could it be related to this? https://github.com/huggingface/transformers/issues/36343 

I notice that when I do setup_python, there seems to be a warning with numpy. 
image

juliawu:
> Could it be related to this? [huggingface/transformers#36343](https://github.com/huggingface/transformers/issues/36343)
> 
> I notice that when I do setup_python, there seems to be a warning with numpy. image

Great find! I'll try and see if downgrading `numpy` fixes things.

juliawu:
Closing because more progress is being made in #5377

PR #5373, opened 2025-08-06
source filtering for vai

alyssasuiyiguo:


server/services/datacommons.py
```diff
@@ -509,6 +509,14 @@ def search_statvar(query, places, sv_only):
   })
 
 
+def filter_statvars(stat_vars, places):
```
gmechali, line 512: here and everywhere, this isn't just places right? It s all entities?

Can you name it accordingly? "entities" is probably fine.

gmechali (approved):
Just a nit on naming, but looks great otherwise!

PR #5377, opened 2025-08-07
Upgrade Package dependencies 

gmechali:


build/cdc_data/Dockerfile
```diff
@@ -52,7 +52,7 @@ ENV PATH="/workspace/venv/bin:$PATH"
 RUN pip3 install -r ./import/simple/requirements.txt
 
 # Install data requirements.
-RUN pip3 install torch==2.2.2 --extra-index-url https://download.pytorch.org/whl/cpu \
+RUN pip3 install torch==2.8.0 --extra-index-url https://download.pytorch.org/whl/cpu \
```
dwnoble, line 55: @hqpho @keyurva  do you know if we have automated tests for the cdc data script to verify this won't break anything?

gmechali: Actually looks like we may need to use 2.7.0

juliawu: +1, torch upgrade may not be needed, because the root import errors were coming from sentence-transformers.

PR #5378, opened 2025-08-07
Fix dataset selector bug

alyssasuiyiguo:
Fixes a [bug](https://buganizer.corp.google.com/issues/433923263) in the stat var dataset selector where if a dataset was selected, both the source and the dataset would be cleared.

[Demo of fix](http://screencast/cast/NjUyNTgwNzAzOTA4NjU5Mnw4NzBmNzA1My0yYQ)


dwnoble (commented):
Thanks Alyssa! could you add a unit test for this?

dwnoble (approved):
Thanks @alyssasuiyiguo !

PR #5401, opened 2025-08-15
API code dialog

nick-nlb:
## Description

This PR implements the API code dialog that displays corresponding REST V2 API endpoints for the given chart. The dialog is introduced on page 28 of the [Rolling UI deck](https://docs.google.com/presentation/d/1ztyZwu8b3YzU1MMAFPI3iC-Hm00CE63rPb7eP6vGZlw/edit?pli=1&slide=id.g33b18c2fc5a_1_69#slide=id.g33b18c2fc5a_1_69).

The API dialog displays the endpoints necessary to pull the data displayed in the chart, and allows the user to copy those endpoint code snippets to the clipboard.  

## Design Considerations

We have the following two key considerations:

* The construction of endpoints is tile-specific, with each tile type responsible for defining how its own REST V2 API endpoints are constructed. This is similar in principle to the existing `getDataCSV` callback.
* The API dialog itself should be free to render the endpoint in any format. We currently render it in `curl`, but this could later be GET, Python, etc. In other words, the creation of endpoint definitions on the tile level and the conversion of those to their final displayed endpoint should be decoupled.

To accomplish this, the PR introduces the concept of an _observation spec_ to represent a particular call to the V2 `observation` endpoint. The observation spec is a JS object that contains enough information (date, stat vars, filters) to later construct the endpoint. Given a particular spec, we can render the endpoint call in a variety of formats.

Each tile type provides a callback that returns a list of observation specs for a particular chart. This callback is passed to the API dialog/button in the chart footer (following the pattern of the existing `getDataCSV` download callback).

The API dialog then converts the observation specs into actual endpoint code (currently curl), using a conversion function provided in the observation spec library file.

## Notes

Unlike the CSV data download, there are not always simple REST V2 API endpoints that correspond to exactly the data as shown in the chart. In those cases, we always give the user endpoints that will at least contain all the data used to construct the chart.

As an example, there is no single endpoint that will deliver per capita results. Instead, we provide two separate endpoints to the user. One will give the standard numerator data, and the other will give the user the denominator values. The user can use that to calculate per capital values after retrieving the data.

### Line chart

line_chart

### Line chart - stat vars with different facets (necessitating two calls).

map_chart_different_facets

### Scatter chart - per capita

scatter_chart_per_capita

### Map chart

map_chart

## Tests

This PR adds a web driver test to verify the existence of the API button, open the API dialog and verify that the endpoint contains expected key parts.

## Testing

The API code dialog can be viewed for all line, scatter, map, bar and highlight charts in both the NL search and the new viz tools. One way to test the endpoints themselves is to use a program like Postman (where you can paste the curl call directly and run it).

## Feature Flag

This functionality is behind the `metadata_modal` feature flag.

### DIFF
--- a/static/js/tools/stat_var/dataset_selector.test.tsx
+++ b/static/js/tools/stat_var/dataset_selector.test.tsx
@@ -1,0 +1,84 @@
+/**
+ * @jest-environment jsdom
+ */
+// Copyright 2025 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import { jest } from "@jest/globals";
+import Adapter from "@wojtekmaj/enzyme-adapter-react-17";
+import Enzyme, { shallow } from "enzyme";
+import React from "react";
+
+import { updateHash } from "../../utils/url_utils";
+import { DatasetSelector } from "./dataset_selector";
+
+Enzyme.configure({ adapter: new Adapter() });
+
+// Mock the updateHash function
+jest.mock("../../utils/url_utils", () => ({
+  updateHash: jest.fn(),
+}));
+
+const mockProps = {
+  dataset: "dataset1",
+  datasets: [
+    { dcid: "dataset1", name: "Dataset 1" },
+    { dcid: "dataset2", name: "Dataset 2" },
+  ],
+  source: "source1",
+  sources: [
+    { dcid: "source1", name: "Source 1" },
+    { dcid: "source2", name: "Source 2" },
+  ],
+};
+
+describe("DatasetSelector", () => {
+  beforeEach(() => {
+    // Clear mock calls before each test
+    (updateHash as jest.Mock).mockClear();
+  });
+
+  it("renders the component with initial props", () => {
+    const wrapper = shallow(<DatasetSelector {...mockProps} />);
+    expect(
+      wrapper.find("#dataset-selector-source-custom-input").prop("value")
+    ).toEqual("source1");
+    expect(
+      wrapper.find("#dataset-selector-dataset-custom-input").prop("value")
+    ).toEqual("dataset1");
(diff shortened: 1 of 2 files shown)


static/js/shared/observation_specs.ts
```diff
@@ -0,0 +1,311 @@
...
+ *
+ * @param spec The observation specification to convert.
+ * @param apiRoot The root URL for the Data Commons API.
+ * @returns A formatted cURL command string.
+ */
+export function observationSpecToCurl(
+  spec: ObservationSpec,
+  apiRoot?: string
+): string {
+  const apiUrl = `${(apiRoot || DEFAULT_API_ROOT).replace(
```
dwnoble, line 236: `apiRoot` might not be the right value to use for the apiUrl. In the web component api, apiRoot is set to the "website api", not the v2 api root.

See this JSFiddle for an example (i had to edit the code to force the experiment to be enabled in chart_footer.tsx): https://jsfiddle.net/yo5vLu4n/ 

it's pretty hairy, but if an apiRoot url is set, i think it should go something like:
- if apiRoot is set, that is a custom DC website URl
- For custom DCs, the "v2" api root is <apiRoot>/core/api.  For example: https://undata-prod-datacommons-web-service-107186651273.us-central1.run.app/core/api/v2/node?nodes=dc/topic/undata/sdg_2.4.1&property=-%3E*
- Additionally, if apiRoot is set, no API key is necessary

Screenshot 2025-08-20 at 4 06 30 PM

static/js/shared/observation_specs.ts
```diff
@@ -0,0 +1,311 @@
...
+        .join(",")}]}`
+    );
+  }
+
+  const jsonBody = params.map((p) => `    ${p}`).join(",\n");
+  const dataPayload = `-d '{\n${jsonBody}\n  }'`;
+
+  return [
+    `curl -X POST \\`,
+    `  -H "X-API-Key: {API_KEY}" \\`,
```
dwnoble, line 285: from the mocks, I think this should be ${API_ROOT}. that way users can do something like

```
export API_KEY=their-key
```

then run the command directly

dwnoble: Also, I think it's worth mentioning in the copy that users should replace ${API_KEY} with their actual api key, or export the variable. Wdyt @miss-o-soup ?
Screenshot 2025-08-20 at 4 10 16 PM

dwnoble (commented):
It's looking great!

I like this functionality for a v1. For a v2, we should consider:
(1) Adding an "API Code" button to the highlight tile
(2) Right now the curl request just fetches the observation values, but it doesn't get all of the values from the CSV export in the "Download" modal like the place names, place iso codes, or variable names. Should we consider adding these down the road?

cc @miss-o-soup

dwnoble (approved):
Thank you!!

PR #5404, opened 2025-08-17
Webdriver Test - Stat var hierarchy filter

nick-nlb:
## Description

The stat var hierarchy's `test_filter` test was failing intermittently.

### The issue

The reason for the intermittent failures was that, after the test selected "County", it attempted to wait for the loader to disappear (indicating that the new filtered values were ready). However, `wait_for_loading` was checking against a loader with id "screen", whereas the loader in the stat var hierarchy tool had an id of "browser-screen".

Because of this, the test continued right away, without waiting for the call to resolve and the component to update. Usually, especially locally, this still resulted in a passing test because of the speed with which the updates happened. However, if the update takes time, the race condition results in a failing test.

### Solution

To solve this, we are now looking for the change in the content of the relevant section of the hierarchy, rather than the status of the spinner.

Only once the text of the "Agriculture" element changes, we check the numbers and make sure the number has gone down.

## Other minor updates

Besides the direct fix, this PR makes other minor updates for clarity of identifiers and simplifying how the regex retrieves the numbers.  It also follows a suggestion by the auto review to simplify and reuse the selector path. This is a change from the original method used, but it does simplify the test's code.

### DIFF
--- a/server/lib/nl/explore/gemini_prompts.py
+++ b/server/lib/nl/explore/gemini_prompts.py
@@ -77,7 +77,7 @@ The available statistical variables are the following: {stat_var_titles}
 Do not include ```json ``` in the output.
 """
 
-RELATED_QUESTIONS_PROMPT = """
+FOLLOW_UP_QUESTIONS_PROMPT = """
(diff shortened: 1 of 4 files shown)


dwnoble (approved):
Nice find nick!

PR #5421, opened 2025-08-20
[CSS fix]

gmechali:
Should not have added this max-width since it interferes with mobile width


clincoln8 (approved):
Thanks for the fix!

PR #5428, opened 2025-08-21
(UN) Fixed map tooltip overflow issue

dwnoble:
Fixed issue where long variable names were overflowing out of the tooltip on the map

### Before:
tooltip-before

### After:
tooltip-after


beets (commented):
i'm also a little confused about how this is the fix (the extra semi-colon and leading 0's should be fixed by the css compiler). i tried this on the explore page on autopush. 

is there another file missing, or is this specific to custom dc?

dwnoble:
> i'm also a little confused about how this is the fix (the extra semi-colon and leading 0's should be fixed by the css compiler). i tried this on the explore page on autopush.
> 
> is there another file missing, or is this specific to custom dc?

Whoops i think i just didn't check in my changes 😅  - added them now.

beets (approved):

PR #5429, opened 2025-08-22
Include current day's results in screenshot endpoint by default

hqpho:
The `server/lib/gcs.list_folder` function uses `end_offset` which is exclusive in the GCS API. This caused an issue in the screenshot viewer where items from the specified end date (which defaults to today) were not being included in the results.

This change addresses the issue by modifying the screenshot endpoint and `list_folder` to make `end_offset` optional and omit it from the call to GCS when it's not provided explicity.

### DIFF
--- a/server/lib/gcs.py
+++ b/server/lib/gcs.py
@@ -13,6 +13,8 @@
 # limitations under the License.
 
 import collections
+from datetime import datetime
+from datetime import timedelta
 
 from google.cloud import storage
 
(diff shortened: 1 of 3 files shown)


server/lib/gcs.py
```diff
@@ -59,10 +61,22 @@ def list_folder(bucket_name, prefix, start_offset='', end_offset=''):
   storage_client = storage.Client()
   bucket = storage_client.get_bucket(bucket_name)
   start_offset = prefix + '/' + start_offset
-  end_offset = prefix + '/' + end_offset
+  final_end_offset = None
+  if end_offset:
+    try:
+      # The end_offset parameter is exclusive. To make it inclusive, we add one
+      # day to the end_offset date.
```
juliawu, line 68: Please add a docstring to `list_folder()` that mentions the end_offset argument is now inclusive to prevent future confusion.

server/lib/gcs.py
```diff
@@ -59,10 +61,22 @@ def list_folder(bucket_name, prefix, start_offset='', end_offset=''):
...
+      end_date = datetime.strptime(end_offset, '%Y_%m_%d')
+      next_day = end_date + timedelta(days=1)
+      final_end_offset = prefix + '/' + next_day.strftime('%Y_%m_%d')
+    except ValueError:
+      # Consider logging a warning that end_offset has an invalid format.
+      # `final_end_offset` will remain `None` as initialized before.
```
juliawu, line 74: I like the idea of adding a warning about invalid format. Could you add a quick logging statement here?

juliawu (approved):
Thanks for the fix! Just some quick comments.

Just out of curiosity, why doesn't omitting `end_date` in server/routes/screenshot/html.py work?

hqpho:
> Thanks for the fix! Just some quick comments.
> 
> Just out of curiosity, why doesn't omitting `end_date` in server/routes/screenshot/html.py work?

It does! Thanks for the suggestion!

server/lib/gcs.py
hqpho: I'll do you one better, modified to return a 400 if either start_date or end_date is in the wrong format.

server/lib/gcs.py
hqpho: Changed back to being exclusive and added explanatory text in the UI.

PR #5432, opened 2025-08-22
Explore Page Results Header

nick-nlb:
## Description

This PR implements the new results header component on the explore page that appears on a successful search

This is introduced on page 69 of the [Rolling UX deck](https://docs.google.com/presentation/d/1ztyZwu8b3YzU1MMAFPI3iC-Hm00CE63rPb7eP6vGZlw/edit?pli=1&slide=id.g33c056d52a1_2_127#slide=id.g33c056d52a1_2_127).

We display the place (and a list of the parent places) and provide a link to the knowledge graph (similar to the place overview page). 

We are also now showing the searched query on the page.

## Design Considerations

The final specifications differ somewhat from the UX mockup. 

* Rather than display only the place name, we display the related places, similar to the place overview page.
* However, unlike the place overview page, we truncate the related places. This is described below.
* When we have more than one place, we display a general message with a tooltip menu that shows all of the places.
* Unlike the single place view, we don't display "State in", "Country in", due to space constraints. 

### Truncation rules

* When the place is smaller than a country (i.e., a city, county, etc) we truncate at the country: "Texas, State in United States of America", "Toronto, City in Ontario, Canada".
* When the place is a country, we truncate at the continent: "United States of America, Country in North America".
* Otherwise we don't truncate (for continents, world).

## Notes

This PR involved a fairly substantial overhaul of the `ResultHeaderSection` component. Because the new functionality is behind a feature flag, the original component is now called `ResultHeaderSectionLegacy`. The feature flag determines which is displayed. The legacy component will be removed when the feature flag is dropped.

It is still not yet determined if "State" will be capitalized, or if we will lowercase them to "state". The latter is preferable but will require translations of each place type in the various languages.

### Single Place

Screenshot 2025-08-22 at 8 47 05 AM

Screenshot 2025-08-22 at 8 48 27 AM

Screenshot 2025-08-22 at 8 48 58 AM

### Multiple Places

Screenshot 2025-08-22 at 8 49 44 AM

## Feature Flag

This functionality is behind the `explore_result_header` feature flag, enabled in `local` and `autopush`.

## Tests

Several of the explore page tests were affected by these updates. This is complicated by the fact that the tests can run in both environments where the feature flag is set and where it isn't.

I've added in a new utility class in the tests that allow us to select an element any one of a list (usually a pair) of selectors. This allows us to distinguish which version of the the feature-flagged page we are looking at, and test differently based on that. The tests now pass in both environments.

Unrelated to this PR but related to one of the updated tests, I've also changed one of the tests to skip (rather than fail) when the topics list doesn't appear. The topic list doesn't show a certain percentage of the time (due to a weighted feature flag) and so this test was failing a certain percentage of the time.


server/webdriver/base_utils.py
```diff
@@ -112,6 +113,32 @@ def find_elem(
   return elems[0] if elems else None
 
 
+def find_any_of_elems(
```
dwnoble, line 116: Nice!

static/js/i18n/i18n_messages.ts
```diff
@@ -168,4 +168,19 @@ export const messages = defineMessages({
     description:
       "Text to display when we have a single place, and need to add the 'in' before the place. For example, 'in California'.",
   },
+  allAbout: {
+    id: "all_about",
+    defaultMessage: "All about",
```
dwnoble, line 173: This might be better suited as "All about {place}", in case a translation would prefer to re-order the "all about" translation after the place description

static/js/apps/explore/result_header_section.tsx
```diff
@@ -56,24 +351,96 @@ export function ResultHeaderSection(
...
+        const parentPlacesMap: Record<string, NamedTypedPlace[]> = {};
+        const newPlaces: NamedTypedPlace[] = [];
+
+        responses.forEach((response) => {
+          newPlaces.push(response.place);
+          let parents = response.parentPlaces;
+          const place = response.place;
+
+          /*
+            We apply the following rules here to truncate the list of parent places:
```
dwnoble, line 381: Would it make sense to extract this logic to a function? I don't know if i have strong feelings about it either way, so a genuine ask here.

dwnoble (commented):
The new functionality is looking good!

PR #5440, opened 2025-08-27
Add Custom DC test env file to gitignore

hqpho:
This file is pulled from Secret Manager by `ensure_cdc_test_env_file` in run_test.sh.


beets (approved):
thanks for the update!

PR #5445, opened 2025-08-29
Update to build logs

beets:
Clean up some of our build logs to reduce noise and improve visibility of errors. Mostly to reduce scrolling 👀 📜 

* Reduces wireit logging in our ci scripts to reduce unnecessary noise.
  e.g., for `website-pull-request-npm`:  removes ~2k log lines [before](https://github.com/user-attachments/assets/8fe1ac6c-770a-4a8c-b8b7-4ce352077d29), [after](https://github.com/user-attachments/assets/1196b28a-1ab0-4b00-87cb-7c450b1e556c)

* Updates some webpack progress logging as well, to reduce verbosity, add color, etc. When building js packages locally, overrides the default progress bar, that looks like many lines of this:
  ```
  <s> [webpack.Progress] 1% setup before run
  ```
  to a single (or few) line(s) like the following:
  ```
  webpack: 92% sealing asset processing SourceMapDevToolPlugin ...
  ```
  which results in 27 lines for `./run_test.sh -b`, vs 689 before.

  In ci (or non-TTY outputs), progress logging is suppressed.


beets:
thanks all for the reviews! 

i realized that the TTY check doesn't quite work (it essentially turned off the plugin in all cases), so updated to use an env flag for the ci runs (so we should still see the output locally). just ran tests in ci and locally and things work as expected.

will merge this in, and can make updates if we run into any issues with the logging. also open to suggestions for improvements to this :)

PR #5448, opened 2025-08-30
Add short-term search indicators endpoint 

clincoln8:
Adds /api/nl/search-indicators endpoint which calls the nl_search_vars endpoint to search the embeddings indexes for statvars and topics.

See https://docs.google.com/document/d/1cACOCOzowcSndYdgDsd4S4-fi_p5l4d-GYjZGQJlsFo/edit?resourcekey=0-ySczpup9r4mrwpV5KKUIUQ&tab=t.0 for endpoint overview.

I also had an auto-save formatter on when I started this PR which caused a lot of single quotes to be changed to double quote. This makes this change look a lot larger than it actually is, apologies for the noise!


server/routes/nl/api.py
```diff
@@ -47,3 +105,212 @@ def search_vector():
...
+
+    threshold = threshold_override if threshold_override else nl_result.get(
+        "scoreThreshold", 0.0)
+
+    for query, q_result in nl_result.get("queryResults", {}).items():
+      ranked_indicators = zip(q_result.get("SV", []),
+                              q_result.get("CosineScore", []))
+      for dcid, score in ranked_indicators:
+        if score >= threshold:
+          all_dcids_to_enrich.add(dcid)
```
keyurva, line 234: Wondering if we should break based on `max_candidates_per_index` here so we are not processing more than we need to.

Also, should we have a default for this? Say 10?

server/routes/nl/api.py
```diff
@@ -47,3 +105,212 @@ def search_vector():
...
+      name_nodes = sv_data.get("arcs", {}).get("name", {}).get("nodes", [])
+      name = name_nodes[0].get("value") if name_nodes else None
+      desc_nodes = (sv_data.get("arcs", {}).get("description",
+                                                {}).get("nodes", []))
+      description = desc_nodes[0].get("value") if desc_nodes else None
+      sv_info_map[dcid] = {"name": name, "description": description}
+
+  # Step 4: Assemble the final response, preserving original ranking and using
+  # the Pydantic models for validation and serialization.
+  response_query_results: Dict[str, Dict[str, IndexResponse]] = {}
```
keyurva, line 250: nit and optional: Consider using `dict` (`list`, `set`, etc.) vs their typing counterparts here and elsewhere.

server/services/datacommons.py
```diff
@@ -320,206 +338,255 @@ def v2node_paginated(nodes, prop, max_pages=1):
...
-  return post(url, {'queries': queries})
+    url = f"{url}&skip_topics={skip_topics}"
+  return post(url, {"queries": queries},
+              api_key=api_key,
+              log_extreme_calls=False)
+
+
+def nl_search_vars_in_parallel(queries,
+                               index_types: List[str],
+                               skip_topics="") -> Dict[str, Dict]:
```
keyurva, line 438: nit: Wondering if this method can accept a `bool` for `skip_topics` and do the param value conversion here?

server/routes/nl/api.py
```diff
@@ -47,3 +105,212 @@ def search_vector():
   return dc.nl_search_vars(queries,
                            idx.split(','),
                            skip_topics=request.args.get('skip_topics', ''))
+
+
+@bp.route("/search-variables", methods=["GET"])
+def search_variables():
```
keyurva, line 111: This method has gotten quite long. Consider breaking it up into smaller helper methods if possible.

server/routes/nl/api.py
```diff
@@ -47,3 +105,212 @@ def search_vector():
...
+  """Performs variable search for given queries, with optional filtering.
```
keyurva, line 112: Love the detailed doc!

keyurva (commented):
Thanks Christie!

server/routes/nl/api.py
clincoln8: Good call on the break since we're not doing place existence checks yet. I'll need to refactor that when we add place existence.

Any reason why default would be specifically 10 and not more?
nl_search_vars has a hard coded max of 40 indicators or 60 statvars: https://github.com/datacommonsorg/website/blob/db8a3ce73fefa59c10cdee57e0f8b7ea46041767/nl_server/search.py#L30-L34

keyurva: > Any reason why default would be specifically 10 and not more?

Nope. Higher is fine as long as there is one.

server/routes/nl/api.py
```diff
@@ -47,3 +105,212 @@ def search_vector():
...
+      type=lambda v: v.lower() in ["true", "1"],
+  )
+  skip_topics_arg = "true" if skip_topics else ""
+
+  # Step 1: Get search results from the NL server in parallel.
+  #
+  nl_results_by_index = dc.nl_search_vars_in_parallel(
+      queries, indices, skip_topics=skip_topics_arg)
+
+  # Step 2: Collect all candidate DCIDs that pass the threshold for enrichment.
```
dwnoble, line 220: Would it make sense to extract each of these steps into their own function?

server/routes/nl/api.py
clincoln8: agreed, it's broken up into smaller methods now. thanks for pointing that out!

server/routes/nl/api.py
```diff
@@ -47,3 +108,300 @@ def search_vector():
...
+            Indicator(dcid=dcid,
+                      name=sv_info.get("name"),
+                      description=sv_info.get("description"),
+                      type_of=sv_info.get("type_of"),
+                      score=score,
+                      search_descriptions=sentences))
+
+      index_results.append(
+          IndexResult(index=index,
+                      default_threshold=nl_result.get("scoreThreshold"),
```
keyurva, line 244: Do we need to worry about the case where `nl_result.get("scoreThreshold")` is `None`?

server/routes/nl/api.py
```diff
@@ -47,3 +108,300 @@ def search_vector():
...
+
+    query_results.append(QueryResult(query=query, results=index_results))
+
+  final_response = SearchVariablesResponse(results=query_results,
+                                           response_metadata=response_metadata)
+  return final_response
+
+
+def _parse_request_args(
+) -> tuple[list[str], list[str], float | None, int | None, list[str], bool]:
```
keyurva, line 255: Consider using data classes for long tuples like this.

keyurva (approved):
Thanks for the updates!

server/routes/nl/api.py
clincoln8: in line 381 we set the effective threshold to 0.0 if it's missing from the NL response. Then it makes sense that the default_threshold is listed as None in the response.

server/routes/nl/api.py
```diff
@@ -18,25 +18,86 @@
...
+PARAM_INDEX = "index"
+PARAM_THRESHOLD = "threshold"
+PARAM_LIMIT_PER_INDEX = "limit_per_index"
+PARAM_INCLUDE_TYPES = "include_types"
+TYPE_TOPIC = "Topic"
+
+# Type alias for a DCID string
+Dcid = str
+# Type alias for a tuple of (DCID, score)
+IndicatorScore = tuple[str, float]
```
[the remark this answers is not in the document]
clincoln8: Not particularly other than I forget that dataclasses exist 😅 I've updated IndicatorScore to be a dataclass

server/routes/nl/api.py
```diff
@@ -18,25 +18,104 @@
...
+class IndicatorScore:
+  dcid: str
+  score: float
+
+
+# Type alias for the nested dictionary: {query: {index: [(dcid, score), ...]}}
+TruncatedResults = dict[str, dict[str, list[IndicatorScore]]]
+
+
+@dataclasses.dataclass
```
dwnoble, line 52: For consistency I'd consider using a pydantic object here. They can be treated as a drop-in replacement for dataclasses, and pydantic may help simplify the `_parse_request_args` logic

dwnoble (approved):

server/routes/nl/api.py
clincoln8: converted to pydantic models!

PR #5449, opened 2025-08-31
Explore Page Result Header Adjustments

nick-nlb:
## Description

This PR implements adjustments to the explore page header as requested in the following screenshot:

padding-requests

It involves four changes: 

* "All about" becomes "About"
* The eyebrow text above the query becomes grey
* The space between the eyebrow and the query becomes smaller.
* The padding between the query and the charts (or the topics and the charts if topics are showing) is made smaller.

## Notes

To implement the padding changes, we had to adjust some of the padding in the blocks themselves (because most of the extra whitespace in the "before" picture is padding inside the first block). We did this in a very targeted way in the CSS, to strip the padding only from the very first block, and only on this page.

Also note that the original screenshot in the request showed topics. These are now dropped completely in local and autopush, and topics will not appear again together with this new header (although you can force them to show locally by altering the flags).

### Mobile

mobile

### Desktop

full-screen

### DIFF
--- a/static/webpack.config.js
+++ b/static/webpack.config.js
@@ -232,10 +232,10 @@ const config = {
 };
 
 // Outputs webpack build progress in a single terminal line (if in a TTY environment).
-var interactiveProgressHandler = (percentage, message, ...args) => {
+const interactiveProgressHandler = (percentage, message, ...args) => {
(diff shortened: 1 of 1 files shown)


beets (approved):
thanks!

PR #5450, opened 2025-08-31
Bar Charts - Inconsistent Facets

nick-nlb:
## Description

This PR updates the facet selector so that when we are in "grouped" mode (i.e., for bar charts), we no longer display inconsistent facets.  An inconsistent facet is one that does not appear across all stat vars available in the chart.

This is row number 3. in the [Bugbash](https://docs.google.com/spreadsheets/d/1VtmR1H3VOBbEL_D0OxP9z0ZSFmD3H6k_ZldraM11CjA/edit?gid=690212228#gid=690212228), "Bar charts - Age distribution issues".

## Notes

I did not completely remove the previous functionality of keeping and allowing suggestion of inconsistent facets, but rather placed that functionality behind a flag set to false. This is for the easy recovery of the original functionality, which may be useful for testing and exploration of the data later.

### Before

Before

### After

After

## Tests

The bar chart facet selector test needed to be updated because the chart being tested now no longer has facet choices.

We are now testing another chart. Because this chart's sources do not change when we change facet, we are opening up the metadata dialog to verify the facet was changed (similar functionality to the per capita test).

## Testing

More than one consistent facet, facet selector visible:

http://localhost:8080/explore#q=Jobs+in+the+United+States

Facet options now inconsistent, facet selector invisible.

http://localhost:8080/explore#q=Age+distribution+in+the+United+States

(This age distribution example shows the flag and inconsistent facets if you set the `SHOW_INCONSISTENT_FACETS` flag to this.

### DIFF
--- a/static/css/nl_interface.scss
+++ b/static/css/nl_interface.scss
@@ -812,6 +812,10 @@ $loading-dot-color: #999;
   padding-bottom: 1.5em;
 }
 
+#dc-explore #subject-page-main-pane .col-12 {
+  padding-top: 0;
+}
+
 .examples-container {
   display: flex;
   flex-direction: column;
--- a/static/css/shared/_subject_page.scss
+++ b/static/css/shared/_subject_page.scss
@@ -42,6 +42,14 @@
   }
 }
 
+#dc-explore #subject-page-main-pane {
+  .category:first-of-type {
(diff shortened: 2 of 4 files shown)


beets (approved):
thanks for the fix and the updated test

PR #5453, opened 2025-08-31
Date range suppression by provenance

nick-nlb:
## Description

This PR implements suppression lists by provenance name for date ranges and measurement methods.

The reason for this is that certain very common provenances (Wikipedia-related) have incorrect or concerning date ranges.

This is row number 7. in the [Bugbash](https://docs.google.com/spreadsheets/d/1VtmR1H3VOBbEL_D0OxP9z0ZSFmD3H6k_ZldraM11CjA/edit?gid=690212228#gid=690212228), '"Year data issues from Wikipedia data".

## Notes 

The bugbash entry calls out the date range, but the Wikipedia measurement method also has a problem whereby there are multiple descriptions for the measurement method in the graph and the first one is incorrect (we always take the first node for measurement methods).

### Before

Facet Selector

before - facet selector

Metadata Modal

before - metadata modal

### After

Facet Selector

after - facet selector

Metadata Modal

after - metadata modal

### DIFF
--- a/server/webdriver/shared_tests/vis_timeline_test.py
+++ b/server/webdriver/shared_tests/vis_timeline_test.py
@@ -156,7 +156,7 @@ class VisTimelineTestMixin():
 
     # Check for the existence of the message
     header_message = chart_timeline.find_element(
-        By.XPATH, ".//header/p[text()='Displaying all the datasets available']")
+        By.XPATH, ".//header/p[text()='One dataset available for this chart']")
     self.assertIsNotNone(header_message)
 
   def test_manually_enter_options(self):
--- a/server/webdriver/tests/explore_test.py
+++ b/server/webdriver/tests/explore_test.py
@@ -15,6 +15,7 @@
 import re
 
 from selenium.webdriver.common.by import By
+from selenium.webdriver.support import expected_conditions as EC
 from selenium.webdriver.support.ui import WebDriverWait
 
 from server.webdriver import shared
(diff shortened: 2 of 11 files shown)


static/js/tools/shared/metadata/metadata_fetcher.ts
```diff
@@ -35,6 +35,14 @@ import {
  * This module contains helper functions for fetching and processing metadata.
  */
 
+/* 
+  The following are lists of provenances that contain known incorrect or otherwise
+  concerning date ranges or measurement methods. The respective date range or
+  measurement method will not be included in the compiled metadata.
+ */
+const dateRangeSuppressionProvenances = ["Wikidata", "WikipediaStatsData"];
+const measurementMethodSuppressionProvenances = ["WikipediaStatsData"];
```
beets, line 44: super nit: since these are constants, please update to use all caps for the constant names:

https://google.github.io/styleguide/jsguide.html#naming-constant-names

beets: please also add a todo that these are temporary (and should be fixed upstream in the data source)

static/js/tools/shared/metadata/metadata_fetcher.ts
```diff
@@ -448,32 +464,50 @@ export async function fetchMetadata(
       const key = matchedSeries?.seriesKey ?? {};
       const unit = key.unit || facetInfo.unit;
 
+      const sourceName = provenanceData?.source?.[0]?.name;
+      const provenanceName =
+        provenanceData?.isPartOf?.[0]?.name ||
+        provenanceData?.name?.[0]?.value ||
+        importName;
```
beets, line 471: (for later): rather than duplicate this logic across the file, could we move this logic upstream into flask? please add a TODO for now.

beets (approved):
thanks for the fix. approving for now, but please address the comment before merging

PR #5456, opened 2025-09-01
Facet Selector - Scroll to selected option

nick-nlb:
## Description

With this PR, when a user opens the facet selector after already having selected an option in that same facet selector previously, the facet selector will automatically appear scrolled with that item centered in view.

This is row number 5. in the [Bugbash](https://docs.google.com/spreadsheets/d/1VtmR1H3VOBbEL_D0OxP9z0ZSFmD3H6k_ZldraM11CjA/edit?gid=690212228#gid=690212228), "Would be nice that when you open the data selector dialog it goes to the place where the data is selected, vs just the top".

## Notes

I tried a version first where it opened scrolled to the top, and have it automatically scrolled down smoothly to the selected item, but this turned out to be a bit "dizzying". Now, we rather use a useLayoutEffect to actually scroll to that item before the user sees anything, and the effect is nicer.

Another note is that the diff is fairly large but a lot of that is the result of increased indentation because of the forwardRef.

## Testing

A convenient page test this is:

http://localhost:8080/explore#q=Population+of+the+world

The "Population in the World" chart has a large number of options, allowing the scrolling to be tested easily.

### DIFF
--- a/server/webdriver/shared_tests/vis_timeline_test.py
+++ b/server/webdriver/shared_tests/vis_timeline_test.py
@@ -156,7 +156,7 @@ class VisTimelineTestMixin():
 
     # Check for the existence of the message
     header_message = chart_timeline.find_element(
-        By.XPATH, ".//header/p[text()='Displaying all the datasets available']")
+        By.XPATH, ".//header/p[text()='One dataset available for this chart']")
     self.assertIsNotNone(header_message)
 
   def test_manually_enter_options(self):
--- a/server/webdriver/tests/explore_test.py
+++ b/server/webdriver/tests/explore_test.py
@@ -15,6 +15,7 @@
 import re
 
 from selenium.webdriver.common.by import By
+from selenium.webdriver.support import expected_conditions as EC
 from selenium.webdriver.support.ui import WebDriverWait
 
 from server.webdriver import shared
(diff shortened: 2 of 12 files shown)


static/js/shared/facet_selector/facet_selector_grouped_content.tsx
```diff
@@ -97,8 +107,36 @@ export function FacetSelectorGroupedContent({
     : "";
   const totalStatVars = facetList.length;
 
+  useLayoutEffect(() => {
```
beets, line 110: please add a comment here describing the intent of the scroll

static/js/shared/facet_selector/facet_selector_standard_content.tsx
```diff
@@ -48,8 +54,40 @@ export function FacetSelectorStandardContent({
   mode,
 }: FacetSelectorStandardContentProps): ReactElement {
   const theme = useTheme();
+  const itemRefs = useRef(new Map<string, HTMLElement>());
+  const [isVisible, setIsVisible] = useState(false);
+  const hasScrolledRef = useRef(false);
+
+  useLayoutEffect(() => {
```
beets, line 61: ditto - please leave a comment describing the effect

beets (approved):
thanks - just a couple of small comments to leave a note on each effect block.

PR #5459, opened 2025-09-02
Ranking Tile - Chart embed citations

nick-nlb:
## Description

This PR wires the combined facets and stat var to facet map into the ChartEmbed (download data/SVG) dialog.

This is row number 12. in the [Bugbash](https://docs.google.com/spreadsheets/d/1VtmR1H3VOBbEL_D0OxP9z0ZSFmD3H6k_ZldraM11CjA/edit?gid=690212228#gid=690212228), 'Ranking chart download does not include sources'.

## Notes

The metadata modal itself (which appears at the top of each ranking unit section) did not need this combined list, because each ranking unit has its own metadata button. However, the download dialog does, because it appears once at the bottom of the ranking chart. This combined list matches the data that the dialog is providing.

### Before

ranking-download-dialog-no-citations

### After

ranking-download-dialog-citations

### DIFF
--- a/static/js/components/tiles/ranking_tile.tsx
+++ b/static/js/components/tiles/ranking_tile.tsx
@@ -137,7 +137,7 @@ export function RankingTile(props: RankingTilePropType): ReactElement {
   ]);
 
   /**
-    This hook mergers all the facets across ranking units, providing a single
+    This hook merges all the facets across ranking units, providing a single
     list that is sent into the ChartEmbed (data/svg download) component. This
     allows the dialog to build a citation string that matches the data it provides.
    */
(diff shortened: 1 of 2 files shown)


static/js/components/tiles/ranking_tile.tsx
```diff
@@ -136,6 +136,36 @@ export function RankingTile(props: RankingTilePropType): ReactElement {
     variables,
   ]);
 
+  /**
+    This hook mergers all the facets across ranking units, providing a single
```
beets, line 140:
```suggestion
    This hook merges all the facets across ranking units, providing a single
```

beets (approved):
thanks for the fix. as a followup, could we add the modal to the screenshot test suite?

PR #5471, opened 2025-09-05
Allow targeted lint fixing in run_test.sh

clincoln8:
This allows us to run either 
```
./run_test.sh -f py
```
or 
```
./run_test.sh -f npm
```
to fix lint related to local changes.
We can still run
```
./run_test.sh -f
```
to fix both client code and server code in one shot.

### DIFF
--- a/run_test.sh
+++ b/run_test.sh
@@ -148,22 +148,31 @@ function run_npm_lint_test {
 
 # Fixes lint
 function run_lint_fix {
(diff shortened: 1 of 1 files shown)


beets:
thanks for the fix -- much needed! 

sorry, i was testing adding an auto-assignment review group, and i ended up getting assigned

beets (approved):
thanks for the udpates!

PR #5490, opened 2025-09-10
[Vertex AI Search] Major Refactoring of the script to Generate NL Metadata, Add BigQuery Diff mode and other things for productionization

gmechali:
Major refactoring to the script to generate NL Metadata for Vertex AI:
- Adds logging throughout the script to get a better understanding as execution is in progress.
- Adds a BigQueryDiffs run mode as discussed previously to diff the content of the GCS Bucket with the latest content of BigQuery to ONLY process NEW stat vars.
- Adds a Compaction run mode, as the bigquery diff run mode keeps adding new files on each run, a compaction script will run monthly to merge all the files to avoid bloat.
- Improves failure retry mode to be able to run on the periodic/ folder easily.
- Refactors helper functiosn and gemini-related functions into helper files.
- Updates Readme and comments.
- Improved tests

Tested: Ran this script many times in bigquery diff run mode with a max stat vars to generate many files in a testing directory, then ran compaction script. Also tested in Datcom-nl project, see [Cloud run Jobs](https://pantheon.corp.google.com/run/jobs/details/us-central1/stat-var-metadata-generator/executions?e=13803378&inv=1&invt=Ab1zPA&mods=-monitoring_api_staging&project=datcom-nl) and [output file](https://pantheon.corp.google.com/storage/browser/metadata_for_vertexai_search/gmechali_csv_testing/periodic;tab=objects?e=13803378&inv=1&invt=Ab1m6g&mods=-monitoring_api_staging&project=datcom-ci&prefix=&forceOnObjectsSortingFiltering=false)

To View in action, see the GCP [Workflow](https://pantheon.corp.google.com/workflows/workflow/us-central1/testing-workflow/executions?e=13803378&inv=1&invt=Ab1zPA&mods=-monitoring_api_staging&project=datcom-nl) which reprocesses all the stat vars in BigQuery Diff mode, then runs the compaction mode.


tools/nl/nl_metadata/README.md
```diff
@@ -21,7 +21,7 @@ Before running the scripts, you need to set up your API keys.
 2.  Edit the newly created `.env` file and add your API keys:
     *   `DC_API_KEY`: Your Data Commons API key.
     *   `GEMINI_API_KEY`: A single Gemini API key, used for local runs.
-    *   `GEMINI_API_KEYS`: A comma-separated list of Gemini API keys, used for parallelized Cloud Run jobs (`run_add_metadata.py`). Separate API keys are used per Cloud Run job to prevent 429 (resource exhaustion) errors from Gemini. 
+    *   `GEMINI_API_KEYS`: A comma-separated list of Gemini API keys, used for parallelized Cloud Run jobs (`trigger_nl_metadata_job.py`). Separate API keys are used per Cloud Run job to prevent 429 (resource exhaustion) errors from Gemini.
```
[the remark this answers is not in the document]
gmechali: yeah we did. It was convenient while testing to be able to keep the keys configured and just run the command with different flags based on if we wanted to run all 12 partitions, or a small scale test.

To be honest, I dont know if this script will be long lived in the end. The GCP Workflow i configured doesn't use it so this is just for manual triggering. I expect that very shortly we will decide this isnt needed. Any time we want to run it, we can manually trigger the workflow.

However, just in case I do want to keep it for now :) More to come on this!

tools/nl/nl_metadata/gemini_service.py
```diff
@@ -0,0 +1,107 @@
...
+from google.genai import types
+import google.genai as genai
+from schemas import StatVarMetadata
+from utils import split_into_batches
+
+
+async def generate_alt_sentences(
+    gemini_client: genai.Client, gemini_config: types.GenerateContentConfig,
+    gemini_prompt: str, sv_metadata: list[dict[str, str | list[str]]],
+    delay: int) -> list[dict[str, str | list[str]]]:
```
dhotchkiss, line 13: I think ideally we'd use durations in the future (flag support example: https://source.corp.google.com/piper///depot/google3/pyglib/flags/contrib/duration_flag.py)

gmechali: Added as a Todo.

gmechali:
> I don't particularly love how nested the asyncio calls are in the code but appreciate the overall refactoring. Ty, let's not have perfect being the enemy of good. Let's merge and if we keep around this function at the of the year we can reconsider if a second refactor is appropriate.

Yeah I understand, definitely trying to strike a balance on parallel vs partition. I think it's done right currently, but the nestedness of the asyncio calls definitely makes it a bit obfsucated to udnerstand where the parallelization comes into play.
It's essenitally just parallelizing the Gemini calls since those are what takes a very long time. 
Sounds good on reconsidering changs later on if we find this to be hard to maintain.

PR #5491, opened 2025-09-11
[StatVar Autocomplete] Quality improvements + Full Stat Var search

gmechali:
Made a few changes to increase quality of results:
- On NGrams, only pass in the portion of the query after a LOCATION as identified by the Language Client. This does add some small latency to it (about 250ms from very rough approximation), but it significantly improves the results, so probably worth it.
- For all suggestions, append "on earth" if there are no places in the query to make it an executable query.
- Redirect the user to the /explore?p=Earth&sv=StatVar for exact StatVar selections.

https://screencast.googleplex.com/cast/NjAzNTMzNjE2ODczNDcyMHwwZmM5YjQwOS03ZA


clincoln8 (approved):
Thanks for the update!

Were you able to fix the spacing seen in https://screenshot.googleplex.com/4GP2ZeR6CtV5Zhg or will that be a follow up?

PR #5494, opened 2025-09-15
Tighten flakey webdriver tests

juliawu:
This PR updates our webdriver tests to tighten some flakey tests.

Changes made:
* Fixed a Custom DC test that was trying to call a function that no longer exists.
* Adding more waits to prevent a race condition of a test looking for elements before they've loaded.
* Increased the timeout on waits for scatter charts to load. The scatter charts can be particularly slow to load.
* Swapped to using the shared `charts_rendered` wait condition across all visualization tool tests for consistency.
* Updated the helper functions for interacting with the google maps autocomplete place search to be less brittle.

Testing strategy:
Ran `./run_test.sh -f` locally multiple times. All tests pass each time.

### DIFF
--- a/static/js/components/elements/tabs/tab_set.tsx
+++ b/static/js/components/elements/tabs/tab_set.tsx
@@ -52,10 +52,13 @@ export const TabSet = ({
   const scrollContainerRef = useRef<HTMLDivElement>(null);
   const tabSetRef = useRef<HTMLDivElement>(null);
 
-  const alignmentMap: Record<TabSetAlignment, string> = {
-    left: "flex-start",
-    center: "center",
-    right: "flex-end",
+  const marginStyles: Record<TabSetAlignment, string> = {
+    left: "",
+    center: `
+      margin-left: auto;
+      margin-right: auto;
+    `,
+    right: "margin-left: auto;",
   };
 
   const [indicatorStyle, setIndicatorStyle] = useState<{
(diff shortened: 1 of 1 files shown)


server/webdriver/shared_tests/scatter_test.py
```diff
@@ -30,6 +29,10 @@
     'Count_Person_AsianAlone&svpy=0-14-1&svdy=Count_Person&svny=Asian_Alone&pcy=1'\
     '&epd=geoId/06&epn=California&ept=County'
 
+# Scatter plots can take extra long to load
+# This is a custom, longer timeout to use for charts we know are slow
+LONG_TIMEOUT = 120  # seconds
```
beets, line 34: let's move this to the shared library so all the timeout constants are together

server/webdriver/shared_tests/scatter_test.py
```diff
@@ -57,6 +60,9 @@ def test_charts_from_url(self):
 
     # Wait until the chart has loaded.
     shared.wait_for_loading(self.driver)
+    WebDriverWait(self.driver, LONG_TIMEOUT).until(shared.charts_rendered)
```
beets, line 63: how about adding a variant of wait_elem that we use here and elsewhere for things we know are slower? something like wait_elem_slow

beets (commented):
thanks for deflaking these tests!

server/webdriver/shared_tests/scatter_test.py
juliawu: Good idea. Done!

server/webdriver/shared_tests/scatter_test.py
juliawu: wait_elem explicitly uses `presence_of_element_located` as an expected condition, so instead of a wait_elem, I added a new `wait_for_charts_to_render`. I noticed the way we currently use shared.charts_rendered has an inefficient double wait, so I've also updated the function to more clearly be an explicit wait condition called by wait_for_charts_to_render.

I also didn't add a wait_elem_slow because I think we can just use the timeout parameter in wait_elem

beets (approved):
thanks for the fixes and refactor! lgtm

PR #5496, opened 2025-09-15
[search-indicators] Read typeOf from dcid instead of name

clincoln8:
Context: https://github.com/datacommonsorg/agent-toolkit/pull/55/files/cd3e5786aa04f47baef47c265ba079c1df0f6f8e#r2349599413

* Custom Nodes might not have `name` set: 
image

* Base Nodes have both `name` and `dcid`: 
image

So we should read typeOf from dcid and not name.

Tested locally and typeOf is still set: 
image

### DIFF
--- a/server/routes/nl/api.py
+++ b/server/routes/nl/api.py
@@ -16,8 +16,12 @@
 import json
 
 import flask
-from flask import Blueprint, Response, request
-from pydantic import BaseModel, ConfigDict, Field
+from flask import Blueprint
+from flask import request
+from flask import Response
+from pydantic import BaseModel
+from pydantic import ConfigDict
+from pydantic import Field
 
 from server.services import datacommons as dc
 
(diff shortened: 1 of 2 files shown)


jm-rivera:
All of our variables have the name set, so I think the issue is that the right info does come with the name - rather than the name being missing.
We've actually run into this issue in a few places where custom nodes are missing information that exists depending on what call is made to the API.

clincoln8:
> All of our variables have the name set, so I think the issue is that the right info does come with the name - rather than the name being missing. We've actually run into this issue in a few places where custom nodes are missing information that exists depending on what call is made to the API.

@jm-rivera The way I worded the PR description might have been confusing. It's the name of the node that the typeOf property points to that's missing the name.
For example Custom_Stat_Var.**typeOf**.node only has (dcid = "StatisticalVariable") where as Base_Stat_Var.**typeOf**.node has both (dcid="StatisticalVariable", name="StatisticalVariable"). Prior to this PR, I was using the typeOf.node.name whereas I should really be using typeOf.node.dcid anyways. 

Does that seem accurate to you?

PR #5498, opened 2025-09-16
[Vertex AI Search] Properly log the gemini service errors + Add Image Build & Deploy

gmechali:
This will ensure that GCP Error Reporting will fire on those logs so we can be alerted if we start receiving too many errors.

This also creates the CloudBuild file to push the image to Artifact registry and deploy the Cloud Run Job to use the new latest image after the new one is available.
I have also setup the Cloud Build Trigger for periodic rebuild of all binaries to run this cloud build file ensuring we keep our tools StatVar Metadata Generator in sync every day without requiring manual intervention.


tools/nl/nl_metadata/cloudbuild.push_image.yaml
```diff
@@ -0,0 +1,50 @@
...
+  - name: 'gcr.io/google.com/cloudsdktool/cloud-sdk'
+    id: DEPLOY
+    entrypoint: gcloud
+    args:
+      - run
+      - jobs
+      - update
+      - stat-var-metadata-generator
+      - '--image'
+      - 'us-west1-docker.pkg.dev/datcom-ci/cloud-run-source-deploy/nl-metadata-image:latest'
```
dhotchkiss, line 44: nit: I don't love seeing single arguments spread over multiple lines; it's not obvious which are together and which could be sorted. 

consider inlining the key/val pairs. e.g
```
args:
  - run
  - jobs
  - ...
  - '--region uscentral1'
  - '--project datcom-nl'
```

PR #5501, opened 2025-09-17
Highlight Result Tile Padding Fix

nick-nlb:
## Description

This PR fixes the bottom padding in the highlight result section.

Previously, the subtopic section inside the highlight was rendering a bottom border (correctly) but without padding.

### Before

Screenshot 2025-09-17 at 3 49 56 PM

### After

Screenshot 2025-09-17 at 12 24 13 PM

## Testing

This scenario can be replicated at the following link:

http://localhost:8080/explore?p=country/IND&sv=Amount_EconomicActivity_GrossDomesticProduction_Nominal&unit=USDollar&imp=WorldDevelopmentIndicators&obsPer=P1Y&chartType=TIMELINE_WITH_HIGHLIGHT&origin=aim

### DIFF
--- a/static/css/shared/_subject_page.scss
+++ b/static/css/shared/_subject_page.scss
@@ -115,6 +115,7 @@ article.category {
 
 .highlight-result-title .subtopic:last-of-type {
   border-bottom: 1px solid rgba(0, 0, 0, 0.12);
+  padding-bottom: 1.5em;
 }
 
 .subtopic.notitle:not(:first-of-type) {


beets:
nick - could you add this page to the screenshot suite so we catch future regressions?
https://github.com/datacommonsorg/website/blob/b89b181adbe212e0ab9e0158cda8311cda6c6cf4/server/webdriver/screenshot/local/page.json#L183

PR #5503, opened 2025-09-17
Highlight Result Tile Snapshot

nick-nlb:
## Description

This PR adds a webdriver snapshot for an explore page that demonstrates a highlight result.

It is a follow-up to: #5501


beets (approved):
thanks for the follow up!

PR #5504, opened 2025-09-17
update version number

shixiao-coder:
Updating the version number from 1.5 to 2.5 for Gemini pro  models


server/integration_tests/test_data/detection_api_multivar/numberofpoorhispanicwomenwithphd/debug_info.json
beets, line 1: the changes in this file are a loss -- it looks like we've stripped too much from the query, so the matches are slightly less relevant.

server/integration_tests/test_data/detection_api_multivar/numberofpoorhispanicwomenwithphd/debug_info.json
```diff
@@ -167,46 +162,45 @@
     },
     "Query": "number of poor hispanic women with phd",
     "SV": [
-      "dc/06f0jf8xvzw4f",
```
beets, line 170: this is a really relevant sv: https://datacommons.org/browser/dc/06f0jf8xvzw4f

clincoln8: Looks like dc/06f0jf8xvzw4f just got demoted to second highest, not deleted. It's at line 166 now.

server/integration_tests/test_data/detection_api_multivar/numberofpoorhispanicwomenwithphd/debug_info.json
clincoln8: This doesn't actually seem like much of a loss to me in terms of before and after the model update. They are technically both a loss considering this query is being interpreted as a MultiSV query when it really shouldn't be. I think this is because we don't actually have a statvar that captures "income of women with phd"

In fact, this is actually conforming more to the prompt than before: https://screenshot.googleplex.com/9JHyqQyt4DPbbc2

server/routes/experiments/biomed_nl/api.py
```diff
@@ -27,7 +27,7 @@
 from server.routes.experiments.biomed_nl.traversal import PathFinder
 import server.routes.experiments.biomed_nl.utils as utils
 
-GEMINI_PRO = 'gemini-1.5-pro'
+GEMINI_PRO = 'gemini-2.5-pro'
```
clincoln8, line 30: fyi, this will require a push to dev to make this change "live".

server/integration_tests/test_data/detection_api_context/stateswithhighestphds/chart_config.json
```diff
@@ -1,6 +1,16 @@
 {
   "childEntityType": "State",
   "classifications": [
+    {
```
clincoln8, line 4: I'm not really sure what this section of this output means. @beets Any pointers?

clincoln8 (approved):
Thanks Xiao, congrats on the speedy first PR!

PR #5514, opened 2025-09-22
Webdriver Test Tightening

nick-nlb:
## Description

This PR contains measures to tighten four tests, all of which exhibited intermittent failures due to race conditions.

* `test_bar_select_different_facet` and `test_per_capita_metadata`: These tests occasionally failed via the same mechanism. The presence of the element was waited for, but this sometimes resolved before the content tested for afterwards was ready. We resolved this by waiting for the relevant element to be visible and with that text.
*  `test_no_facet_choices_available` and `test_select_different_facet`: These tests both assumed the existence of certain elements after the chart finished loading (which was usually true). We now explicitly wait for them.

## Notes

These were based on reports by @juliawu (thank you!). I could replicate the first two issues. However, I was unable to replicate the last two. Nevertheless, I did identify those possible routes to a race condition in those latter tests and resolved them.

### DIFF
--- a/server/webdriver/shared_tests/vis_timeline_test.py
+++ b/server/webdriver/shared_tests/vis_timeline_test.py
@@ -20,7 +20,8 @@ from selenium.webdriver.support import expected_conditions as EC
 from selenium.webdriver.support.ui import WebDriverWait
 
 from server.webdriver.base_utils import find_elem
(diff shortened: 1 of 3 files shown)


dwnoble (approved):
Thank you!

PR #5520, opened 2025-09-24
[Lint Error Fix] Let Gemini CLI Fix All Lint Errors - down to 54

gmechali:


static/js/apps/visualization/app_context.tsx
```diff
@@ -187,13 +187,15 @@ export function AppContextProvider(
...
+        setChildPlaceTypes((prevChildPlaceTypes) => {
+          if (_.isEqual(newChildPlaceTypes, prevChildPlaceTypes)) {
+            return prevChildPlaceTypes;
+          }
+          return newChildPlaceTypes;
+        });
       }
     );
-  }, [places, visTypeConfig]);
+  }, [places, visTypeConfig, isContextLoading]);
```
gmechali, line 198: some of these changes could potentially lead to timing changes. So perhaps it's safer to revert, but overall these all look coorect to me.

static/js/apps/topic_page/page_selector.tsx
```diff
@@ -120,24 +112,6 @@ export function PageSelector(props: PageSelectorPropType): JSX.Element {
   );
 }
 
-function getPlaceOptions(
```
clincoln8, line 123: just double checking, we don't need this method?

static/js/components/nl_search_bar/auto_complete_input.tsx
```diff
@@ -131,18 +131,30 @@ export function AutoCompleteInput(
...
         setSampleQuestionText,
         setDynamicPlaceholdersEnabled,
         setDynamicPlaceholdersDone
       );
     }
-  }, []);
+  }, [
+    inputText,
+    props.enableDynamicPlaceholders,
+    setSampleQuestionText,
```
clincoln8, line 154: I didn't know that the state setter methods would/could count as dependencies for useEffect, interesting!

clincoln8 (approved):
Thanks for the cleanup!!

static/js/apps/topic_page/page_selector.tsx
dwnoble: This looks correct to me too.

static/js/apps/visualization/app_context.tsx
dwnoble: I think this change and the changes below are fine, but can you do some manual tests in the visualization tools around changing the place types?

gmechali:
Thanks! did some more manual testing on vis tools, it looks good! I ll check autopush heavily after it s in as well!

PR #5521, opened 2025-09-24
[Vertex AI Search] Make GEmini only generate sentences and merge it into the JSON with logic.

gmechali:
We previously gave Gemini the entire JSON object and asked it to append the generated Sentences. THis opened up the possibility of Gemini hallucinations. 

With these changes, I'm now giving Gemini the entire JSON object, and telling it to return just a json with DCID + Generated Sentences, and then mathcing the returned objects back with the previous JSONs generated from BigQuery.
This prevents Gemini from hallucinating and modifying core values within the JSON object.

Tested by running the script and storing the output in the staging directory. Verified the GCS files look good!


dwnoble (approved):
Thanks Gabe!

PR #5530, opened 2025-09-25
Facet Terminology

nick-nlb:
## Description

This PR changes the terminology “dataset” to “facet” across the explore page and visualization tools. 

The word “facet” (what was dataset) is visible to the user in the facet selector button.

## Notes

The word “dataset” itself appears throughout the site (outside of the explore and the tools) but in these cases it refers to “datasets” more generally (or conceptually) and do not refer to facets in the sense that we are using the word here. These usages of the term are unchanged.

## Screenshots

Screenshot 2025-09-25 at 2 17 02 PM

Screenshot 2025-09-25 at 2 25 46 PM

localhost_8080_tools_visualization_(Screen - Full HD)


miss-o-soup:
Thank you so much for the updates on this Nick!

I noticed that the actual text within the dialog says "Select the data source that you would like... "
I think we should update that too to say "Select the facet that you would like... "

In other things, since we have different dialogs for when one facet is selected vs multiple ones. What about adjusting the language for the plurality of things. So when the user selects one face, the TITLE would read:
- Select a facet
And for multiple ones it would say what we currently have everywhere:
- Select facets

PR #5553, opened 2025-09-29
Removed legacy place page summary configuration

dwnoble:
Removed legacy place page summary configuration.

* Removed tools/summaries/* for generating and updating place page summaries & uploading them to kubernetes
* Removed place page summary kubernetes configuration mappings
* Removed a few unused references to reading the place page summary shards from the filesystem
* Added new script `scripts/delete_place_summary_config.sh` for deleting place page kubernetes configuration. After the next deployment, we should run this script to clean up the old configurations (cc @n-h-diaz )


gmechali:
The issue seems to be with this file: https://github.com/datacommonsorg/website/blob/d899e629b95bcc081990d77c8dcbb5b92f5a8eb6/deploy/helm_charts/dc_website/templates/config_files_as_cm.yaml#L11
Which had redefined how we pass in the place summaries in the cloud deploy pipeline. And now that you removed the place summary configs, this is being passed in erroneously. I think we just need to remove all the place summary configs from here as well.

PR #5554, opened 2025-09-30
[StatVar Autocomplete ] Only suggest queries for places we have data on

gmechali:
For Stat Var Autocomplete, when there is no place identified in the request, until now we defaulted to just show "on Earth". However, most of our data is not at the Earth level, which caused frequent queries to fail on StatVar Autocomplete redirection.

With this change, we don't delay the autocompletion, as Stat vars become available in the same time frame, but we later add a randomly selected place for which we know we have data.

See screencast:
https://screencast.googleplex.com/cast/NTI2MTk1MDk3MDc1NzEyMHw3Y2I0NWIzNS03ZQ

Note that if the user selects an autocomplete suggestions where it's autocompleting the entire query, we redirect it to the explore page with URL parameter injection using the SV and P params, like 
http://localhost:8080/explore#p=country/USA&sv=Annual_Emissions_CarbonDioxide_FossilFuelOperations&q=Annual%20Amount%20of%20Emissions%3A%20Fossil%20Fuel%20Operations%2C%20Carbon%20Dioxide%20in%20United%20States

However, when the autocomplete suggestion only replaces a partial query (ex: "Tell me about the life expectancy" --> The life expectancy SV will only match the tail of the query), we execute the query via the NL endpoint.

For selecting places, we only use the TopPlaces of type Country, State, AA1 or EurostatNUTS1


static/js/components/nl_search_bar/auto_complete_input.tsx
```diff
@@ -246,10 +272,20 @@ export function AutoCompleteInput(
...
+          response.data.predictions || []
         );
+        setResults(newResults);
         setBaseInputLastQuery(query);
+        const statVarDcids = newResults
+          .filter((result) => result.matchType === STAT_VAR_SEARCH)
+          .map((result) => result.dcid);
+        const dcidsToFetch = statVarDcids.filter((dcid) => !statVarInfo[dcid]);
+        if (dcidsToFetch.length > 0) {
+          getStatVarInfo(dcidsToFetch).then((resp) => {
```
dwnoble, line 285: Nit: consider flattening this call (& the surrounding axios calls) to use `await` for less nesting

dwnoble, line 285: nit: add error handling

dwnoble (commented):
Thanks Gabe! A few questions:

(1) is there a way to get nearby or more relevant place selections?  Example: https://screenshot.googleplex.com/5fv3oSfFpYFQJSM
(2) if i keep typing out a variable name, for example Unemployment rate of a Population in`, it can lose the variable and start picking up a new one: https://screenshot.googleplex.com/3oBWeYqrJT8TnyC - what do you think? (here's how it looks in prod w/out variable detection https://screenshot.googleplex.com/4Zog3mT2bc399XU)
(3) Now that autocomplete is starting to get more complex, can we add some search evals for testing to avoid regressions?
(4) With testing and evaluations in mind, would it make sense to move this autocomplete logic (and potentially _all_) logic to a backend endpoint? I wonder if it would be easier to test if we just sent a single endpoint a query and got back a list of autocomplete suggestions. Alternative, if it does make sense to keep this logic split between the frontend & backend, what do you think of migrating all of the autocomplete logic to a typescript service class that. likewise just takes an input string and returns the autocomplete candidates?

static/js/components/nl_search_bar/auto_complete_input.tsx
gmechali: done.

dwnoble:
> > Thanks Gabe! A few questions:
> > (1) is there a way to get nearby or more relevant place selections? Example: https://screenshot.googleplex.com/5fv3oSfFpYFQJSM
> 
> So Im fetching the Top Places from the StatVar Info API call. we return a placeSummary that lists out a few places per place type. Right now im doing Country + State + AA1 + EurostatsNuts1. The reason I chose these is they give pretty good coverage of our data. If I choose just country, then we won't suggest a place that we have data on for State-only statvars (which do exist). Long term, we can also try to be more picky about what goes in the Top Places?
> 
> > (2) if i keep typing out a variable name, for example Unemployment rate of a Population in`, it can lose the variable and start picking up a new one: https://screenshot.googleplex.com/3oBWeYqrJT8TnyC - what do you think? (here's how it looks in prod w/out variable detection https://screenshot.googleplex.com/4Zog3mT2bc399XU)
> 
> Right, it's a very hard balance to strike on the weights. To some extent there's logic in giving a higher weight to the tail matches, while in this case it shows it's not working super well, but in other cases it shows imrpoved results.
> 
> What's actually going on is we give all the results a score based on matching and the number of words is a factor in this score. But then in order to counteract situations some situations, I made stop words count negatively in the score. For example, if the query is Comparing life expectancy to the median age, my objective is that results for "life expectancy to the media age", which may have LifeExpectancyPerson as a result, would be deranked because of the stop words in between. However it has the drawback that "Unemployment Rate of a population" gets also a bit deranked. I still see it in the first page of results though.
> 
> (3) Now that autocomplete is starting to get more complex, can we add some search evals for testing to avoid regressions?
> 
> Yes, I ll work on that a bit later, if we launch it :)
> 
> (4) With testing and evaluations in mind, would it make sense to move this autocomplete logic (and potentially _all_) logic to a backend endpoint? I wonder if it would be easier to test if we just sent a single endpoint a query and got back a list of autocomplete suggestions. Alternative, if it does make sense to keep this logic split between the frontend & backend, what do you think of migrating all of the autocomplete logic to a typescript service class that. likewise just takes an input string and returns the autocomplete candidates?
> 
> So all of the autocompletion logic in the autocomplete.py API endpoint. The FE logic is just dealing with the complexity of rendering it, and specifically replacing the correct sub part. And in this PR we're adding the place DCID for a place that SV has data on. It's true we could move this part to the API as well. The nice thing about it in the FE is we can actually render everything and just substitute the place once it comes, so it doens't slow down autocomplete. Happy to revisit if you'd like. Maybe it can be a follow up?


Sounds good, Gabe.

For (3) and (4): I'm on board with shipping this as an experiment and worrying about evals later.

If the metrics look good, we should look into a refactor. If performance is good enough, i like the separation of moving itinto a backend endpoint to manage the complexity. If it makes sense to keep the place existence check on the frontend, that can work too if we build some thoughtful interfaces around it. We can tackle that and document the workflow once we see this is working.

PR #5558, opened 2025-09-30
Update gemini flashlite

shixiao-coder:
Updating the Gemini version from 1.5 Pro to 2.5 Flash Lite when disable feature flag of enable_gemini_2_5_flash, testing NL Eval with Flash Lite in dev


server/lib/nl/detection/llm_api.py
```diff
@@ -190,3 +195,16 @@ def _extract_answer(resp: str) -> str:
         return '{"UNSAFE": true}'
 
   return '\n'.join(ans)
+
+
+def detect_model_name() -> str:
+  if is_feature_enabled(ENABLE_GEMINI_2_5_FLASH_FLAG,
+                        request=request) and is_feature_enabled(
+                            ENABLE_GEMINI_2_5_FLASH_LITE_FLAG, request=request):
+    return 'error: both Gemini 2.5 Flash and Flash Lite feature flags are enabled'
```
gmechali, line 204: We definitely don't want to risk this - If we enable both we should just pick one.

if ENABLE_GEMINI_2_5_FLASH_LITE_FLAG:
  --> return _GEMINI_2_5_FLASH_LITE
ELSE IF ENABLE_GEMINI_2_5_FLASH_FLAG:
  --> return _GEMINI_2_5_FLASH
else:
  log an error, return 1.5 PRO

server/routes/dev_datagemma/api.py
```diff
@@ -85,7 +79,11 @@ def datagemma_query():
     return 'error: must provide a query field', 400
   if not mode or mode not in [_RIG_MODE, _RAG_MODE]:
     return f'error: must provide a mode field with values {_RIG_MODE} or {_RAG_MODE}', 400
-  dg_result = _get_datagemma_result(query, mode)
+  model_name = detect_model_name()
+  if 'error' in model_name:
```
gmechali, line 83: The error should be logged within detect_model_name().

And we should always return a model name, so you dont have to keep checking it downstream.

server/lib/nl/detection/llm_api.py
shixiao-coder: done

server/lib/nl/detection/llm_api.py
```diff
@@ -190,3 +192,19 @@ def _extract_answer(resp: str) -> str:
...
+
+
+def detect_model_name() -> str:
+  if is_feature_enabled(ENABLE_GEMINI_2_5_FLASH_FLAG,
+                        request=request) and is_feature_enabled(
+                            ENABLE_GEMINI_2_5_FLASH_LITE_FLAG, request=request):
+    logging.error(
+        'error: Both Gemini 2.5 Flash and Flash Lite feature flags are enabled, please enable only one. Force to use 1.5 Pro under this condition'
+    )
+    return _GEMINI_1_5_PRO
```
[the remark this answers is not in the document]
shixiao-coder: same as above

PR #5560, opened 2025-09-30
Highlight tile - API dialog

nick-nlb:
## Description

This PR implements the API button for highlight charts.

The highlight charts are different than other tiles in that they don't have a standard chart footer.  Because of this, the API button will appear in the tile sources component (as per the first screenshot).

The functionality of the PR can be divided into the following
- The addition of the calculation of the observation spec for the highlight chart. This is similar to the line tile's, but it takes a "LATEST" for the date.
- The addition of a new variant of the API button that does not show the icon.
- The updating of the tile sources component to accept an optional observation spec (passed only by the highlight tile). If this exists, the API button will be displayed.
- Nothing changes in the API dialog itself.

### Original Request

APIonTiles

### Highlight tile with button

Screenshot 2025-09-30 at 9 35 18 AM

### Dialog when button is clicked

Screenshot 2025-09-30 at 9 35 33 AM


static/js/tools/shared/metadata/tile_sources.tsx
```diff
@@ -133,6 +145,19 @@ export function TileSources(props: {
               </span>
             </>
           )}
+          {getObservationSpecs && (
+            <>
+              <span {...{ part: "source-separator" }}> • </span>
```
dwnoble, line 150: Thanks for adding the "part" attributes!

PR #5574, opened 2025-10-02
Reduce website staging V3 mirror fraction

hqpho:
This value got bumped when we switched to using Cloud Deploy, since we went from using Helm charts in website repo to using the ones in mixer repo with overrides defined in skaffold.yaml.

- Before Cloud Deploy: https://github.com/datacommonsorg/website/blob/afea0ba1dc70d1680146a4cbafd04bfc6f26a062/deploy/helm_charts/envs/staging.yaml#L48
- After Cloud Deploy: https://github.com/datacommonsorg/website/blob/afea0ba1dc70d1680146a4cbafd04bfc6f26a062/skaffold.yaml#L321

### DIFF
--- a/skaffold.yaml
+++ b/skaffold.yaml
@@ -25,7 +25,6 @@ kind: Config
 metadata:
   name: website
 
-
 build:
   artifacts:
     # Artifacts for local development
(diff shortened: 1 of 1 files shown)


gmechali (approved):
ah thanks!

PR #5579, opened 2025-10-03
API Dialog - Python and Syntax Highlighting

nick-nlb:
## Description

This PR implements the following features in the API dialog:
* Syntax highlighting
* Python endpoints

## Libraries

To effect the highlighting itself, we used Prism (prism.js), as it is lightweight, extensible and very well used.

## Notes

To allow for multiple languages, we introduced a "language spec", object that describe a given language and indicate the conversion function that will be used to convert an observation spec into the API call in that language.

Some languages and protocols such as Python are well suited to single scripts that can contain multiple API calls. Others, such as cURL, need to have each call separated. Whether a particular language is "single script" or "multiple endpoint" is indicated by that languages spec. The language will have an appropriate generator function based on what type it is.

Instructions are provided in the comments of the API dialog component that describe how a new language can be added.

### Scrolling

We had discussed the idea of having the scrolling within the elements that contain the code (to keep the height of the dialog stable). However, this results in double-scrolling when the dialog is small (and is also inconsistent because of the fact that some languages will contain more than one code section.

Instead, we allowed the dialog to handle the scrolling as per normal, and put in a fade transition to help smooth the effect.

## Screenshots

Screenshot 2025-10-04 at 11 44 48 AM

Screenshot 2025-10-02 at 8 17 13 PM

## Tests

The wedriver test for the API dialog has been updated for changes made by this PR, as well as augmented to test language switching via the select box.


miss-o-soup:
This looks really good Nick, thank you!
I love the clarity of how it looks. Could we make the Python option as the default one? The one that is listed first?
Per research this is the one that would more sense for data analysts, thank you!

miss-o-soup:
Thank you for the follow up, questions and suggestions.

Since we have the dropdown menu with options we could add the 3rd one as
well.
I was curious to learn more about the Python options, from what I read my
understanding is that the Python (Data Commons Python API) is desirable
because Automatically returns data in convenient Python objects and this
might help best data analysts and developers focused on data insights and
rapid development. Is that understanding correct?

Assuming the understanding of the above is true, it sounds like we might
want to list "Python (Data Commons Python API)" first in the list.
Looking forward to hearing your thoughts,

Adriana

PR #5584, opened 2025-10-03
[DRAFT] Add flag to disable google maps in flask and terraform

juliawu:


deploy/helm_charts/dc_website/templates/deployment.yaml
```diff
@@ -138,6 +138,8 @@ spec:
               value: {{ .Values.website.evalTool.enabled | quote }}
             - name: ENABLE_DATAGEMMA
               value: {{ .Values.website.dataGemma.enabled | quote }}
+            - name: ENABLE_GOOGLE_MAPS
```
dwnoble, line 141: I dont think we'll need to actually update the helm charts - those are just for our main DC environments, and we can default to enabling maps api keys

deploy/helm_charts/dc_website/values.yaml
```diff
@@ -44,6 +44,9 @@ website:
   dataGemma:
     enabled: false
 
+  googleMaps:
```
dwnoble, line 47: ditto here for helm

deploy/terraform-datacommons-website/modules/helm/main.tf
```diff
@@ -100,6 +100,11 @@ resource "helm_release" "datcom_website" {
     value = var.enable_ingress
   }
 
+  set {
```
dwnoble, line 103: This terraform module isn't actually used anymore 😬 - can you take a look at `deploy/terraform-custom-datacommons` ?

And make sure you go through the setup on https://docs.datacommons.org/custom_dc/deploy_cloud.html to test your TF changes. At the end of this, you should see the environment variable getting set here in cloud run: https://screenshot.googleplex.com/BLxQgH3HA6JjK2Q

static/js/components/google_map.tsx
```diff
@@ -207,7 +207,7 @@ export class GoogleMap extends React.Component<
   }
 
   render(): JSX.Element {
-    if (!this.state.shouldShowMap) {
+    if (!globalThis.enableGoogleMaps || !this.state.shouldShowMap) {
```
dwnoble, line 210: if for some reason enableGoogleMaps is undefined, make sure it defaults to a truthy value

PR #5919, opened 2026-01-21
Convert custom entities sample to use explicit schema

kmoscoe:
Also fix the config.json for the main sample to specify the format.

You can see it running at https://bullie.svl.corp.google.com:8080


keyurva (approved):
Thanks Kara!

PR #6088, opened 2026-03-17
Feature/image vulnerablities fixes

DevVegeta:


build/cdc_data/Dockerfile
```diff
@@ -74,8 +75,16 @@ COPY build/cdc_data/run.sh .
 # Make script executable.
 RUN chmod +x run.sh
 
+# Python packages upgrade to fix vulnerabilities
+RUN pip3 install --upgrade --no-cache-dir \
```
dwnoble, line 79: Rather than putting these version inline in the docker image, can they be moved to a requirements.txt file?

dwnoble (changes requested):

build/cdc_data/Dockerfile
DevVegeta: Hi @dwnoble, I encountered a conflict error after adding it to the requirements.txt file; however, upgrading the package in the Dockerfile worked as expected.

PR #6154, opened 2026-03-31
Update Saint Martin name in custom DC test goldens

juliawu:
We've been getting alerts about failing custom DC tests, that are a result of the V2 APIs returning "Saint Martin" for the name of dcid country/MFA instead of "Saint Martin (French part)". This PR updates the goldens to match V2 api behavior, which will be needed after the next data release.


juliawu:
Closing as obsolete.

PR #6161, opened 2026-04-01
[v0/v1 migration] Add v2 migration flag to production

nick-nlb:
## Description

This PR adds the `use_v2_api` flag to production, initially set to false.


gmechali (approved):
Thanks Nick! 

So as a reminder, when you submit this, all the production nodes will restart. Though it shouldnt affect the actual code running since it's set to false. But pls only submit this during work hours to ensure if there's an issue, people are ready to respond!

juliawu (approved):

PR #6176, opened 2026-04-06
Rolling production to 20% using Gemini 3 for Flash and Flash Lite from 2.5

shixiao-coder:
This change enables 20% usage of Gemini 3. Will be merged once staging and autopush looks fine


gmechali (approved):
Thanks Xiao!

PR #6178, opened 2026-04-06
Add DC_INSTRUCTIONS_DIR to env.sample in website

kmoscoe:
Also deprecate ADMIN_SECRET, since /admin page no longer exists.


keyurva (approved):
Thanks. Can you verify it works as expected before merging this?

custom_dc/env.list.sample
```diff
@@ -69,3 +76,5 @@ ENABLE_MODEL=true
 # Deprecated. Use OUTPUT_DIR instead.
 # The data path of the files stored in Google Cloud Storage, in the form gs://<BUCKET>/<FOLDER>.
 GCS_DATA_PATH=
+# Deprecated. No longer used.
```
keyurva: Can we remove this entirely?

keyurva (approved):

custom_dc/env.list.sample
kmoscoe: The entire section?

keyurva: I meant the ADMIN_SECRET part.

kmoscoe: Done.

PR #6185, opened 2026-04-07
Add support for DC_INSTRUCTIONS_DIR in Docker convenience script

kmoscoe:
This PR updates the run_cdc_dev_docker.sh script to enable both local and GCS paths for custom instructions

- adds a new section in run_service() for running with custom instructions
- refactors the run_service() function to construct Docker commands according to variable and flag settings (thanks to @clincoln8 )
- refactors the run_data function in the same way
- adds a new function to check if a service container is already running on port 8080 and kills it if so
- fixes server paths to use FLASK_ENV (I'm not sure how the previous "CUSTOM_DIR" even worked!)
- removes some details from the help

I've tested it running all types of run flags (release, latest, custom build) as well as different settings in the env.list file (local and remote gs paths).


run_cdc_dev_docker.sh
```diff
@@ -24,39 +24,14 @@ Usage:
...
-
-./run_cdc_dev_docker.sh [--env_file|-e <env.list file path>] [--actions|-a run|build_run] 
- [--image|-i <custom image name:tag>]
-
-./run_cdc_dev_docker.sh [--env_file|-e <env.list file path>] [--actions|-a run] 
-  [--schema_update|-s]
-
-All others will be ignored. The script will infer the correct container based on the 
-env.list directory settings and/or the 'build_run' setting. 
+If no options are set, the default is '--env_file $PWD/custom_dc/env.list --actions run --container all --release stable': all containers are run, using the Data Commons-provided 'stable' image.
```
keyurva: Can the changes in this file be in a separate PR? I'm not very familiar with this file.

kmoscoe: Yes, I'll split them into 2. Sorry about that.

run_cdc_dev_docker.sh
```diff
@@ -24,39 +24,14 @@ Usage:
   [--release|-r latest|stable] [--image|-i <custom image name:tag>] 
   [--package|-p <package name:tag>] [--schema_update|-s]
 
-If no options are set, the default is '--env_file $PWD/custom_dc/env.list --actions run --container all --release stable'
-All containers are run, using the Data Commons-provided 'stable' image.
```
clincoln8, line 28: Is the deletion of lines 28-51 intensional?  Is this documented somewhere else?

run_cdc_dev_docker.sh
clincoln8: nit: I think I like the "before" better where `all containers are run, using the Data Commons-provided 'stable' image.` is in a separate line 
```suggestion
If no options are set, the default is '--env_file $PWD/custom_dc/env.list --actions run --container all --release stable'

All containers are run using the Data Commons-provided 'stable' image.
```

run_cdc_dev_docker.sh
```diff
@@ -195,64 +162,68 @@ run_data() {
...
+      instructions_mount=""
+    else
+      instructions_mount="$DC_INSTRUCTIONS_DIR:$DC_INSTRUCTIONS_DIR"
     fi
-  # Regular mode
-  else
-  # Custom-built image
-  if [ -n "$IMAGE" ]; then
-    log_notice "Starting Docker services container with custom image '${IMAGE}'..."
+    log_notice "$message"
```
clincoln8: nit: move this line up to line 176 so that it comes right after it's set in the `Set IMAGE variable and output message` block

run_cdc_dev_docker.sh
```diff
@@ -195,64 +162,68 @@ run_data() {
...
-  else 
-    if [ "$RELEASE" == "latest" ]; then
-     docker pull gcr.io/datcom-ci/datacommons-services:latest
-    fi
-    log_notice "Starting Docker services container with '${RELEASE}' release..."
+    $IMAGE
+
+  # Regular mode with local instructions and local service
+  else
+    log_notice "$message"
```
clincoln8: if you move line 186 up to line 176, then this line case be deleted.

run_cdc_dev_docker.sh
```diff
@@ -195,64 +162,68 @@ run_data() {
 
 # Run service container
 run_service() {
```
clincoln8, line 184: I think we can use an array to programmatically add docker run args as needed and have a single docker command call

```
run_service() {
  local message
  
  # 1. Set IMAGE variable and output message
  if [ -n "$IMAGE" ]; then
    message="Starting Docker services container with custom image '${IMAGE}'..."
  elif [ "$RELEASE" == "latest" ]; then
    message="Starting Docker services container with latest release..."
    IMAGE="gcr.io/datcom-ci/datacommons-services:latest"
  else
    message="Starting Docker services container with stable release..."
    IMAGE="gcr.io/datcom-ci/datacommons-services:stable"
  fi

  log_notice "$message"

  # 2. Define base Docker arguments shared by all modes
  local docker_args=(
    -it
    --env-file "$ENV_FILE"
    -p 8080:8080
    -e DEBUG=true
    -v "$PWD/server/templates/custom_dc/$FLASK_ENV:/workspace/server/templates/custom_dc/$FLASK_ENV"
    -v "$PWD/static/custom_dc/$FLASK_ENV:/workspace/static/custom_dc/$FLASK_ENV"
  )

  # 3. Conditionally add GCP Credentials (needed if data OR instructions are remote)
  if [[ "$service_hybrid" == true || "$instructions_hybrid" == true ]]; then
    check_app_credentials
    docker_args+=(
      -e GOOGLE_APPLICATION_CREDENTIALS=/gcp/creds.json
      -v "$HOME/.config/gcloud/application_default_credentials.json:/gcp/creds.json:ro"
    )
  fi

  # 4. Conditionally mount Input/Output directories (Local data)
  if [[ "$service_hybrid" != true ]]; then
    docker_args+=(
      -v "$INPUT_DIR:$INPUT_DIR"
      -v "$OUTPUT_DIR:$OUTPUT_DIR"
    )
  fi

  # 5. Conditionally mount Instructions directory (Local instructions)
  if [[ "$instructions_hybrid" != true ]]; then
    docker_args+=( -v "$DC_INSTRUCTIONS_DIR:$DC_INSTRUCTIONS_DIR" )
  fi

  # 6. Execute the Docker command
  docker run "${docker_args[@]}" "$IMAGE"
}
```

clincoln8 (commented):
Thanks Kara! Just want to make sure that deletion is intentional.

run_cdc_dev_docker.sh
kmoscoe: Yes. The permutations got too numerous so the text got too long. 
The bottom line is that the script will handle invalid combinations etc., so it's not worth getting into the details here.

run_cdc_dev_docker.sh
kmoscoe: This is amazing, thank you so much for these huge improvements. I didn't know how to do fancy things in shell like append (+=), specify arguments etc.! By the way, why is the docker_args[@] array needed? i.e. why not just docker run "$docker_args" "$IMAGE" at the end?

Also, is "local" required to declare a local variable? (I just assumed that if it was declared inside a function, it's scope would be local automatically.) Or is it just an indicator to the reader?

I'm going to make all these changes, and I will do something similar for the data container too.

run_cdc_dev_docker.sh
kmoscoe: Since this is addressed in your last comment, I'll just resolve this.

run_cdc_dev_docker.sh
kmoscoe: Again, since you've refactored the entire function, I'll just resolve this.

run_cdc_dev_docker.sh
kmoscoe: Oh, I got the answer to the first question: it's actually an array rather than just a long string, is that right? Still am not sure what the @ symbol denotes. I'll look it up.

kmoscoe: got it -- preserves spaces between strings :-)

run_cdc_dev_docker.sh
```diff
@@ -163,97 +132,128 @@ upload() {
...
-  if [ "$data_hybrid" == true ]; then
+  message+="..."
+
+  log_notice "$message"
+
+# 2. Define base Docker arguments shared by all modes
+  local docker_args=(
+    -it
+    --env-file "$ENV_FILE"
+    -p 8080:8080
```
clincoln8: is this required? I don't see this on the previous commands.

clincoln8 (commented):
Nice!

clincoln8 (approved):

run_cdc_dev_docker.sh
kmoscoe: Yeah, I think you're right that it's not required. Removing.

PR #6190, opened 2026-04-09
Switching Gemini 3 to low level thinking to reduce latency

shixiao-coder:
The low level latency offers similar overall NL performance and comparable running time for evaluation against Gemini 2.5 Flash

### DIFF
--- a/server/lib/nl/detection/llm_api.py
+++ b/server/lib/nl/detection/llm_api.py
@@ -100,8 +100,7 @@ def detect_with_gemini(query: str, history: List[List[str]],
     config = types.GenerateContentConfig(
         temperature=_TEMPERATURE,
         safety_settings=_GEMINI_CONFIG.safety_settings,
-        thinking_config=types.ThinkingConfig(thinking_level="low")
-    )
+        thinking_config=types.ThinkingConfig(thinking_level="low"))
   else:
     config = _GEMINI_CONFIG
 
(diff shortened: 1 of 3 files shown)


server/lib/nl/detection/llm_api.py
```diff
@@ -96,9 +96,17 @@ def detect_with_gemini(query: str, history: List[List[str]],
       'gemini_model',
       f'{api_version}/{model_name}',
   )
+  if "gemini-3" in model_name:
```
gmechali: Can you use the exact model name(s)

gmechali: Should we make a helper funciton to generate the gemini config, pass in the name, check if it's a gemini 3 model and update the thinking budget if so? 

The static _GEMINI_CONFIG is now not enough, so it might be time :)

server/lib/utils/gemini_utils.py
```diff
@@ -42,6 +42,9 @@ def call_gemini(
       "response_mime_type": "application/json",
       "response_schema": schema
   } if schema else {}
+  if "gemini-3" in gemini_model:
```
gmechali: If you make a helper, you can reuse it here. And it probably belongs in this file anyways.

server/lib/nl/detection/llm_api.py
shixiao-coder: Guess, we need that. For any future model update, we may need to change/update configs. Doing an update on it

server/lib/utils/gemini_utils.py
shixiao-coder: The config setup is different between llm_api.py and gemini_utils.py. I'll do separate helper function for now

gmechali (approved):

PR #6199, opened 2026-04-12
[v0/v1 migration] /bulk/info/variable-group

nick-nlb:
## Issue

[b/491885197] (https://b.corp.google.com/issues/491885197)

## Description

This PR implements the migration of `v1/bulk/info/variable-group` to `v2`.

The change is gated behind the `use_v2_api` flag.

## Notes

The core of the migration is very simple: a flag-mediated gate that determines which endpoint is called. 

The complication comes (that makes up most of the diff) comes from the fact that the `v2` endpoint is no longer able to provide a `definition` along with each stat var. This definition is used (relatively rarely) in the natural language search, in order to find sibling stat vars to provide further exploration topics for the user.

Because this is no longer available in the `v2` endpoint, the functionality had to be reconstructed via direct v2 calls. 

A discussion of the methodology, testing, fidelity and latency considerations of this can be found [at this link](https://docs.google.com/document/d/1LEZI_6-2wkwpmv0ELYIKg28wqJJpsyZSZohl7CJ3-HE/edit?resourcekey=0-_UcPAttv9jeT4qas_rIGhw&tab=t.0). (Message if access is required).

This document describes how the `definition` functionality is used and analyses the latency and fidelity implications of moving that functionality to Flask and `v2`.

## Testing

There are two aspects of the NL search that are affected by the "definitions". These are described in the document as Flow 1 and Flow2.

Flow 1 is rarely invoked, but can be seen in the following query (which should produce the same results for "Related" charts populated at the bottom of the results section).

* [V2](http://localhost:8080/explore#q=population+of+people+aged+85+and+over+in+Seattle&client=ui_query)
* [V1](http://localhost:8080/explore?disable_feature=use_v2_api#q=population+of+people+aged+85+and+over+in+Seattle&client=ui_query)

Flow 2 is much more common, and is the primary driver of latency discrepancies between `v1` and `v2`. 

This flow is invoked on a standard query such as:

[Query]()

## Goldens

This PR also includes explicit directives to the integration tests to use v1. The goldens would have to be regenerated for v2 at some point before the feature flag is dropped.


server/lib/nl/common/variable_group.py
```diff
@@ -50,6 +68,18 @@ def _get_descendant_sv_nodes(
...
+  sv_definitions = {}
+  if use_v2:
+    all_child_svs = []
+    for data in resp.get("data", []):
+      if not (info := data.get("info")):
+        continue
+      for child_sv in info.get("childStatVars", []):
+        if 'id' in child_sv:
+          all_child_svs.append(child_sv['id'])
+    sv_definitions = dc.get_variable_definitions(all_child_svs)
```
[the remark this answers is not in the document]
nick-nlb: I noticed this just before creating the PR - a fix to this was staged but not committed when the PR was made. This is an artifact of an earlier version and has been removed.

server/lib/nl/common/variable.py
```diff
@@ -121,6 +128,74 @@ def limit_extended_svs(sv: str, ext_svs: Set[str], limit: int) -> Dict:
...
+  # Batch 3: Fetch variable group info for all sibling SVGs
+  svg_siblings_info = {'data': []}
+  if all_sibling_svgs:
+    svg_siblings_info = dc.get_variable_group_info(list(all_sibling_svgs), [])
+
+  # Collect all child SVs from sibling groups
+  all_sibling_child_svs = set()
+  for item in svg_siblings_info.get('data', []):
+    for c in item.get('info', {}).get('childStatVars', []):
+      if 'id' in c:
```
n-h-diaz, line 175: If we're only using child stat vars from the variable group info response in the new path, have you tried using node with "memberOf", for example https://api.datacommons.org/v2/node?key=&nodes=dc/g/Demographics&property=%3C-memberOf
(not sure if this would be any faster now, but could help when moving to spanner)

server/services/datacommons.py
```diff
@@ -390,6 +395,81 @@ def get_variable_group_info(nodes: List[str],
...
+          prop_val = vn.get("dcid") or vn.get("value")
+          if prop_val:
+            chunk_constraint_props.add(prop_val)
+
+    # Initialize merged arcs with Step 1 results
+    merged_arcs = {
+        node: node_arcs.get("arcs", {}) for node, node_arcs in data1.items()
+    }
+
+    # Step 2: Fetch the specific constraint properties for this chunk
```
n-h-diaz: Instead of making two calls, could we do something like ->* and then filter by the constraint properties?

(not sure if you've already considered this / how the performance compares)

n-h-diaz (approved):
Thanks!

I had a few comments on some alternative api fetches to consider, but not blocking for this PR

server/services/datacommons.py
nick-nlb: I just implemented this and did a performance comparison. Over the "Standard 150" queries that I've been doing the tests on so far, the results were almost identical to the double-fetch (I put them into a new tab in the doc).

However, it does simplify that section of code by quite a bit, so given the very similar performance, I've committed it.

server/lib/nl/common/variable.py
nick-nlb: I had a go at this but had trouble reproducing the same results (not a question of latency). I'll go with the current version for now but revisit after!

PR #6207, opened 2026-04-14
Defer facet enrichment until dialog opens

nick-nlb:
## Issue

[b/502594306](https://buganizer.corp.google.com/issues/502594306)

## Description

For our charts, we first make a light call to get basic facet information for that chart (so that we know how many available facets we have). Once that's resolved, we make a heavier call to enrich those facets with the metadata required for the facet modal.

A note that this used to be done via v1 calls collated on the front end. It is now a single /metadata/facet Flask endpoint that uses only v2 calls.

The issue that this bug seeks to resolve is:

The enrichment call, which is relatively heavy, is made right after the basic call to get the facets, as soon as the chart begins rendering on the frontend. However, we don't actually need any of that enriched data until the user opens the dialog.

This means that on an explore page, where we may have dozens of charts, we make dozens of metadata/facet calls, most of which never server a purpose unless the user opens the dialog.

This is not a new issue, but became more apparent during the move to the Flask endpoint.

The enrichment of data is now deferred to when the user actually opens the dialog box (at which time, a spinner will be displayed as the metadata is fetched).

## Testing

The content of the facet selector modals is not be affected by this change. However, the behavior is slightly different, in that clicking on the facet modal will produce a spinner (much like the metadata modal) before displaying the content.

This change affects the explore pages and the tools:

[Explore](http://localhost:8080/explore#q=jobs+in+texas&client=ui_query)
[Scatter](http://localhost:8080/tools/scatter#svx%3DDifferenceRelativeToBaseDate2006_Max_Temperature_RCP45%26dx%3DCount_Person%26svy%3DPercent_Person_WithCoronaryHeartDisease%26dy%3DCount_Person%26epd%3Dcountry%2FUSA%26ept%3DCounty%26qd%3D1%26dd%3D1)
[Map](http://localhost:8080/tools/map#%26sv%3DCount_Worker_NAICSAccommodationFoodServices%26pc%3D0%26denom%3DCount_Person%26pd%3DgeoId%2F48%26ept%3DCounty)
[Timeline](http://localhost:8080/tools/timeline#place=geoId%2F0606000%2CgeoId%2F2511000%2CgeoId%2F2603000%2CgeoId%2F1777005%2CgeoId%2F1225175%2CgeoId%2F4815976&statsVar=Median_Income_Person)
[Download](http://localhost:8080/tools/download#pt=Country&place=Earth&sv=Count_Person&dtType=LATEST&facets=%7B%22Count_Person%22%3A%221389069806%22%7D)

A note when opening the facet selector on the explore page: as with the metadata modal, if you open the facet selector while the page is still undergoing heavy chart loading, the spinner will spin for some time before showing the modal. This is not because the call itself is slow, but rather because the call doesn't resolve until all previously queued (unrelated) calls have finished.

This is not an issue in production, as the explore page and related charts all load much faster.


static/js/shared/hooks/use_facet_enrichment.tsx
```diff
@@ -0,0 +1,126 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { useCallback, useEffect, useMemo, useRef, useState } from "react";
+
+import { FacetSelectorFacetInfo } from "../facet_selector/facet_selector";
+
+export interface FacetEnrichmentResult {
+  /**
```
nick-nlb, line 22: Minor note here: the move to TSDoc comments here is a divergence from our older pattern of using `//`, but is the more canonical way of documenting contracts. I don't mind to go back if we prefer consistency, but otherwise, we can use TDDocs going forward.

static/js/components/subject_page/block.tsx
```diff
@@ -439,6 +441,44 @@ export function Block(props: BlockPropType): ReactElement {
...
+  }-${blockSVs.map((sv) => sv.statVar).join(",")}`;
+  const {
+    facetList: enrichedFacetList,
+    loading: enrichmentLoading,
+    onModalOpen: onFacetSelectorOpen,
+    totalFacetCount,
+  } = useFacetEnrichment(
+    facetListCacheKey,
+    facetList,
+    useCallback(async () => {
```
dwnoble, line 458: Nit: Add a comment here describing this callback

dwnoble (approved):
Thanks @nick-nlb !

PR #6210, opened 2026-04-15
Add NL sidebar for website autopush on dc-mixer-default to enable v2/resolve for SV search through NL server

shixiao-coder:
This is the first PR to try enable NL sidebar for dc-mixer-default in website

It will work on autopush, later PR will be done for staging and prod once this is fine


skaffold.yaml
```diff
@@ -219,6 +219,7 @@ profiles:
               mixer.hostProject: "datcom-website-autopush"
               mixer.serviceName: "website-esp.endpoints.datcom-website-autopush.cloud.goog"
               serviceGroups.recon.enableEmbeddingsSidecar: "false"
+              serviceGroups.default.enableEmbeddingsSidecar: "true"
```
gmechali, line 222: https://github.com/datacommonsorg/mixer/blob/303a7931ac3e5f9436571fbd0c4b6432182b1de4/deploy/helm_charts/mixer/values.yaml#L441

In mixer, we set enableEmbeddingsSidecar to true for recon, which is why the line above sets it to false for website's recon.

Is it intentional to put this in defualt? Why have it in different mixer service groups across website and mixer?

gmechali, line 222: Xiao, separate note is that this provides an added risk at deployment. So just make sure to monitor when this gets released to autopush, staging and prod and just double check this looks healthy.
Might want a note to secondary oncaller when he does the release.

clincoln8: For website, the recon workload isn't used: https://github.com/datacommonsorg/website/blob/3f077cbb82865df6380e9d15abf48bb29b3254f3/deploy/helm_charts/envs/autopush.yaml#L95-L96

resolve traffic is routed to the default workload. Which is why for website-autopush we want the embeddings server in default and not recon

clincoln8 (approved):
Thanks Xiao! Make sure to notify oncall thread about change to deployment for autopush

PR #6215, opened 2026-04-16
[DCP - Spanner] Enable Spanner Reads in CDC with Stale Reads

gmechali:
This modification to the run.sh passes in the spanner configuration into the CDC service, enabling CDC's mixer to read from their Spanner instance.

For legacy CDC instance, this will remain disabled, and Cloud SQL will remain the read path.
For CDC on DCP, the Terraform script will pass in the Spanner settings, and the USE_STALE_READS as environment variables.

TESTED: I built a custom datacommons-services image from local code using this run.sh script. Then through my terraform script, deployed the CDC service with Spanner enabled, and USE_STALE_READS enabled. I was able to read nodes inserted into my Spanner DB. Future work will enable testing the stale reads.


gmechali:
I assume that eventually we will want to stop using the datacommons-services image, and instead just ship the mixer image. If so, we will be be able to get rid of a lot of the run.sh logic. 

This should not break existing CDC clients either.

build/cdc_services/run.sh
```diff
@@ -71,6 +71,52 @@ if [[ $ENABLE_MODEL == "true" ]]; then
...
+EOF
+    MIXER_ARGS+=("--spanner_graph_info=/tmp/spanner_graph_info.yaml")
+fi
+
+if [[ $USE_STALE_READS == "true" ]]; then
+    echo "Stale reads enabled via environment variable."
+    USE_STALE_READS_FLAG="true"
+    # Stale reads requires these to be true too
+    USE_SPANNER_GRAPH="true"
+fi
```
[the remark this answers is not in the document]
gmechali: Terraform will validate the setup. Lmk if you think I should also make a change here.

n-h-diaz (approved):

build/cdc_services/run.sh
```diff
@@ -71,7 +71,42 @@ if [[ $ENABLE_MODEL == "true" ]]; then
...
+    
+    MIXER_ARGS+=("--spanner_graph_info=$SPANNER_CONFIG_YAML")
+fi
+
+# If any feature flag needs to be enabled, generate the file
+if [[ $USE_SPANNER_GRAPH == "true" ]]; then
+    cat << EOF > /tmp/cdc_feature_flags.yaml
+flags:
+  UseSpannerGraph: $USE_SPANNER_GRAPH
+  SpannerGraphDatabase: $GCP_SPANNER_DATABASE_NAME
```
n-h-diaz: (this flag overrides the spanner_graph_info.database, so doesn't need to be set if they'll just be the same)

gmechali: Done.

PR #6220, opened 2026-04-17
[Web Components] Facet selection

nick-nlb:
## Description

The various web components contain props to control the data displayed, but until now did not contain the ability to indicate specific facets.

This PR implements the first round of facet-based filtering functionality: the ability to specify facets by ID.

There are three props available, depending on how you wish to apply the facet filtering:
* **facetId**: If you wish to apply a single facet across all variables, this is the simplest way to do this. It acts as a broadcast value for all variables in the component.
* **facetIds**: If you wish to apply multiple facet IDs across multiple variables, use `facetIds`. This takes a space-delimited list of facet IDs (matching the variable list), and applies each facet ID to the respective variable positionally.
* **facetMapping**: This takes a JSON string mapping variable DCIDs to facet IDs (e.g., `'{"Count_Person": "123", "Median_Income_Household": "456"}'`). This is the **recommended** approach for clarity and unambiguity, as it explicitly links a facet to a specific variable regardless of their order in the props.

### Conflict Resolution

If a consumer provides multiple facet props simultaneously, the component resolves the facet for each variable using the following order of precedence:
1.  **`facetMapping`**: The component first checks if the variable has an explicit mapping in this JSON object.
2.  **`facetIds`**: If not found in the mapping, it checks for a facet ID at the matching index in the `facetIds` list.
3.  **`facetId`**: If neither of the above provides a facet ID, it falls back to the global `facetId`.

*Although the resolution order described above is deterministic, it is recommended only to use one of the three in any one component to avoid confusion*

### Behavior of `facetIds` (Multiple Variables)

*   **Single Value Broadcast**: As a convenience, if `facetIds` contains exactly **one** ID (e.g., `facetIds="123"`), that ID will be applied to *all* variables, effectively behaving the same as `facetId`.
*   **Partial Lists**: If `facetIds` contains more than one ID but fewer than the number of variables, variables at indices that exceed the list length will fall back to the global `facetId` (if provided) or default resolution.

## Notes

Facet Ids are currently transient in that the same facet may not have the same id over time. This will change in upcoming schema changes. However, in the meantime, use of these props should take this transience into account.

* It is better not to use hard coded facet ids directly within an embedding [i.e., like what we do on the example page] until facet ids become fixed.
* However, you can use this to flexibly update web components, for example, via an external facet selector or via an agent-based tool that uses the web-component to display facet-scoped information, because in those cases, the use of the facet and web component is not permanent, but rather using a facet id that was immediately looked up.

## Testing and Examples

There is a page available only on dev that demonstrates this functionality across the various web components.

When running the website locally, you can access this page at: 

[Dev Example Page](http://localhost:8080/dev/facet-examples)


server/routes/dev/html.py
```diff
@@ -34,3 +34,10 @@ def subject_page_tiles():
   return flask.render_template('dev/subject_page_tiles.html',
                                maps_api_key=flask.current_app.config.get(
                                    'MAPS_API_KEY', ''))
+
+
+@bp.route('/facet-examples')
+def facet_examples():
+  if os.environ.get('FLASK_ENV') == 'production':
```
keyurva: nit: Consider wrapping this in a util function like `isProd()` or `isEnvironment('production')`. Currently the `FLASK_ENV` var is accessed from multiple places and starting to keep it confined will be useful.

static/library/utils.ts
```diff
@@ -157,3 +157,51 @@ export function getObservationDateRange(response: PointApiResponse): {
...
+ * Resolves the facet ID for a given variable based on mapping, positional list, or broadcast fallback.
+ *
+ * @param variable The variable DCID.
+ * @param index The index of the variable in the list.
+ * @param facetMapping Optional parsed mapping of variable to facet ID.
+ * @param facetIds Optional positional list of facet IDs.
+ * @param facetId Optional fallback facet ID.
+ * @returns The resolved facet ID or empty string.
+ */
+export function getFacetId(
```
keyurva, line 178: Is the preference order of various facet parameters documented somewhere? If not, consider capturing it.

keyurva (approved):
Thanks Nick! This is very cool. Appreciate the super-fast turnaround!

server/routes/dev/html.py
nick-nlb: Good call! There is already a similar util function for checking if we are in a testing environment. I added one in for production that matches, and used it here.

static/library/utils.ts
nick-nlb: Done!

PR #6225, opened 2026-04-21
100% rolling up to Gemini 3

shixiao-coder:
This PR will fulling rollout Gemini from 2.5 to 3.

A following PR will be generated to remove the code logic and this flag if prod works fine with Gemini 3


gmechali (approved):
LGTM - thanks Xiao!

PR #6226, opened 2026-04-21
Modify the code to only keep Gemini 3 related setup and remove all Gemini 2.5 setup

shixiao-coder:
This PR update all calls that uses feature flag for Gemini 3 and removes all logics, constant related to Gemini 2.5


server/lib/nl/explore/related.py
gmechali: instead of setting use_think_config and gemini_model on lines 317 and 318, can you just set them here?  Avoids redundancy

server/lib/feature_flags.py
```diff
@@ -33,7 +33,6 @@
 ENABLE_STAT_VAR_AUTOCOMPLETE = 'enable_stat_var_autocomplete'
 ENABLE_NL_AGENT_DETECTOR = 'enable_nl_agent_detector'
 NEW_RANKING_PAGE = 'new_ranking_page'
-ENABLE_GEMINI_3_FLASH = 'enable_gemini_3_flash'
```
gmechali, line 36: After this PR is released to 2 prod cycles, make sure to delete the feature flag from the feature flag .json files!

2 prod cycles so that if we have to do a rollback at some point, it's still safe!

gmechali (approved):

server/lib/feature_flags.py
shixiao-coder: kk

PR #6237, opened 2026-04-24
[v0/v1 migration] Upate stat var hierarchy to match behavior between Spanner and BT

nick-nlb:
## Issue

[b/505058380](https://b.corp.google.com/issues/505058380)

## Description

When mixer is using BigTable and using the bulk variable group info endpoint to construct the hierarchy, if no filtering is applied, then mixer hardcodes all items as having data. The reason for this may have been for efficiency purposes, but it does produce inaccurate results.  The Spanner flow is producing accurate results.

However, the stat var hierarchy relies on that behavior to show all stat vars, even those that have no data, as long as no filtering is applied.

When using Spanner, certain stat var groups were showing, but without their children, because of the differences in the results that the different databases were returning.

Because Spanner is returning more accurate results in this this case, in order to restore parity between the two, we have updated the frontend to explicitly mimic the BT behavior regardless of the database being used (rather than updating mixer to have Spanner mimic BT).

### Spanner, before change

7vJwAmrKmzRH5Fi

### Spanner, after change

image

## Testing

When running locally, with spanner, the 12 UN Data Thematic Areas -> 1: Poverty and food security -> Inequality and social protection item exists, indicates that there is one child, but does not display that child.

In this branch, the child will appear (as it does when using V1).

[Stat Var Explorer](http://localhost:8080/tools/statvar)

### DIFF
--- a/static/js/stat_var_hierarchy/stat_var_group_node.tsx
+++ b/static/js/stat_var_hierarchy/stat_var_group_node.tsx
@@ -182,6 +182,8 @@ export class StatVarGroupNode extends React.Component<
         svgOnSvPath.add(path[level]);
       }
     }
+    // No filtering is active if no places, data sources, or entity count thresholds are indicated.
+    // In this case, we are in full hierarchy exploration mode.
     const noFilteringActive =
       this.props.entities.length === 0 &&
       !this.props.dataSource &&
--- a/static/js/stat_var_hierarchy/stat_var_group_node.tsx
+++ b/static/js/stat_var_hierarchy/stat_var_group_node.tsx
@@ -182,16 +182,22 @@ export class StatVarGroupNode extends React.Component<
         svgOnSvPath.add(path[level]);
       }
     }
-    const childSV = this.props.showAllSV
-      ? this.state.childSV
-      : this.state.childSV.filter(
-          (sv) => sv.hasData || sv.id in this.context.svPath
-        );
-    const childSVG = this.props.showAllSV
-      ? this.state.childSVG
-      : this.state.childSVG.filter((svg) => {
-          return svg.descendentStatVarCount > 0 || svgOnSvPath.has(svg.id);
-        });
+    const noFilteringActive =
+      this.props.entities.length === 0 &&
+      !this.props.dataSource &&
+      !this.props.numEntitiesExistence;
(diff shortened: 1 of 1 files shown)


static/js/stat_var_hierarchy/stat_var_group_node.tsx
```diff
@@ -182,16 +182,23 @@ export class StatVarGroupNode extends React.Component<
...
+      this.props.entities.length === 0 &&
+      !this.props.dataSource &&
+      !this.props.numEntitiesExistence;
+    const childSV = noFilteringActive
+      ? this.state.childSV.map((sv) => ({ ...sv, hasData: true }))
+      : this.props.showAllSV
       ? this.state.childSV
       : this.state.childSV.filter(
           (sv) => sv.hasData || sv.id in this.context.svPath
         );
```
juliawu: Supernit: Could you reformat this to make the nested logic more clear? Something like
```typescript
const childSV = noFilteringActive
  ? this.state.childSV.map((sv) => ({ ...sv, hasData: true }))
  : (this.props.showAllSV 
      ? this.state.childSV 
      : this.state.childSV.filter(sv => sv.hasData || sv.id in this.context.svPath)
    );
```

static/js/stat_var_hierarchy/stat_var_group_node.tsx
```diff
@@ -182,16 +182,23 @@ export class StatVarGroupNode extends React.Component<
         svgOnSvPath.add(path[level]);
       }
     }
-    const childSV = this.props.showAllSV
+    const noFilteringActive =
```
juliawu, line 187: Supernit: Could you add a quick comment on what noFilteringActive means, that way future maintainers without the original context don't have to parse through the logic?

juliawu (approved):
Thanks for the fix! Just some nits.

static/js/stat_var_hierarchy/stat_var_group_node.tsx
nick-nlb: I like this formatting suggestion, but the prettier won't allow it! (It reverts to the original).  Really a sign that the ternary operator isn't the best approach here, so I've turned it into an "if/else".

PR #6238, opened 2026-04-24
[v0/v1 migration] Add fixed ordering for top-level hierarchy nodes

nick-nlb:
## Issue

[http://b/505060318](https://b.corp.google.com/issues/505060318)

## Description

The front end was previously rendering the top level of the stat var hierarchy in the order returned by the API. From the perspective of both mixer and the website, this order was effectively non-deterministic (i.e., the order was not being enforced in either mixer or the website), and was therefore being effectively curated upstream.

With the migration to Spanner, the order being returned was no longer the same. 

With this PR, we are now enforcing Data Commons root nodes to be pinned in the order that matches the current production site (rather than relying on an arbitrary order from the database).

### Before

image

### After

image


static/js/stat_var_hierarchy/stat_var_hierarchy.tsx
```diff
@@ -302,6 +302,22 @@ export class StatVarHierarchy extends React.Component<
...
+              STAT_VAR_HIERARCHY_CONFIG.topLevelPinnedOrder?.[a.id] ?? Infinity;
+            const orderB =
+              STAT_VAR_HIERARCHY_CONFIG.topLevelPinnedOrder?.[b.id] ?? Infinity;
+
+            if (orderA !== orderB) {
+              return orderA - orderB;
+            }
+            return (a.displayName || "").localeCompare(b.displayName || "");
+          });
+        }
```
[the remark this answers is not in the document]
nick-nlb: Agreed with the removal of the restriction to `dc/g/Root`. However the comment about the alphabetical order made me reconsider the alphabetical sort that I did put in. Spanner is not arbitrary in what it returns (and is doing sorting). It just didn't match the curated top-level node that we are now pinning. I removed the alphabetical sort. If we don't have pinning, we will fall back to whatever comes out of the database.

static/js/shared/ga_events.test.tsx
juliawu: Should this also be updated to index 1?

juliawu (approved):
Thanks for the fix!

static/js/shared/ga_events.test.tsx
nick-nlb: Good catch!

PR #6239, opened 2026-04-24
[new Ranking Pages] Add 18n messages for ranking page titles with "the World"

juliawu:
As a followup to the comments in #6235 discussing translations, this PR adds translation strings for ranking page titles that contain "the World", so that they will be translated the next time we submit strings for translation. There is no change to UI or functionality.


server/routes/ranking/html.py
```diff
@@ -35,13 +35,16 @@ def ranking(stat_var, place_type, place_dcid=''):
       place_name = place_dcid
   else:
     place_dcid = 'Earth'
-    # TODO(juliawu): Add localization for page titles that use "the World".
+    # TODO(juliawu): Once we have translations for "the World" page titles in i18n_ranking_messages.ts,
+    # remove this default place name and use the translated "world" page titles instead.
```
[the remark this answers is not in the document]
juliawu: While added, they still need to be sent out for translation. I've updated the wording to clarify.

static/js/i18n/i18n_ranking_messages.ts
```diff
@@ -53,4 +65,16 @@ export const rankingMessages = defineMessages({
     description:
       "Subtitle of the page, which shows ranking of the top / highest {rankSize} contained places of a type within a place, computed on a per capita basis, where {rankSize} will be replaced by a number, {pluralPlaceType} will be replaced by the place type of the contained places (could be cities, countries, counties), and {placeName} is the containing place. Please maintain the {rankSize}, {placeName} and {pluralPlaceType} as is in the final translation, and use a gender neutral structure that conveys the same meaning (e.g. dashes to separate).",
   },
+  topPlacesWorldTitle: {
+    id: "ranking-subtitle_top_world",
+    defaultMessage: "Top {rankSize} {pluralPlaceType} in the World",
```
nick-nlb: With `{placeName}`, this gets interpolated as a link. Here with the raw text, as it is the "the World" will not be wrapped in an anchor.

We'll only deal with the actual link creation later, but to prepare for it before these get translated, we can wrap "the World" in a rich text tag: `Top {rankSize} {pluralPlaceType} in <worldLink>the World</worldLink>`.

The translators will then be able to rearrange the string as they see fit for their languages (i.e. `{pluralPlaceType} del <worldLink>mundo</worldLink>`)

juliawu: Great callout! Wrapped "the World" with `<worldLink>` and updated the descriptions.

nick-nlb (approved):
LGTM!

PR #6240, opened 2026-04-24
[Vis Tools] Remove remaining references to vis tool flag

juliawu:
Remove remaining references to the `standardized_vis_tool` feature flag.

# Context
Now that the older visualization tool experiences have been removed in PRs #6170 #6168 #6167 and #6165, this PR removes the remaining references to the `standardized_vis_tool` feature flag. This includes some logic around loading the example links, which are effectively now a no-op because the feature flag is now always "True".

Removing the feature flag entry from the `<env>.json` files will come in a separate PR, so that the flag is removed after changes in this PR are reflected in production.

No changes to UI or functionality were made.

# Testing Strategy
Manually verified that each of the three tools (/tools/map, /tools/scatter, /tools/timeline) have their example links load and behave as expected.


nick-nlb (approved):
LGTM!

PR #6249, opened 2026-04-28
Add a loading spinner to stat var hierarchy nodes

juliawu:
Add a loading spinner to the stat var hiearchy nodes

This PR addresses b/505059972 (Google corp access required).

# Context
Because the new Spanner backed APIs are not yet optimized, we're experiencing some latency when interacting with the stat var hierarchy. Now, there can be a delay between when a user clicks on a node and when the sub-variables load and are shown on screen. Unfortunately, this delay looks like a "freeze" to users, even though the data fetch is still loading in the background.

# The fix
This PR makes the following adjustments to reduce user frustration:

(1) Adds a loading spinner while sub-variables are loading. Now, the node expands immediately on click, and a loading spinner is displayed inside the expanded area if the data is still in flight.

(2) When a loading spinner is present, don't show the "No available statistical variables" yet (because they're still loading).

# Screen recording

https://github.com/user-attachments/assets/6aeb425e-5d4a-4c6e-ba30-8d5e324c3246


static/js/stat_var_hierarchy/stat_var_hierarchy.tsx
```diff
@@ -259,6 +262,7 @@ export class StatVarHierarchy extends React.Component<
 
   private async fetchData(): Promise<void> {
     loadSpinner(SV_HIERARCHY_SECTION_ID);
+    this.setState({ isLoading: true });
```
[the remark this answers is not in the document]
juliawu: Instead of checking state, I added an abortController to the underlying data fetch, so that we can cancel the stale request. This approach has the added benefit of also saving network bandwidth.

static/js/stat_var_hierarchy/stat_var_group_node.tsx
```diff
@@ -242,6 +249,14 @@ export class StatVarGroupNode extends React.Component<
           }
         >
           <>
+            {this.state.isLoading &&
+              _.isEmpty(this.state.childSV) &&
+              _.isEmpty(this.state.childSVG) && (
+                <div className="node-loading-spinner">
```
nick-nlb, line 256: We have a `ProgressActivity` spinner that we could use here: 

```
<div className="node-loading-spinner">
  <div className="spinner-circle"></div>
  <span>Loading...</span>
</div>

<div className="node-loading-spinner">
  <ProgressActivity />
  <span>Loading...</span>
</div>
```

It won't look exactly like the bespoke one here, but it is customizable in terms of color and size if necessary.

static/js/stat_var_hierarchy/stat_var_group_node.tsx
```diff
@@ -272,6 +287,8 @@ export class StatVarGroupNode extends React.Component<
   }
 
   private fetchData(): void {
+    this._dataAbortController?.abort();
+    this._dataAbortController = new AbortController();
```
nick-nlb: For this underscore prefix here: we have a bit of histocal inconsistency in the code base for how we name/prefix private attributes in class-based components. Within this component itself, other attributes like `delayTimer` are private as well, so for consistency, I would go without the underscore here.

With TypeScript, underscores are redundant, as we can explicitly make fields private etc (although we haven't done it in this class, and for consistency could leave it public).

static/css/stat_var_hierarchy.scss
```diff
@@ -226,3 +226,22 @@ $small-font-size: 0.75rem;
   font-style: italic;
   font-size: 0.8rem;
 }
+
+.node-loading-spinner {
+  display: flex;
+  align-items: center;
+  gap: 8px;
+  padding: 4px 15px;
+  color: #ffffff;
```
nick-nlb: I like the use of the CSS variables below. I see we don't have a dc-white (we have some other whites, but those are not semantically the same thing, even if the color is the same). Can we add in a dc-white and use that here? (We have a corresponding dc-white in the theme as well, which is also a pure white).

nick-nlb (approved):
Looks great, and the loading state makes a huge difference given the added latency we are currently facing.

Some minor comments, but otherwise LGTM.

static/js/stat_var_hierarchy/stat_var_group_node.tsx
```diff
@@ -111,6 +114,7 @@ export class StatVarGroupNode extends React.Component<
       errorMessage: "",
       dataFetchedEntities: null,
       selectionCount: 0,
+      isLoading: false,
     };
     this.highlightedStatVar = React.createRef();
     this.dataFetchingEntities = null;
```
nick-nlb, line 120: One minor thing to add here: we're initializing some items here, but not the dataAbortController.

Our project's types are loose (i.e., TypeScript doesn't complain about the fact that the AbortController can (and for a while is) undefined. For the future possibility of us increasing TS type strictness (which I hope we can do one day), can we either initialize it here, or most simply make it `AbortController | undefined`?

static/js/stat_var_hierarchy/stat_var_group_node.tsx
juliawu: Good callout. Removed the leading underscores.

static/js/stat_var_hierarchy/stat_var_group_node.tsx
juliawu: Great callout. Initialized the controllers in both stat_var_group_node.tsx and stat_var_hierarchy.tsx.

static/js/stat_var_hierarchy/stat_var_group_node.tsx
juliawu: Switched to using ProgressActivity.

static/css/stat_var_hierarchy.scss
juliawu: Used the CSS variable for color, I didn't see good entries for spacing.

PR #6256, opened 2026-04-30
Hide the explore search bar from the stat var explorer and data download tools.

nick-nlb:
## Issue

[b/508324881](b/508324881) [internal]

## Description

The three primary visualization tools  hide the search bar in the header. These are now extended to the stat var explorer and the data download tools.


juliawu (approved):
Thank you for the fix!

PR #6258, opened 2026-05-01
Remove v2 post `bulk/info/variable-group` endpoint definition construction

nick-nlb:
## Issue

[b/507068100](b/507068100)

## Related PRs

[Mixer: 1877](https://github.com/datacommonsorg/mixer/pull/1877)

If the above PR is not yet merged into master, you should have the above branch active in your local mixer.

## Description

With Spanner soon to be supporting definitions directly from the `v2/bulk/info/variable-group` endpoint, the costly frontend definitions calculations are now removed. Both `v1` and `v2` operate on the same path, under the assumptions that the definitions will be provided. 

This PR involves a significant reversion of Flask-side definition construction. However, major improvements introduced during the initial work on the `v2` definitions have been kept, most notably query batching to remove an N+1 that had always existed in the original `v1` path.

## Notes

With the addition of the `include_definitions` parameter to the Spanner call (see 1877 linked above), we can now selectively determine which calls to the `v2/bulk/info/variable-group` endpoint calls return definitions. This is done for the sake of latency: calls that require definitions take a non-trivial amount of additional time and compute to resolve, particularly when the call also requests additional filtering.

We take the following approach:

* Calls made in the NL search (which do not involve additional filtering, but do need definitions) request definitions.
* Calls made directly to the flask `variable-group` endpoint (used only by the hierarchy) never need definitions, and so always exclude them.

## Testing

Explore searches should provide largely similar results through either Spanner or BT backed mixers . 

Results may not be exactly the same. For example, in the following search, the first graph provides "Language Spoken at Home" as an explore more option from BT, but does not from Spanner (you can see this below using production for the BT version, but the same result will happen locally). 

This is due to the backfilled Spanner definitions being different in some cases (i.e., in the case below, the base and comparison SVs differ by more than one constraint in the Spanner definitions). The root of this discrepancy is unknown, but it is not due to the functionality of this PR.

[Local search: Population of the United States](http://localhost:8080/explore?enable_feature=enable_stat_var_autocomplete#q=Population+of+the+United+States&client=ui_query)

[Production search: Population of the United States](https://datacommons.org/explore?enable_feature=enable_stat_var_autocomplete#q=Population+of+the+United+States&client=ui_query)

Merging Notes

🚨 **This PR should not be merged in until after the above linked 1877 PR in Mixer**


juliawu (approved):
Thanks for the updates!

PR #6263, opened 2026-05-05
[v0/v1 migration]  Fix cache issue for feature flag overrides in the variable ancestor endpoint

nick-nlb:
## Description

This PR moves the memoization to a helper function that takes the feature flag into account, preventing cache poisoning when the flag override is given on the URL.


juliawu (approved):
Thank you for the fix!

PR #6261, opened 2026-05-05
Update screenshot suite URLs

beets:
This PR curates the list of URLs used for Percy screenshot tests in urls.json. The previous list was excessively long (over 1000 lines) and repetitive, while still lacking coverage for several key features and parameters.

### Key Changes
- Pruned Repetitive URLs: Reduced the file size from over 1000 lines to a curated set of ~150 URLs, focusing on diversity of topics and geographies.
- Improved Feature Coverage: Added missing pages and features to ensure full coverage of the site:
  - Static Pages: Added /about, /faq, and data sources.
  - Landing Pages: Added /place and /browser.
  - Explore with Parameters: Added examples with specific parameters for pinned maps/topics (Eurostat and Census data).
  - Ranking Pages: Added specific ranking pages for different place types and geographies (e.g., US Zip codes, India States).
- Disaster Events: Ensured coverage for specific event pages (e.g., Drought in Algeria).
- Curated Explore Queries: Pruned the long list of natural language queries to a small, diverse set covering different topics (Health, Econ, Education, Environment, Fallbacks) and geographies (US, India, Norway, Iran, Canada, Pakistan).

### Verification
Verified that the updated file parses correctly as valid JSON.

PR #6280, opened 2026-05-11
Fix screenshot URL typo

beets:
I missed the starting slash in one URL from the last PR #6261 :( ptal..


juliawu (approved):
Thanks for the fix!

PR #6292, opened 2026-05-18
[DCP - Enable V2 API] Enable the V2 API on the website when USE_SPANNER is enabled.

gmechali:
This PR copies over the default custom.yaml for custom DC mixer flags. note that this will only ever be used for DCP customers. 
CDC customers never had to use a custom.yaml file.

We also update the v2 api flag based on the use_spanner_graph variable. 

https://github.com/datacommonsorg/mixer/pull/1902 See

### DIFF
--- a/build/cdc_services/run.sh
+++ b/build/cdc_services/run.sh
@@ -15,10 +15,6 @@
 
 set -e
 
-# Workaround for Go runtime.Caller path issue when mounting local binary
-# mkdir -p /workspace/mixer
-# ln -sf /workspace/internal /workspace/mixer/internal
-
 export MIXER_API_KEY=$DC_API_KEY
 # https://stackoverflow.com/a/62703850
 export TOKENIZERS_PARALLELISM=false
--- a/build/cdc_services/run.sh
+++ b/build/cdc_services/run.sh
@@ -15,10 +15,6 @@
 
 set -e
 
-# Workaround for Go runtime.Caller path issue when mounting local binary
-mkdir -p /workspace/mixer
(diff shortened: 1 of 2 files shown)


build/cdc_services/run.sh
```diff
@@ -86,8 +104,8 @@ if [[ $USE_SPANNER_GRAPH == "true" ]]; then
     
     SPANNER_CONFIG_YAML="{project: \"$GCP_PROJECT_ID\", instance: \"$GCP_SPANNER_INSTANCE_ID\", database: \"$GCP_SPANNER_DATABASE_NAME\"}"
     
-    MIXER_ARGS+=("--spanner_graph_info=$SPANNER_CONFIG_YAML" "--use_spanner_graph=true")
-    # Until V2 APIs are enabled 100%, use the flag --use_v2_api=true on your docker run command
+    # 2. Enable V2 API for Mixer
+    MIXER_ARGS+=("--spanner_graph_info=$SPANNER_CONFIG_YAML" "--use_spanner_graph=true" "--feature_flags_path=/workspace/mixer/deploy/featureflags/custom.yaml")
```
clincoln8: i don't think you need the feature flags if you're setting the mixer flag --use_spanner_graph=true

dwnoble (approved):

build/cdc_services/run.sh
gmechali: Thanks Christie, chatted offline. We used to need both use_spanner_graph and V2Divert but Natalie changed that a few weeks ago and I had forgotten. 

Removed the mixer feature flag. We just need to enable_v2_api for the website feature flags.

PR #6293, opened 2026-05-18
[screenshot] update wait times

beets:
Increased screenshot wait times
Removed ranking feature flag

### DIFF
--- a/tools/release_automation/percy_snapshot_script/per_environment_percy_snapshots.py
+++ b/tools/release_automation/percy_snapshot_script/per_environment_percy_snapshots.py
@@ -38,6 +38,7 @@ from selenium.webdriver.support import expected_conditions as EC
 from selenium.webdriver.support.ui import WebDriverWait
 
 WAIT_TIMEOUT = 15
+PAGE_LOAD_SECONDS = 60
 URLS_FILE = "urls.json"
 
 
@@ -99,13 +100,13 @@ def take_single_snapshot(name: str, url: str):
     try:
       WebDriverWait(driver, WAIT_TIMEOUT).until(
           EC.presence_of_element_located((By.TAG_NAME, "body")))
-      time.sleep(5)  # Additional sleep to ensure full rendering
+      time.sleep(PAGE_LOAD_SECONDS)  # Additional sleep to ensure full rendering
       print(f"[PID {os.getpid()}] Page '{name}' loaded. ✅")
     except TimeoutException:
       print(
           f"[PID {os.getpid()}] Warning: Timed out waiting for primary element (body) on '{name}'. Proceeding with fallback sleep. ⏳"
       )
-      time.sleep(5)
+      time.sleep(PAGE_LOAD_SECONDS)
 
     percy_snapshot(driver, name, responsiveSnapshotCapture=True)
     print(f"[PID {os.getpid()}] Snapshot taken for '{name}'. Done. 🎉")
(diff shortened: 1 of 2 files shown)


tools/release_automation/percy_snapshot_script/per_environment_percy_snapshots.py
```diff
@@ -99,13 +100,13 @@ def take_single_snapshot(name: str, url: str):
...
           EC.presence_of_element_located((By.TAG_NAME, "body")))
-      time.sleep(5)  # Additional sleep to ensure full rendering
+      time.sleep(PAGE_LOAD_SECONDS)  # Additional sleep to ensure full rendering
       print(f"[PID {os.getpid()}] Page '{name}' loaded. ✅")
     except TimeoutException:
       print(
           f"[PID {os.getpid()}] Warning: Timed out waiting for primary element (body) on '{name}'. Proceeding with fallback sleep. ⏳"
       )
-      time.sleep(5)
+      time.sleep(PAGE_LOAD_SECONDS)
```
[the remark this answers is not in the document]
nick-nlb: This really less about this PR, and more about how this was originally set up: i.e. if we wait up to 15 seconds (`WAIT_TIMEOUT`) for the body tag to load, and then give it another hard 5 seconds (or 60 seconds), then the question is: why not just include that time in the original `WebDriverWait`. 

We're already allowing a maximum of 15 + 60 seconds for the body to load, so why not just allow 60 seconds for the body to load directly in the WebDriverWait. That way the 60 second fallback can be broken out of rather than being hard.

Then we would have the overall PAGE_LOAD_WAIT happen outside of the try/except (since it is really unrelated to the body tag, and rather for giving time for the pages themselves to load).

It is not something we need to deal with here, but maybe we could look at a TODO to make this part more efficient.

tools/release_automation/percy_snapshot_script/per_environment_percy_snapshots.py
```diff
@@ -38,6 +38,7 @@
 from selenium.webdriver.support.ui import WebDriverWait
 
 WAIT_TIMEOUT = 15
+PAGE_LOAD_SECONDS = 60
```
nick-nlb, line 41: This will include a constant 60 second per page wait on every page, including pages where we probably don't need it. 

I'm thinking about pages like the home page, `/version`, `/about`, `/faq`. 

We could add a flag into the urls.json to flag certain pages a not needing a long wait (because there is extensive CSR, we do still need a short wait, but that could be very short).

nick-nlb, line 41: This bumps the wait from 5 seconds (which we have seen in practice is too short) to 60 seconds. The only downside of 60 seconds is the longer wait time for the tests to run.

Should we try a wait between those two as a first go, to see if that resolves the partial page loads, before increasing it? Maybe even 10 or 15 seconds would be enough, and if not, we could bump it up.

nick-nlb: In the long turn, it would be nice to turn the hard waits into soft waits (i.e., for the appearance or disappearance of a particular element). However, that isn't trivial in our case, because the snapshot test is not page-specific (but the logic required would be). This is something we can revisit (and going forward with any new repos where we use Percy, we can consider this ahead of time. For now, I think we have to keep some form of hard wait.

PR #6301, opened 2026-05-20
Auto-iterate all /v2/node response pages

clincoln8:
* also ensures we don't publish to `:latest` tag for `cdc-services` when pushing dev images

### DIFF
--- a/build/ci/cloudbuild.push_image.yaml
+++ b/build/ci/cloudbuild.push_image.yaml
@@ -38,11 +38,8 @@ steps:
         DOCKER_BUILDKIT=1
         # Bundle container with website, esp and mixer.
         docker build -f build/cdc_services/Dockerfile \
-          --tag gcr.io/$_PROJECT_ID/datacommons-services:${_TAG} \
-          --tag gcr.io/$_PROJECT_ID/datacommons-services:latest \
-          .
+          --tag gcr.io/$_PROJECT_ID/datacommons-services:${_TAG} .
         docker push gcr.io/$_PROJECT_ID/datacommons-services:${_TAG}
-        docker push gcr.io/$_PROJECT_ID/datacommons-services:latest
     waitFor: ["-"]
 
   - id: push-nl-server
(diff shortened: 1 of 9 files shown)


build/ci/cloudbuild.push_image.yaml
```diff
@@ -76,11 +73,8 @@ steps:
         set -e
         DOCKER_BUILDKIT=1
         docker build -f build/website_cron_testing/Dockerfile \
-          --tag gcr.io/$_PROJECT_ID/website-cron-testing:${_TAG} \
-          --tag gcr.io/$_PROJECT_ID/website-cron-testing:latest \
-          .
+          --tag gcr.io/$_PROJECT_ID/website-cron-testing:${_TAG} .
         docker push gcr.io/$_PROJECT_ID/website-cron-testing:${_TAG}
-        docker push gcr.io/$_PROJECT_ID/website-cron-testing:latest
```
gmechali, line 83: Is this also connected to a build trigger and therefore this would make us never push the latest tag? 

Can you just add a check that if _TAG == dev, we don't push latest?

server/config/feature_flag_configs/autopush.json
```diff
@@ -41,6 +47,12 @@
     "owner": "gmechali",
     "description": "Enable stat var autocompletion in the NL Search bar"
   },
+  {
```
gmechali, line 50: Changes to feature flag files should be in their own PR pls :)

server/services/datacommons.py
```diff
@@ -348,6 +353,20 @@ def v2node_paginated(nodes, prop, max_pages=1):
...
     """
+  # We use a sentinel object (DEFAULT_MAX_PAGES) as the default value to
+  # distinguish between the caller not passing max_pages vs. explicitly
+  # passing None.
+  #
+  # - If caller does not pass max_pages, it gets DEFAULT_MAX_PAGES. We then
+  #   use the feature flag to decide the default behavior.
+  # - If caller explicitly passes None, max_pages is None, and we preserve
+  #   the behavior of auto-iterating all pages.
+  if max_pages is DEFAULT_MAX_PAGES:
```
gmechali, line 364: Is there a test we could write to verify this and ensure it doesnt break?

clincoln8:
Will follow up with a different approach where we enable fetchAll on a case-by-case basis

PR #6306, opened 2026-05-20
Enable all-page fetch for StatVar properties in detect-and-fulfill

clincoln8:
StatVar names were falling back to DCIDs when requesting properties for large lists because the results were being truncated by the default Spanner page size limit of 500.

This PR conditionally enables fetching all pages for `property_values` calls in variable.py, controlled by the `enable_nl_v2node_fetchall` feature flag

### DIFF
--- a/server/lib/nl/common/variable.py
+++ b/server/lib/nl/common/variable.py
@@ -23,6 +23,7 @@ import server.lib.fetch as fetch
 import server.lib.nl.common.constants as constants
 import server.lib.nl.common.topic as topic
 import server.lib.nl.common.utils as utils
+from server.lib.feature_flags import is_feature_enabled, ENABLE_NL_V2NODE_FETCHALL
 from server.lib.nl.explore.params import DCNames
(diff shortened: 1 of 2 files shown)


server/lib/nl/common/variable.py
```diff
@@ -132,9 +136,13 @@ def _fetch_indirect_siblings(
   # Batch 1: Fetch parents for all identified SVGs
   svgs_to_expand = list({sv2svg[sv] for sv in svs_needing_indirect})
   parents_resp = {}
+  max_pages = None if is_feature_enabled(
```
keyurva: Consider a `_get_max_pages()` helper since this is used in a few places.

keyurva (approved):
Thanks!

gmechali (approved):
Thanks Christie!!

server/lib/nl/common/variable.py
clincoln8: done

PR #6323, opened 2026-05-28
enable NL sidecar for dc-mixer-default in website prod

shixiao-coder:
The eval test result shows almost no variable eval score changes.

On individual level we saw a few differences here: https://docs.google.com/spreadsheets/d/1wnoMOLMzxulQM9eWmitGiIAOv4q-NKVTLwpWQnXpeQI/edit?resourcekey=0-CiQOAQ6-xS1-xFEguji0Ig&gid=1399122854#gid=1399122854

When debugging locally, it dues to the unstable LLM check the query part (multiSV v.s. singleSV or different metric), v2/resolve itself is not the impacting part


clincoln8 (approved):
Thanks!

PR #6325, opened 2026-05-29
Add new test index for multi-entity statvars

ajaits:
Create new index for base NL statvars with additional multi-entity statvars.


keyurva (approved):
Thanks Ajai!

PR #6327, opened 2026-06-01
add the mixer feature flags in Dockerfile and use the custom.yaml in DCP

shixiao-coder:
This PR copy the feature flag folder in custom DC Dockerfile

It also add logic to include feature_flags_path for DCP only case, it will read the feature flag when use spanner graph.

We also add the logic to read an env variable RESOLVE_WITH_SPANNER_EMBEDDINGS to tune the yaml to do resolve through spanner embeddings


build/cdc_services/Dockerfile
```diff
@@ -153,6 +153,7 @@ WORKDIR /workspace
 COPY --from=mixer /workspace/bin/ bin
 COPY --from=mixer /workspace/esp/ esp
 COPY --from=mixer /workspace/internal/ internal
+COPY --from=mixer /workspace/deploy/featureflags deploy/featureflags
```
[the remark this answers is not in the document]
shixiao-coder: copied the folder into workdir

build/cdc_services/run.sh
```diff
@@ -108,6 +108,29 @@ except Exception as e:
     
     # 2. Enable V2 API for Mixer
     MIXER_ARGS+=("--spanner_graph_info=$SPANNER_CONFIG_YAML" "--use_spanner_graph=true")
+
+    # 3. Use mixer custom feature flags
+    MIXER_ARGS+=('--feature_flags_path=deploy/featureflags/custom.yaml')
+
+    if [[ $RESOLVE_WITH_SPANNER_EMBEDDINGS == "true" ]]; then
```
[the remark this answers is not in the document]
shixiao-coder: updated

gmechali, line 113: Why do you need this? 

For website we needed it because the feature_flags file is also used by the CDC stable image, so we couldnt just override that file.

The custom.yaml you added is only for DCP. So is there a reason why you can't just set the variables in the custom.yaml?

FYI this override logic for website feature flags is meant to be deleted when we stop building the CDC stable image.

shixiao-coder: updated the flags into the mixer custom.yaml and use the env to control reading from the yaml

build/cdc_services/run.sh
```diff
@@ -108,6 +108,11 @@ except Exception as e:
     
     # 2. Enable V2 API for Mixer
     MIXER_ARGS+=("--spanner_graph_info=$SPANNER_CONFIG_YAML" "--use_spanner_graph=true")
+
+    # 3. Use mixer custom feature flags
```
gmechali, line 112: Can you add a comment that for now this just contorls the flag for reading embeddings for spanner, and that eventually we should ALWAUYS pass in the fearture flags!

gmechali (approved):

server/integration_tests/test_data/detection_api_context/stateswithhighestphds/chart_config.json
```diff
@@ -1,6 +1,16 @@
 {
   "childEntityType": "State",
   "classifications": [
+    {
+      "quantity": {
```
gmechali, line 5: Why this change?

shixiao-coder: This is just a flaky Golden change, it sometimes add a trivial value 2.2250738585072014e-308 to to output and we are simply doing text compare and causing the Golden to fail occasionaly

build/cdc_services/run.sh
shixiao-coder: added

PR #6328, opened 2026-06-01
Fix contained related places when pagination occurs

nick-nlb:
## Issue

b/515861634

## Description

When using Spanner, the related places at the bottom of a the places page could be either incomplete or not shown at all. An example of this is the place page for [San Mateo County](http://localhost:8080/place/geoId/06081), where a BT-backed mixer will show a set of places, and a Spanner-backed mixer will not (see screenshots below).

The reason for this was Spanner pagination combined with the way that the Flask backend was making the request. 

Flask was requesting all containedInPlace nodes, regardless of the type, and then filtering out to show relevant child place types within Flask. San Mateo, as an example, contained hundreds of non-place contained nodes (EPAReportingFacility, VehicleCrashIncidents) that are pulled out by the query before any relevant places appear.  With BT, this worked (albeit with unnecessarily large payloads). With Spanner, the pagination means that no relevant places are retrieved on the first page, and no places are then displayed.

## Solution

This PR implements two approaches to solve this problem. The first is to resolve all paginated pages in the `raw_property_values` call. This involved updating the `raw_property_values` in a similar way to how `property_values` had already been updated, in order to allow pagination to be handled. 

This in itself does solve the problem, but the multiple Spanner calls to resolve the pagination add in latency (for the sake of retrieving nodes, many of which we do not need). 

The second arm of the solution is to update the property values call itself to filter by the place types that we are interested, giving us a payload that only consists of types that we will use. In practice, with the Mixer-level filtering in place, pagination will be invoked less often.

## Notes

There was a pre-existing inefficiency in the code whereby the original code fetched all the child places in order to determine the place types, and then later fetched the child places again.

I've updated this to now fetch the child places and the place types at the same time (returning them both at the same time). The new function that does this is largely similar to the the previous one, but I left the previous one intact as it is to limit the surface of the changes (adding in a TODO to migrate other uses of the functions to the new function and to remove the older function).

### Local

In master, this page will produce results locally when using a Mixer backed with BT, and no results when using a mixer backed with Spanner.

In this branch, Spanner will give the same results as BT.

[San Mateo County](http://localhost:8080/place/geoId/06081)

### Production

Production can also serve as a working reference of the page (see the "Places in San Mateo County" section at the bottom of the page).

[San Mateo County](https://datacommons.org/place/geoId/06081)

## Screenshot

The following is the section that is missing with Spanner in master (but available otherwise, and now with this PR in Spanner).

image

### DIFF
--- a/server/lib/fetch.py
+++ b/server/lib/fetch.py
@@ -453,7 +453,7 @@ def multiple_property_values(nodes: List[str],
   return result
 
 
-def raw_property_values(nodes, prop, out=True, constraints=''):
+def raw_property_values(nodes, prop, out=True, constraints='', max_pages=1):
   """Returns full property values data out of REST API response.
 
   The response is the following format:
@@ -470,8 +470,16 @@ def raw_property_values(nodes, prop, out=True, constraints=''):
   }
 
   """
-  resp = dc.v2node(nodes, '{}{}{}'.format('->' if out else '<-', prop,
-                                          constraints))
+  from server.lib.feature_flags import is_feature_enabled
+  from server.lib.feature_flags import USE_V2_API
+
+  if is_feature_enabled(USE_V2_API):
+    resp = dc.v2node_paginated(
+        nodes, '{}{}{}'.format('->' if out else '<-', prop, constraints),
+        max_pages)
+  else:
+    resp = dc.v2node(nodes, '{}{}{}'.format('->' if out else '<-', prop,
+                                            constraints))
   result = {}
   for node, node_arcs in resp.get('data', {}).items():
     result[node] = node_arcs.get('arcs', {}).get(prop, {}).get('nodes', [])
--- a/server/lib/fetch.py
+++ b/server/lib/fetch.py
@@ -454,7 +454,7 @@ def multiple_property_values(nodes: List[str],
 
 
 def raw_property_values(nodes, prop, out=True, constraints='', max_pages=1):
-  """Returns full property values data out of REST API response.
+  """Returns raw property value dictionaries from the REST API, fetching up to max_pages of results.
 
   The response is the following format:
   {
--- a/server/routes/place/utils.py
+++ b/server/routes/place/utils.py
@@ -668,6 +668,72 @@ def get_child_place_types(place: Place) -> list[str]:
   return []
 
 
+# TODO(nick-nlb): look at other uses of the very similar get_child_place_types, and migrate them
+#                 to use this function instead.
+def get_child_places_by_type(
+    place: Place) -> Tuple[List[str], Dict[str, List[str]]]:
+  """
+  Determines the child place types for a given place and returns a tuple containing the
+  matched types and the places grouped by type.
+  """
+  ordered_child_place_types = []
+  # Attempt to directly match a child place type using the custom expressions.
+  for f in PLACE_MATCH_EXPRESSIONS:
+    matched_child_place_type = f(place)
+    if matched_child_place_type:
+      ordered_child_place_types = [matched_child_place_type]
+      break
+
+  if not ordered_child_place_types:
+    # Determine the order of child place types based on the parent type.
+    ordered_child_place_types = DEFAULT_CHILD_PLACE_TYPES
+    for place_type in place.types:
+      if place_type in PLACE_TYPES_TO_CHILD_PLACE_TYPES:
+        ordered_child_place_types = PLACE_TYPES_TO_CHILD_PLACE_TYPES[place_type]
+        break
+
+  # Construct constraint to filter by expected child types.
+  child_types_str = ",".join(ordered_child_place_types)
+  constraints = f"{{typeOf:[{child_types_str}]}}"
+
+  # Fetch child places matching the expected types for the given place.
+  raw_property_values_response = fetch.raw_property_values(
+      nodes=[place.dcid],
+      prop="containedInPlace+",
+      out=False,
+      constraints=constraints,
+      max_pages=None)
+
+  # Collect all types of child places found in the property values response.
+  child_place_types = set()
+  type_to_dcids = {}
+  for raw_property_value in raw_property_values_response.get(place.dcid, []):
+    dcid = raw_property_value.get("dcid")
+    if not dcid:
+      continue
+    types = raw_property_value.get("types", [])
+    child_place_types.update(types)
+    for t in types:
+      if t not in type_to_dcids:
+        type_to_dcids[t] = []
+      type_to_dcids[t].append(dcid)
+
+  child_place_type_candidates = []
+  # Return the child place types that match the ordered list of possible child types.
+  for place_type_candidate in ordered_child_place_types:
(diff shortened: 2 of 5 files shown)


server/lib/fetch.py
```diff
@@ -470,8 +470,16 @@ def raw_property_values(nodes, prop, out=True, constraints=''):
...
+  from server.lib.feature_flags import is_feature_enabled
+  from server.lib.feature_flags import USE_V2_API
+
+  if is_feature_enabled(USE_V2_API):
+    resp = dc.v2node_paginated(
+        nodes, '{}{}{}'.format('->' if out else '<-', prop, constraints),
+        max_pages)
+  else:
+    resp = dc.v2node(nodes, '{}{}{}'.format('->' if out else '<-', prop,
+                                            constraints))
```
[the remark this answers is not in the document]
nick-nlb: This is true, but temporary (and parallels the similar call in `property_values`). The second branch will eventually disappear.

server/routes/place/utils.py
```diff
@@ -631,17 +631,25 @@ def get_child_place_types(place: Place) -> list[str]:
...
+  child_types_str = ",".join(ordered_child_place_types)
+  constraints = f"{{typeOf:[{child_types_str}]}}"
+
+  # Fetch child places matching the expected types for the given place.
+  raw_property_values_response = fetch.raw_property_values(
+      nodes=[place.dcid],
+      prop="containedInPlace",
+      out=False,
+      constraints=constraints,
+      max_pages=None)
```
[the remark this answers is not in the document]
nick-nlb: I don't agree with the principle being stated here, that "highly likely" to contain all the relevant child place types is enough. I would argue that it might be highly likely that the first page will not contain all place types in a place that has so many pages that it overruns the pagination.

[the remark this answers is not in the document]
nick-nlb: As a follow-up to this: while the original assertion that the code-assist reviewer made was incorrect, there is an underlying inefficiency where we pull all the child places in order to get the child place types, collate those place types and then use them to once again pull all the child places. 

I've created a new function that returns both the places types (as the original function did) as well as the places, and added a TODO to consider converting other uses of the original function to this in order to get rid of what is a now very similar function pair with repeated code.

server/lib/fetch.py
```diff
@@ -453,7 +453,7 @@ def multiple_property_values(nodes: List[str],
   return result
 
 
-def raw_property_values(nodes, prop, out=True, constraints=''):
+def raw_property_values(nodes, prop, out=True, constraints='', max_pages=1):
   """Returns full property values data out of REST API response.
```
juliawu: supernit: Can we update this text from saying "full" property values to something that makes the behavior of the new `max_pages` parameter more clear?

server/lib/fetch.py
```diff
@@ -470,8 +470,16 @@ def raw_property_values(nodes, prop, out=True, constraints=''):
   }
 
   """
-  resp = dc.v2node(nodes, '{}{}{}'.format('->' if out else '<-', prop,
-                                          constraints))
+  from server.lib.feature_flags import is_feature_enabled
+  from server.lib.feature_flags import USE_V2_API
```
juliawu, line 474: What do you think of leaving imports at the top of the file instead?

server/lib/fetch.py
```diff
@@ -453,7 +453,7 @@ def multiple_property_values(nodes: List[str],
   return result
 
 
-def raw_property_values(nodes, prop, out=True, constraints=''):
+def raw_property_values(nodes, prop, out=True, constraints='', max_pages=1):
```
juliawu, line 456: When using `raw_property_values(nodes=, prop=)` without specifying `max_pages`, it's a little surprising to have the default at 1 and not `None`. Because `max_pages` is only used in the USE_V2_API=true branch, what do you think of making `max_pages` default to `None` (so we're removing pagination bugs moving forward)? We can set max_pages=1 explicitly in locations where latency is a problem.

server/tests/routes/place/related_places_test.py
```diff
@@ -217,8 +216,8 @@ def test_related_places_california_es(self):
       expected = jsonify(
           RelatedPlacesApiResponse(
               place=Place(dcid=mock_data.CALIFORNIA.dcid,
-                          name=mock_data.CALIFORNIA.dcid,
-                          types=[]),
+                          name=mock_data.CALIFORNIA.name + 'es',
```
juliawu, line 219: Why is `+ 'es'` needed here?

juliawu (commented):
Very nice!

server/lib/fetch.py
nick-nlb: Normally I would (and I remember dealing with this when inserting the original imports in the property_values version of this. Unfortunately it causes circular imports which would require significant refactoring to resolve.

Because of the temporary nature of the import and flag usage, I imported inside the function.

server/lib/fetch.py
nick-nlb: This one (and the property_values version) was done this way to match the original, preexisting pattern from `fetch_triples`.

The idea of defaulting to forcing a full resolve of the pagination did cross my mind, but weighing the counter use of this in the original pattern, as well as the fact that although full resolution would better match what non-Spanner paths do (because there is no pagination), it is otherwise an unusual approach to pagination (where you would normally expect a single page). 

The potential of introducing undiagnosed latency was also a consideration (i.e., it can be beneficial to surface these issues explicitly, so we can determine if the added Spanner latency is acceptable.

Let's discuss offline!

server/tests/routes/place/related_places_test.py
nick-nlb: The function that makes the mock data creates mock Spanish data by appending "es@es" to the name (i.e., "Californiaes@es"). I believe the original logic for this is that many names, like California, are the same in Spanish, so the mock now differentiates it. 

The `@es` is then stripped, but leaving us with the constructed  mocked translation name with "es" at the end.

server/lib/fetch.py
nick-nlb: Done! The "full" property values was actually the original text describing the general purpose of the function from when this function was created, but it wasn't very clear or descriptive (and I also added a note about the max pages).

server/tests/routes/place/related_places_test.py
juliawu: Gotcha, thanks for explaining! The `es@es` with `es` on both sides looks a little strange, but nothing worth blocking the PR over.

server/lib/fetch.py
juliawu: Gotcha. Leaving it here works for me then.

server/lib/fetch.py
juliawu: Per discussion, we'll keep `max_pages=1`, as this would be the ideal behavior when we swap to Spanner and all calls include pagination.

juliawu (approved):
Thanks for the responses! LGTM.

PR #6334, opened 2026-06-03
Synchronize the uv.lock file with the current requirements in the nl_server

nick-nlb:
## Description

This commit synchronizes the uv.lock file with the latest requirements.

Separately from this PR, a bug has been created to make this done automatically in the fugure.


juliawu (approved):
THANK YOU

PR #6346, opened 2026-06-08
[v0/v1 migration] Targeted translations

nick-nlb:
## Issue

[b/515782168](b/515782168)

## Related PR

[Mixer: 1915](https://github.com/datacommonsorg/mixer/pull/1915)

## Description

Until the above PR, Mixer was unable to support the ability filter by terminal nodes in `v2/node` queries, and was thus unable to fetch translations for particular places by language. We could only over-fetch and request all translations for the places sent into the query, and then do the filtering in Flask.

While this was inefficient, it did work when Mixer was backed by BT. However, when using Spanner, the overfetching frequently caused page overruns, and as a consequence, the response was truncated to the first page. This resulted in many pages and charts being only partially translated.

With the linked PR in Mixer, we are now able to filter terminal nodes by the language query operator `$lang`.

In website, we are now able to request translations for specific languages, resulting in much smaller responses from Mixer that rarely paginate (and when they do, paginate to a much lesser degree, resulting in less latency cost).

## Testing

The following two pages demonstrate the issue:

*   `http://localhost:8080/place?hl=ru` (Russian locale place page).
*   `http://localhost:8080/place/country/JPN?q=Japan&hl=ja` (Japanese locale place page).

When using a Spanner-enabled backend:
* In master, these two pages will result in partial translations. On this branch, both will be fully translated.

When using BT:
* The translations will be complete on both master and this version.


server/lib/fetch.py
```diff
@@ -448,8 +448,12 @@ def multiple_property_values(nodes: List[str],
     resp_node_arcs = resp_data.get(node, {}).get('arcs', {})
     result[node] = {}
     for prop in props:
-      prop_nodes = resp_node_arcs.get(prop, {}).get('nodes', [])
-      result[node][prop] = [p.get('dcid') or p.get('value') for p in prop_nodes]
+      # Strip inline constraints (e.g., '{...}') to get the response key return by mixer
+      base_prop = prop.split('{')[0].strip()
```
juliawu, line 452: Do we need to check if `prop.split('{')` returns a non-empty array here? Will indexing `[0]` raise and error if there's no `{}` in prop?

server/routes/place/utils.py
```diff
@@ -500,9 +500,13 @@ def fetch_places(place_dcids: List[str], locale=DEFAULT_LOCALE) -> List[Place]:
       List[Place]: A list of Place objects with names in the specified locale.
   """
   props = ['typeOf', 'name', 'dissolutionDate']
-  # Only fetch names with locale-specific tags if the desired locale is non-english
   if locale != DEFAULT_LOCALE:
-    props.append('nameWithLanguage')
+    from server.lib import i18n
```
juliawu: Is this another circular import issue, or can we move this import to the top of the page? I see we're already importing other sections of `server.lib.i18n`.

juliawu (commented):
Very nice! Just a few comments.

server/lib/fetch.py
nick-nlb: Because we're doing a string split on prop, even if there is no "{", it will split into an array of one string (the original string) and index [0] to retrieve it (this would still work even if prop is an empty string).

So we are okay here!

server/routes/place/utils.py
nick-nlb: Good call - no circular import here. Done!

juliawu (approved):
Thanks for the update!

PR #6350, opened 2026-06-09
Remove draw functions from datacommons JS library

juliawu:
Removes the `draw*` functions from the datacommons JS library.

The `draw*` functions are no longer used in the datacommons JS library (which instead uses the tiles as `Datacommons<type>Component`). This PR removes the following functions and their references:
* drawRanking
* drawBar
* drawLine
* drawMap


dwnoble (approved):
thanks! is the plan to remove js/ranking/old_component in a follow-up?

juliawu:
> thanks! is the plan to remove js/ranking/old_component in a follow-up?

Thanks for the review! Yup, we're going to remove old_component in the followup where I clear out the old ranking pages entirely.

PR #6351, opened 2026-06-10
[v0/v1 migration] Fix bar chart facet selector test

nick-nlb:
## Issue

[b/515840750](b/515840750)

## Description

The facet selection web driver test relied on the order of the facets as displayed in the facet selector (and therefore as given by Mixer) in order to choose what facet was selected.

The order of the facets coming out of Spanner is different than BT, causing the selected index to no longer be the expected facet and thus causing the test to fail.

This PR updates the test to choose the selected facet by looking at the description of the facet in the selector.

Additionally, it makes other tweaks to how the test searches for and waits for content in order to tighten the test (this test was mostly reliable, but did occasionally fail on heavy flake-finding; it now always passes).

## Test

```
export WEBSITE_MIXER_API_ROOT=http://127.0.0.1:8081
./run_test.sh -w --live --no_extract -k "test_bar_select_different_facet" --reruns 0
```


juliawu (approved):
Thanks for the fix!

PR #6353, opened 2026-06-10
Temporarily disable ranking pages with HTTP 503

juliawu:
Temporarily disables ranking pages by returning HTTP 503 in the default case.

In order to proceed with the migration off of v1 mixer APIs, we are temporarily turning down the ranking pages with an HTTP 503 response while we get v2 API and Spanner ready versions of the page ready. For more context, see [this doc](https://docs.google.com/document/d/1rn9Lw5Z8dp64Qv38RYxhXAVem3mYcQ6EOvBfSfaRSfI/edit?tab=t.0#heading=h.gyh1kbjcale2) (Google corp access required.)

Now, when navigating to /ranking/*, we first check whether the `new_ranking_page` feature flag is enabled.
-> If `new_ranking_page` is enabled, show the new v2/observation-backed tile-based ranking page.
-> If `new_ranking_page` is NOT enabled, return a 503.

The new 503 page looks like this:
image

Changes made:
* Adds a return HTTP 503 to the /ranking/* route
* Skips the ranking page webdriver tests
* Removes links to the ranking pages from the place landing pages
* Updates the place landing page webdriver tests to skip the check for ranking page links.

Testing strategy:
* Spin up a local server
* Make sure `new_ranking_page` feature flag is set to `false` in `local.json`
* Navigate to a ranking page like http://localhost:8083/ranking/Count_Person/County and see that a HTTP 503 is returned.
* Navigate to http://localhost:8083/place and see that the page still renders but the ranking pages have been removed.


server/routes/ranking/html.py
```diff
@@ -27,38 +27,43 @@
...
-                               use_new_ranking=use_new_ranking,
-                               sample_questions=json.dumps(
-                                   current_app.config.get(
-                                       'HOMEPAGE_SAMPLE_QUESTIONS', [])))
+  # If new ranking pages are enabled, return the new ranking template.
+  if use_new_ranking:
+    place_name = ''
+    if place_dcid:
+      place_names = place_api.get_i18n_name([place_dcid])
+      place_name = place_names[place_dcid]
```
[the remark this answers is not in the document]
juliawu: Fixed.

server/webdriver/tests/ranking_test.py
```diff
@@ -15,6 +15,7 @@
 from server.webdriver.base_dc_webdriver import BaseDcWebdriverTest
 from server.webdriver.shared_tests.ranking_test import RankingTestMixin
 
-
+# TODO(juliawu): Remove this skip once the ranking pages are fully migrated to v2.
+@pytest.mark.skip(reason="Temporarily disabling ranking tests while completing the v2 migration for the ranking pages")
```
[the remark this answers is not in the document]
juliawu: Fixed.

server/webdriver/cdc_tests/ranking_test.py
```diff
@@ -14,6 +14,7 @@
 from server.webdriver.cdc_tests.cdc_base_webdriver import CdcTestBase
 from server.webdriver.shared_tests.ranking_test import RankingTestMixin
 
-
+# TODO(juliawu): Remove this skip once the ranking pages are fully migrated to v2.
+@pytest.mark.skip(reason="Temporarily disabling ranking tests while completing the v2 migration for the ranking pages")
```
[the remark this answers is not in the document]
juliawu: Fixed.

server/templates/place_landing.html
```diff
@@ -80,6 +80,8 @@ <h3>{% trans %}Places{% endtrans %}</h3>
     </li>
   </ul>
 
+  {# TODO(juliawu): Add back the rankings once the v2 migrated, new ranking pages are completed. #}
+  {% if false %}
```
nick-nlb: Since we are now tying the existence of the ranking pages completely to the `new_ranking_page` (i.e., when that flag is on, ranking pages exist one way or another), could we convert this check to use the flag rather than the hard false?

server/routes/ranking/html.py
```diff
@@ -27,38 +27,41 @@
...
-                               use_new_ranking=use_new_ranking,
-                               sample_questions=json.dumps(
-                                   current_app.config.get(
-                                       'HOMEPAGE_SAMPLE_QUESTIONS', [])))
+  # If new ranking pages are enabled, return the new ranking template.
+  if use_new_ranking:
+    place_name = ''
+    if place_dcid:
+      place_names = place_api.get_i18n_name([place_dcid])
+      place_name = place_names.get(place_dcid, place_dcid)
```
nick-nlb: The handling of the fallback is slightly different here than in the previous code (and might break in edge cases).

The first line is the same, getting place_names. The get_i18n_name function initializes all the dcid keys to empty strings, so place_names might have an empty string for the place_dcid. 

The get function will see the place_dcid index value as `''`' and return it (rather than use the fallback place_dcid), whereas the old code checks for that.

server/webdriver/shared_tests/place_landing_test.py
```diff
@@ -45,16 +45,18 @@ def test_place_landing_en(self):
...
-        By.XPATH, '//*[@id="body"]/ul[2]/li[1]/strong')
-    self.assertEqual(median_income.text, 'Median Income, United States')
-
-    gni = self.driver.find_element(By.XPATH, '//*[@id="body"]/ul[2]/li[4]/a[2]')
-    self.assertEqual(gni.text, 'Gross National Income')
-    self.assertEqual(
-        gni.get_attribute('href'), self.url_ +
-        '/ranking/Amount_EconomicActivity_GrossNationalIncome_PurchasingPowerParity_PerCapita/Country/'
-    )
+    # TODO(juliawu): Uncomment the block below when we complete the v2-compatible ranking pages
```
nick-nlb, line 48: My first thought was that this could also be gated behind the flag (following up from the comment to flag rather than `if false` the template section), but I checked and couldn't find any examples of us ever using feature flags directly like this in the tests. We sometimes force flags on or off, but we never seem to check the status of a flag in the environment the test is running. 

So it might be safer to keep it this, pending understanding of why that is the case.

nick-nlb (commented):
Thank you for this! Hidden in this PR is a monumental moment. A major milestone in our long goodbye to v1!

Should we also set the ranking page flag to false in all environments? (definitely in staging / dev, but I think local should match that as well.

server/routes/ranking/html.py
juliawu: Great callout, I've reverted back to the original, but used `.get(place_dcid, '')` to address the original Gemini comment.

server/templates/place_landing.html
juliawu: Good idea, I've now passed in the `new_ranking_page` flag and am using that in this if statement.

server/webdriver/shared_tests/place_landing_test.py
juliawu: Being able to use the feature flags directly in the tests would be great! In order not to block this PR, I vote we leave implementing this to a future PR.

juliawu:
> Should we also set the ranking page flag to false in all environments? (definitely in staging / dev, but I think local should match that as well.

Yes! I'll do that in a separate PR after this one goes in.

nick-nlb (approved):
Thank you (and goodbye v1)!

PR #6355, opened 2026-06-10
[v0/v1 migration] Remove `v1` path and make `use_v2_api` flag obsolete.

nick-nlb:
## Issue

[b/522384579](b/522384579)

## Description

The `use_v2_api` flag has been set to true for some time in all environments except for custom DC.

With the upcoming turndowns, all environments, including custom DCs, will need to use `v2`. This PR drops the flag usage in the code itself, removing all code associated with the `v1` endpoints, and leaving all environments using `v2`.

## Notes

This PR does not yet remove the flags themselves. This will be done in a subsequent PR.

## Testing

For all environments where the flag had previously been set to true, this should be a noop. 

Areas that are potentially affected are be included are:
* The stat var explorer
* The visualization tools (with the embedded stat var hierarchy)
* The NL search (i.e. a search for "Jobs in Texas")


static/js/stat_var_hierarchy/stat_var_section_input.tsx
```diff
@@ -25,59 +25,15 @@ import React from "react";
 
 import { ASYNC_ELEMENT_CLASS } from "../constants/css_constants";
 import { Context, ContextType } from "../shared/context";
-import { isFeatureEnabled } from "../shared/feature_flags/util";
 import {
   RADIO_BUTTON_TYPES,
   StatVarHierarchyType,
```
juliawu: Is this import still needed?

server/routes/shared_api/stats.py
```diff
@@ -182,6 +182,4 @@ def search_statvar():
       }
     else:
       result = {"matches": [], "statVars": [], "statVarGroups": []}
-  else:
-    result = dc.search_statvar(query, entities, sv_only)
```
juliawu, line 186: This looks like the only usagae of `dc.search_statvar`. Let's also remove it from datacommons.py

juliawu (commented):
Thank you, this is a huge milestone! Left some minor comments.

I've also encountered errors when playing around with the place pages (try http://localhost:8080/place/geoId/48215?category=Crime), but not sure if it's related or not.

Screenshot 2026-06-10 at 4 28 34 PM

static/js/stat_var_hierarchy/stat_var_section_input.tsx
nick-nlb: Removed!

nick-nlb:
> Thank you, this is a huge milestone! Left some minor comments.
> 
> I've also encountered errors when playing around with the place pages (try http://localhost:8080/place/geoId/48215?category=Crime), but not sure if it's related or not.
> 

This bug occurs in master and so is not related (but I will investigate separately).

juliawu (approved):
Thank you!

PR #6363, opened 2026-06-11
Remove dependency on api/ranking in sitemap script

juliawu:
This PR updates the logic for generating the sitemap entries for the top 100 cities in the USA to V2 Python API queries instead of a live request to /api/ranking.

/api/ranking will stop working soon with the deprecation of v1/page/ranking, which is why we need this change. The V2 Python API call can take a while (~15 seconds on my machine), but because the sitemaps are not updated frequently and the process is already quite manual, a 15 second call is acceptable.


tools/sitemap/README.md
```diff
@@ -46,6 +46,43 @@ WHERE
...
+Similarly, the query necessary to get the top 100 most populous US cities requires some heavy lifting, so we query for those DCIDs in Spanner Studio. The result of the query has been saved in `top_100_us_cities.csv`. To update the CSV,
+use the following SQL query in Spanner Studio and export the results as CSV:
+
+```sql
+WITH usa_cities AS (
+    SELECT subject_id AS city_dcid
+    FROM Node
+    WHERE 'City' IN UNNEST(types)
+      AND subject_id LIKE 'geoId/%'
+),
```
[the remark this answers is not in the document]
juliawu: Switched to V2 Python API instead.

tools/sitemap/top_100_us_cities.csv
```diff
@@ -0,0 +1,101 @@
...
+"geoId/2622000","Detroit","1849568"
+"geoId/0455000","Phoenix","1708127"
+"geoId/4260000","Philadelphia","1603797"
+"geoId/4865000","San Antonio","1567118"
+"geoId/0666000","San Diego","1425999"
+"geoId/4819000","Dallas","1345076"
+"geoId/12031","Duval County","1062963"
+"geoId/0668000","San Jose","1037082"
+"geoId/15003","Honolulu County","1016508"
+"geoId/1235000","Jacksonville","1009833"
```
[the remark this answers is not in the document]
juliawu: Switched to V2 Python API instead.

tools/sitemap/main.py
```diff
@@ -142,14 +144,11 @@ def get_global_cities_with_population_over_500k() -> List[str]:
...
   dcids = []
-  response = requests.get(
-      "https://datacommons.org/api/ranking/Count_Person/City/country/USA")
-  city_ranking_data = response.json().get("Count_Person",
-                                          {}).get("rankTop1000",
-                                                  {}).get("info", [])
-  for city in city_ranking_data[:100]:
-    if 'placeDcid' in city:
-      dcids.append(city['placeDcid'])
+  with open(TOP_100_US_CITIES_CSV) as f:
```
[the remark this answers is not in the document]
juliawu: Switched to V2 Python API instead

tools/sitemap/README.md
```diff
@@ -46,6 +46,43 @@ WHERE
...
+city_populations AS (
+    SELECT 
+        c.city_dcid,
+        MAX(CAST(v.value AS FLOAT64)) AS max_population
+    FROM usa_cities c
+    JOIN Observation o ON c.city_dcid = o.observation_about
+    CROSS JOIN UNNEST(o.observations.values) AS v
+    WHERE o.variable_measured = 'Count_Person'
+    GROUP BY c.city_dcid
+),
```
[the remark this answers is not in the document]
juliawu: Switched to V2 Python API instead

tools/sitemap/main.py
```diff
@@ -139,18 +139,36 @@ def get_global_cities_with_population_over_500k() -> List[str]:
...
+    return []
+
+  # Group and deduplicate city DCIDs, keeping the observation with the most recent date
+  city_populations = {}  # maps city_dcid -> (population_value, date_string)
+  for record in records:
+    if record.entity and record.value is not None and record.date:
+      current_value, current_date = city_populations.get(
+          record.entity, (None, ""))
+      # Keep the record with the more recent date. If dates match, select the higher value.
+      if (current_date == "") or (record.date > current_date) or (
```
nick-nlb: nit: the current_date == "" is redundant here, because any record.date will evaluate as greater than "" (and so pass anyway).  We could remove this check.

tools/sitemap/main.py
```diff
@@ -139,18 +139,36 @@ def get_global_cities_with_population_over_500k() -> List[str]:
...
+  except Exception:
+    logging.exception('Got an error while querying for US cities population')
+    return []
+
+  # Group and deduplicate city DCIDs, keeping the observation with the most recent date
+  city_populations = {}  # maps city_dcid -> (population_value, date_string)
+  for record in records:
+    if record.entity and record.value is not None and record.date:
+      current_value, current_date = city_populations.get(
+          record.entity, (None, ""))
```
nick-nlb: It looks like the `None` assignment (in conjunction with the "" assignment for the date) work in tandem to prevent a `None` from ever being used in the boolean below (and giving a type error).

I.e., we are initializing the current_value to None, but the none is never used in the boolean because the corresponding "" cause the expression to short-circuit out before that.

This works, but relies on the order of the check to work. We could make this a bit sturdier by switching the None to `-1`, so that it is the correct type from the beginning. 

This is a nit given the obscurity of the script (and the unlikeliness of it changing in the future).

tools/sitemap/main.py
```diff
@@ -139,18 +139,36 @@ def get_global_cities_with_population_over_500k() -> List[str]:
...
+  try:
+    # Query for cities in USA, filtering on population (Count_Person)
+    response = dc_client.observation.fetch_observations_by_entity_type(
+        date="latest",
+        parent_entity="country/USA",
+        entity_type="City",
+        variable_dcids=["Count_Person"])
+    records = response.to_observation_records()
+  except Exception:
+    logging.exception('Got an error while querying for US cities population')
```
nick-nlb, line 153: This is what we're doing for other similar sections of this file, but I wonder if this shouldn't fail harder (i.e. end the script). 

If one of these calls produces no results, it's better that no changes to the site map could ever happen (given that we will always expect results for each of these).

nick-nlb (commented):
Thank you for doing this! The final final removal of v1.

A couple of small comments inline, and a question:

When I run the script and look at the diff, some of the changes made sense, but I also noticed the addition of counties that weren't there before. It makes sense that these are being pulled, because they are classified as cities in the Knowledge Graph (but are really represented by actual cities).

https://datacommons.org/browser/geoId/12031 (Duvall County, a city in the graph, but the city as we think of it is Jacksonville).

One heuristic we could do (so we don't have to make calls to get the actual types) would be to filter out counties by the number of digits (5) in the FIPS code (this link below explains that counties always have five).

https://transition.fcc.gov/oet/info/maps/census/fips/fips.txt

tools/sitemap/main.py
juliawu: Great callout. Check removed.

tools/sitemap/main.py
juliawu: Good idea, switched to `-1`.

tools/sitemap/main.py
juliawu: I agree with this in principle, but I think we would want a cleanup of the whole sitemap script, which I think is out of scope for this PR

juliawu:
> When I run the script and look at the diff, some of the changes made sense, but I also noticed the addition of counties that weren't there before. It makes sense that these are being pulled, because they are classified as cities in the Knowledge Graph (but are really represented by actual cities).
> 
> https://datacommons.org/browser/geoId/12031 (Duvall County, a city in the graph, but the city as we think of it is Jacksonville).
> 
> One heuristic we could do (so we don't have to make calls to get the actual types) would be to filter out counties by the number of digits (5) in the FIPS code (this link below explains that counties always have five).
> 
> https://transition.fcc.gov/oet/info/maps/census/fips/fips.txt

I added a filtering to skip any DCIDs that follow the "geoId/" + 5 digit FIPS code pattern

nick-nlb (approved):
Thank you Julia!

PR #6369, opened 2026-06-12
Update metadata fetches to use provenance id

nick-nlb:
## Issue

[b/519184012](b/519184012)

## Related PR

[Mixer: 1937](https://github.com/datacommonsorg/mixer/pull/1937)

## Description

The related Mixer PR updates facet definitions so that the provenance id is included. While the provenance id has typically been the import name prefixed with `dc/base/`, this will soon no longer always be the case.

This PR updates the metadata endpoints to use the provenance id directly, only falling back to the prefixed import name if the provenance id is not available.

## Notes

For this PR to have effect, you must be using a version of mixer that contains the updates in the linked PR. 

🚨 This PR should not be merged in until the corresponding Mixer PR is merged in.

### DIFF
--- a/server/routes/shared_api/metadata.py
+++ b/server/routes/shared_api/metadata.py
@@ -84,14 +84,15 @@ def _get_node_name(node_list: list[dict[str, Any]],
   return None
 
 
-def _get_provenance_name(pdata: dict[str, Any], linked_names_map: dict[str,
-                                                                       str],
+def _get_provenance_name(provenance_data: dict[str, Any],
+                         linked_names_map: dict[str, str],
                          import_name: str | None, prov_id: str | None) -> str:
   """Resolves the display name for a provenance node."""
-  if not pdata:
+  if not provenance_data:
     return import_name or (prov_id.split('/')[-1] if prov_id else "")
-  return (_get_node_name(pdata.get('isPartOf', []), linked_names_map) or
-          _get_node_name(pdata.get('name', []), linked_names_map) or
+  return (_get_node_name(provenance_data.get('isPartOf', []),
+                         linked_names_map) or
+          _get_node_name(provenance_data.get('name', []), linked_names_map) or
           import_name or (prov_id.split('/')[-1] if prov_id else ""))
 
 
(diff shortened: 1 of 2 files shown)


server/routes/shared_api/metadata.py
```diff
@@ -84,6 +84,17 @@ def _get_node_name(node_list: list[dict[str, Any]],
   return None
 
 
+def _get_provenance_name(pdata: dict[str, Any], linked_names_map: dict[str,
```
juliawu: Supernit: Can we use a more descriptive name than `pdata` or otherwise define it in the docstring? Took me a second to understand the "p" refers to "provenance".

server/routes/shared_api/metadata.py
```diff
@@ -335,11 +346,13 @@ def _build_metadata_payload(
 
     for fid in active_facets:
       finfo = facets.get(fid, {})
```
juliawu: Similarly, while we're making edits, could we use `facet_id` and `facet_info` instead?

server/routes/shared_api/metadata.py
```diff
@@ -335,11 +346,13 @@ def _build_metadata_payload(
...
-      import_name = finfo.get('importName')
-      if not import_name:
-        continue
+      prov_id = finfo.get('provenanceId')
+      if not prov_id:
+        import_name = finfo.get('importName')
+        if not import_name:
+          continue
+        prov_id = f"dc/base/{import_name}"
```
juliawu, line 356: For my own understanding, why do we need the dc/base prefix here and on line 496?

server/routes/shared_api/metadata.py
```diff
@@ -349,8 +362,8 @@ def _build_metadata_payload(
       mm = finfo.get('measurementMethod')
 
       source_name = _get_node_name(pdata['source'], linked_names_map)
-      prov_name = _get_node_name(pdata['isPartOf'], linked_names_map) or \
-                  _get_node_name(pdata['name'], linked_names_map) or import_name
+      prov_name = _get_provenance_name(pdata, linked_names_map,
+                                       finfo.get('importName'), prov_id)
```
juliawu: can we use the variable `import_name` instead of refetching `finfo.get('importName')` here?

juliawu (approved):
Thank you! Just some minor comments, none of which should block this PR.

server/routes/shared_api/metadata.py
nick-nlb: We discussed offline but to record here:

If we are using either Spanner current schema or Spanner new schema, we will expect the observation call to have returned a full provenance id (that is either prefixed or not, depending on whether it is a custom dc or base dc). 

In BT, we don't get a provenance id directly in the endpoint, and likely won't due to the fact that we are now approaching the end of its lifespan. We know that everything coming out of BT should be prefixed with the legacy `dc/base` prefix (because provenance ids without prefixes will never be applied to data in a BT environment). Mixer therefore returns the facet info with the provenance id, with that hardcoded prefix.

We then have this final fallback on website, where if we don't have a provenance id, we know that we are in a legacy environment (where we always had bare import names and no provenance ids, but knew that provenance ids are just import names with the prefix). So we add the prefix here to cover this possibility, and have website behave the same way it did before.

PR #6387, opened 2026-06-17
feat: Add Croissant JSON-LD metadata to provenance pages

dtulsyann:
This feature dynamically generates and injects Croissant dataset metadata (JSON-LD format) into the Knowledge Graph Browser for provenance nodes (dc/base/*).

- Queries the Mixer V2 Node API to retrieve dataset properties like name, description, license, and creator source.
- Injects the JSON-LD script into the node.html template for SEO and metadata crawlers.

### DIFF
--- a/server/lib/croissant_metadata.py
+++ b/server/lib/croissant_metadata.py
@@ -1,0 +1,125 @@
+# Copyright 2024 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import logging
+import server.services.datacommons as dc
+
+def build_dataset_metadata(dcid: str) -> dict:
+  """Builds a Croissant JSON-LD dictionary for a provenance node."""
+  json_ld_data = {
+      "@context": {
+          "@language": "en",
+          "@vocab": "https://schema.org/",
+          "sc": "https://schema.org/",
+          "cr": "http://mlcommons.org/croissant/",
+          "rai": "http://mlcommons.org/croissant/RAI/",
+          "dct": "http://purl.org/dc/terms/",
+          "citeAs": "cr:citeAs",
+          "column": "cr:column",
+          "conformsTo": "dct:conformsTo",
+          "data": {
+              "@id": "cr:data",
+              "@type": "@json"
+          },
+          "dataType": {
+              "@id": "cr:dataType",
+              "@type": "@vocab"
+          },
+          "examples": {
+              "@id": "cr:examples",
+              "@type": "@json"
+          },
+          "extract": "cr:extract",
+          "field": "cr:field",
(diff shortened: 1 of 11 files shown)


server/lib/croissant_metadata.py
```diff
@@ -0,0 +1,125 @@
...
+      # Fetch its external URL
+      if s_dcid:
+        try:
+          s_resp = dc.v2node([s_dcid], "->url")
+          s_url_nodes = s_resp.get("data", {}).get(s_dcid, {}).get(
+              "arcs", {}).get("url", {}).get("nodes", [])
+          if s_url_nodes:
+            source_obj["url"] = s_url_nodes[0].get("value")
+        except Exception as e:
+          logging.error("Error fetching source URL for %s: %s", s_dcid, e)
```
SandeepTuniki: Either catch a more specific sub-class of Exception, or catch the Exception and then re-raise it after logging the error.

server/lib/croissant_metadata.py
```diff
@@ -0,0 +1,125 @@
...
+
+      json_ld_data["creator"] = [source_obj]
+
+  except Exception as e:
+    logging.error("Error fetching metadata for %s: %s", dcid, e)
```
SandeepTuniki, line 165: Same comment as the previous one about catch the root Exception class.

server/routes/browser/html.py
```diff
@@ -47,7 +49,15 @@ def browser_node(dcid):
       node_name = api_name
   except Exception as e:
     logging.info(e)
+
+  json_ld_data = {}
+  # Provenance nodes start with dc/base/
+  if dcid.startswith("dc/base/"):
```
SandeepTuniki: If the intent is to check when the user is requesting the provenance page (ex: https://datacommons.org/browser/dc/base/ACSED5YrSurvey), I think there should be a better way to do it than checking for the "dc/base/" substring. This will break for DCP users, as the website is also moving away from this prefix pattern.

SandeepTuniki: Perhaps this folder can give a clue on what's the better way to check the page: https://github.com/datacommonsorg/website/tree/master/server/routes/browser

server/templates/browser/node.html
```diff
@@ -21,6 +21,11 @@
    {% set title = node_name + ' - Knowledge Graph' %}
 
    {% block head %}
+    {% if json_ld_data %}
+    <script type="application/ld+json">
+      {{ json_ld_data | tojson | safe }}
```
SandeepTuniki: What does the `safe` filter (or pipe?) do here?

SandeepTuniki (changes requested):

server/lib/croissant_metadata.py
dtulsyann: Added the raise keyword to re-throw the exception after logging it.

server/lib/croissant_metadata.py
dtulsyann: Added the raise keyword to re-throw the exception after logging it.

server/routes/browser/html.py
dtulsyann: Replaced the hardcoded string prefix check with a robust API call (fetch.property_values([dcid], 'typeOf')).

server/templates/browser/node.html
dtulsyann: I actually removed the `| safe` filter entirely! Flask's `| tojson` filter automatically handles HTML-escaping securely and wraps the final output in a Markup object, which makes adding | safe completely redundant here.

SandeepTuniki:
Closing this PR as the changes got merged through another PR.

PR #6388, opened 2026-06-17
Enable use_v2_api feature flag on custom DC

juliawu:
Updates the `use_v2_api` to default to `true` on custom DC


nick-nlb (approved):

nick-nlb:
With the `use_v2_api` flag now being removed, this PR is closed.

PR #6394, opened 2026-06-18
Disable NL server startup when RESOLVE_WITH_SPANNER_EMBEDDINGS is true

clincoln8:
Condition NL server startup and Mixer model arguments on RESOLVE_WITH_SPANNER_EMBEDDINGS not being true in cdc_services/run.sh.

This is to reduce start up time for cdc-services cloud run job.


shixiao-coder (approved):
LGTM

PR #6396, opened 2026-06-18
Remove dynamic enabling now obsolete `use_v2_api` feature flag.

nick-nlb:
## Goal

Remove dynamic enabling of the obsolete `use_v2_api` feature flag from the CDC services startup script, aligning with the deprecation and removal of `v1` API paths.

## Description

The `use_v2_api` flag has already been removed from usage within website (frontend/Flask). This PR follows that with a removal of the final reference to that flag in the CDC services startup script. We will follow this with a removal of the flags themselves.

PR #6404, opened 2026-06-22
[v0/v1 migration] Remove use_v2_api feature flag

nick-nlb:
## Related PRs
[6355](https://github.com/datacommonsorg/website/pull/6355)
[6396](https://github.com/datacommonsorg/website/pull/6396)

## Description

Remove `use_v2_api` feature flag from all environment configurations.

This is a follow-up to 6355 and 6396 (linked above) that removed all usage of those flags from the codebase.


gmechali:
Hey Nick! The rule is never submit a change that includes production.json AND another file. Can you remove the change to production.json, submitt his PR with all the other files, then put up a separate PR just for production.json?

nick-nlb:
> Hey Nick! The rule is never submit a change that includes production.json AND another file. Can you remove the change to production.json, submitt his PR with all the other files, then put up a separate PR just for production.json?

Updated! Thank you for the sugggestion. 

In this case the separation has an additional benefit that we can do this merge now (we need to wait until production actually has 6355 before merging a change the prod flag).

nick-nlb:
This is on hold pending a separate removal of the production version of this flag (after the next production release).

juliawu (approved):

PR #6411, opened 2026-06-25
Download tool

pablonoel:
This PR address the Download Tool refactor, it includes fixing the Facet selector and streamline the interface.

- Complete download page.tsx refactor
- Move state elements into contex.tsx ( as per maps / scatter tools use )
- Add Theme / App wrapper
- Remove Date
- Fix Facet selector
- Refactor URL/Path handling
- Remove Preview button, implement auto-preview 

localhost_8080_tools_download(Laptop XS) (2)

localhost_8080_tools_download(Laptop XS) (1)

localhost_8080_tools_download(Laptop XS)

Things to consider:

- Download tool uses its own EnclosedPlacesSelector and FacetSelector, we should consider that to be shared among the rest of the tools. 
- There might be space for further refactor of the loadStateFromURL and some other elements
- There is no current change on the html/styles structure, that will be implemented separately


pablonoel:
@nick-nlb ready for review, let me know if you need more context.

static/js/i18n/i18n_tool_messages.ts
```diff
@@ -17,6 +17,29 @@
 import { defineMessages } from "react-intl";
 
 export const toolMessages = defineMessages({
+  downloadToolTitle: {
+    id: "download_visualization_tool_name",
```
nick-nlb, line 21: These are defined here, but never referenced (the text here is also still that of the map explorer)

static/js/tools/download/context.ts
```diff
@@ -0,0 +1,130 @@
...
+  const [facetError, setFacetError] = useState(false);
+  const facetsReqObj = useRef({});
+
+  return {
+    options: {
+      value: selectedOptions,
+      set: setSelectedOptions,
+    },
+    preview: {
+      options: previewOptions,
```
nick-nlb: These are now no longer referenced anywhere?

static/js/tools/download/page.tsx
```diff
@@ -428,60 +209,64 @@ export function Page(props: PagePropType): ReactElement {
...
+                    onModalOpen={onModalOpen}
+                  />
+                </div>
               </div>
             )}
             <Row className="d-lg-none">
               <Col>
                 <Button color="primary" onClick={toggleSvModalCallback}>
-                  {intl.formatMessage(toolMessages.selectAVariableInstruction)}
+                  {intl.formatMessage(
```
nick-nlb: In this (and a couple of other places) the prettier rules need to be applied.

static/js/tools/download/page.tsx
```diff
@@ -194,28 +108,28 @@ export function Page(props: PagePropType): ReactElement {
...
       .catch(() => {
-        setFacetError(true);
+        facets.setError(true);
       })
       .finally(() => {
-        setFacetLoading(false);
+        facets.setLoading(false);
       });
-  }, [selectedOptions, dataCommonsClient]);
+  }, [options.value]);
```
nick-nlb: This looks like the cause of the stale state at the base of the facet selection button. Without facets being in this dependency array, we don't update the facet. 

We could add facets directly to this, and solve the issue, but this ends up being a fairly inefficient effect (triggering multiple times per update - i.e. triggering when the facets change, but [correctly] dropping out of the effect early because of the condition at the top).

We could  solve this by destructuring the stable references we need out of the facets and only use those in the hook.

static/js/tools/download/page.tsx
```diff
@@ -511,147 +296,122 @@ export function Page(props: PagePropType): ReactElement {
...
-        statVarsList.forEach((sv) => (emptySvInfo[sv] = {}));
-        options.selectedPlace = { dcid: place, name: place, types: [] };
-        options.selectedStatVars = emptySvInfo;
-        setSelectedOptions(options);
-        setPreviewOptions(options);
-      });
-  }
+  useEffect(() => {
+    loadStateFromURL(store);
+  }, []);
```
nick-nlb: This is an incomplete dependency array (although unlike the previous one, the behavior is what we want - initializing once on mount).

In the past, we've left these as warnings, but following a forward-looking principal of not leaving new warnings: 

- We could add a comment to disable lint consideration, but I don't really like leaving the dependency array dishonest.
- It can be a sign that we can rethink how and where this state synching is done.

nick-nlb (commented):
Thank you Pablo for all of this! 

There are some inline comments, but I'll put a couple of behavioral notes here.

Mismatch of responsive breakpoints.
- View in a wide desktop view
- Select a location (say “United States”)
- Select a breakdown (say “State”)
- You will see “Pick a statistical variable in the left pane” as well as the statistical variable explorer on the left.
- Slowly shrink the viewport
- At a certain point, we flip into mobile/tablet, where the lefthand explorer disappears.
- At this point, the message correctly switches to “Pick a statistical variable using the “select variable” button above.”
- However, the button does not appear.
- Shrink the viewport further, and the button will appear.

Facet Selector State state correspond with selected stat vars [I've suggested a fix in the comments]
- Select “United States”
- Select “State”
- Select  Health: Medical Condition: Diabetes: Percentage…”
- You will see two facets (these are correct).
- Now select Total Population
- We expect 15 facets now be available (but we only see the two).
- Refresh the page - you will see fifteen.

static/js/i18n/i18n_tool_messages.ts
pablonoel: This will be implemented as part of the standardization of the components, and style refactor.

static/js/tools/download/context.ts
pablonoel: No!, they will be removed as they are no longer necessary.

static/js/tools/download/page.tsx
pablonoel: Done!

static/js/tools/download/page.tsx
pablonoel: Yes, the new reference looks like

  const { setList, setLoading, setError, reqObj: facetsReqObj } = facets;

static/js/tools/download/page.tsx
pablonoel: Agreed, I refactored the way the loadStateFromURL works to accept only the setOptions

static/js/i18n/i18n_tool_messages.ts
nick-nlb: Let's implement these as part of this PR:
* Correcting the messages (these reference the map tool)
* Applying them the tool

static/js/tools/download/page.tsx
```diff
@@ -194,28 +111,28 @@ export function Page(props: PagePropType): ReactElement {
...
-        setFacetLoading(false);
+        setLoading(false);
       });
-  }, [selectedOptions, dataCommonsClient]);
+  }, [options.value, setList, setLoading, setError, facetsReqObj]);
 
-  const facetListCacheKey = selectedOptions
-    ? `${selectedOptions.selectedPlace.dcid}-${selectedOptions.enclosedPlaceType}`
+  const facetListCacheKey = options.value
+    ? `${options.value.selectedPlace.dcid}-${options.value.enclosedPlaceType}`
```
nick-nlb: The key here isn't including the stat var, and so the facet selector doesn't refresh when we change stat vars (it stales at the first stat var selected).

```
${options.value.selectedPlace.dcid}-${options.value.enclosedPlaceType}-${Object.keys(options.value.selectedStatVars).join(",")}
```

static/js/tools/download/page.tsx
```diff
@@ -253,405 +169,277 @@ export function Page(props: PagePropType): ReactElement {
...
-      if (_.isEmpty(val)) {
-        urlParams.delete(key);
-      } else {
-        urlParams.set(key, val);
-      }
-    }
-    window.location.hash = urlParams.toString();
-  }
+    return (): void => {
+      window.onhashchange = null;
```
nick-nlb: This pattern of directly setting the `window.onhashchange` (and then unsetting it here in the cleanup) will overwrite change listeners registered by other components.

Something like this is safer:

```
useEffect(() => {
  const handleHashChange = (): void => {
    loadStateFromURL(setOptions);
  };
  window.addEventListener("hashchange", handleHashChange);
  return (): void => {
    window.removeEventListener("hashchange", handleHashChange);
  };
}, [setOptions]);
```

I know that we've used this pattern in other tools - we should eventually refactor that.

nick-nlb (changes requested):
Thank you Pablo!

Have a look at the comments - and could we also update the description, screenshots to reflect the design changes since the PR was originally created?

nick-nlb:
@pablonoel Some questions to consider:

- Now that we are guaranteed tidy long because we only have a single stat var, should the Date, Value and Source columns have the stat var removed (it serves no disambiguating purpose). The stat var could instead be incorporated into the download name.
- Should the button to download be converted to our current button component and style set?

static/js/tools/download/page.tsx
pablonoel: Agreed!

static/js/tools/download/page.tsx
pablonoel: Done!

pablonoel:
> @pablonoel Some questions to consider:
> 
>     * Now that we are guaranteed tidy long because we only have a single stat var, should the Date, Value and Source columns have the stat var removed (it serves no disambiguating purpose). The stat var could instead be incorporated into the download name.
> 
>     * Should the button to download be converted to our current button component and style set?

Yes to both, I can take care of the button.

pablonoel:
localhost_8080_tools_download(Laptop SM) (1)
localhost_8080_tools_download(Laptop SM) (2)
localhost_8080_tools_download(Laptop SM) (3)
localhost_8080_tools_download(Laptop SM) (4)

static/js/tools/shared/vis_tools/vis_tool_instructions_box.tsx
```diff
@@ -28,6 +28,16 @@ import { intl } from "../../../i18n/i18n";
 import { toolMessages } from "../../../i18n/i18n_tool_messages";
 
 const toolInstructions = {
+  download: {
+    variable: {
+      desktop: intl.formatMessage(
+        toolMessages.infoBoxInstructionsVariableDesktopMap
```
nick-nlb: Although the text of this message is the same as the Map version, it is semantically a different concept (the map tool text versus the download tool text). Can we convert these to messages for the download tool?

nick-nlb (changes requested):
Thank you for the updates!

One item left in the comments, as well as another item to still be addressed:
The removal of the stat var from the column names (Date:Median_Age_Person -> Date), and the move of the stat var name into the file download.

static/js/tools/shared/vis_tools/vis_tool_instructions_box.tsx
pablonoel: Done!

pablonoel:
Feature flag implementation is up, there is some extra logic that needed to be implemented on the stat_var_chooser in order to accommodate for both ways to coexist, other than that when the feature flag is inactive, things should look and work exactly as they where.

server/config/feature_flag_configs/autopush.json
```diff
@@ -23,6 +23,12 @@
     "owner": "nick-nlb",
     "description": "Diverts traffic to Spanner database backend."
   },
+  {
+    "name": "download_tool",
```
nick-nlb: `download_tool` isn't really descriptive of what this flag is (we are always using a download tool, either old or new).

`use_new_download_tool`

(also then moving this to its new spot alphabetically).

server/routes/tools/html.py
```diff
@@ -136,13 +137,23 @@ def stat_var():
 
 @bp.route('/download')
 def download():
-  # List of DCIDs displayed in the info page for download tool
-  # NOTE: EXACTLY 2 EXAMPLES REQUIRED.
-  info_places = _load_example_file('download', default=[])
+  download_tool = is_feature_enabled(DOWNLOAD_TOOL_FEATURE_FLAG,
```
nick-nlb: Likewise, download_tool here doesn't describe what this boolean really represents. It, along with the constant here, can be renamed to correspond to the update to the flag name itself.

static/js/shared/facet_selector/facet_selector.tsx
```diff
@@ -43,6 +43,10 @@ import { facetSelectionComponentMessages } from "../../i18n/i18n_facet_selection
 import { messages } from "../../i18n/i18n_messages";
 import { FacetSelectionCriteria } from "../../types/facet_selection_criteria";
 import { findMatchingFacets } from "../../utils/data_fetch_utils";
+import {
+  DOWNLOAD_TOOL_FEATURE_FLAG,
```
nick-nlb: The frontend flags should also be similarly updated to more actively represent what the flag is doing.

static/js/shared/types.ts
```diff
@@ -69,8 +74,24 @@ export const StatVarHierarchyType = {
...
 
+/**
+ * Returns whether the given stat var hierarchy type should render as a
+ * radio button (single select) instead of a checkbox (multi select).
+ *
+ * DOWNLOAD only behaves as single-select when the new download tool is
+ * enabled - the legacy download tool still supports selecting multiple
+ * stat vars via checkboxes.
+ */
+export function isRadioButtonType(type: string): boolean {
```
nick-nlb: It feels like this function belongs more with the stat var hierarchy, rather than in this generic type function (it will never be used anywhere outside of those files specifically associated with the hierarchy). Can we move this to the primary hierarchy and export it from there?

nick-nlb (changes requested):
Thank you for the updates!

In addition to inline comments:

## Column headers

Now that the tool displays data in tidy long, we no longer need to have the stat var in the headers. I.e., `Source:Median_Income_Person` should just be `Source`. The same applies for `Date` and `Value`. 

However, let's include the stat var in the download file name itself (now that it is "global" to the file that is downloaded).

## Download button reactivity

When you click the download button, it takes a second before the download is indicated by the browser, and the button itself gives you no indication that the click was received and that something is happening.

We solved this in the download dialog by animating a temporary check icon in the button that provides feedback to the user. You can see this by going to the explore page, opening any chart and clicking Copy Values. What do you think of using this here? (open to ideas on this).

pablonoel:
> Thank you for the updates!
> 
> In addition to inline comments:
> 
> ## Column headers
> Now that the tool displays data in tidy long, we no longer need to have the stat var in the headers. I.e., `Source:Median_Income_Person` should just be `Source`. The same applies for `Date` and `Value`.
> 
> However, let's include the stat var in the download file name itself (now that it is "global" to the file that is downloaded).
> 
> ## Download button reactivity
> When you click the download button, it takes a second before the download is indicated by the browser, and the button itself gives you no indication that the click was received and that something is happening.
> 
> We solved this in the download dialog by animating a temporary check icon in the button that provides feedback to the user. You can see this by going to the explore page, opening any chart and clicking Copy Values. What do you think of using this here? (open to ideas on this).

Yes! changed the headers and good idea to implement the download button, all is done in the next push

nick-nlb (approved):
Thank you Pablo for all of this!

PR #6416, opened 2026-06-30
feat: Add Croissant JSON-LD metadata for Datasets

dtulsyann:
- Dynamically injects Croissant JSON-LD metadata into the <head> of Dataset pages in the Data Commons browser.

- Introduces the 'show_croissant_json_ld' feature flag to safely toggle this behavior across different environments.

### DIFF
--- a/server/config/feature_flag_configs/production.json
+++ b/server/config/feature_flag_configs/production.json
@@ -1,5 +1,11 @@
 [
   {
+    "name": "show_croissant_json_ld",
+    "enabled": false,
+    "owner": "dtulsyan",
+    "description": "Feature flag to enable Croissant JSON-LD for Dataset nodes."
+  },
+  {
     "name": "autocomplete",
     "enabled": true,
     "owner": "gmechali",
--- a/server/config/feature_flag_configs/production.json
+++ b/server/config/feature_flag_configs/production.json
@@ -1,11 +1,5 @@
 [
   {
-    "name": "show_croissant_json_ld",
-    "enabled": false,
-    "owner": "dtulsyan",
-    "description": "Feature flag to enable Croissant JSON-LD for Dataset nodes."
-  },
-  {
     "name": "autocomplete",
     "enabled": true,
     "owner": "gmechali",
--- a/server/config/feature_flag_configs/staging.json
+++ b/server/config/feature_flag_configs/staging.json
@@ -1,11 +1,5 @@
 [
   {
-    "name": "show_croissant_json_ld",
-    "enabled": false,
-    "owner": "dtulsyan",
-    "description": "Enables Croissant JSON-LD metadata on dataset pages."
-  },
-  {
     "name": "autocomplete",
     "enabled": true,
     "owner": "gmechali",
@@ -94,6 +88,12 @@
     "enabled": false,
     "owner": "javiervazquez",
     "description": "Makes the page overview feature, generated from query and relevant stat vars, generally available."
+  },
+  {
+    "name": "show_croissant_json_ld",
+    "enabled": false,
+    "owner": "dtulsyan",
+    "description": "Enables Croissant JSON-LD metadata on dataset pages."
   },
   {
     "name": "use_v2_api",
--- a/server/config/feature_flag_configs/staging.json
+++ b/server/config/feature_flag_configs/staging.json
@@ -1,5 +1,11 @@
 [
   {
+    "name": "show_croissant_json_ld",
+    "enabled": false,
+    "owner": "dtulsyan",
+    "description": "Enables Croissant JSON-LD metadata on dataset pages."
+  },
+  {
     "name": "autocomplete",
     "enabled": true,
     "owner": "gmechali",
--- a/server/lib/croissant_metadata.py
+++ b/server/lib/croissant_metadata.py
@@ -16,18 +16,19 @@ import logging
 import server.lib.feature_flags as feature_flags_lib
 import server.services.datacommons as dc
 import server.lib.fetch as fetch
+import requests
 
 
 def build_dataset_metadata(dcid: str) -> dict:
   """Builds a Croissant JSON-LD dictionary for a Dataset node."""
 
+  if not feature_flags_lib.is_feature_enabled(
+      feature_flags_lib.CROISSANT_JSON_LD_FEATURE):
+    return {}
+
   # Verify the node is actually a Dataset node
   node_types = fetch.property_values([dcid], 'typeOf').get(dcid, [])
   if 'Dataset' not in node_types:
-    return {}
-
-  if not feature_flags_lib.is_feature_enabled(
-      feature_flags_lib.CROISSANT_JSON_LD_FEATURE):
     return {}
 
   json_ld_data = {
@@ -128,14 +129,14 @@ def build_dataset_metadata(dcid: str) -> dict:
                          {}).get("nodes", [])
           if s_url_nodes:
             source_obj["url"] = s_url_nodes[0].get("value")
-        except Exception as e:
+        except (ValueError, requests.exceptions.RequestException) as e:
           logging.warning("Error fetching source URL for %s: %s", s_dcid, e)
-          raise
+          # Continue without the URL
 
       json_ld_data["creator"] = [source_obj]
 
-  except Exception as e:
+  except (ValueError, requests.exceptions.RequestException) as e:
     logging.error("Error fetching metadata for %s: %s", dcid, e)
-    raise
+    return {}
 
   return json_ld_data
--- a/server/lib/croissant_metadata.py
+++ b/server/lib/croissant_metadata.py
@@ -1,0 +1,141 @@
+# Copyright 2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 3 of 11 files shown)


server/lib/croissant_metadata.py
```diff
@@ -0,0 +1,142 @@
...
+        try:
+          s_resp = dc.v2node([s_dcid], "->url")
+          s_url_nodes = ((((s_resp.get("data") or {}).get(s_dcid) or
+                           {}).get("arcs") or {}).get("url") or
+                         {}).get("nodes", [])
+          if s_url_nodes:
+            source_obj["url"] = s_url_nodes[0].get("value")
+        except (ValueError, requests.exceptions.RequestException) as e:
+          logging.warning("Error fetching source URL for %s: %s", s_dcid, e)
+          # Continue without the URL
```
SandeepTuniki: Stray comment. We can remove this.

server/lib/croissant_metadata.py
```diff
@@ -0,0 +1,142 @@
...
+          "replace": "cr:replace",
+          "separator": "cr:separator",
+          "source": "cr:source",
+          "subField": "cr:subField",
+          "transform": "cr:transform"
+      },
+      "@type":
+          "Dataset",
+      "conformsTo":
+          "http://mlcommons.org/croissant/1.0",
```
SandeepTuniki: Should this be `1.1` to account for the updated spec https://docs.mlcommons.org/croissant/docs/croissant-spec-1.1.html?

server/routes/browser/html.py
```diff
@@ -47,7 +48,11 @@ def browser_node(dcid):
       node_name = api_name
   except Exception as e:
     logging.info(e)
+
+  json_ld_data = croissant_metadata_lib.build_dataset_metadata(dcid)
```
SandeepTuniki: I'm wondering what is the best way to conditionally trigger this code only for the dataset pages?

I just spoke to Daksh offline and he noted the pattern in the rest of the codebase is to follow something like this:

```py
if 'Dataset' not in node_types:
  ...
```

But this requires first making an api call. We probably should avoid making an api call to mixer to figure out what page the user requested.

@nick-nlb do you know of a better way to check this?

SandeepTuniki (changes requested):

server/lib/croissant_metadata.py
dtulsyann: Removed.

server/lib/croissant_metadata.py
dtulsyann: Yes. Changed to 1.1

server/config/feature_flag_configs/production.json
```diff
@@ -1,4 +1,10 @@
 [
+  {
+    "name": "show_croissant_json_ld",
+    "enabled": false,
+    "owner": "dtulsyan",
+    "description": "Feature flag to enable Croissant JSON-LD for Dataset nodes."
```
nick-nlb: The rule for adding (or removing) feature flags is: don't submit a change that includes production.json and another file. Can you remove the change to production.json, submit this PR with all the other files, then put up a separate PR after just for production.json?

server/config/feature_flag_configs/staging.json
```diff
@@ -1,4 +1,10 @@
 [
+  {
+    "name": "show_croissant_json_ld",
```
nick-nlb: For the non-production flag lists, can we move the flag entry to its alphabetical place? (in this case below the `page_overview_ga`entry).

server/routes/browser/html.py
nick-nlb: I agree that we should avoid making that extra API call. It would create an extra round-trip to Mixer for almost every browser entry, even though very few are actually nodes of type `Dataset`.

There are a couple of other options:

- Prefix sniffing: every Dataset DCID begins with the prefix `dc/d`. I verified directly in the database that this is true for all of our base DC nodes. Rather than making an explicit API call to get the `typeOf`, we could look at the prefix of the DCID and use that to determine the type. The upside of this is that the check becomes zero-cost. The downside is that we are relying on a technique that might be fragile. We have to be very sure that this will always be an invariant that we can rely on. Will this prefixing always be the case, and will it work on all custom DCs? Mixer does rely in this invariant (and engages in prefix sniffing for datasets), but I don’t feel totally comfortable with it.
- Multi-property value call: We are actually already making separate API right above this already (and doing this on every browser page): `api_name = shared_api.names([dcid]).get(dcid)`. We could convert this to a multi-property value call, where we ask for the name and typeOf together in a single API call: `fetch.multiple_property_values([dcid], ["name", "typeOf"])`. This way, we still only have a single-round trip call to mixer, and I believe that both with BT and Spanner, the additional load would be minimal. It gives us a true, canonical typeOf for the node without relying on the prefix, and without an additional API call.

I prefer the second option, but let me know your thoughts.

If we do the second option, we can get both the name and the type by doing something like this:

```
    node_info = fetch.multiple_property_values([dcid], ["name", "typeOf"])
    dcid_info = node_info.get(dcid, {})
    names = dcid_info.get("name", [])
    if names:
      node_name = names[0]
    node_types = dcid_info.get("typeOf", [])
```

Then we would conditionally call the build_dataset_metadata based on the node_types (and the flag).

dtulsyann: I completely agree! The second option is definitely the cleanest approach.

I have updated `html.py` to use `fetch.multiple_property_values` so we can fetch both `name` and `typeOf` in a single network round-trip. I also removed the redundant `typeOf` API check from `croissant_metadata.py` so we aren't querying it twice.

server/config/feature_flag_configs/production.json
dtulsyann: Done! I've completely removed production.json from this PR in my latest push. I'll open a separate PR just for that one.

server/config/feature_flag_configs/staging.json
dtulsyann: I've updated all the non-production config files so the feature flags are now in alphabetical order.

server/routes/browser/html.py
```diff
@@ -41,13 +43,23 @@ def bio_browser_main():
 @bp.route('/<path:dcid>')
 def browser_node(dcid):
   node_name = dcid
+  node_types = []
   try:
-    api_name = shared_api.names([dcid]).get(dcid)
-    if api_name:
-      node_name = api_name
+    node_info = fetch.multiple_property_values([dcid], ["name", "typeOf"])
```
nick-nlb: Can we gate this change behind the flag as well?

I.e. if `show_croissant_json_ld` is false, then we do the `api_name = shared_api.names([dcid]).get(dcid)` call that we did before.

This update will then be closer to a no-op for environments wherethe flag is false or not set.

server/routes/browser/html.py
```diff
@@ -41,13 +43,23 @@ def bio_browser_main():
...
+    names = dcid_info.get("name", [])
+    if names:
+      node_name = names[0]
+    node_types = dcid_info.get("typeOf", [])
   except Exception as e:
     logging.info(e)
+
+  json_ld_data = {}
+  if 'Dataset' in node_types:
+    json_ld_data = croissant_metadata_lib.build_dataset_metadata(dcid)
```
nick-nlb: Relating to the above comment, we can move the flag check from outside this build_dataset_metadata. I.e., rather than calling build_dataset_metadata regardless of the flag (and then returning an empty result if the flag is set). 

This aligns with the idea of the flag as "flow control". I.e., the flag doesn't make it impossible to call the function (maybe we want to call it from some other context or script even when the flag is false), but it makes it so the flag isn't called in the flow we are dealing with. 

It also simplifies implementation of the first comment that I made, where, if the flag is not set, we will do the original api call, and then not call build_dataset_metadata at all. If the flag is set, we do the new multiple properties API call and then call build_dataset_medatadata.

It would also make the test a bit simpler too.

server/lib/croissant_metadata.py
```diff
@@ -0,0 +1,135 @@
...
+import requests
+
+import server.lib.feature_flags as feature_flags_lib
+import server.lib.fetch as fetch
+import server.services.datacommons as dc
+
+
+def build_dataset_metadata(dcid: str) -> dict:
+  """Builds a Croissant JSON-LD dictionary for a Dataset node."""
+
```
nick-nlb: As per the Codacy call-out: can we remove the blank line between the docstring and the function. We haven't always been consistent about it, but this aligns with PEP257 (https://pep257.readthedocs.io/en/latest/error_codes.html). It will also have Codacy pass.

nick-nlb (commented):
Thank you for the updates! A couple of more comments. 

The most important one is to gate the API change - the move to pulling the typeOf behind the feature flag (so that we are closer to a functional no-op when the flag is not set to true).

server/lib/croissant_metadata.py
dtulsyann: Done, the blank line has been removed!

server/routes/browser/html.py
dtulsyann: Done! I moved the feature flag check up into html.py. Now it safely gates both the API call and the dataset logic. If the flag is off, it just falls back to the original shared_api.names call.

server/routes/browser/html.py
dtulsyann: Handled this along with the comment above.

SandeepTuniki (approved):

PR #6431, opened 2026-07-07
feat: Add extended metadata and validations

dtulsyann:
- Enforce strict Croissant 1.1 validation requirements including name, url, license, creator, and datePublished.
- Add 'show_croissant_extended_feature' flag to gate extended dataset properties ( FileObject and recordSets).
- Add unit tests to strictly verify schema generation, extended payloads, and edge cases.

## How to Test
**1. Automated Unit Tests:**
Verify that the updated unit tests in `server/tests/lib/croissant_metadata_test.py` pass successfully.

**2. Test Missing Fields (Validation Rejection):**
- Start your local server (`./run_server.sh`)
- Navigate to a node that lacks a proper source/creator, dataset name, license, description etc.
- Look at the `metadata` script tag in the `<head>` of the DOM. The `application/ld+json` Croissant payload should *not* be injected into the page.

**3. Test Successful Extended Payload:**
- Navigate to a known good dataset, such as: `http://localhost:8080/browser/dc/d/UsCensusBureau_AmericanCommunitySurveyAcs`
- Open the Developer Tools (F12) -> Elements tab.
- Inspect the `<script type="application/ld+json">` tag in the `<head>`.
- Verify the following:
  - The root-level `name` and `url` fields exist.
  - The `creator` block exists.
  - The `license` field exists and is deterministic.
  - Ensure that the extended array nodes (like `distribution` and `recordSet`) do not contain invalid empty arrays if no provenances matched the filters.


server/config/feature_flag_configs/local.json
```diff
@@ -89,6 +89,12 @@
     "owner": "javiervazquez",
     "description": "Makes the page overview feature, generated from query and relevant stat vars, generally available."
   },
+  {
+    "name": "show_croissant_extended_feature",
+    "enabled": false,
```
nick-nlb, line 101: We probably do want to set this to true in local, autopush, especially if it is intended to go in and be invoked soon. As you have correctly done here, it should not change production at all)

server/lib/croissant_metadata.py
```diff
@@ -88,39 +208,113 @@ def build_dataset_metadata(dcid: str) -> dict:
...
+            if d > max_refresh_date:
+              max_refresh_date = d
+
+        # Extended Feature: distributions and recordSets
+        if extended_enabled:
+          p_download_urls = p_props.get("downloadUrl", [])
+          # contentUrl is required for FileObjects in Croissant 1.1
+          if p_download_urls:
+            prov_name = prov_node.get("name", prov_dcid)
+            prov_id = prov_name.replace(" ", "_")
```
nick-nlb: We are removing spaces here, but I'm not sure if this will cover all examples that might affect parsing/understanding of these ids.

Either way, could we extract this to a function that we can then add unit tests for (like `_sanitize_id`)? (with the unit tests indicating that both correct and edge cases are covered)?

Depending on what characters need to be avoided, the regex could be something like `sanitized = re.sub(r'[^\w\-]+', '_', name)` (this would swap out "#" etc as well - so it depends on what the requirements are for these ids).

server/lib/croissant_metadata.py
```diff
@@ -20,62 +20,182 @@
...
+    return False
+
+  if not json_ld_data.get("datePublished"):
+    logging.info("Skipping Croissant JSON-LD for %s: Missing 'datePublished'",
+                 dcid)
+    return False
+
+  if not json_ld_data.get("creator"):
+    logging.info(
+        "Skipping Croissant JSON-LD for %s: Missing 'creator' or 'creator url'",
```
nick-nlb: The logging line is somewhat ambiguous: it's checking if creator exists at all (which fits "Missing creator"), but the specific callout to "creator url" implies that even if creator exists, we should log if there is no url. This isn't being done here. Either the log message or the check should be updated.

server/lib/croissant_metadata.py
```diff
@@ -88,39 +208,113 @@ def build_dataset_metadata(dcid: str) -> dict:
...
 
-    # Fetch source
-    source_nodes = (arcs.get("isPartOf") or {}).get("nodes", [])
-    if source_nodes:
-      s_node = source_nodes[0]
-      s_name = s_node.get("name", s_node.get("value", ""))
-      s_dcid = s_node.get("dcid", "")
+    # 5. Provenance Handling
+    prov_resp = dc.v2node([dcid], "<-isPartOf")
+    prov_data = (prov_resp.get("data") or {}).get(dcid) or {}
```
nick-nlb, line 239: There are a few lines similar to this, where we are assuming that the call was completely successful and as expected (i.e., it does have "data" in the dictionary). Since we are catching the request exception below, we are most likely okay, but in case we do get a response but it is not as expected, we would get an attribute error here.

Although unlikely, we could add AttributeError to the catch below to fully harden this.

server/lib/croissant_metadata.py
```diff
@@ -88,39 +208,113 @@ def build_dataset_metadata(dcid: str) -> dict:
...
+
+            file_object = _build_file_object(prov_id, prov_name, p_desc,
+                                             p_props.get("url", []),
+                                             p_download_urls)
+            json_ld_data["distribution"].append(file_object)
+            json_ld_data["recordSet"].append(_build_record_set(prov_id))
+
+      # Apply final fallbacks
+      if not fallback_license_found and unique_licenses:
+        json_ld_data["license"] = list(unique_licenses) if len(
```
nick-nlb: Because unique_licenses is a set, the order is non-deterministic. In a sense, we might not care about the order of licenses, but in order to make this fully-deterministic (for example, for tests) we could sort the list (i.e., `sorted(list(unique_licenses))`

server/lib/croissant_metadata.py
```diff
@@ -20,62 +20,182 @@
...
+                 dcid)
+    return False
+
+  if not json_ld_data.get("creator"):
+    logging.info(
+        "Skipping Croissant JSON-LD for %s: Missing 'creator' or 'creator url'",
+        dcid)
+    return False
+
+  if extended_enabled and "distribution" in json_ld_data:
```
nick-nlb: It seems like this function is really doing two separate things :validating the dataset, and cleaning it up, and so the cleanup seems like an unanticipated side effect of the validation. 

I would move the cleanup (behind the extended_enabled flag) to another function and call it above here, leaving this function to simply validate.

nick-nlb (commented):
Thank you for this Daksh. Some comments in the review (all relatively minor). 

The adjustment to how the id is sanitized itself might be optional (although we should roll it out into a function that can be tested regardless).

In terms of the PR description itself:

Can we include a more detailed description that gives a breakdown of how to test the functionality. In this case, it might include URLs of datasets in the browser, and a description of what changes to expect.

server/lib/croissant_metadata.py
dtulsyann: Updated the log message to 'Missing creator'.

server/lib/croissant_metadata.py
dtulsyann: Done.

server/lib/croissant_metadata.py
dtulsyann: Added AttributeError.

server/lib/croissant_metadata.py
dtulsyann: I creator another function `_cleanup_dataset` to clean the empty arrays.

SandeepTuniki (approved):

server/config/feature_flag_configs/local.json
dtulsyann: It's not needed as of now.

server/lib/croissant_metadata.py
dtulsyann: Here we are using `provenance name` for `prov_id` and `prov_name` and provenance name is safe to be used as an id here.

PR #6432, opened 2026-07-07
Fractional diversion to spanner

nick-nlb:
## Issue

[b/520048943](b/520048943)

## Description

This PR implements fractional diversion to spanner, with the fraction in question mediated by a new `divert_to_spanner` flag with a `rollout_percentage` attribute.

There are two major components to this update:
* The first (and simplest) part is the cohort determination itself.
* The more complex update was to the caching system, to allow us to partition the caches between the two cohorts.

### Cohort Determination

We deterministically allocate users to the Spanner cohort based on a lightweight MD5 hash of User-Agent and client IP (extracting the first, original client IP from the `X-Forwarded-For`) with a server salt. It can be noted that the purpose of the hash is to prioritize even distribution of cohort assignments, rather than for it to be cryptographically secure. Related to that, we are not concerned with the possibility that the first entry in the IP list in `X-Forwarded-For` may be spoofed. Given the purpose of the rollout, a client forcing itself into a cohort is not considered a problem.

### Custom Caching for Partitioning

To prevent cache contamination between the legacy and Spanner cohorts, we now partition the cached responses. This ensures a request routed to the Bigtable backend does not retrieve a cached response generated by Spanner, or vice-versa.

We implement this partitioning at the cache client backend layer rather than at individual call sites. A new `CohortAwareBackendMixin` intercepts all standard cache operations (`get`, `set`, `add`, `inc`, `dec`, `has`, `delete`, `delete_many`, `get_many`, `set_many`). If a request context is present and assigned to the Spanner cohort (`g.use_spanner` is `True`), the mixin appends a suffix (`__cohort:spanner`) to the cache key before sending it to the parent backend client.

Key invalidation methods (`delete` and `delete_many`) are designed to be cohort-blind. They proactively evict both the base key and the Spanner-suffixed key to prevent stale cache entries.

We apply this mixin to all standard Flask-Caching storage clients (`RedisCache`, `FileSystemCache`, `SimpleCache`, `NullCache`) by registering custom backend factories in the Flask configuration. This makes the key partitioning transparent to application decorators (`@cache.cached`, `@cache.memoize`) and direct backend calls.

### Setting up Mixer 

To set up testing of this PR, you will need to run a local version of mixer that:
* has `--use_spanner_graph=false` (i.e., we are still primarily on BT and require diversion to go to Spanner)
* has the `UseSpannerGraph: true` directive set to true in `deploy/featureflags/local.yaml` (i.e., Spanner is enabled)

With the `local.yaml` file so updated, you can start mixer with:

```
./run_server.sh \
    --feature_flags_path=$PWD/deploy/featureflags/local.yaml \
    --spanner_graph_info="$(cat deploy/storage/spanner_graph_info.yaml)" \
    --use_spanner_graph=false \
    --use_base_bigtable=true \
    --embeddings_server_url=http://localhost:6060
```

### Website

Once mixer is set up, you must run your local Flask server with the `-l` directive:

```
./run_server.sh -m -l
```

There are three possible scenarios, depending on your feature flag setting in `local.json`:
* percentage rollout is set to 100: no hash is computed and we are guaranteed Spanner. All page hits should go to Spanner (this can be verified by viewing your mixer console logs after hitting an explore or other page).
* percentage rollout is set to 0: no hash is computed and we are guaranteed BT. This can likewise be verified by reviewing mixer's console logs
* A percentage greater than zero but less than 100: you will either be sent to BT or Spanner depending on your IP address. You will have to experiment with the flag value depending on where you are to determine what percentage rollout will put you into each respective cohort.

### Cache testing

A note that your local webserver will not cache by default (it uses the null cache). In order to test caching, you will need to patch your local webserver temporarily to override this behavior and use the file cache.

Apply the following temporary patch:

```
diff --git a/server/lib/cache.py b/server/lib/cache.py
index b8c073ca3..79b5429ab 100644
--- a/server/lib/cache.py
+++ b/server/lib/cache.py
@@ -63,7 +63,10 @@ class CohortAwareBackendMixin:
     # Only suffix if in request context and assigned to Spanner
     if has_request_context() and g.get('use_spanner',
                                        False) and self._is_isolatable_key(key):
-      return f"{key}{self.COHORT_SUFFIX}"
+      suffixed = f"{key}{self.COHORT_SUFFIX}"
+      print(f"--> [CACHE ACCESS - SPANNER COHORT] {key} -> {suffixed}")
+      return suffixed
+    print(f"--> [CACHE ACCESS - REGULAR COHORT] {key}")
     return key
 
   def get(self, key, *args, **kwargs):
@@ -219,8 +222,12 @@ if cfg.USE_MEMCACHE or REDIS_HOST:
     })
 else:
   # For some instance with fast updated data, we may not want to use memcache.
+  # Temporarily configured as FileSystemCache for local caching verification.
   cache = Cache(
-      config={'CACHE_TYPE': 'server.lib.cache.cohort_aware_null_cache_factory'})
+      config={
+          'CACHE_TYPE': 'server.lib.cache.cohort_aware_filesystem_cache_factory',
+          'CACHE_DIR': os.path.join(Path(__file__).parents[2], '.cache')
+      })
 
 
 def should_skip_cache():
```

This will do two things: it will add a print to console indicating the cohort decided for any particular call. More significantly, it will set up local caching. 

In your `.cache` directory, you will see files created with each call made to mixer. Notably, on any given page (say, "Jobs in Texas") you should see a set of files added for a given cohort. While in that cohort, no new files should be added on subsequent visits (unless you delete the cache files first). However, subsequent loads of the page should be much faster.  Then, if you switch cohorts, you will see a new set of files created the first time you revisit the same page.

A note that, regardless of the cohort you are to be assigned to, the call made when the server starts up will be in the default BT cohort (as there is no request context to assign). The log generated by this initial call can be ignored.

## Follow-up PRs

A follow-up PR will implement fixed cohorting for certain IP addresses. The strategy here will be to allow the environment to specify IP addresses that will be guaranteed to be placed in either of the two cohorts. This will allow us to deterministically test either cohort from certain locations.

Finally, once Spanner has been fully rolled out (and set to be the only backend available), we will undo the updates we have made to the cache here, returning the cache to the simple, unpartitioned cache that existed before this PR.

## Codacy

Some of the Codacy errors are incompatible with our Python linter and cannot be fixed. Others are inconsistent with similar docstrings in the same file (and elsewhere in our repository).


dwnoble (approved):

n-h-diaz (approved):
Thanks Nick for walking through the testing with me! Very cool :)

PR #6443, opened 2026-07-10
Paginate descendent places

nick-nlb:
## Description

With Spanner [correctly] paginating the results of `v2` node, the `descendent_places` call in Flask was only delivering partial results (the first page) for larger collections of descendents.

This PR sets `default max_pages` to `None` in `descendent_places` to allow fetching all pages (triggering a fetch of all places).

## Screenshots

The following screenshots are taken using Spanner as a backend:

### Before

image

### After

image


n-h-diaz (approved):
Thanks!

PR #6458, opened 2026-07-16
feat: Add Explore dataset link if json-ld data exists

dtulsyann:
- Add an inline 'Explore dataset' link next to the 'typeOf: Dataset' header in the Knowledge Graph browser.

- Gate the new UI and its associated API fetch behind the presence of json_ld_data, ensuring it only appears when valid Croissant metadata has been generated.

### DIFF
--- a/static/js/browser/app.tsx
+++ b/static/js/browser/app.tsx
@@ -23,6 +23,7 @@ import React from "react";
 
 import { GoogleMap } from "../components/google_map";
 import { ASYNC_ELEMENT_HOLDER_CLASS } from "../constants/css_constants";
+import { ArrowForward } from "../components/elements/icons/arrow_forward";
 import { StatVarHierarchyType } from "../shared/types";
 import { StatVarHierarchy } from "../stat_var_hierarchy/stat_var_hierarchy";
 import { ImageSection } from "./image_section";
@@ -46,11 +47,13 @@ interface BrowserPagePropType {
   statVarId: string;
   nodeTypes: string[];
   shouldShowStatVarHierarchy: boolean;
+  hasJsonLdData?: boolean;
 }
 
 interface BrowserPageStateType {
   provenanceNames: { [key: string]: string };
   dataFetched: boolean;
+  datasetSource?: string;
 }
 
 function setListenerForScrollTo(): void {
@@ -102,6 +105,7 @@ export class BrowserPage extends React.Component<
     this.state = {
       dataFetched: false,
       provenanceNames: {},
+      datasetSource: "",
     };
   }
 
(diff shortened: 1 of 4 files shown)


SandeepTuniki (approved):

static/js/browser/app.tsx
```diff
@@ -218,18 +237,38 @@ export class BrowserPage extends React.Component<
   }
 
   private fetchData(): void {
-    axios
-      .get("/api/browser/provenance")
-      .then((resp) => {
-        const provenance = resp.data;
+    const promises: Promise<any>[] = [axios.get("/api/browser/provenance")];
```
nick-nlb, line 240: The use of `any` as a type: there are legacy instances of this in the codebase, but we now avoid it as it circumvents TypeScript entirely.

static/js/browser/app.tsx
```diff
@@ -218,18 +237,38 @@ export class BrowserPage extends React.Component<
...
         for (const provId in provenance) {
           const provenanceName = provenance[provId].name;
           provenanceNames[provId] = provenanceName;
         }
+
+        let datasetSource = "";
+        if (resps.length > 1) {
+          const isPartOfResp = resps[1].data[this.props.dcid] || [];
+          if (isPartOfResp.length > 0) {
+            datasetSource = isPartOfResp[0].dcid;
```
nick-nlb, line 264: Typing the responses would then make `isPartOfResp` a `Node` (and so to be safe, we would want to coalesce the `.dcid` with `|| ""` (because it is strictly speaking an optional aspect of Node).

static/js/browser/app.tsx
```diff
@@ -218,18 +237,38 @@ export class BrowserPage extends React.Component<
...
+      promises.push(
+        axios.get(
+          `/api/node/propvals/out?prop=isPartOf&dcids=${encodeURIComponent(
+            this.props.dcid
+          )}`
+        )
+      );
+    }
+
+    Promise.all(promises)
```
nick-nlb, line 251: With the `Promise.all` pattern as implemented here, it will fail completely if either of the calls fail (including the new one). We could be more defensive about this and allow the page to display with just the result of the first call.

nick-nlb (commented):
Thank you for this Daksh!

The PR is already in (and looks good) but a couple of comments below (we can decide offline whether or not to follow-up).

Some general comments on the description of the PR:

* In addition to the description of what the PR does, we should include steps for viewing the functionality (including localhost links). In this case something like localhost:8080/browser/dc/d/AustralianBureauOfStatistics_AustraliaStatistics
* Where relevant, the we should includes steps to verify and test the functionality.
* If there are visual changes, including screenshots can be helpful too, as they allow people without local repos to get an immediate snapshot of the changes.

PR #6472, opened 2026-07-21
Update facet matching logic to fallback to provenanceId

nick-nlb:
## Description

Spanner no longer includes import names. However, we do use import names as a component for matching facets (in order to preselect a facet in the highlighted chart section of the explore page).

This PR adds logic similar to that found in the metadata API, to provide fallback logic for when the import name is not provided in the facet, but is given as a facet selector.  Import names will not be provided in facet logic going forward; however, the fallback logic will handle existing URLs to prevent existing links from changing behavior.

This error was surfaced in a webdriver test (test_highlight_chart_facet_selector). This test now passes in production Spanner.

### DIFF
--- a/static/js/shared/stat_types.ts
+++ b/static/js/shared/stat_types.ts
@@ -23,6 +23,7 @@
 // TODO (nick-next): comment the client library version to document that `StatMetadata` matches.
 
 export interface StatMetadata {
+  provenanceId?: string;
   importName?: string;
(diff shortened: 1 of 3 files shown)


n-h-diaz (approved):
Thank you for the fix! 

(if we use the base prefix elsewhere, maybe can reuse here, but only if it doesn't complicate things)

PR #6476, opened 2026-07-21
Mitigate graphql errors with smaller request

beets:
We are getting frequent (though not constant) 503 errors from that line in the workflow runs. Trying to mitigate with this fix.


keyurva (approved):
Thanks!

juliawu (approved):

PR #6489, opened 2026-07-23
Remove example with missing data

nick-nlb:
## Description

This PR tentatively removes one of the scatter plot examples, pending whether data ingestion can be completed in time. If data ingestion makes this data available in Spanner, this PR will be discarded.

## Notes

This should not be merged if data for the existing example gets in. Additionally, as it is a publicly visible change to website copy, **the change should be confirmed and approved by a lead before being merged in**. If required, the example can be changed.


beets (approved):

nick-nlb:
This PR is now no longer necessary.

PR #6505, opened 2026-07-24
fix(build): remove latest tag from cloudbuild.push_image.yaml

clincoln8:
## Goal
Remove the automatic pushing of the `:latest` tag in `build/ci/cloudbuild.push_image.yaml` during manual/dev container image builds.

## Key Aspects of the Approach
* Removed `--tag ...:latest` build options and `docker push ...:latest` steps for both `datacommons-services` and `website-cron-testing` images in `build/ci/cloudbuild.push_image.yaml`.
* Guarantees that ad-hoc/dev image pushes (e.g. via `./scripts/push_image.sh`) only push to the explicitly provided tag (`_TAG`, such as `dev-<git-hash>`), preserving tag isolation for downstream environments.

## Testing & Verification
* Verified `build/ci/cloudbuild.push_image.yaml` step definitions against image tagging expectations.

### DIFF
--- a/build/ci/cloudbuild.push_image.yaml
+++ b/build/ci/cloudbuild.push_image.yaml
@@ -57,10 +57,8 @@ steps:
           --build-arg MIXER_HASH=$(cat mixer_hash.txt) \
           --build-arg WEBSITE_HASH=$(cat website_hash.txt) \
           --tag gcr.io/$_PROJECT_ID/datacommons-services:${_TAG} \
-          --tag gcr.io/$_PROJECT_ID/datacommons-services:latest \
           .
         docker push gcr.io/$_PROJECT_ID/datacommons-services:${_TAG}
-        docker push gcr.io/$_PROJECT_ID/datacommons-services:latest
     waitFor: ["resolve-hashes"]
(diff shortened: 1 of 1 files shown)


build/ci/cloudbuild.push_image.yaml
```diff
@@ -95,10 +93,8 @@ steps:
         DOCKER_BUILDKIT=1
         docker build -f build/website_cron_testing/Dockerfile \
           --tag gcr.io/$_PROJECT_ID/website-cron-testing:${_TAG} \
-          --tag gcr.io/$_PROJECT_ID/website-cron-testing:latest \
```
gmechali, line 98: my guess is our CI isn't going to automatically put it back in so if we submit this without the change to the trigger, we will never push a latest again?

I dont think it's worth your time rn tbh. A minor nuisance for now that we will fix :)

gmechali: chatted oflfine, seems like it uses a different one from the trigger.

gmechali (approved):
thanks!

PR #6533, opened 2026-07-28
add logic to read numerical value from config

shixiao-coder:
This change add theshold from app config and use this logic to set the spanner embeddings theshold from it


server/lib/feature_flags.py
```diff
@@ -132,6 +133,20 @@ def is_feature_enabled(feature_name: str, app=None, request=None) -> bool:
...
+
+  if not is_feature_enabled(feature_name, app, request):
+    return default_value
+  feature_flags = app.config.get('FEATURE_FLAGS', {})
+  flag_config = feature_flags.get(feature_name, {})
+
+  if not flag_config:
+    return default_value
+
+  return flag_config.get('numerical_value', default_value)
```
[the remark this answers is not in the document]
shixiao-coder: added app with resolve_flask_app

server/tests/lib/feature_flag_test.py
```diff
@@ -279,3 +280,26 @@ def test_spanner_diversion_ip_overrides(self):
...
+    mock_feature_flags(self.app, [TEST_FEATURE_FLAG], True)
+    with self.app.test_request_context():
+      self.assertEqual(
+          get_feature_flag_value(TEST_FEATURE_FLAG, 0.5, self.app), 0.5)
+
+    # Test disabled flag returns default_value
+    mock_feature_flags(self.app, [TEST_FEATURE_FLAG], False)
+    with self.app.test_request_context():
+      self.assertEqual(
+          get_feature_flag_value(TEST_FEATURE_FLAG, 0.5, self.app), 0.5)
```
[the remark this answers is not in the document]
shixiao-coder: this is intentional, all tests are passing app

server/app_env/_base.py
```diff
@@ -123,3 +123,5 @@ class Config:
   # Target parameter for v2/resolve API calls when resolver=indicator
   V2_RESOLVE_INDICATORS_TARGET = os.environ.get('V2_RESOLVE_INDICATORS_TARGET',
                                                 '')
+  # Threshold for Spanner vector search embeddings resolution in NL search.
+  SPANNER_EMBEDDING_THRESHOLD = 0.7
```
gmechali: https://github.com/shixiao-coder/mixer/blob/fb6bfc6164751f66ebb461f4960caacf0f3ec738/internal/server/spanner/spanner_config/default.yaml

hey xiao, why can't we just use this file where you already put spanner search related configurations? wouldn't that be a better place?

shixiao-coder: That one is used as a v2/resolve level filter, the purpose is to control v2/resolve quality for all use case (website, MCP etc.). I would prefer to separate it with the threshold we provide to website. This is the same/similar behavior through NL server, website are having a different threshold with mixer. lmk if there is a preference to unifying it

server/integration_tests/test_data/demo_fallback/query_1/chart_config.json
```diff
@@ -9,13 +9,34 @@
               {
                 "tiles": [
                   {
-                    "type": "PLACE_OVERVIEW"
```
shixiao-coder, line 12: verified the change is reasonable, it will filter out all SV with threshold 0.7 and able to capture SV with threshold 0.6

server/integration_tests/test_data/e2e_correlation_simple_place/foreignbornvs.nativeborninsunnyvale/chart_config.json
```diff
@@ -10,308 +10,115 @@
                 "tiles": [
                   {
                     "statVarKey": [
-                      "Count_Person_5OrMoreYears_ForeignBorn"
+                      "Count_Person_5OrMoreYears_ForeignBorn",
```
shixiao-coder: restores back to two SVs as the test intended

server/integration_tests/test_data/e2e_correlation_simple_place/nativebornvs.medianincomeinsunnyvale/chart_config.json
```diff
@@ -10,55 +10,28 @@
                 "tiles": [
                   {
                     "statVarKey": [
-                      "Median_Income_Person_15OrMoreYears_WithIncome_ForeignBorn"
+                      "Count_Person_ForeignBorn"
```
shixiao-coder: properly filter-in MultiSV after threshold decrease

server/integration_tests/test_data/e2e_edge_cases2/whatcrimesareconsideredfeloniesvs.misdemeanorsintheus/chart_config.json
```diff
@@ -14,16 +14,16 @@
                     },
                     "statVarKey": [
                       "Count_CriminalActivities_CombinedCrime_scatter",
-                      "Count_CriminalActivities_AggravatedAssault_scatter"
```
shixiao-coder: heuristic dection type can already find SVs above threshold and didn't goes to LLM detection like before

server/integration_tests/test_data/e2e_single_date/whatwastheaveragehousepricefor2brhouseinmountainviewdecadeago/chart_config.json
```diff
@@ -145,24 +177,21 @@
                       "geoId/0649670"
                     ],
                     "statVarKey": [
-                      "Count_HousingUnit_WithTotal2Rooms_multiple_place_bar_block_2016",
```
shixiao-coder, line 148: we have more SVs to show with lower threshold and not showing the related like before

server/integration_tests/test_data/e2e_toolformer_rag_mode/foreignbornvs.nativeborninsunnyvale/chart_config.json
```diff
@@ -9,20 +9,23 @@
...
                 "tiles": [
                   {
-                    "barTileSpec": {
-                      "maxPlaces": 15,
-                      "maxVariables": 15,
-                      "sort": "DESCENDING"
-                    },
-                    "comparisonPlaces": [
-                      "geoId/0677000",
-                      "geoId/06"
```
shixiao-coder, line 19: we are able to capture the 2 SV like before

server/integration_tests/test_data/multisv/query_3/chart_config.json
```diff
@@ -4,6 +4,35 @@
     "categories": [
       {
         "blocks": [
+          {
```
shixiao-coder: Lowering the theshold would cause multiSV to be captured and the related SVs to be excluded by multiSV by filter_with_single_var and exclude all top ranked SVs as a result.

This happens under test=filter_test setup. Without this setup, the query keeps the same

server/integration_tests/test_data/e2e_edge_cases2/countiesincaliforniawhereincomeisover50000/chart_config.json
```diff
@@ -4,6 +4,459 @@
     "categories": [
       {
         "blocks": [
+          {
```
shixiao-coder: under test=filter_test cases, this test will include more SVs with lower threshold, however order is not deterministic

the lowering of threshold include MultiSV into the case and causing the fulfillment to go through filter_with_dual_vars (changes the order from v2/observation) instead of filter_with_single_var (only return 1 chart, keep the order from v2/node)

server/integration_tests/test_data/place_detection_e2e_dc/query_2/chart_config.json
```diff
@@ -73,10 +73,335 @@
               {
                 "tiles": [
                   {
+                    "barTileSpec": {
```
shixiao-coder: lowering the threshold causing us to capture 20+ SVs&topics instead of previous 2 SV&topics only.
The fulfillment logic allows max 2 individual chart before the group Bar chart for SVs > 3.
When we have only 2 SV, we allows 10 individual chart from topic and causing the reranking to happen between charts
This is the expected behavior

beets (commented):
Sorry for the late review - but for changes like this, we should split it with: 
- application logic update to add support for the flag (but keeping the value stable)
- a follow on PR where we start playing with the threshold.

The flag support should be straightforward and we can hone in on the right value for the flag after.

Separately, any reason why we shouldn't extend support to do this via feature flag instead? Changes to the thresholds using the env files require full binary rebuilds. I believe we have this support in the mixer feature flags already.

gmechali (approved):

PR #6536, opened 2026-07-28
[v0/v1 migration] Remove deprecated ranking pages

nick-nlb:
## Description

This PR removes the deprecated legacy ranking page and its associated components, routes and tests.


juliawu (approved):
Hooray for un-launches! Thank you for the cleanup.

PR #6538, opened 2026-07-28
[feature-flags] Add PR warning annotations, oncaller CODEOWNERS, and deployment review reminders for production flag updates

beets:
## Summary
This PR improves safety, review visibility, and team coordination around production feature flag changes in the `website` repository.

## Changes
- **CODEOWNERS (.github/CODEOWNERS)**:
  - Fixed a directory path typo (`server/config/feature_flag_config/` → `server/config/feature_flag_configs/`).
  - Assigned `@datacommonsorg/oncallers` as the Code Owner for `server/config/feature_flag_configs/production.json`.
- **GitHub PR Review Checklist (.github/workflows/feature-flag-checks.yml)**:
  - Created a new GitHub Actions check that runs on pull requests modifying `server/config/feature_flag_configs/production.json`.
  - Surfaces an inline GitHub warning annotation (`::warning::`) across the PR **Conversation** and **Files changed** tabs with a 6-point review checklist:
    1. **Staging Validated**: Ensuring exact changes were tested in staging.
    2. **Rolling Restart**: Reminding that merging triggers a rolling pod restart in production.
    3. **Cache Invalidation**: Reminding to clear Redis cache if API responses change.
    4. **Time Window**: Encouraging global working hours and requiring lead approval for Fri–Sun updates.
    5. **Oncall Communication**: Advising coordination and announcement on the oncall thread.
    6. **Type Safety**: Verifying binary type compatibility.
- **CLI Deployment Safety (scripts/update_gcs_feature_flags.sh)**:
  - Added a reminder to conduct a careful review, coordinate with the oncaller, and announce updates on the oncall thread before confirming a Kubernetes cluster restart.

## Testing
- Verified workflow YAML syntax and multi-line annotation message formatting.
- Verified bash syntax for interactive prompts in `update_gcs_feature_flags.sh`.

PR #6543, opened 2026-07-29
[feature-flags] Update workflow to post to the PR comment thread directly

beets:
The last PR #6538 added the note in the check which won't be as visible. Updating it to be inline with the PR comments -- will need to test this post-merge with a benign change that I won't submit.


juliawu (approved):
Thanks! I like this checklist.

beets:
verified that this worked!

image

PR #6547, opened 2026-07-30
add the feature flag use_config_threshold_for_spanner_embedding to gate using new config 0.6 threshold from default 0.7 threshold

shixiao-coder:
This PR add the feature flag in non prod environment and default to false

When false, we will read default value 0.7 from the code,
When true, we will set config value to be 0.6 based on test result https://docs.google.com/spreadsheets/d/12d9BPJagWS5z2WA8En9JY71TyGrCYpIBQ4LqI0JJ2v4/edit?gid=1132426848#gid=1132426848


gmechali (approved):
Make sure you submit this one first! And that it gets to staging before you ensble the flag in staging. That might only be next week, given that staging was cut a few hours ago. 

https://github.com/datacommonsorg/website/pull/6533

PR #6548, opened 2026-07-30
add the use_config_threshold_for_spanner_embedding feature flag to prod following the flag for other env setup

shixiao-coder:
This PR add the feature flag in prod environment and default to false

When false, we will read default value 0.7 from the code,
When true, we will set config value to be 0.6 based on test result https://docs.google.com/spreadsheets/d/12d9BPJagWS5z2WA8En9JY71TyGrCYpIBQ4LqI0JJ2v4/edit?gid=1132426848#gid=1132426848


juliawu (approved):

gmechali (commented):
Can you ping us when your PR made it to production github.com/datacommonsorg/website/pull/6533
So we can approve this then. A bit too early for this PR :)

gmechali (approved):
approved, thanks!

PR #6557, opened 2026-08-03
[v0/v1 migration] Remove deprecated SDG topic generation tools

nick-nlb:
## Description

This PR removes now unsed SDG topic generation tools (that used the soon-to-be-deprecated v1/bulk/variable/group endpoint).


dwnoble (approved):

dwnoble:
Thanks Nick! There may also be some gitignore lines here to get rid of: https://github.com/datacommonsorg/website/blob/master/.gitignore#L72-L73

nick-nlb:
> Thanks Nick! There may also be some gitignore lines here to get rid of: https://github.com/datacommonsorg/website/blob/master/.gitignore#L72-L73

Done! I also noticed that the following two (similar) directives in the gitignore are also not relevant and they are removed too.

PR #6558, opened 2026-08-03
Refine download CSV headers and preview UI

pablonoel:
Rename tidy CSV header constants to friendlier labels (e.g. "ISO code", "Date", "Value") to match the new download tool. Simplify page layout by removing an extra wrapper and render Preview directly. Improve Preview: move stat-var titles into the preview header, make header/button responsive, adjust padding/gaps, and add table styling (responsive overflow, borders, font sizing, and cell padding) for better readability and UX.

localhost_8080_tools_download(Laptop SM)

localhost_8080_tools_download(Laptop SM) (1)


server/routes/shared_api/csv.py
```diff
@@ -34,18 +34,18 @@
...
+    "Date",
+    "Import name",
+    "Measurement method",
+    "Observation period",
+    "Provenance URL",
+    "Scaling factor",
+    "Unit",
+    "Unit display name",
+    "Value",
     "Variable properties name",
```
kmoscoe: This should just be Name.

server/routes/shared_api/csv.py
```diff
@@ -34,18 +34,18 @@
...
-    "Variable observation metadata unitDisplayName",
-    "Variable observation value",
+    "Date",
+    "Import name",
+    "Measurement method",
+    "Observation period",
+    "Provenance URL",
+    "Scaling factor",
+    "Unit",
+    "Unit display name",
```
kmoscoe: You shouldn't need to have this. It should just say "Unit".

server/routes/shared_api/csv.py
pablonoel: Wild do!

server/routes/shared_api/csv.py
pablonoel: Will do!

nick-nlb: There are still some issues with Unit.

The preview is actually displaying the Unit DCID. This is what is called "Unit".

The actual download contains both "Unit" (the Unit DCID), and then "Unit display name". Additionally, the "Unit display name" is the last column, rather than being next to the Unit DCID.

* Move what is currently called the "Unit display name" column to be adjacent to "Unit DCID".
* Should we display what is currently the "Unit display name" in the preview, rather than the Unit DCID? (I.e. USD rather than USDollar). @kmoscoe or @miss-o-soup could confirm.
* We could then rename that as "Unit", and call the other column "Unit DCID" (pending confirmation above).

Either way, the current implementation still contains "Unit display name", and so the comment isn't yet resolved.

static/js/tools/download/preview.tsx
```diff
@@ -42,6 +42,17 @@ import { DATE_ALL, DownloadOptions } from "./context";
 const NUM_ROWS = 7;
 const SECTION_ID = "preview-section";
 const DOWNLOADED_RESET_DELAY_MS = 1500;
+// Columns hidden from the preview table (still included in the downloaded
+// CSV).
+const PREVIEW_HIDDEN_COLUMNS = new Set<string>([
```
nick-nlb, line 49: Can we add a comment here indicating the file in which the CSV columns are defined, as well as a note that these need to be in sync, and so if the columns in one place are updated, they need to be updated in the other place.

After that, let's add in a comment in the corresponding file where the CSV file is defined, indicating the same (if those column titles are renamed, the versions in this file also need to be renamed.

nick-nlb (commented):
Thank you for this Pablo.

There are three remaining things left:

* Ensure that the Unit names are handled as per @kmoscoe and @miss-o-soup 's spec
* Add the comments referenced here
* Once the columns are finalized, we need to update the tests/mocks to match these new columns (I have updated the tests to bring up to the current state).

server/routes/shared_api/csv.py
pablonoel: Noted, pushed the swap of position and naming, both crosschecked with @miss-o-soup. 

Original field key | CSV column name (TIDY_CSV_HEADER_ROW) | In preview table?
-- | -- | --
place (entity dcid) | Entity DCID | Yes
entity.name | Entity name | Yes
entity.isoCode | ISO code | Yes
sv (variable dcid) | Variable DCID | Yes
variable_props[sv].name | Variable name | Yes
observation.date | Date | Yes
observation.value | Value | Yes
facet.unitDisplayName | Unit | Yes
facet.unit | Unit DCID | Hidden
facet.measurementMethod | Measurement method | Yes
facet.provenanceUrl | Provenance URL | Yes
facet.importName | Import name | Hidden
facet.observationPeriod | Observation period | Hidden
facet.scalingFactor | Scaling factor | Hidden

static/js/tools/download/preview.tsx
pablonoel: Comments added.

nick-nlb (approved):
Thank you Pablo!

PR #6567, opened 2026-08-05
[oncall cleanup] Disable periodic Node.js query testing

n-h-diaz:
Turn down periodic Node.js query testing across automated builds

### Description
This change introduces an explicit toggle (`enableNodejs: false`) to disable periodic Node.js query tests in automated GKE cron executions by default. This reduces automated test churn and email alert noise from `datacommons-alerts+tests@google.com` while the Node.js server is scheduled for future deprecation. Manual testing remains fully supported via existing CLI scripts or by toggling the flag.

### Changes
* Add `enableNodejs: false` to default Helm configuration in `values.yaml` and explicitly under `cronTesting` in `autopush.yaml`.
* Expose `enableNodejs` in the `cron-testing-config` Kubernetes ConfigMap and wire the `ENABLE_NODEJS` environment variable into the GKE CronJob template.
* Update `run_website_cron_tests.sh` to check `ENABLE_NODEJS == "true"` before running query tests and differ notifications.
* Update documentation in `gke/README.md` and `tools/nl/nodejs_query_differ/README.md` explaining the test toggle and deprecation status.


juliawu (approved):
Thanks for the changes! I like the added consistency of having an `enableNodejs` to match `enableSanity` and `enableAdversarial`.

Just flagging to all reviewers that setting the default to false means this would also stop the Nodejs query testing cron for the dev environment.

Also a suggestion: Maybe we want to update deploy/helm_charts/envs/staging.yaml to use `enableNodejs: true` instead of `nodejsApiRoot: ""` for clarity, but I don't think that blocks this PR.

n-h-diaz:
> Thanks for the changes! I like the added consistency of having an `enableNodejs` to match `enableSanity` and `enableAdversarial`.
> 
> Just flagging to all reviewers that setting the default to false means this would also stop the Nodejs query testing cron for the dev environment.
> 
> Also a suggestion: Maybe we want to update deploy/helm_charts/envs/staging.yaml to use `enableNodejs: true` instead of `nodejsApiRoot: ""` for clarity, but I don't think that blocks this PR.

Thanks for the review! Updated dev yaml to explicitly set enableNodejs to false, and also set enableNodejs false on staging for consistency

PR #6576, opened 2026-08-11
[oncall] Fix Explore sanity test context sharing and remove unsupported sample query

n-h-diaz:
## 1. Problem
During the weekly staging promotion sanity tests, the test suite was failing on the Explore health query: ["Does obesity correlate with lack of sleep in US counties?"](https://staging.datacommons.org/explore?enable_feature=enable_stat_var_autocomplete#q=Does%20obesity%20correlate%20with%20lack%20of%20sleep%20in%20US%20counties%3F) with repeated timeouts, despite the query resolving successfully to charts when visited directly in a browser.

## 2. Root Cause
1. SPA Context Accumulation in Test Runner: sanity.py iterates through all landing page sample queries sequentially within a single browser session. Because all explore query URLs share the same path (/explore) and only differ by the URL hash fragment (/explore#client=ui_landing&q=...), Selenium's driver.get() updates the hash without triggering a full page reload or unmounting the React app. As a result, savedContext.current in app.tsx accumulated conversational history from all preceding queries.

2. Suppression of LLM Fallback & Correlation Misrouting: When evaluating "Does obesity correlate with lack of sleep in US counties", heuristic SV matching misses "lack of sleep" under the recent Spanner embedding resolution.

* Without prior context: need_llm() triggers Gemini LLM fallback, correctly extracting both metrics and fulfilling 29 chart tiles.
* With prior context: Because the previous query ("How does child immunizations correlate with life expectancy in Asia") had existing SVs in context, need_llm() assumed the query was an incomplete follow-up and suppressed LLM fallback. The fulfillment handler then attempted a contextual correlation (context-sv-correlation) with the previous query's Asian child immunization variables, yielding 0 charts and falling back to a single PLACE_OVERVIEW tile for country/USA.
* The frontend's shouldSkipPlaceOverview() detected the single place overview tile and automatically redirected to /place/country/USA, where .explore-charts does not exist, triggering the test timeout.

3. Unsupported Demographics Sample Query: Removing the context sharing revealed that ["Which state in the US has most people of Caribbean origin?"](https://datacommons.org/explore?#q=Which%20state%20in%20the%20US%20has%20most%20people%20of%20Caribbean%20origin%3F) in topics.json had no matching statistical variable in Data Commons and was previously only passing by relying on context from earlier immigrant queries.

## 3. Changes
1. Isolate Explore Test Cases (sanity.py):
Navigates to about:blank prior to loading each explore URL in WebsiteSanityTest.explore(), ensuring each query is evaluated in a clean session without in-memory state carrying across test cases.
2. Remove Unsupported Example Query (topics.json):
Removed "Which state in the US has most people of Caribbean origin?" from the Demographics landing page examples (aligned with [#6514](https://github.com/datacommonsorg/website/pull/6514)).

## 4. Testing
* Verified Does obesity correlate with lack of sleep in US counties? now passes in ~1.3s.
* Ran full sanity suite locally against staging; all 100+ landing and homepage queries pass.


nick-nlb (approved):
Thank you Natalie, especially for the detailed write-up and explanation of the problem and solution.

PR #6579, opened 2026-08-12
remove the legacy feature flag of gemini 3

shixiao-coder:
The gemini 3 feature flag was use when we roll up from 2.5 to 3, it's no longer needed, all non prod env has removed this feature flag


nick-nlb (approved):
LGTM - thanks Xiao

gmechali (approved):

PR #6581, opened 2026-08-12
Overhaul entities sample to use new schema guidelines

kmoscoe:
Running at http://bullie.svl.corp.google.com:8080


custom_dc/sample/entities/schema.mcf
```diff
@@ -0,0 +1,86 @@
...
+measuredProperty: dcid:medicalCondition
+statType: dcid:count
+memberOf: dcid:AlaskaHospitals
+
+Node: dcid:Count_ReturnPatients_Pneumonia
+typeOf: dcs:StatisticalVariable
+name: "Count of pneumonia patients returned"
+description: "Per-hospital number of pneumonia patients returned unplanned in a 30-day period"
+populationType: dcid:ReturnPatient
+measuredProperty: dcid:cma/medicalCondition
```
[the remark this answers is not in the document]
kmoscoe: fixed.

custom_dc/sample/entities/schema.mcf
```diff
@@ -0,0 +1,86 @@
...
+populationType: dcid:Patient
+measuredProperty: dcid:medicalCondition
+statType: dcid:count
+memberOf: dcid:AlaskaHospitals
+
+Node: dcid:Count_ReturnPatients_HeartAttack
+typeOf: dcs:StatisticalVariable
+name: "Count of heart attack patients returned"
+description: "Per-hospital number of heart attack patients returned unplanned in a 30-day period"
+populationType: dcid:ReturnPatient
```
carojahn: Given it's a condition of returning, the populationType feels more like patients with a constraintProperty returned (this also makes it easier to compare against all patients

carojahn (changes requested):
Left one comment - not strongly opinionated but curious to get your take on why you made it a separate populationType

custom_dc/sample/entities/schema.mcf
kmoscoe: It's a good point, and I tried to come up with a justification in the doc for choosing between a population type / new class and a constraint property, but now I'm not so sure it was valid!  I think the problem here is what would the constraint property be? "returned"? It seems kind of weird: a property should reference a class but I don't know what the class would be here. Let's discuss this.

kmoscoe: OK, I've redone this as we discussed. PTAL.

gmechali (approved):

kmoscoe:
> Left one comment - not strongly opinionated but curious to get your take on why you made it a separate populationType

custom_dc/sample/entities/alaska_hospitals_unplanned_visits.csv
```diff
@@ -1,127 +1,127 @@
...
-CCN21309,2023-06-30,EDAC_30_PN_TOTAL,,P3Y
-CCN21309,2023-06-30,EDAC_30_PN_RETURN,,P3Y
-CCN21310,2023-06-30,EDAC_30_AMI_TOTAL,,P3Y
-CCN21310,2023-06-30,EDAC_30_AMI_RETURN,,P3Y
-CCN21310,2023-06-30,EDAC_30_HF_TOTAL,,P3Y
-CCN21310,2023-06-30,EDAC_30_HF_RETURN,,P3Y
-CCN21310,2023-06-30,EDAC_30_PN_TOTAL,,P3Y
-CCN21310,2023-06-30,EDAC_30_PN_RETURN,,P3Y
+CCN20001,2023-06-30,Count_Patients_HeartAttack,432,P3Y
+CCN20001,2023-06-30,Count_HeartAttack_ReturnPatients,92,P3Y
```
carojahn, line 3: I am assuming that these StatVars without Patient are the count of "heart attack occurrences"? While something like "count_patients_heartattack" measures the actual number of patients?

Wondering if there is a need to aggregate across all diseases for "count_heartattack_returnpatients" and others, in which case an alternative could be "count_medicalcondition_returningpatient_heartattack". Not strongly opinionated but curious how we'd justify the condition as a constraint in one and populationType in the other.

kmoscoe: No, both of these measures just record the actual number of patients, both returned and not returned. The actual EDAC (excess days in acute measure), which is the principal measure, and measures just the additional days of the returned patients, is not included here. Maybe I should do so to make it clearer, WDYT?

kmoscoe: (The original naming I used was actually misleading.)

kmoscoe: BTW, the CMS keeps those 3 medical conditions as separate metrics. So I think they're not being compared to anything else.

kmoscoe: Yeah, in fact the EDAC is really the primary measure: https://qualitynet.cms.gov/inpatient/measures/edac. The patient counts are more like "additional info". So I definitely need to rework this. Thanks a lot for pointing it out -- not sure what I was thinking!

kmoscoe: OK, I've now added new statvars for the actual time measurement. It made more sense to keep regular patients and return patients as separate classes because the duration metrics only relate to return patients.

gmechali (approved):
good from my perspective.

carojahn (approved):
discussed offline, single enum for a constraint property doesn't make sense.

PR #6583, opened 2026-08-13
[On hold] [Fixit] Place population in query context

nick-nlb:
## Issue

[b/546625182](b/546625182)

## Related PR

[6585](https://github.com/datacommonsorg/website/pull/6585) - this PR is a mitigation method to solve the query sequence itself

## Description

It was recently discovered that the demo sequence (https://datacommons.org/demo) is broken. Between the second and third queries, the place context is lost (causing the sequence to lose the initial tracked countries and fall back to Earth).

The context becomes lost because of how the fulfillment handlers propagated the place list from query to query. During the query fulfillment process, SV candidates are detected and ranked. The fulfillment loop then iterates through each candidate, passing them to a handler in which the candidate undergoes existence checks. If the data is too sparse, the candidate is discarded. 

Before this PR, handlers guarded the population of state.uttr.answerPlaces with a `rank == 0` criteria, likely under the assumption that index 0 would always represent the rendered result. However, when the 0th-ranked candidate failed existence checks and was discarded, subsequent candidates (evaluated with rank >= 1) successfully generated charts but never populated state.uttr.answerPlaces. 

Because of this the utterance finished with an empty answerPlaces list, causing follow-up queries referencing "these places" to fall back to Earth.

## Solution

This PR updates the guard to `answerPlaces` population by populating the answerPlaces list with the highest ranked candidate that passes the data existence checks supplies places to `answerPlaces`, rather than just the very first candidate (that may have been discarded).

## Why now?

As mentioned above, this problem was surfaced through the `/demo` query sequence, which was previously working. The code that populates `answerPlaces` existed in its current form at a time when the demo was working. It is likely that recent changes to data and embeddings (and thus the candidate list) has changed. While the demo sequence may have worked in the past, other sequences would have failed (i.e., this is not a new problem).

## Notes

This PR regenerates goldens (the goldens associated with sequences change with this PR as expected). It also fixes a spelling error in one of the searches.

## Testing

You can test the fix by running the demo sequence locally in `master` and then in this PR (or alternatively, in this PR and in autopush). In this PR, the place context will remain until the end of the sequence. In master, the place context is lost.

## Related PR

There is an unrelated error in the demo itself, where, when the next query in the sequence is typed in, it changes not only the query in the search, but the query as it is rendered in the search results themselves. This will be fixed in a follow-up PR.


server/lib/nl/fulfillment/ranking_across_places.py
```diff
@@ -80,9 +80,9 @@ def populate(state: PopulateState,
     chart_vars.svs = exist_svs
 
     # Maybe set Answer Places.  Be very conservative here.
-    # We only do this if this is the first chart, user requested ranking
-    # result and this is for 1 SV.
-    if (rank == 0 and len(chart_vars.svs) == 1 and
+    # We only do this if answer places have not been set yet, user requested
```
shixiao-coder, line 83: We may need a bit more context for this one. May be good to ask this in the channel to understand if there exists special requirement for rank_across_places

server/lib/nl/fulfillment/time_delta_across_places.py
```diff
@@ -151,7 +151,7 @@ def populate(state: PopulateState, chart_vars: ChartVars, places: List[Place],
                                       sv_place_latest_date=sv_place_latest_date,
                                       sv_place_facet=sv_place_facet)
 
-    if rank == 0 and field == 'abs' and ranked_places:
+    if not state.uttr.answerPlaces and ranked_places:
```
shixiao-coder, line 154: do we need the field == 'abs'?

shixiao-coder (commented):
This is fine with me, may want to double check within the Experience team chat

server/lib/nl/fulfillment/time_delta_across_places.py
nick-nlb: Between abs, pct and pc, "abs" will always be processed first, so if there is an "abs" candidate, it will take precedence and the functionality will remain unchanged. If there isn't an "abs" candidate, then the functionality does change, but in a way sustains the fallback that we are aiming for overall with this PR (i.e., a non "abs" result that contains sufficient data can be used to propagate places). 

So the removal of the 'abs' guard does align with the core philosophy of this PR (moving away from slot-based propagation of places).

PR #6584, opened 2026-08-13
Update simulated typing to only update the query in the input

nick-nlb:
## Description

During `/demo` autoplay, `AutoPlay`'s `inputQuery` callback was calling `setQuery` on every typed character. This caused the search result header (under "Data related to your research question") to mutate live with each keystroke before the query was actually submitted.

image

This PR removes `setQuery` from the `inputQuery` callback so that simulated typing only updates the search bar input via `setStoreQueryString`, leaving the active search results unchanged until the new query is submitted.


dhotchkiss (approved):
Good catch. Thanks Nick!

PR #6587, opened 2026-08-17
there's a new chart added for Goldens, seems to be data related changes

shixiao-coder:


gmechali (approved):
lgtm - pls send to data team too!

ajaits (approved):
The US Census PEP imports had a autorefresh last week and data could be from that. Seems good to proceed.

PR #6590, opened 2026-08-18
feat(nl): add dynamic graph fallback for un-cached topic and peer group members

clincoln8:
## Summary
In Data Commons Platform (DCP), custom datasets and custom topics are ingested dynamically into Spanner without generating static `topic_cache.json` files. 

Previously:
1. `_members()` in `server/lib/nl/common/topic.py` only checked `TOPIC_CACHE`. When a dynamic custom topic was queried, `get_members()` returned empty, terminating topic traversal.
2. Helper functions (`is_topic`, `is_svg`, `is_svpg`, `is_sv`) relied on hardcoded `dc/topic/` or `c/topic/` string prefixes. Custom topic DCIDs (e.g. `custom/topic/DisplacedPersons` or UN topics) were misclassified as raw Statistical Variables, bypassing `_topic_chart_vars()` and failing observation existence checks.

### Changes
1. **Dynamic Graph Fallback in `topic.py`**:
   - When a node is missing from `TOPIC_CACHE` (or returns empty), falls back to calling `_prop_val_ordered(node, prop + 'List')` to retrieve `relevantVariableList` / `memberList` from Cloud Spanner / Mixer.
   - Batch-fetches missing nodes in `_members_raw()` and `_parents_raw()` via `fetch.raw_property_values()`.
   - Iterates over all list strings in `sv_list` across provenances with set deduplication in `_prop_val_ordered()`.
2. **Schema-Driven Candidate Type Capture**:
   - Extended `VarCandidates` in `shared/lib/detected_variables.py` with `sv2types` to capture `typeOf` and `types` metadata returned from `/v2/resolve`.
   - Propagated `sv2types` through `Utterance` to explore fulfillment.
3. **Schema-Driven Classification & Feature Flag Safety**:
   - Added `enable_schema_driven_topic_resolution` feature flag (defaults to `false` in `production.json` for Base DC safety; auto-enabled on DCP container startup via `update_dcp_flags.py`).
   - Updated `is_topic`, `is_svg`, `is_svpg`, and `is_sv` to check `sv2types` and generalized `/topic/`, `/svg/`, and `/svpg/` namespace patterns when the flag is enabled.
   - Updated `explore/topic.py` to route custom topics to `_topic_chart_vars()`.
4. **Display Name Fallback**:
   - Updated `get_sv_name()` to strip namespace prefixes (`re.sub(r'^[^/]+/(topic|svpg|g)/', '', sv)`) for uncurated custom topics.
5. **Unit Tests**:
   - Added unit tests in `server/tests/lib/nl/topic_test.py` covering fallback traversal, schema-driven classification, flag toggling, and serialization roundtrips.

## Testing
- `uv run --project server --group test python3 -m pytest server/tests/lib/nl/` (all 349 passed).
- Verified live end-to-end on `calinc-dev` with image `calinc-topics-03`.

### DIFF
--- a/build/cdc_services/update_dcp_flags.py
+++ b/build/cdc_services/update_dcp_flags.py
@@ -77,12 +77,6 @@ def update_website_flags():
             ):
                 flag['enabled'] = True
 
-        # Support explicit environment variable overrides
-        if 'ENABLE_SCHEMA_DRIVEN_TOPIC_RESOLUTION' in os.environ:
-            for flag in data:
-                if flag.get('name') == 'enable_schema_driven_topic_resolution':
-                    flag['enabled'] = os.environ['ENABLE_SCHEMA_DRIVEN_TOPIC_RESOLUTION'].lower() == 'true'
-                    
         serialized = json.dumps(data, indent=2)
         with open(web_ff_path, 'w') as f:
             f.write(serialized)
--- a/build/cdc_services/update_dcp_flags.py
+++ b/build/cdc_services/update_dcp_flags.py
@@ -70,8 +70,18 @@ def update_website_flags():
             data = json.load(f)
             
         for flag in data:
-            if flag.get('name') in ('enable_nl_v2node_fetchall', 'use_v2_resolve_for_nl_search_vars'):
+            if flag.get('name') in (
+                'enable_nl_v2node_fetchall',
+                'use_v2_resolve_for_nl_search_vars',
+                'enable_schema_driven_topic_resolution'
+            ):
                 flag['enabled'] = True
+
+        # Support explicit environment variable overrides
+        if 'ENABLE_SCHEMA_DRIVEN_TOPIC_RESOLUTION' in os.environ:
+            for flag in data:
+                if flag.get('name') == 'enable_schema_driven_topic_resolution':
+                    flag['enabled'] = os.environ['ENABLE_SCHEMA_DRIVEN_TOPIC_RESOLUTION'].lower() == 'true'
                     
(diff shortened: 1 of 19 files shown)


build/cdc_services/update_dcp_flags.py
```diff
@@ -70,8 +70,18 @@ def update_website_flags():
...
-            if flag.get('name') in ('enable_nl_v2node_fetchall', 'use_v2_resolve_for_nl_search_vars'):
+            if flag.get('name') in (
+                'enable_nl_v2node_fetchall',
+                'use_v2_resolve_for_nl_search_vars',
+                'enable_schema_driven_topic_resolution'
+            ):
                 flag['enabled'] = True
+
+        # Support explicit environment variable overrides
+        if 'ENABLE_SCHEMA_DRIVEN_TOPIC_RESOLUTION' in os.environ:
```
gmechali: Why rely on this env var? Any reason not to just enable this in custom.json? 

are you going to tie this to a teraform variable?

gmechali (approved):
Thanks Christie for threading this needle, can't imagine how delicate and annoying this must have been, and to do it on such a short timeline 👏 

Adding this code now will bring in some more urgency on cleaning this up for Base DC website longer term. Hopefully post dcp public preview we can take this back to attention because it looks really really brittle throughout. :(

build/cdc_services/update_dcp_flags.py
clincoln8: thanks for catching, that was unnecessary since line 76 sets the flag to true for DCP instances.

PR #6594, opened 2026-08-18
bump threshold to 0.6 for production

shixiao-coder:
Bumping to fully using threshold of 0.6 after discussed offline with Dave and Gabe


gmechali (approved):
Thanks Xiao!

PR #6608, opened 2026-08-26
Remove nl server

shixiao-coder:


beets (commented):
Hi Xiao - could you please add description for context, and include testing notes. I'm also curious about the e2e test changes -- I would've expected those to be a no-op...

PR #6613, opened 2026-08-28
rebase goldens as discussed with Rohit

shixiao-coder:


nick-nlb:
These goldens are not correct and should not be merged in (the data underlying the goldens is under active work, after which we can finalize the goldens).

PR #6635, opened 2026-09-05
fix(frontend): remove provenance url from facet metadata enrichment payload

nick-nlb:
## Issue

[556879875](b/556879875)

## Description

Cloud Armor Rule 2000 (OWASP CRS false positive) blocked `POST /api/metadata/facets` requests because the payloads contained the provenanceUrl. This provenanceUrl was a part of the unenriched facet, but was not required by the `/api/metadata/facets` endpoint in order to fulfill its purpose of enriching the facets.

This PR does the following:
* It strips the external `provenanceUrl` URLs from the request payload, to prevent those Cloud Armor false positives.
* It improves the fallback display in the event that a facet enrichment request does fail, displaying unenriched facet information rather than blank lines.
* It adds tests for this functionality.

## Note

The diff is quite large but most of this is tests. Functional changes make up a small portion of the lines.

### Enriched facets (successful call)

image

### Unenriched facets (simulated unsuccessful call)

image

## Next Steps

In Autopush or Staging, we should validate that this solves the initial Cloud Armor false positive by enabling the rule and verifying that these calls still resolve successfully.

Once that has been confirmed, and once this is in production, the rule can be re-enabled in production.

## Codacy

The codacy complaint (a long line) is fixed, but the codacy report and error below is staled. The issue doesn't exist anymore (and the actual report is showing "Up to standards" if you drill into the codacy report).

### DIFF
--- a/static/js/shared/facet_selector/facet_option_content.test.tsx
+++ b/static/js/shared/facet_selector/facet_option_content.test.tsx
@@ -1,0 +1,201 @@
+/**
+ * Copyright 2026 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { ThemeProvider } from "@emotion/react";
+import { render, RenderResult, screen } from "@testing-library/react";
+import React from "react";
+
+import theme from "../../theme/theme";
+import { FacetOptionContent } from "./facet_option_content";
+
+describe("FacetOptionContent", () => {
+  const renderWithTheme = (ui: React.ReactElement): RenderResult =>
+    render(<ThemeProvider theme={theme}>{ui}</ThemeProvider>);
+
+  // Test: Primary title derivation from provenanceName and no detail pollution.
+  // Situation: Facet has provenanceName and provenanceId, but no sourceName or importName.
+  // Expectation: Title displays provenanceName; the raw provenanceId segment is not rendered as a detail item.
+  it("renders provenanceName as title and does not leak provenanceId segment into details", () => {
+    renderWithTheme(
+      <FacetOptionContent
+        metadata={{
+          provenanceName: "U.S. Census Bureau",
+          provenanceId: "dc/base/Census",
+        }}
+      />
+    );
+
+    expect(screen.getByText("U.S. Census Bureau")).toBeTruthy();
+    expect(screen.queryByText("Census")).toBeNull();
+  });
+
+  // Test: Fallback to displayName.
+  // Situation: Facet lacks provenanceName, but displayName is supplied.
+  // Expectation: Title displays displayName.
+  it("renders displayName as title when provenanceName is absent", () => {
+    renderWithTheme(
+      <FacetOptionContent
+        displayName="Custom Display Source"
+        metadata={{
+          provenanceId: "dc/base/Census",
+        }}
+      />
+    );
+
+    expect(screen.getByText("Custom Display Source")).toBeTruthy();
+  });
+
+  // Test: Fallback to sourceName.
+  // Situation: Facet has sourceName but no provenanceName or displayName.
+  // Expectation: Title displays sourceName.
+  it("renders sourceName as title when provenanceName is absent", () => {
+    renderWithTheme(
+      <FacetOptionContent
+        metadata={{
+          sourceName: "Census.gov",
+          provenanceId: "dc/base/Census",
+        }}
+      />
+    );
+
+    expect(screen.getByText("Census.gov")).toBeTruthy();
+  });
+
+  // Test: Precedence of importName over provenanceId segment.
+  // Situation: Facet has importName and provenanceId, but no provenanceName or sourceName.
+  // Expectation: Title displays importName.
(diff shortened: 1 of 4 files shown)


static/js/tools/shared/metadata/metadata_fetcher.ts
```diff
@@ -85,11 +85,20 @@ export async function fetchFacetsWithMetadata(
       ...(surface ? getSurfaceHeader(surface) : {}),
     };
 
+    const sanitizedFacets: FacetResponse = {};
+    for (const [sv, svFacets] of Object.entries(facets)) {
+      sanitizedFacets[sv] = {};
+      for (const [facetId, facet] of Object.entries(svFacets)) {
+        sanitizedFacets[sv][facetId] = { ...facet };
+        delete sanitizedFacets[sv][facetId].provenanceUrl;
```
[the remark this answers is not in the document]
nick-nlb: This can't be done the way suggested above given the projects strict linting set-up (provenanceUrl being a signed a value and it never being used). "Can be slow" might be relevant in a hot loop, but is not an issue here.

static/js/shared/facet_selector/facet_option_content.tsx
```diff
@@ -185,13 +193,22 @@ export function FacetOptionContent({
...
+        {(metadata.measurementMethodDescription ||
+          metadata.measurementMethod) &&
+          primaryTitle !==
+            (metadata.measurementMethodDescription ||
+              metadata.measurementMethod) && (
+            <li>
+              {metadata.measurementMethodDescription ||
+                metadata.measurementMethod}
+            </li>
+          )}
```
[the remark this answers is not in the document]
nick-nlb: I agree with the criticism, but not with the solution. We have other helper variables declared locally above the return, and this would fit much more naturally as one of those.

static/js/tools/shared/metadata/metadata_fetcher.ts
clincoln8, line 93: nit: can we document / add a comment about why we're explicitly stripping provenanceUrl here for future reference?

static/js/shared/facet_selector/facet_option_content.tsx
```diff
@@ -185,13 +196,25 @@ export function FacetOptionContent({
...
```
clincoln8, line 208: Can we remove in favor of using the methodDetail helper block below? I think measurementMethod would render twice right now. Do any tests pick this up?

```suggestion
```

clincoln8 (approved):
Thanks Nick! Just make sure that the measurementMethod only appears once before submitting.

static/js/shared/facet_selector/facet_option_content.tsx
dwnoble: This block and the `methodDetail` block right below it render the same value, so the measurement method shows up twice whenever it is set. Looks like 199cc893b added the `methodDetail` version without removing this one.

I rendered it locally with `{provenanceName, sourceName, measurementMethod, measurementMethodDescription}` and got detail items `["Census.gov", "Census PEP Survey", "Census PEP Survey"]`. This is the normal enriched path, since `enrich_facets` sets `measurementMethodDescription` whenever `measurementMethod` is present, so it is not limited to the failure case.

static/js/shared/facet_selector/facet_option_content.tsx
```diff
@@ -117,6 +125,9 @@ export function FacetOptionContent({
         : humanizedPeriod;
   }
 
+  const methodDetail =
+    metadata.measurementMethodDescription || metadata.measurementMethod;
```
dwnoble: Falling back to the raw `measurementMethod` might work against `MEASUREMENT_METHODS_SUPPRESSION_PROVENANCES` in `server/routes/shared_api/metadata.py`: the server deliberately omits `measurementMethodDescription` for those provenances because the method is considered flawed, but the facet still carries `measurementMethod`, so we would now display it.

Since the component cannot tell "suppressed" from "enrichment failed", is it worth marking unenriched facets in the fetcher, or dropping the suppressed `measurementMethod` server side? Fine to leave as is if that suppression list is effectively dead, since it compares against `provenanceName` while `WikipediaStatsData` looks like an import name.

static/js/shared/facet_selector/facet_option_content.test.tsx
```diff
@@ -0,0 +1,201 @@
...
+        metadata={{
+          measurementMethod: "CensusSurvey",
+        }}
+      />
+    );
+
+    expect(screen.getByText("CensusSurvey")).toBeTruthy();
+    const detailItems = container.querySelectorAll("ul li");
+    expect(detailItems.length).toBe(0);
+  });
```
dwnoble: This case passes with the duplicate above because `primaryTitle` equals the method, so both blocks are skipped. A case where the description differs from the title, asserting the full detail list rather than just presence, would have caught it.

static/js/tools/shared/metadata/metadata_fetcher.ts
```diff
@@ -85,11 +85,20 @@ export async function fetchFacetsWithMetadata(
...
+      }
+    }
```
dwnoble: Now that we merge against the local facets rather than relying on the echoed payload, the request only needs the fields the endpoint actually reads: `importName`, `provenanceId`, `measurementMethod`, `unit`. Building the payload from that allowlist instead of deleting `provenanceUrl` would keep any future free-text field on `StatMetadata` from reintroducing the same false positive. Not blocking.

dwnoble (approved):

static/js/tools/shared/metadata/metadata_fetcher.ts
nick-nlb: Done!

static/js/shared/facet_selector/facet_option_content.tsx
nick-nlb: Ack! I had already done that but hadn't committed and pushed it somehow. It's now pushed (and I added in a test that would have caught the issue you mentioned). Thank you for catching that.

